PR #1137, opened 2024-04-17
feat: token version

peterpeterparker:
## Motivation

Because the UI/UX of custom tokens revolves around enabling and disabling tokens, I think it is worth adding a timestamp (UPDATE: replaced by a version counter following review) to the information to prevent use cases where the user operates on multiple devices simultaneously and would overwrite changes made on one device with those from another.

It may be a not-too-frequent scenario but, given that we are developing this new feature which is not yet live, I think it is the right time to add these kinds of assertions.

That's why this PR adds a `version: Option<u64>` to the existing `UserTokens` and `CustomTokens`.

## Notes

In this PR, we are adding new information within a struct of stable memory that already exists on mainnet but, given that we introduce it as an option, it should be backward compatible. There is also a test about it.

Given that the backend sets the version when the entity is saved, it would be handy to return the new entity when setter and add functions are called. However, the Oisy frontend does not currently require it as we are reloading all the user tokens after changes. Therefore, this PR does not change those functions' returns.

For backwards compatibility, I did not modified the existing `remove_from_user_token` function which accept an ID.

Did files will be updated in a separate PR that also adapts the frontend code.


src/shared/src/types.rs
```diff
@@ -28,8 +28,18 @@ pub mod transaction {
     }
 }
 
+pub type Timestamp = u64;
+
```
bitdivine, line 32: Is the idea that when a client requests a change, you wish to make sure that the client has an up-to-date version of the account?  Essentially the timestamp is being used as a mutation counter.  Is that right?  If so, I think there might be  a problem in that `time()` returns the same value within a block, so multiple changes within a single block may all store the same time.  At least theoretically timestamps are not adequate, although they may be fine in practice.  It would be sounder to use a change counter.  But let's make sure I understand the goal first. :-D

src/backend/tests/it/upgrade.rs
```diff
@@ -74,24 +102,72 @@ fn test_upgrade_add_user_token() {
...
+    let results = update_call::<Vec<UserToken>>(&pic_setup, caller, "list_user_tokens", ());
 
     assert!(results.is_ok());
 
-    assert_tokens_eq(results.unwrap(), expected_tokens);
-}
\ No newline at end of file
+    let update_token: UserToken = UserToken {
+        symbol: Some("Updated".to_string()),
+        ..results.unwrap().get(0).unwrap().clone()
```
bitdivine, line 155:
```suggestion
        ..results.unwrap().swap_remove(0)
```
avoids a clone.

src/backend/tests/it/utils/assertion.rs
```diff
@@ -1,5 +1,6 @@
 use shared::types::custom_token::CustomToken;
 use shared::types::token::UserToken;
+use shared::types::TokenTimestamp;
 
 pub fn assert_tokens_eq(results_tokens: Vec<UserToken>, expected_tokens: Vec<UserToken>) {
```
bitdivine, line 5:
```suggestion
pub fn assert_tokens_eq(results_tokens: &[UserToken], expected_tokens: &[UserToken]) {
```
If you pass by reference there is no need to clone.  Also, you have more flexibility as you can use any slice, not just a vector.

src/backend/tests/it/upgrade.rs
```diff
@@ -74,24 +102,72 @@ fn test_upgrade_add_user_token() {
...
+
+    let updated_results = update_call::<Vec<UserToken>>(&pic_setup, caller, "list_user_tokens", ());
+
+    let expected_tokens: Vec<UserToken> = vec![update_token.clone()];
+
+    assert!(updated_results.is_ok());
+
+    let results_tokens = updated_results.unwrap();
+
+    assert_tokens_eq(results_tokens.clone(), expected_tokens.clone());
```
bitdivine, line 171:
```suggestion
    assert_tokens_eq(&results_tokens, &expected_tokens);
```

src/backend/tests/it/utils/assertion.rs
```diff
@@ -1,5 +1,6 @@
 use shared::types::custom_token::CustomToken;
 use shared::types::token::UserToken;
+use shared::types::TokenTimestamp;
```
bitdivine, line 4: Suggestion:

If you add these traits to UserToken:
```
    pub struct UserToken {
```
you can do `assert_eq!(&one_token_list, &another_token_list, "My lists aren't the same!")` without needing a dedicated comparison function.

If the default definition of `Eq` does not suit, it is possible to define a custom one.

src/backend/tests/it/utils/assertion.rs
```diff
@@ -19,6 +20,25 @@ pub fn assert_custom_tokens_eq(
     assert_eq!(results_tokens.len(), expected_tokens.len());
 
     for (token, expected) in results_tokens.iter().zip(expected_tokens.iter()) {
-        assert!(expected == token);
+        assert!(expected.token == token.token);
+        assert_eq!(expected.enabled, token.enabled);
+    }
+}
+
+pub fn assert_some_tokens_timestamp<T>(tokens: Vec<T>)
```
bitdivine, line 28:
```suggestion
pub fn assert_some_tokens_timestamp<T>(tokens: &[T])
```
If you pass by ref there is no need to clone.

src/backend/tests/it/utils/assertion.rs
```diff
@@ -19,6 +20,25 @@ pub fn assert_custom_tokens_eq(
...
+where
+    T: TokenTimestamp,
+{
+    for token in tokens.iter() {
+        assert!(token.get_timestamp().is_some());
```
bitdivine, line 33:
```suggestion
        assert!(token.get_timestamp().is_some(), "Token has no timestamp: {}", token);
```
This will print which token has failed the assertion.  You will need to add the Debug trait to the type but this is generally a useful trait!

src/backend/tests/it/utils/assertion.rs
```diff
@@ -19,6 +20,25 @@ pub fn assert_custom_tokens_eq(
...
+    }
+}
+
+pub fn assert_none_tokens_timestamp<T>(tokens: Vec<T>)
+where
+    T: TokenTimestamp,
+{
+    for token in tokens.iter() {
+        assert!(token.get_timestamp().is_none());
```
bitdivine, line 42: Likewise, I would recommend adding a message to the assertion to say which token is flawed.

src/backend/tests/it/upgrade.rs
```diff
@@ -1,13 +1,40 @@
...
     CALLER, CALLER_ETH_ADDRESS, WEENUS_CONTRACT_ADDRESS, WEENUS_DECIMALS, WEENUS_SYMBOL,
 };
 use crate::utils::pocketic::{setup_with_custom_wasm, update_call, upgrade};
-use candid::Principal;
-use shared::types::token::UserToken;
+use candid::{CandidType, Deserialize, Principal};
+use lazy_static::lazy_static;
+use shared::types::token::{ChainId, UserToken};
 
 const BACKEND_V0_0_13_WASM_PATH: &str = "../../backend-v0.0.13.wasm.gz";
```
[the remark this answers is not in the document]
peterpeterparker: In another project, I recently underwent an issue which required to migrate 6 versions in a row therefore the idea to explicitely hard-wright the version number to explicitely test a particular upgrade. We can make it more generic for additional test but here, I explicitely want to test the current upgrade.

src/backend/tests/it/utils/assertion.rs
peterpeterparker: Sure we can do that in another PR. Original did not wanted to change the existing struct but, we are passed that now ;)

src/backend/tests/it/utils/assertion.rs
peterpeterparker: I should have run clippy before submitting the PR.

src/backend/tests/it/utils/assertion.rs
peterpeterparker: Fun fact, I explicitely did not added stacktrace in the test. No particular reason. Added now.

src/shared/src/types.rs
bitdivine: I would write `changecount = changecount.unwrap_or_default() + 1;`.  That would be enough.  But _if_ you would find the timestamp useful in other ways we can still look at whether block timestamp is enough.

peterpeterparker: Ah gotcha! No sure let's do that, when you wrote "change counter" I assumed you were expected something more complicated. I'll implement a `changecount`. I might name the variable `version`, if I reckon correctly in few SQL projects that was the name of the field used for such purpose. Will do and ping you when done.

peterpeterparker: i.e. I'll modify this PR.

peterpeterparker:
Printing out the token in the assertions of the test requires implementing display. I don't think that's part of this PR, so I've reverted printing out and we can do that in another PR.

```
error[E0277]: `T` doesn't implement `std::fmt::Display`
  --> src/backend/tests/it/utils/assertion.rs:33:80
   |
33 |         assert!(token.get_timestamp().is_some(), "Token has no timestamp: {}", token);
   |                                                                                ^^^^^ `T` cannot be formatted with the default formatter
   |
   = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
   = note: this error originates in the macro `$crate::const_format_args` which comes from the expansion of the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider further restricting this bound
   |
30 |     T: TokenTimestamp + std::fmt::Display,
   |                       +++++++++++++++++++

```

bitdivine (approved):
LGTM.  Thanks!

PR #1256, opened 2024-05-02
fix: prevent client to set an incorrect starting version

peterpeterparker:


peterpeterparker:
I'm merging this as we implemented it together with @bitdivine.

PR #1259, opened 2024-05-02
fix: Nodejs uses utf8 not utf-8

bitdivine:
# Motivation
Nodejs refers to UTF-8 as `utf8` rather than `utf-8`.  The list of nodejs encodings is [here](https://github.com/nodejs/node/blob/main/lib/buffer.js#L615).

De-facto this has no effect as `utf-8` is supported as an alternative: https://github.com/nodejs/node/blob/main/lib/buffer.js#L717 

Meh, probably not worth changing.

# Changes
- For `readFile` and `writeFile` use `utf8` rather than `utf-8`.

# Tests
Existing tests should suffice.

### DIFF
--- a/scripts/build.csp.mjs
+++ b/scripts/build.csp.mjs
@@ -94,7 +94,7 @@ const injectLinkPreloader = (indexHtml) => {
  * 3. we use our custom script loader to load the main.js script
  */
 const extractStartScript = (htmlFile) => {
-	const indexHtml = readFileSync(htmlFile, 'utf-8');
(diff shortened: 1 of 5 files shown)


peterpeterparker (approved):
I somehow reckon that both `utf8` or `utf-8` can be provided - i.e. that NodeJS has builtin aliases. Themselves they seem to use `utf-8` sometimes ([example](https://github.com/nodejs/node/blob/08c3256527eb48374d8c43ba904e3f7d17f026a0/lib/internal/main/check_syntax.js#L36)).

That said, for sure we can use utf8, thanks!

PR #1260, opened 2024-05-06
build: Update Sns Tokens

sa-github-api:


peterpeterparker:
Catalyze Index is now up-to-date but, OpenFPL Index is out of cycles, so two changes but, only one correct.

Manually updated Catalyze in #1263 and made batch crash if one Index is out of cycles in #1262.

PR #1289, opened 2024-05-13
Created custom component SendInputAmount that has an optional Max button

AntonioVentilii-DFINITY:
# Motivation

Introduce a central `SendInputAmount` component for enhanced reuse and feature integration, such as the "Max" button functionality.

# Changes

- Created the `SendInputAmount` component.
- Created the `MaxButton` sub-component.
- Created unittest for `SendInputAmount`.

# Tests

The test `SendInputAmount.spec.ts` was run successfully.


src/frontend/src/lib/components/send/SendInputAmount.svelte
```diff
@@ -0,0 +1,62 @@
...
+	export let calculateMax: (() => number | undefined) | null = null;
+
+	let error: Error | undefined;
+
+	let onMax = () => {
+		amount = calculateMax?.();
+	};
+
+	const validate = () => {
+		error = undefined;
```
peterpeterparker, line 24: nth: I would set `error = undefined` only if required. That way you avoid a brief rerender if the error is set, then undefined, then set again.

Long story short:

```
const validate = () => {
   if (invalidAmount(amount)) {
        error = undefined;
        return;
   }
   
   ...
}
```

src/frontend/src/lib/components/send/SendInputAmount.svelte
```diff
@@ -0,0 +1,62 @@
...
+<Input
+	name="amount"
+	inputType="currency"
+	required
+	bind:value={amount}
+	decimals={tokenDecimals}
+	{placeholder}
+	spellcheck={false}
+>
+	<svelte:component
```
peterpeterparker, line 53: I would personally use a fragment here.

```
<svelte:fragment slot="inner-end">
    {#if nonNullish(calculateMax)}
        <MaxButton on:click={onMax} />
     {/if}
</svelte:fragment>
```

That way I would not create a `svelte:component` with `this=null`.

But it seems to work too, so just sharing the pattern in case you like it.

src/frontend/src/tests/components/send/SendInputAmount.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+		const input: HTMLInputElement | null = container.querySelector('input');
+		expect(input?.value).toBe(props.amount.toString());
+	});
+
+	it('should not render a max button if calculateMax is not provided', () => {
+		const { calculateMax: _, ...newProps } = props;
+		const { container } = render(SendInputAmount, { props: newProps });
+
+		const button: HTMLButtonElement | null = container.querySelector('button');
```
peterpeterparker, line 24: this test would be more future prone if we use the `data-tid` selector instead of querying for the existance of a button

ditto for other tests

src/frontend/src/tests/components/send/SendInputAmount.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+		expect(button).not.toBeNull();
+		expect(getByText(en.core.text.max)).toBeInTheDocument();
+	});
+
+	it('should trigger max value', async () => {
+		const { container } = render(SendInputAmount, { props });
+
+		const button: HTMLButtonElement = container.querySelector('button') as HTMLButtonElement;
+		if (!button) throw new Error('Button not found');
```
peterpeterparker, line 40: if interesting, we have a utils called `assertNonNullish` which assert that something is not null or undefined, throw and error and inform typescript about the fact that the element is valid

```
assertNonNullish(button);
```

tsconfig.spec.json
```diff
@@ -1,20 +1,38 @@
 {
-	"extends": "./tsconfig.json",
+	"extends": "./.svelte-kit/tsconfig.json",
```
peterpeterparker, line 2: can you please provide configuration - `tsconfig` and vite config -changes in a separate PR.

peterpeterparker (commented):
Thanks! Few minor comments.

tsconfig.spec.json
AntonioVentilii-DFINITY: i actually need it for the test, otherwise it would not run correctly

src/frontend/src/tests/components/send/SendInputAmount.spec.ts
AntonioVentilii-DFINITY: Solved

src/frontend/src/lib/components/send/SendInputAmount.svelte
AntonioVentilii-DFINITY: agreed

src/frontend/src/lib/components/common/MaxButton.svelte
```diff
@@ -0,0 +1,17 @@
+<script lang="ts">
+	import { i18n } from '$lib/stores/i18n.store';
+
+	export let disabled = false;
+</script>
+
+<button data-tid="max-button" on:click|preventDefault class="text-blue font-medium" {disabled}>
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: removed

src/frontend/src/lib/components/common/MaxButton.svelte
```diff
@@ -0,0 +1,17 @@
...
+	{$i18n.core.text.max}
+</button>
+
+<style lang="scss">
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: removed

peterpeterparker:
Reviewed here but, ultimately merged in another PR #1293.

PR #1292, opened 2024-05-14
test: setup jsdom and testing-library

peterpeterparker:
# Motivation

Setup vitest to default to jsdom and add testling-library to support testing Svelte components

# Changes

- add required dev libraries
- setup jsdom and plugin in vitest config
- add ts config and setup file to load `jest-dom`


vitest.config.ts
```diff
@@ -39,6 +40,7 @@ export default defineConfig(
 			]
 		},
 		test: {
+			environment: 'jsdom',
 			globals: true,
 			watch: false,
 			silent: true
```
AntonioVentilii-DFINITY, line 46: The only doubt I would have is to add here `setupFiles: ['./vitest.setup.ts']` and create the file `vitest.setup.ts` that inside has simply:

```typescript
import '@testing-library/jest-dom';
```

The reason is because there are some functions, like `expect(element).toBeInTheDocument()` that belongs to it, so we would need to import it in each test module for the components.
Not an issue, just a redundancy I think.

AntonioVentilii-DFINITY (commented):
See small doubt about `vitest.config.ts`

vitest.config.ts
peterpeterparker: Good point!

There seems to be no other path than having a separated setup files, so added it to this PR and also added only the necessary to the TS config.

AntonioVentilii-DFINITY (approved):

PR #1294, opened 2024-05-14
refactor: modified ETH Amount Input component to use SendInputAmount generic component

AntonioVentilii-DFINITY:
# Motivation

Components `SendAmount` should use the generic component `SendInputAmount` that provides the Max button optionality.

# Changes

- Refactored component and its parents and peers to use the bound `fee` variable.
- Modified the validation function to return an `Error` object or `undefined`.
- Created a `calculateMax` function for each component, using function `getMaxTransactionAmount` from `token.utils.ts` file.
- Created test for `functiongetMaxTransactionAmount`.

# Tests

Manual test in the local development web-interface to confirm the correct behavior. Unittest for `functiongetMaxTransactionAmount`.

PR #1299, opened 2024-05-16
feat: add 25% to amount to calculate fee

peterpeterparker:
# Motivation

So this morning I was looking to transfer some ERC20 transactions for PR [1294](https://github.com/dfinity/oisy-wallet/pull/1294#discussion_r1602609249) but, none of my transactions were successful. Root cause of the problem was the gas fee being entirely consumed.

A few weeks ago the cross-chain team recommended us to add a 10% to the fee as buffer, so I tried to implement this solution but, even with this amount it falled close enough.

That's why this PR adds 25% to the calculation of the ERC20 fees.

# Changes

- in `getErc20FeeData` adds a 25% to the fee result

# Test

Example of transaction that ended in error without enough gas:

https://sepolia.etherscan.io/tx/0x5b76eedfb25b6a11ad72a2b882ac624eec612c38fc57eb676fdc1e510b4429ba
https://sepolia.etherscan.io/tx/0x2dc1183471247604cf979d46d773ce21343c2db82ec15af7fc374898df9985d3

Example of transaction that succeeded with this PR:

https://sepolia.etherscan.io/tx/0xd0c119d94493c1f6cda5011414733722a7b207f68aa265219067192c820d078e

### DIFF
--- a/src/frontend/src/eth/components/fee/FeeContext.svelte
+++ b/src/frontend/src/eth/components/fee/FeeContext.svelte
@@ -28,7 +28,6 @@
 	} from '$icp-eth/utils/cketh.utils';
 	import { ckEthMinterInfoStore } from '$icp-eth/stores/cketh.store';
 	import type { Token } from '$lib/types/token';
-	import { amountForFeePlusTenPercent } from '$eth/utils/send.utils';
 
 	export let observe: boolean;
 	export let destination = '';
@@ -76,7 +75,7 @@
 
 			const erc20GasFeeParams = {
 				contract: $sendToken as Erc20Token,
-				amount: amountForFeePlusTenPercent(amount),
+				amount: parseToken({ value: `${amount ?? '1'}` }),
 				sourceNetwork,
 				...params
 			};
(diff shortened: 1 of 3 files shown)


AntonioVentilii-DFINITY:
I was looking that in the code we use Infura to get the estimate.

However, I recall that Etherscan is a more reliable source and even provides history of recent Gas Fee through the [Gas Tracker API](https://docs.etherscan.io/api-endpoints/gas-tracker). That could be leveraged at some point to have a better buffer.

Do you think it is worth to leave there a `TODO` so that at some point we switch to a more _real-time_ buffer and remove the hard-coded one?

AntonioVentilii-DFINITY (commented):
I was looking that in the code we use Infura to get the estimate.

However, I recall that Etherscan is a more reliable source and even provides history of recent Gas Fee through the [Gas Tracker API](https://docs.etherscan.io/api-endpoints/gas-tracker). That could be leveraged at some point to have a better buffer.

Do you think it is worth to leave there a `TODO` so that at some point we switch to a more _real-time_ buffer and remove the hard-coded one?

peterpeterparker:
Your input is excellent! 👍 

To share some context, I'm not sure if we had it explicitly in the Roadmap, but generally speaking, we are aware that we should improve the topic fee. This includes improving its estimation and the related UX, such as allowing the user to modify the fees or replay transactions that fail.

We can consider it within this general story or even separately. So instead of a TODO, I suggest that we create a JIRA task. Even if your input is probably more accurate, we should validate the hypothesis. Would you like to open the JIRA task yourself or should I do so?

Meanwhile, unless you think this small improvement is not a good idea, I would merge this PR. It's still better than failing.

AntonioVentilii-DFINITY (approved):

PR #1300, opened 2024-05-16
fix: classified MaxButton as non-submit button type to prevent being triggered on keydown

AntonioVentilii-DFINITY:
# Motivation

Whenever the Send form was triggered with a `keydown` event (consequentially triggering the `submit` event), the MaxButton was triggered too, overwriting the current amount in the Input component.

# Changes

Change the type of button of MaxButton component to be non-submit.

# Tests

Manual testing in local development deployment.


peterpeterparker (approved):
Thx!

PR #1302, opened 2024-05-17
fix: included checks for IC and ICRC tokens in function to calculate max amount

AntonioVentilii-DFINITY:
# Motivation

The function `getMaxTransactionAmount` was checking only if the token was **ETH** supported. That would have led to wrong calculations for **IC** and **ICRC** tokens.

# Changes

Created checks for supported **IC** and **ICRC** tokens, similar to **ETH**.

# To-Do

When we will improve the input `fee`, we should improve the function to receive more token fee data (such as `feeTokenId`) and compare with the balance token, instead of having a series of checks.


peterpeterparker (commented):
As I told you few times, I think we need two distinctive functions.

One for ETH and for IC.

This because on IC there are probably additional logic (fees) to deduct notably for ckBTC and ckETH conversion.

There I think we should not merge this PR.

peterpeterparker:
I double checked NNS-dapp, they do not removes additional fees so, let's do not remove additional fees as well for now in ckBTC and ckETH but, this seems not precise enough.

src/frontend/src/icp/utils/ic.utils.ts
```diff
@@ -0,0 +1,4 @@
+import { ICP_TOKEN_ID } from '$env/tokens.env';
+import type { TokenId } from '$lib/types/token';
+
+export const isSupportedIcTokenId = (tokenId: TokenId): boolean => ICP_TOKEN_ID === tokenId;
```
peterpeterparker, line 4: You can remove this. Generally speaking we should already have enough selector in place.

src/frontend/src/icp/utils/icrc.utils.ts
```diff
@@ -131,3 +133,6 @@ export const buildIcrcCustomTokenMetadataPseudoResponse = ({
 		...(nonNullish(icon) ? [icon] : [])
 	];
 };
+
+export const isSupportedIcrcTokenId = (tokenId: TokenId): boolean =>
```
peterpeterparker, line 137: Ditto.

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -22,7 +24,14 @@ export const getMaxTransactionAmount = ({
 	tokenDecimals: number;
 	tokenId: TokenId;
 }): number => {
-	if (isSupportedEthTokenId(tokenId)) {
+	// TODO: When we will improve the input fee, we should improve the function to receive more token
+	// fee data (such as feeTokenId) and compare with the balance token, instead of having a series
+	// of checks.
+	if (
+		isSupportedEthTokenId(tokenId) ||
```
peterpeterparker, line 31: Instead of `tokenId` you can use and pass `tokenStandard`. That way you can do:

```
if (tokenStandard === 'erc20') {
   return Math.max(Number(balance - fee), 0) / 10 ** tokenDecimals;
```

or if you want to inline:

```
return Math.max(Number(balance - (tokenStandard === "erc20" ? fee : 0n)), 0) / 10 ** tokenDecimals
```

peterpeterparker (commented):
Thanks. Few suggestions.

src/frontend/src/lib/utils/token.utils.ts
AntonioVentilii-DFINITY: even better! thanks!

src/frontend/src/icp/utils/ic.utils.ts
AntonioVentilii-DFINITY: Cool! I will look them up, where to find them

AntonioVentilii-DFINITY:
I will create a different PR with a different topic, to be cleaner

PR #1308, opened 2024-05-20
fix: early click on max UI race condition

peterpeterparker:
# Motivation

Fees are not instantly calculated, but the "Max" button is instantly available. This can lead to a UI race condition where the user clicks on the "Max" button early, which calculates the maximum amount without considering the fees. Milliseconds later, an error will inform the user that the balance is insufficient to cover the amount because the fees are ultimately estimated.

# Changes

- `MaxButton` always displayed in send amount input but disabled if no function is provided
- Provide `calculateMax` only if fees are loaded

### DIFF
--- a/src/frontend/src/eth/components/send/SendAmount.svelte
+++ b/src/frontend/src/eth/components/send/SendAmount.svelte
@@ -55,14 +55,16 @@
 		}
 	};
 
-	$: calculateMax = (): number => {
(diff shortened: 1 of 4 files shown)


AntonioVentilii-DFINITY (approved):

peterpeterparker:
Closed and reopened to try to restart jobs.

PR #1317, opened 2024-05-21
feat: adapted component IcSendAmout to use the new common component SendInputAmount

AntonioVentilii-DFINITY:
# Motivation

The input component for the amount of IC tokens should use the same common component of its ETH counterparty: `SendInputAmount`.

# Changes

- Adapted `IcSendAmount` to have the proper standard validation function.
- Created a function to pass to `SendInputAmount` to calculate the max amount possible.
- Since the fee for IC and ICRC tokens is fixed, there is no need to dynamically refresh the amount nor the fee provided to the input.

# Note

Unrelated to this PR, I included a comment on unused variable `triggerCalculateMax` in `SendInputAmount` and disabled the Lint rule _no-unused-vars_ only for that line.


src/frontend/src/lib/components/send/SendInputAmount.svelte
```diff
@@ -22,6 +22,8 @@
 		amount = calculateMax?.();
 	};
 
+	// This function is used for test scopes to forcibly trigger the onMax function.
```
peterpeterparker, line 25: The function `triggerCalculateMax` is used in `SendAmount` component.

src/frontend/src/icp/components/send/IcSendAmount.svelte
```diff
@@ -27,85 +31,70 @@
 	let fee: bigint | undefined;
 	$: fee = ($token as IcToken).fee;
 
-	const validate = () => {
-		if (invalidAmount(amount)) {
-			amountError = undefined;
-			return;
-		}
-
+	$: customValidate = (userAmount: BigNumber) => {
```
peterpeterparker, line 34: Should we provide the return type here? If yes, in another PR given that it's the same omission in `SendAmount.svelte`

```
$: customValidate = (userAmount: BigNumber): Error | undefined => {
```

peterpeterparker (approved):
LGTM, thanks.

Please just have a look at the comment before merging the PR.

src/frontend/src/lib/components/send/SendInputAmount.svelte
AntonioVentilii-DFINITY: will remove the comment

src/frontend/src/icp/components/send/IcSendAmount.svelte
AntonioVentilii-DFINITY: right! will do

peterpeterparker:
FYI I merged the PR assuming you just did not hit the "merge" button @AntonioVentilii-DFINITY.

PR #1324, opened 2024-05-22
feat: addition of QR button

AntonioVentilii-DFINITY:
# Motivation

We would like to have an icon on the address input that opens the QR scanner and returns destination address and, possibly, the amount.

# Changes

Created QR button common component.


src/frontend/src/lib/components/common/QRButton.svelte
```diff
@@ -0,0 +1,7 @@
+<script lang="ts">
+	import { IconQRCodeScanner } from '@dfinity/gix-components';
+</script>
+
+<button type="button" data-tid="qr-code-scanner-button" class="icon-only" on:click|preventDefault>
```
peterpeterparker, line 5: - the accessibility labels are missing
- do we need the data-tid?
- do we need the icon-conly class?
- do we need preventDefault given that the button is of type button?

AntonioVentilii-DFINITY: - accessibility: will add it
- data-tid: I would leave it, for future E2E testing scopes
- icon-only: not necessary, but it is always good to leave the class definition no? I can remove them if you want
- preventDefault: I would leave it there, because it may happen the same behavior of the Max button, where a keyboard event could trigger it involuntarily

src/frontend/src/lib/components/common/QRButton.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+	import { IconQRCodeScanner } from '@dfinity/gix-components';
+</script>
+
+<button
+	type="button"
+	data-tid="qr-code-scanner-button"
+	on:click|preventDefault
+	aria-label="Scan QR code"
```
peterpeterparker, line 9: It needs an i18n key

src/frontend/src/lib/components/common/QRButton.svelte
peterpeterparker: - accessibility thx
- data-tid ok
- icon-only is not a class of oisy so let's remove it
- preventDefault ok

src/frontend/src/lib/components/common/QRButton.svelte
peterpeterparker: also "Scan QR code" seems a bit generic no as call to action?

peterpeterparker (approved):
LGTM, thx

PR #1329, opened 2024-05-22
feat: created utility function to decode QR code

AntonioVentilii-DFINITY:
# Motivation

As first step to creating the QR code modal, we need to set up a function that decode a possible correct QR code and manages each case.

# Changes

- New `decodeQrCode` function in utils that checks for possible errors and token compatibility.
- Tests of such function that include the creation of a mock Token object.

# Tests

Test `qr-code.utils.spec.ts` was correctly completed.
**Note**: we mock the return of third-party function `decodePayment` of library [@dfinity/ledger-icrc](https://github.com/dfinity/ic-js/tree/main/packages/ledger-icrc), since the testing of such function is already done there.

### DIFF
--- a/src/frontend/src/lib/types/qr-code.ts
+++ b/src/frontend/src/lib/types/qr-code.ts
@@ -1,8 +1,8 @@
-export type QrResult = 'success' | 'cancelled' | 'token_incompatible';
+export type QrStatus = 'success' | 'cancelled' | 'token_incompatible';
 
 export type QrResponse = {
-	result: QrResult;
-	identifier?: string;
+	status: QrStatus;
+	destination?: string;
 	token?: string;
 	amount?: number;
 };
--- a/src/frontend/src/lib/types/qr-code.ts
+++ b/src/frontend/src/lib/types/qr-code.ts
@@ -1,0 +1,8 @@
+export type QrResult = 'success' | 'cancelled' | 'token_incompatible';
+
+export type QrResponse = {
+	result: QrResult;
+	identifier?: string;
+	token?: string;
+	amount?: number;
+};
--- a/src/frontend/src/lib/utils/qr-code.utils.ts
+++ b/src/frontend/src/lib/utils/qr-code.utils.ts
@@ -1,0 +1,35 @@
+import type { QrResponse, QrResult } from '$lib/types/qr-code';
+import { decodePayment } from '@dfinity/ledger-icrc';
+import { isNullish, nonNullish, type Token } from '@dfinity/utils';
+
+export const decodeQrCode = ({
+	result,
+	code,
+	requiredToken
+}: {
+	result: QrResult;
+	code?: string | undefined;
+	requiredToken?: Token | undefined;
+}): QrResponse => {
+	if (result !== 'success') {
+		return { result };
+	}
+
+	if (isNullish(code)) {
+		return { result: 'cancelled' };
+	}
+
+	const payment = decodePayment(code);
+
+	if (isNullish(payment)) {
+		return { result: 'success', identifier: code };
+	}
+
+	const { token, identifier, amount } = payment;
+
+	if (nonNullish(requiredToken) && token.toLowerCase() !== requiredToken.symbol.toLowerCase()) {
+		return { result: 'token_incompatible' };
+	}
+
+	return { result: 'success', identifier, token, amount };
+};
--- a/src/frontend/src/lib/utils/qr-code.utils.ts
+++ b/src/frontend/src/lib/utils/qr-code.utils.ts
@@ -1,35 +1,35 @@
-import type { QrResponse, QrResult } from '$lib/types/qr-code';
+import type { QrResponse, QrStatus } from '$lib/types/qr-code';
 import { decodePayment } from '@dfinity/ledger-icrc';
 import { isNullish, nonNullish, type Token } from '@dfinity/utils';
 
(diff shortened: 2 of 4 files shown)


src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -0,0 +1,35 @@
+import type { QrResponse, QrResult } from '$lib/types/qr-code';
+import { decodePayment } from '@dfinity/ledger-icrc';
+import { isNullish, nonNullish, type Token } from '@dfinity/utils';
+
+export const decodeQrCode = ({
+	result,
+	code,
+	requiredToken
+}: {
+	result: QrResult;
```
peterpeterparker, line 10: nitpick: `result` feels a bit unconventional maybe for a variable / parameter name?

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -0,0 +1,35 @@
...
+	code?: string | undefined;
+	requiredToken?: Token | undefined;
```
peterpeterparker, line 12: nitpick: it's "required" but it's optional at the same time?

src/frontend/src/lib/types/qr-code.ts
```diff
@@ -0,0 +1,8 @@
+export type QrResult = 'success' | 'cancelled' | 'token_incompatible';
+
+export type QrResponse = {
+	result: QrResult;
```
peterpeterparker, line 4: we need the result in the response given that we decode it?

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -0,0 +1,35 @@
...
+}): QrResponse => {
+	if (result !== 'success') {
+		return { result };
+	}
+
+	if (isNullish(code)) {
+		return { result: 'cancelled' };
+	}
+
+	const payment = decodePayment(code);
```
peterpeterparker, line 22: can `decodePayment` throws an error? if yes, we let the consumer catch those, that's the idea?

src/frontend/src/tests/mocks/tokens.mock.ts
```diff
@@ -0,0 +1,9 @@
+import type { Token } from '@dfinity/utils';
+
+export const ICP_SYMBOL = 'ICP';
+
+export const ICP_TOKEN: Token = {
```
peterpeterparker, line 5: why not using existing env constant `ICP_TOKEN`?

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -0,0 +1,35 @@
...
+
+	if (isNullish(payment)) {
+		return { result: 'success', identifier: code };
+	}
+
+	const { token, identifier, amount } = payment;
```
peterpeterparker, line 28: Maybe more future prone?

```
const {identifier, ...rest} = payment;
```

src/frontend/src/lib/types/qr-code.ts
```diff
@@ -0,0 +1,8 @@
...
+	identifier?: string;
```
peterpeterparker, line 5: `identifier` is probably a old reference to the old ICP account idenfitier, should we already use a more generic term?

src/frontend/src/tests/mocks/tokens.mock.ts
AntonioVentilii-DFINITY: I tried at first, but it raises an issue with some other env variables that need to be fixed. For example VITE_APP_VERSION. So, to simplify the tests, it's better to mock, since we only need it as constant value to mock input and returns.

src/frontend/src/lib/types/qr-code.ts
AntonioVentilii-DFINITY: I will use `destination` since it is the standard way of defining the construction of a generic transaction QR code and it matches the rest of the code.

src/frontend/src/lib/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: According to the [source code](https://github.com/dfinity/ic-js/blob/cec0dc35454c24f24957c4190332b7719d5dd65b/packages/ledger-icrc/src/utils/payment.utils.ts#L26), there is not much error to return. The unmatched code is returned as undefined and any possible error that could come from parsing the results is already treated.

src/frontend/src/lib/types/qr-code.ts
AntonioVentilii-DFINITY: we will use it to pass possible errors and/or success situations to the components

p.s. i changed it to `status`

src/frontend/src/lib/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: i changed it to

```
const { token, identifier: destination, amount } = payment;
```

src/frontend/src/tests/mocks/tokens.mock.ts
peterpeterparker: If one day we got time, I think it would be worth a try not to mock it. It can be particularly useful if we start reusing some test for unit and E2E tests.

peterpeterparker (approved):
LGTM, thx

PR #1337, opened 2024-05-22
feat: ckErc20 estimated fee symbol ckEth instead of Eth

peterpeterparker:
# Motivation

The estimated fee for a converstion ckErc20 -> Erc20 should be display with the ckEth symbol instead of ETH.

### DIFF
--- a/src/frontend/src/icp/components/fee/EthereumEstimatedFee.svelte
+++ b/src/frontend/src/icp/components/fee/EthereumEstimatedFee.svelte
@@ -46,6 +46,11 @@
 			? maxTransactionFeePlusEthLedgerApprove + CKERC20_TO_ERC20_MAX_TRANSACTION_FEE
 			: maxTransactionFeePlusEthLedgerApprove;
 
+	let feeSymbol: string;
+	$: feeSymbol = ckEr20
+		? tokenCkEth?.symbol ?? $ckEthereumNativeToken.symbol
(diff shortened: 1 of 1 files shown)


src/frontend/src/icp/components/fee/EthereumEstimatedFee.svelte
```diff
@@ -46,6 +46,11 @@
 			? maxTransactionFeePlusEthLedgerApprove + CKERC20_TO_ERC20_MAX_TRANSACTION_FEE
 			: maxTransactionFeePlusEthLedgerApprove;
 
+	let feeSymbol: string;
+	$: feeSymbol = ckEr20
+		? tokenCkEth?.symbol ?? $ckEthereumNativeToken.symbol
```
AntonioVentilii-DFINITY, line 51: is it ok to default to `ckEthereumNativeToken` ? conceptually, if `tokenCkEth` is `nullish`, shouldn't we raise an error? can it be even possible to be `nullish` if `ckEr20` is true?

peterpeterparker: Conceptually agree it's an error but, 1 ckEth === 1 Ethereum, so actually displaying with `ckEthereumNativeToken` isn't wrong neither. Therefore I thought that preventing this edge case this way felt acceptable.

AntonioVentilii-DFINITY (approved):

PR #1339, opened 2024-05-22
feat: created QR modal component

AntonioVentilii-DFINITY:
# Motivation

For the QR code integration, a new Wizard modal was created to add a further step in the send modals that could leverage the GIX component [QR Code Reader](https://gix.design/components/qr-code-reader). It mostly replicates the code in `nns-dapp`.

# Changes

- Created component that renders `QRCodeReaderModal` and that uses `decodeQrCode` function.
- Inside the component, there is an additional `WizardStep` so that the QR Code Modal can be rendered inside the current modal it is called from.

# Tests

There could be no end-to-end test yet because the GIX component `QRCodeReaderModal` has no `testId` prop to call it and to trigger events. [PR #426](https://github.com/dfinity/gix-components/pull/426) was created and approved in gix-components repo, but we would need to wait for the next release, I imagine.


src/frontend/src/lib/components/send/QRCodeModal.svelte
```diff
@@ -0,0 +1,85 @@
...
+	};
+</script>
+
+<WizardModal
+	steps={stepsPlusQr}
+	bind:currentStep
+	bind:this={modal}
+	on:nnsClose
+	testId="qr-code-modal"
+	disablePointerEvents={currentStep?.name === 'Sending'}
```
peterpeterparker, line 79: just thinking out at loud thanks to your code: we have 'Sending' coming few times in the code base, maybe we should extract a constant at some point?

src/frontend/src/lib/i18n/en.json
```diff
@@ -240,7 +241,8 @@
 			"erc20_data_undefined": "Erc20 transaction Data cannot be undefined or null.",
 			"data_undefined": "Transaction Data cannot be undefined or null.",
 			"no_identity_calculate_fee": "No identity provided to calculate the fee for its principal.",
-			"invalid_address": "The address is invalid. Please try again with a valid address identifier."
+			"invalid_address": "The address is invalid. Please try again with a valid address identifier.",
```
peterpeterparker, line 244: I would not use "address identifier" as it relates to old ICP address identifieir. Maybe "destination" or "wallet address" instead?

peterpeterparker (commented):
Thanks. Minor comments.

src/frontend/src/lib/components/send/QRCodeModal.svelte
```diff
@@ -0,0 +1,85 @@
...
+			resolveQrCodePromise = resolve;
+		})
+			.then(({ status, code }) => {
+				if (status === 'token_incompatible') {
+					toastsError({
+						msg: { text: $i18n.send.error.incompatible_token }
+					});
+					return Promise.reject(new Error('Token incompatible'));
+				}
+				return decodeQrCode({ status, code, expectedToken });
```
peterpeterparker, line 57: Just thought at an additional question: shouldn't be this a property as well? i.e. can we decode Ethereum and IC QR code with the same decoder?

src/frontend/src/lib/components/send/QRCodeModal.svelte
AntonioVentilii-DFINITY: Yes! I wanted to talk with you about that!

But it's ok, if I do a separate PR about it? Because, the idea is to check all the places that compare steps to hard-coded strings.

src/frontend/src/lib/i18n/en.json
AntonioVentilii-DFINITY: Agreed!

AntonioVentilii-DFINITY:
```json
"invalid_destination": "The destination is invalid. Please try again with a valid wallet address or destination."
```

what do you think?

src/frontend/src/lib/components/send/QRCodeModal.svelte
AntonioVentilii-DFINITY: Yes, agreed, it is better.

So, I double-checked before writing it down and for token/destination/amount it is the same schema for both ETH and IC, but for ETH and ERC20 they have more optional parameter, that we could at some point use too (for example gasLimit).

Furthermore, for ERC20, actually I need to extract the token address from the QR code.

src/frontend/src/lib/components/send/QRCodeModal.svelte
```diff
@@ -0,0 +1,85 @@
...
+	const goToStep = (stepName: string) => {
+		const stepNumber = stepsPlusQr.findIndex(({ name }) => name === stepName);
+		modal.set(stepNumber);
+	};
+
+	let resolveQrCodePromise:
+		| (({ status, code }: { status: QrStatus; code?: string }) => void)
+		| undefined = undefined;
+
+	export const scanQrCode = async ({
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: removed for now, i think we will need it when integrated with the rest

src/frontend/src/lib/components/send/QRCodeModal.svelte
peterpeterparker: One thing I should mention is that the ICRC standard for IC qr codes will most probably slightly differs - i.e. the current implementation is more "a placeholder" kind of than the specification. So ultimately we most probably gonna need different decoder.

But, we can do that later for sure.

Here the discussed spec: https://github.com/dfinity/ICRC/issues/22

src/frontend/src/lib/i18n/en.json
peterpeterparker: Good idea!

src/frontend/src/lib/components/send/QRCodeModal.svelte
peterpeterparker: Definitely in another PR 👍

src/frontend/src/lib/components/send/QRCodeModal.svelte
```diff
@@ -0,0 +1,85 @@
...
+		{
+			name: STEP_QRCODE,
+			title: $i18n.send.text.scan_qr
+		}
+	];
+
+	let modal: WizardModal;
+
+	const goToStep = (stepName: string) => {
+		const stepNumber = stepsPlusQr.findIndex(({ name }) => name === stepName);
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Theoretically is not needed right? We are using that function just internally and the usage does not allow for anything else than the the QR code step and the current step.

src/frontend/src/lib/components/send/QRCodeModal.svelte
AntonioVentilii-DFINITY: I am creating a new PR #1346 for that

src/frontend/src/lib/components/send/QRCodeModal.svelte
```diff
@@ -0,0 +1,85 @@
...
+	} from '@dfinity/gix-components';
+	import { type Token } from '@dfinity/utils';
+	import { toastsError } from '$lib/stores/toasts.store';
+	import type { QrResponse, QrStatus } from '$lib/types/qr-code';
+	import { decodeQrCode } from '$lib/utils/qr-code.utils';
+
+	export let steps: WizardSteps;
+	export let currentStep: WizardStep | undefined = undefined;
+
+	const STEP_QRCODE = 'QRCode';
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: I was re-thinking about this:

I am not sure about being a property, because it should be just an additional feature for each of the modals (IC and ETH, and future ones). So i would just use it internally in the QRCodeModal component.

The parent component would not need it, right? The only way to trigger this modal step would be through `scanQrCode` function.

I will create an enum for that (so that it is not an hardcoded string), but I think not to make it a property.

What do you think?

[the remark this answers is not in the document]
peterpeterparker: Nevermind my comment actually I think I mixed Wizard steps and Progression steps. We can iterate after the PR if required anyway. Sorry for the noise.

src/frontend/src/lib/components/send/QRCodeModal.svelte
AntonioVentilii-DFINITY: Done with PR #1350

peterpeterparker (approved):
LGTM, thx

PR #1344, opened 2024-05-23
test: created env variables useful for running tests with token stores

AntonioVentilii-DFINITY:
# Motivation

There is a ENV variable error, when trying to run some test using the token stores, for example when:

```typescript
import { ICP_TOKEN, SEPOLIA_TOKEN } from '$env/tokens.env';
import { tokens } from '$lib/derived/tokens.derived';

```

Screenshot 2024-05-23 at 12 58 56

# Changes

In `vitest.config.ts`, we will use the same `VITE_APP_VERSION` and `VITE_DFX_NETWORK` as defined in `vite.config.ts` (for example that we take it from package.json for the version).

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,16 +1,9 @@
 import { sveltekit } from '@sveltejs/kit/vite';
 import { svelteTesting } from '@testing-library/svelte/vite';
-import { readFileSync } from 'node:fs';
-import { fileURLToPath } from 'node:url';
 import { resolve } from 'path';
-import type { UserConfig } from 'vite';
+import { type UserConfig } from 'vite';
 import { defineConfig } from 'vitest/config';
-
-const file = fileURLToPath(new URL('package.json', import.meta.url));
-const json = readFileSync(file, 'utf8');
-const { version } = JSON.parse(json);
-
-const network = process.env.DFX_NETWORK ?? 'local';
+import { defineViteReplacements } from './vite.utils';
 
 export default defineConfig(
 	(): UserConfig => ({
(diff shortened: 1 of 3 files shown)


vitest.config.ts
```diff
@@ -1,9 +1,17 @@
 import { sveltekit } from '@sveltejs/kit/vite';
 import { svelteTesting } from '@testing-library/svelte/vite';
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
 import { resolve } from 'path';
 import type { UserConfig } from 'vite';
 import { defineConfig } from 'vitest/config';
 
+const file = fileURLToPath(new URL('package.json', import.meta.url));
```
peterpeterparker, line 9: In order to avoid code duplication, I suggest to rather move the common code in existing `vite.utils.ts`, expose a const or function within the utils that returns the data and use it in both `vite.config` and `vitest.config`.

peterpeterparker: For example

```
export const defineReplacements = (): { VITE_APP_VERSION: string; VITE_DFX_NETWORK: string } => {
	const file = fileURLToPath(new URL('package.json', import.meta.url));
	const json = readFileSync(file, 'utf8');
	const { version } = JSON.parse(json);

	// npm run dev = local
	// npm run build = local
	// dfx deploy = local
	// dfx deploy --network ic = ic
	// dfx deploy --network staging = staging
	const network = process.env.DFX_NETWORK ?? 'local';

	return {
		VITE_APP_VERSION: JSON.stringify(version),
		VITE_DFX_NETWORK: JSON.stringify(network)
	};
};
```

AntonioVentilii-DFINITY: I am not completely convinced about `VITE_DFX_NETWORK` since theoretically for VITEST should be always local. Or not?

peterpeterparker (approved):
LGTM, thx

PR #1346, opened 2024-05-23
feat: created URN decoder function to treat generic URN cases

AntonioVentilii-DFINITY:
# Motivation

To improve the range of URNs that can be decoded, first of all, we created a custom `decodeUrn` function that takes into consideration the schema and the possible cases of the currently onboarded tokens:

- [IC & ICRC](https://github.com/dfinity/ICRC/issues/22)
- [ETH and ERC20](https://eips.ethereum.org/EIPS/eip-681)
- [BTC](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)

To be noted that, for the ERC20 tokens, the process is not the same as the ICRC tokens: ERC20 tokens provide their respective contract address as additional URN parameter. That means that even if the prefix is `ethereum`, to find out the correct token, we need to loop among the ERC20 token addresses.

# Changes

- Created new function `decodeUrn` that receives as input a generic URN and returns it deconstructed into its single parts, including any number of parameters.
- Created types for the URN decoder return that can have some compulsory numeric and/or string parameters; otherwise the function will try to parse the parameter value as number or return as a string if failed.
- Created URN generator function in `mocks` folder to be used in the tests.
- Created tests of function `decodeUrn` that loop among all tokens in the token store.

# Tests

To perform tests, we created function `generateUrn` in the `mocks` folder. Then, the tests are done in loop for every token in the token store. 

# To-Do

- Use the `decodeUrn` instead of `decodePayment` from package `@dfinity/ledger-icrc` and adapt the tests.
- As suggestion: instead of using our own URN generator, we should use a third-party provider, to be sure of the current usage and schemas.

### DIFF
--- a/src/frontend/src/lib/types/qr-code.ts
+++ b/src/frontend/src/lib/types/qr-code.ts
@@ -1,3 +1,5 @@
+import { z } from 'zod';
+
 export type QrStatus = 'success' | 'cancelled' | 'token_incompatible';
 
 export type QrResponse = {
@@ -7,27 +9,36 @@ export type QrResponse = {
 	amount?: number;
 };
 
-type DecodedUrnBase = {
-	prefix: string;
-	destination: string;
-	networkId?: string;
-	functionName?: string;
-};
+export const URN_NUMERIC_PARAMS = ['amount', 'value'] as const;
 
-export const urnNumericParams = ['amount', 'value'] as const;
+export const URN_STRING_PARAMS = ['address'] as const;
 
-type NumericParams = {
-	[K in (typeof urnNumericParams)[number]]?: number;
-};
+const DecodedUrnBaseSchema = z.object({
+	prefix: z.string(),
+	destination: z.string(),
+	ethereumChainId: z.string().optional(),
+	functionName: z.string().optional()
+});
 
-export const urnStringParams = ['address'] as const;
+const NumericParamsSchema = URN_NUMERIC_PARAMS.reduce(
+	(acc, param) => {
+		acc[param] = z.number().optional();
+		return acc;
+	},
+	{} as Record<(typeof URN_NUMERIC_PARAMS)[number], z.ZodNumber | z.ZodOptional<z.ZodNumber>>
+);
 
-type StringParams = {
-	[K in (typeof urnStringParams)[number]]?: string;
-};
+const StringParamsSchema = URN_STRING_PARAMS.reduce(
+	(acc, param) => {
+		acc[param] = z.string().optional();
+		return acc;
+	},
+	{} as Record<(typeof URN_STRING_PARAMS)[number], z.ZodString | z.ZodOptional<z.ZodString>>
+);
 
-export type DecodedUrn = DecodedUrnBase &
-	NumericParams &
-	StringParams & {
-		[key: string]: string | number | undefined;
-	};
+export const DecodedUrnSchema = DecodedUrnBaseSchema.extend({
+	...NumericParamsSchema,
+	...StringParamsSchema
+}).catchall(z.union([z.string(), z.number()]).optional());
+
+export type DecodedUrn = z.infer<typeof DecodedUrnSchema>;
--- a/src/frontend/src/lib/types/qr-code.ts
+++ b/src/frontend/src/lib/types/qr-code.ts
@@ -6,3 +6,28 @@ export type QrResponse = {
 	token?: string;
 	amount?: number;
 };
+
+type DecodedUrnBase = {
+	prefix: string;
+	destination: string;
+	networkId?: string;
+	functionName?: string;
+};
+
+export const urnNumericParams = ['amount', 'value'] as const;
+
+type NumericParams = {
+	[K in (typeof urnNumericParams)[number]]?: number;
+};
+
+export const urnStringParams = ['address'] as const;
+
+type StringParams = {
+	[K in (typeof urnStringParams)[number]]?: string;
+};
+
+export type DecodedUrn = DecodedUrnBase &
+	NumericParams &
+	StringParams & {
+		[key: string]: string | number | undefined;
+	};
--- a/src/frontend/src/lib/utils/qr-code.utils.ts
+++ b/src/frontend/src/lib/utils/qr-code.utils.ts
@@ -1,6 +1,7 @@
 import {
-	urnNumericParams,
-	urnStringParams,
+	DecodedUrnSchema,
+	URN_NUMERIC_PARAMS,
+	URN_STRING_PARAMS,
 	type DecodedUrn,
 	type QrResponse,
 	type QrStatus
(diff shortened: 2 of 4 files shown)


peterpeterparker (commented):
Before reviewing the PR; what does "URN" stands for? This notion is new to Oisy.

AntonioVentilii-DFINITY:
> reviewing

It is basically a type of "URL" that, in our case, is returned by a QR code for the transactions we deal with. The format and the schema have a specific definition type (you even did it in the specifications of [this issue](https://github.com/dfinity/ICRC/issues/22)).

It is just a fancy name to differentiate it from URLs and URIs. We can call it URL if you prefer, but I wanted to use what your schema and the other schemas are using.

Reference: https://en.wikipedia.org/wiki/Uniform_Resource_Name#:~:text=A%20Uniform%20Resource%20Name%20(URN,to%20exist%20or%20becomes%20unavailable.

src/frontend/src/lib/types/qr-code.ts
```diff
@@ -6,3 +6,28 @@ export type QrResponse = {
 	token?: string;
 	amount?: number;
 };
+
+type DecodedUrnBase = {
```
peterpeterparker, line 10: should we use zod to parse the decoded values to ensure those have the expected format?

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -1,7 +1,49 @@
...
+	urnNumericParams,
+	urnStringParams,
+	type DecodedUrn,
+	type QrResponse,
+	type QrStatus
+} from '$lib/types/qr-code';
 import { decodePayment } from '@dfinity/ledger-icrc';
 import { isNullish, nonNullish, type Token } from '@dfinity/utils';
 
+export const decodeUrn = (urn: string): DecodedUrn | undefined => {
```
peterpeterparker, line 11: Maybe `decodeQrCodeUrn` to make its usage explicit?

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -1,7 +1,49 @@
...
+	const regex = /^([a-zA-Z]+):([a-zA-Z0-9\-.]+)(@(\d+))?(\/([a-zA-Z]+))?(\?(.*))?$/;
```
peterpeterparker, line 12: Can you add some documentation about the source of the regex?

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -1,7 +1,49 @@
...
+	const match = urn.match(regex);
+	if (isNullish(match)) {
+		return undefined;
+	}
+
+	const [_, prefix, destination, , networkId, , functionName, , queryString] = match;
+
+	const params: { [key: string]: string | number | undefined } = {};
+
+	if (queryString) {
```
peterpeterparker, line 23: `nonNullish`

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -1,7 +1,49 @@
...
+		const queryParams = new URLSearchParams(queryString);
+		queryParams.forEach((value, key) => {
```
peterpeterparker, line 25: nitpick: would be nice to map the params in a functional way, that's generally the approach in the dapp

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -1,7 +1,49 @@
...
+	if (isNullish(match)) {
+		return undefined;
+	}
+
+	const [_, prefix, destination, , networkId, , functionName, , queryString] = match;
+
+	const params: { [key: string]: string | number | undefined } = {};
+
+	if (queryString) {
+		const queryParams = new URLSearchParams(queryString);
```
peterpeterparker, line 24: could this throw an exception and if yes, it's up to the caller to handle those?

src/frontend/src/lib/types/qr-code.ts
```diff
@@ -6,3 +6,28 @@ export type QrResponse = {
...
+	prefix: string;
+	destination: string;
+	networkId?: string;
+	functionName?: string;
+};
+
+export const urnNumericParams = ['amount', 'value'] as const;
```
peterpeterparker, line 17: I would expose constant in capital letters for consistancy wdyt?

`URN_NUMERIC_PARAMS`

src/frontend/src/lib/types/qr-code.ts
```diff
@@ -6,3 +6,28 @@ export type QrResponse = {
 	token?: string;
 	amount?: number;
 };
+
+type DecodedUrnBase = {
+	prefix: string;
+	destination: string;
+	networkId?: string;
```
peterpeterparker, line 13: We might confuse `networkId: string` with `networkId: NetworkId`, would it be worth using a slightly different name?

src/frontend/src/lib/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: right! always forget!

src/frontend/src/lib/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: on it!

src/frontend/src/lib/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: not really, it would just mean that there are not extra parameters in the URN. So it will be returned nothing as `params`

src/frontend/src/lib/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: changed

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -1,7 +1,84 @@
...
+	const parseQueryString = (qs: string): { [key: string]: string | number | undefined } => {
+		try {
+			return [...new URLSearchParams(qs).entries()].reduce(
+				(acc, entry) => ({
+					...acc,
+					...processParam(entry)
+				}),
+				{}
+			);
+		} catch (error) {
```
peterpeterparker, line 58: nitpick: `error: unknown`

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -1,7 +1,84 @@
...
+			console.error('Invalid query string:', error);
```
peterpeterparker, line 59: I understand this to mean that if the query parameters are invalid, we still decode the rest as valid. If this is correct, shouldn't we rather avoid decoding anything if we are unable to decode the query parameters as well?

I'm not sure if in real life one decodes a QR code, it's surprising or not if not everything is read. So, in a conservative approach, I would spontaneously think about discarding everything.

What do you think? Or maybe I just misunderstood?

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -1,7 +1,84 @@
...
+		prefix,
+		destination,
+		...(networkId && { networkId }),
+		...(functionName && { functionName }),
+		...params
+	};
+
+	const result = DecodedUrnSchema.safeParse(decodedUrn);
+	if (!result.success) {
+		console.error(result.error);
```
peterpeterparker, line 76: should we add a label? should we use `warn` when we expecitely ignores error?

Like `console.warn("QR code cannot be decoded", result.error)`

peterpeterparker (commented):
Looks good, just remaining minor questions

src/frontend/src/lib/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: Yes, sorry, forgot to mention, I addressed it in the last commit

peterpeterparker: Gotcha. Can you either replace the `console.error` with a comment or replace it with a `console.warn` given that we one purpose catch the error and return undefined instead?

peterpeterparker (approved):
LGTM, thanks

Just added a tiny last comment, if you can just please resolve it before merge would be appreciated.

PR #1350, opened 2024-05-24
refactor: created enums for the wizard steps related to send flow

AntonioVentilii-DFINITY:
# Motivation

The Wizard steps for the send flow were hard-coded strings in the code. We want them to be enums as to be better controlled and compared.
There is no clear advantage in creating different sets of enums for ETH and for IC, since the Wizard steps are in the end customized for each case at modal-level, but are based on the same basic flow.

# Changes

- Adding enums referring to the Wizard steps of the send flow.
- Adapt the send scripts to use such enums.

# To-Do

All the hard-coded Wizard steps should be transformed as enum. We will work on it in another PR.

### DIFF
--- a/src/frontend/src/lib/enums/progress-steps.ts
+++ b/src/frontend/src/lib/enums/progress-steps.ts
@@ -53,3 +53,9 @@ export enum ProgressStepsUpdateBalanceCkBtc {
 	RELOAD = 'reload',
 	DONE = 'done'
 }
+
+export enum ProgressStepsSendStepName {
+	SEND = 'Send',
+	REVIEW = 'Review',
+	SENDING = 'Sending'
+}
--- a/src/frontend/src/lib/enums/progress-steps.ts
+++ b/src/frontend/src/lib/enums/progress-steps.ts
@@ -53,9 +53,3 @@ export enum ProgressStepsUpdateBalanceCkBtc {
 	RELOAD = 'reload',
 	DONE = 'done'
 }
-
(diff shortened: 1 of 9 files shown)


peterpeterparker (commented):
`SendWizardStep` defines the steps of the progress component used within the wizards to display the effective action stages.

Although the steps share the same names, I think that we should better not reuse them for the wizard steps since each serves a distinctive purpose. Future steps added to one component may be entirely unrelated to the other.

In a separate PR, if the current step names are unclear, we could refactor by using a "Progress" suffix. For example, `SendWizardStep` could become `SendProgressStep`.

For the current PR, I think we should rather create a new set of enums, such as `SendWizardStepName`.

AntonioVentilii-DFINITY:
Have a look at the new commit, let me know if the name fits

src/frontend/src/lib/enums/progress-steps.ts
```diff
@@ -53,3 +53,9 @@ export enum ProgressStepsUpdateBalanceCkBtc {
 	RELOAD = 'reload',
 	DONE = 'done'
 }
+
+export enum ProgressStepsSendStepName {
```
peterpeterparker, line 57: Yeah that's the thing, this is not a "Progress" step. This is a step of the "Wizard".

So I suggest to move it to a `wizard-steps.ts` module and name it `WizardStepsSend`.

peterpeterparker (approved):
LGTM, thx

PR #1353, opened 2024-05-28
refactor: changed QRCodeModal to be normal component instead of modal

AntonioVentilii-DFINITY:
# Motivation

After a few tries, I found out that the modal created in PR #1339 is not the most efficient way to have the QR code scan. That is why we modified the component to be a normal component to be called as a step in the Send Wizard Modal.

# Changes

- `QRCodeModal` was renamed `QRCodeScan`.
- The code of the component is now the replica of a normal step of a generic Wizard Modal.

# Tests

Created small test to check if the timeout was correctly triggered.

# Note

We decided to dispatch events to navigate back, in accordance to the other steps of the Wizard Modal.


src/frontend/src/lib/components/send/QRCodeScan.svelte
```diff
@@ -0,0 +1,96 @@
...
+	import { createEventDispatcher, onDestroy, onMount } from 'svelte';
+	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
+
+	export let expectedToken: Token;
+	export let destination: string | undefined;
+	export let amount: number | undefined;
+
+	const dispatch = createEventDispatcher();
+
+	let timeoutId: NodeJS.Timeout | undefined;
```
peterpeterparker, line 18: Given that this PR aims to be a refactoring only, can you provide the feature `timeout`, which seems to be a newest thing,  and its motivation in a separate PR please?

AntonioVentilii-DFINITY: done

src/frontend/src/lib/i18n/en.json
```diff
@@ -244,7 +244,8 @@
 			"data_undefined": "Transaction Data cannot be undefined or null.",
 			"no_identity_calculate_fee": "No identity provided to calculate the fee for its principal.",
 			"invalid_destination": "The destination is invalid. Please try again with a valid wallet address or destination.",
-			"incompatible_token": "The token is incompatible. Please try again with a compatible token."
+			"incompatible_token": "The token is incompatible. Please try again with a compatible token.",
+			"qr_scan_timeout": "No QR code found. Please try again."
```
peterpeterparker, line 248: This label should as well be provided in another PR related to the timeout

src/frontend/src/tests/lib/components/send/QRCodeScan.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+const expectedToken = ICP_TOKEN;
+
+vi.mock('$lib/stores/toasts.store');
+
+describe('QRCodeReaderComponent', () => {
+	afterEach(() => {
+		vi.clearAllMocks();
+	});
+
+	it('should show timeout error on QR scan timeout', async () => {
```
peterpeterparker, line 17: ditto

PR #1354, opened 2024-05-29
refactor: changed function decodeQrCode to use interal decodeQrCodeUrn

AntonioVentilii-DFINITY:
# Motivation

Instead of using the `decodePayment` function from package `@dfinity/ledger-icrc`, we will use the function `decodeQrCodeUrn` that treats a broader spectrum of tokens and transactions.

# Changes

- Changed the decoding function in `decodeQrCode`.
- Created an internal method to match specific Ethereum tokens based on the decoded QR code..

# Tests

Modified the original test of `decodeQrCode` to adapt to the new code.

### DIFF
--- a/src/frontend/src/lib/utils/qr-code.utils.ts
+++ b/src/frontend/src/lib/utils/qr-code.utils.ts
@@ -1,3 +1,6 @@
+import { erc20Tokens } from '$eth/derived/erc20.derived';
+import { enabledEthereumTokens } from '$eth/derived/tokens.derived';
+import type { EthereumNetwork } from '$eth/types/network';
 import {
 	DecodedUrnSchema,
 	URN_NUMERIC_PARAMS,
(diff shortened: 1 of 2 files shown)


AntonioVentilii-DFINITY:
> For ICP/ICRC, I think we should continue to use `decodePayment` as this function will be reviewed/improved when the standard is accepted and because we want to use the same libraries for the IC as we and the community do.

So, do you think it is better to put separate QR decoding functions in each folder `frontend/src/icp` and `frontend/src/eth`, or shall I treat it in the common function with some IF condition?

peterpeterparker:
> So, do you think it is better to put separate QR decoding functions in each folder 

Yes, I would go for separation of concern with `frontend/src/icp` and `frontend/src/eth`

AntonioVentilii-DFINITY:
I will move this to a draft while I work separately on it with #1356

AntonioVentilii-DFINITY:
Closed this one and opened PR #1359

PR #1356, opened 2024-05-29
refactor: created two different functions for ICP and for ETH to decode the QR code

AntonioVentilii-DFINITY:
# Motivation

Since we are going to use different decoders to deconstruct the QR code URN, we are going to create separate functions for each universe.

# Changes

Moved `decodeQrCode` function in two different utils modules.

# Tests

Created a set of tests for each new function.


peterpeterparker (approved):
LGTM, thx

PR #1358, opened 2024-05-29
feat: add QRButton component in Destination Input

AntonioVentilii-DFINITY:
# Motivation

We include the `QRButton` component into the common `SendInputDestination`, to be used in the main Send Modals.

# Changes

- Included `QRButton` in the `inner-end` of the destination input.
- Added a prop function that is triggered by the button.


peterpeterparker:
Did you double check with Artem? Looks definitely good to me but, as we discussed the fact that the QR code reader can also read e.g. amount, I was wondering if it makes sense to have it within the destination field?

peterpeterparker (approved):
LGTM, thx

PR #1359, opened 2024-05-30
refactor: changed ETH function decodeQrCode to use internal decodeQrCodeUrn

AntonioVentilii-DFINITY:
# Motivation

Instead of using the `decodePayment` function from package `@dfinity/ledger-icrc`, for ETH we will use the function `decodeQrCodeUrn` that treats a broader spectrum of tokens and transactions.

# Changes

- Changed the decoding function in `decodeQrCode` for the ETH universe.
- Created an internal method to match specific Ethereum tokens based on the decoded QR code.

# Tests

Modified the original test of `decodeQrCode` to adapt to the new code.

### DIFF
--- a/src/frontend/src/eth/utils/qr-code.utils.ts
+++ b/src/frontend/src/eth/utils/qr-code.utils.ts
@@ -1,6 +1,10 @@
-import type { QrResponse, QrStatus } from '$lib/types/qr-code';
-import { decodePayment } from '@dfinity/ledger-icrc';
+import { erc20Tokens } from '$eth/derived/erc20.derived';
+import { enabledEthereumTokens } from '$eth/derived/tokens.derived';
+import type { EthereumNetwork } from '$eth/types/network';
+import { type QrResponse, type QrStatus } from '$lib/types/qr-code';
+import { decodeQrCodeUrn } from '$lib/utils/qr-code.utils';
 import { isNullish, nonNullish, type Token } from '@dfinity/utils';
+import { get } from 'svelte/store';
 
 export const decodeQrCode = ({
 	status,
@@ -19,13 +23,58 @@ export const decodeQrCode = ({
 		return { status: 'cancelled' };
 	}
 
-	const payment = decodePayment(code);
+	const payment = decodeQrCodeUrn(code);
 
 	if (isNullish(payment)) {
 		return { status: 'success', destination: code };
 	}
 
-	const { token, identifier: destination, amount } = payment;
+	const { prefix, destination, amount, address, ethereumChainId, functionName } = payment;
+
+	const normalizeChainId = (chainId: string): string => {
+		if (chainId.startsWith('0x')) {
+			return parseInt(chainId, 16).toString();
+		}
+		return chainId;
+	};
+
+	const matchEthereumToken = (
+		address: string | undefined,
+		functionName: string | undefined,
+		ethereumChainId: string | undefined
+	): string | undefined => {
+		const parsedEthereumChainId = nonNullish(ethereumChainId)
+			? normalizeChainId(ethereumChainId)
+			: 1n;
+		console.warn('parsedEthereumChainId', parsedEthereumChainId);
+
+		if (nonNullish(address) && functionName === 'transfer') {
+			const erc20TokensList = get(erc20Tokens);
+			return (
+				erc20TokensList.find(
+					(token) =>
+						token.address.toLowerCase() === address.toLowerCase() &&
+						(token.network as EthereumNetwork).chainId.toString() ===
+							parsedEthereumChainId.toString()
+				)?.symbol || undefined
+			);
+		}
+
+		const enabledEthereumTokensList = get(enabledEthereumTokens);
+		return (
+			enabledEthereumTokensList.find(
+				(token) =>
+					(token.network as EthereumNetwork).chainId.toString() === parsedEthereumChainId.toString()
+			)?.symbol || undefined
+		);
+	};
+
+	const token =
+		prefix === 'ethereum' ? matchEthereumToken(address, functionName, ethereumChainId) : undefined;
+
+	if (isNullish(token)) {
+		return { status: 'token_incompatible' };
+	}
 
 	if (nonNullish(expectedToken) && token.toLowerCase() !== expectedToken.symbol.toLowerCase()) {
 		return { status: 'token_incompatible' };
--- a/src/frontend/src/tests/eth/utils/qr-code.utils.spec.ts
+++ b/src/frontend/src/tests/eth/utils/qr-code.utils.spec.ts
@@ -1,19 +1,20 @@
-import { ICP_TOKEN } from '$env/tokens.env';
+import { SEPOLIA_TOKEN } from '$env/tokens.env';
+import type { EthereumNetwork } from '$eth/types/network';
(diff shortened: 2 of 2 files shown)


src/frontend/src/eth/utils/qr-code.utils.ts
```diff
@@ -1,6 +1,10 @@
-import type { QrResponse, QrStatus } from '$lib/types/qr-code';
-import { decodePayment } from '@dfinity/ledger-icrc';
+import { erc20Tokens } from '$eth/derived/erc20.derived';
+import { enabledEthereumTokens } from '$eth/derived/tokens.derived';
+import type { EthereumNetwork } from '$eth/types/network';
+import { type QrResponse, type QrStatus } from '$lib/types/qr-code';
+import { decodeQrCodeUrn } from '$lib/utils/qr-code.utils';
 import { isNullish, nonNullish, type Token } from '@dfinity/utils';
+import { get } from 'svelte/store';
```
peterpeterparker, line 7: If possible, please avoid using store imperatively in utils and rather pass the state or desired variable as parameter. That way the utils can be use dynamically.

src/frontend/src/eth/utils/qr-code.utils.ts
```diff
@@ -19,13 +23,58 @@ export const decodeQrCode = ({
...
 
 	if (isNullish(payment)) {
 		return { status: 'success', destination: code };
 	}
 
-	const { token, identifier: destination, amount } = payment;
+	const { prefix, destination, amount, address, ethereumChainId, functionName } = payment;
+
+	const normalizeChainId = (chainId: string): string => {
+		if (chainId.startsWith('0x')) {
```
peterpeterparker, line 35: do we have maybe already a utils for this?

src/frontend/src/eth/utils/qr-code.utils.ts
```diff
@@ -19,13 +23,58 @@ export const decodeQrCode = ({
...
+			return parseInt(chainId, 16).toString();
+		}
+		return chainId;
+	};
+
+	const matchEthereumToken = (
```
peterpeterparker, line 41: can we use an object as parameter for consistency reason?

src/frontend/src/eth/utils/qr-code.utils.ts
```diff
@@ -19,13 +23,58 @@ export const decodeQrCode = ({
...
+		address: string | undefined,
+		functionName: string | undefined,
+		ethereumChainId: string | undefined
+	): string | undefined => {
+		const parsedEthereumChainId = nonNullish(ethereumChainId)
+			? normalizeChainId(ethereumChainId)
+			: 1n;
```
peterpeterparker, line 48: wouldn't be safer to not fallack on what I guess is ethereum production and ignore or throw an error if the chain is not supported?

src/frontend/src/eth/utils/qr-code.utils.ts
```diff
@@ -19,13 +23,58 @@ export const decodeQrCode = ({
...
+		console.warn('parsedEthereumChainId', parsedEthereumChainId);
+
+		if (nonNullish(address) && functionName === 'transfer') {
+			const erc20TokensList = get(erc20Tokens);
+			return (
+				erc20TokensList.find(
+					(token) =>
+						token.address.toLowerCase() === address.toLowerCase() &&
```
peterpeterparker, line 56: Not for this PR, just thinking out at loud: we really should create a utils to compare address, I already repeated the `toLowerCase` to many times.

AntonioVentilii-DFINITY: Agreed!

src/frontend/src/eth/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: Yes, it makes sense

src/frontend/src/eth/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: you mean to parse a string to int right?

I found [`hexStringToUint8Array`](https://github.com/dfinity/ic-js/blob/0e762d9405c1f58ccfae2af6e73cab5214ed44d0/packages/utils/src/utils/arrays.utils.ts#L52) in `@dfinity/utils`.

Or you mean a `normalizeChainId` util?

peterpeterparker: I mean `normalizeChainId` and `startsWith('0x')`

src/frontend/src/eth/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: I didn't quite understand the question. You mean to pass `matchEthereumToken` as property?

peterpeterparker:
```
const matchEthereumToken = (address: string | undefined,
		functionName: string | undefined,
		ethereumChainId: string | undefined
	): string | undefined => {
```

=> (for example)

```
const matchEthereumToken = ({address, functionName, ethereumChainId}: {address: string | undefined,
		functionName: string | undefined,
		ethereumChainId: string | undefined}
	): string | undefined => {
```

src/frontend/src/eth/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: Not that I know of. I searched in both `@dfinity/utils` and `ethers.js` but I could not find something similar. Shall I create it in our utils?

src/frontend/src/eth/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: Ah! Right! I forgot sorry!

src/frontend/src/eth/utils/qr-code.utils.ts
peterpeterparker: It would be nice I think, no?

It can be extracted after we merge this PR.

src/frontend/src/eth/utils/qr-code.utils.ts
peterpeterparker: Generally speaking I would avoid starting adding deep comparison has the root feature of Oisy is and should remain to be the usage of JavaScript symbol - we are discussing a special case here.

Ok, works for me, if it can be omitted, then we do not crash an error but fallback. However, then it should not be `1n` but the chain id of the currently selected network I think - i.e. either Ethereum or Sepolia.

src/frontend/src/eth/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: will do!

src/frontend/src/eth/utils/qr-code.utils.ts
AntonioVentilii-DFINITY: perfect! changing it

AntonioVentilii-DFINITY: perfect! changing it

peterpeterparker (approved):
LGTM, thx

PR #1360, opened 2024-05-30
feat: included timeout error on component QRCodeScan

AntonioVentilii-DFINITY:
# Motivation

It is better to have a timeout for the QR code scan, so that after a certain amount of seconds, the component unmounts itself.

# Changes

- Included a `timeout` error status among the status results of a QR code scan.
- Added 30 seconds of timeout in component `QRCodeScan`, after which, it will resolve itself, returning the error as a message.

# Tests

We created a new test for the component to check the correct return triggered by the `timeout` error.

### DIFF
--- a/src/frontend/src/lib/components/send/QRCodeScan.svelte
+++ b/src/frontend/src/lib/components/send/QRCodeScan.svelte
@@ -5,7 +5,7 @@
 	import type { Token } from '@dfinity/utils';
 	import type { QrStatus } from '$lib/types/qr-code';
 	import { i18n } from '$lib/stores/i18n.store';
-	import { createEventDispatcher, onMount } from 'svelte';
+	import { createEventDispatcher, onDestroy, onMount } from 'svelte';
 	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
 
 	export let expectedToken: Token;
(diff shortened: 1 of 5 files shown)


peterpeterparker:
I'll review that one next week, if that's ok?

peterpeterparker:
> It is better to have a timeout for the QR code scan

@AntonioVentilii-DFINITY is this proposition a standard pattern in other similar wallet or e-banking app? Also 30 seconds is a standard?

We do not have such timeout in NNS dapp.

AntonioVentilii-DFINITY:
> > It is better to have a timeout for the QR code scan
> 
> @AntonioVentilii-DFINITY is this proposition a standard pattern in other similar wallet or e-banking app? Also 30 seconds is a standard?
> 
> We do not have such timeout in NNS dapp.

Not really, it is just to avoid continuing scanning if there is no intention anymore from the user, and the user does not click the "back" button.

AntonioVentilii-DFINITY:
> > Not really
> 
> In that case, I personally would not implement such a pattern. I feel like, for example, 30 seconds might be too short. Moreover, if it's not an industry standard, it can be a surprising UX for the users - i.e. I would stick to expected behavior. That said, just my two cents.

Perfect, then I Close this PR

PR #1361, opened 2024-05-30
feat: included QR Code Scan step in ETH Wizard Steps

AntonioVentilii-DFINITY:
# Motivation

The QRCodeScan component will be used as an additional step inside the Send Wizard Modal. It is better in this way, since we could keep the transitions between Wizard steps and the links between one step and the other.

```typescript
export const sendWizardSteps = (i18n: I18n): WizardSteps => [
  {
    name: WizardStepsSend.SEND,
    title: i18n.send.text.send
  },
  {
    name: WizardStepsSend.REVIEW,
    title: i18n.send.text.review
  },
  {
    name: WizardStepsSend.SENDING,
    title: i18n.send.text.sending
  ,
  {
    // New additional step for QR code scanning
    name: WizardStepsSend.QR_CODE_SCAN,
    title: i18n.send.text.scan_qr
  }
];
```

The Wizard processes the flows in an ordered way, step-by-step. However, nowhere in the code, the step `SENDING` triggers the `modal.next` function. So, there is no risk for the user to move

# Changes

Included an additional step among the Wizard steps, based on conditional of `currentStep`.


peterpeterparker (approved):
LGTM, thx

src/frontend/src/eth/components/send/SendTokenWizard.svelte
```diff
@@ -248,6 +250,8 @@
 				{/if}
 			</svelte:fragment>
 		</SendForm>
+	{:else if currentStep?.name === WizardStepsSend.QR_CODE_SCAN}
+		<QRCodeScan expectedToken={$sendToken} bind:destination bind:amount {decodeQrCode} />
```
peterpeterparker, line 254: Actually, is it correct to use `bind:` here?  Shouldn't it be just passing the props `{destination} {amount}`?

AntonioVentilii-DFINITY: I think so, since the QRCodeScan must change those amounts. Otherwise, the amount and destination will not change.

peterpeterparker: Got it!

PR #1362, opened 2024-05-30
refactor: moved IC Wizard Steps in separate file

AntonioVentilii-DFINITY:
# Motivation

We moved the Wizard Steps for IC in a separate config files, to be centralized and consistent with the ETH. This will be helpful when we will have to add the QR Code scan step.
Furthermore, it is a first step for when we will want to further refactor and unify the Wizard Steps for all the universes, avoiding repetition of code.

# Changes

- Created an IC Send Wizard Step list.
- Changed component `IcSendModal` to use the list.


peterpeterparker (approved):
LGTM, thx

PR #1364, opened 2024-05-31
refactor: moving BTC Network-related functions to replicate the same functions of ETH

AntonioVentilii-DFINITY:
# Motivation

To prepare for BTC integration, we moved some base functions about BTC network to the same file where are the similar ones for ETH, to centralize them.

# Changes

- Create env variables for the list of Bitcoin networks.
- Create  bool functions to check current Network ID.

### DIFF
--- a/src/frontend/src/icp/utils/ic-send.utils.ts
+++ b/src/frontend/src/icp/utils/ic-send.utils.ts
@@ -1,8 +1,4 @@
-import {
-	BTC_MAINNET_NETWORK_ID,
-	BTC_TESTNET_NETWORK_ID,
-	ETHEREUM_NETWORK_ID
-} from '$env/networks.env';
+import { BTC_MAINNET_NETWORK_ID, ETHEREUM_NETWORK_ID } from '$env/networks.env';
 import {
 	CKBTC_LEDGER_CANISTER_IDS,
 	CKERC20_LEDGER_CANISTER_IDS,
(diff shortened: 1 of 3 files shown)


src/frontend/src/icp/utils/ic-send.utils.ts
```diff
@@ -45,7 +41,7 @@ export const isTokenCkErc20Ledger = ({ ledgerCanisterId }: Partial<IcToken>): bo
 	nonNullish(ledgerCanisterId) && CKERC20_LEDGER_CANISTER_IDS.includes(ledgerCanisterId);
 
 export const isNetworkIdBTC = (networkId: NetworkId | undefined): boolean =>
-	nonNullish(networkId) && [BTC_MAINNET_NETWORK_ID, BTC_TESTNET_NETWORK_ID].includes(networkId);
+	nonNullish(networkId) && isNetworkIdBitcoin(networkId);
```
peterpeterparker, line 44: we do not need the `nonNullish(networkId)` check here I think, it's deferred within `isNetworkIdBitcoin(networkId)`?

so we probably also do not need `isNetworkIdBTC` which can be replaced by `isNetworkIdBitcoin` or do we want to keep both?

AntonioVentilii-DFINITY: Honestly, I would have done as you suggest, but I see that for ETH we were doing like that eheh!

In that case, I would change ETH too (in another PR), what do you think?

peterpeterparker: Did not check the code but, spontaneously I would say that the ETH related existing probably make sense and should not be modified. This because between ICP/ETH/Sepolia and Mainnet/derived network/specific network we might need the different functions but, again, did not checked the code.

AntonioVentilii-DFINITY: perfect, then i implement the simpler solution for BTC and, in case, in future, I will check ETH, on a separate PR, if it can be simplified too

peterpeterparker (approved):
LGTM, thx

PR #1373, opened 2024-06-04
feat: added code to trigger the QR Code Scan step on ICP

AntonioVentilii-DFINITY:
# Motivation

To trigger the QR Code scan step, we created specific events `icQRCodeScan` and `icQRCodeBack`.

# Changes

- Create event `icQRCodeScan` that is dispatched by the QR Code Button.
- Create event `icQRCodeBack` that is dispatched by the Cancel button in the QR Code Scan step.
- Assign them to navigate to the appropriate Wizard Step, using new function `goToWizardStep`. 

# Test
Manual testing done on local environment.

### DIFF
--- a/src/frontend/src/icp/components/send/IcSendModal.svelte
+++ b/src/frontend/src/icp/components/send/IcSendModal.svelte
@@ -44,7 +44,9 @@
 		type EthereumFeeContext as EthereumFeeContextType
 	} from '$icp/stores/ethereum-fee.store';
 	import { WizardStepsSend } from '$lib/enums/wizard-steps';
-	import { icSendWizardSteps } from '$icp/config/ic-send.config';
+	import { icSendWizardStepsWithQrCodeScan } from '$icp/config/ic-send.config';
+	import { icDecodeQrCode } from '$icp/utils/qr-code.utils';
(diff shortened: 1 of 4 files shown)


src/frontend/src/icp/components/send/IcSendModal.svelte
```diff
@@ -177,6 +179,11 @@
 	setContext<EthereumFeeContextType>(ETHEREUM_FEE_CONTEXT_KEY, {
 		store: initEthereumFeeStore()
 	});
+
+	const goToWizardStep = (stepName: WizardStepsSend) => {
```
peterpeterparker, line 183: Should we extract this in a utils to avoid duplicated code?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/icp/components/send/IcSendModal.svelte
AntonioVentilii-DFINITY: Yes, agreed! I was going to do it in the next PR eheh

AntonioVentilii-DFINITY: Changing it with PR #1378

PR #1374, opened 2024-06-04
fix: corrected ERC20 amount property, according to EIP-681

AntonioVentilii-DFINITY:
# Motivation

There was a mistake when choosing what property consider when the QR code refers to ERC-20 tokens. According to [EIP-681](https://eips.ethereum.org/EIPS/eip-681):

> Requesting payments in [ERC-20](https://eips.ethereum.org/EIPS/eip-20) tokens involves a request to call the `transfer` function of the token contract with an `address` and a `uint256` typed parameter, containing the _beneficiary address_ and the _amount in atomic units_, respectively. For example, requesting a Unicorn to address `0x8e23ee67d1332ad560396262c48ffbb01f93d052` looks as follows: `ethereum:0x89205a3a3b2a69de6dbf7f01ed13b2108b2c43e7/transfer?address=0x8e23ee67d1332ad560396262c48ffbb01f93d052&uint256=1`

# Changes

The function `decodeQrCode` was adapted to get the prop uint256 in case the QR code refers to an ERC20 token, defined by the existence of the prop `functionName`.

# Tests

Changed test of decodeQrCode to consider the correct prop.

### DIFF
--- a/src/frontend/src/eth/utils/qr-code.utils.ts
+++ b/src/frontend/src/eth/utils/qr-code.utils.ts
@@ -4,6 +4,7 @@ import { type QrResponse, type QrStatus } from '$lib/types/qr-code';
 import type { Token } from '$lib/types/token';
 import { decodeQrCodeUrn } from '$lib/utils/qr-code.utils';
 import { hexStringToUint8Array, isNullish, nonNullish } from '@dfinity/utils';
+import { formatEther } from 'ethers/lib/utils';
 
 export const decodeQrCode = ({
 	status,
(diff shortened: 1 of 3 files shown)


src/frontend/src/eth/utils/qr-code.utils.ts
```diff
@@ -97,5 +98,12 @@ export const decodeQrCode = ({
 		return { status: 'token_incompatible' };
 	}
 
+	const amount =
+		functionName === 'transfer'
+			? uint256
+			: nonNullish(value)
+				? +formatEther(value.toString())
```
peterpeterparker, line 105: - we do not use `formatEther` anywhere else in the code base, could we potentially use `formatToken` here?

- it's always 18 decimals here even for ERC20 such as USDC?

AntonioVentilii-DFINITY: - changing it;
- well, when it's a token that is a ERC20, the amount is equal to `uint256`, but, using `formatToken`, we will provide the `token.decimals` input, to avoid any doubts

peterpeterparker (approved):
LGTM,thx

AntonioVentilii-DFINITY:
This PR refers to the same issue of PR #1379

PR #1375, opened 2024-06-04
style: changed color of QRButton to be the same as MaxButton

AntonioVentilii-DFINITY:
# Motivation

The color of the QR Button will be the same as the Max Button.

# Changes

### Before

Screenshot 2024-06-04 at 13 24 50

### After

Screenshot 2024-06-04 at 13 24 01


peterpeterparker (approved):
Thx

PR #1377, opened 2024-06-04
style: changed the background of QRCodeScan to have more opacity

AntonioVentilii-DFINITY:
# Motivation

Changed to opacity of the QR Code Scan background to **60%** (value agreed with @artkorotkikh-dfinity) and using the theme color.

# Changes

### Before

Screenshot 2024-06-04 at 14 38 51

### After

Screenshot 2024-06-04 at 15 17 58

### DIFF
--- a/src/frontend/src/lib/components/send/SendQRCodeScan.svelte
+++ b/src/frontend/src/lib/components/send/SendQRCodeScan.svelte
@@ -72,7 +72,7 @@
 	};
 </script>
 
-<div class="stretch">
+<div class="stretch md:min-h-[300px]">
 	<QRCodeReader on:nnsCancel={onCancel} on:nnsQRCode={onQRCode} />
 </div>
 


src/frontend/src/lib/components/send/QRCodeScan.svelte
```diff
@@ -81,3 +81,9 @@
 		{$i18n.core.text.back}
 	</button>
 </ButtonGroup>
+
+<style>
+	.qr-code-wrapper {
+		color: rgba(0, 0, 0, 0.6) !important;
```
peterpeterparker, line 87: Please avoid using `!important` as it will inevitably lead to problems.

We can for example use following here probably `--primary-rgb: 0, 0, 0`

Or maybe something like `--primary-rgb: 50, 20, 105` if we want to use a color of the theme

AntonioVentilii-DFINITY: changed to use `--primary-rgb: 50, 20, 105`

src/frontend/src/lib/components/send/QRCodeScan.svelte
```diff
@@ -81,3 +81,10 @@
 		{$i18n.core.text.back}
 	</button>
 </ButtonGroup>
+
+<style>
+	.qr-code-wrapper {
+		--primary-rgb: 50, 20, 105;
+		color: rgba(var(--primary-rgb), 0.6);
```
peterpeterparker, line 88: this is needed? I had a quick test, using solely `--primary-rgb` seemed to work but maybe I missed something?

AntonioVentilii-DFINITY:
Closed in favor of PR #1387

PR #1379, opened 2024-06-04
fix: corrected ERC20 destination/address property, according to EIP-681

AntonioVentilii-DFINITY:
# Motivation

There was a mistake when choosing what property consider when the QR code refers to ERC-20 tokens. According to [EIP-681](https://eips.ethereum.org/EIPS/eip-681):

> Requesting payments in [ERC-20](https://eips.ethereum.org/EIPS/eip-20) tokens involves a request to call the `transfer` function of the token contract with an `address` and a `uint256` typed parameter, containing the _beneficiary address_ and the _amount in atomic units_, respectively. For example, requesting a Unicorn to address `0x8e23ee67d1332ad560396262c48ffbb01f93d052` looks as follows: `ethereum:0x89205a3a3b2a69de6dbf7f01ed13b2108b2c43e7/transfer?address=0x8e23ee67d1332ad560396262c48ffbb01f93d052&uint256=1`

# Changes

The function `decodeQrCode` was adapted to get the prop `destination` in case the QR code refers to an ERC20 token, defined by the existence of the prop `functionName`.


AntonioVentilii-DFINITY:
This PR refers to the same issue of PR #1374 .

src/frontend/src/eth/utils/qr-code.utils.ts
```diff
@@ -80,7 +94,7 @@ export const decodeQrCode = ({
 	const token =
 		prefix === 'ethereum'
 			? matchEthereumToken({
-					address,
+					tokenAddress: functionName === 'transfer' ? destinationOrTokenAddress : undefined,
```
peterpeterparker, line 97: I'm a bit confuse here, is there a use case for a non nullish `destinationOrTokenAddress` with a `functionName === 'transfer'`? or why `undefined` if not transfer?

AntonioVentilii-DFINITY: According to ETH standard, the prop that I defined as `destinationOrTokenAddress` is the first address present in the QR code and can be:
- `destination` when the QR code refers to the native token ETH  (mainnet or testnet), so it refers to the destination wallet.
- `tokenAddress` when the QR code refers to an ERC20 token, so it refers to the address of the ERC20 token.

To differentiate the two cases, there is the `functionName` parameter equal to `'transfer'` in the QR code.

So, to check if the QR code refers to an ERC20 token, we check if `functionName` is correct and we pass `destinationOrTokenAddress` to prop `tokenAddress`. If it is referring to the native token, `tokenAddress` prop shall be `undefined` (otherwise we would be incorrectly passing the destination wallet).

peterpeterparker: Thanks for the explanation. Can we document this in some way or add a comment in the code?

peterpeterparker (approved):
LGTM, thx

PR #1387, opened 2024-06-05
style: changed the background of QRCodeScan to have more opacity

AntonioVentilii-DFINITY:
# Motivation

Changed to opacity of the QR Code Scan background to **60%** (value agreed with @artkorotkikh-dfinity) and using the theme color.

# Changes

### Before

Screenshot 2024-06-04 at 14 38 51

### After

Screenshot 2024-06-04 at 15 17 58

### DIFF
--- a/src/frontend/src/lib/components/send/SendQRCodeScan.svelte
+++ b/src/frontend/src/lib/components/send/SendQRCodeScan.svelte
@@ -72,7 +72,7 @@
 	};
 </script>
 
-<div class="stretch md:min-h-[300px]">
+<div class="stretch md:min-h-[300px] qr-code-wrapper">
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/send/SendQRCodeScan.svelte
```diff
@@ -81,3 +81,10 @@
 		{$i18n.core.text.back}
 	</button>
 </ButtonGroup>
+
+<style>
```
peterpeterparker, line 85: `<style lang="scss">` maybe? that's generally what we use.

peterpeterparker (approved):
LGTM, thx

PR #1391, opened 2024-06-05
feat: include list of BTC tokens in tokens store, mirroring ETH structure

AntonioVentilii-DFINITY:
# Motivation

Preparing for the BTC onboarding, we created a list of current native Bitcoin tokens to add to the `tokens` store.

# Changes

- Create env variable to enable (or not) the BTC Mainnet (similar to ETH).
- Create store of native BTC tokens: mainnet and testnet.
- Include Bitcoin tokens store in the main `tokens` store.

# Important Note

We created a temporary environment variable `VITE_NETWORK_BITCOIN_ENABLED` that triggers the features related to Bitcoin, defaulting to `false`.
Such variable should be manually set in `.env.development` file when one would like to test locally. 

```
[...]
VITE_NETWORK_BITCOIN_ENABLED=true
```

**It will be removed once all the features are onboarded and reviewed.**

# To-Do

- [ ] Remove flag `NETWORK_BITCOIN_ENABLED` once all the Bitcoin feature are approved.


src/frontend/src/env/networks.btc.env.ts
```diff
@@ -14,3 +14,9 @@ export const BITCOIN_CANISTER_IDS: Record<MinterCanisterIdText, CanisterIdText>
 		[IC_CKBTC_MINTER_CANISTER_ID]: 'ghsi2-tqaaa-aaaan-aaaca-cai'
 	})
 };
+
+export const BTC_MAINNET_ENABLED =
+	JSON.parse(import.meta.env.VITE_BITCOIN_MAINNET ?? false) === true;
+
+export const BTC_UNIVERSE_ENABLED =
```
peterpeterparker, line 21: I don't use - on purpose - the term "universe" in Oisy. I guess you can use `NETWORK`?

src/frontend/src/env/networks.btc.env.ts
```diff
@@ -14,3 +14,9 @@ export const BITCOIN_CANISTER_IDS: Record<MinterCanisterIdText, CanisterIdText>
 		[IC_CKBTC_MINTER_CANISTER_ID]: 'ghsi2-tqaaa-aaaan-aaaca-cai'
 	})
 };
+
+export const BTC_MAINNET_ENABLED =
```
peterpeterparker, line 18: This property should probably be added to `.env.example` as well?

src/frontend/src/env/networks.btc.env.ts
AntonioVentilii-DFINITY: changing it

src/frontend/src/env/networks.btc.env.ts
AntonioVentilii-DFINITY: since it is a temporary flag to be used only in local replica until we are ok with Bitcoin features, I thought to not add it there, like we do for example for the Chain Fusion flag

but not strong opinion, if you prefer it there, I will add it

peterpeterparker (approved):
LGTM, thx

PR #1392, opened 2024-06-06
chore: removed network ID parameter in Settings page URL

AntonioVentilii-DFINITY:
# Motivation

Going through the `/settings` page, I found out that the parameter `network` in the URL is not used, nor necessary for the sub-components.

# Changes

Removed the network parameter from the `/settings` page.

### Before:
https://oisy.com/settings/?network=ETH

### After:
https://oisy.com/settings/

# Tests

Manual tests throughout the Settings page on local development env.

### DIFF
--- a/src/frontend/src/lib/components/core/Menu.svelte
+++ b/src/frontend/src/lib/components/core/Menu.svelte
@@ -6,8 +6,7 @@
 	import { OISY_REPO_URL } from '$lib/constants/oisy.constants';
 	import IconWallet from '$lib/components/icons/IconWallet.svelte';
 	import IconChevronDown from '$lib/components/icons/IconChevronDown.svelte';
-	import { networkICP, networkId } from '$lib/derived/network.derived';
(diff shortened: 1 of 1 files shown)


peterpeterparker:
I did not double check but, I would bet the parameter is required otherwise when you go back the network is lost.
e.g. ICP -> Settings -> Back -> ETH 🫠

AntonioVentilii-DFINITY:
> I did not double check but, I would bet the parameter is required otherwise when you go back the network is lost. e.g. ICP -> Settings -> Back -> ETH 🫠

Indeed, it goes back to the "default" network, that right now is ETH. I thought this could be acceptable for now, since we will not have a "network" to go back to, for the  single-page view, no?

peterpeterparker:
I think this should remains in place until it's effectively decided to implement a single-page view and how - i.e. not before.

AntonioVentilii-DFINITY:
Then closing this PR, to revisit in future.

PR #1393, opened 2024-06-06
refactor: make query parameters optional in fetchCoingecko function

AntonioVentilii-DFINITY:
# Motivation

To add additional endpoints to the Coingecko API wrapper (for example [supported_vs_currencies](https://docs.coingecko.com/reference/simple-supported-currencies), we need to adapt function `fetchCoingecko`, to make the query parameters optionals and to make the return type use _generics_.

# Changes

- Change return type of `fetchCoingecko` to use _generics_.
- Make `queryParams` optional.

# Tests

### DIFF
--- a/src/frontend/src/lib/rest/goincecko.rest.ts
+++ b/src/frontend/src/lib/rest/goincecko.rest.ts
@@ -20,7 +20,7 @@ export const simplePrice = async ({
 	ids,
 	...rest
 }: CoingeckoSimplePriceParams): Promise<CoingeckoSimplePriceResponse | null> =>
-	fetchCoingecko({
+	fetchCoingecko<CoingeckoSimplePriceResponse>({
 		endpointPath: 'simple/price',
 		queryParams: joinParams([
 			`ids=${Array.isArray(ids) ? ids.join(',') : ids}`,
@@ -40,7 +40,7 @@ export const simpleTokenPrice = async ({
 	contract_addresses,
 	...rest
 }: CoingeckoSimpleTokenPriceParams): Promise<CoingeckoSimplePriceResponse | null> =>
-	fetchCoingecko({
+	fetchCoingecko<CoingeckoSimplePriceResponse>({
 		endpointPath: `simple/token_price/${id}`,
 		queryParams: joinParams([
 			`contract_addresses=${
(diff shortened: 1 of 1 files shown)


peterpeterparker:
For which use case do we need additional endpoints such as [supported_vs_currencies](https://docs.coingecko.com/reference/simple-supported-currencies)?

AntonioVentilii-DFINITY:
> For which use case do we need additional endpoints such as [supported_vs_currencies](https://docs.coingecko.com/reference/simple-supported-currencies)?

I was planning on expanding the Currencies to show the balance, for example, EUR, CHF, or even the quasi-entire list available.

peterpeterparker:
>  to show the balance, for example, EUR, CHF,

Are you doing that now or are you planing for the (not near) future?

If the first, sure, I'll review the PR.
If the latest, I suggest to review this PR later at that time when such modification is going to be required.

AntonioVentilii-DFINITY:
> > to show the balance, for example, EUR, CHF,
> 
> Are you doing that now or are you planing for the (not near) future?
> 
> If the first, sure, I'll review the PR. If the latest, I suggest to review this PR later at that time when such modification is going to be required.

I was doing it now eheh! It would have been among the next PRs (I split it in several, to go step-by-step).

peterpeterparker:
Have you discussed with Artem, Mary and team about the integration of this?

Such questions like how do users select which currency they want to display? should this choice be saved and persisted within the backend canister? or is that just a frontfacing choice? should it be saved in the local storage etc?

AntonioVentilii-DFINITY:
> Have you discussed with Artem, Mary and team about the integration of this?
> 
> Such questions like how do users select which currency they want to display? should this choice be saved and persisted within the backend canister? or is that just a frontfacing choice? should it be saved in the local storage etc?
> 
> If yes, maybe you can briefly explains me next week that way I understand how complex (or not) the feature is and can review the PRs accordingly.

For now, it would have been just a way to optionally have a different currency on functions such as

```typescript
export const exchangeRateETHToUsd = async (): Promise<CoingeckoSimplePriceResponse | null> =>
  simplePrice({
    ids: 'ethereum',
    vs_currencies: 'usd'
  });
```

that would default to `'usd'` if not given (so, exactly as current behavior). And then, at some point, add the front-end feature to choose it, like most of wallets do, and maybe persist it.

peterpeterparker:
If not used, I rather like that we do not add unused options in the code base this to avoid complexity and prevent unexpected usages.

AntonioVentilii-DFINITY:
Closing this PR, awaiting the moment we will make the balance multi-currency.

PR #1395, opened 2024-06-07
refactor: change function tokenUrl to use the token network instead of a prop

AntonioVentilii-DFINITY:
# Motivation

When creating the token URL, the network was given as prop. 

```typescript
const tokenUrl = ({
  token: { name },
  networkId,
  path
}: {
  token: Token;
  networkId: NetworkId;
  path: '/transactions/' | '/';
}): string => [...]
```

However, the `Token` object has the `network` property, and it would be better to use directly that prop, to avoid inconsistencies.

# Changes

- Removed `networkId` from function `tokenUrl`.
- Adapted the usage of such function throughout the code:
   - function `transactionsUrl`
   - component `Tokens`

# Tests

Behavior unchanged during manual testing in local development env, on each network.

### DIFF
--- a/src/frontend/src/lib/utils/nav.utils.ts
+++ b/src/frontend/src/lib/utils/nav.utils.ts
@@ -5,13 +5,8 @@ import type { Token } from '$lib/types/token';
 import { isNullish, nonNullish } from '@dfinity/utils';
 import type { LoadEvent, Page } from '@sveltejs/kit';
 
-export const transactionsUrl = ({
-	token,
-	networkId
-}: {
-	token: Token;
-	networkId: NetworkId;
-}): string => tokenUrl({ path: '/transactions/', token, networkId });
+export const transactionsUrl = ({ token }: { token: Token }): string =>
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -24,17 +19,15 @@ export const isSubRoute = (page: Page): boolean =>
 	isRouteTransactions(page) || isRouteSettings(page);
 
 const tokenUrl = ({
-	token: { name },
-	networkId,
+	token: { name, network },
```
peterpeterparker, line 22: If interested, here you can destruct the `network` that way the content of the function does not have to change.

```
const tokenUrl = ({
   token: { name, network: {id: networkId} },
```

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/utils/nav.utils.ts
AntonioVentilii-DFINITY: even better! changing it

PR #1405, opened 2024-06-11
feat(backend): add a specific call for BTC to endpoint ecdsa_public_key

AntonioVentilii-DFINITY:
# Motivation

As first step to request the Bitcoin address for the user's Principal, we created a custom call to [endpoint `ecdsa_public_key` of IC Management API](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-ecdsa_public_key).
The call requires only the `canisterId` and the `network`. The `key_name` is the same as the one used in configs and the `derivation_path` will default to the following values:
- Mainnet: `m/44'/0'/0'/0/0`
- Testnet and Regnet: `m/44'/1'/0'/0/0`

It respects the specifics of [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) and it is an acceptable complexity for the current scope.

# Changes

- Created derivation path utils.
- Created function to call `ecdsa_public_key`.


src/backend/src/lib.rs
```diff
@@ -436,6 +439,22 @@ fn list_custom_tokens() -> Vec<CustomToken> {
     read_state(|s| s.custom_token.get(&stored_principal).unwrap_or_default().0)
 }
 
+async fn bitcoin_ecdsa_public_key_of(canister_id: Principal, network: BitcoinNetwork) -> Vec<u8> {
```
peterpeterparker, line 442: - Is this function suppose to be available for the outside word? Currently it's not exposed
- Why passing `canister_id` as parameter?

src/backend/src/lib.rs
```diff
@@ -436,6 +439,22 @@ fn list_custom_tokens() -> Vec<CustomToken> {
...
+    let name = read_config(|s| s.ecdsa_key_name.clone());
```
peterpeterparker, line 443: We can use the same name for both ETH and Bitcoin?

src/backend/src/lib.rs
```diff
@@ -436,6 +439,22 @@ fn list_custom_tokens() -> Vec<CustomToken> {
...
+    let derivation_path = network_to_derivation_path(network)?;
+    let (key, ) = ecdsa_public_key(EcdsaPublicKeyArgument {
+        canister_id: Some(canister_id),
```
peterpeterparker, line 446: I don't think we should path the `canister_id` or what does it means if both caller and optional `canister_id` are equals? Are the keys derived differently?

src/backend/src/lib.rs
```diff
@@ -436,6 +439,22 @@ fn list_custom_tokens() -> Vec<CustomToken> {
...
+        derivation_path,
+        key_id: EcdsaKeyId {
+            curve: EcdsaCurve::Secp256k1,
+            name,
+        },
+    })
+        .await
+        .expect("failed to get public key");
```
peterpeterparker, line 454: Given that we are going to have two distinct public keys, the error message can probably be more distinctive as well

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+pub fn network_to_derivation_path(network: BitcoinNetwork) -> Vec<Vec<u8>> {
+    // This is a simplified version of the complete flow to create a derivation path.
+    // The documentation on the complete flow can be found at https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki
+    // For the current implementation and to spare computation workload, we are using the default account 0, change 0, and address index 0.
+    let coin_type = match network {
+        BitcoinNetwork::Mainnet => "0",
+        BitcoinNetwork::Testnet => "1",
+        BitcoinNetwork::Regtest => "1",
```
peterpeterparker, line 27: both regtest and testnet are equals to 1, that's correct?

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,32 @@
...
+    };
+    let path = format!("m/44'/{}'/0'/0/0", coin_type);
+    let path_segments: Vec<Vec<u8>> = parse_derivation_path(&path);
```
peterpeterparker, line 30: You don't need `path_segments`,  `parse_derivation_path(&path)` can be the last statement of the function

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,32 @@
+use std::str::FromStr;
+use ic_cdk::api::management_canister::bitcoin::{BitcoinNetwork};
+
+fn parse_derivation_path(path: &str) -> Vec<Vec<u8>> {
+    path.split('/')
```
peterpeterparker, line 5: should we assert the expected format of `path`? To some extension it can be misused by passing anything as input string some day.

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,32 @@
...
+        .skip(1) // Skip the leading 'm'
+        .map(|s| {
+            if s.ends_with('\'') {
+                // Handle hardened derivation (apostrophe indicates hardened)
+                let index = u32::from_str(&s[..s.len() - 1]).unwrap() + 0x80000000;
```
peterpeterparker, line 10: - can unwrap fails here?
- what's 0x80000000 ?

peterpeterparker (changes requested):
I have added a few comments. Please ensure that you also provide tests, as they are required for any PRs related to the backend.

Additionally, it would be beneficial if Max reviews any backend-related PRs.

src/backend/src/lib.rs
AntonioVentilii-DFINITY: same comment as above

src/backend/src/bitcoin_utils.rs
AntonioVentilii-DFINITY: Theoretically, we could give whatever derivation path we would like in `Regtest` (even empty), since it is simply a local node (no connection whatsoever with the `Mainnet` or `Testnet` chains, so it is not linked to any convention or specific rules, like BIP-44).

I put the same value as `Testnet` to have consistency between Local env and the env where we will test it properly linked to `Testnet`.

src/backend/src/bitcoin_utils.rs
AntonioVentilii-DFINITY: agreed

src/backend/src/bitcoin_utils.rs
AntonioVentilii-DFINITY: - it would have failed, but I included the assertion of the string format (from a previous comment) and now there is no opportunity that it could fail without failing the assertion first
- that value is used to represent an _hardened_ key (a key with ', for example 44'): looking up more details, it is the hexadecimal of the smallest 32-bit value with the high-order bit set

src/backend/src/lib.rs
AntonioVentilii-DFINITY: that part is only to adapt it to the arguments of `ecdsa_public_key`, because it needs to be Optional. But the function `bitcoin_ecdsa_public_key_of` per-se does not accept Optional.

Another way of doing it would be below, but I would prefer to avoid that:

```rust
let (key, ) = ecdsa_public_key(EcdsaPublicKeyArgument {
        canister_id: canister_id.unwrap_or(Principal::anonymous()),
        derivation_path,
[...]
```

src/backend/src/lib.rs
peterpeterparker: One way or another, we need tests provided with this PR. So, if it's easier to provide those by adding the endpoints in this PR, let's do that.

Note: we do not need to replicate the second function you listed above - `eth_address_of` - for Bitcoin. Only the first - `caller_eth_address`.

src/backend/src/lib.rs
peterpeterparker: I think you misunderstand my comment. I think that `canister_id` should be set to `None`.

Unless there is a reason not to do so. If there is can you share the references?

src/backend/src/lib.rs
AntonioVentilii-DFINITY: Ok, then I already pushed the complete PR on this (included tests)

src/backend/src/lib.rs
AntonioVentilii-DFINITY: agreed! In the final PR it would have been that, but since I already pushed it here, you will see canister=None

src/backend/src/bitcoin_utils.rs
AntonioVentilii-DFINITY: I think there is a misunderstanding on what is my enum VS what is the enum in the link.

My list is used to mark the `coin_type` in the derivation path (source [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki)). The list of allowed coin types can be found in [SLIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) and it shows the following:

| Coin type  | Path component (`coin_type'`) | Symbol  | Coin                              |
| ---------- | ----------------------------- | ------- | --------------------------------- |
| 0          | 0x80000000                    | BTC     | Bitcoin                           |
| 1          | 0x80000001                    |         | Testnet (all coins)               |
| 2          | 0x80000002                    | LTC     | Litecoin                          |
| 3          | 0x80000003                    | DOGE    | Dogecoin                          |
| ...        | ...                   | ...     | ...                          |

meaning that number 3 is assigned to LiteCoin.

What I do for the derivation path is to give the same `coin_type` to `regtest` as the one of `testnet`: for `regtest` (that is the local Bitcoin node) we actually don't really need a real derivation path that conforms to BIP-44, but I would like to have the same behavior as if it was running in the `testnet` Bitcoin chain.

What do you think?

P.S. I am documenting it in the code.

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,75 @@
+use std::str::FromStr;
+use ic_cdk::api::management_canister::bitcoin::{BitcoinNetwork};
+use k256::sha2;
+use k256::sha2::Digest;
+
+fn parse_derivation_path(path: &str) -> Vec<Vec<u8>> {
+    let re = regex::Regex::new(r"^m/44'(/\d+'?){2}/[01]/\d+$").unwrap();
+    assert!(re.is_match(path), "Invalid BIP-44 derivation path format");
```
bitdivine, line 8: Is there a reason for parsing once in the check above and again below?  Why not parse just once?  You could do something like:
```
// Gets parts:
let parts = path.split()
let master_node = parts.next().expect("Missing master node");
let purpose = parts.next().expect...
...
assert!(parts.next().is_none(), "Too many parts");
// Checks parts:
assert_eq!(master_node, "m", "Invalid master node");
```

Also, I note that the check allows some invalid derivation paths.  For example, bip44 specifies that the account number must be a natural number (0,1,2...) but the regex allows say "00" and "000" etc.  I am not 100% sure whether it matters here, but in many systems having several representations of the same ID is a bug.

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    path.split('/')
+        .skip(1) // Skip the leading 'm'
+        .map(|s| {
+            if s.ends_with('\'') {
```
bitdivine, line 13: You can use:
```
if let Some(num) = s.strip_suffix("'") {
  // Handles hardened key
  u32::from_str(num) ...
} else {
  // Handles unhardened key
  u32::from_str(s)...
}

File: src/backend/src/bitcoin_utils.rs (hunk 8 above)
Comment (bitdivine) [reply src/backend/src/bitcoin_utils.rs:18]: Is there a limit for how large the index may be?  Can I pass in the unhardened value `m/2147483648/...` for example and have that interpreted in the same way as the hardened key `m/0'/` ?

You might also want to use `checked_add()` to ensure that the addition doesn't wrap.

Comment (bitdivine) [reply src/backend/src/bitcoin_utils.rs:18]: I checked.  BIP44 does not specify a maximum.  Given that the top u32 bit is used as a flag, I would say that a sensible maximum is one that keeps the number smaller than any flag bits.  What do you think?

File: src/backend/src/bitcoin_utils.rs [hunk 11]
```diff
@@ -0,0 +1,75 @@
...
+                // Handle hardened derivation (apostrophe indicates hardened)
+                // Value 0x80000000 is added to represent a hardened key.
+                // It is the hexadecimal of the smallest 32-bit value with the high-order bit set
+                // More details: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#Extended_keys
+                let index = u32::from_str(&s[..s.len() - 1]).unwrap() + 0x80000000;
+                index.to_be_bytes().to_vec()
+            } else {
+                let index = u32::from_str(s).unwrap();
```
bitdivine, line 21: How does `u32::from_str()` handle leading zeros?  Does it accept them or return an error?

bitdivine (commented):
Sorry for the delay, here is a start.

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,75 @@
...
+    hasher.finalize().to_vec()
+}
+fn ripemd160(data: &[u8]) -> Vec<u8> {
+    let mut hasher = ripemd::Ripemd160::new();
+    hasher.update(data);
+    hasher.finalize().to_vec()
+}
+
+// Converts a public key to a P2PKH address.
+pub fn public_key_to_p2pkh_address(public_key: &[u8], network: BitcoinNetwork) -> String {
```
bitdivine, line 58: Is there a standard set of test vectors for this?  Or can we get a set of test cases from a fairly authoritative codebase?

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,75 @@
...
+    let result = ripemd160(&sha256(public_key));
+
+    let prefix = match network {
+        BitcoinNetwork::Testnet | BitcoinNetwork::Regtest => 0x6f,
+        BitcoinNetwork::Mainnet => 0x00,
+    };
+    let mut data_with_prefix = vec![prefix];
+    data_with_prefix.extend(result);
+
+    let checksum = &sha256(&sha256(&data_with_prefix.clone()))[..4];
```
bitdivine, line 69: I believe the clone is not needed?

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,75 @@
...
+pub fn public_key_to_p2pkh_address(public_key: &[u8], network: BitcoinNetwork) -> String {
+    // SHA-256 & RIPEMD-160
+    let result = ripemd160(&sha256(public_key));
+
+    let prefix = match network {
+        BitcoinNetwork::Testnet | BitcoinNetwork::Regtest => 0x6f,
+        BitcoinNetwork::Mainnet => 0x00,
+    };
+    let mut data_with_prefix = vec![prefix];
+    data_with_prefix.extend(result);
```
bitdivine, line 67: Given that the sha256 API consumes multiple chunks, what do you think about feeding it the chunks directly?  Something like this:
```suggestion
    let checksum = {
          let mut hasher = sha2::Sha256::new();
          hasher.update(&[prefix]);
          hasher.update(&result);
          sha256(hasher.finalize().to_vec()).truncate(4)
    };
```

src/backend/src/bitcoin_utils.rs
bitdivine: I note that CI doesn't have Rust lints set up, otherwise it would have let you know about this.  I can take care of that.

bitdivine (commented):
Another chunk.

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,75 @@
...
+    // For the current implementation and to spare computation workload, we are using the default account 0, change 0, and address index 0.
+    // For the Regtest network, we are using the Testnet derivation path, even if it is not necessary to conform to BIP-44, since it is a local node.
+    // However, we do that because we would like to have the same behavior of the Testnet network.
+    let coin_type = match network {
+        BitcoinNetwork::Mainnet => "0",
+        BitcoinNetwork::Testnet => "1",
+        BitcoinNetwork::Regtest => "1",
+    };
+    let path = format!("m/44'/{}'/0'/0/0", coin_type);
+    parse_derivation_path(&path)
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: perfect!

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,75 @@
...
+        })
+        .collect()
+}
+
+pub fn network_to_derivation_path(network: BitcoinNetwork) -> Vec<Vec<u8>> {
+    // This is a simplified version of the complete flow to create a derivation path.
+    // The documentation on the complete flow can be found at https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki
+    // For the current implementation and to spare computation workload, we are using the default account 0, change 0, and address index 0.
+    // For the Regtest network, we are using the Testnet derivation path, even if it is not necessary to conform to BIP-44, since it is a local node.
+    // However, we do that because we would like to have the same behavior of the Testnet network.
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: good suggestion! done!

src/backend/src/bitcoin_utils.rs
AntonioVentilii-DFINITY: yes, indeed there is with [BIP-0038](https://github.com/bitcoin/bips/blob/master/bip-0038.mediawiki)

I will implement it in the tests

src/backend/src/bitcoin_utils.rs
AntonioVentilii-DFINITY: last changes deprecated the usage of `u32::from_str()`

src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,75 @@
+use std::str::FromStr;
+use ic_cdk::api::management_canister::bitcoin::{BitcoinNetwork};
+use k256::sha2;
+use k256::sha2::Digest;
+
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Thank you for the suggestion: I included the crate `derivation_path` in the last commit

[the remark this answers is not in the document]
AntonioVentilii-DFINITY: About `# Panics`: in the last version I think there is no need anymore, since it should not raise an error.

src/backend/src/bitcoin_utils.rs
AntonioVentilii-DFINITY: Actually, @bitdivine , I have created a test that uses the `CALLER` canister and must match the output defined by `CALLER_BTC_ADDRESS`.

The test is quite similar to the ETH function with the same scope.

Do you think we still need to try it out on more examples?

src/backend/src/bitcoin_utils.rs
peterpeterparker: Sounds good, thanks.

src/backend/src/bitcoin_utils.rs
bitdivine: I would, if you can.  Where did you get the Bitcoin address msx83kFfWPib2YbnUk6xTpBn85vefcaVX8 from, used in the test?

bitdivine (approved):
LGTM, thanks, but please add more tests, if you can.

AntonioVentilii-DFINITY:
Since it diverged too much from the main branch, I am closing this in favor of a clean PR:
  - #1986

PR #1411, opened 2024-06-12
chore: env variable to disable exchange

peterpeterparker:
# Motivation

The exchange worker does not start locally because it can quickly throw an error due to too many requests, which is a limitation of Coingecko. However, sometimes it is useful to enable it to develop certain UI-related features. I always forget which component I should temporarily remove the LOCAL flag from, not to mention that it's prone to error.

Long story short, adding an environment variable is handy.

# Changes

- Add env `VITE_EXCHANGE_DISABLED` which fallback to `false` - i.e. exchange enabled per default which is the current behavior

# Tests


AntonioVentilii (approved):
Thank you! I wanted to talk to you about that indeed, I was always manually commenting that snippet to test it

AntonioVentilii-DFINITY (approved):

PR #1419, opened 2024-06-12
refactor: re-ordered store tokens to show Ethereum native tokens before ICRC tokens

AntonioVentilii-DFINITY:
# Motivation

Since we will have a single-view with the list of all tokens of all networks, as part of the requirements, the Ethereum native tokens will be put before the list of ICRC tokens, but after ICP.

# Changes

Changed order of store `tokens`.


peterpeterparker (approved):
LGTM, thanks.

I'll ping Mary because I ask myself if the list should not be different - i.e.

ETH
ICP
ERC20
ICRC (with ck first)

PR #1424, opened 2024-06-12
Style/add network logo to token logo

AntonioVentilii-DFINITY:
# Motivation

As part of the single-view requirements, we would like to see a smaller logo that refers to the Token Network when there is the complete list of Tokens. For now, it will be showed only when the user selects Chain Fusion.

Screenshot 2024-06-12 at 18 14 51

# Changes

- Use inception Logo component to create bottom-right smaller logo when there is a prop for that.
- Use token's network icon in the Token Menu.

# Tests

See screenshot above.


peterpeterparker (changes requested):
The `Logo` component should not be modified as it represents a single logo, an icon with a fallback color. It is also a UI component (see path `ui`) that should not be aware of any logic.

Moreover, your implementation contains a recursive component, which I would prefer to avoid, as it seems prone to bundling issues.

Furthermore, we probably want to display those "sub-icons" in other use cases beyond just the Token page.

That's why, in my opinion, the first step is to create a new component that can render a logo for the token and a sub-icon/logo for its network.

AntonioVentilii-DFINITY:
> That's why, in my opinion, the first step is to create a new component that can render a logo for the token and a sub-icon/logo for its network.

I changed the commits to have a new component with a corner icon, let me know what you think.

I would still leave an optional parameter `showNetworkIcon`, because we are not sure if we will show the network logo in case the user selects a single network.

peterpeterparker:
I would rename `LogoWithCorner` to `TokenLogo` and move the component to `$lib/components/tokens`.

src/frontend/src/lib/components/ui/LogoWithCorner.svelte
```diff
@@ -0,0 +1,24 @@
...
+
+	const {
+		icon,
+		name,
+		network: { name: networkName, icon: networkIcon }
+	} = token;
+</script>
+
+<div class="relative">
+	<Logo src={icon} alt={`${name} logo`} {size} {color} />
```
peterpeterparker, line 18: Unrelated to this PR: should we i18n `... logo`?

src/frontend/src/lib/components/ui/LogoWithCorner.svelte
```diff
@@ -0,0 +1,24 @@
+<script lang="ts">
+	import Logo from '$lib/components/ui/Logo.svelte';
+	import type { Token } from '$lib/types/token';
+
+	export let token: Token;
+	export let size: string;
```
peterpeterparker, line 6: If this component becomes meant for `TokenLogo` we can be less generic here and expose for example a property `export let size: 'small' | 'big' = 'big';`.

src/frontend/src/lib/components/ui/LogoWithCorner.svelte
AntonioVentilii-DFINITY: yes! good point! i will do in another PR

src/frontend/src/lib/components/ui/LogoWithCorner.svelte
AntonioVentilii-DFINITY: You mean the `Logo` component right? I had a look throughout the code and it is used elsewhere too. However, the list of all sizes in which it is used is:

- 64px
- 52px
- 20px

So, I can modify it to have  'small' | 'medium' | 'big', what do you think? In any case, it will be in another PR

AntonioVentilii-DFINITY: Done, please check last commit. It is mostly used as `'small'` , so maybe we can default to that?

src/frontend/src/lib/components/ui/Logo.svelte
```diff
@@ -5,9 +5,16 @@
...
-	export let size: string;
+	export let size: 'small' | 'medium' | 'big' = 'medium';
 	export let color: 'dust' | 'off-white' | 'white' = 'dust';
 
+	const sizes = {
+		small: '20px',
+		medium: '52px',
+		big: '64px'
+	};
+	const sizePx = sizes[size];
```
peterpeterparker, line 16: `let` no?

src/frontend/src/lib/components/ui/Logo.svelte
```diff
@@ -5,9 +5,16 @@
 
 	export let src: string | undefined;
 	export let alt = '';
-	export let size: string;
+	export let size: 'small' | 'medium' | 'big' = 'medium';
```
peterpeterparker, line 8: if it's mostly `small`, shouldn't the default be `small`?

peterpeterparker (approved):
LGTM, thanks

Please have a look at my last two comments before merge.

src/frontend/src/lib/components/ui/Logo.svelte
AntonioVentilii-DFINITY: > Done, please check last commit. It is mostly used as `'small'` , so maybe we can default to that?

indeed, see my comment above, i thought the same, agreed

src/frontend/src/lib/components/ui/Logo.svelte
AntonioVentilii-DFINITY: yes, it is better, agreed

src/frontend/src/lib/components/ui/LogoWithCorner.svelte
AntonioVentilii-DFINITY: Created PR #1431 for this

PR #1425, opened 2024-06-12
fix: changed back function to use the previous route if defined

AntonioVentilii-DFINITY:
# Motivation

The "Back to Wallet" button in each `transaction` page was using the networkId defined by default in the current URL.
For example, if we are in `(app)/transactions/?token=ICP&network=ICP`, it would have returned to `(app)/?network=ICP`.

However that is an incorrect behavior when we deal with Chain Fusion, since the URL network parameter does not match Chain Fusion.
For example:
1) We are in `(app)/?network=ChainFusion`
2) We click on the ICP token that redirects to `(app)/transactions/?token=ICP&network=ICP`
3) We click on "Back to Wallet"
4) We are redirected to `(app)/?network=ICP`, **that is not what we would expect**.

# Changes

- Change function `back` in `nav.utils.ts` to have an optional parameter `fromUrl` that, if it is defined, takes priority over `networkId`.
- Change `Back` component to use the URL of navigational variable `fromRoute`.

# Tests

Manual test in local replica.

### Before:

https://github.com/dfinity/oisy-wallet/assets/169057656/2f4953ff-f187-4cc6-8b3a-28eac9ba718e

### After:

https://github.com/dfinity/oisy-wallet/assets/169057656/7da33a0c-f933-4ba8-bd97-8fdac9a7ddf7

### DIFF
--- a/src/frontend/src/lib/utils/nav.utils.ts
+++ b/src/frontend/src/lib/utils/nav.utils.ts
@@ -35,8 +35,16 @@ const tokenUrl = ({
 export const networkParam = (networkId: NetworkId): string =>
 	`network=${networkId.description ?? ''}`;
 
-export const back = async ({ pop, networkId }: { pop: boolean; networkId: NetworkId }) => {
-	const rootUrl = `/?${networkParam(networkId)}`;
+export const back = async ({
+	pop,
+	networkId,
+	fromUrl
+}: {
+	pop: boolean;
+	networkId: NetworkId;
+	fromUrl?: URL;
+}) => {
+	const rootUrl = fromUrl?.toString() ?? `/?${networkParam(networkId)}`;
 
 	if (!pop) {
 		await goto(rootUrl);
--- a/src/frontend/src/lib/components/core/Back.svelte
+++ b/src/frontend/src/lib/components/core/Back.svelte
@@ -16,6 +16,7 @@
 
 <button
 	class="flex gap-0.5 text-white font-bold pointer-events-auto ml-2"
-	on:click={async () => back({ pop: nonNullish(fromRoute), networkId: $networkId })}
+	on:click={async () =>
+		back({ pop: nonNullish(fromRoute), networkId: $networkId, fromUrl: fromRoute?.url })}
 	><IconBack /> {$i18n.navigation.text.back_to_wallet}</button
 >


src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -35,8 +35,16 @@ const tokenUrl = ({
 export const networkParam = (networkId: NetworkId): string =>
 	`network=${networkId.description ?? ''}`;
 
-export const back = async ({ pop, networkId }: { pop: boolean; networkId: NetworkId }) => {
-	const rootUrl = `/?${networkParam(networkId)}`;
+export const back = async ({
+	pop,
```
peterpeterparker, line 39: given that we add `fromUrl?: URL;` as parameter, do we still need `pop` or can we replace it by use this new parameter within the `back` function?

peterpeterparker (commented):
Unrelated to the review: `https://.../?network?ChainFusion` is a bit ugly no? Should we give it a try of modifying the routing so that the abscence of network parameter equals ChainFusion? We can discuss that offline.

src/frontend/src/lib/utils/nav.utils.ts
AntonioVentilii-DFINITY: right! i will change it

AntonioVentilii-DFINITY: Ok, since its scope is to remove the current navigation URL from the history stack, the parameter `pop` is used unrelated to the `fromRoute` two times:

```typescript
if (isRouteTransactions($page)) {
	await back({ pop: true, networkId: $networkId });
}

[...]

await back({ networkId: backToNetworkId, pop: true });
```

I would say to leave it there, what do you think? Its usage is not specifically related to the page the user was coming from.

peterpeterparker: I see your point. After this PR, is it use somewhere else? If know, not that a strong opinion, but I would lean towards removing it anyway because for maintance reason and to scope the complexity I rather like to expose only the options that are effectively used but, we can keep it as well. I like you decide, I approve the PR.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/utils/nav.utils.ts
AntonioVentilii-DFINITY: It is used in:

## `Network.svelte`

```typescript
	const onClick = async () => {
		await switchNetwork(network.id);

		if (isRouteTransactions($page)) {
			await back({ pop: true, networkId: $networkId });
		}

		// A small delay to give the user a visual feedback that the network is checked
		setTimeout(() => dispatch('icSelected'), 500);
	};
```

## `HideTokenModal.svelte`
```typescript
	await hideToken({
		identity: $authStore.identity
	});

	hideProgressStep = ProgressStepsHideToken.UPDATE_UI;

	// We must navigate first otherwise we might land on the default token Ethereum selected while being on network ICP.
	await back({ networkId: backToNetworkId, pop: true });

	await updateUi({
		identity: $authStore.identity
	});

	hideProgressStep = ProgressStepsHideToken.DONE;
```


But I agree that is basically always `true` so, we could change it like this

```typescript
export const back = async (params: { networkId: NetworkId; fromUrl?: URL }) => {
	const { networkId, fromUrl } = params;
	const rootUrl = fromUrl?.toString() ?? `/?${networkParam(networkId)}`;
	await goto(rootUrl, { replaceState: 'fromUrl' in params ? nonNullish(fromUrl) : true });
};
```

AntonioVentilii-DFINITY: Created PR #1428 for this

PR #1426, opened 2024-06-13
feat: mark cketh derived as deprecated

peterpeterparker:
# Motivation

We will have to find another solution than using `cketh.derived` stores to integrate "ChainFusion". This notably for the Receive, Send and Conversion modal.

# Changes

- Mark all stores as deprecated.


peterpeterparker:
I changed the approach. We will try to use the `$token` store as global selector. Therefore, I close this draft.

PR #1428, opened 2024-06-13
chore: removed unused param  from util back

AntonioVentilii-DFINITY:
# Motivation

We noticed in PR #1425, that function `back` had the param `pop` that was actually alway `true`, except when passed the param `fromUrl`. To simplify, we removed such parameter and adapted the function logic.

# Changes

- Changed params and logic of function `back`.
- Adapted usage of such function.

# Tests

Behavior remains unchanged in manual tests in local replica.


peterpeterparker (approved):
LGTM, thx

PR #1423, opened 2024-06-12
Fix/remove icrc test token from chain fusion

AntonioVentilii-DFINITY:
# Motivation

The Chain Fusion network was mistakenly showing the ICRC test tokens together with all the _mainnet_ tokens.
We filter those tokes based on they Ledger Canister ID.

# Changes

- Create utils to distinguish Test Ledger Canisters.
- Include filter in store `networkTokens` to filter the ICRC tokens by environment, based on the property `ledgerCanisterId`.

# Tests

### Before:

Screenshot 2024-06-12 at 17 26 40

### After:

Screenshot 2024-06-12 at 17 26 55

PR #1429, opened 2024-06-13
feat: filter icrc tokens in Chain Fusion

peterpeterparker:
# Motivation

We do not want to list the ck tokens testnes (ckSepoliaETH, ckTESTBTC etc.) on "Chain Fusion".

# Notes

This PR takes few ideas of #1423 but apply the filter on the `networkTokens` derived stores. That way, not just the list is updated but other derived information such as the total of the USD (balance) will also be displayed correctly.

# Changes

- Refactor code to extract `networkTokens` to a separate module. That way it's a bit more obvious that it mixes information. Note that we might want to implement an eslint rule at some point to avoid circular dependency.
- Extract list of ck ledger ids testnet to a constant `CHAIN_FUSION_NETWORKS_IDS`.
- Use constant in new utils `isTokenIcrcTestnet`.
- Replace usage of constants with utils.
- Use the utils to filter `networkTokens`.
- Add some TODOs I spotted while implementing this.


AntonioVentilii-DFINITY (approved):
Better solution than PR #1423 , thank you!

PR #1431, opened 2024-06-13
chore: use i18n for alt string of logos

AntonioVentilii-DFINITY:
# Motivation

It is better to use `i18n` when giving the `alt` string to `Logo` components.

# Changes

Changed all uses of snippet ``alt=`{...} logo` `` to use `i18n.core.alt.logo`.


peterpeterparker (approved):
Cool, thx

PR #1432, opened 2024-06-14
build: Update Rust version

sa-github-api:
A new Rust version available.


peterpeterparker (approved):
According out colleagues:

> It’s ok to use both 1.78 and 1.79. It’s not a bug, but a chance of stack overflow with deep recursions.

I think we are fine as we used to had 1.78 in production without known issues.

peterpeterparker:
No rush. Let' wait a bit more.

PR #1434, opened 2024-06-14
refactor: renamed default network constant

AntonioVentilii-DFINITY:
# Motivation

We renamed `DEFAULT_NETWORK` to `DEFAULT_ETHEREUM_NETWORK`.


src/frontend/src/eth/constants/networks.constants.ts
```diff
@@ -0,0 +1,3 @@
+import { SUPPORTED_ETHEREUM_NETWORKS } from '$env/networks.env';
+
+export const [DEFAULT_ETHEREUM_NETWORK] = SUPPORTED_ETHEREUM_NETWORKS;
```
peterpeterparker, line 3: I think we can / should rename existing `DEFAULT_NETWORK` - i.e. not add a new constant. Unless there is a good reason to do so, maybe I missed something?

peterpeterparker: That way existing usages of `DEFAULT_NETWORK` which are meant to fallback on Ethereum, inherits it.

AntonioVentilii-DFINITY: Ah, I see! But `DEFAULT_NETWORK` is used in other circumstances where it should not be the `DEFAULT_ETHEREUM_NETWORK` but the real default.

For example, when the page is simply `https://(app)/` it selects DEFAULT_NETWORK.

But agreed, we can change these incorrect usages afterwards in another PR

peterpeterparker (approved):
LGTM, thx

PR #1437, opened 2024-06-14
Feature/chain fusion/modal to manage tokens

AntonioVentilii-DFINITY:
# Motivation

The user should be able to manage all the tokens in the list for Chain Fusion. We are going to replicate the ICP Token Management Modal (without some specific features for now).

# Changes

- Create list of manageable tokens: these tokens have the additional `enabled` prop to define if they are to be showed or not.
- Create modal component and sub-components to show the list.

# Tests


peterpeterparker:
We should wait before starting on this - i.e first network, routing and receive/send modal - because everything which would be done regarding merging those modal currently will probably end being incorrect and not used.

PR #1438, opened 2024-06-14
chore: remove additional menu item in TokensMenu for Chain Fusion

AntonioVentilii-DFINITY:
# Motivation

The context menu that opens in the Tokens list presents the additional item of importing ERC20 tokens, by default for all networks, except ICP.
We would like to remove it by default, and present it ONLY when the network has additional menu items to present. 

# Changes

- Created derived store `additionalMenuItem` to define the current network additional item to present, or `undefined`.
- Changed component `TokensMenu` to use the derived to present the choice, or otherwise do not present any choice if `undefined`.

# Tests

See below screenshots of manual replica:

### Before:

Screenshot 2024-06-14 at 10 51 37

### After:

Screenshot 2024-06-14 at 10 52 04


src/frontend/src/lib/derived/tokens-menu.derived.ts
```diff
@@ -0,0 +1,27 @@
+import { networkEthereum, networkICP } from '$lib/derived/network.derived';
```
peterpeterparker, line 1: Why not just simply displayed this option when the route === transaction?

peterpeterparker (commented):
Sorry, I don't understand this PR. Why no just removing the option of the context menu?

AntonioVentilii-DFINITY:
> Sorry, I don't understand this PR. Why no just removing the option of the context menu?

Just because we would like to have the option at some point for other networks or for Chain Fusion, right?

peterpeterparker:
> Just because we would like to have the option at some point for other networks or for Chain Fusion, right?

Ah right. So, if you synchronize merge branch you can now solve that by displaying the option only if `$network` is defined.

AntonioVentilii-DFINITY:
@peterpeterparker I think there is no need to change it, because now it will show the option only if the network has it. So, in case of `undefined` network, it will show nothing.
I tested on replica and it works still

peterpeterparker:
Good point.

> Or do you prefer to have it in the following logic:
>
> if (noNetwork) --> show nothing (or in the future show something)
> if (network1) --> show option1
> if (network2) --> show option2

Maybe? That would follow the existing pattern similar as for example in the `Hero`.

Note that I don't say that it is a nice pattern but, maybe for consistency?

AntonioVentilii-DFINITY:
> Note that I don't say that it is a nice pattern but, maybe for consistency?

Yes, sure, we can always change later when we have more choices and becomes too verbose. 

But I created new PR #1448 to do this, to have a cleaner history.

PR #1442, opened 2024-06-17
feat: tokens order

peterpeterparker:
# Motivation

Review tokens order as currently in production.

# Changes

- Eth
- ICP
- Bitcoin (new)
- Erc20
- Icrc


AntonioVentilii-DFINITY (commented):
Why Bitcoin as last? Wouldn't it be better to be showed at least together with Ethereum, just to show immediatly that is multi-chain?

peterpeterparker:
Good idea 👍.
Done.

AntonioVentilii-DFINITY (approved):
thank you!

PR #1444, opened 2024-06-17
feat: make global state token nullable

peterpeterparker:
# Motivation

We are requesting to transform Oisy from a "mandatory network/mandatory token" selection approach to an "optional network/optional token" pattern. This means we aim to display multiple tokens across networks on the main screen and add call-to-action buttons—Receive and Send—for each of those tokens. Previously, these actions were only accessible if the global token was derived according to the URL navigation.

To achieve this pattern without rewriting the code of the dApp, my idea is to transform the derived token into a writable token—i.e., having a global store that we can set on the fly to switch from one token to another. Of course, this will mean we will be limited to selecting one token at a time, but it is a step in the right direction and resolves what we are requested to achieve.

That's why, as a first step, this PR transforms the derived $token store from an always-defined value to a potentially nullable value. This way, in future PRs, we can modify this information from a derived to a writable form, which can be set and reset as needed.

# Changes

- Transform `$token` to nullable and propagate the change across the entire application.
- Introduces `$tokenWithFallback` that defaults to ETHEREUM and replicates a similar approach to the current pattern, but restricts its usage to very specific cases which we might review at some point. The idea is that it can also serve as a handy object for coding purposes, like when passing an object to a function without having to check for nullability.
- Adds a `NoListener` component which is mounted when the token is undefined.

# Tests

No that well tested. We gonna have to test manually each and every single functions once we are done breaking and implementing everything.


src/frontend/src/icp/components/fee/BitcoinKYTFee.svelte
```diff
@@ -1,29 +1,28 @@
...
 	let ckBTC = false;
-	$: ckBTC = isTokenCkBtcLedger($token as IcToken);
+	$: ckBTC = isTokenCkBtcLedger($tokenWithFallback as IcToken);
 
 	let btcNetwork = false;
 	$: btcNetwork = isNetworkIdBitcoin(networkId);
 
 	let kytFee: bigint | undefined = undefined;
-	$: kytFee = $ckBtcMinterInfoStore?.[$tokenId]?.data.kyt_fee;
+	$: kytFee = nonNullish($tokenId) ? $ckBtcMinterInfoStore?.[$tokenId]?.data.kyt_fee : undefined;
```
AntonioVentilii-DFINITY, line 25: I am not 100% sure (maybe 90%), but I think that it would return `undefined` either way, if you pass `$tokenId` nullish here

`$: kytFee = $ckBtcMinterInfoStore?.[$tokenId]?.data.kyt_fee;`

you already tried?

src/frontend/src/icp/components/fee/EthereumFeeContext.svelte
```diff
@@ -18,16 +18,18 @@
...
+	$: ckEr20 = isTokenCkErc20Ledger($tokenWithFallback as IcToken);
 
 	let ethNetwork = false;
 	$: ethNetwork = isNetworkIdEthereum(networkId);
 
 	let maxTransactionFeeEth: bigint | undefined = undefined;
-	$: maxTransactionFeeEth = $eip1559TransactionPriceStore?.[$tokenId]?.data.max_transaction_fee;
+	$: maxTransactionFeeEth = nonNullish($tokenId)
+		? $eip1559TransactionPriceStore?.[$tokenId]?.data.max_transaction_fee
+		: undefined;
```
AntonioVentilii-DFINITY, line 32: ditto

src/frontend/src/lib/derived/balances.derived.ts
```diff
@@ -1,18 +1,19 @@
...
 import { balancesStore } from '$lib/stores/balances.store';
 import { nonNullish } from '@dfinity/utils';
 import type { BigNumber } from '@ethersproject/bignumber';
 import { derived, type Readable } from 'svelte/store';
 
 export const balance: Readable<BigNumber | undefined | null> = derived(
-	[balancesStore, tokenId],
-	([$balanceStore, $tokenId]) => $balanceStore?.[$tokenId]?.data
+	[balancesStore, token],
+	([$balanceStore, $token]) => (nonNullish($token) ? $balanceStore?.[$token.id]?.data : undefined)
```
AntonioVentilii-DFINITY, line 9: ditto

src/frontend/src/icp/components/fee/BitcoinKYTFee.svelte
peterpeterparker: Yeah, I wish too.

> Svelte: Type null cannot be used as an index type.
> Svelte: Type undefined cannot be used as an index type.

Capture d’écran 2024-06-17 à 16 36 36

src/frontend/src/icp/components/fee/EthereumFeeContext.svelte
peterpeterparker: See https://github.com/dfinity/oisy-wallet/pull/1444#discussion_r1642923779

src/frontend/src/lib/derived/balances.derived.ts
peterpeterparker: See https://github.com/dfinity/oisy-wallet/pull/1444#discussion_r1642923779

AntonioVentilii-DFINITY (approved):

PR #1446, opened 2024-06-17
feat: sepolia chainlink

peterpeterparker:
# Motivation

Add support for ckSepoliaLINK.

# Note

I picked the logo from https://chain.link/brand-assets. We might want to ask internallly if we are allowed to use this logo.
We can proceed but, question should be solved before deploying on mainnet.


AntonioVentilii-DFINITY (approved):
Look good!

For another PR maybe, I just noticed that there is a file called `network.ircrc.env.ts` that should be `network.icrc.env.ts` , right?

peterpeterparker:
Good catch -> #1449

peterpeterparker:
We reviewed the logo and agreed we can use it as the usage complies with Chainlink guidelines.

PR #1447, opened 2024-06-17
feat: sepolia pepe

peterpeterparker:
# Motivation

Add support for ckSepoliaPEPE.

# Note

Similarly to #1446, we might want to discuss if we can use the logo I download from https://cryptologos.cc/pepe for this token.


AntonioVentilii-DFINITY (approved):

peterpeterparker:
Given that Pepe website does not provide any guidelines, we are guessing that anyone can use the logo and therefore we can use it in the dApp.

PR #1450, opened 2024-06-17
feat: created list of network tokens with prop to be showed or not

AntonioVentilii-DFINITY:
# Motivation

As first step to make all the tokens manageable from a show/hide modal, we add prop `enabled` to `networkTokens` derived.

# Changes

- Create new derived `manageableNetworkTokens` to add prop enabled to `networkTokens`.
- Create new derived `enabledNetworkTokens` to be used as official list of tokens.
- Adapt the components that were using `networkTokens` to use `enabledNetworkTokens`.

### DIFF
--- a/src/frontend/src/lib/components/exchange/ExchangeBalance.svelte
+++ b/src/frontend/src/lib/components/exchange/ExchangeBalance.svelte
@@ -3,10 +3,10 @@
 	import { exchangeInitialized, exchanges } from '$lib/derived/exchange.derived';
 	import { balancesStore } from '$lib/stores/balances.store';
 	import { usdValue } from '$lib/utils/exchange.utils';
-	import { networkTokens } from '$lib/derived/network-tokens.derived';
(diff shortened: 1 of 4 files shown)


peterpeterparker (commented):
> to make all the tokens manageable from a show/hide modal,

Not sure what do you mean. What do we need this extra flag?

AntonioVentilii-DFINITY:
@peterpeterparker I would add it so that they will be "manageable" from the Chain Fusion modal to show/hide tokens. Right now, only ICRC tokens have this propriety, but even ETH, ERC20 (and the rest) should have it

peterpeterparker:
Mmmmh not sure that's correct. `enabled` is a flag parts of the backend data, not sure we should mock it at some higher level. Can we have a look in a call tomorrow?

AntonioVentilii-DFINITY:
Closing this, because we are going to go in another way to create the Manage Tokens modal

PR #1451, opened 2024-06-17
Refactor/extract button menu

AntonioVentilii-DFINITY:
# Motivation

We extract the component

```svelte
<button
  class="flex gap-2 items-center no-underline hover:text-blue active:text-blue"
  aria-label={...}
  on:click={...}
>
 [...]
</button>
```

to generalize its usage.

# Changes

- Create new UI component `ButtonMenu`.
- Expand its usage through the code.

### DIFF
--- a/src/frontend/src/lib/components/ui/ButtonMenu.svelte
+++ b/src/frontend/src/lib/components/ui/ButtonMenu.svelte
@@ -1,0 +1,15 @@
+<script lang="ts">
+	export let label: string;
+	export let onClick: () => void;
+	export let ariaLabel: string | undefined = undefined;
+</script>
+
+<button
+	class={`flex gap-2 items-center no-underline hover:text-blue active:text-blue`}
+	aria-label={ariaLabel ?? label}
+	on:click={onClick}
+>
+	<slot name="before-label"></slot>
+	{label}
+	<slot name="after-label"></slot>
+</button>
--- a/src/frontend/src/eth/components/tokens/ManageTokensMenuButton.svelte
+++ b/src/frontend/src/eth/components/tokens/ManageTokensMenuButton.svelte
@@ -1,14 +1,13 @@
 <script lang="ts">
 	import { i18n } from '$lib/stores/i18n.store';
 	import { modalStore } from '$lib/stores/modal.store';
+	import ButtonMenu from '$lib/components/ui/ButtonMenu.svelte';
 </script>
 
 <div class="content">
-	<button
-		class="flex gap-2 items-center no-underline hover:text-blue active:text-blue"
-		aria-label={$i18n.tokens.import.text.title}
-		on:click={modalStore.openAddToken}
-	>
-		{`+ ${$i18n.tokens.import.text.title}`}
-	</button>
+	<ButtonMenu
+		label={`+ ${$i18n.tokens.import.text.title}`}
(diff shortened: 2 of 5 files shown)


src/frontend/src/lib/components/ui/ButtonMenu.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang="ts">
+	export let label: string;
+	export let onClick: () => void;
```
peterpeterparker, line 3: Not sure what we defined last time we discussed this. Aren't sticking to the current pattern for consistency - bubbling event- for now?

```
// ButtonMenu
<button on:click />

// Consumer
<ButtonMenu on:click={myFunction} />
```

AntonioVentilii-DFINITY: you mean, to reduce the number of props

actually, agreed, better

peterpeterparker: > component would have the event on:click, right?

Only if within the component `on:click` is present

peterpeterparker: It's Svelte v5 this pattern disappear but, in our code base, it is use everywhere. So for consistency I would stick to that but, not sure anymore what we decided when we discussed that.

AntonioVentilii-DFINITY: perfect, will do!

src/frontend/src/lib/components/ui/ButtonMenu.svelte
```diff
@@ -0,0 +1,14 @@
...
+	export let ariaLabel: string | undefined = undefined;
+</script>
+
+<button
+	class={`flex gap-2 items-center no-underline hover:text-blue active:text-blue`}
+	aria-label={ariaLabel ?? label}
+	on:click
+>
+	<slot name="before-label"></slot>
+	{label}
```
peterpeterparker, line 12: Sorry did not noticed in the first review, why not just `<slot />` here?

Then

```
<ButtonMenu label={$i18n.tokens.manage.text.title} on:click={modalStore.openIcManageTokens} />
```

becomes

```
<ButtonMenu on:click={modalStore.openIcManageTokens} >
   {$i18n.tokens.manage.text.title} 
</ButtonMenu>
```

It's closer to the rest of the code base. WDYT? Or rather like to pass it as prop?

AntonioVentilii-DFINITY: So, at first actually I didn't even put the slot, because I prefer to pass the label as prop, instead of component in slot.
However, there is one usage that requires a slot

```svelte
// old code
<button
	class="flex gap-2 items-center no-underline hover:text-blue active:text-blue"
	aria-label={$i18n.navigation.alt.more_settings}
>
	<IconSettings />
	{$i18n.settings.text.title}
</button>
```

and I thought this could be a nice solution, since most usage do not require additional components

AntonioVentilii-DFINITY:
> I approve the PR and let you decide. Thanks!

I had a look around to some UI libraries (MUI, Ant Design, Chakra UI) to get some inspiration: they all tend to give the button label as slot.

So, I will do as you suggested, and then merge

PR #1452, opened 2024-06-17
style: include Chain Fusion logo

AntonioVentilii-DFINITY:
# Motivation

We introduce the Chain Fusion logo in its first version.

# Changes

- Add Chain Fusion logo as SVG.
- Include the logo in the `NetworksSwitcher` component.

### DIFF
--- a/src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
+++ b/src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
@@ -5,7 +5,6 @@
 	import IconMorePlain from '$lib/components/icons/IconMorePlain.svelte';
 	import Network from '$lib/components/networks/Network.svelte';
 	import { selectedNetwork } from '$lib/derived/network.derived';
-	import { nonNullish } from '@dfinity/utils';
 	import { networksMainnets, networksTestnets } from '$lib/derived/networks.derived';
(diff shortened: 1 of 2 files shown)


AntonioVentilii-DFINITY:
We could discuss if at some point we can extract the `Img` component to try and re-use the `Logo` component.

src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
```diff
@@ -31,7 +32,17 @@
 	aria-label={$i18n.networks.title}
 >
 	<div class="w-full h-full md:w-[28px] md:h-[28px]">
-		{#if nonNullish($selectedNetwork?.icon)}
+		{#if isNullish($selectedNetwork)}
```
peterpeterparker, line 35: What about inlining instead of duplicating properties?

```
<Img
				src={$selectedNetwork?.icon ?? chainFusion}
				alt={replacePlaceholders($i18n.core.alt.logo, {
					$name: $selectedNetwork?.name ?? $i18n.networks.chain_fusion
				})}
				width="100%"
				height="100%"
				rounded
			/>
```

 WDYT?

AntonioVentilii-DFINITY: yes! i have avoided it to keep the IF logic, but I prefer it too

peterpeterparker (approved):
LGTM, thx

PR #1455, opened 2024-06-17
Feature/chain fusion/testnet single page token list

AntonioVentilii-DFINITY:
# Motivation

To show only the test tokens for the Chain Fusion testnet selection, we first create a store for the Chain Fusion settings, and we use it to simplify the logic of derived `networkTokens`.
The next step will be to adapt NetworksSwitcher to set the store if Chain Fusion is selected or not.

# Changes

- Create utils to point out test tokens.
- Create store and derived for Chain Fusion with prop `onlyTestnets`.
- Adapt derived `networkTokens` to use both the above.


peterpeterparker:
Same obvervarion about testnet as in other PR (https://github.com/dfinity/oisy-wallet/pull/1456#issuecomment-2174950521), we agreed to not implement Chain Fusion testnet.

PR #1456, opened 2024-06-17
refactor: extract Chain Fusion network button

AntonioVentilii-DFINITY:
# Motivation

To be consistent, we create a separate button for Chain Fusion in `NetworksSwitcher`. After this PR, we will re-use it for the Chain Fusion only-testnets selection.

# Changes

- Add testnet Chain Fusion icon.
- Create separate button, re-using `NetworkButton`.
- Adapt code.


peterpeterparker:
I think we agreed, in last Tuesday's late evening call with Mary, not to implement and add a Chain Fusion testnet.
Was this re-discussed and changed?

AntonioVentilii-DFINITY:
I thought they had change their mind, since I saw that in Figma they included Chain Fusion. I will ask Mary

peterpeterparker:
Gotcha. Artem was not present in the call, maybe he did not get the information. Let's discuss it in today's meeting.

AntonioVentilii-DFINITY:
This was useful in case of Chain Fusion Network.

As discussed in the call, we will not have a Chain Fusion testnet, since it is not a network

PR #1462, opened 2024-06-18
feat: reset and deprecate token store

peterpeterparker:
# Motivation

See comment in code of the PR.

# Changes

- Deprecate token store.
- Add a function `reset()` to the store (for consistency reason).


AntonioVentilii-DFINITY (approved):
Looks good!

Maybe add a TODO flag just to be sure? I generally do a search to check possible to-dos

PR #1470, opened 2024-06-18
feat: open and close send modals

peterpeterparker:
# Motivation

Set the global `$token` store before opening the "Send" modal and reset the store on close.

# Changes

- Refactor function to close and open modal+callback in a services that are a bit more generic.
- Pass token to `Send` and `IcSend`.
- Set `$token` before opening modals.
- Implement plus fix usage of the close action because we want to reset the `$token` only in case of modal, not when the modals are use from the hero on the transactions page.


src/frontend/src/icp/components/receive/IcReceive.svelte
```diff
@@ -34,13 +34,15 @@
...
+		await loadTokenAndRun({ token, callback });
 	};
 
+	const dispatch = createEventDispatcher();
+
 	const close: CloseModalAndResetToken = () => {
 		modalStore.close();
-		tokenStore.set(null);
+		// We are resetting the token in the parent. That way we can know if we are using a global page $token or a selected token.
+		dispatch('nnsClose');
```
AntonioVentilii-DFINITY, line 45: if we are thinking of abandoning the page `$token` at some point, maybe it is worth to start pointing the places where will we modify once done? some TODO here maybe? or we revisit the entire code once done?

AntonioVentilii-DFINITY (approved):
i just added a comment in one of the files, more a doubt really

src/frontend/src/icp/components/receive/IcReceive.svelte
peterpeterparker: Not sure, ultimatelly I kind of feel like `$token` should just not exist at all anymore, so it would be like adding TODO everywhere ;)

PR #1471, opened 2024-06-18
feat(scripts): script to help creating new ERC20 env file

AntonioVentilii-DFINITY:
# Motivation

Created util script to help with the steps 3 to 5 explained with PR #1459 when adding a new ERC20 token:

> ...
> 3. Create a new source environment file in [src/frontend/src/env] by cloning [src/frontend/src/env/tokens.usdc.env.ts] and renaming `usdc` to the token's name.
> 4. **Adapt the content of the tokens:** 
>     1. Find the contract address on the ckETH dashboard [production](https://sv3dd-oaaaa-aaaar-qacoa-cai.raw.icp0.io/dashboard) or [testnet](https://jzenf-aiaaa-aaaar-qaa7q-cai.raw.icp0.io/dashboard) in the table "Supported ckERC20 tokens". 
>     2. Obtain the token name, decimals, and symbol on Etherscan using the contract address (Select "Contract > Read contract" to query various information from the ABI).
> 5. **Set up the token** for the Ethereum network by listing the new token in the twin tokens arrays of 
> ...

# Changes

- Created script `add.tokens.erc20.mjs`.
- Add explanation to `HACKING.md.`

# Tests

Manual testing to verify if newly created files would match existing files.


package.json
```diff
@@ -70,12 +70,14 @@
 		"@typescript-eslint/eslint-plugin": "^7.2.0",
 		"@typescript-eslint/parser": "^7.2.0",
 		"autoprefixer": "^10.4.19",
+		"cheerio": "^1.0.0-rc.12",
```
peterpeterparker, line 73: Let's not add dependency unless it's necessary. Do we really need this cheerio?

package.json
```diff
@@ -70,12 +70,14 @@
...
 		"dotenv": "^16.4.5",
 		"eslint": "^8.57.0",
 		"eslint-config-prettier": "^9.1.0",
 		"eslint-plugin-local-rules": "^2.0.1",
 		"eslint-plugin-svelte": "^2.35.1",
 		"jsdom": "^24.1.0",
+		"node-fetch": "^3.3.2",
```
peterpeterparker, line 80: NodeJS now supports fetch out of the box.

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+import { fileURLToPath } from 'node:url';
+import path from 'path';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+const rootPath = resolve(__dirname, '../');
+
+const envFile = process.env.NODE_ENV === 'production' ? '.env' : '.env.development';
+const envPath = resolve(rootPath, envFile);
+dotenv.config({ path: envPath });
```
peterpeterparker, line 16: Can't we use `config({ path: `.env.${ENV}` });` as in other scripts?

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+
+const getArgValue = (argName) => {
+	const argIndex = process.argv.indexOf(argName);
+	return argIndex > -1 && argIndex < process.argv.length - 1 ? process.argv[argIndex + 1] : null;
```
peterpeterparker, line 20: there is not overflow in JS if I'm not wrong so maybe you can just do `return process.argv[argIndex + 1]` here?

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+};
+
+const ETHERSCAN_API_KEY = getArgValue('--etherscan-api-key') || process.env.VITE_ETHERSCAN_API_KEY;
```
peterpeterparker, line 23: prefer nullish coalescing instead of `||` which is less prone to unwanted errors.

ps: we should set an eslint rule for that or did we already?

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+
+if (!ETHERSCAN_API_KEY) {
```
peterpeterparker, line 25: `isNullish`? 😉

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+	console.error(
+		`Missing VITE_ETHERSCAN_API_KEY. Please provide it in ${envFile} or via --etherscan-api-key argument.`
+	);
+	process.exit(1);
+}
+
+const PROD_DASHBOARD_URL = 'https://sv3dd-oaaaa-aaaar-qacoa-cai.raw.icp0.io/dashboard';
```
peterpeterparker, line 32: should we read those canister ids from dfx.json?

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+		console.log(`File ${newFilePath} already exists.`);
+		return;
+	} catch (err) {
+		// File does not exist, proceed with creation
+	}
+
+	const mainnetToken = `${tokenDetails.symbol}_TOKEN`;
+	const testnetToken = `SEPOLIA_${tokenDetails.symbol}_TOKEN`;
+
+	const newContent = `import { ETHEREUM_NETWORK, SEPOLIA_NETWORK } from '$env/networks.env';
```
peterpeterparker, line 82: This seems difficult to maintain. Have you thought about for example reading the `tokens.usdc.env` file and doing search and replace in the content?

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+	const filePath = path.join(SRC_PATH, 'tokens.erc20.env.ts');
+	let content = await fs.readFile(filePath, 'utf8');
+
+	const mainnetRegex = /const ERC20_TWIN_TOKENS_MAINNET: RequiredErc20Token\[] = \[([\s\S]*?)];/;
+	const sepoliaRegex = /const ERC20_TWIN_TOKENS_SEPOLIA: RequiredErc20Token\[] = \[([\s\S]*?)];/;
+
+	const mainnetMatch = content.match(mainnetRegex);
+	const sepoliaMatch = content.match(sepoliaRegex);
+
+	if (mainnetMatch) {
```
peterpeterparker, line 141: this block of code and next one are similar, maybe worth extracting a local function?

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+	const importStatement = `import { ${mainnetToken}, ${testnetToken} } from '$env/${newFileName.replace('.ts', '')}';\n`;
+	content = importStatement + content;
+
+	await fs.writeFile(filePath, content);
+	console.log(`Updated ${filePath}`);
+};
+
+const fetchSupportedTokens = async (url) => {
+	try {
+		const html = await fetchHtml(url);
```
peterpeterparker, line 168: Few notes:

- Why searching for tokens online when those Oisy is supporting are listed in `tokens.ckerc20.json`? This json file is updated on a weekly basis and can also be updated thanks to a script or manually in the CI.

- If we have to fetch those information online, reading an HTML file to extract the list of token seems unreliable. Moreover, to some extension given that the outcome should be processed manually, not really secure. Why not calling the ck orchestrator canisters to get the list of supported tokens?

But, again, we have a json file. Maybe I just do not understand what the script is doing here?

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+	const testnetTokenMap = new Map(
+		testnetTokens.map((token) => [token.symbol.replace('ckSepolia', 'ck'), token.contractAddress])
+	);
+
+	const tokensToProcess = [];
+
+	for (const token of prodTokens) {
+		const testnetContractAddress = testnetTokenMap.get(token.symbol);
+		if (!testnetContractAddress) {
+			continue;
```
peterpeterparker, line 202: nth (opiniated comment, feel free to ignore): it would be nice to implement a more functional pattern and avoid the use of `continue` which I find difficult to read and prone to a variety of errors when code is modified later one.

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+		}
+
+		const tokenName = token.symbol.slice(2);
+		const tokenFilePath = path.join(SRC_PATH, `tokens.${tokenName.toLowerCase()}.env.ts`);
+		try {
+			await fs.access(tokenFilePath);
+		} catch (err) {
+			tokensToProcess.push({
```
peterpeterparker, line 210: When I use `access` this is how I check if the error is not found or another error:

```
export const junoConfigExist = async (params: {filename: ConfigFilename}): Promise<boolean> => {
  try {
    const {configPath} = junoConfigFile(params);
    await access(configPath);
    return true;
  } catch (err: unknown) {
    if (err instanceof Error && 'code' in err && (err as NodeJS.ErrnoException).code === 'ENOENT') {
      return false;
    } else {
      throw err;
    }
  }
};
```

sharing just in case in intersting for this use case.

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+				tokenName,
+				symbol: token.symbol,
+				contractAddress: token.contractAddress,
+				testnetContractAddress
+			});
+		}
+	}
+
+	if (tokensToProcess.length === 0) {
+		console.log('All prod tokens that have a test token are set.');
```
peterpeterparker, line 220: I don't understand this message. What are you asserting here?

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+		return;
+	}
+
+	for (const { tokenName, contractAddress, testnetContractAddress } of tokensToProcess) {
+		console.log('--------------------------------');
+		console.log(`Creating file for token ${tokenName}`);
+		const tokenDetails = await fetchTokenDetails(contractAddress);
+
+		const { newFileName, mainnetToken, testnetToken } = await createEnvFile(
```
peterpeterparker, line 229: nth: not against using an object as parameter for `createEnvFile({tokenName, contractAddress, ...etc})` as well

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,264 @@
...
+	);
+
+	tokensToProcess.forEach(({ tokenName }) => {
+		console.log(
+			`Example: For token ${tokenName}, the SVG file should be: ${SRC_DIR}/${tokenName.toLowerCase()}.svg`
+		);
+	});
+};
+
+main().catch(console.error);
```
peterpeterparker, line 264: for what it's worth, top level `await` is possible

peterpeterparker (commented):
Nice, cool idea 👍

Few questions and suggestions.

package.json
AntonioVentilii-DFINITY: ah tks! i was in doubt

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: done

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: changed!

so, we have that rule set, but the `scripts` folder was not included in `tsconfig.eslint.json`

```json
"include": [
		"ambient.d.ts",
		"non-ambient.d.ts",
		"./types/**/$types.d.ts",
		"./vite.config.js",
		"./vite.config.ts",
		"./src/**/*.js",
		"./src/**/*.ts",
		"./src/**/*.svelte",
		"./vitest.setup.ts",
		"./vitest.config.ts",
		"./tailwind.config.js",
		"./svelte.config.js",
		"./postcss.config.js",
		"./playwright.config.ts",
		"./e2e/**/*.js",
		"./e2e/**/*.ts",
		"./scripts/*.mjs"  <-- to be included
	],
```

i can do it in another PR; let's note that we need to create the exception that for those specific files we accept `console.log`

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: right! we can use there too! thank you

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: yes, that was my first iteration, but the specifics of searching what needed to be replaced are as difficult to maintain as creating an entire new file, i thought: a lot of specific regex strings that should be re-verified either way in case we change something.

the difference would be that in this case is more clear visually when developing

I am not really sure to be honest, what do you think? which one would be the better way to reduce maintainance?

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: done

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: done

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: i didn't even think about that! cool tks!

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: yes, i will use, better

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: done

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: no, for sure it is better to use the json, I did that as first, but I thought that since it is inside `src` folder, you would have preffered to have on-way interactions between `scripts` --> `src`

i will use the json here

peterpeterparker: > for sure it is better to use the json

Then let's just read the json file.

```
const obj = JSON.parse(readFileSync(pathToJson));
```

scripts/add.tokens.erc20.mjs
peterpeterparker: Both seems to have some trade-off in terms of maintenance. Then let's go with your implementation.

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: Created PR #1505 to help getting the ERC20 contract address

AntonioVentilii-DFINITY: done

package.json
AntonioVentilii-DFINITY: removed

scripts/add.tokens.erc20.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+if [ -n "$1" ]; then
+  ETHERSCAN_API_KEY=$1
+  node ./scripts/add.tokens.erc20.mjs --etherscan-api-key "$ETHERSCAN_API_KEY"
+else
+  node ./scripts/add.tokens.erc20.mjs
+fi
+
+#npm run format && npm run lint
```
peterpeterparker, line 10: You want to uncomment this? Or remove it its a leftover?

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,262 @@
...
+};
+
+const flattenEnvironmentData = (data) => {
+	return Object.entries(data).reduce((acc, [environment, values]) => ({
+		...acc,
+		[environment]: flattenData(values)
+	}), {});
+};
+
+const loadSupportedTokens = () => {
```
peterpeterparker, line 169: maybe rename the function `readSupportedTokens`?

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,262 @@
...
+	const jsonPath = path.resolve(DATA_DIR_PATH, 'tokens.ckerc20.json');
+	return flattenEnvironmentData(JSON.parse(readFileSync(jsonPath, 'utf-8')));
+};
+
+const filterTokens = async (prodTokens, testnetTokens) => {
```
peterpeterparker, line 174: Does this approach actually works when the tokens exists on testnet but, has not yet been proposed and executed on mainnet?

For example ckSepoliaPEPE is on testnet but, the proposal is not yet on mainnet.

scripts/add.tokens.erc20.sh
AntonioVentilii-DFINITY: good question: do you think i shall put it there for good measure? or in an automatic job we will put it in the job and not in the script?

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: no, for now it works only for tokens that are both testnet and mainnet, i thought as a first step to do it like that, then we could improve

what do you think?

scripts/add.tokens.erc20.sh
peterpeterparker: Maybe there that way when we run it in the CI but also locally, the output is formatted?

scripts/add.tokens.erc20.mjs
peterpeterparker: Ah, I think we should change it then 😉. The typical flow is deploy a ckSepolia token on mainnet for testnet / test purpose and then few days/weeks later, make a NNS proposal to enable the related production token. So, we want to integrate testnet earlier as this week with ckSepoliaPepe without ckPepe yet. Does that make sense?

AntonioVentilii-DFINITY: ah! yes, for sure it makes sense, let me modify it and come back

scripts/add.tokens.erc20.sh
AntonioVentilii-DFINITY: actually, different change in the last commit: i included the optional run of `npm run format && npm run lint` only if files were created

otherwise it would have run everytime

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: updated! but it became another script basically eheh! please review it all, thank you!

AntonioVentilii-DFINITY:
@peterpeterparker updated! but it became another script basically eheh! please review it all, thank you!

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,377 @@
...
+		(acc, [environment, values]) => ({
+			...acc,
+			[environment]: flattenData(values)
+		}),
+		{}
+	);
+};
+
+const readSupportedTokens = () => {
+	const jsonPath = path.resolve(DATA_DIR_PATH, 'tokens.ckerc20.json');
```
peterpeterparker, line 292: You can use `resolve`  instead of `path.resolve` given that you are already `import { resolve } from ...`

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,377 @@
...
+	const { production: prodTokens, staging: testnetTokens } = flattenEnvironmentData(
+		JSON.parse(readFileSync(jsonPath, 'utf-8'))
+	);
+	return [...prodTokens, ...testnetTokens];
+};
+
+const parseTokens = (tokens) =>
+	Object.values(
+		tokens.reduce((acc, { symbol, erc20ContractAddress }) => {
+			const mainSymbol = symbol.replace('Sepolia', '').slice(2);
```
peterpeterparker, line 302: Here we are assuming that all token are prefixed with "Sepolia" but, it would be more accurate to not concatenate all tokens in `readSupportedTokens` and leverage their `testnet` or `mainnet` classification here.

i.e.

```
const tokens = readSupportedTokens(); // just read the json no manipulation

const tokensToProcess = parseTokens(tokens); // here object.entries and manipulation
```

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,377 @@
...
+			const contractAddress = !symbol.includes('ckSepolia') ? erc20ContractAddress : undefined;
+			const testnetContractAddress = symbol.includes('ckSepolia')
+				? erc20ContractAddress
+				: undefined;
+
+			acc[mainSymbol] = {
+				mainSymbol,
+				contractAddress: acc[mainSymbol]?.contractAddress || contractAddress,
```
peterpeterparker, line 310: The linter still does not catch nullish coalish `??` instead of `||`?

scripts/add.tokens.erc20.mjs
```diff
@@ -0,0 +1,377 @@
...
+
+	const mainnetTokenCreated = existingFileContent.split(/\W+/).includes(mainnetToken ?? '');
+	const testnetTokenCreated = existingFileContent.split(/\W+/).includes(testnetToken ?? '');
+
+	if (mainnetTokenCreated && testnetTokenCreated) {
+		console.log(`Variables for token ${symbol} already exists in the environment file ${fileName}`);
+		return { fileName, mainnetToken: undefined, testnetToken: undefined };
+	}
+
+	const { name, decimals } = await fetchTokenDetails({
```
peterpeterparker, line 127: Do I understand correctly that you are fetching the token details either for production or testnets? If correct, isn't this inaccurate? Can't token information, notably name, be potentially different on mainnet and testnet? Shouldn't we always fetch their corresponding information?

AntonioVentilii-DFINITY: indeed, I was incorrectly assuming they would be the same, so I wanted to reduce the fetching time.

But it is not the same (as you pointed out for example, USD has different names for Mainnet and Sepolia), so I will change it

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: Strangely enough no! I am not really sure what happened, because a few times, it caught it. Why not this time I am not sure. I will investigate

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: yes, makes sense! better!

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: As discussed, this is solved

scripts/add.tokens.erc20.mjs
AntonioVentilii-DFINITY: done

peterpeterparker (approved):
LGTM, thanks

PR #1472, opened 2024-06-18
feat: use receive icon instead of qr code

peterpeterparker:
# Screenshot

Capture d’écran 2024-06-18 à 14 07 15


peterpeterparker:
Discussed with Artem, actually "QR code" as icon is what we want. Therefore I close this PR.

PR #1474, opened 2024-06-18
refactor: moved token manage modal (and sub-components) to icp-eth folder

AntonioVentilii-DFINITY:
# Motivation

To create a general modal to manage tokens for Chain Fusion, we try and modify the existing `IcManageTokensModal` component used for ICP, since it is the most similar to the final result we would like to achieve.
As first step, we move it (and the related sub-components) to the common `icp-eth` folder.


peterpeterparker (approved):
LGTM, thanks

PS: Are you planing to remove the `Ic` prefix from the file name and usage of the component in another PR?

AntonioVentilii-DFINITY:
yes, correct, i would just wait a bit, right before adding non-ICRC tokens, just to be consistent

PR #1477, opened 2024-06-18
refactor: changed indexing in component IcManageTokens

AntonioVentilii-DFINITY:
# Motivation

To start generalizing its usage, we change the way that tokens are indexed in IcManageTokens component. Before they were using property `ledgerCanisterId`, that is unique for token of type `IcrcCustomToken`. However, we would prefer to use it for the more generic type that could be `Token` or a close child.

# Changes

Changed indexing of modifiedTokens variable to use token ID and network ID, which combination is considered unique for any token in our list throughout the code.

### DIFF
--- a/src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
@@ -18,8 +18,6 @@
 	import { ICP_TOKEN } from '$env/tokens.env';
 	import { sortIcTokens } from '$icp/utils/icrc.utils';
 	import { replacePlaceholders } from '$lib/utils/i18n.utils';
-	import type { TokenId } from '$lib/types/token';
-	import type { NetworkId } from '$lib/types/network';
 
 	const dispatch = createEventDispatcher();
 
@@ -73,44 +71,33 @@
 	$: tokens = filteredTokens.map(({ id, network, enabled, ...rest }) => ({
 		id,
 		network,
-		enabled: modifiedTokens[network.id]?.[id]?.enabled ?? enabled,
+		enabled: modifiedTokens[`${network.id.description}-${id.description}`]?.enabled ?? enabled,
 		...rest
 	}));
 
 	let noTokensMatch = false;
 	$: noTokensMatch = tokens.length === 0;
 
-	let modifiedTokens: Record<NetworkId, Record<TokenId, IcrcCustomToken>> = {};
+	let modifiedTokens: Record<string, IcrcCustomToken> = {};
 	const onToggle = ({ detail: { id, network, ...rest } }: CustomEvent<IcrcCustomToken>) => {
 		const { id: networkId } = network;
+		const { [`${networkId.description}-${id.description}`]: current, ...tokens } = modifiedTokens;
 
-		modifiedTokens[networkId] = modifiedTokens[networkId] || {};
-
-		if (nonNullish(modifiedTokens[networkId][id])) {
-			delete modifiedTokens[networkId][id];
-			if (Object.getOwnPropertySymbols(modifiedTokens[networkId]).length === 0) {
-				delete modifiedTokens[networkId];
-			}
+		if (nonNullish(current)) {
+			modifiedTokens = { ...tokens };
 			return;
 		}
 
-		modifiedTokens[networkId][id] = { id, network, ...rest };
+		modifiedTokens = {
+			[`${networkId.description}-${id.description}`]: { id, network, ...rest },
+			...tokens
+		};
 	};
 
 	let saveDisabled = true;
-	$: saveDisabled = Object.getOwnPropertySymbols(modifiedTokens).every(
-		(networkId) => Object.getOwnPropertySymbols(modifiedTokens[networkId]).length === 0
-	);
+	$: saveDisabled = Object.keys(modifiedTokens).length === 0;
 
-	const save = () =>
-		dispatch(
-			'icSave',
-			Object.getOwnPropertySymbols(modifiedTokens).flatMap((networkId) =>
-				Object.getOwnPropertySymbols(modifiedTokens[networkId]).map(
-					(tokenId) => modifiedTokens[networkId][tokenId]
-				)
-			)
-		);
+	const save = () => dispatch('icSave', Object.values(modifiedTokens));
 </script>
 
 <div class="mb-4">
--- a/src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
@@ -105,8 +105,10 @@
 	const save = () =>
 		dispatch(
 			'icSave',
-			Object.values(modifiedTokens).flatMap((subObject) =>
-				Object.values(subObject as Record<TokenId, IcrcCustomToken>)
+			Object.getOwnPropertySymbols(modifiedTokens).flatMap((networkId) =>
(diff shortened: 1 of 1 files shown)


peterpeterparker:
This breaks the current feature - i.e. if I test the feature, I cannot save and edit the list of tokens. This will be fixed in another PR?

AntonioVentilii-DFINITY:
Wait, I tested it locally and it was not breaking it, can we see it in a call tomorrow?

AntonioVentilii-DFINITY:
my mistake sorry, I had forgot to adapt the last part of the code

please re-try now

src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
```diff
@@ -69,40 +70,47 @@
 			);
 
 	let tokens: IcrcCustomToken[] = [];
-	$: tokens = filteredTokens.map(({ ledgerCanisterId, enabled, ...rest }) => ({
-		ledgerCanisterId,
-		enabled: modifiedTokens[`${ledgerCanisterId}`]?.enabled ?? enabled,
+	$: tokens = filteredTokens.map(({ id, network, enabled, ...rest }) => ({
+		id,
```
peterpeterparker, line 74: I think we do not need to explicitely add the `id` here given it is contained in the spread `...rest`

src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
```diff
@@ -69,40 +70,47 @@
...
-		detail: { ledgerCanisterId, enabled, ...rest }
-	}: CustomEvent<IcrcCustomToken>) => {
-		const { [`${ledgerCanisterId}`]: current, ...tokens } = modifiedTokens;
+	let modifiedTokens: Record<NetworkId, Record<TokenId, IcrcCustomToken>> = {};
+	const onToggle = ({ detail: { id, network, ...rest } }: CustomEvent<IcrcCustomToken>) => {
+		const { id: networkId } = network;
 
-		if (nonNullish(current)) {
-			modifiedTokens = { ...tokens };
+		modifiedTokens[networkId] = modifiedTokens[networkId] || {};
```
peterpeterparker, line 87: `??`

src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
```diff
@@ -69,40 +70,47 @@
...
+
+		if (nonNullish(modifiedTokens[networkId][id])) {
+			delete modifiedTokens[networkId][id];
```
peterpeterparker, line 90: Is there a way to avoid the use of `delete` - i.e. implement a more functional approach? 

A bit an opiniated feedback but, somehow I remember having read some articles explaining that `delete` was a bad practice.

If it would become too verbose let's keep those but, generally I try to avoid those.

src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
```diff
@@ -69,40 +70,47 @@
...
-
-	const save = () => dispatch('icSave', Object.values(modifiedTokens));
+	$: saveDisabled = Object.getOwnPropertySymbols(modifiedTokens).every(
+		(networkId) => Object.getOwnPropertySymbols(modifiedTokens[networkId]).length === 0
+	);
+
+	const save = () =>
+		dispatch(
+			'icSave',
+			Object.getOwnPropertySymbols(modifiedTokens).flatMap((networkId) =>
```
peterpeterparker, line 108: Using e.g. a reducer would we manage to do build the result in a single iteration?

peterpeterparker (commented):
It works out now, thanks!

Few questions.

src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
AntonioVentilii-DFINITY: i extract it because I use it for the enabled line after:

```svelte
	$: tokens = filteredTokens.map(({ id, network, enabled, ...rest }) => ({
		id,
		network,
		enabled: modifiedTokens[network.id]?.[id]?.enabled ?? enabled,   <--- here
		...rest
	}));
```

src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
AntonioVentilii-DFINITY: right!!!

but why didn't lint flag it?!

src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
AntonioVentilii-DFINITY: well, the first iteration I did was not a nested record, but

`let modifiedTokens: Record<string, IcrcCustomToken | undefined> = {};`

where the index was

`${networkId.description}-${id.description}`

a bit more verbose, but it takes one loop off, so it is preferable: changing it

src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
AntonioVentilii-DFINITY: yes, removing it!

in another PR, I will add the lint rule eheh

src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
peterpeterparker: Gotcha

src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
```diff
@@ -69,32 +68,28 @@
...
 
-	let modifiedTokens: Record<CanisterIdText, IcrcCustomToken> = {};
-	const onToggle = ({
-		detail: { ledgerCanisterId, enabled, ...rest }
-	}: CustomEvent<IcrcCustomToken>) => {
-		const { [`${ledgerCanisterId}`]: current, ...tokens } = modifiedTokens;
+	let modifiedTokens: Record<string, IcrcCustomToken> = {};
+	const onToggle = ({ detail: { id, network, ...rest } }: CustomEvent<IcrcCustomToken>) => {
+		const { id: networkId } = network;
+		const { [`${networkId.description}-${id.description}`]: current, ...tokens } = modifiedTokens;
```
peterpeterparker, line 84: Cool idea. Not sure, maybe it's too much and not useful but, we can also create a symbol. But maybe that's too much and not useful. Just sharing in case the question I asked my self would not be too dumb 😅.

```
const { [Symbol(`${networkId.description}-${id.description}`)]: current, ...tokens } = modifiedTokens;

peterpeterparker (approved):
LGTM, thx

src/frontend/src/icp-eth/components/tokens/IcManageTokens.svelte
AntonioVentilii-DFINITY: the only particularity with symbols is that apparently if you use `Object.keys` it returns an empty list, you should use `Object.getOwnPropertySymbols` instead

I found out yesterday searching, I cannot find the link anymore... but I tested and it really returned empty ([remember this comment?](https://github.com/dfinity/oisy-wallet/pull/1477#issuecomment-2176327858))


I really have no preference, it was just to reduce the code that changed, you tell me if you feel strong on that, @peterpeterparker

peterpeterparker: Agree with you, that was just an idea/question that crossed my mind.

PR #1478, opened 2024-06-18
feat: display exchange currency compact

peterpeterparker:
# Motivation

On one hand we have now less space to display balances, on the other, if you've got lots of money, numbers are hard to read anyway with some much zero in the dollars 😉.

So, let's display the balance in dollars in a compact way.

# Screenshots

Capture d’écran 2024-06-18 à 14 54 39
Capture d’écran 2024-06-18 à 14 53 54


AntonioVentilii-DFINITY (approved):
Good idea! Ahahah! I wish I could be part of the use case

PR #1486, opened 2024-06-19
feat(lint): created local rule to use isNullish and nonNullish

AntonioVentilii-DFINITY:
# Motivation

We would like to enforce the usage of functions `isNullish` and `nonNullish` in the code whenever possible.

# Changes

- Created custom local lint rule for `isNullish` usage.
- Created custom local lint rule for `nonNullish` usage.
- Fix issues created by new rules.

# Tests

It raised issues, correctly, and we fixed it.


src/frontend/src/eth/derived/erc20.derived.ts
```diff
@@ -116,7 +117,7 @@ export const enabledErc20TokensAddresses: Readable<ContractAddressText[]> = deri
 
 export const erc20UserTokensInitialized: Readable<boolean> = derived(
 	[erc20UserTokensStore],
-	([$erc20UserTokensStore]) => $erc20UserTokensStore !== undefined
+	([$erc20UserTokensStore]) => nonNullish($erc20UserTokensStore)
```
peterpeterparker, line 120: Are you sure this was not checked for `undefined` on purpose given that `null` in store is often used to express the fact that something was explicitely set to `null`?

src/frontend/src/icp-eth/services/cketh.services.ts
```diff
@@ -24,7 +24,7 @@ export const loadCkEthMinterInfo = async ({
 	const minterInfoInStore = get(ckEthMinterInfoStore);
 
 	// We try to load only once per session the helpers (ckETH and ckErc20) contract addresses
-	if (minterInfoInStore?.[tokenId] !== undefined) {
+	if (nonNullish(minterInfoInStore?.[tokenId])) {
```
peterpeterparker, line 27: ditto

src/frontend/src/lib/components/ui/Json.svelte
```diff
@@ -21,7 +22,7 @@
 		| 'undefined';
 
 	const getValueType = (value: unknown): ValueType => {
-		if (value === null) {
+		if (isNullish(value)) {
```
peterpeterparker, line 25: I think we want to display `undefined` here and the `=== null` is a choice

src/frontend/src/lib/derived/address.derived.ts
```diff
@@ -8,7 +8,7 @@ export const addressNotLoaded: Readable<boolean> = derived([addressStore], ([$ad
 );
 
 export const address: Readable<OptionAddress> = derived([addressStore], ([$addressStore]) =>
-	$addressStore === null ? null : $addressStore?.address
+	isNullish($addressStore) ? null : $addressStore?.address
```
peterpeterparker, line 11: ditto

src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -71,7 +71,7 @@ export const loadRouteParams = ($event: LoadEvent): RouteParams => {
 	const token = searchParams?.get('token');
 
 	const replaceEmoji = (input: string | null): string | null => {
-		if (input === null) {
+		if (isNullish(input)) {
```
peterpeterparker, line 74: not a call to action, just saying if interesting: input cannot be undefined here

src/frontend/src/routes/+layout.svelte
```diff
@@ -74,7 +74,7 @@
 		}
 
 		// We want to display a spinner until the authentication is loaded. This to avoid a glitch when either the landing page or effective content (sign-in / sign-out) is presented.
-		if ($authStore === undefined) {
+		if (isNullish($authStore)) {
```
peterpeterparker, line 77: ditto

peterpeterparker (commented):
Cool stuffs.

`[].find` or `[].filter` cannot return `null`. I'm fine using `isNullish` and `nonNullish` for those array operators but, if there is a way to improve the eslint rule, I have to say that I wont' be against neither - i.e. that would be actually nice.

AntonioVentilii-DFINITY:
@peterpeterparker ah! I didn't think that some of the conditions were in place with a purpose...

so, either I put the @lint-exception in place there, or I can just discard this PR, what do you think? for example, if we want to keep the rule of this pr we would do:

```typescript
// eslint-disable-next-line nullish-check -- it is ok to display undefined becasue.....
if (value === null) {
```

peterpeterparker:
> I didn't think that some of the conditions were in place with a purpose...

Yep. Checkout the definition of the store and how we handle errors in the services. On error we set `null`.

> so, either I put the @lint-exception

Lint-exception sounds good to me, I think it's a cool stuffs to have your eslint rule. 👍
We should just pay attention at the changes and not do a bulk change.

src/frontend/src/lib/components/ui/Json.svelte
AntonioVentilii-DFINITY: right! I am putting the eslint exception here

AntonioVentilii-DFINITY:
Since the code changed too much since I last reviewed this PR, I will open a new one (https://github.com/dfinity/oisy-wallet/pull/2355) and close this one

PR #1488, opened 2024-06-19
chore: change aria label of IcManageTokenToggle

AntonioVentilii-DFINITY:
# Motivation

Previous label was saying:

"Hide zero balances"

and now 

"Show or hide token"

### DIFF
--- a/src/frontend/src/lib/i18n/en.json
+++ b/src/frontend/src/lib/i18n/en.json
@@ -275,7 +275,8 @@
 			"hide_zero_balances": "Hide zero balances",
 			"all_tokens_with_zero_hidden": "All tokens with zero balance are currently hidden.",
 			"initializing": "Initializing...",
-			"updating_ui": "Updating the UI..."
+			"updating_ui": "Updating the UI...",
+			"show_hide_token": "Show or hide token"
 		},
 		"details": {
 			"title": "Token details",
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/i18n/en.json
```diff
@@ -275,7 +275,8 @@
 			"hide_zero_balances": "Hide zero balances",
 			"all_tokens_with_zero_hidden": "All tokens with zero balance are currently hidden.",
 			"initializing": "Initializing...",
-			"updating_ui": "Updating the UI..."
+			"updating_ui": "Updating the UI...",
+			"show_hide_token": "Show or hide token"
```
peterpeterparker, line 279: Shouldn't we replace globally `hide_zero_balances` by `show_hide_token` for UX consistency?

AntonioVentilii-DFINITY: the only "incorrect" usage I found was here, do you see it somewhere else?

peterpeterparker: Ah now I get it.

peterpeterparker (approved):
LGTM, thanks for the fix!

Note: if we want to get fancy, we can display an aria label depending of the state

```
aria-label={checked ? "Hide token" : "Add token"}
```

peterpeterparker (approved):
LGTM, thanks

PR #1489, opened 2024-06-19
lint: include rule no-delete-var

AntonioVentilii-DFINITY:
# Motivation

We want to include the ESLint rule `no-delete-var` ([doc](https://eslint.org/docs/latest/rules/no-delete-var))


peterpeterparker (approved):
Thanks

PR #1491, opened 2024-06-19
revert: moving back component IcManageTokenToggle back to icp folder

AntonioVentilii-DFINITY:
# Motivation

Since we will use a different toggle component for non-ICRC tokens, we move back the IcManageTokenToggle.


peterpeterparker (approved):
LGTM, thx

PR #1492, opened 2024-06-19
refactor: changed types in ManageTokens to use the generic Token type

AntonioVentilii-DFINITY:
# Motivation

To use ManageTokens in a more generic way, we change some token list types from IcrcCustomToken to Token. However, we still need an util to differentiate between them since we access specific props.

# Changes

- Created util `isIcrcCustomToken`.
- Changed types for lists `filteredTokens`, `tokens` and `modifiedTokens`.

# Tests

Local replica behavior unchanged.


src/frontend/src/icp/utils/icrc.utils.ts
```diff
@@ -131,3 +131,6 @@ export const buildIcrcCustomTokenMetadataPseudoResponse = ({
 		...(nonNullish(icon) ? [icon] : [])
 	];
 };
+
+export const isIcrcCustomToken = (token: Partial<IcrcCustomToken>): token is IcrcCustomToken =>
```
peterpeterparker, line 135: A token can be of type Icrc Custom if disabled too, so maybe either rename the util or revisit its logic.

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
```diff
@@ -150,7 +157,12 @@
 					{token.symbol}
 				</span>
 
-				<IcManageTokenToggle slot="action" {token} on:icToken={onToggle} />
+				<!-- We add to break it like this because it was giving a wrong rendering with the IF statement -->
```
peterpeterparker, line 160: What do you mean? "We ad to break it like" what?

AntonioVentilii-DFINITY: it is a typo: We **had** to break it...

but basically, if i put the IF statement by itself like this:

```svelte
{#if isIcrcCustomToken(token)}
  <IcManageTokenToggle slot="actions" {token} on:icToken={onToggle} />
{/if}
```

the layout result is wrong, and apparently it is really the IF statement: i tested with and without, and that was the difference

Screenshot 2024-06-19 at 09 27 13

peterpeterparker: Ah then you can remove the comment. That's standard Svelte 4 behavior. You indeed need a fragment for the slot if you use such a if.

peterpeterparker:
```
if
   div slot // Not allowed
   
fragment slot
   if
      div // Allowed
```

src/frontend/src/icp/utils/icrc.utils.ts
AntonioVentilii-DFINITY: yes, agreed, but that function just checks the existence of the properties (not their value) and that would identify the type

```typescript
export const isIcrcCustomToken = (token: Partial<IcrcCustomToken>): token is IcrcCustomToken =>
	nonNullish(token.ledgerCanisterId) && nonNullish(token.enabled);
```

shall i call it 

isTypeIcrcCustomToken

peterpeterparker: >  but that function just checks the existence

If the goals is to check the existence of the properties we should not test if they are undefined null or defined but rather check if they are present in my opinion.

For example:

```
ledgerCanisterId in token && indexCanisterId in token && enabled in token
```

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
AntonioVentilii-DFINITY: ah ok, i didn't know that... removing

 is there some doc on that for me to read? i swear I searched a lot but nothing explicitly saying that

peterpeterparker: Don't know where we learned it anymore spontaneously.

It's deprecated in Svelte v5 anyway 😜

PR #1493, opened 2024-06-19
feat: created component ManageTokenToggle to be used for generic token

AntonioVentilii-DFINITY:
# Motivation

We will need to have a generic toggle button to use for object Token. However we create the custom type DisplayToken, to facilitate the usage:

`export type DisplayToken = Token & { show?: boolean };`

# Changes

- Created type `DisplayToken`.
- Create `ManageTokenToggle` that is a copy of `IcManageTokenToggle` but without the specific ICRC features.

### DIFF
--- a/src/frontend/src/lib/types/token.ts
+++ b/src/frontend/src/lib/types/token.ts
@@ -25,3 +25,5 @@ export type RequiredToken = Required<Token>;
 export type OptionToken = Token | undefined | null;
 export type OptionTokenId = TokenId | undefined | null;
 export type OptionTokenStandard = TokenStandard | undefined | null;
+
+export type DisplayToken = Token & { show?: boolean };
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/types/token.ts
```diff
@@ -25,3 +25,5 @@ export type RequiredToken = Required<Token>;
 export type OptionToken = Token | undefined | null;
 export type OptionTokenId = TokenId | undefined | null;
 export type OptionTokenStandard = TokenStandard | undefined | null;
+
+export type DisplayToken = Token & { show?: boolean };
```
peterpeterparker, line 29: Not a strong opinion but, I would maybe declare it in a separate module. That way `Token` remains kind of scoped. Maybe a module `manage-token.ts`.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/types/token.ts
AntonioVentilii-DFINITY: on it

PR #1495, opened 2024-06-19
fix: function isIcrcCustomToken shall include icp standard

AntonioVentilii-DFINITY:
# Motivation

There was an issue with ICP due to PR #1492 : it was not considered as an ICRC custom token type.

### Before

Screenshot 2024-06-19 at 12 33 48

### After

Screenshot 2024-06-19 at 12 35 18

### DIFF
--- a/src/frontend/src/icp/utils/icrc.utils.ts
+++ b/src/frontend/src/icp/utils/icrc.utils.ts
@@ -132,5 +132,5 @@ export const buildIcrcCustomTokenMetadataPseudoResponse = ({
 	];
 };
 
-export const isIcrcCustomToken = (token: Partial<IcrcCustomToken>): token is IcrcCustomToken =>
+export const isCustomTokenIC = (token: Partial<IcrcCustomToken>): token is IcrcCustomToken =>
 	(token.standard === 'icp' || token.standard === 'icrc') && 'enabled' in token;
--- a/src/frontend/src/icp/utils/icrc.utils.ts
+++ b/src/frontend/src/icp/utils/icrc.utils.ts
@@ -132,5 +132,5 @@ export const buildIcrcCustomTokenMetadataPseudoResponse = ({
 	];
 };
 
-export const isCustomTokenIC = (token: Partial<IcrcCustomToken>): token is IcrcCustomToken =>
+export const icTokenContainsEnabled = (token: Partial<IcrcCustomToken>): token is IcrcCustomToken =>
 	(token.standard === 'icp' || token.standard === 'icrc') && 'enabled' in token;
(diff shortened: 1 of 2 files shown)


src/frontend/src/icp/utils/icrc.utils.ts
```diff
@@ -133,4 +133,4 @@ export const buildIcrcCustomTokenMetadataPseudoResponse = ({
 };
 
 export const isIcrcCustomToken = (token: Partial<IcrcCustomToken>): token is IcrcCustomToken =>
-	token.standard === 'icrc' && 'enabled' in token;
+	(token.standard === 'icp' || token.standard === 'icrc') && 'enabled' in token;
```
peterpeterparker, line 136: - I would rename the function to `isCustomTokenIC` then
- You can maybe shorten `["icp", "icrc"].includes(token.standard)` (if interesting and not sure TS likes it)?

AntonioVentilii-DFINITY: yes on the first, on the second shall be `["icp", "icrc"].includes(token.standard ?? '')` since it am getting the partial `IcrcCustomToken`

still ok?

peterpeterparker: I wonder if that will be solved with TypeScript 5.5 but, let's keep your implementation. Thanks for checking.

src/frontend/src/icp/utils/icrc.utils.ts
```diff
@@ -132,5 +132,5 @@ export const buildIcrcCustomTokenMetadataPseudoResponse = ({
 	];
 };
 
-export const isIcrcCustomToken = (token: Partial<IcrcCustomToken>): token is IcrcCustomToken =>
-	token.standard === 'icrc' && 'enabled' in token;
+export const isCustomTokenIC = (token: Partial<IcrcCustomToken>): token is IcrcCustomToken =>
```
peterpeterparker, line 135: Please accept my apologies for commenting again. I am really sorry, but I just had a second thought. The term isCustmoTokenIC is actually misleading because there is no "ICP custom token" in the backend and in Oisy, there are ERC20 and ICRC custom tokens.

Can we call it something like "objContainsEnabled" or another more descriptive name? Any better ideas would also be appreciated.

Again, sorry for the inconvenience.

AntonioVentilii-DFINITY: eheh indeed... it was bothering me a little the name too...

what about `icTokenContainsEnabled`?

peterpeterparker: Yeah way better thanks. Already approving so that you can go one without waiting on me.

peterpeterparker (approved):
Thanks!

PR #1496, opened 2024-06-19
Feature/chain fusion/include network tokens in manage modal

AntonioVentilii-DFINITY:
# Motivation

We include the list of all non IC tokens in component `ManageTokens`, respecting the visual order that the user sees in the single-view page (basically the ICRC tokens are last).
Furthermore we provide the ManageTokenToggle to the new tokens (for now disabled).

# Changes

- Created `allTokens` as merged list of `networkTokens` and `allIcrcTokens`. Note that we remove the duplicates, but IC tokens take priority. 
- Change index of loop EACH for `tokens`: it was giving duplicates indices, since the symbol may be repeated for different tokens on different networks (ICP native vs ICP Ethereum network).
- Use `ManageTokenToggle` for non-IC tokens.

# To-Do

- [ ] Remove ckLocal IC tokens from the list in case of Chain Fusion

# Tests

See below a local replica simulation:

Screen Recording 2024-06-19 at 13 16 20

### DIFF
--- a/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
@@ -159,7 +159,7 @@
 	</button>
 {:else}
 	<div class="container md:max-h-96 pr-2 pt-1 overflow-y-auto">
-		{#each tokens as token (token.symbol)}
+		{#each tokens as token (`${token.network.id.description}-${token.id.description}`)}
 			<Card>
 				{token.name}
 
--- a/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
@@ -160,7 +160,7 @@
 	</button>
 {:else}
 	<div class="container md:max-h-96 pr-2 pt-1 overflow-y-auto">
-		{#each tokens as token (`${token.network.id.description}-${token.id.description}`)}
+		{#each tokens as token (token.symbol)}
 			<Card>
 				{token.name}
 
--- a/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
@@ -20,6 +20,7 @@
 	import { replacePlaceholders } from '$lib/utils/i18n.utils';
 	import type { Token } from '$lib/types/token';
 	import { networkTokens } from '$lib/derived/network-tokens.derived';
+	import ManageTokenToggle from '$lib/components/tokens/ManageTokenToggle.svelte';
 
 	const dispatch = createEventDispatcher();
 
@@ -178,6 +179,8 @@
 				<svelte:fragment slot="action">
 					{#if icTokenContainsEnabled(token)}
 						<IcManageTokenToggle {token} on:icToken={onToggle} />
+					{:else}
+						<ManageTokenToggle {token} />
 					{/if}
 				</svelte:fragment>
 			</Card>
--- a/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
@@ -77,12 +77,13 @@
 
 	let filteredTokens: Token[] = [];
 	$: filteredTokens = isNullishOrEmpty(filterTokens)
-		? allIcrcTokens
-		: allIcrcTokens.filter(
-				({ name, symbol, alternativeName }) =>
-					name.toLowerCase().includes(filterTokens.toLowerCase()) ||
-					symbol.toLowerCase().includes(filterTokens.toLowerCase()) ||
-					(alternativeName ?? '').toLowerCase().includes(filterTokens.toLowerCase())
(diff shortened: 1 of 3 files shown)


AntonioVentilii-DFINITY:
P.S. I noticed a delay in the overflow menu, that is being solved with PR #1497

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
```diff
@@ -51,6 +53,22 @@
 		)
 	].sort(sortIcTokens);
 
+	let allTokens: Token[] = [];
+	$: allTokens = [
```
peterpeterparker, line 57: Can't we use the global store `$tokens` to get all the tokens?
Also shouldn't we filter test tokens?

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
```diff
@@ -60,12 +78,13 @@
...
-		? allIcrcTokens
-		: allIcrcTokens.filter(
-				({ name, symbol, alternativeName }) =>
-					name.toLowerCase().includes(filterTokens.toLowerCase()) ||
-					symbol.toLowerCase().includes(filterTokens.toLowerCase()) ||
-					(alternativeName ?? '').toLowerCase().includes(filterTokens.toLowerCase())
+		? allTokens
+		: allTokens.filter(
+				(token) =>
+					token.name.toLowerCase().includes(filterTokens.toLowerCase()) ||
```
peterpeterparker, line 84: We compare token with their ids. Is there a specific reason why not to here?

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
```diff
@@ -141,7 +160,7 @@
 	</button>
 {:else}
 	<div class="container md:max-h-96 pr-2 pt-1 overflow-y-auto">
-		{#each tokens as token (token.symbol)}
+		{#each tokens as token (`${token.network.id.description}-${token.id.description}`)}
```
peterpeterparker, line 163: Is this related to this PR?

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
AntonioVentilii-DFINITY: I actually kept the same searching logic: when the user search for a token, it checks the name, the symbol and, if there is, the alternative name

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
AntonioVentilii-DFINITY: Created PR #1498

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
peterpeterparker: I see, thanks for the feedback.

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
AntonioVentilii-DFINITY: I used `networkTokens` since it will keep the same behavior if we select ICP network: if it's Chain Fusion, `$tokens` is the same as `$networkTokens` but without the test tokens (Sepolia for example).

However, you are right that the ICP test tokens appear there, I wanted to do a separate PR to remove them.

Do you prefer here?

peterpeterparker: Is that the expected behavior from Artem and Mary? I actually would expect that when I access "Manage Tokens" I can always manage all my tokens regardless of the selected network. 🤷‍♂️

peterpeterparker: Note that my comment above about testnet was incorrect I think. For ERC20 I can also set custom ERC20 on Sepolia.

AntonioVentilii-DFINITY: > Is that the expected behavior from Artem and Mary? I actually would expect that when I access "Manage Tokens" I can always manage all my tokens regardless of the selected network. 🤷‍♂️

Ah, that I didn't know, I thought the user should manage only the tokens in the selected network. Let me ask them

AntonioVentilii-DFINITY: I wrote to Artem and he said

Ok, it looks like it indeed m

> akes sense to show only the tokens related to the selected network. Is it possible?
> Maybe some small hint under the search would help with the context?
> e.g.
> Managing token for Internet Computer
> Managing token for ChainFusion

he made edits in figma

https://www.figma.com/design/2Wi7w1kenBrOhKgKo4iD3Z/OISY-Layouts?node-id=1354-5273&t=JO5QzcGQyyYI2EhX-0

peterpeterparker: Thanks for double checking. I also quickly discussed with him to be sure I understand everything. Let's do that then.

peterpeterparker (approved):
LGTm, thx

PR #1497, opened 2024-06-19
fix: now the overflow menu closes after opening a new modal

AntonioVentilii-DFINITY:
# Motivation

Somewhere along the past weeks, I removed incorrectly the behavior that would close the Overflow Menu once clicked on it to open a modal.

### Before

Screen Recording 2024-06-19 at 13 54 39

### After

Screen Recording 2024-06-19 at 13 54 58


peterpeterparker (approved):
LGTM, thanks

I guess ultimately, we will have only one single `ManageTokensMenuButton` component.

PR #1498, opened 2024-06-20
fix: change index key of loop EACH in ManageTokens

AntonioVentilii-DFINITY:
# Motivation

Since there could be tokens with same symbol once we merge into a single list all of them (for example native ICP and Ethereum ICP), we change the key used in the EACH loop to avoid duplicates error.

# Changes

Changed key of loop to `${token.network.id.description}-${token.id.description}`.


peterpeterparker (approved):
LGTM, thx

PR #1499, opened 2024-06-20
refactor: created new variables for symbols of Ethereum networks

AntonioVentilii-DFINITY:
# Motivation

It will be useful to have separate variables for the string representation of Ethereum networks symbol. Otherwise we would need to use the `.description` prop. Also, for consistency with the other networks.


peterpeterparker (approved):
LGTM, thx

PR #1500, opened 2024-06-20
feat: AddTokenReview component has its related functions inside to be more re-useful

AntonioVentilii-DFINITY:
# Motivation

To make the component `AddTokenReview` more usable in other modals too (not strictly the ETH modal), we move the related functions inside the component.

# Changes

- Moved `save` function from `AddTokenModal` to `AddTokenReview`.
- Created bindable variables modal and saveProgressStep in AddTokenReview.
- Instead of `save` executing the `close` function, it dispatches `icClose` event that is now listened by `AddTokenModal`.
- Removed unused imports and variables in `AddTokenModal`.

# Tests

Behavior unchanged for Ethereum network in local replica.

### DIFF
--- a/src/frontend/src/eth/components/tokens/AddTokenModal.svelte
+++ b/src/frontend/src/eth/components/tokens/AddTokenModal.svelte
@@ -2,18 +2,8 @@
 	import { WizardModal, type WizardStep, type WizardSteps } from '@dfinity/gix-components';
 	import AddTokenReview from '$eth/components/tokens/AddTokenReview.svelte';
 	import AddTokenForm from '$eth/components/tokens/AddTokenForm.svelte';
-	import type { Erc20Metadata } from '$eth/types/erc20';
-	import { isNullishOrEmpty } from '$lib/utils/input.utils';
-	import { toastsError } from '$lib/stores/toasts.store';
 	import { i18n } from '$lib/stores/i18n.store';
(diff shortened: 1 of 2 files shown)


peterpeterparker (changes requested):
We cannot / should not move the "save" function within the "review" step because that particular component is destroyed when the token is saved, as the progress screen is displayed during this event.

AntonioVentilii-DFINITY:
Strange, it is working either way... you mean that for good practice we should not, right?

peterpeterparker:
No, I really mean that I had issues because I tried this pattern as well, and I deliberately chose a pattern where the main action is executed in a persistent component, in the parent, such as the modal. The function might run until completion, though sometimes I've observed it being destroyed prematurely. When it does finish, some context might be lost.

That's why I believe we shouldn't change this pattern.

What you can do is extract the save function into a service. Then, either the service can decide what to do if it's dealing with Ethereum or IC, or alternatively, the modal can call either Service A or Service B, depending on the context.

AntonioVentilii-DFINITY:
Ah understood! 

However the function is quite dependent on stores and local variables, so it will become too much of a verbose thing, I think... I will do differently: I will add the function in `ManageTokens`

i will close this one

peterpeterparker:
> quite dependent on stores and local variables

Yeah pass their values as variable of the service.

PR #1501, opened 2024-06-20
feat: AddTokenForm can be used as non-first step in a modal now

AntonioVentilii-DFINITY:
# Motivation

Since we would like to use `AddTokenForm` as non-first step of a Wizard modal (for example for the single ManageTokens), it changes to have an optional variable to define if it is the first step of the modal or not.
I think that the default should be that it is not the first step, that is why the variable defaults to `false`.

# Changes

- Add variable `isFirstWizardStep` that defaults to `false`.
- Change AddTokenModal to use the variable as `true`.

### DIFF
--- a/src/frontend/src/eth/components/tokens/AddTokenForm.svelte
+++ b/src/frontend/src/eth/components/tokens/AddTokenForm.svelte
@@ -6,6 +6,8 @@
 	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
 
 	export let contractAddress = '';
+
+	// TODO: remove this variable as soon as we deprecate AddTokenModal that is specific to Ethereum network
 	export let isFirstWizardStep: boolean = false;
 
 	let invalid = true;
--- a/src/frontend/src/eth/components/tokens/AddTokenForm.svelte
+++ b/src/frontend/src/eth/components/tokens/AddTokenForm.svelte
@@ -6,6 +6,7 @@
 	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
 
 	export let contractAddress = '';
+	export let isFirstWizardStep: boolean = false;
 
 	let invalid = true;
 	$: invalid = isNullishOrEmpty(contractAddress);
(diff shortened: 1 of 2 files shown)


peterpeterparker:
Why do we need this? Isn't `AddTokenModal` not use anymore and `AddTokenForm` about to be integrated in the "Manage Token > Add custom token" wizard - i.e. in `ManageTokensModal` step `import`?

AntonioVentilii-DFINITY:
Correct! But since for now we will need it in both (until we remove AddTokenModal) this is the way I thougth to use it in both. Otherwise I shall create a new AddTokenForm that is the exact same, but with a different "back" button.

You prefer it?

peterpeterparker:
My point is that we cannot deploy main to production in its current state. Therefore, I’m unsure if it's worth the effort to maintain backward compatibility with the "Manage tokens" feature. We might as well deprecate / delete `AddTokenModal` and proceed with the integration.

However, if you believe it's not too much effort to maintain both, then let's go ahead with that approach. Please add a `TODO:` comment so we can easily locate and remove the relevant code later.

AntonioVentilii-DFINITY:
Maybe it is better that I create a new component that is the copy, so that when we deprecate `AddTokenModal` it should be easy to find the unused components too and the old `AddTokenForm` will fall in that category.
What do you think? And if you agree, any opinion on the new component name? `ECR20AddTokenForm`  (to be similar to `IcAddTokenForm`)?

peterpeterparker:
Maybe, but reviewing the work won't be easy because it will be difficult to distinguish between new code and copied code.

So, I would rather follow your suggestion: add a variable, assuming we don't have to add too many variables, or go bold and deprecate the feature while keeping what we need. Once again, in my opinion, we cannot deploy main in production as long as the "Manage Tokens" feature is not implemented anyway.

Both works for me, your call as you are implementing the feature. If you pick the first, please add a TODO and I'll approve the PR.

AntonioVentilii-DFINITY:
I think I can control it better with the first option. I included the TO-DO, have a look

peterpeterparker (approved):
LGTM, thx

PR #1502, opened 2024-06-20
feat: created new component AddTokenByNetwork

AntonioVentilii-DFINITY:
# Motivation

This new component allows the user to select the network to add a specific token (for now either IC or Ethereum).
It uses the already existing components `IcAddTokenForm` and `AddTokenForm`.

Preview:

Screenshot 2024-06-20 at 08 06 20
Screenshot 2024-06-20 at 08 06 28


src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
```diff
@@ -0,0 +1,40 @@
...
+	import { i18n } from '$lib/stores/i18n.store';
+	import { Dropdown, DropdownItem } from '@dfinity/gix-components';
+	import { networksMainnets } from '$lib/derived/networks.derived';
+	import { ETHEREUM_NETWORK_SYMBOL, ICP_NETWORK_SYMBOL } from '$env/networks.env';
+	import IcAddTokenForm from '$icp/components/tokens/IcAddTokenForm.svelte';
+	import AddTokenForm from '$eth/components/tokens/AddTokenForm.svelte';
+	import { networkId } from '$lib/derived/network.derived';
+	import { isNullish } from '@dfinity/utils';
+
+	export let dropdownNetwork: string | undefined;
```
peterpeterparker, line 11: - any particular reason to prefix this with "dropdown" keyword? we generally do not use UI component prefix sofar if I'm not wrong
- any particular reason to use a `string` instead of using the network id?

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
```diff
@@ -0,0 +1,40 @@
...
+	export let tokenData: Record<string, string>;
+
+	let ledgerCanisterId: string = '';
```
peterpeterparker, line 14: There shouldn't be a callback that reset the variable if the network is switched or UX wise we want to keep the previous values in memories when the user select networks back and forth?

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
```diff
@@ -0,0 +1,40 @@
...
+	let contractAddress: string = '';
+
+	$: tokenData = {
+		ledgerCanisterId,
+		indexCanisterId,
+		contractAddress
+	};
+</script>
+
+{#if isNullish($networkId)}
```
peterpeterparker, line 25: Assuming the user can manage all tokens regardless of the network, I would say we do not need this check.

If Mary and Artem say the user select tokens for the specific network, then instead of not displaying the dropdown why not display it either as a paragraph (label+value) or disabling the dropdown that way we provide the user a feedback to the user about which network will be use.

But, I would still tend to allowing user to always administrate their all tokens regardless of the network.

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
```diff
@@ -0,0 +1,40 @@
...
+	<div class="stretch pt-8">
```
peterpeterparker, line 26: I think `stretch` is for the all content - i.e. `IcAddTokenForm` and `AddTokenForm` should be within that div.

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
```diff
@@ -0,0 +1,40 @@
...
+		<label for="selectedNetwork" class="font-bold px-4.5">{$i18n.tokens.manage.text.select_network}:</label>
+		<Dropdown name="selectedNetwork" bind:selectedValue={dropdownNetwork}>
```
peterpeterparker, line 28: For consistency and UX reason, I would display undefined when the network has not been select yet.

See for example `SendNetwork`.

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
```diff
@@ -0,0 +1,40 @@
...
+			{#each $networksMainnets as network}
+				<DropdownItem value={network.id.description ?? ''}>{network.name}</DropdownItem>
+			{/each}
+		</Dropdown>
+	</div>
+{/if}
+
+{#if dropdownNetwork === ICP_NETWORK_SYMBOL}
```
peterpeterparker, line 36: Have you try to add a `in:fade` effect here? Does it providers a smoother experience when the network is modified?

peterpeterparker:
In addition to the comment, noticed that the spacing is a bit off but it might be related to the `stretch` class comment I provided above.

341386878-e9c579ab-9a00-4e49-ad8a-169e63d99a04
341386853-6b1806dd-60c3-4fe4-885f-8e583f5d0145

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
AntonioVentilii-DFINITY: - i wanted to call it `selectedNetwork`, but I didn't want it to confound it with the derived `selectedNetwork`, do you have some idea of a better name?
- I tried symbol, but `Dropdown` component accepts only string as `selectedValue` prop and it can be bound only to the real variable not the prop: we cannot do `bind:selectedValue={dropdownNetwork.description}`

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
AntonioVentilii-DFINITY: According to Artem, we should show filter by network, but I agree with you to leave the choice. The only thing is that it will default to the selected network is any, right?

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
AntonioVentilii-DFINITY: corrected

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
AntonioVentilii-DFINITY: good point! included

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
peterpeterparker: - `managedNetwork` or `customTokensNetwork`?
- Ah yeah right. I really don't say it's a good pattern but, should we do something similar to `SendNetwork` component using `name`?

peterpeterparker: Actually isn't the feature really close to what the `SendNetwork` does, is there a chance to have or reuse a same component for both or that's too different? Just an idea that crossed my mind right now, not sure it's a good idea...

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
peterpeterparker: I would do:

- If the modal is open on "Chain Fusion", then the default value is `undefined`.
- If the modal is open with a selected network, then the default is the network and the value cannot be changed

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
peterpeterparker: I would not default to IC.

AntonioVentilii-DFINITY: cool

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
AntonioVentilii-DFINITY: cool!

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
AntonioVentilii-DFINITY: I was looking at it right now

I would prefer to use the same, but it has some variables that are specific to the Send process (`destination` for example) and some function too (onDestinationAddressInput for example)

I can adapt this one to use the same logic for selection thoough... (and same format, style, strings, etc)

peterpeterparker: Sounds cool, thanks

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
peterpeterparker: Works for me in terms of UX, I'm just a bit concern about the state because the parent/modal component will ends up knowing about values that should not be persisted.

e.g. user select IC, enter ledger ID, then select Ethereum, enter contract address, finish the process and when the `save` is called, there is a valid ledger ID in memory.

But I guess that if we really pay attention...

AntonioVentilii-DFINITY: if you want (in another PR) we can improve this doing something like modifying slightly the icNext event in `AddTokenByNetwork`, for example

```svelte
const icNext = () => {
  contractAddress = '';
  dispatch('icNext');
  }

<IcAddTokenForm on:icBack on:icNext={icNext} bind:ledgerCanisterId bind:indexCanisterId />
```

and same logic for `AddTokenForm`

peterpeterparker: something like that or a clean onDestroy of the component would be nice indeed

peterpeterparker: in another pr sure

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
AntonioVentilii-DFINITY: done

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
```diff
@@ -0,0 +1,63 @@
...
+
+<div class="stretch pt-8">
+	<label for="network" class="font-bold px-4.5">{$i18n.tokens.manage.text.network}:</label>
+
+	<div id="network" class="mb-4 mt-1 pt-0.5">
+		<Dropdown name="network" bind:selectedValue={networkName}>
+			<option disabled selected value={undefined} class="hidden"
+				><span class="description">{$i18n.tokens.manage.placeholder.select_network}</span></option
+			>
+			{#each $networksMainnets as network}
```
peterpeterparker, line 40: Why not Sepolia as well? User can set up custom tokens on this particular chain.

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
```diff
@@ -0,0 +1,63 @@
...
+				<DropdownItem value={network.name}>{network.name}</DropdownItem>
+			{/each}
+		</Dropdown>
+	</div>
+
+	<div class="mt-8">
```
peterpeterparker, line 46: should we wrap this in a `if nonNullish(network)` just for the state of the art of not having a useless div in the DOM?

AntonioVentilii-DFINITY: yes!

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
AntonioVentilii-DFINITY: no particular reason to be honest, ii think it was an overview

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
```diff
@@ -0,0 +1,66 @@
...
+	import { isNetworkIdICP } from '$lib/utils/network.utils';
+	import { isNetworkIdEthereum } from '$lib/utils/network.utils.js';
+
+	export let network: Network | undefined;
+	export let tokenData: Record<string, string>;
+
+	let networkName: string | undefined = network?.name;
+	$: networkName,
+		(network = nonNullish(networkName)
+			? $networksMainnets.find(({ name }) => name === networkName)
```
peterpeterparker, line 19: Should it be `$networks` here too?

peterpeterparker (approved):
LGTM, thanks

Added one last comment, please have a look at it before merge.

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
AntonioVentilii-DFINITY: yes! good spot!

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
AntonioVentilii-DFINITY: created PR #1507 for this

PR #1503, opened 2024-06-20
feat: included AddTokenReview in component ManageTokensModal

AntonioVentilii-DFINITY:
# Motivation

We add the component `AddTokenReview` (specific to Ethereum network) in the generic modal `ManageTokensModal`.

# Changes

- Copied function `save` from Ethereum component `AddTokenModal` and pasted it in `ManageTokensModal`.
- Created new variables `erc20Metadata` and `erc20ContractAddress` to pass to `AddTokenReview`.
- Included `AddTokenReview` and created conditional rendering based on network.

### DIFF
--- a/src/frontend/src/icp-eth/components/tokens/ManageTokensModal.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/ManageTokensModal.svelte
@@ -14,6 +14,16 @@
 	import { toastsError } from '$lib/stores/toasts.store';
 	import { saveCustomTokens } from '$icp/services/ic-custom-tokens.services';
 	import type { IcrcCustomToken } from '$icp/types/icrc-custom-token';
+	import { selectedNetwork } from '$lib/derived/network.derived';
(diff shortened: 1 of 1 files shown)


src/frontend/src/icp-eth/components/tokens/ManageTokensModal.svelte
```diff
@@ -90,6 +100,69 @@
 		}
 	};
 
+	let erc20Metadata: Erc20Metadata | undefined;
+
+	const saveErc20Token = async () => {
```
peterpeterparker, line 105: If we are already there, I would rather like to have this in a service or at least everything we can extract. Or are you planing to refactor that afterwards?

AntonioVentilii-DFINITY: i was planning to do it afterwrds actually, so that we can move step by step, it seems the best way to have smaller PRs

what do you think? I refactor first?

AntonioVentilii-DFINITY:
You are right, better refactor first... closing this

PR #1504, opened 2024-06-20
chore: set type ProgressStepsAddToken to variable saveProgressStep

AntonioVentilii-DFINITY:
# Motivation

Variable `saveProgressStep` should not be a generic string but part of enum `ProgressStepsAddToken`.


peterpeterparker (approved):
Thx

PR #1506, opened 2024-06-20
refactor: extract save function from component AddTokenModal

AntonioVentilii-DFINITY:
# Motivation

Since we would like to have a common component for the AddTokenReview, we first extract the save function and use it as service.

# Changes

- Mirror function `save` to `saveErc20Contract`.
- Adapt component `AddTokenModal`.

# Tests

Manual test on replica showed unchanged behavior.


src/frontend/src/eth/services/erc20.services.ts
```diff
@@ -83,3 +87,83 @@ export const loadErc20Contracts = async (): Promise<{ success: boolean }> => {
...
+	onError
+}: {
+	contractAddress: string;
+	metadata: Erc20Metadata | undefined;
+	updateSaveProgressStep: (step: ProgressStepsAddToken) => void;
+	modalNext: () => void;
+	onSuccess: () => void;
+	onError: () => void;
+}): Promise<void> => {
+	const $i18n = get(i18n);
```
peterpeterparker, line 106: For consistency reason, expect i18n, can you pass those state's values as variables of the function?

peterpeterparker (approved):
Already approving, just on minor suggestion

src/frontend/src/eth/services/erc20.services.ts
AntonioVentilii-DFINITY: on it

PR #1507, opened 2024-06-20
fix: tokenData are passed now from AddTokenByNetwork consistent with the chosen network

AntonioVentilii-DFINITY:
# Motivation

To be more consistent with the selected network from the dropdown, the data returned are set when the component is destroyed.

# Changes

Prop `tokenData` is defined with `onDestroy`.


src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
```diff
@@ -23,11 +24,13 @@
 	let indexCanisterId: string;
 	let erc20ContractAddress: string;
 
-	$: tokenData = {
-		ledgerCanisterId,
-		indexCanisterId,
-		erc20ContractAddress
-	};
+	onDestroy(() => {
+		tokenData = isNetworkIdICP(network?.id)
```
peterpeterparker, line 28: Can you add a comment about why we do this as it's quite an original pattern?

peterpeterparker (approved):
LGTM, thx

AntonioVentilii-DFINITY:
@peterpeterparker check the new version, maybe it is better?

p.s. I tried to do inline, 

```typescript
$: tokenData = isNetworkIdICP(network?.id)
		? { ledgerCanisterId, indexCanisterId }
		: isNetworkIdEthereum(network?.id)
			? { erc20ContractAddress }
			: {};		
```
but it was giving an error because it was interpreting this as assigning undefined keys

src/frontend/src/icp-eth/components/tokens/AddTokenByNetwork.svelte
AntonioVentilii-DFINITY: check now, I updated the logic, maybe it is better?

peterpeterparker (approved):
I like it better!

Thanks

PR #1508, opened 2024-06-20
refactor: extract save function from component ManageTokensModal

AntonioVentilii-DFINITY:
# Motivation

Since we would like to use `ManageTokensModal` in a more generic way, we first extract the save function and use it as service.

# Changes

- Mirror function `save` to `saveIcrcCustomToken `.
- Adapt component `ManageTokensModal`.

# Tests

Manual test on replica showed unchanged behavior.


src/frontend/src/icp/services/ic-custom-tokens.services.ts
```diff
@@ -41,3 +46,53 @@ export const saveCustomTokens = async ({
...
+	onError
+}: {
+	tokens: Pick<IcrcCustomToken, 'enabled' | 'version' | 'ledgerCanisterId' | 'indexCanisterId'>[];
+	updateSaveProgressStep: (step: ProgressStepsAddToken) => void;
+	modalNext: () => void;
+	onSuccess: () => void;
+	onError: () => void;
+}) => {
+	const $i18n = get(i18n);
+	const $authStore = get(authStore);
```
peterpeterparker, line 64: can you pass the `identity` as parameter of the function for consistency reason?

peterpeterparker (approved):
LGTM, minor suggestion

src/frontend/src/icp/services/ic-custom-tokens.services.ts
AntonioVentilii-DFINITY: doing it, and do you want to keep the same name for the `updateSaveProgressStep` and call it `progress`?

peterpeterparker: Ah yeah why not, it's generally what I used so far

AntonioVentilii-DFINITY: cool

PR #1509, opened 2024-06-20
refactor: as temporary precaution, we disable non-ICRC manage toggle

AntonioVentilii-DFINITY:
# Motivation

As temporary precaution we disable all non-ICRC manage toggle, until we are ok with the complete flow.

# To-Do

- [ ] Remove hard-coded condition to disable the toggle


peterpeterparker (commented):
Maybe rather:

```
// TODO: set to false and uncomment once we are OK with flow of show/hide ERC20 tokens
let disabled = true;
// $: disabled = !["icrc", "erc20"].includes(token.category);
```

AntonioVentilii-DFINITY:
done, but probably we will need to revisit the logic to disable, since we haven't defined yet which one we enable/disable about etherem/erc20 , right

peterpeterparker:
Definitely we should review before merging but, I think there is a good chance it's correct 🤞

peterpeterparker (approved):
LGTM, thx

PR #1510, opened 2024-06-20
feat: we turn on the listener to event icShowOrHideToken for ManageTokenToggle

AntonioVentilii-DFINITY:
# Motivation

Since we disabled all non-ICRC token toggle with PR #1509 , we are safe to turn on the event listener for `ManageTokenToggle`


peterpeterparker (approved):
LGTM, thx

PR #1513, opened 2024-06-21
build: Update Rust version

sa-github-api:
A new Rust version available.


peterpeterparker (approved):
We have increased the stack size in PR #1515. Therefore we can upgrade Rust.

PR #1519, opened 2024-06-21
feat: adapted component ManageTokensModal to be able to treat both IC and Ethereum tokens

AntonioVentilii-DFINITY:
# Motivation

We now proceed to merge the flow of adding Erc20 tokens into the general modal `ManageTokensModal`.

# Changes

- Substituted component `IcAddTokenForm` with component `AddTokenByNetwork` to have the network choice (default to current network, if defined).
- Included use of ERC20 component `AddTokenReview`.
- Create additional variables necessary for the merged flow.

# Tests

On local replica, adding tokens wroked as expected.

### DIFF
--- a/src/frontend/src/icp-eth/components/tokens/ManageTokensModal.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/ManageTokensModal.svelte
@@ -7,10 +7,17 @@
 	import { i18n } from '$lib/stores/i18n.store';
 	import { addTokenSteps } from '$lib/constants/steps.constants';
 	import InProgressWizard from '$lib/components/ui/InProgressWizard.svelte';
-	import IcAddTokenForm from '$icp/components/tokens/IcAddTokenForm.svelte';
 	import { authStore } from '$lib/stores/auth.store';
 	import { saveIcrcCustomToken } from '$icp/services/ic-custom-tokens.services';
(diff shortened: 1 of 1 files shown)


src/frontend/src/icp-eth/components/tokens/ManageTokensModal.svelte
```diff
@@ -50,6 +57,20 @@
 		]);
 	};
 
+	const saveErc20Token = async () => {
+		await saveErc20Contract({
+			contractAddress: erc20ContractAddress,
+			metadata: erc20Metadata,
+			chainId: $selectedChainId,
```
peterpeterparker, line 64: I think we have an issue here on "Chain Fusion" because`$selectedChainId` and `$selectedEthereumNetwork` are either undefined or ethereum (did not checked the code). I guess they should be related to the network the user has selected when they added the contract manually no?

AntonioVentilii-DFINITY: yes! it was an overview on my side, correcting it

peterpeterparker: Good idea

peterpeterparker (approved):
LGTM, thx

PR #1521, opened 2024-06-21
chore: remove param chainId from saveErc20Contract because is redundant

AntonioVentilii-DFINITY:
# Motivation

In service `saveErc20Contract`, variable `chainId` can be extracted from param `network` since is of type `EthereumNetwork`.

# Changes

- Removed param `chainId` from `saveErc20Contract`.
- Adapted usage of function `saveErc20Contract`.


peterpeterparker (approved):
LGTM, thx

PR #1522, opened 2024-06-21
refactor: renamed and moved button component that is part of the TokensMenu list

AntonioVentilii-DFINITY:
# Motivation

Since it is used for all the networks, the component is moved and renamed to `ManageTokensMenuButton`.


peterpeterparker (approved):
Thx

PR #1526, opened 2024-06-21
fix: token list in ManageTokens shows ICRC tokens when required

AntonioVentilii-DFINITY:
# Motivation

I noticed an incongruence: we discussed with Artem that if a network is selected, then the tokens that can be managed are only the one related to it.
However the list was always including ICRC tokens too.
We fix it with this PR.

# Changes

Changed list of allTokens in component ManageTokens to include ICRC tokens only for IC network and Chain Fusion (undefined network).

# Tests

Test on local replica show correct behavior for each selected network.


peterpeterparker (approved):
LGTm, thx

PR #1550, opened 2024-06-24
feat(backend): set user token

peterpeterparker:
# Motivation

Given that we are required to extend the dApp to allow users to enable and disable ERC20 tokens as well, we decided to take a simplistic approach - i.e., extending the current entities with a similar mechanism as the one we implemented for `CustomToken` by adding an "enabled" flag to `UserToken`. We discussed using the "custom token" types, but extending the stable tree map we already have allows us to avoid a migration at this time that does not necessarily add value for end users. This new approach also means that user tokens (ERC20) won't be added and deleted anymore in the state, which is why this PR deprecates the related function and adds a setter, again similar to the custom token approach.

# Changes

# Tests

### DIFF
--- a/scripts/test.backend.sh
+++ b/scripts/test.backend.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 
 POCKET_IC_SERVER_VERSION=3.0.1
-OISY_UPGRADE_VERSION=v0.0.13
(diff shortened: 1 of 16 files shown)


peterpeterparker:
I restarted with a clean branch in #1554.

PR #1632, opened 2024-07-01
build: Update ckErc20 Tokens

sa-github-api:
Modifications have been made to the list of ckErc20 tokens deployed by the Orchestrator.


peterpeterparker (approved):
ckOCT

bitdivine (commented):
I would recommend formatting this file so that the diff is shown properly.  What do you think, @peterpeterparker ?

peterpeterparker:
It's actually on purpose inlined - not formatted - @bitdivine, that way we load the less bytes possible at runtime.

Ehhamid611:
bot-tokens-ckerc20-update

PR #1638, opened 2024-07-01
feat: support for ckOCT

peterpeterparker:
# Motivation

We want to support ckOCT. This PR provides the quick configuration.


peterpeterparker:
There is a long [thread](https://dfinity.slack.com/archives/C05M7P123HR/p1719500811650029?thread_ts=1719496763.144909&cid=C05M7P123HR) on Slack which explains why we do it this way or why I think it's better, notably in terms of UI/UX on the IC, to pre-bundle those information.

PR #1639, opened 2024-07-01
chore: remove unused to_string()

peterpeterparker:
# Motivation

We do not need to cast `CALLER.to_string()` in our test. Just `CALLER` is fine.


bitdivine (approved):
thanks!

PR #1640, opened 2024-07-01
feat: Pretty-print tokens.json

bitdivine:
# Motivation
The list of SNS tokens is pre-fetched and stored in a file as one long JSON line.  An interesting question is why it is pre-fetched, but that aside, when it is updated, the diffs shown in the PRs are impossible to compare.  Formatting the JSON would make it possible to see what has changed.

Ditto for erc20 tokens.

# Changes
- In the command that updates the list of SNS tokens, format the output JSON.
- In packages.json, run the command to format the code after updating tokens.
- Run the commands to update the list of tokens.

# Tests
I ran this locally and confirmed that the JSON was pretty-printed:
```
npm run build:tokens-sns
```
Likewise for ck tokens.


bitdivine:
I would expect source code to be designed for codegov and the build for machines.  :-D  I would expect that the build minimizes and compresses the file.  I don't see the file in the build directory: `find build/ | grep erc`  Is it inside one of the chunks?

bitdivine:
I looked for ckOCT.  It actually appears in three chunks, always minimized.  The three chunks are, in my case:
```
./build/_app/immutable/chunks/index.BzFh_Swp.js.gz
./build/_app/immutable/workers/icrc-wallet.worker-tkBRa3T_.js.gz
./build/_app/immutable/workers/ckbtc-update-balance.worker-Bvu0-GOH.js.gz
```

peterpeterparker:
You are right. It looks like because we are importing the JSON as modules, the bundler takes care of inlining it within the JS code. So, we are good.

Can you then remove also those two lines in `.prettierignore`:

```
src/frontend/src/env/tokens.sns.json
src/frontend/src/env/tokens.ckerc20.json
```

And run `npm run format` and commit the formatted files within this PR?
Or you rather like to provide that in a separate PR?

bitdivine:
I guess that the formatter should probably be run after updating the tokens otherwise the CI job to update the tokens may fail CI due to unexpected formatting.

Done.  Added formatting in packages.json to the token update commands and run the token update commands.

peterpeterparker (approved):
Thanks!

peterpeterparker:
Allowind myself to merge it too, as I am about to deploy on staging.

PR #1641, opened 2024-07-01
test: list tokens before and after set

peterpeterparker:
# Motivation

As recommended in #1554, list tokens before and after set in tests.


bitdivine (approved):
Thanks!

PR #1642, opened 2024-07-01
refactor (backend): Simpler search

bitdivine:
# Motivation
The current code parses every single Eth address when comparing.  This adds a little unnecessary complexity, assuming that Eth addresses have a unique string representation.

# Changes
- Compare eth addresses as strings

# Tests
Existing CI should suffice, as the expected behaviour is unchanged.


peterpeterparker (approved):
LGTM, thx

peterpeterparker:
Allowed myself to merge the PR in case I decide to deploy staging today - i.e. to be sure it's already in.

PR #1682, opened 2024-07-03
chore(dfx): mkdir in docker image

peterpeterparker:
# Motivation

Try to solve another dfx issue when building the Docker image:

> #22 72.95 Error: Failed to save pem: Failed to write PEM to file: Failed to encrypt PEM file: /home/apprunner/.config/dfx/identity/___temp___minter/identity.pem.encrypted
> #22 72.95 Using identity: "minter".
> #22 72.95 Error: Identity must exist: Identity minter does not exist at '/home/apprunner/.config/dfx/identity/minter/identity.json'.


peterpeterparker:
This PR is not the solution for the issue. Just keeping it as draft as a reminder that we should solve the problem, if we can.

peterpeterparker:
We do not need this. It does not work anyway.

PR #1701, opened 2024-07-08
style(frontend): network dropdown in manage tokens will display 'Select network' only once

AntonioVentilii-DFINITY:
# Motivation

The Network selection dropdown when adding a new token will show the "Select Network" option as disabled, but only the first time, when the network is still not selected.

# Changes

### Before

Screenshot 2024-07-08 at 10 58 12

### After

Screenshot 2024-07-08 at 10 57 34

Screenshot 2024-07-08 at 10 57 50


peterpeterparker (approved):
LGTM, thanks!!

PR #1707, opened 2024-07-08
refactor(frontend): ERC20 twin token address comes from JSON

AntonioVentilii-DFINITY:
# Motivation

There is a redundancy of data: the ERC20 tokens have their address both in the env files and in the JSON file of the respective ck token.

To avoid repetition, we will fetch the contract address directly from the ckErc20 JSON file, since the correct flow implies that the so-called ERC20 "twin tokens" have their env files **ONLY** when there is the respective ckErc20 token.

However, to be type-safe, we need to remove the possibility of `undefined` data in the JSON file.

# Changes

- Remove `z.undefined()` from type `envTokens`.
- Create constants for Twin Token symbol for each ERC20 token.
- Use maps `ckErc20Production` and `ckErc20Staging` to get the token address.

# Tests

Unchanged behavior on local replica.


AntonioVentilii-DFINITY:
@peterpeterparker 

I thought a little more about the discussion about this PR. Do you think it is ok to raise an error when the address is missing?

```typescript
export const getTwinTokenAddress = (twinTokenSymbol: string): string => {
	const address = (ckErc20Production[twinTokenSymbol] ?? ckErc20Staging[twinTokenSymbol])
		?.erc20ContractAddress;
	if (isNullish(address)) {
		throw new Error(`Address not found for token: ${twinTokenSymbol}`);
	}
	return address;
};
```

peterpeterparker:
> Do you think it is ok to raise an error when the address is missing?

Spontaneously I fear that this would lead to a whitescreen of death at runtime if applied to building the env file. That's why currently all environment variable are skipped if not properly set.

AntonioVentilii-DFINITY:
Right!

Another solution could be to create constants of Twin Tokens and then make them as type for the safe-parsing of the JSON:

```typescript
// in the env file
export const USDC_TWIN_TOKEN = 'ckUSDC';
export const SEPOLIA_USDC_TWIN_TOKEN = 'ckSepoliaUSDC';
...

// in the type file
const tokenSymbolsArray = [USDC_TWIN_TOKEN, SEPOLIA_USDC_TWIN_TOKEN, ...........] as const;
type TokenSymbol = typeof tokenSymbolsArray[number];
const envTokenSymbol = z.enum(tokenSymbolsArray);
...
```

What do you think?

AntonioVentilii-DFINITY:
In my last commit, I tried to remove the possible `undefined` value from the JSON, but it raised an issue during test...

peterpeterparker:
I saw that 😉

AntonioVentilii-DFINITY:
So! Found the issue: there was a circular reference so that `ckErc20Production` and `ckErc20Staging` were not loaded properly when used in the tokens env files.

To solve it, I moved them in a separate file.

However, maybe it is batter to do this as first step in another PR?

@peterpeterparker , what do you think?

peterpeterparker:
Darn, circular dependency 😱

Yeah would be great to split that in another PR, agree.

AntonioVentilii-DFINITY:
Created PR #1729 to avoid the circular reference

AntonioVentilii-DFINITY:
@peterpeterparker with the last commit, it should be ready for review

peterpeterparker:
Thanks. I tried it out. The issue I have with this solution is that the linter is fine with an invalid type, but the app actually fails at runtime.

For example:

```
export const LINK_TOKEN: RequiredErc20Token = {
	id: LINK_TOKEN_ID,
	network: ETHEREUM_NETWORK,
	standard: 'erc20',
	category: 'default',
	name: 'ChainLink Token',
	symbol: LINK_SYMBOL,
	decimals: LINK_DECIMALS,
	icon: link,
	// HERE "YOLO"
	address: ckErc20Production['YOLO'].erc20ContractAddress,
	exchange: 'erc20',
	twinTokenSymbol: LINK_TWIN_TOKEN_SYMBOL
};
```

If fine for the linter

Capture d’écran 2024-07-10 à 13 16 17

But crash at runtime

Capture d’écran 2024-07-10 à 13 16 35

peterpeterparker:
I would like to preserve both type and runtime safety.

image

AntonioVentilii-DFINITY:
Ah wow! For me it didn't happen, it didn't raise an error

AntonioVentilii-DFINITY:
I think I am at an impasse eheh!

The only solution I can think of to match runtime and types: fallback the address to an empty string

```typescript
address: ckErc20Production[PEPE_TWIN_TOKEN_SYMBOL]?.erc20ContractAddress ?? ''
```

and then check for non empty string using `isNullishOrEmpty` every time the tokens are used (that honestly are not that much)

peterpeterparker:
Thanks for trying implementing the changes we discussed.

I'll have another look and we can think about a solution.

Let's put this task in DRAFT meanwhile.

peterpeterparker:
@AntonioVentilii-DFINITY what do you think of PR 👉 #1761
It addresses the same duplication.

AntonioVentilii-DFINITY:
For me it makes sense, if we accept that the ERC20 tokens variables are not "complete" statically.

I will close this one.

PR #1713, opened 2024-07-08
feat(frontend): save ckErc20 logos when running the update

AntonioVentilii-DFINITY:
# Motivation

We would like to save the logo of every ckErc20 token, fetching it automatically from the ICRC Canister.


scripts/build.tokens.ckerc20.mjs
```diff
@@ -78,6 +79,37 @@ const ORCHESTRATOR_PRODUCTION_ID = Principal.fromText('vxkom-oyaaa-aaaar-qafda-c
...
+
+const saveTokenLogo = async (canisterId, name) => {
+	const logoName = name.toLowerCase().replace('ck', '').replace('sepolia', '');
+	const file = join(LOGO_FOLDER, `${logoName}.svg`);
+
+	if (existsSync(file)) {
+		return;
+	}
+
+	const agent = await createAgent({
```
peterpeterparker, line 92: should we move the `agent` level up? that way the same variable can be used in `saveTokenLogo` and `orchestratorInfo`?

scripts/build.tokens.ckerc20.mjs
```diff
@@ -78,6 +79,37 @@ const ORCHESTRATOR_PRODUCTION_ID = Principal.fromText('vxkom-oyaaa-aaaar-qafda-c
...
+		identity: new AnonymousIdentity(),
+		host: 'https://icp-api.io'
+	});
+
+	const { metadata } = IcrcLedgerCanister.create({
+		agent,
+		canisterId
+	});
+
+	const data = await metadata({});
```
peterpeterparker, line 102: It's the default but, to make it explicit I would specify `{ certified: true }` to express the fact that we make an update call. WDYT?

scripts/build.tokens.ckerc20.mjs
```diff
@@ -78,6 +79,37 @@ const ORCHESTRATOR_PRODUCTION_ID = Principal.fromText('vxkom-oyaaa-aaaar-qafda-c
...
+
+	const logoData = data.find((item) => item[0] === 'icrc1:logo')[1].Text;
```
peterpeterparker, line 104: I would handle the fact that `icrc1:logo` is `undefined`. It should not happens in production but, it's probably safer given that per definition the logo is not mandatory. 

Particularly for staging.

Maybe we just stacktrace a warning like `console.warning("No token found for ...")`?

scripts/build.tokens.ckerc20.mjs
```diff
@@ -89,6 +121,13 @@ const findCkErc20 = async () => {
 	};
 
 	writeFileSync(join(DATA_FOLDER, 'tokens.ckerc20.json'), JSON.stringify(tokens, jsonReplacer, 8));
+
+	for (const [name, { ledgerCanisterId }] of Object.entries({
```
peterpeterparker, line 125: NTH: `await Promise.allSettled(Object.entrie().map(...)`

On one hand, that's an opiniated feedback (😅), on the other it might be interesting to continue to generate the logo if just one ledger is not accessible?

scripts/build.tokens.ckerc20.mjs
AntonioVentilii-DFINITY: yes, makes sense!

scripts/build.tokens.ckerc20.mjs
AntonioVentilii-DFINITY: done!

scripts/build.tokens.ckerc20.mjs
AntonioVentilii-DFINITY: yes, always better to have the check

scripts/build.tokens.ckerc20.mjs
AntonioVentilii-DFINITY: indeed, it shouldn't happen, but better to put it there

peterpeterparker (approved):
LGTM, thx

PR #1718, opened 2024-07-08
feat(frontend): add Ethereum Mainnet Shiba Inu token

AntonioVentilii-DFINITY:
# Motivation

Since `ckSHIB` was approved in production, we add Shiba Inu token (mainnet) to the list.


peterpeterparker (approved):
LGTm, thx

PR #1719, opened 2024-07-08
feat(frontend): include Shiba Inu in the ICRC token list

AntonioVentilii-DFINITY:
# Motivation

After updating the JSON file with the new ckERC20 tokens, and after having created the Shiba Inu env variables for the ERC20 token, we can include ckSHIB in the ICRC token list.


peterpeterparker (approved):
LGTM, thanks

PR #1733, opened 2024-07-10
feat(env): Add Proof Of Humanity issuer to environment

lmuntaner:
# Motivation

We would like to present the Proof Of Humanity credentials in Oisy.

In this PR, I introduce the issuer into the enviornment for Oisy. I added the entry to dfx.json pointing to a dummy issuer that will return any requested credential. As canister id, I used the canister id that Modclub gave us and we can use to test it in mainnet.

# Changes

* Update Internet Identity version to support VCs.
* New entry in dfx.json `poh_issuer`.
* Deploy the `poh_issuer` in `deploy.sh`.
* Add issuer canister id to constants.ts to confirm that it's added by the scripts.

# Tests

I even tested that we can present a credential using the JS library and passing all the necessary details.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -16,13 +16,24 @@
 		},
 		"internet_identity": {
 			"type": "custom",
-			"candid": "https://github.com/dfinity/internet-identity/releases/download/release-2023-11-17/internet_identity.did",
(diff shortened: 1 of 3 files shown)


lmuntaner:
@peterpeterparker @AntonioVentilii-DFINITY ready for review

peterpeterparker (approved):
LGTM, thx

PR #1736, opened 2024-07-10
feat(frontend): pin specific tokens on top

AntonioVentilii-DFINITY:
# Motivation

According to a recent requirement, the list of tokens that shall be pinned on top in every token list is (in order):

1. ICP
2. BTC (when it will be enabled)
4. ETH
5. ckBTC
6. ckETH
7. ckUSDC

Since we will sort them further in the future, we create a specific list of pinned tokens.

# Changes

- Create list of pinned tokens.
- Create function to pin those on top.
- Create new derived for the sorted tokens (it will be used for further sorting).


src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -0,0 +1,32 @@
...
+import { BTC_MAINNET_TOKEN } from '$env/tokens.btc.env';
+import { ETHEREUM_TOKEN, ICP_TOKEN } from '$env/tokens.env';
+import type { Token } from '$lib/types/token';
+import { nonNullish } from '@dfinity/utils';
+
+const PINNED_TOKENS_ORDER = [
+	ICP_TOKEN,
+	BTC_MAINNET_TOKEN,
+	ETHEREUM_TOKEN,
+	{ id: Symbol('ckBTC'), network: ICP_NETWORK },
```
peterpeterparker, line 11: This feels like it could be "dangerous". We are using symbols because we want to be absolutely sure we are referencing the same unique token across the entire application. By creating a particular symbol here, we are creating a new reference. This is a pattern that is prone to leading to errors, so I'm not sure what the exact goal is?

peterpeterparker (commented):
Just a single comment to start with, not really had a look to the rest.

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: I agree, I was trying to get the list of ckBTC, ckETH and ckUSDC here, but we don't really have it somewhere... shall i create it as Twin Tokens?

peterpeterparker: Ideally we should not create new information. I think we probably have already enough data. There is probably a way. For example I would use the existing `ICRC_CHAIN_FUSION_DEFAULT_LEDGER_CANISTER_IDS`.

AntonioVentilii-DFINITY: done, have a look

src/frontend/src/icp/derived/icrc.derived.ts
```diff
@@ -105,6 +107,17 @@ export const sortedIcrcTokens: Readable<IcrcCustomToken[]> = derived(
 	([$icrcTokens]) => $icrcTokens.sort(sortIcTokens)
 );
 
+/**
+ * The list of Icrc tokens that are default for Chain Fusion, in the order provided by the static list.
+ */
+export const icrcChainFusionDefaultTokens: Readable<IcrcCustomToken[]> = derived(
+	[icrcTokens],
```
peterpeterparker, line 114: `icrcTokens` contains all tokens. For performance reason, could we derive the information from a subset such `icrcDefaultTokens`?

src/frontend/src/icp/derived/icrc.derived.ts
```diff
@@ -105,6 +107,17 @@ export const sortedIcrcTokens: Readable<IcrcCustomToken[]> = derived(
...
+	([$icrcTokens]) =>
+		ICRC_CHAIN_FUSION_DEFAULT_LEDGER_CANISTER_IDS.map((canisterId) =>
```
peterpeterparker, line 116: Curiosity question, is this implementation more performant than direcly filtering `$icrcTokens`?

```
$icrcTokens.filter(ledgerCanisterId => ICRC_CHAIN_FUSION_DEFAULT_LEDGER_CANISTER_IDS.includes(ledgerCanisterId))
```

Really just a curiosity question.

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -0,0 +1,27 @@
...
+import { nonNullish } from '@dfinity/utils';
+
+export const pinTokensAtTop = ({
+	$tokens,
+	$tokensToPin
+}: {
+	$tokens: Token[];
+	$tokensToPin: Token[];
+}): Token[] => {
+	const pinnedTokens = $tokensToPin
```
peterpeterparker, line 11: For performance reason, can't we do this in a single shot / iteration? 

The `pinnedTokens` are equals to the `tokensToPin` no?

src/frontend/src/icp/derived/icrc.derived.ts
AntonioVentilii-DFINITY: The only reason I didn't do that was to respect the order of `ICRC_CHAIN_FUSION_DEFAULT_LEDGER_CANISTER_IDS`, since I was not sure if it is the same of `icrcTokens`. For sure it would be more performant your way, but it does not guarantee the order given by `ICRC_CHAIN_FUSION_DEFAULT_LEDGER_CANISTER_IDS`

src/frontend/src/icp/derived/icrc.derived.ts
AntonioVentilii-DFINITY: I wasn't sure that `icrcDefaultTokens` would contain `ICRC_CHAIN_FUSION_DEFAULT_LEDGER_CANISTER_IDS`

I will change it

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: Yes and no, in the sense that the `Token` object coming from `tokens` is for sure more complete than the ones in `tokensToPin`. It is not guaranteed that the object is exactly the same in the two lists.

Theoretically, `tokensToPin` could be even a simple list of `id` and `networkId`. So, the function `pinTokensAtTop` should receive a list to re-order and the way to order it.

I thought of doing like that. Do you prefer that the list `tokensToPin` has already the complete `Token` objects?

peterpeterparker: > Yes and no, in the sense that the Token object coming from tokens is for sure more complete than the ones in tokensToPin.

Why? Which fields?

Typescript wise they are the same so if there are actually not we should be careful and more descriptive. You implementation then make sense but, we should solve this.

On the contrary, if they are the same, then I would try to improve the perf.

AntonioVentilii-DFINITY: understood; indeed the list tokensToPin should be better. What if:

```typescript
export type TokenToPin = Pick<Token, 'id'> & { network: Pick<Network, 'id'> };


const tokensToPin: Readable<TokenToPin[]> = derived(
	[icrcChainFusionDefaultTokens],
...


export const pinTokensAtTop = ({
	$tokens,
	$tokensToPin
}: {
	$tokens: Token[];
	$tokensToPin: TokenToPin[];
}): Token[] => {
...
```

peterpeterparker: Yeah sounds cool!

In this PR or another PR? What do you rather like?

AntonioVentilii-DFINITY: better in this PR I think

src/frontend/src/icp/derived/icrc.derived.ts
AntonioVentilii-DFINITY: all good this one @peterpeterparker ? does it make sense?

peterpeterparker (approved):
LGTM, thanks

PR #1737, opened 2024-07-10
feat(frontend): request market cap to Coingecko in the same request of price

AntonioVentilii-DFINITY:
# Motivation

As request for sorting, we will need the market cap of the ERC20 tokens. So we request it to Coingecko.

NOTE: for sorting we don't really need the market cap for ICP, BTC and ETH because the logic of sorting them (or their ck* tokens) is different.

### DIFF
--- a/src/frontend/src/lib/services/exchange.services.ts
+++ b/src/frontend/src/lib/services/exchange.services.ts
@@ -29,7 +29,8 @@ export const exchangeRateERC20ToUsd = async (
 	simpleTokenPrice({
 		id: 'ethereum',
 		vs_currencies: 'usd',
-		contract_addresses: contractAddresses.map(({ address }) => address.toLowerCase())
+		contract_addresses: contractAddresses.map(({ address }) => address.toLowerCase()),
+		include_market_cap: true
 	});
 
 export const syncExchange = (data: PostMessageDataResponseExchange | undefined) =>


src/frontend/src/lib/services/exchange.services.ts
```diff
@@ -29,7 +29,8 @@ export const exchangeRateERC20ToUsd = async (
 	simpleTokenPrice({
 		id: 'ethereum',
 		vs_currencies: 'usd',
-		contract_addresses: contractAddresses.map(({ address }) => address.toLowerCase())
+		contract_addresses: contractAddresses.map(({ address }) => address.toLowerCase()),
+		include_market_cap: true
```
peterpeterparker, line 33: What does this flag? Does it provide more information? Do we use it somewhere?

Can you either add a TODO about what will be done or consumes the information?
Also do we need types?

AntonioVentilii-DFINITY: This would be just the first step, so that we would use the market cap in another PR to sort the tokens.

For the types: we already have the types in place:

```typescript
export interface CoingeckoSimplePrice {
	usd: number;
	usd_market_cap?: number;
	usd_24h_vol?: number;
	usd_24h_change?: number;
	last_updated_at?: number;
}
```

peterpeterparker: Ah if the type is already there, then it's a bit more comprehensive. That said, does setting `include_market_cap` to `true` implies that the type `usd_market_cap` becomes optional anymore?

peterpeterparker: I think we should have types that are in sync with our third party APIs. So if the type becomes mandatory by adding a flag, I would update the type at the same time but, no strong opinion.

peterpeterparker: Your call.

AntonioVentilii-DFINITY: make sense, let me change it, but i will do it in a separate PR, to make it type-safe linked to the parameters

AntonioVentilii-DFINITY: create PR #1738 for this

PR #1738, opened 2024-07-10
chore(frontend): improve response types for Coingecko requests

AntonioVentilii-DFINITY:
# Motivation

After PR #1737 , the response type of function `exchangeRateERC20ToUsd` returns the prop `usd_market_cap` as `number` (and not possibly `undefined`). To be type-safe, we create a new type specific to this function.

# Changes

- Create type `CoingeckoSimpleTokenPriceResponse` that remove optionality of `usd_market_cap`.
- Use the type for function `exchangeRateERC20ToUsd`.
- Adapt function that fetches prices to have variable type on response.

### DIFF
--- a/src/frontend/src/lib/rest/goincecko.rest.ts
+++ b/src/frontend/src/lib/rest/goincecko.rest.ts
@@ -1,7 +1,10 @@
 import type {
+	CoingeckoResponse,
 	CoingeckoSimpleParams,
+	CoingeckoSimplePrice,
 	CoingeckoSimplePriceParams,
 	CoingeckoSimplePriceResponse,
+	CoingeckoSimpleTokenPrice,
 	CoingeckoSimpleTokenPriceParams,
 	CoingeckoSimpleTokenPriceResponse
 } from '$lib/types/coingecko';
@@ -21,7 +24,7 @@ export const simplePrice = async ({
 	ids,
 	...rest
 }: CoingeckoSimplePriceParams): Promise<CoingeckoSimplePriceResponse | null> =>
-	fetchCoingecko<CoingeckoSimplePriceResponse>({
+	fetchCoingecko<CoingeckoSimplePrice>({
 		endpointPath: 'simple/price',
 		queryParams: joinParams([
 			`ids=${Array.isArray(ids) ? ids.join(',') : ids}`,
@@ -41,7 +44,7 @@ export const simpleTokenPrice = async ({
 	contract_addresses,
 	...rest
 }: CoingeckoSimpleTokenPriceParams): Promise<CoingeckoSimpleTokenPriceResponse | null> =>
-	fetchCoingecko<CoingeckoSimpleTokenPriceResponse>({
+	fetchCoingecko<CoingeckoSimpleTokenPrice>({
 		endpointPath: `simple/token_price/${id}`,
 		queryParams: joinParams([
 			`contract_addresses=${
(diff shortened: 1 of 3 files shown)


AntonioVentilii-DFINITY:
@peterpeterparker 

I actualy have done a first version a bit more complex that relies directly on the value of input parameters for the props of the output.
For example: if `include_market_cap` is `true`, then `usd_market_cap` is not `undefined`; if `include_24hr_vol` is `true`, then `usd_24h_vol` is not `undefined`; etc etc.

You can see the commit here: [
commit d9d836ffbbc1714a04af76c8659657d87cd44df2](https://github.com/dfinity/oisy-wallet/pull/1738/commits/d9d836ffbbc1714a04af76c8659657d87cd44df2)

I think it is overkill, but theoretically it is the consistent way of doing this. Let me know if you prefer that commit or this one

src/frontend/src/lib/types/coingecko.ts
```diff
@@ -53,3 +53,12 @@ export interface CoingeckoSimplePrice {
 }
 
 export type CoingeckoSimplePriceResponse = Record<CoingeckoCoinsId | string, CoingeckoSimplePrice>;
+
+export interface CoingeckoSimpleTokenPrice extends CoingeckoSimplePrice {
```
peterpeterparker, line 57: isn't `Omit<CoingeckoSimplePrice, 'usd_market_cap'>`

or better

```
export type CoingeckoSimpleTokenPrice = Omit<CoingeckoSimplePrice, 'usd_market_cap'> & Required<Pick<CoingeckoSimplePrice, 'usd_market_cap'>
```

src/frontend/src/lib/types/coingecko.ts
```diff
@@ -53,3 +53,12 @@ export interface CoingeckoSimplePrice {
...
+	usd_market_cap: number;
+}
+
+export type CoingeckoSimpleTokenPriceResponse = Record<
```
peterpeterparker, line 61: Just an idea. What do you think?

```
export type CoingeckoResponse<T> = Record<CoingeckoCoinsId | string, T>;
```

which you can use in both `CoingeckoSimplePriceResponse` and `CoingeckoSimpleTokenPriceResponse`

```
export type CoingeckoSimpleTokenPriceResponse = CoingeckoResponse<CoingeckoSimpleTokenPrice>
```

that way you can declare `fetchCoingecko` as returning a type extending response

```
const fetchCoingecko = async <T extends >({
  ...
  }): Promise<CoingeckoResponse<T> | null> => {
```

and modify caller to pass the inner type for T

```
export const simpleTokenPrice = async ({
	id,
	contract_addresses,
	...rest
}: CoingeckoSimpleTokenPriceParams): Promise<CoingeckoSimpleTokenPriceResponse | null> =>
	fetchCoingecko<CoingeckoSimpleTokenPrice>({
```

The advantage is that `fetchCoingecko` returns become a bit more descriptive.

src/frontend/src/lib/rest/goincecko.rest.ts
```diff
@@ -50,13 +51,13 @@ export const simpleTokenPrice = async ({
 		])
 	});
 
-const fetchCoingecko = async ({
+const fetchCoingecko = async <T>({
```
peterpeterparker, line 54: Maybe?

```
const fetchCoingecko = async <T extends CoingeckoSimplePrice | CoingeckoSimpleTokenPrice>({
```

src/frontend/src/lib/types/coingecko.ts
AntonioVentilii-DFINITY: agreed! one doubt: why would `fetchCoingecko` have a type extending? can it be the same if:

```typescript
const fetchCoingecko = async <T>({
	endpointPath,
	queryParams
}: {
	endpointPath: string;
	queryParams: string;
}): Promise<CoingeckoResponse<T> | null> => {
...
```

AntonioVentilii-DFINITY: ah! i saw your next comment!

src/frontend/src/lib/rest/goincecko.rest.ts
AntonioVentilii-DFINITY: yes, more descriptive indeed!

peterpeterparker (approved):
LGTM, thx

PR #1739, opened 2024-07-11
fix(frontend): missing new Coingecko Response types in exchange stores

AntonioVentilii-DFINITY:
# Motivation

I just noticed that in PR #1738 , I forgot to update the types in the stores that use Coingecko rest wrappers.

I am not using the dynamic type `T extends CoingeckoSimplePriceResponse | CoingeckoSimpleTokenPriceResponse` because the stores effectively provide a record of possibly both types, and not selectively only one of them.

### DIFF
--- a/src/frontend/src/lib/stores/exchange.store.ts
+++ b/src/frontend/src/lib/stores/exchange.store.ts
@@ -1,11 +1,17 @@
-import type { CoingeckoSimplePriceResponse } from '$lib/types/coingecko';
+import type {
+	CoingeckoSimplePriceResponse,
+	CoingeckoSimpleTokenPriceResponse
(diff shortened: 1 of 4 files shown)


peterpeterparker (commented):
LGTM, thanks

PS.: Should we create a type for the union to avoid duplicating it?

```
type CoingeckoResponses = CoingeckoSimplePriceResponse | CoingeckoSimpleTokenPriceResponse;
```

AntonioVentilii-DFINITY:
agreed! done

peterpeterparker (approved):
I actually thought I already approved the PR but, I just commented it earlier. Sorry for that.

Coolio, super! Thanks!

PR #1740, opened 2024-07-11
feat(frontend): include a list of ERC20 tokens that will soon have the ckERC20 token

AntonioVentilii-DFINITY:
# Motivation

There is a pipeline of ERC20 tokens that will be onboarded with the respective ckERC20 token. However, it was requested to already include the Ethereum contracts in the list of available token.

We include them as tokens which details are fetched at runtime.

We need to remember to remove them once the respective ckERC20 token is approved, so that the normal flow of onboarding can function properly.

### DIFF
--- a/src/frontend/src/env/tokens.erc20.env.ts
+++ b/src/frontend/src/env/tokens.erc20.env.ts
@@ -10,9 +10,77 @@ import type { EthereumNetwork } from '$eth/types/network';
 import { mapAddressStartsWith0x } from '$icp-eth/utils/eth.utils';
 import type { TokenId } from '$lib/types/token';
 
+// TODO: remember to remove the ERC20 from here once the ckERC20 is implemented. Following the normal flow, the ERC20 variables should be created on a separate file.
+
+const ERC20_CONTRACT_ADDRESS_WBTC: Erc20Contract = {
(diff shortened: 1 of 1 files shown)


peterpeterparker (commented):
1. Does few of those contract addresses work also on Sepolia as well? 

For example, Wrapped Ether could be useful. That's one of the token supported by Uniswap which I often use when I test the integration with WalletConnect.

2. In addition to the addresses, we need their logo. Once logo added, the related mapper function `mapErc20Icon` must be extended. Did you plan to solve this in another PR?

AntonioVentilii-DFINITY:
1. they are not directly the same contract in sepolia, but I can try and find them if you prefer. The issue is that I was unable to find it on the official sites, or github pages. Only searching for the token + sepolia, I can find a sepolia.etherscan page. However, I would not rely on that. Do you have a better source?

2. Yes, you are right. I have the logo ready here. Do you prefer in the same PR?

peterpeterparker (approved):
1. No worries then, thanks for checking

2. Another PR is fine. Without logo a color is derived so it does not arm to have those yet on main and the logo later.

LGTM, thanks

AntonioVentilii-DFINITY:
2. done with PR #1742

PR #1741, opened 2024-07-11
feat(frontend): Add service to request credentials and store them

lmuntaner:
# Motivation

We want to request the Proof Of Humanity credential to give users rewards on their ck tokens.

In this PR, I introduce the service that will take care of requesting the credential 

# Changes

* Install `@dfinity/verifiable-credentials`.
* Add app constants for POH and II origin.
* Add error message in auth key for when requesting credentials fails.
* Add new store userProfileStore using initStorateStore and types (that will be removed when we implement them in the backend which is still pending to be approved by Max).
* New service `requestPohCredential` to request credentials and store it in userProfileStore.

# Tests

Unit tests still pending.

I tested manually by starting the flow from settings and it was working.


lmuntaner:
@peterpeterparker @AntonioVentilii-DFINITY could you take an initial look?

* Is the PR size ok? Should it be bigger or smaller?
* Regarding testing, should I add a test for the new service only which mocks the imported helper? Should I test the store?

Thanks!

AntonioVentilii-DFINITY:
For me it seems all good. I imagine this is just the first of a series of PR, since it has some placeholders for a successful query, right?

However, I defer to @peterpeterparker for a deeper review.

src/frontend/src/lib/constants/app.constants.ts
```diff
@@ -9,6 +11,12 @@ export const INTERNET_IDENTITY_CANISTER_ID = LOCAL
 	? import.meta.env.VITE_LOCAL_INTERNET_IDENTITY_CANISTER_ID
 	: undefined;
 
+export const INTERNET_IDENTITY_ORIGIN = LOCAL
```
peterpeterparker, line 14: There is no II on staging  actually - i.e. here staging === mainnet

src/frontend/src/lib/constants/app.constants.ts
```diff
@@ -17,6 +25,12 @@ export const POH_ISSUER_CANISTER_ID = LOCAL
 			? import.meta.env.VITE_IC_POH_ISSUER_CANISTER_ID
 			: undefined;
 
+export const POH_ISSUER_ORIGIN = nonNullish(POH_ISSUER_CANISTER_ID)
```
peterpeterparker, line 28: Should we have `ic0.app` in some constant?

src/frontend/src/lib/services/request-poh-credential.services.ts
```diff
@@ -0,0 +1,82 @@
...
+import {
+	requestVerifiablePresentation,
+	type VerifiablePresentationResponse
+} from '@dfinity/verifiable-credentials/request-verifiable-presentation';
+import { get } from 'svelte/store';
+
+// This credential type is defined by the issuer Decide AI
+const POH_CREDENTIAL_TYPE = 'ProofOfUniqueness';
+
+export const requestPohCredential = async ({
```
peterpeterparker, line 20: Nth: We do not have a linter for this but I think we do not have much function that returns a `boolean`. We mostyl use more descriptive results like `Promise<{result?: "ok" | "error", data?: XYZ}>` or `Promise<{success: boolean}`. Particularly given the function is named `requestPohCredential`, it would be a bit more descriptive. WDYT?

src/frontend/src/lib/services/request-poh-credential.services.ts
```diff
@@ -0,0 +1,82 @@
...
+	credentialSubject
+}: {
+	credentialSubject: Principal;
+}): Promise<boolean> => {
+	const i18Copy = get(i18n);
```
peterpeterparker, line 25: I know that code comes from II just by reading `i18Copy` 😉.

We mostly directly inline access `get(i18n)` or we destruct what we need.

src/frontend/src/lib/services/request-poh-credential.services.ts
```diff
@@ -0,0 +1,82 @@
...
+				},
+				credentialSubject
+			},
+			onError() {
+				toastsError({
+					msg: { text: i18Copy.auth.error.error_requesting_poh_credential }
+				});
+				reject();
+			},
+			onSuccess(response: VerifiablePresentationResponse) {
```
peterpeterparker, line 56: is there a way to extract the `onSuccess` in a separate function for readability or it's a bit tricky because we need to pass resolve?

Just asking for reading the PR, maybe in the editor it's easier to follow

src/frontend/src/lib/services/request-poh-credential.services.ts
```diff
@@ -0,0 +1,82 @@
...
+				if ('Ok' in response) {
+					// TODO: GIX-2646 Add credential to backend and load user profile
+					const fakeCredentialSummary = {
```
peterpeterparker, line 59: - It's called `fake...`?
- Can we set a type given that we build the object here? Might be useful in the future if some more fields gets addded and need to be populated.

src/frontend/src/lib/services/request-poh-credential.services.ts
```diff
@@ -0,0 +1,82 @@
...
+						credential_type: POH_CREDENTIAL_TYPE,
+						verified_date_timestamp: BigInt(Date.now()),
```
peterpeterparker, line 61: I guess there is a little chance that the `Date.now` provides different value but, should we assign `Date.now()` to a variable and use that variable to assign the value, that way we are certain that `verified_date_timestamp`, `created_timestamp` and `updated_timestamp` have same timestamp and `expire_date_timestamp` is exactly calculated based on that particular timestamp?

peterpeterparker:
> Is the PR size ok? 

Ok for me 👍

> Regarding testing, should I add a test for the new service only which mocks the imported helper? Should I test the store?

Our current agreement with @AntonioVentilii-DFINITY is that we should push our selves to make those E2E happens has we, or at least me, want to bet the most on E2E - i.e. more E2E than unit - and we currently provide unit tests when we introduce something new like particular utils.

That being said, I bet we are not a push in term of unit tests at the api and service level neither.

src/frontend/src/lib/services/request-poh-credential.services.ts
lmuntaner: This is temporary. This object won't be created here. That's why I added the comment with the TODO.

src/frontend/src/lib/constants/app.constants.ts
lmuntaner: Done!

src/frontend/src/lib/constants/app.constants.ts
lmuntaner: Yes, makes sense. Done!

src/frontend/src/lib/services/request-poh-credential.services.ts
lmuntaner: Makes sense, done!

src/frontend/src/lib/services/request-poh-credential.services.ts
lmuntaner: Hehe, I destructured it.

src/frontend/src/lib/services/request-poh-credential.services.ts
lmuntaner: Agree, done!

src/frontend/src/lib/services/request-poh-credential.services.ts
lmuntaner: I moved it to the helper. This is temporary, no object will be created here because it will come from the backend.

lmuntaner:
> > Is the PR size ok?
> 
> Ok for me 👍
> 
> > Regarding testing, should I add a test for the new service only which mocks the imported helper? Should I test the store?
> 
> Our current agreement with @AntonioVentilii-DFINITY is that we should push our selves to make those E2E happens has we, or at least me, want to bet the most on E2E - i.e. more E2E than unit - and we currently provide unit tests when we introduce something new like particular utils.
> 
> That being said, I bet we are not a push in term of unit tests at the api and service level neither.

I like the idea of betting on e2e tests as well. That test might be added once I have the full feature only though.

peterpeterparker (approved):
LGTM, thx

PR #1743, opened 2024-07-12
feat(backend): Add types to Rust and Candid

lmuntaner:
# Motivation

We want Oisy to become a relying party and store credentials for users.

In this PR, I introduce the new endpoints needed for the new functionality along with the new Rust and Candid types.

# Changes

* Add endpoints and types to candid interface.
* Add endpoints to the backend returning mock data.
* Add types to the types file.

# Tests

No new functionality was added yet. Therefore, no new tests were a

### DIFF
--- a/src/backend/src/lib.rs
+++ b/src/backend/src/lib.rs
@@ -2,7 +2,6 @@ use crate::assertions::{assert_token_enabled_is_some, assert_token_symbol_length
 use crate::guards::{caller_is_allowed, caller_is_not_anonymous};
 use crate::token::{add_to_user_token, remove_from_user_token};
 use candid::{CandidType, Deserialize, Nat, Principal};
-use ic_cdk::api::time;
 use core::ops::Deref;
 use ethers_core::abi::ethereum_types::{Address, H160, U256, U64};
 use ethers_core::types::transaction::eip2930::AccessList;
@@ -12,6 +11,7 @@ use ic_cdk::api::management_canister::ecdsa::{
     ecdsa_public_key, sign_with_ecdsa, EcdsaCurve, EcdsaKeyId, EcdsaPublicKeyArgument,
     SignWithEcdsaArgument,
 };
+use ic_cdk::api::time;
 use ic_cdk_macros::{export_candid, init, post_upgrade, query, update};
 use ic_stable_structures::{
     memory_manager::{MemoryId, MemoryManager, VirtualMemory},
@@ -24,9 +24,11 @@ use shared::http::{HttpRequest, HttpResponse};
 use shared::metrics::get_metrics;
 use shared::std_canister_status;
 use shared::types::custom_token::{CustomToken, CustomTokenId};
-use shared::types::user_profile::{AddCredentialRequest, GetUsersRequest, GetUsersResponse, OisyUser, UserCredential, UserProfile};
 use shared::types::token::{UserToken, UserTokenId};
 use shared::types::transaction::SignRequest;
+use shared::types::user_profile::{
+    AddCredentialRequest, GetUsersRequest, GetUsersResponse, OisyUser, UserCredential, UserProfile,
+};
 use shared::types::{Arg, InitArg};
 use std::borrow::Cow;
 use std::cell::RefCell;
@@ -498,7 +500,9 @@ fn get_users(request: GetUsersRequest) -> GetUsersResponse {
     let data: Vec<OisyUser> = Vec::new();
     GetUsersResponse {
         data,
-        limit_response: request.limit_response.unwrap_or(DEFAULT_LIMIT_GET_USERS_RESPONSE),
+        limit_response: request
+            .limit_response
+            .unwrap_or(DEFAULT_LIMIT_GET_USERS_RESPONSE),
         total_users: 0,
     }
 }
--- a/src/backend/src/lib.rs
+++ b/src/backend/src/lib.rs
@@ -499,15 +499,13 @@ fn get_or_create_user_profile() -> UserProfile {
 #[allow(clippy::needless_pass_by_value)]
 fn get_users(request: GetUsersRequest) -> GetUsersResponse {
     // TODO: Implement https://dfinity.atlassian.net/browse/GIX-2650
-    // WARNING: Do not read all the BTreeMap directly, but put a limit on how much one reads
-    // due to concerns on running out of cycles.
-    let data: Vec<OisyUser> = Vec::new();
+    // WARNING: The value `DEFAULT_LIMIT_GET_USERS_RESPONSE` must also be determined by the cycles consumption when reading BTreeMap.
+    let users: Vec<OisyUser> = Vec::new();
     GetUsersResponse {
-        data,
-        limit_response: request
-            .limit_response
+        users,
+        matches_max_length: request
+            .matches_max_length
             .unwrap_or(DEFAULT_LIMIT_GET_USERS_RESPONSE),
-        total_users: 0,
     }
 }
 
--- a/src/backend/src/lib.rs
+++ b/src/backend/src/lib.rs
@@ -27,7 +27,8 @@ use shared::types::custom_token::{CustomToken, CustomTokenId};
 use shared::types::token::{UserToken, UserTokenId};
 use shared::types::transaction::SignRequest;
 use shared::types::user_profile::{
-    AddUserCredentialRequest, GetUsersRequest, GetUsersResponse, OisyUser, UserCredential, UserProfile,
+    AddUserCredentialRequest, GetUsersRequest, GetUsersResponse, OisyUser, UserCredential,
+    UserProfile,
 };
 use shared::types::{Arg, InitArg};
 use std::borrow::Cow;
--- a/src/backend/src/lib.rs
+++ b/src/backend/src/lib.rs
@@ -27,7 +27,7 @@ use shared::types::custom_token::{CustomToken, CustomTokenId};
 use shared::types::token::{UserToken, UserTokenId};
 use shared::types::transaction::SignRequest;
 use shared::types::user_profile::{
-    AddCredentialRequest, GetUsersRequest, GetUsersResponse, OisyUser, UserCredential, UserProfile,
+    AddUserCredentialRequest, GetUsersRequest, GetUsersResponse, OisyUser, UserCredential, UserProfile,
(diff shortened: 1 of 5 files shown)


src/backend/src/lib.rs
```diff
@@ -470,6 +475,38 @@ fn list_custom_tokens() -> Vec<CustomToken> {
     read_state(|s| s.custom_token.get(&stored_principal).unwrap_or_default().0)
 }
 
+#[update(guard = "caller_is_not_anonymous")]
```
bitdivine, line 478: Is the plan to panic on error rather than return a Result?

bitdivine (approved):
LGTM, thanks!

src/shared/src/types.rs
```diff
@@ -100,3 +100,57 @@ pub mod custom_token {
...
+    }
+
+    #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
+    pub struct GetUsersRequest {
+        pub updated_after_timestamp: Option<u64>,
+        pub limit_response: Option<u64>,
+    }
+
+    #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
+    pub struct OisyUser {
```
peterpeterparker, line 145: Why not just `User` out of curiosity?

src/shared/src/types.rs
```diff
@@ -100,3 +100,57 @@ pub mod custom_token {
...
+        pub principal: Principal,
+        pub pouh_verified: bool,
```
peterpeterparker, line 147: It's maybe also interesting to know when it was verified?

src/shared/src/types.rs
```diff
@@ -100,3 +100,57 @@ pub mod custom_token {
...
+    }
+
+    #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
+    pub struct GetUsersResponse {
```
peterpeterparker, line 151: Sharing in case interesint, that's the result interface of Juno's satellite:

```
#[derive(Default, CandidType, Deserialize, Clone)]
    pub struct ListResults<T> {
        pub items: Vec<(Key, T)>,
        pub items_length: usize,
        pub items_page: Option<usize>,
        pub matches_length: usize,
        pub matches_pages: Option<usize>,
    }
```

Docs:

```
{
  items: []; // The data - array of documents
  items_length: bigint; // The number of documents - basically items.length
  items_page?: bigint; // If the query is paginated, at what page (starting from 0) do the items find the place
  matches_length: bigint; // The total number of matching results
  matches_pages?: bigint; // If the query is paginated, the total number (starting from 0) of pages
}
```

If we aim for pagination thought it might be interesting to share.

src/shared/src/types.rs
```diff
@@ -100,3 +100,57 @@ pub mod custom_token {
...
+    #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
+    pub struct UserCredential {
+        pub credential_type: CredentialType,
+        pub verified_date_timestamp: Option<u64>,
+        pub expire_date_timestamp: Option<u64>,
+    }
+
+    // Used in the endpoint
+    #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
+    pub struct UserProfile {
```
peterpeterparker, line 120: The user profile should, in my opinion, similarly to `Token` and `CustomToken` contains a `version: Option<Version>` and implement and the `TokenVersion` should be implemented for it.

Not sure if we also want to introduce a version field within the `UserCredential`

src/shared/src/types.rs
```diff
@@ -100,3 +100,57 @@ pub mod custom_token {
...
+        pub credentials: Vec<UserCredential>,
+        pub created_timestamp: u64,
+        pub updated_timestamp: u64,
+    }
+
+    #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
+    pub struct UserProfileStorage {
```
peterpeterparker, line 127: What's that struct for?

peterpeterparker (commented):
PR is a draft but, given that Max already approved it, I add a look too.

src/backend/src/lib.rs
lmuntaner: I have followed the same pattern as the other endpoints.

I would prefer to return an Error, shall I do that even when the pattern is not common in the other endpoints of the canister?

src/backend/src/lib.rs
```diff
@@ -470,6 +475,38 @@ fn list_custom_tokens() -> Vec<CustomToken> {
...
+    UserProfile {
+        credentials,
+        created_timestamp: time(),
+        updated_timestamp: time(),
+    }
+}
+
+#[query]
+#[allow(clippy::needless_pass_by_value)]
+fn get_users(request: GetUsersRequest) -> GetUsersResponse {
```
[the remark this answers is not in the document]
lmuntaner: Yes, that's correct. When I implement it I will add ths limitation.

src/shared/src/types.rs
lmuntaner: This is the type stored in stable memory. `UserProfile` is the one returned to the client.

I had to use two because I couldn't use a Map of credential_type to CredentialSummary in Candid and instead return a vector of CredentialSummary.

src/shared/src/types.rs
lmuntaner: It was even called `PrivateUser`. This is more external and I believe `User` seems like an internal type instead.

src/backend/src/lib.rs
peterpeterparker: In the frontend, you still need to use `try...catch` even if a `Result` is used. Therefore, unless there are specific details that must be included in the result, trapping errors, similarly to other end points of the backend, might fits better?

src/shared/src/types.rs
lmuntaner: That's very interesting. We might end up adding those. This will be used by the Rewards Canister, so we'll try to keep it simple.

src/shared/src/types.rs
lmuntaner: Oh, that adds quite come complexity. How does that work? What's the advantage of having this?

src/backend/src/lib.rs
```diff
@@ -470,6 +475,38 @@ fn list_custom_tokens() -> Vec<CustomToken> {
     read_state(|s| s.custom_token.get(&stored_principal).unwrap_or_default().0)
 }
 
+#[update(guard = "caller_is_not_anonymous")]
+#[allow(clippy::needless_pass_by_value)]
+#[allow(unused_variables)]
+fn add_credential(request: AddCredentialRequest) {
```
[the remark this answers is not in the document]
lmuntaner: Done!

src/backend/src/lib.rs
lmuntaner: I prefer to follow the current pattern unless otherwise specified.

There are different reasons why the validation might fail, but I don't think that any of them is something that the user could do something about. The user follows a flow and has no control over how the credential is built.

src/backend/src/lib.rs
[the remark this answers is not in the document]
peterpeterparker: I would add the guard already. It makes clear from the start that it was never meant to be public and also less prone to being forgotten.

src/shared/src/types.rs
peterpeterparker: I see. I just feld "OisyUser" not super meaningful but, does not have a better naming to suggest.

src/shared/src/types.rs
peterpeterparker: I'm not sure I got it but, all good sure.

src/shared/src/types.rs
peterpeterparker: Sure. Reading the struct again:

- Shouldn't `data` be renamed to `users`?
- What's `limit_response` ? Feels like a call to action to prepare the response but, here it's already the response

src/shared/src/types.rs
peterpeterparker: > What's the advantage of having this?

Prevent data overwrite if the same entity is edited multiple times at the same times.

Think at a user who use the dApp at the same time on their mobile phone and laptop for example.

> How does that work?

Basically the caller must provide the version that matches the entity in the smart contract otherwise the update is rejected.

Kind of similar to what's done in SQL database.

src/backend/src/lib.rs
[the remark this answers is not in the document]
lmuntaner: Added

src/shared/src/types.rs
lmuntaner: It's to tell the client what's the maximum size it will bet. Most of the times, it will be the same as in the request, except if the backend limit is 1,000 and the client asks for 50,000. In that case, the `limit_response` will be different in the request and response. Request is `50,000` and response is `1,000`.

Maybe the name is confusing. Therefore, I renamed to `matches_max_length`.

src/shared/src/types.rs
lmuntaner: Thanks! I added the `version` field.

lmuntaner:

src/shared/src/types.rs
lmuntaner: Naming is hard...

src/shared/src/types.rs
peterpeterparker: Thanks

src/shared/src/types.rs
peterpeterparker: Ah got it! For consistency, should we call it `StoredUserProfile`? It's a bit different but, we already use  `StoredPrincipal` to store principal as key of the maps.

```
type CustomTokenMap = StableBTreeMap<StoredPrincipal, Candid<Vec<CustomToken>>, VMem>;

MEMORY_MANAGER.with(|mm| State {
...
            custom_token: CustomTokenMap::init(mm.borrow().get(USER_CUSTOM_TOKEN_MEMORY_ID)),
        })
```

peterpeterparker (approved):
LGTM, thx

src/shared/src/types.rs
lmuntaner: Ah, sure, that makes sense.

I'll rename it in another PR.

PR #1745, opened 2024-07-12
feat(frontend): token store is transformed into context

AntonioVentilii-DFINITY:
# Motivation

Referring to PR #1462 

# Changes

# Tests

### DIFF
--- a/src/frontend/src/icp/components/send/IcSend.svelte
+++ b/src/frontend/src/icp/components/send/IcSend.svelte
@@ -22,7 +22,7 @@
 <SendButton on:click={openSend} {compact} />
 
 {#if $modalIcSend && $modalStore?.data === modalId}
-	<TokenContext>
+	<TokenContext {token}>
 		<IcSendModal networkId={ICP_NETWORK_ID} on:nnsClose />
 	</TokenContext>
 {/if}
--- a/src/frontend/src/icp/components/send/IcSend.svelte
+++ b/src/frontend/src/icp/components/send/IcSend.svelte
@@ -6,6 +6,7 @@
 	import SendButton from '$lib/components/send/SendButton.svelte';
 	import type { Token } from '$lib/types/token';
 	import { loadTokenAndRun } from '$icp/services/token.services';
+	import TokenContext from '$lib/components/tokens/TokenContext.svelte';
 
 	export let token: Token;
 	export let compact = false;
@@ -21,5 +22,7 @@
 <SendButton on:click={openSend} {compact} />
 
 {#if $modalIcSend && $modalStore?.data === modalId}
-	<IcSendModal networkId={ICP_NETWORK_ID} on:nnsClose />
+	<TokenContext>
+		<IcSendModal networkId={ICP_NETWORK_ID} on:nnsClose />
+	</TokenContext>
 {/if}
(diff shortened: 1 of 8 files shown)


src/frontend/src/lib/stores/token.store.ts
```diff
@@ -1,33 +1,53 @@
...
- * Unfortunately, this architecture is not ideal. If the global token store is used for other features on the main screen in the future, conflicts may arise.
- * Therefore, this store is deprecated and should eventually be replaced with a method that passes down the selected token, ideally through contexts.
- * In other words, for the Hero, Receive, and Send features, we should avoid accessing a global store and instead use a dedicated state.
- * TODO: pass down token with context instead of global store.
- * @deprecated This approach works for now but does not align with the new architectural requirements.
- */
-export const token = initTokenStore();
+export const TOKEN_CONTEXT_KEY = Symbol('token');
+
+export const getTokenStoreFromContext = (): TokenStore =>
```
peterpeterparker, line 52: I would not proceed this with a big bang migration but, I've got the feeling that it would be easier to introduce the context piece by piece. For example there are already 40 changes in this PR, that's quite a lot. Introducing a new context might help providing smaller PR.

peterpeterparker (commented):
Few feedbacks not really linked to the code but rather the approach. Let me know what you think.

AntonioVentilii-DFINITY:
I think I understood the idea: check my last commit. That would be the first step, and then Pr-by-PR, we adapt the rest?

src/frontend/src/icp/components/send/IcSend.svelte
```diff
@@ -21,5 +22,7 @@
 <SendButton on:click={openSend} {compact} />
 
 {#if $modalIcSend && $modalStore?.data === modalId}
-	<IcSendModal networkId={ICP_NETWORK_ID} on:nnsClose />
+	<TokenContext>
```
peterpeterparker, line 25: That's a coo idea!

src/frontend/src/lib/components/tokens/TokenContext.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+	import { setContext } from 'svelte';
+	import { initTokenStore } from '$lib/stores/token.store';
+
+	const tokenContextKey = {};
```
peterpeterparker, line 5: Shouldn't this component takes a `token` as property and use it to create the context?

That way the context is never undefined or empty.

src/frontend/src/lib/stores/token.store.ts
```diff
@@ -31,3 +32,8 @@ const initTokenStore = (): TokenStore => {
  * @deprecated This approach works for now but does not align with the new architectural requirements.
  */
 export const token = initTokenStore();
+
+export const TOKEN_CONTEXT_KEY = Symbol('token');
+
+export const getTokenStoreFromContext = (): TokenStore =>
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Done! I went with the second option, because I think, until we move to the top-level, we will use more the function `getContextToken`

src/frontend/src/lib/components/tokens/TokenContext.svelte
AntonioVentilii-DFINITY: You are right! Fixed

AntonioVentilii-DFINITY:
The approach will be different: we will create context for each situation. For example, we will use the Send context when the Send modal is opened. So for now, we close this one

PR #1746, opened 2024-07-12
feat(frontend): base sorting logic for token list

AntonioVentilii-DFINITY:
# Motivation

According to a recent requirement, the sorting logic of the token list, by default, should be:

1. Market cap
2. Name
3. Network Name

# Changes

- Create sorting function using both token list and exchange data.
- Apply the logic **BEFORE** pinning tokens on top (as per requirements).

# Tests

Created tests for the new function.


peterpeterparker (commented):
Are we fine merging this into the main branch, knowing that it implies that UI/UX-wise this can lead to some weird behavior for the users, with tokens moving around if exchange values are changing or are not yet loaded?

AntonioVentilii-DFINITY:
In the code, the exchange rates (and consequently the market cap) are refreshed every 1 minute.

I talked with Mary last week and she said that it is ok to refresh the order of the tokens at the same time as we refresh the rates.

I wanted to do a separate PR to make it so, during refresh, it would disable the clicking on the tokens list, or something similar (open to suggestions), to avoid weird behavior. Shall I do it in this one?

However, there could be more path to propose:

1) extend the refreshing interval
or
2) refreshing only when it changes pages and/or after X minutes of inactivity passes

peterpeterparker (approved):
LGTM; thanks

peterpeterparker:
Thanks for the details.

Refreshing the rate is a core functional behavior, sorting tokens for display purpose is a UI thing.

Therefore I think the refresh rate should NOT be changed, nor should it be disabled. The potential UI glitch should be solve at the UI level.

peterpeterparker:
> In the code, the exchange rates (and consequently the market cap) are refreshed every 1 minute.

Note that there can be a glitch at minute zero.

i.e. Oisy renders the tokens but, exchange is still fetching the rates. Rates pops in, sorting is modified.
In other words, there can be a glitch at boot time.

PR #1751, opened 2024-07-15
style(frontend): "Connect with" padding

peterpeterparker:
# Motivation

I feel like the "Connect with" button is a bit compressed. Adding just a bit of padding make it a bit more pleasant.

# Screenshots

Capture d’écran 2024-07-15 à 08 10 01
Capture d’écran 2024-07-15 à 08 10 22


AntonioVentilii-DFINITY (approved):
Very true!

PR #1752, opened 2024-07-15
style(frontend): external link full width

peterpeterparker:
# Motivation

"View on explorer" does not match the full with in the context menu which is a bit inconsistent given that other actions (buttons) of the same menu do.

# Screenshots

Capture d’écran 2024-07-15 à 08 16 56


AntonioVentilii-DFINITY (approved):
Just a small comment: shouldn't it be `fullWidth`?

peterpeterparker:
> Just a small comment: shouldn't it be fullWidth?

Definitely, thanks!

PR #1759, opened 2024-07-15
refactor(frontend): extract a const for 1inch address

peterpeterparker:
# Motivation

There are other constants now so we can also extract a constant for 1inch.


AntonioVentilii-DFINITY (approved):
thank you! it was on my list of to-do eheh

PR #1761, opened 2024-07-15
feat(frontend): use addresses from JSON env file for  ERC20 twin token

peterpeterparker:
# Motivation

We have redundancy in our environment configuration. ERC20 contract addresses for twin tokens, provided by the orchestrator and extracted to JSON files, duplicate the values we are coding. This PR aims to address this duplication.

# Changes

- Create an ERC20 type for the environment with the field "address" omitted.
- Create a mapper that retrieves the addresses from the JSON environment file by mapping the symbols.

# Notes

Symbols are not unique, but given that we are manually mapping the tokens—i.e., manually listing which tokens should be mapped—we can assume we know that we are mapping a unique symbol.

For example: USDC === (ckUSDC)


AntonioVentilii:
Stale (>1 year, no activity) and the surrounding env/ERC20 code has changed substantially. Closing — please reopen against current main if still desired.

PR #1762, opened 2024-07-15
refactor(frontend): extract utils for mapCkErc20Data

peterpeterparker:
# Motivation

In #1761 we added some more utils for `$env` therefore we can also extract the existing mapper to a utility as well for consistency.


AntonioVentilii:
Stale (>1 year, no activity) and depends on #1761, which is also being closed. Closing — please reopen against current main if still desired.

PR #1768, opened 2024-07-15
chore(frontend): change string to authenticate with II

AntonioVentilii-DFINITY:
# Motivation

As per recent request, we were asked to change the message to use Internet Identity Authetication from "Connect with" to "Authenticate with"

# Before

Screenshot 2024-07-15 at 16 38 03

# After

Screenshot 2024-07-15 at 16 36 31


peterpeterparker (approved):
Code looks good to me.

That big "Authenticate" button feels really weird UX wise but, not my call.

PR #1773, opened 2024-07-16
feat(devops): Add Init Args

lmuntaner:
# Motivation

To validate credentials Oisy backend will need some information about the environment. Such as the canister ids of II and the POUH issuer.

In this PR, I add new fields to InitArg and pass them on deploy.

# Changes

* Add fields to InitArg.
* Gather the environment information in deploy.backend and pass it as init arguments.
* Update candid file with new fields and type.
* Update the config on post_upgrade and not only on init.

# Tests

I tested by deploying locally the backend to an already deployed canister and it worked.


scripts/deploy.backend.sh
```diff
@@ -14,18 +14,67 @@ case $ENV in
     ;;
 esac
 
+II_CANISTER_ID="$(dfx canister id internet_identity --network "$ENV")"
```
peterpeterparker, line 17: This is not incorrect for a deployment on mainnet (production and staging)?

scripts/deploy.backend.sh
```diff
@@ -14,18 +14,67 @@ case $ENV in
...
+POUH_ISSUER_CANISTER_ID="$(dfx canister id pouh_issuer --network "$ENV")"
```
peterpeterparker, line 18: ditto

scripts/deploy.backend.sh
```diff
@@ -14,18 +14,67 @@ case $ENV in
...
+# URL used by II-issuer in the id_alias-verifiable credentials (hard-coded in II)
+# Represents more an ID than a URL
+II_VC_URL="https://identity.ic0.app"
```
peterpeterparker, line 21: The contrary, that's invalid for local and ci development no?

scripts/deploy.backend.sh
```diff
@@ -14,18 +14,67 @@ case $ENV in
...
+# URL used by issuer in the issued verifiable credentials (tipically hard-coded)
+# Represents more an ID than a URL
+# TODO: To be confirmed by DECIDE AI team
+POUH_ISSUER_VC_URL="https://www.elna.ai"
+
+
+if [[ "$ENV" = "local" || -z "${ENV}" ]]; then
```
peterpeterparker, line 28: For readability reason, I would move the initialization of any variable depending of the environment in the existing `case` block?

src/backend/src/lib.rs
```diff
@@ -154,36 +155,63 @@ pub struct Config {
     pub ecdsa_key_name: String,
     // A list of allowed callers to restrict access to endpoints that do not particularly check or use the caller()
     pub allowed_callers: Vec<Principal>,
+    pub supported_credentials: Option<Vec<SupportedCredential>>,
+    /// Root of trust for checking canister signatures.
+    pub ic_root_key_raw: Option<Vec<u8>>,
```
peterpeterparker, line 160: Why do we need this root key in the state? Can't it be derived on the fly when needed? If it has to be saved in the state, what are the security implication?

src/backend/src/lib.rs
```diff
@@ -154,36 +155,63 @@ pub struct Config {
...
+}
+
+fn update_config(arg: InitArg) {
+    mutate_state(|state| {
+        let ic_root_key_raw = match arg
+            .ic_root_key_der
+            .map(|der_key| extract_raw_root_pk_from_der(&der_key))
+        {
+            Some(Ok(root_key)) => Some(root_key),
+            Some(Err(_)) => None,
```
peterpeterparker, line 170: It's an acceptable error - i.e. if it fails it's ok to let the init and upgrade of the canister succeed? How do we notice if this fails?

src/backend/src/lib.rs
```diff
@@ -154,36 +155,63 @@ pub struct Config {
...
+    pub ic_root_key_raw: Option<Vec<u8>>,
+}
+
+fn update_config(arg: InitArg) {
+    mutate_state(|state| {
+        let ic_root_key_raw = match arg
+            .ic_root_key_der
+            .map(|der_key| extract_raw_root_pk_from_der(&der_key))
+        {
+            Some(Ok(root_key)) => Some(root_key),
```
peterpeterparker, line 169: There is no enforcement in place to ensure that no root_key args is passed when the canister is deployed on mainnet, is that safe enough?

src/backend/src/lib.rs
```diff
@@ -154,36 +155,63 @@ pub struct Config {
...
+            Some(Err(_)) => None,
+            None => match extract_raw_root_pk_from_der(IC_ROOT_PK_DER) {
+                Ok(root_key) => Some(root_key),
+                Err(_) => None,
+            },
+        };
+        state
+            .config
+            .set(Some(Candid(Config {
+                ecdsa_key_name: arg.ecdsa_key_name,
```
peterpeterparker, line 179: Not sure but you might be able to destruct the arg here to be less version.

```
Config {
   ic_root_key_raw,
   ..arg
}
```

But agin, not sure.

src/backend/src/lib.rs
```diff
@@ -154,36 +155,63 @@ pub struct Config {
     pub ecdsa_key_name: String,
     // A list of allowed callers to restrict access to endpoints that do not particularly check or use the caller()
     pub allowed_callers: Vec<Principal>,
+    pub supported_credentials: Option<Vec<SupportedCredential>>,
+    /// Root of trust for checking canister signatures.
+    pub ic_root_key_raw: Option<Vec<u8>>,
+}
+
+fn update_config(arg: InitArg) {
```
peterpeterparker, line 163: Given that the function does not replace a single fiel but entirely overwrite the config, maybe `init_config` instead of `update_config`?

src/backend/src/lib.rs
```diff
@@ -154,36 +155,63 @@ pub struct Config {
...
+    match arg {
+        Some(Arg::Init(arg)) => update_config(arg),
+        Some(Arg::Upgrade) => {
+            read_state(|s| {
+                let _ = s.config.get().as_ref().expect(
+                    "config is not initialized: reinstall the canister instead of upgrading",
+                );
+            });
+        }
+        None => {
```
peterpeterparker, line 207: `None` is acceptable? If yes you can maybe spare some code duplication by replacing `Some(Arg::Upgrade)` and `None` with `_`.

If `None` is not acceptable, maybe trap?

src/shared/src/types.rs
```diff
@@ -1,10 +1,22 @@
 use candid::{CandidType, Deserialize, Principal};
 use std::fmt::Debug;
 
+#[derive(CandidType, Deserialize)]
+pub struct SupportedCredential {
+    pub credential_type: String,
+    pub ii_origin: String,
```
peterpeterparker, line 7: What happens if we support both `ic0.app` and `internetcomputer.org`? Not an issue with a single `ii_origin`?

scripts/deploy.backend.sh
lmuntaner: Why would this be incorrect for production and staging?

scripts/deploy.backend.sh
lmuntaner: Also not, the proper canister id will be set in dfx.json in this case.

scripts/deploy.backend.sh
lmuntaner: No, that's not correct. Maybe I'll rename the variable, but in the library it's using this name.

That's why I added a comment about, but this is more of an ID than a URL. The credential will have this url hardcoded and we need to compare it with this.

scripts/deploy.backend.sh
lmuntaner: Yes, good point.

src/backend/src/lib.rs
lmuntaner: The root key in a local replica changes every time. We already discussed with the crypto team to have a fixed one, but that won't happen in the short term.

Moreover, the root key is not accessible from within a canister.

Therefore, the root key needs to be passed as a parameter on init (or upgrade).

This is how we have been keeping the root key in all the demos we have been doing and the feedback from security was to have a hardcoded mainnet key.

src/backend/src/lib.rs
lmuntaner: Good point, how about `set_config` then. Init seems to be done only once.

src/backend/src/lib.rs
lmuntaner: I believe so, but I wonder whether we should ask for a security review after this PR?

src/backend/src/lib.rs
lmuntaner: Mhm, I don't know. My default is always to keep running code. But maybe in this case we want to panic and skip the upgrade. I'll add a `panic` instead.

src/backend/src/lib.rs
lmuntaner: Sure!

src/backend/src/lib.rs
lmuntaner: I believe it's ok yes. One could upgrade without an argument and we still want to check the config.

I'll use `_` thanks!

src/shared/src/types.rs
lmuntaner: Not an issue, the credentials from both will have the same value.

I believe this is again a naming confusion.

As an example you can try to the VC flow with the dummy [relying party](https://l7rua-raaaa-aaaap-ahh6a-cai.icp0.io/) and dummy issuer (default values).

After receiving the credentials, you can decode them and then decode the alias credential.

The "iss" field is what this `ii_origin` means. It's the issuer.

Screenshot 2024-07-16 at 13 13 48

I used this names (issuer_origin also) because it's the name used i[n the library](https://github.com/dfinity/verifiable-credentials-sdk/blob/a116a2e53a13e4a1f6f7e8429960a7d0d6142e70/rust-packages/ic-verifiable-credentials/src/lib.rs#L45) that Bartosz created to validate the credentials.

I'm ok changing the name here and then mapping it properly. I also found it confusing from the coding perspective, but it made sense from the standard perspective.

scripts/deploy.backend.sh
peterpeterparker: Right, it works for production. Doesn't for staging though

> ❯ dfx canister id internet_identity --network staging
> Error: Cannot find canister id. Please issue 'dfx canister create internet_identity --network staging'.

scripts/deploy.backend.sh
peterpeterparker: Right.

scripts/deploy.backend.sh
peterpeterparker: Oh it's an ID. Yeah would be nice to rename if possible. Not necessary in this PR but, generally speaking.

scripts/deploy.backend.sh
```diff
@@ -4,28 +4,74 @@ case $ENV in
...
+POUH_ISSUER_VC_URL="https://www.elna.ai"
+
 if [ -n "${ENV+1}" ]; then
   dfx deploy backend --argument "(variant {
     Init = record {
          ecdsa_key_name = \"$ECDSA_KEY_NAME\";
          allowed_callers = vec {};
+         supported_credentials = opt vec {
+            record {
+              credential_type = \"ProofOfUniqueness\";
```
peterpeterparker, line 50: Seeing it now, it's always "ProofOfUniqueness" - i.e. shouldn't the type be a variant / enum instead of String?

src/backend/src/lib.rs
peterpeterparker: Thanks for the exaplantion, did not knew. It feels a bit weird no?
All good.

src/backend/src/lib.rs
```diff
@@ -154,36 +155,60 @@ pub struct Config {
...
+
+fn set_config(arg: InitArg) {
+    let InitArg {
+        ecdsa_key_name,
+        allowed_callers,
+        supported_credentials,
+        ic_root_key_der,
+    } = arg;
+    mutate_state(|state| {
+        let ic_root_key_raw =
```
peterpeterparker, line 171: With the help of ChatGPT, if interesting, the two lines version of this block:

```
let ic_root_key_raw = extract_raw_root_pk_from_der(
    &ic_root_key_der.unwrap_or_else(|| IC_ROOT_PK_DER.to_vec())
).map(Some).unwrap_or_else(|_| panic!("There was an error parsing the root key"));
```

src/backend/tests/it/utils/pocketic.rs
```diff
@@ -91,6 +91,8 @@ fn init_arg() -> Arg {
     Arg::Init(InitArg {
         ecdsa_key_name: format!("master_ecdsa_public_key_{}", SUBNET_ID).to_string(),
         allowed_callers: vec![Principal::from_text(CALLER).unwrap()],
+        ic_root_key_der: None,
+        supported_credentials: None,
```
peterpeterparker, line 95: You provide a test for the upgrade in another PR?

src/shared/src/types.rs
peterpeterparker: I'm not sure I still get it but, thanks for the explanation. I think you are right we should use the naming convetion even if they do feel a bit weird at first sight.

That said, I'm still unsure what happens when a credential is issued on iss ic0.app but used on internetcomputer.org but, definitely nothin to do with this PR.

scripts/deploy.backend.sh
lmuntaner: Can I create an enum with a set of strings in Candid or Rust?

I would need the mapper of an enum / variant to a string at some point anyway.

That would mean that we'll need to add the types to the enum when we want to support new credentials. Which kind of defeats the purpose to have it as a parameter as well, doesn't it?

src/backend/src/lib.rs
lmuntaner: 🎉

src/backend/tests/it/utils/pocketic.rs
lmuntaner: What do you mean a test for the upgrade?

Are there already upgrade tests?

src/shared/src/types.rs
lmuntaner: internetcomputer.org will issue the same credential on ic0.app because both have the identity.ic0.app hardcoded.

scripts/deploy.backend.sh
lmuntaner: As discussed in [this comment](https://github.com/dfinity/oisy-wallet/pull/1773#discussion_r1679239909) we chose to keep the name according to the convention.

src/backend/src/lib.rs
peterpeterparker: Maybe. Generally speaking we have been keeping them in the loop of Oisy each time we have build a new feature.

scripts/deploy.backend.sh
peterpeterparker: I think so. In Rust you create an enum which translate to a variant in Candid.

For example this enum: https://github.com/junobuild/juno/blob/e41584c130736ddea8c8e2bd0a444119136599d4/src/libs/shared/src/types.rs#L32C5-L36C6

Becomes this variant in Candid: https://github.com/junobuild/juno/blob/e41584c130736ddea8c8e2bd0a444119136599d4/src/declarations/observatory/observatory.did.d.ts#L10

So I guess here we can have 

```
 #[derive(CandidType, Serialize, Deserialize, Clone)]
pub enum CredentialType {
       ProofOfUniqueness,
}
```

But, I have no clue of the credentials mechanism, just a suggestion if the type are defined in some ways.

src/backend/tests/it/utils/pocketic.rs
peterpeterparker: Sure, for tokens when we added the flag `version` and `enabled`. There are also utils to dowload a particular wasm of the backend and then proceed with an upgrade.

See https://github.com/dfinity/oisy-wallet/tree/main/src/backend/tests/it/upgrade

scripts/deploy.backend.sh
lmuntaner: The credential type in Rust is a String:

https://github.com/dfinity/verifiable-credentials-sdk/blob/a116a2e53a13e4a1f6f7e8429960a7d0d6142e70/rust-packages/ic-verifiable-credentials/src/issuer_api.rs#L94

src/backend/src/lib.rs
lmuntaner: I'll do that then!

lmuntaner: I'll also point to this PR as well.

peterpeterparker (approved):
LGTM, thanks

Might be worth to have a quick review from @bitdivine too particularly as we are touching init and post upgrade here.

scripts/deploy.backend.sh
```diff
@@ -4,28 +4,74 @@ case $ENV in
...
   "ic")
     ECDSA_KEY_NAME="key_1"
     WALLET="yit3i-lyaaa-aaaan-qeavq-cai"
+    # For security reasons, mainnet root key will be hardcoded in the backend canister.
+    ic_root_key_der="null"
     ;;
   *)
     ECDSA_KEY_NAME="dfx_test_key"
+    # At the time of writing dfx outputs incorrect JSON with dfx ping (commas between object
+    # entries are missing).
```
bitdivine, line 19: It _seems_ to work with dfx 0.20.0, as defined in dfx.json, for mainnet and localhost:
```
$ dfx ping ic | jq -r '.root_key | reduce .[] as $item ("opt vec { "; "\(.) \($item):nat8;") + " }"'
opt vec {  48:nat8; 129:nat8; 130:nat8; 48:nat8; 29:nat8; 6:nat8; 13:nat8; 43:nat8; 6:nat8; 1:nat8; 4:nat8; 1:nat8; 130:nat8; 220:nat8; 124:nat8; 5:nat8; 3:nat8; 1:nat8; 2:nat8; 1:nat8; 6:nat8; 12:nat8; 43:nat8; 6:nat8; 1:nat8; 4:nat8; 1:nat8; 130:nat8; 220:nat8; 124:nat8; 5:nat8; 3:nat8; 2:nat8; 1:nat8; 3:nat8; 97:nat8; 0:nat8; 129:nat8; 76:nat8; 14:nat8; 110:nat8; 199:nat8; 31:nat8; 171:nat8; 88:nat8; 59:nat8; 8:nat8; 189:nat8; 129:nat8; 55:nat8; 60:nat8; 37:nat8; 92:nat8; 60:nat8; 55:nat8; 27:nat8; 46:nat8; 132:nat8; 134:nat8; 60:nat8; 152:nat8; 164:nat8; 241:nat8; 224:nat8; 139:nat8; 116:nat8; 35:nat8; 93:nat8; 20:nat8; 251:nat8; 93:nat8; 156:nat8; 12:nat8; 213:nat8; 70:nat8; 217:nat8; 104:nat8; 95:nat8; 145:nat8; 58:nat8; 12:nat8; 11:nat8; 44:nat8; 197:nat8; 52:nat8; 21:nat8; 131:nat8; 191:nat8; 75:nat8; 67:nat8; 146:nat8; 228:nat8; 103:nat8; 219:nat8; 150:nat8; 214:nat8; 91:nat8; 155:nat8; 180:nat8; 203:nat8; 113:nat8; 113:nat8; 18:nat8; 248:nat8; 71:nat8; 46:nat8; 13:nat8; 90:nat8; 77:nat8; 20:nat8; 80:nat8; 95:nat8; 253:nat8; 116:nat8; 132:nat8; 176:nat8; 18:nat8; 145:nat8; 9:nat8; 28:nat8; 95:nat8; 135:nat8; 185:nat8; 136:nat8; 131:nat8; 70:nat8; 63:nat8; 152:nat8; 9:nat8; 26:nat8; 11:nat8; 170:nat8; 174:nat8; }


$ dfx --version
dfx 0.20.0
```
But if you are getting errors in some cases, for sure, dfx needs to get fixed! :-)

src/backend/src/lib.rs
```diff
@@ -154,36 +155,57 @@ pub struct Config {
...
+        allowed_callers,
+        supported_credentials,
+        ic_root_key_der,
+    } = arg;
+    mutate_state(|state| {
+        let ic_root_key_raw = match extract_raw_root_pk_from_der(
+            &ic_root_key_der.unwrap_or_else(|| IC_ROOT_PK_DER.to_vec()),
+        ) {
+            Ok(root_key) => root_key,
+            Err(msg) => panic!("{}", format!("Error parsing root key: {msg}")),
```
bitdivine, line 175: So the idea is that if an incorrect argument is passed, the canister dies in the post-upgrade hook rather than proceeding?

src/backend/src/lib.rs
```diff
@@ -154,36 +155,57 @@ pub struct Config {
...
+        };
+        state
+            .config
+            .set(Some(Candid(Config {
+                ecdsa_key_name,
+                allowed_callers,
+                supported_credentials,
+                ic_root_key_raw: Some(ic_root_key_raw),
+            })))
+            .expect("setting config should succeed");
```
bitdivine, line 185: If I read correctly, `.set()` can fail only if the data is too big to fit into memory.  The memory is expanded if necessary, so it tentatively looks to me as if the expect could be `.unwrap_or_else(|_| unreachable!("This data structure is too small to exhaust stable memory."))`?  I guess the rest of memory could be filled with accounts, and this is taking just the last few bytes so then I guess it could happen.  OK!  Please shout if I missed something but at the moment `expect` looks correct to me.

bitdivine (approved):
LGTM, thanks!

src/declarations/backend/backend.did
```diff
@@ -1,3 +1,4 @@
+type AddUserCredentialRequest = record { credential_jwt : text };
 type Arg = variant { Upgrade; Init : InitArg };
```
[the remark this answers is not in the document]
lmuntaner: This was tested manually and I intend to add an upgrade test in a following PR.

scripts/deploy.backend.sh
lmuntaner: Thanks! I changed it. I copied it from another project.

src/backend/src/lib.rs
lmuntaner: That is correct.

src/backend/src/lib.rs
lmuntaner: Better be on the safe side.

PR #1775, opened 2024-07-16
feat(frontend): tokens list in manage modal will be sorted by market cap and some pinned

AntonioVentilii-DFINITY:
# Motivation

For the `ManageTokens` list, we sort by marketcap first and then we pin some tokens on top. The logic is the same as per the list of tokens shown in the main page.
Since we want to persist the order of the list during the usage of the modal, we use static data for the exchanges data.

# Changes

- Create static variable for exchanges data, set with `onMount`.
- Create list of sorted tokens.

# Tests

Manual test on local replica showed correct sorting.

### DIFF
--- a/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
@@ -32,6 +32,10 @@
 	import type { LedgerCanisterIdText } from '$icp/types/canister';
 	import { filterTokensForSelectedNetwork } from '$lib/utils/network.utils';
 	import type { IcCkToken } from '$icp/types/ic';
+	import { pinTokensAtTop, sortTokens } from '$lib/utils/tokens.utils';
+	import { exchanges } from '$lib/derived/exchange.derived';
+	import { tokensToPin } from '$lib/derived/tokens.derived';
+	import type { ExchangesData } from '$lib/types/exchange';
 
 	const dispatch = createEventDispatcher();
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
```diff
@@ -84,6 +88,23 @@
 		$pseudoNetworkChainFusion
 	]);
 
+	// To avoid strange behavior when the exchange data changes (for example, the tokens may shift
+	// since some of them are sorted by market cap), we store the exchange data in a variable during
+	// the life of the component.
+	let exchangesStaticData: ExchangesData | undefined;
+	onMount(() => {
```
peterpeterparker, line 95: - While Svelte seems to be fine having multiple `onMount` declared, I would not be against having a single declaration / usage. Would you mind using a single `onMount`?

- Should we create a deep copy to even strenghness the fact that we do not want to dynamically update the list? `exchangesStaticData = nonNullish($exchanges) ? { ...$exchanges } : undefined;`

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
```diff
@@ -84,6 +88,23 @@
...
+		exchangesStaticData = $exchanges;
+	});
+
+	let allTokensSorted: Token[] = [];
+	$: exchangesStaticData,
```
peterpeterparker, line 100: I don't think you need `$: exchangesStaticData,` here.

```
$: allTokensSorted = etc.
```

Should do no given that `exchangesStaticData` is use as param of the assignation?

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
AntonioVentilii-DFINITY: - on it! do you prefer in this PR or another?
- makes sense

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
AntonioVentilii-DFINITY: since it is initialized as `undefined`, i was worried that it would not consider the change when it moves from `undefined` to `nonNullish`  on `onMount` event

but you were right, it works either way

removed it

AntonioVentilii-DFINITY:
About the complexity of the component, I agree. Maybe I will start to move out something...

for example, we could create some `derived` about `tokensForManageModal` (or something similar), moving out the list of tokens

peterpeterparker:
> we could create some derived about tokensForManageModal (or something similar), moving out the list of tokens

Maybe not a global state but, yeah, something like that

peterpeterparker (approved):
LGTM, thx

PR #1776, opened 2024-07-16
feat(backend): Add updated_timestamp to OisyUser

lmuntaner:
# Motivation

Rewards canister will also need the updated timestamp for each user.

# Changes

* Add "updated_timestamp" to OisyUser.
* Add new field in candid file.

# Tests

Not necessary.

### DIFF
--- a/src/backend/backend.did
+++ b/src/backend/backend.did
@@ -48,7 +48,11 @@ type InitArg = record {
   ecdsa_key_name : text;
   allowed_callers : vec principal;
 };
-type OisyUser = record { "principal" : principal; pouh_verified : bool };
+type OisyUser = record {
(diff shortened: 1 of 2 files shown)


lmuntaner:

This was requested by @frederikrothenberger

frederikrothenberger (approved):
Thanks!

peterpeterparker (approved):
LGTM, thanks.

Btw. in Juno I actually create a type for the timestamp and do not mention "timestamp" in the field. I use then pattern "something_at" which feels like a common naming convention in SQL but, don't really have statistics for. Just a feeling and sharing if interesting.

```
type Timestamp = u64;

struct Something {
   updated_at: Timestamp;
}
```

PR #1777, opened 2024-07-16
feat(rust): Move all Rust dependencies into the workspace

bitdivine:
# Motivation
For fast builds, it helps to have all dependencies defined in the workspace.  This way, dependencies can be pulled and built regardless of which packages we have.

# Changes
- Move all Rust dependencies into the workspace.
- Add a test to ensure that all the rust dependencies are in the workspace.

# Tests
See CI.  Builds should complete as usual.


peterpeterparker (approved):
LGTM, thanks.

Just thinking out at loud: another option would be to move "format" and "lint" checks within the Docker build? Don't know if possible, again just a thought.

peterpeterparker:
The Docker image builds the Wasm which is then released. Adding linter and formatter checks within the container might make sense to have the same tooling do everything?

I'm also not familiar with caching on GitHub Actions, but is it maybe easier (and more performant) to cache the resulting images rather than caching the filesystem of the actions?

Really, I'm just stricly thinking out at loud. Not that much an idea.

PR #1779, opened 2024-07-16
feat(frontend): pin tokens with balance at top

AntonioVentilii-DFINITY:
# Motivation

As final requirement for sorting, we pin the tokens with balance on top, per descending USD amount and after per token amount (if USD value is not provided).

# Changes

- Create sorting function to pin tokens with balance on top.
- Apply sorting to the `sortedTokens` derived.

# Tests

Created test for function `pinTokensWithBalanceAtTop`.


src/frontend/src/lib/derived/tokens.derived.ts
```diff
@@ -31,7 +32,11 @@ export const tokensToPin: Readable<TokenToPin[]> = derived(
 );
 
 export const sortedTokens: Readable<Token[]> = derived(
-	[tokens, tokensToPin, exchanges],
-	([$tokens, $tokensToPin, $exchanges]) =>
-		pinTokensAtTop({ $tokens: sortTokens({ $tokens, $exchanges }), $tokensToPin })
+	[tokens, tokensToPin, exchanges, balancesStore],
+	([$tokens, $tokensToPin, $exchanges, $balancesStore]) =>
+		pinTokensWithBalanceAtTop({
```
peterpeterparker, line 37: I'm a bit concerned about the performant. This PR introduces more `map` and `filter` and `sort` over the list of tokens and various stores. I'm sure it's fine when user have few tokens but, I fear it becomes laggy when the amount of data grows. 

For example here there is `pinTokensWithBalanceAtTop` that manipulate the list which then call `pinTokensAtTop` that manipulates the list which then use `sortTokens` that manipulate the list as well.

Is there any potential of performance improvements?

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -40,3 +45,44 @@ export const sortTokens = ({
...
+	$balancesStore
+}: {
+	$tokens: Token[];
+	$exchanges: ExchangesData;
+	$balancesStore: CertifiedStoreData<BalancesData>;
+}): Token[] => {
+	const tokensWithBalanceToPin: TokenToPin[] = $tokens
+		.map((token) => ({
+			...token,
+			usdValue: nonNullish($exchanges?.[token.id])
```
peterpeterparker, line 61: - In which type is `usdValue` defined?
- If the `usdValue` is now used for sorting purpose as well, should we integrate in the global store and global type or should we keep it as an optional value calculated on the fly? For example for the same token we gonna access the exchange store to present the value in dollar to sort the token and then for render purpose - i.e. twice at least.

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -40,3 +45,44 @@ export const sortTokens = ({
 			a.name.localeCompare(b.name) ||
 			a.network.name.localeCompare(b.network.name)
 	);
+
+export const pinTokensWithBalanceAtTop = ({
```
peterpeterparker, line 49: Can you add a comment to summarize the logic?

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -40,3 +45,44 @@ export const sortTokens = ({
...
+		.map((token) => ({
+			...token,
+			usdValue: nonNullish($exchanges?.[token.id])
+				? usdValue({
+						token,
+						balances: $balancesStore,
+						exchanges: $exchanges
+					})
+				: 0,
+			balance: Number(
```
peterpeterparker, line 68: ditto

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: - it is not defined for any type, that is why i put it in a concatenation mapping/filter/sorting
- I wanted to talk with you about that: I agree, since we are already calculating it twice, better to put it in a single place. My question is: in a PR before this one or in a PR after this one?

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: on it

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: on it

src/frontend/src/lib/derived/tokens.derived.ts
AntonioVentilii-DFINITY: As a first thought about it, I am thinking that ultimately we will have to have 3 sorts separately:
1. tokens with balance/usd_value
2. tokens that are forced to be pinned on top
3. all the others sorted by market_cap/name

So, the best way to reduce excessive calculation could be starting to add some props already to the tokens (as you suggested in the next comment)

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: on it

AntonioVentilii-DFINITY: check PR #1785 please

AntonioVentilii-DFINITY: adapted code to use the new prop

src/frontend/src/lib/derived/tokens.derived.ts
peterpeterparker: In this PR you do `.map().filter().filter()` to compute `tokensWithBalanceToPin`. We can start by using a `.reduce()` maybe?

AntonioVentilii-DFINITY: check last commit please

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -40,3 +44,60 @@ export const sortTokens = ({
...
+
+				const usdBalance = token.usdBalance ?? 0;
+
+				if (usdBalance > 0 || balance > 0) {
+					acc.push({ ...token, balance });
+				}
+
+				return acc;
+			},
+			[] as (TokenUi & { balance: number })[]
```
peterpeterparker, line 89: I'm not 100% sure, you gonna add the balance to `TokenFinancialData` in another PR correct, that's what we discussed?

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -40,3 +44,60 @@ export const sortTokens = ({
...
+						value: $balancesStore?.[token.id]?.data ?? BigNumber.from(0),
+						unitName: token.decimals,
+						displayDecimals: token.decimals
+					})
+				);
+
+				const usdBalance = token.usdBalance ?? 0;
+
+				if (usdBalance > 0 || balance > 0) {
+					acc.push({ ...token, balance });
```
peterpeterparker, line 84: nth: immutable answer `return [...acc, {...token, balance}]`

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: correct! I was working on it with PR #1798

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: Right! maybe there is some lint for that too

PR #1780, opened 2024-07-16
refactor(docker): Build dependencies without individual packages

bitdivine:
# Motivation
We would like to add a command line tool and we may briefly have another canister before it is moved to a separate repository.  In its current form, the docker build fails when additional packages are added.

# Changes
- When pre-building dependencies, get the list of dependencies from Cargo.toml rather than duplicating the list.

# Tests
- Docker builds and tests should work as before.


Dockerfile
```diff
@@ -39,10 +39,7 @@ COPY Cargo.toml .
 COPY src/backend/Cargo.toml src/backend/Cargo.toml
 COPY src/shared/Cargo.toml src/shared/Cargo.toml
 ENV CARGO_TARGET_DIR=/cargo_target
-RUN mkdir -p src/backend/src \
-    && touch src/backend/src/lib.rs \
-    && mkdir -p src/shared/src \
-    && touch src/shared/src/lib.rs \
+RUN yq -oy .workspace.members[] Cargo.toml | xargs -I{} cargo new --lib {} \
```
peterpeterparker, line 42: Spontaneously, I would actually be happy if we should not changed this because it's similar code as in II.

The all Dockerfile and scripts is basically what's done in II.

It's nice to have similar approach and way to build, that one way project improvements benefits the other.

bitdivine: You are right.  This change would benefit II as well.

peterpeterparker: Cool then!

PR #1782, opened 2024-07-17
fix(devops): Fix docker yq installation

bitdivine:
# Motivation
The installation of `yq` in the Dockerfile does not work; it installs an empty executable instead.  Fun fact: An empty executable is the simplest possible executable that always returns successfully.  So the check whether `yq` was installed also returned successfully, even though `yq` was not installed.

# Changes
- Change the curl flags to actually download `yq`.
- Change the check that `yq` is installed to guard against accepting an empty executable.

# Tests
In local development, I was now able to use `yq` successfully further down the Dockerfile.


peterpeterparker (approved):
LGTM

PR #1784, opened 2024-07-17
test(devops): Add upgrade test for new credentials init args

lmuntaner:
# Motivation

We introduced new init args and we'd like to test that the canister is upgradable from a previous canister without them.

# Changes

* Add an optional argument to `setup_with_custom_wasm` and `upgrade_latest_wasm` which are the arguments and set it to `None` in all current calls.
* Add v0.0.25 backend wasm.
* New init types for the v0.0.25 canister.
* New upgrade test credentials_init_args.

# Tests

Only test changes.

### DIFF
--- a/src/backend/tests/it/upgrade/credentials_init_args.rs
+++ b/src/backend/tests/it/upgrade/credentials_init_args.rs
@@ -1,0 +1,32 @@
+use crate::upgrade::constants::BACKEND_V0_0_25_WASM_PATH;
+use crate::upgrade::types::{ArgV0_0_25, InitArgV0_0_25};
+use crate::utils::mock::CALLER;
+use crate::utils::pocketic::{setup_with_custom_wasm, update_call, upgrade_latest_wasm};
+use candid::{encode_one, Principal};
+use shared::types::token::UserToken;
+use shared::types::user_profile::UserProfile;
+
+#[test]
+fn test_upgrade_credential_init_args() {
+    // Deploy a released canister
+    let arg = ArgV0_0_25::Init(InitArgV0_0_25 {
+        ecdsa_key_name: "master_ecdsa_public_key_fscpm-uiaaa-aaaaa-aaaap-yai".to_string(),
+        allowed_callers: vec![Principal::from_text(CALLER).unwrap()],
+    });
+    let encoded_arg = encode_one(arg).unwrap();
+    let pic_setup = setup_with_custom_wasm(BACKEND_V0_0_25_WASM_PATH, Some(encoded_arg));
+
+    // Test a call
+    let caller = Principal::from_text(CALLER).unwrap();
+    let result = update_call::<Vec<UserToken>>(&pic_setup, caller, "list_user_tokens", ());
+
+    assert!(result.is_ok());
+
+    // Upgrade canister with new wasm
+    upgrade_latest_wasm(&pic_setup)
+        .unwrap_or_else(|e| panic!("Upgrade canister failed with error: {}", e));
+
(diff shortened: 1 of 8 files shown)


lmuntaner:
@bitdivine @peterpeterparker the PR with the upgrade test

src/backend/tests/it/upgrade/credentials_init_args.rs
```diff
@@ -0,0 +1,32 @@
...
+    let pic_setup = setup_with_custom_wasm(BACKEND_V0_0_25_WASM_PATH, Some(encoded_arg));
+
+    // Test a call
+    let caller = Principal::from_text(CALLER).unwrap();
+    let result = update_call::<Vec<UserToken>>(&pic_setup, caller, "list_user_tokens", ());
+
+    assert!(result.is_ok());
+
+    // Upgrade canister with new wasm
+    upgrade_latest_wasm(&pic_setup)
```
peterpeterparker, line 26: LGTM. I was just wondering, given the recent changes added extra field to the `Config`, should we get the `caller_eth_address` before the upgrade, then proceed with the upgrade and get the same address again for comparison purposes? This way, we ensure that even though we added fields in the config, the struct remains intact and the derived key is still derived correctly. Or do you think that would not add much value?

lmuntaner: That makes a lot of sense.

I had to add the encoded arg as optional parameter to the helper to upgrade as well, but it's still working.

Thanks!

src/backend/tests/it/upgrade/credentials_init_args.rs
```diff
@@ -0,0 +1,39 @@
...
+    let allowed_callers = vec![Principal::from_text(CALLER).unwrap()];
+    // Deploy a released canister
+    let initial_arg = ArgV0_0_25::Init(InitArgV0_0_25 {
+        ecdsa_key_name: ecdsa_key_name.clone(),
+        allowed_callers: allowed_callers.clone(),
+    });
+    let encoded_initial_arg = encode_one(initial_arg).unwrap();
+    let pic_setup = setup_with_custom_wasm(BACKEND_V0_0_25_WASM_PATH, Some(encoded_initial_arg));
+
+    // Test a call
```
peterpeterparker, line 20: Maybe `// Get ETH address before upgrade for post-upgrade test` or something?

src/backend/tests/it/upgrade/credentials_init_args.rs
```diff
@@ -0,0 +1,39 @@
...
+    });
+    let encoded_updated_arg = encode_one(updated_arg).unwrap();
+
+    // Upgrade canister with new wasm
+    upgrade_latest_wasm(&pic_setup, Some(encoded_updated_arg))
+        .unwrap_or_else(|e| panic!("Upgrade canister failed with error: {}", e));
+
+    let after_upgrade_result = update_call::<String>(&pic_setup, caller, "caller_eth_address", ());
+
+    assert_eq!(initial_result.unwrap(), after_upgrade_result.unwrap());
```
peterpeterparker, line 38: Maybe `.expect("Initial ETH address err")` instead of `unwrap`?

peterpeterparker (approved):
LGTM, thanks

src/backend/tests/it/upgrade/credentials_init_args.rs
lmuntaner: Done!

src/backend/tests/it/upgrade/credentials_init_args.rs
lmuntaner: Good point, done!

PR #1789, opened 2024-07-17
style(frontend): IconChevronDown will inherits the color from the parent

AntonioVentilii-DFINITY:
# Motivation

Since we will be using icon IconChevronDown in other components, with other colors, we make the color inheritable.


peterpeterparker (approved):
LGTM, thx

PR #1790, opened 2024-07-17
feat(backend): Implement get_or_create_user_profile

lmuntaner:
# Motivation

Manage users's credentials.

In this PR, implement the get_or_create_user_profile.

# Changes

* Change the endpoint to update and not query. This was a mistake from before.
* Add the `version` field in StoredUserProfile because it was missing. It's still not used because we are not updating the user profile yet.
* Add two new maps in the state to store the user profiles.
* Implement a default StoredUserProfile and a converter of UserProfile.
* New helper `get_or_create` to get or create a user profile.
* Implement the endpoint get_or_create_user_profile.

# Tests

* Add two tests for get_or_create_user_profile.


lmuntaner:
@bitdivine @peterpeterparker please review

src/backend/src/lib.rs
```diff
@@ -126,6 +135,8 @@ pub struct State {
     /// Introduced to support a broader range of user-defined custom tokens, beyond just ERC20.
     /// Future updates may include migrating existing ERC20 tokens to this more flexible structure.
     custom_token: CustomTokenMap,
+    user_profile: UserProfileMap,
+    user_profile_updated: UserProfileUpdatedMap,
```
peterpeterparker, line 139: Maybe I'm dumb but, I dont not understand why there are two Maps?

src/backend/src/user_profile.rs
```diff
@@ -0,0 +1,22 @@
...
+    user_profile_map: &mut StableBTreeMap<(u64, StoredPrincipal), Candid<StoredUserProfile>, VMem>,
+    user_profile_updated_map: &mut StableBTreeMap<StoredPrincipal, u64, VMem>,
+) -> Candid<StoredUserProfile> {
+    if let Some(updated) = user_profile_updated_map.get(&principal) {
+        user_profile_map
+            .get(&(updated, principal))
+            .expect("Failed to fetch user from user profile map but it's present in updated map")
+    } else {
+        let now = time();
+        let default_profile = StoredUserProfile::default(now);
```
peterpeterparker, line 17: Not sure why we pass now as a parameter instead of scoping the creating of this timestamp?

Moreover, if we don't pass anything we can implement the default trait

```
impl Default for StoredUserProfile {
    fn default() -> Self {
```

src/shared/src/impls.rs
```diff
@@ -45,3 +48,34 @@ impl TokenVersion for CustomToken {
...
+            created_timestamp: now,
+            updated_timestamp: now,
+            version: None,
+        }
+    }
+}
+
+impl UserProfile {
+    #[must_use]
+    pub fn from_stored(user: StoredUserProfile) -> UserProfile {
```
peterpeterparker, line 67: That's the pattern we have for tokens

```
impl From<&Token> for CustomTokenId {
    fn from(token: &Token) -> Self {
```

For consistency, should we replicate it here?

src/backend/backend.did
```diff
@@ -99,7 +99,7 @@ service : (Arg) -> {
   caller_eth_address : () -> (text);
   eth_address_of : (principal) -> (text);
   get_canister_status : () -> (CanisterStatusResultV2);
-  get_or_create_user_profile : () -> (UserProfile) query;
+  get_or_create_user_profile : () -> (UserProfile);
```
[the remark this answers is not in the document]
lmuntaner: I thought that Oisy didn't rely on query calls and only on update calls for security reasons, isn't that the case?

src/backend/src/lib.rs
lmuntaner: The key of the first map is a tuple of updated and principal. That means I can't select only based on principal. But it allows me to select by updated, which will be necessary for get_users.

At the same time, we can't use only the principal as key, because due to the cycle consumption, we might not be able to iterate through the whole map to filter per updated at.

In the end the solution was two maps:
* One to get profiles filtered by updated.
* One to get the updated by principal, and then use the pair to get the user profile from the other map.

Does that make sense?

src/backend/src/user_profile.rs
lmuntaner: I wanted to keep the implementation of the class pure. `time` can only used within the canister, this way we could create one in a test for example.

What's the advantage of using the the default trait?

src/backend/src/lib.rs
```diff
@@ -28,26 +27,34 @@ use shared::types::custom_token::{CustomToken, CustomTokenId};
...
 mod guards;
 mod token;
+mod user_profile;
 
 type VMem = VirtualMemory<DefaultMemoryImpl>;
 type ConfigCell = StableCell<Option<Candid<Config>>, VMem>;
 type UserTokenMap = StableBTreeMap<StoredPrincipal, Candid<Vec<UserToken>>, VMem>;
 type CustomTokenMap = StableBTreeMap<StoredPrincipal, Candid<Vec<CustomToken>>, VMem>;
+/// Map of (`updated_timestamp`, `user_principal`) to `UserProfile`
+type UserProfileMap = StableBTreeMap<(u64, StoredPrincipal), Candid<StoredUserProfile>, VMem>;
```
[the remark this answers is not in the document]
lmuntaner: Good point. I added the `Timestamp` type but I kept the comment because it specifies which timestamp is used as key. I believe that's important. And having two types `UpdatedTimestamp` and `CreatedTimestamp` is a bit too much, wdyt?

src/backend/src/lib.rs
```diff
@@ -505,16 +516,19 @@ fn add_user_credential(request: AddUserCredentialRequest) {
...
 fn get_or_create_user_profile() -> UserProfile {
-    // TODO: Implement https://dfinity.atlassian.net/browse/GIX-2648
-    let credentials: Vec<UserCredential> = Vec::new();
-    UserProfile {
-        credentials,
-        created_timestamp: time(),
-        updated_timestamp: time(),
-        version: None,
-    }
+    ic_cdk::println!("in da get_or_create_user_profile");
```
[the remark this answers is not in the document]
lmuntaner: thanks!

src/shared/src/impls.rs
lmuntaner: Done thanks!

src/backend/backend.did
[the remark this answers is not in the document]
peterpeterparker: Query+update

src/backend/src/lib.rs
peterpeterparker: > One to get profiles filtered by updated.

Thanks, I understand the logic now. However, I'm wondering for which features we need to get a list of users by timestamp?

src/backend/src/user_profile.rs
peterpeterparker: Not a strong opinion but, I rarely create object by passing a timestamp as parameter.

i.e. I rarely did something like that

```
class Something {
   constructor(private now: number) {}
}

const tmp = new Something(Date.now());
```

I'm also a Rust noob but, have the feeling that generally `::default()` accept, is used with, no arguments.

For those reason, if you would not like to use the time in the implementation, maybe instead of using `default` we an implementation named `from` or `from_something` or `with_something`?

```
let default_profile = StoredUserProfile::from_timestamp(now);
```

That said, I can live with `default(now)` so not that a strong opinion neither. If you want to go with that for sure I'm fine with it.

src/backend/src/lib.rs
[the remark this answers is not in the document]
peterpeterparker: Comment is not really useful to me but, for sure I don't mind if it's there 🙂

src/backend/backend.did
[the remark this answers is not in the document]
lmuntaner: Ah, same as in NNS dapp. In that case, yes. I'll split it.

src/backend/src/lib.rs
lmuntaner: For the Rewards canister that Frederik is working on.

src/backend/src/user_profile.rs
lmuntaner: I agree with you on passing now. But I thought that in this case it would be beneficial because the "now" needs to be executed within a canister. Which means that if I put it in the class, then that class won't be usable in a test for example.

lmuntaner:

src/backend/src/lib.rs
peterpeterparker: I see. Now I understand the features but, I'm not sure about the proposed structure of the information I have to say.

Currently those are the maps defined in the PR:

```
type UserProfileMap = StableBTreeMap<(Timestamp, StoredPrincipal), Candid<StoredUserProfile>, VMem>;
type UserProfileUpdatedMap = StableBTreeMap<StoredPrincipal, Timestamp, VMem>;
```

This works fine but, I'm not sure it serves best the first goal of the feature - having user profiles.

For me, the main required features are:

- On sign-in, a profile is resolved for the user principal
- If no principal exists, a profile is created
- Maybe in the future, user can edit their profile in the settings page, for example setting a nickname

And additionally, for gameficication reasons, we also need

- A list of the user principals which completed the verification after a time x

For the main features, I got the feeling that it would probably be performant to have a map principal <> profile

```
type UserProfileMap = StableBTreeMap<StoredPrincipal, Candid<StoredUserProfile>, VMem>;
```

That way, on each login there could be a single access to a map and any additional requests are also limited to getting the information directly for the caller.

Regarding the additional functions, depends if we are fine duplicating or not the information but, I feel like it's more an "historical" data map than a profile related information.

```
type VerifiedUserProfileHistoryMap = StableBTreeMap<Timestamp, StoredPrincipal, VMem>;
// or
type VerifiedUserProfileHistoryMap = StableBTreeMap<Timestamp, Candid(<StoredRelevantInformation>), VMem>;
```

Not sure if I explain really well what I mean and if it makes sense?

lmuntaner: Those are good points, but I think it's better to do something more generic than storing the data needed by that endpoint now. Moreover, using only the Timestamp by key could be problematic if two share the same timestamp. Very unlikely, but I don't think it's worth the risk.

I believe that @bitdivine already has a use case to retrieve more user data. So, with my proposed implementation, a new endpoint could provide whatever is needed easily and without migrating.

I shared the architecture before implementing and we all discussed it there. Maybe you missed it: https://docs.google.com/document/d/1R-VFUzHsXLewSqMxEpEH3kDWOGdPUQcwGP4Ia5ucB48/edit#heading=h.b9uawa2tmu48

The current architecture is also very performant for the main feature. It just requires two accesses instead of one, but both are constant time.

peterpeterparker: >  Moreover, using only the Timestamp by key could be problematic 

Agree.

> Maybe you missed it

Absolutely.

> It just requires two accesses instead of one, but both are constant time.

Personally, I think it's a cost we should spare because the main goal does not requires it or in other words because it deserves the main goal - i.e. I think we should design firstly the feature for the main feature - i.e. user needs to have profile that are retrieved on sign-in - and implement additional features around it.

That said, of course if Max and you already agree on this architecture, definitely fine for me!

lmuntaner: I don't think it's such a big cost, and it's more flexible than copying the data for the current use case.

peterpeterparker: Like I said as you wish.

bitdivine: By the way, user data is also in say the custom token map.  So it's not as if all the user profile data is in this structure.  Is it worth making the name reflect the actual content, e.g. `user_credentials`?

bitdivine (commented):
Just some first thoughts...

src/backend/src/lib.rs
bitdivine: My feeling is that this works, and while there are a lot of policy decisions that we might change, it isn't a big deal to migrate the data to another structure should we wish to at some point.

src/backend/backend.did
```diff
@@ -56,6 +57,7 @@ type OisyUser = record {
   pouh_verified : bool;
   updated_timestamp : nat64;
 };
+type Result = variant { Ok : UserProfile; Err : GetUserProfileError };
```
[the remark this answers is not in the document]
lmuntaner: Yes, it's generated.

I guess we should move to generate Rust from Candid, hehe.

bitdivine (approved):

src/backend/src/lib.rs
```diff
@@ -505,16 +516,34 @@ fn add_user_credential(request: AddUserCredentialRequest) {
...
+fn create_user_profile() -> UserProfile {
+    let stored_principal = StoredPrincipal(ic_cdk::caller());
+
+    mutate_state(|s| {
+        let stored_user = create_profile(
+            stored_principal,
+            &mut s.user_profile,
+            &mut s.user_profile_updated,
+        );
+        UserProfile::from(&stored_user.clone())
```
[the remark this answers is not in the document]
lmuntaner: That's correct, I can remove the clone if I don't return the result wrapped in `Candid`, which is not necessary.

src/backend/src/lib.rs
lmuntaner: I'd rather keep it as UserProfile because the point is to add more fields to it and not have it only for credentials.

That was recommended and agreed with @peterpeterparker

src/backend/backend.did
```diff
@@ -97,9 +99,10 @@ type UserTokenId = record { chain_id : nat64; contract_address : text };
 service : (Arg) -> {
   add_user_credential : (AddUserCredentialRequest) -> ();
   caller_eth_address : () -> (text);
+  create_user_profile : () -> (UserProfile);
```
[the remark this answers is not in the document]
lmuntaner: Done!

PR #1791, opened 2024-07-17
style(frontend): NetworksSwitcher is moved from the header to the tokens list

AntonioVentilii-DFINITY:
# Motivation

As recent requirements, we move the component `NetworkSwitcher` from the header to the tokens list.

# Changes

- Moved component `NetworkSwitcher` into `Tokens`.
- Created new UI component `ButtonSwitcher` with specific style and used in `NetworkSwitcher`.
- Adapted popover to appear left-to-right

# Results

Screenshot 2024-07-17 at 13 25 04
Screenshot 2024-07-17 at 13 25 11
Screenshot 2024-07-17 at 13 25 25
Screenshot 2024-07-17 at 13 25 52


src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
```diff
@@ -26,18 +25,15 @@
...
 	{disabled}
-	imgSrc={$selectedNetwork?.icon ?? chainFusion}
-	imgAlt={replacePlaceholders($i18n.core.alt.logo, {
-		$name: $selectedNetwork?.name ?? $i18n.networks.chain_fusion
-	})}>{$selectedNetwork?.name ?? $i18n.networks.chain_fusion} <IconChevronDown /></ButtonHero
+	>{$selectedNetwork?.name ?? $i18n.networks.chain_fusion} <IconChevronDown /></ButtonSwitcher
 >
 
-<Popover bind:visible anchor={button} direction="rtl">
+<Popover bind:visible anchor={button} direction="ltr">
```
peterpeterparker, line 36: `ltr` is the default maybe?

peterpeterparker (approved):
LGTM, thanks

Note: there is maybe some logic that redirects the user to the tokens page if the network is changed on another page, like on transactions or settings page. Should we keep that logic or clean it up in another PR?

src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
AntonioVentilii-DFINITY: yes, you are right, just checked. I will remove it

AntonioVentilii-DFINITY:
Well, since it is a "insurance" logic (right?) I would keep it: if something should happen, for whatever reason, that the network is changed in another page, it is always better to fallback to the main list, no?

peterpeterparker:
Makes sense.

PR #1792, opened 2024-07-17
style(frontend): remove bold from network switcher

AntonioVentilii-DFINITY:
# Motivation

After PR #1791 , I noticed that the all the items in the popover of `NetworksSwitcher` are now bold (different from how it was before). That is because it is inherited from `<h2>` in the `Header` component that is:

```html
<div class="flex justify-between mb-6 pb-1 items-center">
	<h2 class="text-base"><slot /></h2>

	<slot name="end" />
</div>
```

Since `Header` is used elsewhere, we will change `NetworksSwitcher`.

## Before

Screenshot 2024-07-17 at 15 34 36

## After

Screenshot 2024-07-17 at 15 35 11


AntonioVentilii-DFINITY:
i didn't do that because `<ul>` is used in other two components in the code and I didn't want to cause alterations

peterpeterparker:
No no all good, thanks!

PR #1794, opened 2024-07-17
chore(frontend): use i18n for WalletConnect button

AntonioVentilii-DFINITY:
# Motivation

We use `i18n` for the strings in the WalletConnect button.


peterpeterparker (approved):
Thanks

PR #1795, opened 2024-07-17
refactor(frontend): icon for WalletConnect button inherits color

AntonioVentilii-DFINITY:
# Motivation

To be able to re-use the WalletConnect icon from the button, we make it inherit the color from the parent.


peterpeterparker (approved):
LGTM, thx

PR #1798, opened 2024-07-18
refactor(frontend): new prop balance in token type

AntonioVentilii-DFINITY:
# Motivation

Since we are using the token balance in a few instances, we set it as prop in type `TokenUi`. That means that we need to change the `token` global store too.

# NOTE

The new prop `balance` is defined as non-optional, so that the type-checks raise an issue when using `Token` or `TokenUi`, and it is easier to keep a consistent flow.
For example, if it is optional, there is no issue with casting from `Token` to `TokenUi`, while in certain instances we want to avoid the casting, since we require the balance prop (even if `undefined` as value)

# Changes

- Add prop `balance` to `TokenFinancialData`.
- Change derived store `enabledNetworkTokensUi` to set the property.
- Adapt the `balance` global store (in another PR we will focus on remove this redundant store).
- Create new derived `pageTokenUi`.
- Adapt the token context types.

# Tests

No changes in local replica.


AntonioVentilii-DFINITY:
@peterpeterparker  let me know if you want to break this PR into smaller parts, but mostly it is just substituting the usage of the new `TokenUi` where needed

AntonioVentilii-DFINITY:
I will close this one and create a new PR from scratch when we merge [branch style(frontend)/new-ui-design-july-2024](https://github.com/dfinity/oisy-wallet/tree/style(frontend)/new-ui-design-july-2024) into `main`

PR #1800, opened 2024-07-18
chore(frontend): remove unnecessary IcToken casting

AntonioVentilii-DFINITY:
# Motivation

I noticed that casting the token as `IcToken` it is not really necessary since those function accepts `Partial<IcToken>` as type.


peterpeterparker (approved):
Nice, thx

PR #1801, opened 2024-07-18
refactor(frontend): unify IcToken casting under a global store

AntonioVentilii-DFINITY:
# Motivation

Since we cast global stores `token` and `tokenWithFallback` to IcToken a few times, it is better to unify under a global store, so that it is easier to control it, and eventually remove it in future.

# Changes

- Create global store `tokenAsIcToken`.
- Create global store `tokenWithFallbackAsIcToken`.
- Adapt the code.


peterpeterparker (approved):
LGTM, thx

PR #1806, opened 2024-07-18
style(frontend): blur token list when is logged out

AntonioVentilii-DFINITY:
# Motivation

While the user in logged out we blur the token list:

Screenshot 2024-07-18 at 14 53 16


peterpeterparker (approved):
Not sure about the no pointer nor the blur personally but, I assumed it's the requirements, so bit it. PR itself LGTM, thanks.

PR #1807, opened 2024-07-18
feat(backend): Gate api methods

bitdivine:
# Motivation
When we split the backend into two canisters, we will need to disable APIs gradually and selectively:

```mermaid
timeline
    title Data Migration
    Now : backend (user data read-write, eth read-sign)
    Pre-migration : backend (user data read-write, eth read-sign)
       : new_backend (all user-facing APIs disabled)
    In-migration : backend (readonly)
       : new_backend (all user-facing APIs disabled)
    Data transfer completed : backend (readonly)
       : new_backend (user data read-write)
    UI uses new backend
       : backend (signing read-sign)
       : new_backend (user data read-write)
```

Note:
- When threshold keys are readonly, users will be able to get their Ethereum public key and so see their balances but will not be able to transact.
- When user data is readonly, users will be able to see their tokens as usual but not be able to create new accounts or edit their accounts.

# Changes
- Add init args to selectively disable API methods.
- Apply guards to all user facing methods that access either user data or threshold signatures.

# Tests
- Unit tests are included to  ensure that the enums behave as expected.

# TODO
- PR with an API method that enables or disables features without reinstall.
- Test that after signing is disabled, the new canister will indeed not provide public keys or sign anything.


anedos-dfinity:
If I understand correctly, this architecture allows for the Oisy BE migration to occur while Oisy is still live - but read-only - is this the idea behind it ?

bitdivine:
This _does_ support readonly during migration but the API toggles are needed with or without.  For example, if reads are still enabled on the signer or signing on the backend after migration and we call the wrong canister because a user hasn't refreshed so is using an old client, we are in a worse state than before.

src/backend/backend.did
```diff
@@ -10,6 +10,7 @@ type AddUserCredentialRequest = record {
   current_user_version : opt nat64;
   credential_spec : CredentialSpec;
 };
```
bitdivine, line 12: We wish to be able to make features either fully enabled, fully disabled, or enabled for read methods only.  This type is used to represent which of these states a feature is in.

src/backend/tests/it/utils/pocketic.rs
```diff
@@ -108,6 +108,7 @@ pub(crate) fn init_arg() -> Arg {
             issuer_origin: ISSUER_ORIGIN.to_string(),
             credential_type: CredentialType::ProofOfUniqueness,
         }]),
+        api: None,
```
frederikrothenberger, line 111: @bitdivine: Will you add some integration tests that exercise the API disabling?

bitdivine: Disabling APIs matters in the migration, so I was going to test that in the migration e2e tests.  After the migration is complete I expect all of this will be deleted again. :smile:

frederikrothenberger (approved):
LGTM, a little bit sad that the CDK annotations do not allow multiple guards...

bitdivine:
Agreed!  At least, I didn't test but it looks that way based on this option: https://github.com/dfinity/cdk-rs/blob/59795716487fbb8a9910ac503bcea1e0cb08c932/src/ic-cdk-macros/src/export.rs#L12  No idea whether changing the option to a vec would "just work" or whether more is needded.  I see no hints either way in the macro attribute code.

PR #1811, opened 2024-07-19
feat(backend): Implement add_user_credential

lmuntaner:
# Motivation

Validate and store credentials presented by users.

This PR implements the endpoint `add_user_credential`. It verifies a credential and then stores a related data in the StoredUserProfile.

# Changes

* Install `ic-verifiable-credentials`.
* Change request and response types of `add_user_credential`. Response types will be helpful to give feedback to the user and the new parameters in the request are needed to make the endpoint generic of any credential.
* Change backend generated types after changes in Candid.
* New helper `get_credential_config` to get the related config entries of a credential and issuer.
* New helper `add_credential` to store the credential info after it has been verified.
* Add a supported credential in the default init of the tests.
* Impelement `add_user_credential` using the library and new helpers.

# Tests

* Add two tests for `add_user_credential`.


src/backend/tests/it/utils/pocketic.rs
```diff
@@ -99,7 +101,13 @@ fn init_arg() -> Arg {
         ecdsa_key_name: format!("master_ecdsa_public_key_{}", SUBNET_ID).to_string(),
         allowed_callers: vec![Principal::from_text(CALLER).unwrap()],
         ic_root_key_der: None,
-        supported_credentials: None,
+        supported_credentials: Some(vec![SupportedCredential {
+            ii_canister_id: II_CANISTER_ID.to_string(),
```
lmuntaner, line 105: I want to convert this argument along with `issuer_canister_id` to Principal in another PR before we deploy a version of Oisy with this arguments.

lmuntaner:
@peterpeterparker @bitdivine please review

src/backend/src/lib.rs
```diff
@@ -511,9 +515,33 @@ fn list_custom_tokens() -> Vec<CustomToken> {
 
 #[update(guard = "caller_is_not_anonymous")]
 #[allow(clippy::needless_pass_by_value)]
-#[allow(unused_variables)]
-fn add_user_credential(request: AddUserCredentialRequest) {
-    // TODO: Implement https://dfinity.atlassian.net/browse/GIX-2649
+fn add_user_credential(request: AddUserCredentialRequest) -> Result<(), AddUserCredentialError> {
```
peterpeterparker, line 518: For consistency, given that no other public functions returns a result, shouldn't this function trap as well in case of error?

src/backend/src/lib.rs
```diff
@@ -511,9 +515,33 @@ fn list_custom_tokens() -> Vec<CustomToken> {
...
+    let user_principal = ic_cdk::caller();
+    let stored_principal = StoredPrincipal(user_principal);
+    let current_time_ns = time() as u128;
```
peterpeterparker, line 521: I would cast the time where the cast is required, assuming it's for a specific usage. This for readability reasons.

src/backend/src/user_profile.rs
```diff
@@ -35,3 +40,31 @@ pub fn create_profile(
...
+    user_profile_map: &mut StableBTreeMap<(u64, StoredPrincipal), Candid<StoredUserProfile>, VMem>,
+    user_profile_updated_map: &mut StableBTreeMap<StoredPrincipal, u64, VMem>,
+) -> Result<(), AddUserCredentialError> {
+    if let Ok(user_profile) = get_profile(principal, user_profile_map, user_profile_updated_map) {
+        let now = time();
+        let user_credential = UserCredential {
+            credential_type: credential_type.clone(),
+            verified_date_timestamp: Some(now),
+        };
+        let mut credentials = user_profile.credentials.clone();
```
peterpeterparker, line 56: Not sure, should we set the `user_profile` as mutable to direcly modif the credentials and spare this clone?

src/backend/src/user_profile.rs
```diff
@@ -35,3 +40,31 @@ pub fn create_profile(
...
+        credentials.insert(credential_type.clone(), user_credential);
+        let updated_profile = StoredUserProfile {
+            created_timestamp: user_profile.created_timestamp,
+            updated_timestamp: now,
+            version: Some(user_profile.version.map_or(1, |v| v + 1)),
```
peterpeterparker, line 61: Can you implement a similar trait as `TokenVersion` for user profile.

Maybe we should actually rename `TokenVersion` and reuse it for tokens and any types of entity that has a field version.

src/backend/src/user_profile.rs
```diff
@@ -35,3 +40,31 @@ pub fn create_profile(
...
     }
 }
+
+pub fn add_credential(
+    principal: StoredPrincipal,
+    credential_type: &CredentialType,
+    user_profile_map: &mut StableBTreeMap<(u64, StoredPrincipal), Candid<StoredUserProfile>, VMem>,
+    user_profile_updated_map: &mut StableBTreeMap<StoredPrincipal, u64, VMem>,
+) -> Result<(), AddUserCredentialError> {
+    if let Ok(user_profile) = get_profile(principal, user_profile_map, user_profile_updated_map) {
```
peterpeterparker, line 50: It would maybe be nice to scope the update of the user profile in an implementation? I mean not the persistence, just the contruction of the updated object.

src/backend/src/lib.rs
lmuntaner: I wanted to give feedback to the frontend and show different UX. I believe that an error with the credentials should be reported to the user so that they can report it to the issuer. Versus a problem with Oisy backend can have a generic error to try again.

And I don't think that using trap to differentiate the errors is very developer friendly.

src/backend/src/lib.rs
lmuntaner: Done!

src/backend/src/user_profile.rs
lmuntaner: Discussed offline. I implemented the TokenVersion and therefore cloned a whole user_profile and credentials.

src/backend/src/user_profile.rs
lmuntaner: Done!

src/backend/src/user_profile.rs
lmuntaner: Good idea, done!

lmuntaner:

src/backend/src/lib.rs
peterpeterparker: Developer needs to try / catch anyway but works for me if the feature requires to display diffent UX in case of errors.

peterpeterparker (approved):
LGTM, thx

As usual, I'm happy if @bitdivine review it as well before merge.

lmuntaner:
> LGTM, thx
> 
> As usual, I'm happy if @bitdivine review it as well before merge.

@bitdivine is on holidays this week and we need to finish this before I go on holidays next week.

Is it ok if I merge and then we ask @bitdivine to review it? This endpoint is not yet used and when used will be behind a feature flag. Until then, the whole backend could change without any disruption.

peterpeterparker:
No risk, no fun 😉

Joking. Sure makes sense, let's just ensure he reviews the code before going live.

PR #1812, opened 2024-07-19
feat(frontend): add informative modals when not signed in

AntonioVentilii-DFINITY:
# Motivation

When user is not signed in, we display two new voices in the header:

1. What can I do with my wallet (from now on "What")
2. How does Oisy work (from now on "How")

However, when the screen is too small, we collapse them into a button that shows a popover menu.

# Changes

- Add new texts into i18n JSON.
- New modal for "What".
- New modal for "How".
- Common component for the button to open both, and create derived components for "What" and "How".
- Menu component that shows both buttons with larger screens and single menu button (+popover) otherwise.
- Add new `AboutMenu` component in `Header`.

# Tests

### Big screens

Screenshot 2024-07-22 at 11 05 16
Screenshot 2024-07-22 at 11 05 27

### Smaller screens

Screenshot 2024-07-19 at 15 09 59
Screenshot 2024-07-19 at 15 10 22
Screenshot 2024-07-22 at 11 05 44


src/frontend/src/lib/components/hero/about/AboutMenu.svelte
```diff
@@ -0,0 +1,43 @@
...
+</script>
+
+<div class="hidden md:flex gap-4">
+	<AboutWhat />
+	<AboutHow />
+</div>
+
+<div class="flex md:hidden">
+	<button
+		class="user icon desktop-wide !bg-black !bg-opacity-30 !border !border-white !border-opacity-30"
```
AntonioVentilii-DFINITY, line 23: @peterpeterparker  IIRC, I think you don't like it when we use "!"; do you have any suggestion here?

src/frontend/src/lib/components/hero/about/AboutHow.svelte
```diff
@@ -0,0 +1,16 @@
...
+	import AboutItem from '$lib/components/hero/about/AboutItem.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+
+	export let asMenuItem = false;
+	export let onClick: (() => void) | undefined = undefined;
+</script>
+
+<AboutItem
+	labelAsTitle={$i18n.about.how.text.title}
```
peterpeterparker, line 11: can we use slot to pass text?

src/frontend/src/lib/components/hero/about/AboutHow.svelte
```diff
@@ -0,0 +1,16 @@
...
+	labelAsMenuItem={$i18n.about.how.text.sub_title}
+	openModal={modalStore.openAboutHow}
+	{asMenuItem}
+	{onClick}
```
peterpeterparker, line 15: would it work by bubbling the `on:click`?

src/frontend/src/lib/components/hero/about/AboutMenu.svelte
peterpeterparker: Yeah that definitely cannot go in main 😉.

src/frontend/src/lib/components/hero/about/AboutMenu.svelte
```diff
@@ -0,0 +1,43 @@
...
+		<ul class="flex flex-col gap-4 list-none">
+			<li><AboutWhat asMenuItem onClick={() => (visible = false)} /></li>
+			<li><AboutHow asMenuItem onClick={() => (visible = false)} /></li>
+		</ul>
+	</Popover>
+</div>
+
+{#if $modalAboutWhat}
+	<AboutWhatModal />
+{/if}
```
peterpeterparker, line 39: `else if` just in case there is ever a race condition that would lead to display two modals at the same time?

src/frontend/src/lib/components/hero/about/AboutMenu.svelte
peterpeterparker: e.g. don't use the `user` style class if that the root cause of the issue

src/frontend/src/lib/components/hero/about/AboutHow.svelte
AntonioVentilii-DFINITY: check last version, is that what you meant?

src/frontend/src/lib/components/hero/about/AboutHow.svelte
AntonioVentilii-DFINITY: you mean re-dispatching the event? like  `dispatch('click')` in AboutItem?

because i tried to call on:click from there and it didn't work:

`<li><AboutWhat asMenuItem on:click={() => (visible = false)} /></li>`

AntonioVentilii-DFINITY: in my first try it didn't work, but I found out that to bubble it i need to repeat the event

```
class={asMenuItem ? '' : 'text-center text-white text-lg font-bold px-4 whitespace-nowrap'}
on:click={handleClick}
on:click
```

AntonioVentilii-DFINITY: changed

anedos-dfinity:
@AntonioVentilii-DFINITY myself and Artem have updated the content and style of the modals, just FYI.

AntonioVentilii-DFINITY:
@anedos-dfinity thank you for the heads-up! I saw the notification, working on it

src/frontend/src/lib/components/hero/about/AboutItem.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+	export let openModal: () => void;
+	export let asMenuItem = false;
+</script>
+
+<button
+	class={asMenuItem ? '' : 'text-center text-white text-lg font-bold px-4 whitespace-nowrap'}
+	on:click={openModal}
```
peterpeterparker, line 8: this line can be removed

src/frontend/src/lib/components/hero/about/AboutHow.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
+	import AboutItem from '$lib/components/hero/about/AboutItem.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+
+	export let asMenuItem = false;
+</script>
+
+<AboutItem openModal={modalStore.openAboutHow} {asMenuItem} on:click>
```
peterpeterparker, line 10: `on:click={modalStore.openAboutHow}`

src/frontend/src/lib/components/hero/about/AboutMenu.svelte
```diff
@@ -0,0 +1,41 @@
...
+	<button
+		class="about icon desktop-wide"
+		bind:this={button}
+		on:click={() => (visible = !visible)}
+		aria-label={replaceOisyPlaceholders($i18n.about.text.title)}><IconMenu /></button
+	>
+
+	<Popover bind:visible anchor={button} direction="rtl">
+		<ul class="flex flex-col gap-4 list-none">
+			<li><AboutWhat asMenuItem on:click={() => (visible = false)} /></li>
```
peterpeterparker, line 31: nth: you can create a function `const clost = () => visible = false` that way `on:click={close}` and no code duplication

src/frontend/src/lib/components/hero/about/AboutWhat.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
+	import AboutItem from '$lib/components/hero/about/AboutItem.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+
+	export let asMenuItem = false;
+</script>
+
+<AboutItem openModal={modalStore.openAboutWhat} {asMenuItem} on:click>
```
peterpeterparker, line 10: `on:click={modalStore.openAboutWhat}`

src/frontend/src/lib/components/icons/IconMenu.svelte
```diff
@@ -0,0 +1,9 @@
+<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
peterpeterparker, line 1: Can you add a comment about the source?

src/frontend/src/lib/styles/global/button.scss
```diff
@@ -167,4 +168,9 @@ button {
 			}
 		}
 	}
+
+	&.about {
+		background: rgba(0, 0, 0, 0.3);
```
peterpeterparker, line 173: nth: black and white colors are defined as css variable as rgb, can we reuse those variable here too?

src/frontend/src/lib/components/hero/about/AboutHow.svelte
AntonioVentilii-DFINITY: in this case, I cannot bubble the on:click (it says `Duplicate attribute click`). So I would need to re-dispatch the event:

```typescript

const handleClick = (event: MouseEvent) => {
modalStore.openAboutHow();
dispatch('click', event);
};
```

not an issue, it is just that I would need to do the same both in `AboutWhat` and `AboutHow` components. Not really sure if I prefer it like that. Any strong opinion?

src/frontend/src/lib/components/icons/IconMenu.svelte
AntonioVentilii-DFINITY: done

src/frontend/src/lib/styles/global/button.scss
AntonioVentilii-DFINITY: ah right! however, I cannot find a way to adjust the opacity of those variables. I tried a few things and didn't work, for example

1. rgba(var(--color-black), 30%)
2. mixin color.with-opacity

src/frontend/src/lib/components/hero/about/AboutHow.svelte
peterpeterparker: Ah no I get what you are trying to do

```
on:click
on:click={openModal}
```

Maybe it works, but let's try to avoid it. Maybe you are but, I'm not exactly sure how Svelte behaves when handling two clicks and how safe it will be in the future?

Moreover, as we already discussed, even though we are in this weird place between Svelte 4 and 5, let's stick to bubbling events for consistency.

```
const doSomething = () => {
   // do something if needed
   dispace("custmoEvent")
}

on:click={doSomething}
```

Agree?

src/frontend/src/lib/styles/global/button.scss
peterpeterparker: Thanks for trying. Indeed there are two issues preventing us to do so.

- Unlike white, black is not redeclared as a variable so `--color-black` is unknown 😅
- Even if it would be know it would be set as rgb so we can tweak it, it would render to `rgba(rgb(0,0,0),0.4)`

Alright so let's forget about it.

src/frontend/src/lib/components/hero/about/AboutHow.svelte
AntonioVentilii-DFINITY: ah ok! so re-dispatching it is! on it

src/frontend/src/lib/components/hero/about/AboutItem.svelte
AntonioVentilii-DFINITY: done

AntonioVentilii-DFINITY:
Changed to draft until I update it with the latest changes from Figma

AntonioVentilii-DFINITY:
@peterpeterparker updated with latest changes, please review

src/frontend/src/lib/components/hero/about/AboutHow.svelte
```diff
@@ -0,0 +1,20 @@
...
+	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+	import { createEventDispatcher } from 'svelte';
+
+	export let asMenuItem = false;
+
+	const dispatch = createEventDispatcher();
+
+	const openModal = () => {
+		modalStore.openAboutHow();
+		dispatch('click');
```
peterpeterparker, line 14: Should we maybe inverse the order - i.e. dispatch before open modal - to have two backdrop at the top of each other for a shorter period?

If we inverse the order, should we keep 'click' or give it a particular name like `dispatch('icOpenModal')`?

peterpeterparker (approved):
LGTM, already approving.

Just shared a last question if you can have a look before merge.

src/frontend/src/lib/components/hero/about/AboutHow.svelte
AntonioVentilii-DFINITY: - inverse the order: yes, done
- change name: good question... to be consistent, I would say yes: `icOpenAboutModal` ?

peterpeterparker: `icOpenAboutModal` sounds cool if that works and makes sense for you?

PR #1813, opened 2024-07-19
styel(frontend): adjust WalletConnect menu item aligned with the rest

AntonioVentilii-DFINITY:
# Motivation

There is a slight misalignment between the item WalletConnect and the rest, due to the dimensions of the icon.

### Before

Screenshot 2024-07-19 at 19 22 39

### After

Screenshot 2024-07-19 at 19 22 29


src/frontend/src/lib/components/icons/IconWalletConnect.svelte
```diff
@@ -1,5 +1,9 @@
 <!-- source: https://walletconnect.com/brand -->
-<svg fill="none" height="16.6" viewBox="0 0 480 332" width="24" xmlns="http://www.w3.org/2000/svg"
+<script lang="ts">
+	export let size = '20';
```
peterpeterparker, line 3: The size of the icon was original 16.6 / 24 - i.e. rectangle
In this per we are setting a new size of 20 / 20 - i.e square

Does this visually distort the image on screen?
Should we apply a width equals to size * 24 / 16.6?

AntonioVentilii-DFINITY: the general size given by width and height just define the borders, it does not distort the image: apart from making it bigger or smaller, but keeps the proportions. I opted for a square ratio to keep it consistent with the rest, but also because it is easier to adjust in the list

### 16.6 x 24
Screenshot 2024-07-20 at 08 40 58


### 20 x 20
Screenshot 2024-07-20 at 08 41 10

### Going from 20x20 to 13.8x20 (that is the same proportions as 16.6x24)

https://github.com/user-attachments/assets/799b2761-76f2-4103-ad29-0a3243f65574

peterpeterparker: Thanks for the detailed explanation 👍

peterpeterparker (approved):
LGTM, thx

PR #1817, opened 2024-07-22
refactor(backend): Change SupportedCredential type

lmuntaner:
# Motivation

Set the config with the proper type (Principal) as soon as possible. In this case, when the init arguments are passed on deploy.

# Changes

* Set ii_canister_id and issuer_canister_id fields of SupportedCredentials from InitArg as Principal instead of a string.
* Change client types.
* Remove unnecessary conversion to Principal.

# Tests

* Fix init args in tests.
* Tested deploy to the local replica.


lmuntaner:

peterpeterparker (approved):
LGTM, thx

PR #1819, opened 2024-07-22
style(frontend): update style of primary buttons

AntonioVentilii-DFINITY:
# Motivation

We update the style of primary buttons. Furthermore, the `border-radius` for every button type is now `4px`.

### Before
image

### After
image

### DIFF
--- a/src/frontend/src/lib/styles/global/colors.scss
+++ b/src/frontend/src/lib/styles/global/colors.scss
@@ -16,7 +16,7 @@
 	--color-medium-purple: theme(colors.medium-purple);
 	--color-platinum: theme(colors.platinum);
 	--color-mountain-meadow: theme(colors.mountain-meadow);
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/styles/global/colors.scss
```diff
@@ -16,6 +16,7 @@
 	--color-medium-purple: theme(colors.medium-purple);
 	--color-platinum: theme(colors.platinum);
 	--color-mountain-meadow: theme(colors.mountain-meadow);
+	--color-dodger-blue: #016dfc;
```
peterpeterparker, line 19: Can you configure the new color as a tailwind variable as well?

AntonioVentilii-DFINITY: done!

peterpeterparker (approved):
Nice, thanks

PR #1821, opened 2024-07-22
refactor(frontend): updated the Authenticate button when logged out

AntonioVentilii-DFINITY:
# Motivation

We change the Authenticate button to be simpler:

Screenshot 2024-07-22 at 15 16 16
Screenshot 2024-07-22 at 15 16 24

P.S. we delete `ButtonIC` component that is deprecated.

### DIFF
--- a/src/frontend/src/lib/components/hero/HeroSignIn.svelte
+++ b/src/frontend/src/lib/components/hero/HeroSignIn.svelte
@@ -22,6 +22,5 @@
 </div>
 
 <ButtonIC on:click={async () => await signIn({})} testId="login-button">
-	<svelte:fragment slot="action">{$i18n.auth.text.authenticate_with}</svelte:fragment>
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/hero/HeroSignIn.svelte
```diff
@@ -22,6 +22,5 @@
 </div>
 
 <ButtonIC on:click={async () => await signIn({})} testId="login-button">
```
peterpeterparker, line 24: Instead of modifying it, maybe we can remove that `ButtonIc` and just use a standard primary button?

AntonioVentilii-DFINITY: right! good one! done

peterpeterparker (approved):
LGTM, thx

PR #1822, opened 2024-07-22
style(frontend): remove ICP duplicated icon in main receive modal

AntonioVentilii-DFINITY:
# Motivation

We removed separate the ICP addresses and make them as single groups, removing the duplicated icon and adding a small description.

Furthermore we re-arrange the items in `ReceiveAddressWithLogo` component to show the action buttons above the notes.

Screenshot 2024-07-24 at 13 54 51
Screenshot 2024-07-24 at 13 55 01


src/frontend/src/lib/components/ui/Logo.svelte
```diff
@@ -36,6 +37,7 @@
 	class:bg-off-white={color === 'off-white' && !loaded}
 	class:bg-white={color === 'white' && !loaded}
 	class:opacity-10={!loaded}
+	class:invisible
```
peterpeterparker, line 40: Opiniated answer but, I'm not a big fan of having elements in the DOM that are always invisible for aligment. Can we not render the element and use CSS to align the elements?

peterpeterparker (commented):
While looking at the screenshot I also noticed this text aligment which is a bit unfortunate, WDYT? Can we improve that?

350998378-eae5c6ff-a385-4acc-ab1c-dd308d17e303

src/frontend/src/lib/components/ui/Logo.svelte
AntonioVentilii-DFINITY: eheh I agree with you actually. This was the quick solution, not the optimal, let me check

AntonioVentilii-DFINITY: can it be something like this? or you meant rendering nothing at all and use just CSS to align?

```svelte
{#if invisible}
	<div style={`width: ${...}; height: ${...};`}></div>
{:else}
	<Logo ... />
{/if}
```

AntonioVentilii-DFINITY: otherwise:

```svelte
<div class="flex-none w-14" slot="icon">
	{#if !invisibleLogo}
		<Logo
			color="white"
			src={token.network.icon}
			alt={replacePlaceholders($i18n.core.alt.logo, { $name: token.network.name })}
			size="medium"
		/>
	{/if}
</div>
```

AntonioVentilii-DFINITY:
> While looking at the screenshot I also noticed this text aligment which is a bit unfortunate, WDYT? Can we improve that?

yes, for sure! I will work it on a separate PR, since it will require a bit of refactoring of component `ReceiveAddressWithLogo`

src/frontend/src/lib/components/ui/Logo.svelte
peterpeterparker: I would use a CSS grid and set start to 1 or 2 depending if there is a logo or none to render.

AntonioVentilii-DFINITY: Check last commit please

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/components/ui/Logo.svelte
peterpeterparker: Thx

PR #1826, opened 2024-07-23
Start POUH credential

lmuntaner:
# Motivation

Demo the POUH in Oisy.

In this PR, I add an env var to render a button that will start the credential flow in the settings page.

# Changes

* New env var `VITE_POUH_ENABLED`.
* Add a button in Settings that uses the `requestPouhCredential` service.
* Add window features when starting vc flow.
* Set busy screen during vc flow.

# Tests

Not necessary. Only for the demo.


lmuntaner:
Closing because it's not necessary anymore.

PR #1827, opened 2024-07-23
feat(backend): Implement get_users endpoint

lmuntaner:
# Motivation

The Rewards canister needs to have the list of Oisy users and whether they have a verified POUH credential or not.

In this PR, get_users is implemented, which is the endpoint that Rewards canister will call to get Oisy users.

# Changes

* New helper `get_oisy_users`.
* New helper in user_profile `store_user` to avoid code duplication.
* New implementation of OisyUser `from_profile` to convert one from a StoredUserProfile.
* Implement `get_users`.

# Tests

* Test the new endpoint.

### DIFF
--- a/src/backend/src/lib.rs
+++ b/src/backend/src/lib.rs
@@ -31,8 +31,8 @@ use shared::types::custom_token::{CustomToken, CustomTokenId};
 use shared::types::token::{UserToken, UserTokenId};
 use shared::types::transaction::SignRequest;
 use shared::types::user_profile::{
-    AddUserCredentialError, AddUserCredentialRequest, GetUserProfileError, GetUsersRequest,
-    GetUsersResponse, OisyUser, StoredUserProfile, UserProfile,
+    AddUserCredentialError, AddUserCredentialRequest, GetUserProfileError, ListUsersRequest,
+    ListUsersResponse, OisyUser, StoredUserProfile, UserProfile,
 };
 use shared::types::{Arg, InitArg, SupportedCredential, Timestamp};
 use std::borrow::Cow;
@@ -580,15 +580,15 @@ fn get_user_profile() -> Result<UserProfile, GetUserProfileError> {
 
 #[query(guard = "caller_is_allowed")]
 #[allow(clippy::needless_pass_by_value)]
-fn get_users(request: GetUsersRequest) -> GetUsersResponse {
-    // WARNING: The value `DEFAULT_LIMIT_GET_USERS_RESPONSE` must also be determined by the cycles consumption when reading BTreeMap.
+fn list_users(request: ListUsersRequest) -> ListUsersResponse {
+    // WARNING: The value `DEFAULT_LIMIT_LIST_USERS_RESPONSE` must also be determined by the cycles consumption when reading BTreeMap.
 
-    let (users, limit_users_size): (Vec<OisyUser>, u64) =
+    let (users, matches_max_length): (Vec<OisyUser>, u64) =
         read_state(|s| get_oisy_users(&request, &s.user_profile));
 
-    GetUsersResponse {
+    ListUsersResponse {
         users,
-        matches_max_length: limit_users_size,
+        matches_max_length,
     }
 }
 
--- a/src/backend/src/lib.rs
+++ b/src/backend/src/lib.rs
@@ -166,12 +166,6 @@ impl Storable for StoredPrincipal {
     }
 }
 
-impl StoredPrincipal {
-    fn clone_principal(self) -> Principal {
-        self.0.clone()
-    }
-}
-
 #[derive(CandidType, Deserialize)]
 pub struct Config {
     pub ecdsa_key_name: String,
@@ -590,7 +584,7 @@ fn get_users(request: GetUsersRequest) -> GetUsersResponse {
     // WARNING: The value `DEFAULT_LIMIT_GET_USERS_RESPONSE` must also be determined by the cycles consumption when reading BTreeMap.
 
     let (users, limit_users_size): (Vec<OisyUser>, u64) =
-        read_state(|s| get_oisy_users(request, &s.user_profile));
+        read_state(|s| get_oisy_users(&request, &s.user_profile));
 
     GetUsersResponse {
         users,
--- a/src/backend/src/lib.rs
+++ b/src/backend/src/lib.rs
@@ -22,6 +22,7 @@ use ic_stable_structures::{
 };
 use ic_verifiable_credentials::validate_ii_presentation_and_claims;
 use k256::PublicKey;
+use oisy_user::get_oisy_users;
 use serde_bytes::ByteBuf;
 use shared::http::{HttpRequest, HttpResponse};
 use shared::metrics::get_metrics;
@@ -42,6 +43,7 @@ use user_profile::{add_credential, create_profile, get_profile};
 mod assertions;
 mod config;
 mod guards;
+mod oisy_user;
 mod token;
 mod user_profile;
 
@@ -60,7 +62,6 @@ const USER_CUSTOM_TOKEN_MEMORY_ID: MemoryId = MemoryId::new(2);
 const USER_PROFILE_MEMORY_ID: MemoryId = MemoryId::new(3);
 const USER_PROFILE_UPDATED_MEMORY_ID: MemoryId = MemoryId::new(4);
 
-const DEFAULT_LIMIT_GET_USERS_RESPONSE: u64 = 10_000;
 const MAX_SYMBOL_LENGTH: usize = 20;
 
 thread_local! {
@@ -162,6 +163,12 @@ impl Storable for StoredPrincipal {
         Self(Principal::from_slice(
             Blob::<29>::from_bytes(bytes).as_slice(),
         ))
+    }
+}
+
+impl StoredPrincipal {
+    fn clone_principal(self) -> Principal {
+        self.0.clone()
     }
 }
 
@@ -580,14 +587,14 @@ fn get_user_profile() -> Result<UserProfile, GetUserProfileError> {
 #[query(guard = "caller_is_allowed")]
 #[allow(clippy::needless_pass_by_value)]
 fn get_users(request: GetUsersRequest) -> GetUsersResponse {
-    // TODO: Implement https://dfinity.atlassian.net/browse/GIX-2650
     // WARNING: The value `DEFAULT_LIMIT_GET_USERS_RESPONSE` must also be determined by the cycles consumption when reading BTreeMap.
-    let users: Vec<OisyUser> = Vec::new();
+
+    let (users, limit_users_size): (Vec<OisyUser>, u64) =
+        read_state(|s| get_oisy_users(request, &s.user_profile));
+
     GetUsersResponse {
         users,
-        matches_max_length: request
-            .matches_max_length
-            .unwrap_or(DEFAULT_LIMIT_GET_USERS_RESPONSE),
+        matches_max_length: limit_users_size,
     }
 }
 
--- a/src/backend/src/oisy_user.rs
+++ b/src/backend/src/oisy_user.rs
@@ -2,27 +2,32 @@ use crate::{Candid, StoredPrincipal, VMem};
 use candid::Principal;
 use ic_stable_structures::StableBTreeMap;
 use shared::types::{
-    user_profile::{GetUsersRequest, OisyUser, StoredUserProfile},
+    user_profile::{ListUsersRequest, OisyUser, StoredUserProfile},
     Timestamp,
 };
 use std::ops::Bound;
 
-const DEFAULT_LIMIT_GET_USERS_RESPONSE: usize = 10_000;
+const DEFAULT_LIMIT_LIST_USERS_RESPONSE: usize = 10_000;
+pub const PRINCIPAL_MIN: Principal = Principal::from_slice(&[]);
+
+fn get_limit_users_size(request: &ListUsersRequest) -> usize {
+    match request.matches_max_length {
+        Some(num) => match usize::try_from(num) {
+            Ok(val) if val <= DEFAULT_LIMIT_LIST_USERS_RESPONSE => val,
+            _ => DEFAULT_LIMIT_LIST_USERS_RESPONSE,
+        },
+        None => DEFAULT_LIMIT_LIST_USERS_RESPONSE,
+    }
+}
 
 pub fn get_oisy_users(
-    request: &GetUsersRequest,
+    request: &ListUsersRequest,
     user_profile_map: &StableBTreeMap<(u64, StoredPrincipal), Candid<StoredUserProfile>, VMem>,
 ) -> (Vec<OisyUser>, u64) {
-    let limit_users_size: usize = match request.matches_max_length {
-        Some(num) => match usize::try_from(num) {
-            Ok(val) if val <= DEFAULT_LIMIT_GET_USERS_RESPONSE => val,
-            _ => DEFAULT_LIMIT_GET_USERS_RESPONSE,
-        },
-        None => DEFAULT_LIMIT_GET_USERS_RESPONSE,
-    };
+    let limit_users_size: usize = get_limit_users_size(&request);
 
     let start_bound: Bound<(Timestamp, StoredPrincipal)> = match request.updated_after_timestamp {
-        Some(updated) => Bound::Included((updated, StoredPrincipal(Principal::anonymous()))),
+        Some(updated) => Bound::Included((updated, StoredPrincipal(PRINCIPAL_MIN))),
         None => Bound::Unbounded,
     };
     let users: Vec<OisyUser> = user_profile_map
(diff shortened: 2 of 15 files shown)


lmuntaner:

src/backend/src/lib.rs
```diff
@@ -580,14 +581,14 @@ fn get_user_profile() -> Result<UserProfile, GetUserProfileError> {
 #[query(guard = "caller_is_allowed")]
 #[allow(clippy::needless_pass_by_value)]
 fn get_users(request: GetUsersRequest) -> GetUsersResponse {
```
peterpeterparker, line 583: unrelated to this PR but, it crossed my mind: should the endpoint actually be named `list_users` instead of `get_users`?

src/backend/src/lib.rs
```diff
@@ -580,14 +581,14 @@ fn get_user_profile() -> Result<UserProfile, GetUserProfileError> {
...
-    // TODO: Implement https://dfinity.atlassian.net/browse/GIX-2650
     // WARNING: The value `DEFAULT_LIMIT_GET_USERS_RESPONSE` must also be determined by the cycles consumption when reading BTreeMap.
-    let users: Vec<OisyUser> = Vec::new();
+
+    let (users, limit_users_size): (Vec<OisyUser>, u64) =
```
peterpeterparker, line 586: Just ideas:

- You can rename `limit_users_size` to `matches_max_length` if interesting.

- Should we create a type `type OisyUsers = Vec<OisyUser>`?

src/backend/src/oisy_user.rs
```diff
@@ -0,0 +1,35 @@
...
+};
+use std::ops::Bound;
+
+const DEFAULT_LIMIT_GET_USERS_RESPONSE: usize = 10_000;
+
+pub fn get_oisy_users(
+    request: &GetUsersRequest,
+    user_profile_map: &StableBTreeMap<(u64, StoredPrincipal), Candid<StoredUserProfile>, VMem>,
+) -> (Vec<OisyUser>, u64) {
+    let limit_users_size: usize = match request.matches_max_length {
```
peterpeterparker, line 16: nth: maybe we can extract this in a function

src/backend/src/oisy_user.rs
```diff
@@ -0,0 +1,35 @@
...
+        Some(num) => match usize::try_from(num) {
+            Ok(val) if val <= DEFAULT_LIMIT_GET_USERS_RESPONSE => val,
+            _ => DEFAULT_LIMIT_GET_USERS_RESPONSE,
+        },
+        None => DEFAULT_LIMIT_GET_USERS_RESPONSE,
+    };
+
+    let start_bound: Bound<(Timestamp, StoredPrincipal)> = match request.updated_after_timestamp {
+        Some(updated) => Bound::Included((updated, StoredPrincipal(Principal::anonymous()))),
```
peterpeterparker, line 25: `Principal::anonymous` is correct?

This is the solution of Frederik for min and max: https://forum.dfinity.org/t/stable-structure-stablebtreemap-principal-in-key/22974/2?u=peterparker

src/backend/src/oisy_user.rs
```diff
@@ -0,0 +1,35 @@
...
+        None => Bound::Unbounded,
+    };
+    let users: Vec<OisyUser> = user_profile_map
```
peterpeterparker, line 28: I never used `Bound::Unbounded` in a range, that works out? Cool

src/backend/src/oisy_user.rs
```diff
@@ -0,0 +1,35 @@
...
+        .range((start_bound, Bound::Unbounded))
+        .take(limit_users_size)
```
peterpeterparker, line 30: This isn't really performance, should we use a `limit_max_timestamp` instead of limiting the number of users that way we can limit the range?

src/backend/src/user_profile.rs
```diff
@@ -3,9 +3,24 @@ use ic_cdk::api::time;
...
 };
 
+fn store_user(
+    principal: StoredPrincipal,
+    timestamp: Timestamp,
+    user: &StoredUserProfile,
+    user_profile_map: &mut StableBTreeMap<(u64, StoredPrincipal), Candid<StoredUserProfile>, VMem>,
+    user_profile_updated_map: &mut StableBTreeMap<StoredPrincipal, u64, VMem>,
+) {
+    if let Some(old_updated) = user_profile_updated_map.get(&principal) {
```
peterpeterparker, line 16: Just a note: I still have trouble understanding the purpose of each map when I read the PRs. If we can find better names, I would be in favor of it.

src/backend/src/user_profile.rs
```diff
@@ -3,9 +3,24 @@ use ic_cdk::api::time;
 use ic_stable_structures::StableBTreeMap;
 use shared::types::{
     user_profile::{AddUserCredentialError, GetUserProfileError, StoredUserProfile},
-    CredentialType, Version,
+    CredentialType, Timestamp, Version,
 };
 
+fn store_user(
```
peterpeterparker, line 9: `store_profile` or `store_user_profile` for consistency?

src/backend/src/user_profile.rs
```diff
@@ -3,9 +3,24 @@ use ic_cdk::api::time;
...
+    principal: StoredPrincipal,
+    timestamp: Timestamp,
+    user: &StoredUserProfile,
+    user_profile_map: &mut StableBTreeMap<(u64, StoredPrincipal), Candid<StoredUserProfile>, VMem>,
```
peterpeterparker, line 13: nth: if we create types for those StableBTreeMaps it would make the parameters a bit less verbose

src/backend/tests/it/get_users.rs
```diff
@@ -0,0 +1,186 @@
...
+use candid::Principal;
+use ic_verifiable_credentials::issuer_api::CredentialSpec;
+use pocket_ic::PocketIc;
+use shared::types::user_profile::{
+    AddUserCredentialError, AddUserCredentialRequest, GetUsersRequest, GetUsersResponse, OisyUser,
+    UserProfile,
+};
+
+fn create_users(pic_setup: &(PocketIc, Principal), start: u8, end: u8) {
+    for i in start..=end {
```
peterpeterparker, line 16: advance maybe the timers sometimes in the test to not have all the users created with the same timestamp?

src/backend/tests/it/get_users.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    let caller = Principal::from_text(CALLER).unwrap();
+
+    let arg = GetUsersRequest {
+        matches_max_length: None,
+        updated_after_timestamp: None,
+    };
+    let get_users_response = query_call::<GetUsersResponse>(&pic_setup, caller, "get_users", arg);
+
+    assert_eq!(
```
peterpeterparker, line 38: in this test and other below, in addition to the length can we also compare the returned users themselves?

See for example `assert_custom_tokens_eq`

src/backend/tests/it/get_users.rs
```diff
@@ -0,0 +1,186 @@
...
+        get_users_response
+            .expect("Call get_users failed")
+            .users
+            .len(),
+        users_count_after_timestamp as usize + 1
+    );
+}
+
+#[test]
+fn test_get_users_returns_requested_users_count() {
```
peterpeterparker, line 115: Feature wise are we interested in the "count" value? If yes, we should add a dedicated function as `.count()` on a stable map is less costly than listing, mapping, getting the result and counting. Or that's already the plan?

src/backend/tests/it/get_users.rs
```diff
@@ -0,0 +1,186 @@
...
+    pocketic::{query_call, setup, update_call},
+};
+use candid::Principal;
+use ic_verifiable_credentials::issuer_api::CredentialSpec;
+use pocket_ic::PocketIc;
+use shared::types::user_profile::{
+    AddUserCredentialError, AddUserCredentialRequest, GetUsersRequest, GetUsersResponse, OisyUser,
+    UserProfile,
+};
+
```
peterpeterparker, line 14: All tests have positive assertions. Can we add some that test errors?
Notably, not being able to call get_users for an unknown caller.

Maybe I missed it, but is there also a test that returns fewer users than matches_max_length and a test that returns fewer users than the number of users in memory?

src/backend/src/oisy_user.rs
lmuntaner: Why isn't this performant?

We discussed this with Frederik and Max in the document and offline.

If we use an expiration, we never know how many there will be and the size limit is important for two reasons:
* Size limit of the response. Messages have a limit of 2Mb.
* Instruction consumption limit. If the canister reads too many from stable memory, it might run on a limit of instructions. Therefore, we need to limit how much we read. Hence the `take`.

In the [document](https://docs.google.com/document/d/1R-VFUzHsXLewSqMxEpEH3kDWOGdPUQcwGP4Ia5ucB48/edit#heading=h.b9uawa2tmu48), we discussed filtering only the lower bound with a timestamp, and the upper bound with a number of elements.

src/backend/src/user_profile.rs
lmuntaner: I agree and I can do it in another PR. It requires moving some types around because some are more private and can't be moved so easily.

src/backend/src/user_profile.rs
lmuntaner: I understand that and I'll happily change the names when we have new ones.

src/backend/tests/it/get_users.rs
lmuntaner: * Not being able to call for not allowe caller -> Will do!
* returns fewer users than matches_max_length -> Will do!
* returns fewer users than the number of users in memory -> Yes, `test_list_users_returns_requested_users_count` and `test_list_users_returns_filtered_users_by_updated`.

src/backend/tests/it/get_users.rs
lmuntaner: I do that outside this function when needed.

For example, in `test_list_users_returns_filtered_users_by_updated` I create two sets of 10 users with different timestamps.

This way the test controls the time and can manage it.

* Do you see a reason why making sure each user has a different timestamp is important?
* Would you then assume that some X time has passed after calling `create_users`.

src/backend/src/oisy_user.rs
peterpeterparker: range + take + map is less performant than just range

but if you all three are fine with that, for sure so be it

src/backend/tests/it/get_users.rs
lmuntaner: We are not yet interested in the count. But we might be when we implement the summary endpoint. When we come to that, we'll add the new state, I agree.

src/backend/tests/it/get_users.rs
peterpeterparker: > Do you see a reason why making sure each user has a different timestamp is important?

I had once issue because pocket-ic was too fast and therefore had to explicitely add some delay to test both "fast" and "slow" scenarios. I was thinking that here it could be particular useful because the maps are ordered by timestamps.

> I do that outside this function when needed.

If we cover use cases where all users do not have the same time stamp it's good, just felt like it would be nice to have this tested explicitely for above reasons.

src/backend/tests/it/get_users.rs
lmuntaner: Good idea, I added it to this test, but I don't see the value in doing it everywhere.

Creating the expected users adds boilerplate and complexity and I don't see how that will improve the other tests.

peterpeterparker: > I don't see how that will improve the other tests.

Iv'e got the feeling that it can proove itself to be useful when we will start adding more fields in the profile. So I see it as an "investement" for the future.

src/backend/src/oisy_user.rs
lmuntaner: It seems so... Google and ChatGPT helped me here.

Just in case I asked a question to the execution guys: https://dfinity.slack.com/archives/CGZJ7G1J6/p1721734840980959

src/backend/src/lib.rs
lmuntaner: Done

src/backend/src/oisy_user.rs
lmuntaner: done

src/backend/src/oisy_user.rs
lmuntaner: Done!

src/backend/src/user_profile.rs
lmuntaner: done

src/backend/src/oisy_user.rs
```diff
@@ -0,0 +1,40 @@
...
+use candid::Principal;
+use ic_stable_structures::StableBTreeMap;
+use shared::types::{
+    user_profile::{ListUsersRequest, OisyUser, StoredUserProfile},
+    Timestamp,
+};
+use std::ops::Bound;
+
+const DEFAULT_LIMIT_LIST_USERS_RESPONSE: usize = 10_000;
+pub const PRINCIPAL_MIN: Principal = Principal::from_slice(&[]);
```
peterpeterparker, line 11: It does not need to be public maybe?

src/backend/src/oisy_user.rs
peterpeterparker: Cool, good to know!

src/backend/tests/it/get_users.rs
peterpeterparker: Ok

src/backend/src/oisy_user.rs
lmuntaner: Indeed! I removed the `pub`.

src/backend/tests/it/get_users.rs
lmuntaner: Good point!

Like this it made sense also that create_users had some delay creating each user so that each has a different timestmamp as well.

src/backend/src/oisy_user.rs
```diff
@@ -0,0 +1,40 @@
...
+use ic_stable_structures::StableBTreeMap;
+use shared::types::{
+    user_profile::{ListUsersRequest, OisyUser, StoredUserProfile},
+    Timestamp,
+};
+use std::ops::Bound;
+
+const DEFAULT_LIMIT_LIST_USERS_RESPONSE: usize = 10_000;
+const PRINCIPAL_MIN: Principal = Principal::from_slice(&[]);
+
```
bitdivine, line 12: Note: The `get_` prefix is normally not used according to the Rust naming guidelines: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

The same applies to various other getters below.

src/backend/tests/it/list_users.rs
```diff
@@ -0,0 +1,280 @@
...
+        pic_setup.0.advance_time(Duration::new(10, 0));
+        let caller = Principal::self_authenticating(i.to_string());
+        let response = update_call::<UserProfile>(&pic_setup, caller, "create_user_profile", ());
+        let timestamp = pic_setup.0.get_time();
+        let timestamp_nanos = timestamp
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards")
+            .as_nanos();
+        let expected_user = OisyUser {
+            updated_timestamp: timestamp_nanos as u64,
```
bitdivine, line 28: This is fine, but just so that you are aware, in places where wrap-arounds could be critical, it is conventional to do something like:
```
u64::try_from(foo).expect("This conversion should never fail because XYZ")
```

lmuntaner: But in tests casting is ok, right?

src/shared/src/impls.rs
```diff
@@ -131,3 +133,16 @@ impl From<&StoredUserProfile> for UserProfile {
...
 }
+
+impl OisyUser {
+    #[must_use]
+    pub fn from_profile(user: &StoredUserProfile, principal: Principal) -> OisyUser {
+        OisyUser {
+            principal,
+            pouh_verified: user
+                .credentials
+                .contains_key(&CredentialType::ProofOfUniqueness),
```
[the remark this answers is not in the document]
lmuntaner: Good question.

The credential type means two things:
* The credential_type value of the credential.
* The type of credential Oisy considers that to be.

The enum `CredentialType` is about the second one: a way to differentiate credentials.

Yet, I implemented the formatting of the enum:

```
impl fmt::Display for CredentialType {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        match self {
            CredentialType::ProofOfUniqueness => write!(f, "ProofOfUniqueness"),
        }
    }
}
```

So that I could compare the value of the enum with the first `credential_type` in a credential when adding a credential.

I'm not so sure what you mean by "setting the enum variant representatives". Do you mean to add a value in the enum: `ProofOfUniqueness { id: u32 },`?

Thinking again about it, I'm also not so happy with the current solution being scalable to multiple issuers. The question is, is the effort to make it scalable worth the time? How many credentials will Oisy support? less than ten? hundreds?

I believe it will be more towards less than ten, we should be able to use an enum and different entries per each SupportedCredential.

PR #1831, opened 2024-07-23
refactor(frontend): remove Receive and Send buttons from the token list

AntonioVentilii-DFINITY:
# Motivation

Removed Receive and Send buttons from the token list.


peterpeterparker (approved):
LGTM, thx

PR #1835, opened 2024-07-23
chore(backend): Uninstall ic-verifiable-credentials

lmuntaner:
# Motivation

ic-verifiable-credentials makes the wasm size exceed 2Mb.

# Changes

* Uninstall the library and comment all the code that uses the library.
* add_user_credential adds the credential without verifying.

# Tests


peterpeterparker:
I'm not sure yet - the error msg could as well be a false positive. i.e. another error but wrong error msg.

The wasm seems to be around 1Mb when I build it locally.

lmuntaner:
Closing, the wasm size is ok when gzipped.

PR #1836, opened 2024-07-23
fix(frontend): Fix mainnet domain

lmuntaner:
# Motivation

The mainnet domain for the issuer was set to ic0.app but it should be the new icp0.io

# Changes

* Change `MAINNET_DOMAIN`.

# Tests

Only testeable not in local.


peterpeterparker (approved):
Thanks

PR #1838, opened 2024-07-23
fix(frontend): solved navigational loop between /settings and /transactions

AntonioVentilii-DFINITY:
# Motivation

There was a bug so that the `Back` component would always go in loop between the last two navigational page. That was not acceptable when going back from `/settings` to `/transactions`, then from `/transactions` to the main page, since it would go back to `/settings`.


AntonioVentilii-DFINITY:
Closing in favor of PR #1869

PR #1843, opened 2024-07-23
feat(ci): script to check unused components

AntonioVentilii-DFINITY:
# Motivation

The new script will check for unused `svelte` components (excluding `/routes` folder).

### DIFF
--- a/scripts/check.unused.mjs
+++ b/scripts/check.unused.mjs
@@ -2,6 +2,7 @@
 
 import { readFileSync, readdirSync } from 'node:fs';
 import { basename, resolve } from 'node:path';
+import { findFiles } from './utils.mjs';
 
 const RED = '\x1b[31m';
 const GREEN = '\x1b[32m';
@@ -11,14 +12,7 @@ const DATA_DIR = 'src/frontend/src';
 const DATA_DIR_PATH = resolve(process.cwd(), DATA_DIR);
 
 const findSvelteFiles = (dir) =>
-	readdirSync(dir, { withFileTypes: true }).flatMap((file) => {
-		const res = resolve(dir, file.name);
-		return file.isDirectory()
-			? findSvelteFiles(res)
-			: file.isFile() && res.endsWith('.svelte')
-				? [res]
-				: [];
-	});
+	findFiles({ dir, extensions: ['.svelte'], ignoreDirs: ['routes'] });
 
 const searchForFileName = ({ filename, dir }) =>
 	readdirSync(dir, { withFileTypes: true }).some((file) => {
@@ -28,8 +22,7 @@ const searchForFileName = ({ filename, dir }) =>
 			: file.isFile() && readFileSync(res, 'utf-8').includes(filename);
 	});
 
-const checkFileUsage = ({ filename, dir }) =>
-	filename.startsWith('+') || searchForFileName({ filename, dir });
+const checkFileUsage = ({ filename, dir }) => searchForFileName({ filename, dir });
 
 const logResult = (isUsed) => process.stdout.write(isUsed ? `${GREEN}.${NC}` : `${RED}x${NC}`);
 
@@ -50,6 +43,7 @@ const main = async () => {
 
 	if (unusedFiles.length === 0) {
 		console.log(`${GREEN}No unused components found.${NC}`);
+		process.exit(1);
 	} else {
 		unusedFiles.forEach((file) => console.log(`${RED}Unused Svelte file: ${file}${NC}`));
 	}
--- a/scripts/check.unused.mjs
+++ b/scripts/check.unused.mjs
@@ -42,4 +42,7 @@ const main = async () => {
 	}
 };
 
-main().catch(console.error);
+main().catch((err) => {
+	console.error(err);
+	process.exit(1);
+});
--- a/scripts/check.unused.mjs
+++ b/scripts/check.unused.mjs
@@ -29,10 +29,11 @@ const main = async () => {
 
 	if (unusedFiles.length === 0) {
 		console.log(`${GREEN}No unused components found.${NC}`);
-		process.exit(1);
+		process.exit(0);
 	} else {
 		console.log(`${RED}Found ${unusedFiles.length} unused component(s).${NC}`);
 		unusedFiles.forEach((file) => console.log(`${RED}Unused Svelte file: ${file}${NC}`));
+		process.exit(1);
 	}
 };
 
--- a/scripts/check.unused.mjs
+++ b/scripts/check.unused.mjs
@@ -14,25 +14,30 @@ const DATA_DIR_PATH = resolve(process.cwd(), DATA_DIR);
 const findSvelteFiles = (dir) => findFiles({ dir, extensions: ['.svelte'] });
 
 const main = async () => {
-	console.log(`${NC}Scanning ${DATA_DIR} folder to find all .svelte files`);
+	console.log(`${NC}Scanning ${DATA_DIR} folder to find all .svelte files\n`);
 
 	const allSvelteFiles = findSvelteFiles(DATA_DIR_PATH);
 
-	const potentialUnusedFiles = allSvelteFiles.filter((file) => !dirname(file).includes('routes'));
+	let potentialUnusedFiles = allSvelteFiles.filter((file) => !dirname(file).includes('routes'));
 
-	const unusedFiles = allSvelteFiles.reduce((acc, file) => {
+	allSvelteFiles.forEach((file) => {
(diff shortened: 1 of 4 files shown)


scripts/check.unused.mjs
```diff
@@ -0,0 +1,58 @@
...
+
+const logResult = (isUsed) => process.stdout.write(isUsed ? `${GREEN}.${NC}` : `${RED}x${NC}`);
+
+const main = async () => {
+	console.log(`${NC}Scanning ${DATA_DIR} folder to find all .svelte files`);
+	console.log(`  ${GREEN}.${NC} means the .svelte is imported in another file`);
+	console.log(`  ${RED}x${NC} means the .svelte is not imported and should likely be removed`);
+
+	const svelteFiles = findSvelteFiles(DATA_DIR_PATH);
+	const unusedFiles = svelteFiles.filter((file) => {
```
peterpeterparker, line 42: I'm not sure what the best option is for performance reasons. If I understand correctly, we are reading the content of all files for every single file.

For example, we find a.svelte and then we check every file to see if a.svelte is present.

Would it be more efficient to build a list of all filenames and then visit each file to read its content and eliminate the entries until there are no more entries to eliminate or all files have been browsed? If ultimately the list is not empty, those remaining entries are the leaves.

Just thinking that if we add the script to the CI if would be nice if it runs the fastest.

scripts/check.unused.mjs
```diff
@@ -0,0 +1,58 @@
...
+		return !isUsed;
+	});
+
+	console.log('\n');
+
+	if (unusedFiles.length === 0) {
+		console.log(`${GREEN}No unused components found.${NC}`);
+	} else {
+		unusedFiles.forEach((file) => console.log(`${RED}Unused Svelte file: ${file}${NC}`));
+	}
```
peterpeterparker, line 55: `process.exit(1)` if there are unused files?

peterpeterparker (commented):
I wonder if this could potentially be solved with TSC or Svelte language tools—i.e., getting the tree of Svelte modules that is built and comparing it with the list of files.

scripts/check.unused.mjs
AntonioVentilii-DFINITY: right!

scripts/check.unused.mjs
```diff
@@ -0,0 +1,58 @@
...
+import { basename, resolve } from 'node:path';
+
+const RED = '\x1b[31m';
+const GREEN = '\x1b[32m';
+const NC = '\x1b[0m'; // No Colour
+
+const DATA_DIR = 'src/frontend/src';
+const DATA_DIR_PATH = resolve(process.cwd(), DATA_DIR);
+
+const findSvelteFiles = (dir) =>
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: on it!

scripts/check.unused.mjs
```diff
@@ -0,0 +1,58 @@
...
+const searchForFileName = ({ filename, dir }) =>
+	readdirSync(dir, { withFileTypes: true }).some((file) => {
+		const res = resolve(dir, file.name);
+		return file.isDirectory()
+			? searchForFileName({ filename, dir: res })
+			: file.isFile() && readFileSync(res, 'utf-8').includes(filename);
+	});
+
+const checkFileUsage = ({ filename, dir }) =>
+	filename.startsWith('+') || searchForFileName({ filename, dir });
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: done!

scripts/check.unused.mjs
AntonioVentilii-DFINITY: you mean store the content of each file, right? We would visit them just once and store the content (separately or all together). I think it would be just a memory "issue" at that point, no?

peterpeterparker: No I mean something like pseud code:

```
const listOfAllSvelteFiles = findSvelteFiles(); // [src/yolo.svelte, src/+page.svelte]

cont listOfPotentialUnusedFiles = [...listOfAllSvelteFiles]);

for (const file of listOfAllSvelteFiles) {
   const content = readFile(file);
   
   const usedFiles = listOfPotentialUnusedFiles.filter(file => content.include(file));
   
   listOfPotentialUnusedFiles = [...listOfPotentialUnusedFiles.filter(file => !usedFiles.includes(file))];
   
   if (listOfPotentialUnusedFiles.length === 0) {
      process.exit(0); // No used files
  }
}

listOfPotentialUnusedFiles.forEach(file => console.log("Unused file:", file));
```

While we have to iterate over an array, I was just thinking that that way, even if naive, maybe there is a chance that we would have to read less content and therefore less file system access which potentially is slower than filtering the array?

Just an idea of course, that might be less performant.

AntonioVentilii-DFINITY: Ah! Right! Now I understand! It makes sense indeed! Check now please

scripts/check.unused.mjs
```diff
@@ -0,0 +1,39 @@
...
+
+	console.log('\n');
+
+	if (unusedFiles.length === 0) {
+		console.log(`${GREEN}No unused components found.${NC}`);
+		process.exit(1);
+	} else {
+		console.log(`${RED}Found ${unusedFiles.length} unused component(s).${NC}`);
+		unusedFiles.forEach((file) => console.log(`${RED}Unused Svelte file: ${file}${NC}`));
+	}
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: ditto

scripts/check.unused.mjs
```diff
@@ -0,0 +1,39 @@
...
+	const unusedFiles = allSvelteFiles.reduce((acc, file) => {
+		const content = readFileSync(file, 'utf-8');
+		return acc.filter((f) => !content.includes(basename(f)));
+	}, potentialUnusedFiles);
+
+	console.log('\n');
+
+	if (unusedFiles.length === 0) {
+		console.log(`${GREEN}No unused components found.${NC}`);
+		process.exit(1);
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Ah right! it is inverted!

scripts/check.unused.mjs
```diff
@@ -0,0 +1,39 @@
...
+const findSvelteFiles = (dir) => findFiles({ dir, extensions: ['.svelte'] });
+
+const main = async () => {
+	console.log(`${NC}Scanning ${DATA_DIR} folder to find all .svelte files`);
+
+	const allSvelteFiles = findSvelteFiles(DATA_DIR_PATH);
+
+	const potentialUnusedFiles = allSvelteFiles.filter((file) => !dirname(file).includes('routes'));
+
+	const unusedFiles = allSvelteFiles.reduce((acc, file) => {
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Yes, you are right: I was trying to leave it more functional, but this way it can be faster indeed.

However, I still leave the check in the end too, since it should consider the improbable situation where we arrive there with empty `unusedFiles`: for example, if `allSvelteFiles` is empty (again, it is very improbable, but for consistency...)

scripts/check.unused.mjs
```diff
@@ -0,0 +1,45 @@
...
+		console.log(`${GREEN}No unused components found.${NC}`);
+		process.exit(0);
+	} else {
+		console.log(`${RED}Found ${potentialUnusedFiles.length} unused component(s).${NC}`);
+		potentialUnusedFiles.forEach((file) => console.log(`${RED}Unused Svelte file: ${file}${NC}`));
+		process.exit(1);
+	}
+};
+
+main().catch(console.error);
```
peterpeterparker, line 45: `process.exit(1)` in case of error?

peterpeterparker (commented):
LGTM, thanks

In another PR should we add it to `frontend-checks.yml` or another yml ci job?

scripts/check.unused.mjs
AntonioVentilii-DFINITY: adjusted

AntonioVentilii-DFINITY:
> In another PR should we add it to frontend-checks.yml or another yml ci job?

That means that we must remove a deprecated element in the same PR where we end up not using it anymore. Is that ok? An alternative would be creating a job that runs every saturday/sunday night that auto-removes deprecated components?

peterpeterparker:
Good point! A weekly job sounds cool 👍

peterpeterparker (approved):

AntonioVentilii-DFINITY:
Done with PR #1872

PR #1844, opened 2024-07-23
feat(ci): script to check unused i18n keys

AntonioVentilii-DFINITY:
# Motivation

New script to check if there are unused keys from the `i18n` file.

For dynamic usage it just checks if each part of the key is in the file AND if somewhere in the file `get(i18n)` is used.
For example:

```typescript
const { auth: authI18n } = get(i18n);
...
toastsError({
  msg: { text: authI18n.error.no_pouh_credential }
});
```

or

```typescript
const {
  init: {
    error: { eth_address_unknown, loading_balance }
  }
} = get(i18n);
```


scripts/check.i18n.mjs
```diff
@@ -0,0 +1,48 @@
...
+const extractKeys = (obj, prefix = '') =>
+	Object.keys(obj).reduce((res, el) => {
+		if (typeof obj[el] === 'object') {
+			return [...res, ...extractKeys(obj[el], `${prefix}${el}.`)];
+		}
+		return [...res, `${prefix}${el}`];
+	}, []);
+
+// It checks if the key is used in the content or if the key is used in a dynamic way
+const checkKeyUsage = ({ key, files }) =>
```
peterpeterparker, line 20: We can maybe implement the same optimization as those we implemented in `check.unused.mjs` ? Keep a list of potential unused keys and remove those that matching and exist the loop as soon as there are no more entries?

AntonioVentilii-DFINITY: Good one! Done!

peterpeterparker (approved):
LGTM, thx

PR #1847, opened 2024-07-24
feat(frontend): Add service to load user profile

lmuntaner:
# Motivation

We want to use the profile stored in the backend in the frontend.

In this PR, a new service is introduced `loadUserProfile` which loads gets or creates a profile and stores it in the store. It also performs an update call to store the certified data in the background.

# Changes

* New api function `getUserProfile` and `createUserProfile`.
* New error `UserProfileNotFoundError`.
* New service `loadUserProfile`.

# Tests

* Unit tests for `loadUserProfile`.
* I also tested triggering the service when the user visited settings and it worked as expected.

### DIFF
--- a/src/frontend/src/lib/api/backend.api.ts
+++ b/src/frontend/src/lib/api/backend.api.ts
@@ -1,12 +1,12 @@
 import type {
 	CustomToken,
+	GetUserProfileError,
 	SignRequest,
 	UserProfile,
 	UserToken
 } from '$declarations/backend/backend.did';
 import { getBackendActor } from '$lib/actors/actors.ic';
 import type { ECDSA_PUBLIC_KEY } from '$lib/types/address';
-import { UserProfileNotFoundError } from '$lib/types/errors';
 import type { OptionIdentity } from '$lib/types/identity';
 import type { Identity } from '@dfinity/agent';
 import type { QueryParams } from '@dfinity/utils';
@@ -121,15 +121,9 @@ export const createUserProfile = async ({
 export const getUserProfile = async ({
 	identity,
 	certified = true
-}: { identity: Identity } & QueryParams): Promise<UserProfile> => {
+}: { identity: Identity } & QueryParams): Promise<
+	{ Ok: UserProfile } | { Err: GetUserProfileError }
+> => {
 	const { get_user_profile } = await getBackendActor({ identity, certified });
-	const response = await get_user_profile();
-	if ('Ok' in response) {
-		return response.Ok;
-	}
-	const err = response.Err;
-	if ('NotFound' in err) {
-		throw new UserProfileNotFoundError();
-	}
-	throw new Error('Unknown error');
+	return get_user_profile();
 };
--- a/src/frontend/src/lib/api/backend.api.ts
+++ b/src/frontend/src/lib/api/backend.api.ts
@@ -1,6 +1,12 @@
-import type { CustomToken, SignRequest, UserToken } from '$declarations/backend/backend.did';
+import type {
+	CustomToken,
+	SignRequest,
+	UserProfile,
+	UserToken
+} from '$declarations/backend/backend.did';
 import { getBackendActor } from '$lib/actors/actors.ic';
 import type { ECDSA_PUBLIC_KEY } from '$lib/types/address';
+import { UserProfileNotFoundError } from '$lib/types/errors';
 import type { OptionIdentity } from '$lib/types/identity';
 import type { Identity } from '@dfinity/agent';
 import type { QueryParams } from '@dfinity/utils';
@@ -101,4 +107,29 @@ export const setUserToken = async ({
 }): Promise<void> => {
 	const { set_user_token } = await getBackendActor({ identity });
 	return set_user_token(token);
+};
+
+export const createUserProfile = async ({
+	identity
+}: {
+	identity: Identity;
+}): Promise<UserProfile> => {
+	const { create_user_profile } = await getBackendActor({ identity });
+	return create_user_profile();
+};
+
+export const getUserProfile = async ({
+	identity,
+	certified = true
+}: { identity: Identity } & QueryParams): Promise<UserProfile> => {
+	const { get_user_profile } = await getBackendActor({ identity, certified });
+	const response = await get_user_profile();
+	if ('Ok' in response) {
+		return response.Ok;
+	}
+	const err = response.Err;
+	if ('NotFound' in err) {
+		throw new UserProfileNotFoundError();
+	}
+	throw new Error('Unknown error');
 };
--- a/src/frontend/src/lib/services/load-user-profile.services.ts
+++ b/src/frontend/src/lib/services/load-user-profile.services.ts
@@ -56,12 +56,14 @@ export const loadUserProfile = async ({ identity }: { identity: Identity }): Pro
 		let profile = await queryUnsaveProfile({ identity });
 		if (isNullish(profile)) {
 			profile = await createUserProfile({ identity });
+			userProfileStore.set({ key: 'user-profile', value: profile });
 		} else {
+			// We set the store before the call to load the certified profile.
+			userProfileStore.set({ key: 'user-profile', value: profile });
 			// We don't wait for this to complete because we want a smooth UX
 			// the uncertified data is enough for the user to start using the app.
 			loadCertifiedUserProfile({ identity });
 		}
-		userProfileStore.set({ key: 'user-profile', value: profile });
 	} catch (err: unknown) {
 		const { settings } = get(i18n);
 		toastsError({
--- a/src/frontend/src/lib/services/load-user-profile.services.ts
+++ b/src/frontend/src/lib/services/load-user-profile.services.ts
@@ -8,13 +8,31 @@ import type { Identity } from '@dfinity/agent';
 import { isNullish } from '@dfinity/utils';
 import { get } from 'svelte/store';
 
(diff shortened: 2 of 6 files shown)


lmuntaner:

src/frontend/src/lib/api/backend.api.ts
```diff
@@ -102,3 +108,28 @@ export const setUserToken = async ({
...
+	const { create_user_profile } = await getBackendActor({ identity });
+	return create_user_profile();
+};
+
+export const getUserProfile = async ({
+	identity,
+	certified = true
+}: { identity: Identity } & QueryParams): Promise<UserProfile> => {
+	const { get_user_profile } = await getBackendActor({ identity, certified });
+	const response = await get_user_profile();
```
peterpeterparker, line 126: I would maybe implement interpreting the response in a service and keep the function "dumb"? The api is basically just a wrapper proxy.

peterpeterparker: Plus it allows to internationalize error message and avoid hardcoding "Unknown error" text.

src/frontend/src/lib/services/load-user-profile.services.ts
```diff
@@ -0,0 +1,54 @@
...
+import { createUserProfile, getUserProfile } from '$lib/api/backend.api';
+import { i18n } from '$lib/stores/i18n.store';
+import { userProfileStore } from '$lib/stores/settings.store';
+import { toastsError } from '$lib/stores/toasts.store';
+import { UserProfileNotFoundError } from '$lib/types/errors';
+import type { Identity } from '@dfinity/agent';
+import { isNullish } from '@dfinity/utils';
+import { get } from 'svelte/store';
+
+const getUncertifiedProfile = async ({
```
peterpeterparker, line 11: can we use the prefix `unsafe` instead of `uncertified`?

src/frontend/src/lib/services/load-user-profile.services.ts
```diff
@@ -0,0 +1,54 @@
...
+	} catch (err) {
+		// We ignore the error because this is a background task.
+		console.error('Failed to load certified user profile.', err);
+	}
+};
+
+export const loadUserProfile = async ({ identity }: { identity: Identity }): Promise<void> => {
+	try {
+		let profile = await getUncertifiedProfile({ identity });
+		if (isNullish(profile)) {
```
peterpeterparker, line 39: I would appreciate if you ping prodsec about this logic - i.e. if answer to query is undefined then create.

I'm sure it's fine but, just for the state of the art as you never know.

src/frontend/src/lib/services/load-user-profile.services.ts
```diff
@@ -0,0 +1,54 @@
...
+			profile = await createUserProfile({ identity });
+		} else {
+			// We don't wait for this to complete because we want a smooth UX
+			// the uncertified data is enough for the user to start using the app.
+			loadCertifiedUserProfile({ identity });
```
peterpeterparker, line 44: Two things come to my mind:

1. If the profile does not need to be created, I would already set the profile in the store as uncertified and not block the user. Then load with an update, and if the update fails, log out the user with an error message.

Otherwise, each login might end up being slower by a few seconds.

Or do you plan to not load the user profile on sign-in but, only when required?

2. Do we want to cache the profile on the client side as we do for the ETH address? If the profile is cached in IndexedDB (idb), we skip the query, skip the check to create it, and just do an update call in the background, which also signs out the user if unsuccessful.

src/frontend/src/lib/services/load-user-profile.services.ts
```diff
@@ -0,0 +1,54 @@
...
+		}
+		userProfileStore.set({ key: 'user-profile', value: profile });
+	} catch (err) {
+		const { settings } = get(i18n);
+		toastsError({
+			msg: { text: settings.error.loading_profile }
+		});
+		console.error('Failed to load user profile.', err);
```
peterpeterparker, line 52: Can be removed I would say. Instead, pass the `err` to `toastsError`?

src/frontend/src/lib/services/load-user-profile.services.ts
```diff
@@ -0,0 +1,54 @@
...
+		let profile = await getUncertifiedProfile({ identity });
+		if (isNullish(profile)) {
+			profile = await createUserProfile({ identity });
+		} else {
+			// We don't wait for this to complete because we want a smooth UX
+			// the uncertified data is enough for the user to start using the app.
+			loadCertifiedUserProfile({ identity });
+		}
+		userProfileStore.set({ key: 'user-profile', value: profile });
+	} catch (err) {
```
peterpeterparker, line 47: `err: unknown`

src/frontend/src/lib/types/errors.ts
```diff
@@ -0,0 +1,5 @@
+export class UserProfileNotFoundError extends Error {
+	constructor() {
+		super('User profile not found');
```
peterpeterparker, line 3: If it needs to be i18n, I would remove the error msg here

src/frontend/src/lib/services/load-user-profile.services.ts
lmuntaner: Done https://dfinity.slack.com/archives/C01EWN833KN/p1721824932805649

src/frontend/src/lib/services/load-user-profile.services.ts
lmuntaner: > If the profile does not need to be created, I would already set the profile in the store as uncertified and not block the user. Then load with an update

This is what the code does, isn't it?

I asked prodsec about caching the profile.

Moreover, I need to change the current caching because it caches all user profiles with the same key and we should add the principal in the key so that different users can be cached in the same device. This will come in another PR.

Yet, I don't see why we should follow a different path if we have the cache data.

Isn't the unsafe but updated data better than waiting for an update call?

If there is cached data, it means that getting the profile will work (or fail in other ways than NotFound), and it will skip creating. No need to specify it in the code.

src/frontend/src/lib/api/backend.api.ts
lmuntaner: Done! I wasn't sure either which pattern you preferred.

src/frontend/src/lib/services/load-user-profile.services.ts
lmuntaner: Done

src/frontend/src/lib/services/load-user-profile.services.ts
lmuntaner: done

src/frontend/src/lib/services/load-user-profile.services.ts
lmuntaner: done

lmuntaner:

lmuntaner:

src/frontend/src/lib/services/load-user-profile.services.ts
peterpeterparker: We can go on with the PR and iterate after their review - i.e. given that we asked I think it's fine to continue with the PR. Of course we should wait on their feedback before going live but, it should not block us.

lmuntaner: This is not yet used, so it's not even blocking us from deploying `main`.

src/frontend/src/lib/services/load-user-profile.services.ts
peterpeterparker: > This is what the code does, isn't it?
> ....
> Isn't the unsafe but updated data better than waiting for an update call?

Right; Sorry, I missed your comment ("We don't wait for this to complete...") in previous review.

Nevertheless, should we perhaps move the `userProfileStore.set` call before `loadCertifiedUserProfile`? Even if it's 99.999999% unlikely, code wise, there is maybe a potential race condition where the profile could first be set by the `update` in the store and then overwritten by the `query`?

src/frontend/src/lib/services/load-user-profile.services.ts
```diff
@@ -0,0 +1,72 @@
...
+		return response.Ok;
+	}
+	const err = response.Err;
+	if ('NotFound' in err) {
+		throw new UserProfileNotFoundError();
+	}
+	throw new Error('Unknown error');
+};
+
+const queryUnsaveProfile = async ({
```
peterpeterparker, line 29: typo? `unsave`

lmuntaner: No... I expect `unsave` to be a word and this is camelcase.

peterpeterparker: Ah. Well, I've got no idea what `unsave` means in this context? I understand it as "query the list of profile that have been reverted from being saved (unsaved)"

lmuntaner: ah, sorry, hahaha, it should be Unsafe, the problem is the "f", I thought you were saying that it should be a lower case "u", not "U" 😂

lmuntaner:

src/frontend/src/lib/services/load-user-profile.services.ts
peterpeterparker: hahaha glad my english is not that bad 😅

peterpeterparker (approved):
LGTM, thanks

PR #1849, opened 2024-07-24
feat(frontend): support for II alternative origins

peterpeterparker:
# Motivation

Andronikos want the II authentication of `staging.oisy.com` to be derived with `tewsx-xaaaa-aaaad-aadia-cai.icp0.io`.

Therefore, this PR adds support for [II alternative origins](https://internetcomputer.org/docs/current/developer-docs/identity/internet-identity/alternative-origins/)

# Changes

- Add a script to create a static `.well-known/ii-alternative-origins` at build time
- Extend authentication to support derivation origin
- Add optional environment variables


AntonioVentilii-DFINITY (approved):
LGTM!

Note: shouldn't we update the `.env.example` too?

peterpeterparker:
> shouldn't we update the .env.example too?

I would say no. It's a special use case related to Oisy itself.

PR #1850, opened 2024-07-24
chore(frontend): remove CSS class 'ic' that is deprecated

AntonioVentilii-DFINITY:
# Motivation

The CSS class `ic` is deprecated.


peterpeterparker (approved):
Thx!

PR #1851, opened 2024-07-24
feat(devops): Set expected POUH issuer id

lmuntaner:
# Motivation

We want to be able to validate the credentials semantically (check the data is the expected one). One of the fields is `iss`, which is the issuer id (which a url is used most of the times).

In this PR, I change the hardcoded `POUH_ISSUER_VC_URL` and set it depending on the environment.

The values were confirmed by Decide AI team.

# Changes

* Change `deploy.backend.sh` to set `POUH_ISSUER_VC_URL` per environment.

# Tests

I tested deploying it locally and it was working.


lmuntaner:

peterpeterparker (approved):
LGTM, thanks

PR #1854, opened 2024-07-24
chore(frontend): Fix test files with VS Code

lmuntaner:
# Motivation

Have IDE help in test files. At the moment, VS Code couldn't understand the test files with `describe`, `it`, ... And it wasn't auto completing or auto importing.

The fix was to add a tsconfig in the test folder pointing to the expected config.

# Changes

* Add file `tsconfig.json` in `src/frontend/src/tests`.

# Tests

My VS Code is happy again 😄


lmuntaner:
@AntonioVentilii-DFINITY @peterpeterparker I guess that you don't use VS Code, but just in case a developer wants to use VS Code it would be good to merge this. Otherwise, I'm also happy having it in my local directory.

PR #1855, opened 2024-07-24
feat(frontend): auto load erc20 user token if not enabled

peterpeterparker:
# Motivation

When a user converts an ERC20 token to a ckERC20 twin token, the UI needs information about the counterpart token (ERC20). For example, the user might have a CK token (e.g., ckUSDC) enabled, but its token counterpart (e.g., USDC) disabled.

As a result, the UI might lack necessary information. Notably, the balance displayed on the "How to convert" modal wizard might remain zero even if the user has effectively transferred tokens to Oisy (this is a bug that was reported).

Consequently, the user waits indefinitely and cannot proceed with the conversion unless they understand what's happening and finally figure out they should manually add the ERC20 token.

That's why this PR aims to automatically enable ERC20 tokens if required.

# Source

Reported per PM and shared on [Slack](https://dfinity.slack.com/archives/C05M7P123HR/p1721277306898159).

# Changes

- New service function `autoLoadUserToken`.
- On `openCkETHReceive` proceed by calling the function first and if required enable automatically the token.
- Refactor `setUserToken` which is now called manually and automatically.

# Screenshot

"Wait for ..." remains zero for ever:

image


AntonioVentilii-DFINITY (approved):
LGTM!

Separate topic: does the `button` component have a feature like `loading` if the `on:click` function is not finished?

AntonioVentilii-DFINITY:
No, no, nothing, I was thinking wrongly

PR #1863, opened 2024-07-25
chore(frontend): set i18n for TokenCardWithUrl aria label

AntonioVentilii-DFINITY:
# Motivation

The aria label for `TokenCardWithUrl` now uses `i18n`.


peterpeterparker (approved):
Thanks!

PR #1865, opened 2024-07-25
feat(backend): Add issuer to UserCredentials

lmuntaner:
# Motivation

It's useful to remember who have the credentials in case the same credential type has multiple issuers.

# Changes

* Add field `issuer` to `UserCredentials`.

# Tests

* Tests cover new field because they check all fields.


anedos-dfinity:
Good idea!

src/shared/src/types.rs
```diff
@@ -132,6 +132,7 @@ pub mod user_profile {
     pub struct UserCredential {
         pub credential_type: CredentialType,
         pub verified_date_timestamp: Option<Timestamp>,
+        pub issuer: String,
```
peterpeterparker, line 135: Shouldn't this information be a variant?

lmuntaner: This comes from the config, the `issuer_origin` of `SupportedCredential`. There it's a String, and I believe that's ok.

I don't think we need to create each issuer into an entry in an enum.

peterpeterparker: Analytics might need manual work, and we won't be able to accurately use the information in the frontend for a feature. Those were my spontaneous thoughts for my question, but okay.

peterpeterparker (approved):
LGTM, thx

PR #1866, opened 2024-07-25
feat(frontend): add USDT token and link it to ckUSDT

AntonioVentilii-DFINITY:
# Motivation

We add USDT as token in the env folders, and we link it to the new ckUSDT token. Furthermore, we remove the "temporary" USDT token that was there.


src/frontend/src/eth/utils/erc20.utils.ts
```diff
@@ -64,8 +63,6 @@ const mapErc20Icon = (symbol: string): string | undefined => {
 			return rndr;
 		case 'uni':
 			return uniswap;
-		case 'usdt':
```
peterpeterparker, line 67: redo?

AntonioVentilii-DFINITY: we don't use that one for USDT anymore, correct? now the icon is included in the `env` file

peterpeterparker: Ah yeah! thanks

peterpeterparker (approved):
LGTM, thanks

PR #1867, opened 2024-07-25
refactor(frontend): extract sub-component from IcSendModal

AntonioVentilii-DFINITY:
# Motivation

Similar to how it is done for ETH, we extract the sub-component inside the `IcSendModal`. The final scope is to re-use it elsewhere too.

# Changes

- Create new component `IcSendTokenWizard`.
- Adjust component `IcSendModal`.

# Tests

Local test of sending to myself some ICP worked as expected.


peterpeterparker (approved):
LGTm, thx

PR #1868, opened 2024-07-25
refactor(frontend): use Img component in modals About

AntonioVentilii-DFINITY:
# Motivation

We use the dedicated `Img` component in the new `About` modals.


peterpeterparker (approved):
LGTm, thx

PR #1869, opened 2024-07-25
fix(frontend): back

peterpeterparker:
# Motivation

There was a bug so that the `Back` component would always go in loop between the last two navigational page (transactions and settings).

# Changes

- Rename `Back to wallet` to `Back`
- Use native navigation history and goto to execute browser's back on, back
- Replace specific call to function back that were actually meant as navigation with a new function that navigate and replace the history with the root


AntonioVentilii-DFINITY (approved):
Thank you! Better!

PR #1871, opened 2024-07-25
feat(frontend): Verify and add credential to user profile after presentation

lmuntaner:
# Motivation

We want to validate the credentials presented by users and store some metadata in their profiles.

In this PR, the service `requestPouhCredential` is connected to the backend.

# Changes

* New api function `addUserCredential`.
* Load the user profile in settings page and render the credentials only when present.
* Change the parameter of `requestPouhCredential` to the `Identity`.
* Add two new i18n messages.
* Export `loadCertifiedUserProfile`.
* Add a helper `addPouhCredential` to be used within the handle success of `requestPouhCredential`.

# Tests

See screencast of the successful flow.

### DIFF
--- a/src/frontend/src/lib/components/pages/SettingsSignedIn.svelte
+++ b/src/frontend/src/lib/components/pages/SettingsSignedIn.svelte
@@ -1,9 +1,9 @@
 <script lang="ts">
 	import { shortenWithMiddleEllipsis } from '$lib/utils/format.utils';
-	import { KeyValuePairInfo } from '@dfinity/gix-components';
+	import { KeyValuePairInfo, Spinner } from '@dfinity/gix-components';
 	import Copy from '$lib/components/ui/Copy.svelte';
 	import { authRemainingTimeStore, authStore } from '$lib/stores/auth.store';
-	import { nonNullish } from '@dfinity/utils';
+	import { isNullish, nonNullish } from '@dfinity/utils';
 	import { secondsToDuration } from '@dfinity/utils';
 	import type { Principal } from '@dfinity/principal';
 	import NetworksTestnetsToggle from '$lib/components/networks/NetworksTestnetsToggle.svelte';
@@ -15,23 +15,35 @@
 	import { fade } from 'svelte/transition';
 	import { busy } from '$lib/stores/busy.store';
 	import { POUH_ENABLED } from '$lib/constants/credentials.constants';
+	import type { OptionIdentity } from '$lib/types/identity';
+	import { loadUserProfile } from '$lib/services/load-user-profile.services';
+	import { userProfileStore } from '$lib/stores/settings.store';
 
 	let remainingTimeMilliseconds: number | undefined;
 	$: remainingTimeMilliseconds = $authRemainingTimeStore;
 
+	let identity: OptionIdentity;
+	$: identity = $authStore?.identity;
+
 	let principal: Principal | undefined | null;
-	$: principal = $authStore?.identity?.getPrincipal();
+	$: principal = identity?.getPrincipal();
 
 	const getPouhCredential = async () => {
-		if (nonNullish(principal)) {
+		if (nonNullish(identity)) {
 			try {
 				busy.show();
-				await requestPouhCredential({ credentialSubject: principal });
+				await requestPouhCredential({ identity });
 			} finally {
 				busy.stop();
 			}
 		}
 	};
+
+	$: {
+		if (nonNullish(identity) && POUH_ENABLED) {
+			loadUserProfile({ identity });
+		}
+	}
 </script>
 
 <KeyValuePairInfo>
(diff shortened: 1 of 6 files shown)


lmuntaner:
New UX when presenting credentials:

https://github.com/user-attachments/assets/4585193b-343a-45f5-a7bf-e7d6fd7733b9

New UX when user arrives to settings:

https://github.com/user-attachments/assets/098fd7ce-224a-4151-b533-0ca4c7e32df1

lmuntaner:

AntonioVentilii-DFINITY (approved):
LGTM!

src/frontend/src/lib/components/pages/SettingsSignedIn.svelte
```diff
@@ -15,23 +15,35 @@
...
-				await requestPouhCredential({ credentialSubject: principal });
+				await requestPouhCredential({ identity });
 			} finally {
 				busy.stop();
 			}
 		}
 	};
+
+	$: {
+		if (nonNullish(identity) && POUH_ENABLED) {
```
peterpeterparker, line 43: `authSignedIn` instead of `nonNullish(identity)` probably

src/frontend/src/lib/components/pages/SettingsSignedIn.svelte
```diff
@@ -97,21 +109,20 @@
 	</KeyValuePairInfo>
 </div>
 
-{#if POUH_ENABLED}
-	<div class="mt-8">
+{#if POUH_ENABLED && nonNullish($userProfileStore)}
```
peterpeterparker, line 112: should we created a derived store instead of `nonNullish($userProfileStore)`?

src/frontend/src/lib/components/pages/SettingsSignedIn.svelte
```diff
@@ -97,21 +109,20 @@
...
 		<h2 class="text-base mb-6 pb-1">{$i18n.settings.text.credentials_title}</h2>
 
 		<div class="mt-4">
 			<KeyValuePairInfo>
 				<span slot="key" class="font-bold">{$i18n.settings.text.pouh_credential}:</span>
 				<svelte:fragment slot="value">
-					<!-- TODO: Render spinner if no user profile -->
 					{#if $userHasPouhCredential}
-						<output in:fade class="mr-1.5">
+						<output class="mr-1.5">
```
peterpeterparker, line 121: the fade is still useful here I think, it's should probably not be removed as it is meant to smoothness the animation when credentials are requested for the first time

src/frontend/src/lib/i18n/en.json
```diff
@@ -48,6 +48,8 @@
 		},
 		"error": {
 			"no_internet_identity": "No internet identity.",
+			"invalid_pouh_credential": "Sorry, but there was an error while validating the Humanity Credential. Please contact the issuer and request the credential again.",
```
peterpeterparker, line 51: curiosity question: "Humanity Credential" is a name therefore the capital letters?

src/frontend/src/lib/services/request-pouh-credential.services.ts
```diff
@@ -11,31 +11,79 @@ import { i18n } from '$lib/stores/i18n.store';
...
+		}
+		if ('Err' in response) {
+			if ('InvalidCredential' in response.Err) {
+				toastsError({
+					msg: { text: authI18n.error.invalid_pouh_credential }
+				});
+				return { success: false };
+			}
+			// Throw so that it gets handled by the catch block.
+			throw new Error();
```
peterpeterparker, line 57: the linter or the editor does not display a hint here? something like throwing within a try catch is not recommended. Can you return an error?

src/frontend/src/lib/components/pages/SettingsSignedIn.svelte
lmuntaner: I'd still need `nonNullish(identity)` because TS doesn't know that when `authSignedIn` is true, `identity` is defined.

src/frontend/src/lib/components/pages/SettingsSignedIn.svelte
lmuntaner: Maybe, but let's discuss after the refactor of the user profile store on Monday.

src/frontend/src/lib/i18n/en.json
lmuntaner: That's how I see it so far. In the settings, the key is called "Humanity Credential", and this messages refer to that line in settings.

src/frontend/src/lib/services/request-pouh-credential.services.ts
lmuntaner: I want the same behavior as inside the `catch` block.

There are three options:

* Move the code of the `catch` after the `try/catch` block and do nothing here nor inside the `catch` block.
* Duplicated the toast error here instead of throwing.
* Throw an error.

Which one do you prefer?

src/frontend/src/lib/components/pages/SettingsSignedIn.svelte
peterpeterparker: You can pass a `OptionIdentity` all along and the initialization of the actor ultimately throws an error.

src/frontend/src/lib/i18n/en.json
peterpeterparker: Good to know.

src/frontend/src/lib/services/request-pouh-credential.services.ts
peterpeterparker: I would have two different errors since they are distinct. I would create a specific label for the first toast to make it obvious that it throws an error because the response type is not supported.

```
try {
     toastsError({
			msg: { text: authI18n.error.a_new_label_for_response_type_unsupported },
		});

} catch (....
```

PR #1872, opened 2024-07-25
feat(ci): create weekly job to remove unused components

AntonioVentilii-DFINITY:
# Motivation

Weekly job to run script to remove unused `.svelte` files.


scripts/check.unused.mjs
```diff
@@ -36,8 +38,13 @@ const main = async () => {
...
 		process.exit(0);
 	} else {
-		console.log(`${RED}Found ${potentialUnusedFiles.length} unused component(s).${NC}`);
-		potentialUnusedFiles.forEach((file) => console.log(`${RED}Unused Svelte file: ${file}${NC}`));
+		console.log(
+			`${RED}Found ${potentialUnusedFiles.length} unused component(s)${REMOVE_FILES ? ' that will be deleted' : ''}.${NC}`
+		);
+		potentialUnusedFiles.forEach((file) => {
+			console.log(`${RED}Unused Svelte file: ${file}${NC}`);
+			if (REMOVE_FILES) unlinkSync(file);
```
peterpeterparker, line 46: I'm opiniated, would you mind adding `{}` for the if?

```
if (REMOVE_FILES) {
  unlinkSync(file);
}
```

Don't know if it was in article or just in a tweets but, I remember reading some developer advice about always specifing brackets for if, I think they say it was better for maintanability reasons of large code base, so started using that pattern.

.github/workflows/frontend-remove-unused-components.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Prepare
+        uses: ./.github/actions/prepare
+
+      - name: Remove unused files
+        run: ./scripts/check.unused.sh --remove-files
+        continue-on-error: true
```
peterpeterparker, line 20: Is that, `continue-on-error: true` correct?

peterpeterparker (approved):
LGTM, thanks

Minor two questions.

scripts/check.unused.mjs
AntonioVentilii-DFINITY: ah yes, for sure will do

I saw that we do it somehwere else in the code, maybe I will check if there is a lint rule, for another PR

.github/workflows/frontend-remove-unused-components.yml
AntonioVentilii-DFINITY: This is due to the fact that I return `process.exit(1)` in case of unused files, to continue the workflow. We can remove that return code (or return it only when the files are not removed).

peterpeterparker: AAAAAH gotcha! Shouldn't it return `process.exit(0)` if there is no unused files? If I said the contrary in a previous review, sorry for that.

scripts/check.unused.mjs
AntonioVentilii-DFINITY: Done with PR #1876

.github/workflows/frontend-remove-unused-components.yml
AntonioVentilii-DFINITY: what do you think if I add:

```javascript
} else {
  console.log(...);
  potentialUnusedFiles.forEach((file) => {
    console.log(`${RED}Unused Svelte file: ${file}${NC}`);
    if (REMOVE_FILES) {
      unlinkSync(file);
    }
  });
  
  if (REMOVE_FILES) {
    await main();
  }
  
  process.exit(1);
}
```

it will loop until all unused files are removed (maybe caused by the removal of another one), AND it will exit with 0, so we don't need the `continue-on-error`

peterpeterparker: what about something like following:

```
const print = async (file) => console.log(`${RED}Unused Svelte file: ${file}${NC}`);
const remove = async (file) => {
    await print(file);
    unlinkSync(file);
    console.log(`${GREEN}Removed: ${file}${NC}`
}
		
const fn = REMOVE_FILES ? remove : print;
		
await Promise.allSettled(potentialUnusedFiles.map(fn));
```

AntonioVentilii-DFINITY: on it!

But ultimately, I remove the `process.exit(1)` from there, right?

peterpeterparker: I lost track. Let's just make it works ;)

PR #1873, opened 2024-07-25
chore(frontend): move USDC logo to eth folder

AntonioVentilii-DFINITY:
# Motivation

For consistency, we move the USDC logo to the `eth` folder, since it is used only for the ERC20 token (ckUSDC token has a different logo).


peterpeterparker (approved):
Thanks!

PR #1876, opened 2024-07-25
feat(ci): add lint rule 'curly'

AntonioVentilii-DFINITY:
# Motivation

We add the lint rule [curly](https://eslint.org/docs/latest/rules/curly) to force the usage of `{}` after an `if` statement. Furthermore, we fix the parts of the code that would raise an issue.


peterpeterparker (approved):
Nice, thanks!

PR #1879, opened 2024-07-25
refactor(frontend): add 'back' option to IcSendForm

AntonioVentilii-DFINITY:
# Motivation

Since we would like to re-use `IcSendTokenWizard` as an intermediate step in a wizard modal, we add the possibility of having `Back` button instead of just `Close`.

Note: it is similar to what we have in `EthSendTokenWizard`.


peterpeterparker (approved):
LGTM, thx

PR #1882, opened 2024-07-26
feat(frontend): Improvements to present credential flow

lmuntaner:
# Motivation

The main motivation is to address the post-merge comments in #1871 

# Changes

* Add fade effect to user data in profile.
* Use derived store to check if user is signed in.
* Use `OptionIdentity` instead of `Identity` in services and api layer.
* Show a different error if adding credential fails with Oisy error.

# Tests

Tested manually.


lmuntaner:

peterpeterparker (approved):
Thanks

PR #1883, opened 2024-07-26
chore(frontend): Remove Unused Svelte Files

sa-github-api:
The following files were removed because they were not used in the project anymore:
```
$(cat deleted_files.txt)
```

PR #1884, opened 2024-07-26
fix(ci): remove txt file and using env variables for deleted files

AntonioVentilii-DFINITY:
# Motivation

Having tested the workflow with PR #1883 , I noticed that the `.txt` file used to store the deleted files is included in the PR. Furthermore, the list is not printed in the PR body.

So we change it to use ENV variables instead.


peterpeterparker (approved):
LGTM; Thx

PR #1886, opened 2024-07-26
feat(frontend): create SendWizard as content for generic send modal

AntonioVentilii-DFINITY:
# Motivation

We merge `EthSendTokenWizard` and `IcSendTokenWizard` into a single component `SendWizard`, to be used as content for the future, general Send modal.

### DIFF
--- a/src/frontend/src/lib/components/send/SendWizard.svelte
+++ b/src/frontend/src/lib/components/send/SendWizard.svelte
@@ -1,0 +1,54 @@
+<script lang="ts">
+	import { type WizardStep } from '@dfinity/gix-components';
+	import { selectedEthereumNetwork } from '$eth/derived/network.derived';
+	import { ethereumToken } from '$eth/derived/token.derived';
+	import EthSendTokenWizard from '$eth/components/send/EthSendTokenWizard.svelte';
+	import type { Network, NetworkId } from '$lib/types/network';
+	import { token } from '$lib/stores/token.store';
+	import SendTokenContext from '$eth/components/send/SendTokenContext.svelte';
+	import IcSendTokenWizard from '$icp/components/send/IcSendTokenWizard.svelte';
+	import { isNetworkIdEthereum, isNetworkIdICP } from '$lib/utils/network.utils';
+
+	export let destination: string;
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/send/SendWizard.svelte
```diff
@@ -0,0 +1,54 @@
...
+	export let destination: string;
+	export let targetNetwork: Network | undefined;
+	export let networkId: NetworkId | undefined;
+	export let amount: number | undefined;
+	export let sendProgressStep: string;
+	export let currentStep: WizardStep | undefined;
+</script>
+
+{#if isNetworkIdEthereum($token?.network.id)}
+	<SendTokenContext token={$token}>
```
peterpeterparker, line 21: Can we move the context one level down? i.e. it would be nice to really just have a wizard and not other logic in this component.

AntonioVentilii-DFINITY: yes! but I think this is for another PR, since it is used in the same way in the ETH modal, no?

peterpeterparker: Can we have it clean from the start or that's a big change?

AntonioVentilii-DFINITY: for sure it should be in another PR: the context is set before to adjust the parent component of `EthSendTokenWizard` (that is `EthSendTokenModal`) to the modals of 
- `EthSendModal`
- `ConvertToCkETH`
- `ConvertToCkERC20`

In our use-case we use it in the same way of `EthSendModal`, but to put it down, we need to change a bit the way the `SEND-CONTEXT` is used.

What do you think?

peterpeterparker: Ok. Please add a TODO before merge.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/components/send/SendWizard.svelte
AntonioVentilii-DFINITY: oki!

PR #1877, opened 2024-07-25
feat(frontend): new send modal for all tokens

AntonioVentilii-DFINITY:
# Motivation

We include the possibility to send tokens from the main page. To do that, we first create a modal that "merge" both flows for IC and ETH tokens.
The modal will display only the enabled tokens that have non-null balance.

# Changes

- Create new component `SendTokensList` with the desired list of tokens:
   - PR #1887 
- Create `SendModal` with the usage of both IC and ETH send modals:
  - PR #1886 
  - PR #1907 
- Create `Send` button.

# Tests

I sent out ICP and ETH on local replica and it worked as expected, as usual.

PR #1887, opened 2024-07-26
feat(frontend): create tokens list for the generic Send modal

AntonioVentilii-DFINITY:
# Motivation

To have a general Send modal in the main page, we first require the tokens list component that we want to show as first step of the modal.

### DIFF
--- a/src/frontend/src/lib/components/send/SendTokensList.svelte
+++ b/src/frontend/src/lib/components/send/SendTokensList.svelte
@@ -1,0 +1,43 @@
+<script lang="ts">
+	import { BigNumber } from '@ethersproject/bignumber';
+	import Listener from '$lib/components/core/Listener.svelte';
+	import TokensSkeletons from '$lib/components/tokens/TokensSkeletons.svelte';
+	import { enabledNetworkTokensUi } from '$lib/derived/network-tokens.derived';
+	import type { TokenUi } from '$lib/types/token';
+	import { fade } from 'svelte/transition';
+	import { formatToken } from '$lib/utils/format.utils';
+	import CardAmount from '$lib/components/ui/CardAmount.svelte';
+	import ExchangeTokenValue from '$lib/components/exchange/ExchangeTokenValue.svelte';
+	import { balancesStore } from '$lib/stores/balances.store';
+	import { createEventDispatcher } from 'svelte';
+	import TokenCardWithOnClick from '$lib/components/tokens/TokenCardWithOnClick.svelte';
+
+	const dispatch = createEventDispatcher();
+
+	let tokens: TokenUi[];
+	$: tokens = $enabledNetworkTokensUi.filter(({ id: tokenId }) =>
+		($balancesStore?.[tokenId]?.data ?? BigNumber.from(0n)).gt(0n)
+	);
+</script>
+
+<TokensSkeletons>
+	{#each tokens as token (token.id)}
+		<Listener {token}>
+			<div in:fade>
+				<TokenCardWithOnClick {token} on:click={() => dispatch('icSendToken', token)}>
+					<output class="break-all" slot="description">
+						{formatToken({
+							value: $balancesStore?.[token.id]?.data ?? BigNumber.from(0n),
+							unitName: token.decimals
+						})}
+						{token.symbol}
+					</output>
+
+					<CardAmount slot="exchange">
+						<ExchangeTokenValue {token} />
+					</CardAmount>
+				</TokenCardWithOnClick>
+			</div>
+		</Listener>
+	{/each}
+</TokensSkeletons>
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/send/SendTokensList.svelte
```diff
@@ -0,0 +1,43 @@
...
+	import CardAmount from '$lib/components/ui/CardAmount.svelte';
+	import ExchangeTokenValue from '$lib/components/exchange/ExchangeTokenValue.svelte';
+	import { balancesStore } from '$lib/stores/balances.store';
+	import { createEventDispatcher } from 'svelte';
+	import TokenCardWithOnClick from '$lib/components/tokens/TokenCardWithOnClick.svelte';
+
+	const dispatch = createEventDispatcher();
+
+	let tokens: TokenUi[];
+	$: tokens = $enabledNetworkTokensUi.filter(({ id: tokenId }) =>
```
peterpeterparker, line 18: It duplicates code from `TokensSignedIn`, should we create a derived global store?

src/frontend/src/lib/components/send/SendTokensList.svelte
```diff
@@ -0,0 +1,43 @@
...
+		($balancesStore?.[tokenId]?.data ?? BigNumber.from(0n)).gt(0n)
+	);
+</script>
+
+<TokensSkeletons>
+	{#each tokens as token (token.id)}
+		<Listener {token}>
```
peterpeterparker, line 25: Why do we need the redeclare the `Listener` here?

src/frontend/src/lib/components/send/SendTokensList.svelte
```diff
@@ -0,0 +1,43 @@
...
+			<div in:fade>
```
peterpeterparker, line 26: Do we need `fade` ? There is no effect no?

src/frontend/src/lib/components/send/SendTokensList.svelte
```diff
@@ -0,0 +1,43 @@
...
+				<TokenCardWithOnClick {token} on:click={() => dispatch('icSendToken', token)}>
+					<output class="break-all" slot="description">
```
peterpeterparker, line 28: The content will be refactored in another PR right?

Just assuming that's the same content as in `TokensSignedIn`.

src/frontend/src/lib/components/send/SendTokensList.svelte
AntonioVentilii-DFINITY: this is just precautious in case the user clicks on the button so quickly that the tokens are still not loaded. Shall I remove it?

src/frontend/src/lib/components/send/SendTokensList.svelte
AntonioVentilii-DFINITY: yes! makes sense!

src/frontend/src/lib/components/send/SendTokensList.svelte
AntonioVentilii-DFINITY: I misunderstood the usage of `Listener`. You are right, it would just duplicate the Listener, since it will always be initiated together with the one in `TokensSignedIn`.

Removing it

src/frontend/src/lib/components/send/SendTokensList.svelte
AntonioVentilii-DFINITY: Yes, I prefer to do it in another. This one already has a new component

src/frontend/src/lib/components/send/SendTokensList.svelte
peterpeterparker: If we want to be consistent then yes we remove the fade and in addition we gonna have to use `waitWalletReady` when the modal is opened to wait until the tokens are not loaded.

src/frontend/src/lib/components/send/SendTokensList.svelte
AntonioVentilii-DFINITY: check last commit @peterpeterparker

src/frontend/src/lib/components/send/SendTokensList.svelte
AntonioVentilii-DFINITY: ah! but then, we should do this in `TokensSignedIn` too, I image? In another PR

AntonioVentilii-DFINITY: Ah! I understand now! Yes, I had already put it [there](https://github.com/dfinity/oisy-wallet/pull/1877/files#diff-3eaedffe3c4906283cba5482203936b32f7e6ff2997ab193db4956f4b3380410).

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/components/send/SendTokensList.svelte
AntonioVentilii-DFINITY: Done in:

- PR #1931

PR #1888, opened 2024-07-26
style(frontend): remove compact style when printing USD amounts

AntonioVentilii-DFINITY:
# Motivation

As requested the USD amounts will be printed in `#'###.00` format from now on, instead of the compact mode.


peterpeterparker (commented):
The `notation` support should also be removed from `formatUSD`.

peterpeterparker (approved):
LGTM, thx

PR #1889, opened 2024-07-26
refactor(frontend): remove confirmation button from welcome loader

AntonioVentilii-DFINITY:
# Motivation

We remove the confirmation button from the `Loader` component, and we set it to automatically proceed after a small delay, for visual feedback.

# Tests

In local replica worked as expected:

https://github.com/user-attachments/assets/ef6ee3ce-b31d-4e5a-98f8-57b0ec8a8078


src/frontend/src/lib/components/core/Loader.svelte
```diff
@@ -38,7 +38,7 @@
 			return;
 		}
 
-		if (confirm) {
+		if (isNullish(oisy_introduction)) {
```
peterpeterparker, line 41: Why not just removing this step?

src/frontend/src/lib/components/core/Loader.svelte
```diff
@@ -98,6 +94,9 @@
 		}
 
 		await progressAndLoad();
+
+		// Automatically confirm introduction after a delay for visual feedback.
```
peterpeterparker, line 98: See previous question - i.e. why adding this?

peterpeterparker (commented):
In addition, I think the function `confirmIntroduction` and the variable `oisy_introduction` can be removed.

src/frontend/src/lib/components/core/Loader.svelte
AntonioVentilii-DFINITY: done

AntonioVentilii-DFINITY:
I thought we wanted to keep this information in localStorage for other purposes. But we can always re-put it there.

Removed for now

peterpeterparker (approved):
LGTM, thanks

PR #1891, opened 2024-07-26
chore(frontend): Add config for POUH credential in production

lmuntaner:
# Motivation

Support POUH issuer in production.

In this PR, I add the origin and canister id of the issuer for production environment.

# Changes

* Add `"ic"` entry for `pouh_issuer` in dfx.json.
* Add one condition more for `POUH_ISSUER_ORIGIN` in frontend constants.

# Tests

I checked the credential from https://id.decideai.xyz with the dummy relying party to make sure the values are as expected.


dfx.json
```diff
@@ -51,7 +51,9 @@
 			"shrink": false,
 			"remote": {
 				"id": {
-					"staging": "qbw6f-caaaa-aaaah-qdcwa-cai"
+					"staging": "qbw6f-caaaa-aaaah-qdcwa-cai",
+					"beta": "qgxyr-pyaaa-aaaah-qdcwq-cai",
```
lmuntaner, line 55: Do we need this? Why isn't it needed in the other canisters?

Do we also need to support this env in `deploy.backend.sh`?

lmuntaner:

dfx.json
peterpeterparker: According your code we do need canister Ids for "beta" and "ic" to use "https://id.decideai.xyz" so I guess yes there are needed.

peterpeterparker (approved):
Thanks

PR #1893, opened 2024-07-26
feat(frontend): change hero background with new animation

AntonioVentilii-DFINITY:
# Motivation

Hero background is updated with an animation. To do that we use [GLSL (OpenGL Shading Language)](https://en.wikipedia.org/wiki/OpenGL_Shading_Language) written by David Aerne and it requires package [three](https://www.npmjs.com/package/three) (we optimize it too, since we don't require it all).

https://github.com/user-attachments/assets/6ef19506-084f-49ef-b47f-7e0b3521c019

https://github.com/user-attachments/assets/3483c9da-3736-4d04-b780-3bbb0b18f788


src/frontend/src/lib/components/ui/ThreeBackground.svelte
```diff
@@ -0,0 +1,170 @@
...
+
+	let renderer: WebGLRenderer;
+	let scene: Scene;
+	let camera: PerspectiveCamera;
+	let material: ShaderMaterial;
+	let mesh: Mesh;
+
+	onMount(() => {
+		init();
+		return () => {
```
peterpeterparker, line 97: similar to `onDestroy`?

src/frontend/src/lib/components/ui/ThreeBackground.svelte
```diff
@@ -0,0 +1,170 @@
...
+	const handleResize = () => {
+		if (isNullish(container)) {
+			return;
+		}
+		renderer.setSize(container.clientWidth, container.clientHeight);
+		camera.aspect = container.clientWidth / container.clientHeight;
+		camera.updateProjectionMatrix();
+	};
+
+	onMount(() => {
```
peterpeterparker, line 151: Can we not duplicate `onMount`?

src/frontend/src/lib/components/ui/ThreeBackground.svelte
```diff
@@ -0,0 +1,170 @@
...
+		window.addEventListener('resize', handleResize);
```
peterpeterparker, line 152: Why not using `svelte:window`?

src/frontend/src/lib/components/ui/ThreeBackground.svelte
```diff
@@ -0,0 +1,170 @@
...
+		return () => {
+			window.removeEventListener('resize', handleResize);
+		};
+	});
+</script>
+
+<div bind:this={container} class="three-background"></div>
+
+<style>
```
peterpeterparker, line 161: `<style lang="scss">`

src/frontend/src/lib/components/ui/ThreeBackground.svelte
```diff
@@ -0,0 +1,170 @@
...
+	.three-background {
```
peterpeterparker, line 162: If interesting, we do not need the class, we can just style the div given that styles are scoped per default

src/frontend/src/lib/components/ui/ThreeBackground.svelte
```diff
@@ -0,0 +1,170 @@
...
+		position: absolute;
```
peterpeterparker, line 163: we cannot solve the position and size with Tailwind? Like `inset-full` or something?

src/frontend/src/lib/components/ui/ThreeBackground.svelte
```diff
@@ -0,0 +1,170 @@
...
+			mesh.geometry.dispose();
+		};
+	});
+
+	const init = () => {
+		scene = new Scene();
+		camera = new PerspectiveCamera(75, container.clientWidth / container.clientHeight, 0.1, 1000);
+		renderer = new WebGLRenderer();
+		renderer.setSize(container.clientWidth, container.clientHeight);
+		container.appendChild(renderer.domElement);
```
peterpeterparker, line 111: Given that we manipulate the DOM manually here, not using Svelte, are the elements properly cleaned on destroy?

src/frontend/src/lib/components/ui/ThreeBackground.svelte
```diff
@@ -0,0 +1,170 @@
...
+		mesh = new Mesh(geometry, material);
+		scene.add(mesh);
+
+		camera.position.z = 5;
+
+		animate();
+	};
+
+	const animate = () => {
+		requestAnimationFrame(animate);
```
peterpeterparker, line 137: Should there be a `cancelAnimationFrame` on destroy?

src/frontend/src/lib/components/ui/ThreeBackground.svelte
```diff
@@ -0,0 +1,170 @@
...
+		material.uniforms.time.value += 0.001;
+		renderer.render(scene, camera);
+	};
+
+	const handleResize = () => {
+		if (isNullish(container)) {
```
peterpeterparker, line 143: Per defition the container cannot be undefined or null, should the definition be updated?

peterpeterparker (commented):
1. What is the impact on the bundle chunks size? If it increase should three be chunked in a separate chunks? What strategy works best on mainnet?

2. Was it tested in every browser including Safari and Firefox?

src/frontend/src/lib/components/ui/ThreeBackground.svelte
AntonioVentilii-DFINITY: right! I need to do some lint on this... i always forget!

src/frontend/src/lib/components/ui/ThreeBackground.svelte
AntonioVentilii-DFINITY: moved to OnDestroy

src/frontend/src/lib/components/ui/ThreeBackground.svelte
AntonioVentilii-DFINITY: merged

src/frontend/src/lib/components/ui/ThreeBackground.svelte
AntonioVentilii-DFINITY: yes, better! i used tailwind classes directly

src/frontend/src/lib/components/ui/ThreeBackground.svelte
AntonioVentilii-DFINITY: indeed! same comment as above

src/frontend/src/lib/components/ui/ThreeBackground.svelte
AntonioVentilii-DFINITY: that is a good point! because I didn't think of it eheh!

src/frontend/src/lib/components/ui/ThreeBackground.svelte
AntonioVentilii-DFINITY: done

src/frontend/src/lib/components/ui/ThreeBackground.svelte
AntonioVentilii-DFINITY: I am cleaning up with this:

```typescript
cancelAnimationFrame(animationRequestId);
renderer.dispose();
scene.remove(mesh);
material.dispose();
mesh.geometry.dispose();
```

something missing, you think?

src/frontend/src/lib/components/ui/ThreeBackground.svelte
AntonioVentilii-DFINITY: i put it there just as "insurance", let's say: I am afraid on runtime there could be some mismatch so that it may eventually come to be `undefined`

do you think it is useless? shall i remove the check?

AntonioVentilii-DFINITY:
1. As discussed, we will check it on `beta`

2. I tested in both (on mobile too) and it was not giving any issue at first sight (30 seconds).

3. I tested leaving the script running 15 mins and moving around back and forth a few times:

Screenshot 2024-07-29 at 14 58 08

src/frontend/src/lib/components/ui/ThreeBackground.svelte
peterpeterparker: No what I mean is that we are mixing a Svelte approach with a vanilla modification of the DOM but, yeah, maybe it works given that Svelte does not use a virtual dom.

src/frontend/src/lib/components/ui/ThreeBackground.svelte
peterpeterparker: Make sense, then let's be consistent and declare the type as optional too:

```
let container: HTMLDivElement | undefined | null;
```

AntonioVentilii-DFINITY: right! done!

peterpeterparker (approved):
LGTM, thanks

Let's deploy and test. It might need some iteration, or not, let's see.

PR #1897, opened 2024-07-29
chore(frontend): skip selected ckERC20 logos generation

peterpeterparker:
# Motivation

In PR #1896, the logos of ckUSDC and ckUSDT were generated, which is technically correct but incorrect given the requirements that Oisy uses distinctive logos for USDC/ckUSDC and USDT/ckUSDT.

# Changes

There isn't a technical flag to detect this product inconsistency; therefore, this PR adds a hardcoded rule in the script to generate the logos.

### DIFF
--- a/scripts/build.tokens.ckerc20.mjs
+++ b/scripts/build.tokens.ckerc20.mjs
@@ -13,6 +13,16 @@ const agent = await createAgent({
 	host: 'https://icp-api.io'
 });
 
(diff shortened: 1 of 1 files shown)


AntonioVentilii-DFINITY (approved):
LGTM, tks!

AntonioVentilii-DFINITY:
BTW, @anedos-dfinity mentioned that maybe these would be the only coins that will have different logos from ck*

PR #1902, opened 2024-07-29
feat(frontend): New user profile store not persisted in localStorage

lmuntaner:
# Motivation

Simplify how user profile is stored in the frontend.

In this PR, create and use a new userProfileStore that doesn't persist data in localStorage.

# Changes

* New store user-profile.store
* User the new userProfileStore instead of the one from settings.store.
* Remove the userProfileStore from settings.store

# Tests

Tested manually.

### DIFF
--- a/src/frontend/src/lib/stores/user-profile.store.ts
+++ b/src/frontend/src/lib/stores/user-profile.store.ts
@@ -1,0 +1,24 @@
+import type { UserProfile } from '$declarations/backend/backend.did';
+import { writable, type Readable } from 'svelte/store';
+
+// * `undefined` means the store is not loaded yet.
+// * `null` means there was an error.
+// * `UserProfile` is the data.
+export type UserProfileStoreData = UserProfile | undefined | null;
+
+export interface UserProfileStore extends Readable<UserProfileStoreData> {
+	set: (data: UserProfileStoreData) => void;
+	reset: () => void;
+}
+
+const initUserProfileStore = (): UserProfileStore => {
+	const { subscribe, set } = writable<UserProfileStoreData>(undefined);
+
+	return {
+		set: (data: UserProfileStoreData | null) => set(data),
+		reset: () => set(undefined),
+		subscribe
+	};
+};
+
+export const userProfileStore = initUserProfileStore();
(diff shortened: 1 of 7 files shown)


lmuntaner:
@peterpeterparker @AntonioVentilii-DFINITY as discussed offline I changed the userProfileStore. Please review. Thanks!

src/frontend/src/lib/stores/user-profile.store.ts
```diff
@@ -0,0 +1,24 @@
+import type { UserProfile } from '$declarations/backend/backend.did';
+import { writable, type Readable } from 'svelte/store';
+
+// * `undefined` means the store is not loaded yet.
+// * `null` means there was an error.
+// * `UserProfile` is the data.
+export type UserProfileStoreData = UserProfile | undefined | null;
```
peterpeterparker, line 7: Similar to the `address.store`, don't we need the `certified: boolean;` information as well in store?

src/frontend/src/lib/stores/user-profile.store.ts
```diff
@@ -0,0 +1,24 @@
...
+	set: (data: UserProfileStoreData) => void;
+	reset: () => void;
+}
+
+const initUserProfileStore = (): UserProfileStore => {
+	const { subscribe, set } = writable<UserProfileStoreData>(undefined);
+
+	return {
+		set: (data: UserProfileStoreData | null) => set(data),
+		reset: () => set(undefined),
```
peterpeterparker, line 19: `set(null)`

lmuntaner: This would mean that there was an error. `reset` for me is to go back to the initial state, which is set with `writable<UserProfileStoreData>(undefined)`.

Does that make sense?

I can change it though.

src/frontend/src/lib/stores/user-profile.store.ts
lmuntaner: I'll add it, it makes sense.

lmuntaner:

src/frontend/src/lib/stores/user-profile.store.ts
peterpeterparker: Please do. In Oisy, reset === null

lmuntaner: done!

peterpeterparker (approved):
LGTM, thanks

PR #1903, opened 2024-07-29
fix(devops): Add Wasm and did file for orbit

bitdivine:
# Motivation
The `orbit` canister entry in `dfx.json` was missing candid data, causing `npm run generate` to fail.

# Changes
* Add Candid and Wasm information for the Orbit canister.

# Tests
* I have run `npm run generate` locally, with success.  (After deploying canisters locally!)

It would be nice to run this in CI but at present we run `dfx start` and deploy canisters only in docker, which complicates this.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -24,6 +24,9 @@
 			"gzip": true
 		},
 		"orbit": {
+			"type": "custom",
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
Thx

PR #1904, opened 2024-07-29
fix(backend): Add missing issuer field to generated files

bitdivine:
# Motivation
`npm run generate` was blocked for a while.  In the meantime a new `issuer` field was added to `UserCredential`.  That change was not propagated to the generated files.  Now that generation works again, the field needs to be propagated.

# Changes
* Run `npm run generate` and commit the changes.
  * The changes are only the addition of the `issuer` field in auto-generated files.

# Tests
None.  Existing CI should suffice.


lmuntaner (approved):
Thanks!

PR #1915, opened 2024-07-29
feat(backend): List allowed callers

bitdivine:
# Motivation

We need to be able to prove that the backend state is as expected.  For example, that the list of allowed callers does not contain additional entries.

# Changes
- Add an endpoint that returns the backend configuration.
- Allow only allowed principals to check who is allowed.
- Move the `Config` type to the shared library, as it is now externally accessible.

# Tests
- A system test is included that verifies that the config is available and correct to allowed users only.


src/declarations/backend/backend.did.d.ts
```diff
@@ -27,6 +27,12 @@ export interface CanisterStatusResultV2 {
 	module_hash: [] | [Uint8Array | number[]];
 }
 export type CanisterStatusType = { stopped: null } | { stopping: null } | { running: null };
+export interface Config {
+	ecdsa_key_name: string;
+	allowed_callers: Array<Principal>;
+	supported_credentials: [] | [Array<SupportedCredential>];
+	ic_root_key_raw: [] | [Uint8Array | number[]];
```
AntonioVentilii-DFINITY, line 34: just for my info: wouldn't it always be `Uint8Array`? In the sense, is it not redundant with `number[]`? Or we expect that they could be passed as `number` and parsed after?

src/backend/src/lib.rs
```diff
@@ -165,6 +165,14 @@ fn post_upgrade(arg: Option<Arg>) {
     }
 }
 
+/// Show the canister configuration.
+#[query(guard = "caller_is_allowed")]
+#[allow(clippy::needless_pass_by_value)]
```
peterpeterparker, line 170: The function takes no parameters, isn't `needless_pass_by_value` unnecessary?

src/backend/src/lib.rs
```diff
@@ -165,6 +165,14 @@ fn post_upgrade(arg: Option<Arg>) {
...
+#[must_use]
```
peterpeterparker, line 171: Other API endpoints do not specify `must_use` except `http_request`. Should we extend all methods with it or should it be removed here?

src/backend/src/lib.rs
```diff
@@ -165,6 +165,14 @@ fn post_upgrade(arg: Option<Arg>) {
...
+fn config() -> Config {
```
peterpeterparker, line 172: Can you provide tests? Like one that assert the result and one that checks that the function can only be called allowed callers?

bitdivine: Sure thing.  The tests are a lot bigger than this change so I was going to put them in a separate PR but I can include them here.  Done.

src/backend/src/lib.rs
```diff
@@ -107,38 +106,12 @@ pub struct State {
     user_profile_updated: UserProfileUpdatedMap,
 }
 
-#[derive(CandidType, Deserialize)]
```
bitdivine, line 110: Moved to the library of shared types, as it i snow publicly accessible.

src/backend/src/lib.rs
```diff
@@ -107,38 +106,12 @@ pub struct State {
...
-}
-
 fn set_config(arg: InitArg) {
-    let InitArg {
-        ecdsa_key_name,
-        allowed_callers,
-        supported_credentials,
-        ic_root_key_der,
-    } = arg;
+    let config = Config::from(arg);
```
bitdivine, line 110: The conversion has been factored out.

src/shared/src/types.rs
```diff
@@ -32,6 +33,43 @@ pub enum Arg {
     Upgrade,
 }
```
bitdivine, line 35: The Config and the conversion have been moved here from the backend lib.rs.

src/declarations/backend/backend.did.d.ts
peterpeterparker: There is no rules, from experience I've got the feeling that I use more often `Uint8Array` in NodeJS context and `number[]` in browser context but probably depends of the features.

As Max said, we have to live with those types, there are automatically generated and defined by Candid.

src/backend/src/lib.rs
peterpeterparker: Thanks for the exaplanation. Ok then, let's ensure that when we add new endpoints we also use `#[must_use]` and someday let's add it to existing functions.

src/shared/src/types.rs
```diff
@@ -32,6 +33,43 @@ pub enum Arg {
...
+pub struct Config {
+    pub ecdsa_key_name: String,
+    // A list of allowed callers to restrict access to endpoints that do not particularly check or use the caller()
+    pub allowed_callers: Vec<Principal>,
+    pub supported_credentials: Option<Vec<SupportedCredential>>,
+    /// Root of trust for checking canister signatures.
+    pub ic_root_key_raw: Option<Vec<u8>>,
+}
+
+impl From<InitArg> for Config {
```
peterpeterparker, line 46: can you move the impl into module `shared/src/impls`?

it might not be a standard Rust pattern, it just something I do and which we started to do in Oisy too per extension.

peterpeterparker (approved):
LGTM, thanks

src/shared/src/types.rs
bitdivine: Sure thing.

bitdivine: Done

peterpeterparker: Thanks

PR #1918, opened 2024-07-29
refactor(frontend): move sorting by balance to network tokens

AntonioVentilii-DFINITY:
# Motivation

It is not really necessary to sort by balance (and pin the tokens with balance on top) at the tokens level. To spare computational time we apply the function `pinTokensWithBalanceAtTop` to the list of network tokens (that supposedly is smaller or has the same length).

Unrelated: we add some docstring to the sorting functions.

# Changes

- Remove `pinTokensWithBalanceAtTop` from derived `sortedTokens`.
- Create global derived store `sortedNetworkTokensUi` that pins tokens with balance on top (and sorts them).
- Adapt other stores and usage.

# Tests

No changes on local replica.


peterpeterparker (approved):
LGTM, thx

PR #1919, opened 2024-07-30
feat(frontend): Set derivation origin for VC same as logging in

lmuntaner:
# Motivation

Derivation origin might be set when logging in. We need to keep the same derivation origin when performing the Verifiable Credentials flow.

# Changes

* New auth utils helper `getDerivationOrigin`.
* Use new helper in auth store when signing in and in service `requestPouhCredentials`

# Tests

Tested locally, but it must be tested as well in staging.


lmuntaner:

src/frontend/src/lib/utils/auth.utils.ts
```diff
@@ -39,3 +39,11 @@ export const isAlternativeOrigin = (): boolean => {
 };
 
 export const hasDerivationOrigin = (): boolean => notEmptyString(AUTH_DERIVATION_ORIGIN);
```
peterpeterparker, line 41: This does not require to be exposed anymore?

peterpeterparker: same for `isAlternativeOrigin`?

src/frontend/src/lib/utils/auth.utils.ts
```diff
@@ -39,3 +39,11 @@ export const isAlternativeOrigin = (): boolean => {
...
+
+export const getDerivationOrigin = (): { derivationOrigin: string } | Record<string, never> => {
```
peterpeterparker, line 43: Should we reflect the fact that it might return no derivation origin in the function's name?

`getOptionalDerivationOrigin` or `optionalDerivationOrigin` or better name?

src/frontend/src/lib/utils/auth.utils.ts
lmuntaner: Yes, good catch. Done!

src/frontend/src/lib/utils/auth.utils.ts
lmuntaner: Sounds good to me, done!

lmuntaner:

lmuntaner:
@peterpeterparker ready for next review

peterpeterparker (approved):
LGTM, thx

PR #1927, opened 2024-07-30
feat(frontend): use erc20 contract to populate token symbol

peterpeterparker:
# Motivation

The script `add.tokens.erc20.sh` use ckERC20 symbol stripped of `ck` to generate the symbol of the ERC20 counter part. 
However, while importing `WSTETH`, I noticed that the effective symbol is actually `wstETH`.
Therefore the script has to fetch the symbol from the contract (in addition to name and decimals) to generate an accurate symbol.

# Changes

- Use `'function symbol() public view returns (string)'` and apply symbol to corresponding fields of the generated TS code.

### DIFF
--- a/scripts/add.tokens.erc20.mjs
+++ b/scripts/add.tokens.erc20.mjs
@@ -202,7 +202,7 @@ export const ${mainnetToken}: RequiredErc20Token = {
 	const newFileContentTestnet =
 		nonNullish(testnetToken) && !testnetTokenCreated
 			? `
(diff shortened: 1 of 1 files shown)


AntonioVentilii-DFINITY (approved):
LGTM!

There is just a lint warning about an unused variable, before merging: we shall use the testnet-symbol in the code

peterpeterparker:
>  we shall use the testnet-symbol in the code

Good catch. Done.

AntonioVentilii-DFINITY (approved):

PR #1929, opened 2024-07-30
feat(lint): add arrow-body-style rule as warning

AntonioVentilii-DFINITY:
# Motivation

A bit opinionated, but I would like to propose the usage of the lint rule [arrow-body-style](https://eslint.org/docs/latest/rules/arrow-body-style) when needed.

In this same PR, we fix the issues pointed out by the new rule.


peterpeterparker (approved):
I rather like the other but, can definitely live and work with arrow-body-style sure.

LGTM; thanks!

AntonioVentilii-DFINITY:
Ah wait, then we can force the {return..} with the same rule. My idea is to have one of the two and not a mix.

peterpeterparker:
Sounds good to me, thanks!

PR #1930, opened 2024-07-30
refactor(frontend): extract component that shows the fee amount

AntonioVentilii-DFINITY:
# Motivation

We extract the common component that shows the fee amount for ETH/ERC20, with optional validation error.
The transition animation is delayed between the fee and the possible validation error.

https://github.com/user-attachments/assets/badaaa0c-8492-436f-84ef-1ed07b510412


AntonioVentilii-DFINITY:
@peterpeterparker as discussed in a previous PR, I tried to avoid using the `timer` but giving a `delay` to the error transition. However, it seems it is not working properly.

src/frontend/src/icp-eth/components/fee/FeeAmountDisplay.svelte
```diff
@@ -0,0 +1,50 @@
...
+	import { balancesStore } from '$lib/stores/balances.store.js';
+	import { BigNumber } from '@ethersproject/bignumber';
+	import type { TokenId } from '$lib/types/token';
+	import { onDestroy } from 'svelte';
+
+	export let fee: BigNumber;
+	export let feeSymbol: string;
+	export let feeTokenId: TokenId;
+
+	const balance = nonNullish($balancesStore)
```
peterpeterparker, line 17: Still not really convinced by that timeout I have to say.

Delay the animation still does not work?
If not, what about using debounce?

src/frontend/src/icp-eth/components/fee/FeeAmountDisplay.svelte
```diff
@@ -0,0 +1,50 @@
...
+	$: {
+		timer = setTimeout(() => {
+			insufficientFeeFunds = nonNullish(balance) && balance.lt(fee);
+		}, 250);
+	}
+
+	onDestroy(() => clearTimeout(timer));
+</script>
+
+<div transition:fade>
```
peterpeterparker, line 33: There is no if so fade as no effect here

src/frontend/src/eth/components/fee/FeeDisplay.svelte
```diff
@@ -46,38 +40,13 @@
...
-							displayDecimals: EIGHT_DECIMALS
-						}),
-						$symbol: $feeSymbolStore ?? ''
-					})}
-				</p>
-			{/if}
-		</div>
+<div id="balance" class="font-normal px-4.5 mb-4 break-all min-h-6">
+	{#if nonNullish(fee) && nonNullish($feeSymbolStore) && nonNullish($feeTokenIdStore)}
+		<FeeAmountDisplay {fee} feeSymbol={$feeSymbolStore} feeTokenId={$feeTokenIdStore} />
```
peterpeterparker, line 50: Not sure, instead of passing `feeSymbol` and `feeTokenId`, the component `FeeAmountDisplay` cannot retrieve those value from the `FeeContext`?

AntonioVentilii-DFINITY: It is used by `EthereumEstimatedFee` too, and in that case the values are passed differently. We can try and armonize the two flows

peterpeterparker: Ah ok. Then let's keep it as implemented, thanks.

src/frontend/src/icp-eth/components/fee/FeeAmountDisplay.svelte
peterpeterparker: It does not work because there is an issue in the code. `balance` should be reactive and not static.

```
// Not good
const balance = nonNullish($balancesStore) ...

// Better
let balance = nonNullish($balancesStore) ...
```

That said, fixing the issue fixes the animation but, it's a bit glitchy on first load. So here's your code:

```
<script lang="ts">
	import { debounce, nonNullish } from '@dfinity/utils';
	import { fade, slide } from 'svelte/transition';
	import { formatToken } from '$lib/utils/format.utils';
	import { EIGHT_DECIMALS } from '$lib/constants/app.constants';
	import { balancesStore } from '$lib/stores/balances.store.js';
	import { BigNumber } from '@ethersproject/bignumber';

	export let fee: BigNumber;
	export let feeSymbol: string;
	export let feeTokenId: TokenId;

	let balance: BigNumber | undefined;
	$: balance = nonNullish($balancesStore)
		? $balancesStore[feeTokenId]?.data ?? BigNumber.from(0n)
		: undefined;

	let insufficientFeeFunds = false;
	const debounceCheckFeeFunds = debounce(
		() => (insufficientFeeFunds = nonNullish(balance) && balance.lt(fee))
	);

	$: balance, debounceCheckFeeFunds();
</script>

<div transition:fade>
	{formatToken({
		value: fee,
		displayDecimals: EIGHT_DECIMALS
	})}
	{feeSymbol}
</div>
{#if insufficientFeeFunds && nonNullish(balance)}
	<p in:slide={{ duration: 250 }} class="text-cyclamen">
		{replacePlaceholders($i18n.send.assertion.not_enough_tokens_for_gas, {
			$balance: formatToken({
				value: balance,
				displayDecimals: EIGHT_DECIMALS
			}),
			$symbol: feeSymbol ?? ''
		})}
	</p>
{/if}
```

- balance should be reactive
- debounce should be a const to not be reactive
- `in:slide` instead of `transition:slide` is more pleasant

AntonioVentilii-DFINITY: one question: isn't this

```typescript
const debounceCheckFeeFunds = debounce(
  () => (insufficientFeeFunds = nonNullish(balance) && balance.lt(fee))
);

$: balance, debounceCheckFeeFunds();
```
	
	the same as this?
	
```typescript
$: {
  debounce(() => (insufficientFeeFunds = nonNullish(balance) && balance.lt(fee)), 250)();
}
```

in sense of reactivity?

and shouldn't we add `fee` as a reactive component of the debounce? so that if the fee updates, it checks again

peterpeterparker: Not absolutely entirely sure, but from experience, I noticed the first method to be reliable while the second is not always.

The reason is that in the first method, you have one instance of the debounce that debounces a particular function. While in the second method, because of the reactivity, you might end up with multiple instances of the debounce and the function, and therefore the debounce itself does not work properly.

It might not be the case here, but as I said, from experience, I know the first method is reliable.

Does that make sense?

src/frontend/src/icp-eth/components/fee/FeeAmountDisplay.svelte
peterpeterparker: I'm really suprised. I was under the impression that Svelte requires any animation tag to be contained within a direct parent if. Also thought there was a lint check for that. Probably misunderstood it it works. Thanks for the feebdack.

src/frontend/src/icp-eth/components/fee/FeeAmountDisplay.svelte
AntonioVentilii-DFINITY: Yes, understood, makes sense then to use the reliable one. And what about `fee` then? Shouldn't the code be:

`$: balance, fee, debounceCheckFeeFunds();`

peterpeterparker: Yes, absolutely. Good catch.

peterpeterparker (approved):
LGTM, thx

PR #1932, opened 2024-07-30
refactor(backend): Add UserProfileModel to interact with user profiles in stable memory

lmuntaner:
# Motivation

Refactor the code so that the different maps managing user profiles are not mismanaged and we introduce bugs and inconsistent data.

This was requested in https://github.com/dfinity/oisy-wallet/pull/1827/files#r1695268288

# Changes

* Create a new struct UserProfileModel and implement it. This entity will have the responsibility to interact with the stable memory to retrieve and store user profiles.
* Use the new struct in the user_profile module instead of expecting the stable memory structures.

# Tests

Not yet. I want to check that the implementation is the requested one.

### DIFF
--- a/src/backend/src/lib.rs
+++ b/src/backend/src/lib.rs
@@ -38,7 +38,7 @@ use types::{
     Candid, ConfigCell, CustomTokenMap, StoredPrincipal, UserProfileMap, UserProfileUpdatedMap,
     UserTokenMap,
 };
-use user_profile::{add_credential, create_profile, get_profile};
+use user_profile::{add_credential, create_profile, find_profile};
 use user_profile_model::UserProfileModel;
 
 mod assertions;
@@ -493,7 +493,7 @@ fn get_user_profile() -> Result<UserProfile, GetUserProfileError> {
     mutate_state(|s| {
         let mut user_profile_model =
             UserProfileModel::new(&mut s.user_profile, &mut s.user_profile_updated);
-        match get_profile(stored_principal, &mut user_profile_model) {
+        match find_profile(stored_principal, &mut user_profile_model) {
             Ok(stored_user) => Ok(UserProfile::from(&stored_user)),
             Err(err) => Err(err),
         }
--- a/src/backend/src/lib.rs
+++ b/src/backend/src/lib.rs
@@ -39,6 +39,7 @@ use types::{
     UserTokenMap,
 };
 use user_profile::{add_credential, create_profile, get_profile};
+use user_profile_model::UserProfileModel;
 
 mod assertions;
 mod config;
@@ -48,6 +49,7 @@ mod oisy_user;
 mod token;
 mod types;
 mod user_profile;
+mod user_profile_model;
 
 const CONFIG_MEMORY_ID: MemoryId = MemoryId::new(0);
 const USER_TOKEN_MEMORY_ID: MemoryId = MemoryId::new(1);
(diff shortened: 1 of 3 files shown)


bitdivine (commented):
Thank you.  This seems like a useful step towards data integrity!

src/backend/src/user_profile.rs
```diff
@@ -1,62 +1,31 @@
...
         let now = time();
         let default_profile = StoredUserProfile::from_timestamp(now);
-        store_user_profile(
-            principal,
-            now,
-            &default_profile,
-            user_profile_map,
-            user_profile_updated_map,
-        );
+        user_profile_model.stor_new(principal, now, &default_profile);
```
peterpeterparker, line 28: typo `stor`?

lmuntaner: thanks!

src/backend/src/lib.rs
```diff
@@ -455,13 +458,14 @@ fn add_user_credential(request: AddUserCredentialRequest) -> Result<(), AddUserC
         current_time_ns as u128,
     ) {
         Ok(()) => mutate_state(|s| {
+            let mut user_profile_model =
+                UserProfileModel::new(&mut s.user_profile, &mut s.user_profile_updated);
```
bitdivine, line 462: I would recommend making this a method of `State`.  Something like `s.user_profile_model() -> UserProfileModel{...}`.  That way `s.user_profile*` can be made private and accessible only via the model.  However that can be in another PR, just to keep the PRs focused.  I didn't mention it before for this reason but it  might be worth mentioning. :-)

src/backend/src/user_profile_model.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+        self.user_profile_updated_map
+            .insert(user_principal, timestamp);
+        self.user_profile_map
+            .insert((timestamp, user_principal), Candid(new_user.clone()));
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
bitdivine, line 49: I believe that consistency of the `UserProfileModel` can be defined by:

* The number of entries in the two underlying maps is the same.
* For every entry in the principal->timestamps map, there is a corresponding entry in the other map.

I would recommend making a method like this, and checking consistency after any CRUD operations:

```
untested pseudocode!
impl UserProfileModel {
    #[cfg(test)]  // Compiled in only in test code.  Not needed in production.
    fn assert_consistent(&self) {
      assert_eq!(self.user_profile.len(), self.user_profile_timestamps.len(), "The number of entries should be the same");
      for (principal, timestamp) in self.... {
         assert!(self.user_profiles.contains((timestamp, principal)), "...");
      }
    }
}
```
This way it becomes possible to test integrity easily, just by calling `.assert_consistent()`, even after we add various other methods or in proptests.  What do you think?

src/backend/src/user_profile_model.rs
```diff
@@ -0,0 +1,163 @@
...
+        let another_now: Timestamp = now + 400000000;
+        let user_profile_2 = StoredUserProfile::from_timestamp(another_now);
+        user_profile_map.insert(
+            (another_now, user_principal_2),
+            Candid(user_profile_2.clone()),
+        );
+        user_profile_updated_map.insert(user_principal_2, another_now);
+
+        let user_profile_model =
+            UserProfileModel::new(&mut user_profile_map, &mut user_profile_updated_map);
```
bitdivine, line 95: I would insert these using the `user_profile_model.insert()` API but this works too!

bitdivine (approved):
LGTM, thanks!

src/backend/src/lib.rs
lmuntaner: That's a good idea, I didn't think on that.\

I agree, let's keep it in another PR.

But remember that I'm going on holidays and when I'm back I'll be working mostly on II. I'll try to find time to finish the refactor, but feel free to do so without me.

src/backend/src/user_profile_model.rs
lmuntaner: Why? In my mind it doesn't make sense to have a user_profile_model withuot the btrees. That's why I set them in the initialization.

src/backend/src/user_profile_model.rs
lmuntaner: Done!

PR #1936, opened 2024-07-30
feat(frontend): add about modals buttons in the main Menu

AntonioVentilii-DFINITY:
# Motivation

We add `AboutWhat` and `AboutHow` buttons among the items in `Menu`. Furthermore, we include an icon in both of them: it will be used both for the menu when logged in and the menu when logged out (in small screen).

# Changes

- Add `IconInfo` to both button `AboutItem`.
- Include `AboutWhat` and `AboutHow` among the `Menu` items.
- Update the labels: from question to affirmation.
- Unrelated: adjust the spacing and the font size of the buttons.

# Tests

Screenshot 2024-07-30 at 17 19 45
Screenshot 2024-07-30 at 17 19 53
Screenshot 2024-07-30 at 17 20 18
Screenshot 2024-07-30 at 17 20 24

### DIFF
--- a/src/frontend/src/lib/components/core/Menu.svelte
+++ b/src/frontend/src/lib/components/core/Menu.svelte
@@ -17,9 +17,6 @@
 	import { page } from '$app/stores';
 	import AboutHow from '$lib/components/hero/about/AboutHow.svelte';
 	import AboutWhat from '$lib/components/hero/about/AboutWhat.svelte';
-	import { modalAboutHow, modalAboutWhat } from '$lib/derived/modal.derived';
-	import AboutWhatModal from '$lib/components/hero/about/AboutWhatModal.svelte';
-	import AboutHowModal from '$lib/components/hero/about/AboutHowModal.svelte';
 
 	let visible = false;
 	let button: HTMLButtonElement | undefined;
@@ -83,9 +80,3 @@
 		<SignOut on:icLogoutTriggered={hidePopover} />
 	</div>
 </Popover>
-
-{#if $modalAboutWhat}
-	<AboutWhatModal />
-{:else if $modalAboutHow}
-	<AboutHowModal />
-{/if}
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -68,6 +72,11 @@
 
 		<Hr />
 
-		<SignOut on:icLogoutTriggered={() => (visible = false)} />
+		<AboutWhat asMenuItem on:icOpenAboutModal={hidePopover} />
```
peterpeterparker, line 75: `on:click` is not used - i.e. the modal does not appear / work. Is that expected?

AntonioVentilii-DFINITY: I was going to do it in another PR, but it does not make sense... I included the modals here.

src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -68,6 +75,17 @@
...
+		<AboutHow asMenuItem on:icOpenAboutModal={hidePopover} />
+
+		<Hr />
+
+		<SignOut on:icLogoutTriggered={hidePopover} />
 	</div>
 </Popover>
+
+{#if $modalAboutWhat}
```
peterpeterparker, line 87: That does not work I think, aren't the modal duplicated in the UI when you open it? Like two About modals one above another one?

AntonioVentilii-DFINITY: these are inside the Menu component that is rendered once logged in, while the others are inside the AboutMenu that is rendered only when logged out. Is that what you meant?

AntonioVentilii-DFINITY: do you prefer to put them both outside and in the `Header` component?

peterpeterparker: Right, it works for now indeed. Nevertheless, that's prone to duplication, I think it would be worth declaring the modals only once. Maybe we can move them one level higher?

AntonioVentilii-DFINITY: perfect! moving them to `Header`

peterpeterparker (approved):
LGTM, thanks

PR #1937, opened 2024-07-30
feat(frontend): add Oisy short name to env variables

AntonioVentilii-DFINITY:
# Motivation

Include short name for Oisy among the `env` variables.

NOTE: before merging, we modify the `.env.*` files that we use to deploy.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,6 +5,7 @@ VITE_INFURA_API_KEY=
 VITE_ALCHEMY_API_KEY=
 VITE_WALLET_CONNECT_PROJECT_ID=
 VITE_OISY_NAME="Oisy Wallet"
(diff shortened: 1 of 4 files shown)


peterpeterparker:
Actually I was thinking that we should remove the `VITE_OISY_*` title, description and things not related to URL from the environment and transform those into constant. As Oisy becomes a product, not particular reason to have those has options? WDYT?

AntonioVentilii-DFINITY:
Ah yes, sure. It is actually better I think, otherwise every change must be done on each `.env` file. I close this PR and open a new one then

AntonioVentilii-DFINITY:
Done with:

- PR #1943

PR #1940, opened 2024-07-30
fix(frontend): remove sorting function from global store

AntonioVentilii-DFINITY:
# Motivation

Removing sorting function from network tokens store.


peterpeterparker (approved):
LGTM, thx

PR #1942, opened 2024-07-31
chore(frontend): add todos to cache ERC20 metadata

peterpeterparker:
# Motivation

Add the TODOS we have to resolve to reduce the numbers of calls to Infura.

# Changes

- Add few `TODO(GIX-2740)` (Jira task number).

### DIFF
--- a/src/frontend/src/eth/services/erc20.services.ts
+++ b/src/frontend/src/eth/services/erc20.services.ts
@@ -119,7 +119,7 @@ const loadErc20UserTokens = async (params: {
 						},
 						// 1. TODO(GIX-2740): check uf user token is actually a match in the environment static metadata
 						// +
-						// 2. TODO(GIX-2740): check if metadata for address already loaded and reuse - using Infura is not a certified call anyway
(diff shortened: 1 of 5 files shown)


src/frontend/src/eth/services/erc20.services.ts
```diff
@@ -116,6 +117,9 @@ const loadErc20UserTokens = async (params: {
 							version: fromNullable(version),
 							enabled: fromNullable(enabled) ?? true
 						},
+						// 1. TODO(GIX-2740): check uf user token is actually a match in the environment static metadata
+						// +
+						// 2. TODO(GIX-2740): check if metadata for address already loaded and reuse - using Infura is not a certified call anyway
```
AntonioVentilii-DFINITY, line 122: here you mean to cache what we load? I imagine for a specific period of time, to guarantee that any change in metadata is updated?

AntonioVentilii-DFINITY (approved):

src/frontend/src/eth/services/erc20.services.ts
peterpeterparker: I just mean check if metadata is already loaded in store. Added the keyword "store".

PR #1907, opened 2024-07-29
feat(frontend): create SendModal with initial step of choosing the token

AntonioVentilii-DFINITY:
# Motivation

The following is the generic `SendModal` that has as first step the choice of token. The rest will be the same as the modals for ETH and IC.

PR #1944, opened 2024-07-31
refactor(frontend): merge sendWizardSteps

AntonioVentilii-DFINITY:
# Motivation

We merge `sendWizardSteps` and `icSendWizardSteps` into  a single one already being used after PR #1907.


src/frontend/src/lib/config/send.config.ts
```diff
@@ -18,9 +13,21 @@ export const sendWizardSteps = (i18n: I18n): WizardSteps => [
 	{
 		name: WizardStepsSend.SENDING,
 		title: i18n.send.text.sending
-	},
+	}
+];
+
+export const sendWizardStepsWithQrCodeScan = (i18n: I18n): WizardSteps => [
+	...sendWizardSteps(i18n),
```
peterpeterparker, line 20: Why no TOKENS_LIST here?

AntonioVentilii-DFINITY: this is used by the `EthSendModal` and the `IcSendModal`, so we would need an object that has all steps except the tokens list. We can invert and create one that has all and the other one that filters out the TOKENS_LIST step

peterpeterparker: Ahhh now I get it. All good.

peterpeterparker (approved):
LGTM, thx

PR #1945, opened 2024-07-31
feat(devops): Pass flags through to the test

bitdivine:
# Motivation
The backend testing script provides useful setup, but always runs all the tests.  This slows down development.

# Changes
- Pass any command line arguments through to `cargo test`.  This way a test run can be adapted in any way normally supported by `cargo test`.  For example, `./scripts/test.backend.sh foo` will run only those tests that contain `foo` in their name, thus addressing the motivation for this PR.

# Tests
- See existing CI: When no arguments are passed, this runs just as before.
- Commands such as `./scripts/test.backend.sh --frozen test_add_many_custom_tokens_with_index` work as expected, and are significantly faster than running the same without flags.


scripts/test.backend.sh
```diff
@@ -62,4 +62,4 @@ export POCKET_IC_MUTE_SERVER=""
 # Run tests
 
 echo "Running backend integration tests."
-cargo test -p backend
\ No newline at end of file
+cargo test -p backend "${@}"
```
bitdivine, line 65: Note: `"${@}"` is an array of all arguments.  So:
* If there are no arguments, this is a noop.
* A command such as `./scripts/test.backend.sh a b c` changes the above to `cargo test -p backend a b c`
* Spaces and escapes are handled correctly:
  `./scripts/test.backend.sh --frozen "a b c"` changes the above to
  `cargo test -p backend --frozen "a b c"`

peterpeterparker (approved):
Cool, thanks

PR #1946, opened 2024-07-31
refactor(tests): Factor out common pocket-ic setup code

bitdivine:
# Motivation
There are two pocket-ic setup commands.  They are identical apart from the Wasm used.  This means that there is (by now) redundant code and also there is a risk of changes being made to one implementation that should be but are not implemented in the other.

# Changes
- Factor out the common code.

# Tests
- Existing CI should suffice.


peterpeterparker (approved):
LGTM, thx

PR #1948, opened 2024-07-31
refactor(frontend): adapt ThreeBackground size to container

AntonioVentilii-DFINITY:
# Motivation

The component ThreeBackground was not refreshing its size when the container changed (for example, when the user comes back from a token page). It was not clearly visible, but the height of the waves was smaller.

### DIFF
--- a/src/frontend/src/lib/components/ui/ThreeBackground.svelte
+++ b/src/frontend/src/lib/components/ui/ThreeBackground.svelte
@@ -159,6 +159,7 @@ void main(){
 		return {
 			destroy() {
 				resizeObserver.unobserve(node);
+				resizeObserver.disconnect();
 			}
 		};
 	};
--- a/src/frontend/src/lib/components/ui/ThreeBackground.svelte
+++ b/src/frontend/src/lib/components/ui/ThreeBackground.svelte
@@ -112,7 +112,8 @@ void main(){
 		scene = new Scene();
 		camera = new PerspectiveCamera(75, container.clientWidth / container.clientHeight, 0.1, 1000);
 		renderer = new WebGLRenderer();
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/ui/ThreeBackground.svelte
```diff
@@ -152,8 +152,18 @@ void main(){
 		camera.aspect = container.clientWidth / container.clientHeight;
 		camera.updateProjectionMatrix();
 	};
+
+	const resizeAction = (node: HTMLElement) => {
+		const resizeObserver = new ResizeObserver(handleResize);
```
peterpeterparker, line 157: Should the callback be debounced? e.g. if you resize your browser, isn't the obserser triggers lots of time and therefore consumes quite some memory? Or it's fine to resize often?

src/frontend/src/lib/components/ui/ThreeBackground.svelte
```diff
@@ -152,8 +152,18 @@ void main(){
...
+		resizeObserver.observe(node);
+		return {
+			destroy() {
+				resizeObserver.unobserve(node);
```
peterpeterparker, line 161: Shouldn't the observer be disconnected as well?

AntonioVentilii-DFINITY: yes! i forgot to put it in this iteration!

src/frontend/src/lib/components/ui/ThreeBackground.svelte
AntonioVentilii-DFINITY: Sorry, there was a simpler way, please check last commit

peterpeterparker (approved):
LGTM, thx

PR #1949, opened 2024-07-31
fix(frontend): wrong overflow-hidden in Hero

AntonioVentilii-DFINITY:
# Motivation

In Hero component, with ChainFusion background, the overflow was hidden, leading to showing partially the menu popover. This was more clear with bigger menu.

### Before

Screenshot 2024-07-31 at 13 50 12

### After

Screenshot 2024-07-31 at 13 49 56

### DIFF
--- a/src/frontend/src/lib/components/hero/Hero.svelte
+++ b/src/frontend/src/lib/components/hero/Hero.svelte
@@ -21,7 +21,7 @@
 	$: heroContent = usdTotal || summary;
 </script>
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/hero/Hero.svelte
```diff
@@ -62,7 +62,6 @@
 			background: transparent;
 
 			position: relative;
```
peterpeterparker, line 64: `position: relative` still needed?

peterpeterparker (approved):
LGTM, thanks

PR #1950, opened 2024-07-31
feat(backend): Controllers are allowed

bitdivine:
# Motivation
Some `backend` canister APIs are gated to authorized maintainers.  The list of maintainers is kept in a list of `allowed_users`.  Controllers of this canister are also maintainers and, as discussed with the oisy team, it seems reasonable to consider controllers to be allowed to call these endpoints.

# Changes
* Allow canister controllers to call restricted endpoints.
* Update the pocket-ic setup to set controllers.

# Tests
The `config()` API is has tests that only allowed users can call the API.  I have updated this to check that controllers are able to call `config()`.


src/backend/src/guards.rs
```diff
@@ -11,7 +11,8 @@ pub fn caller_is_not_anonymous() -> Result<(), String> {
 }
 
 pub fn caller_is_allowed() -> Result<(), String> {
-    if read_config(|s| s.allowed_callers.contains(&caller())) {
+    let caller = caller();
+    if read_config(|s| s.allowed_callers.contains(&caller)) || ic_cdk::api::is_controller(&caller) {
```
bitdivine, line 15: This is the main change in this PR.  Controllers are now allowed.

src/backend/tests/it/config.rs
```diff
@@ -25,6 +26,18 @@ fn config_is_available_to_allowed_users_only() {
         update_call::<UserProfile>(&pic_setup, nns_root, "config", ()).is_err(),
         "NNS root should not be able to call config"
     );
```
bitdivine, line 28: Here we check that controllers are able to make a protected API call.  There is an existing check that other users are NOT able to make this API call.

src/backend/tests/it/utils/mock.rs
```diff
@@ -1,6 +1,8 @@
 pub const CALLER: &str = "xzg7k-thc6c-idntg-knmtz-2fbhh-utt3e-snqw6-5xph3-54pbp-7axl5-tae";
 pub const CALLER_ETH_ADDRESS: &str = "0xdd7fec4C49CD2Dd4eaa884D22D92503EabA5A791";
 
+pub const CONTROLLER: &str = "l3lfs-gak7g-xrbil-j4v4h-aztjn-4jyki-wprso-m27h3-ibcl3-2cwuz-oqe";
+
```
bitdivine, line 5: Here, and in the file below, we define the canister controller.

src/backend/tests/it/utils/pocketic.rs
```diff
@@ -63,7 +71,7 @@ pub fn upgrade_with_wasm(
         canister_id.clone(),
         wasm_bytes,
         encode_one(&arg).unwrap(),
-        None,
+        Some(controller()),
```
bitdivine, line 74: Updating the Wasm has to be done by a controller.

src/backend/src/guards.rs
peterpeterparker: NTH: For consistency, just `is_controller` - i.e. using the function at the top (don't know how you call that)

```
use ... etc;
use ic_cdk::caller;
use ic_cdk::api::is_controller;

pub fn caller_is_allowed() -> Result<(), String> {
   let caller = caller();
   
   if read_config(|s| s.allowed_callers.contains(&caller)) || is_controller(&caller) {
```

src/backend/tests/it/config.rs
```diff
@@ -25,6 +26,18 @@ fn config_is_available_to_allowed_users_only() {
...
+    // Try a controller
+    assert_eq!(
+        update_call::<Config>(
+            &pic_setup,
+            Principal::from_text(CONTROLLER)
```
peterpeterparker, line 33: You don't want to reuse the expose `controller()` function of the pocket ic utils here? or on the contrary this particular function should be not public?

peterpeterparker (approved):
LGTM, thanks

src/backend/src/guards.rs
bitdivine: Done

src/backend/tests/it/config.rs
bitdivine: Good catch.  Done.

PR #1952, opened 2024-07-31
feat(frontend): remove FET token

AntonioVentilii-DFINITY:
# Motivation

We were asked to remove the FET token.


peterpeterparker (approved):
Code wise, LGTM. Thx.

~~I wonder if any users have transferred FET to their Oisy Wallet and will end up having their funds stuck there. I guess we will know once it's live.~~

peterpeterparker:
Updated the title of the PR from `refactor(frontend)` to `feat(frontend)!` to reflect the fact that this is breaking change.

AntonioVentilii-DFINITY:
> I wonder if any users have transferred FET to their Oisy Wallet and will end up having their funds stuck there. I guess we will know once it's live.

They can still add the token manually, right?

peterpeterparker:
> They can still add the token manually, right?

RIGHT! Thanks.

Removed the `!` from the title.

PR #1953, opened 2024-07-31
feat(frontend): increase debounce time of load balances

peterpeterparker:
# Motivation

I don't know on mainnet but, at least locally, it seems that bumping the debounce time of the load balances from the default 300ms to 500ms can potentially save up to twice the calls to load the balances. While it is not a super solution nor an awesome engineered solution, it feels like given the current situation it's a pragmatic small improvement worth a shot.

### DIFF
--- a/src/frontend/src/icp-eth/components/core/LoaderBalances.svelte
+++ b/src/frontend/src/icp-eth/components/core/LoaderBalances.svelte
@@ -15,7 +15,7 @@
 		]);
 	};
 
-	const debounceLoad = debounce(load);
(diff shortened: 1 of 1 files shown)


AntonioVentilii-DFINITY (approved):
LGTM, tks!

As a note: is it worth to put a TODO there to revisit it later?

peterpeterparker:
> As a note: is it worth to put a TODO there to revisit it later?

Not sure, feels like it might be one of those todo that remains in code forever ;)

Most important is that we solve the TODOs we opened this morning.

PR #1954, opened 2024-07-31
feat(frontend): include maskable icons

AntonioVentilii-DFINITY:
# Motivation

We add maskable icons that are best practice nowadays for modern Android devices and Google products.

Ref: [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Define_app_icons#support_masking).


peterpeterparker (commented):
Are you sure about the transparent background?

I would recommend a plain background to avoid surprises.

AntonioVentilii-DFINITY:
For what is explained around, the background will be automatically taken from the `background_color` prop of the manifest. So theoretically, it needs a transparent background no? or shall we already make it round with the background we want?

peterpeterparker:
Agree with yout but, the documentation also ends with following:

> Finally, give your maskable icon an opaque background color to fill the entire icon area.

This editor - https://progressier.com/maskable-icons-editor - or article -https://intercom.help/progressier/en/articles/6208771-can-a-pwa-icon-have-a-transparent-background - also mention "a solid background"

That's why I'm not sure but, we can try.

peterpeterparker (approved):
LGTM, thanks

AntonioVentilii-DFINITY:
> This editor - https://progressier.com/maskable-icons-editor - or article -https://intercom.help/progressier/en/articles/6208771-can-a-pwa-icon-have-a-transparent-background - also mention "a solid background"

After having read that, if I am not mistaken, we can actually reuse the same exact images for the maskable icons, but just marking them as maskable. I tested it in the website you provided and the icon we have would work for purpose "maskable"

Screenshot 2024-08-02 at 08 49 37

and in this article it says that the default purpose is "any" that allows the icon to be used in any context (maskable, and non-maskable).

@peterpeterparker if you agree, I would close this PR without merging, since it would not add particular difference for now, unless we decide to have different icons for the different purposes.

peterpeterparker:
> if you agree

Not really. In my opinion, the best practice for creating icons is to use square images that are maskable. This means providing icons with a plain background and a square shape, which ensures compatibility across different devices and platforms. Although other formats might work, as you tested, they can be unreliable because we won't be testing on each individual device. Using square, maskable images is a safe and consistent approach which does not require much effort.

Of course, it's your call—this is only my recommendation.

AntonioVentilii-DFINITY:
Talked with Misha (covering Artem) and he said to use the same background of the favicon

PR #1955, opened 2024-07-31
feat(ci): Check that generated bindings are up to date

bitdivine:
# Motivation
There is no check in CI that candid files are up to date.

Secondary (& real!!) motivation: Frustration with `dfx generate` requiring deployment.  Really annoying when trying to iterate an API.  But might as well use the fast command for something useful, now that we have it.

# Changes
-  Added a script that generates candid bindings reasonably quickly &  without local deployment.

# Tests
- Added a test to CI that the bindings are up to date.

### DIFF
--- a/.github/workflows/binding-checks.yml
+++ b/.github/workflows/binding-checks.yml
@@ -40,8 +40,10 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             target
-            !target/wasm32-unknown-unknown/release/backend.wasm
           key: ${{ runner.os }}-cargo-backend-tests-${{ hashFiles('Cargo.lock', 'rust-toolchain.toml') }}
+      - name: Remove the old Wasm to force a rebuild
+        if: steps.backend-wasm-cache.outputs.cache-hit != 'true'
+        run: rm -f target/wasm32-unknown-unknown/release/backend.wasm
       - name: Prepare
         uses: ./.github/actions/prepare
       - name: Install dfx


scripts/did.sh
```diff
@@ -4,9 +4,10 @@ function generate_did() {
   local canister=$1
   canister_root="src/$canister"
```
bitdivine, line 5: Recompile the backend only if the `Wasm` is not there already.

scripts/generate.sh
```diff
@@ -0,0 +1,47 @@
...
+      http*) curl -sSL "$source" >"$destination" ;;
+      *) if test -e "$source"; then cp "$source" "$destination"; else echo "WARNING: $canister_name did file not found at $source"; fi ;;
+      esac
+    done
+}
+
+# Create local .did files
+./scripts/did.sh
+# Download .did files without an explicit url in dfx.json
+./scripts/download.icp.sh
```
peterpeterparker, line 37: We probably don't require download icp, ckbtc and cketh as we do no generate their types. Any particular reason why you download those here?

scripts/generate.sh
```diff
@@ -0,0 +1,47 @@
...
+	be found for a local deployment.  Much faster!
+	EOF
+  exit 0
+}
+
+# Gets all .did files listed in dfx.json.
+#
+# .did files are placed in the local .dfx directory, where the .did files are expected by `dfx generate` and other commands.
+function install_did_files() {
+  jq -r '.canisters | to_entries | .[] | select(.value.candid != null) | "\(.key) \(.value.candid)"' dfx.json |
```
peterpeterparker, line 20: should we add a check that assert if `jq` is installed at the top of the script? "You don't have jq, install jq to execute this script"

.github/workflows/binding-checks.yml
```diff
@@ -0,0 +1,70 @@
...
+          ./cargo-binstall -y --force "cargo-binstall@$BINSTALL_VERSION"
+          rm cargo-binstall
+      - name: Install candid-extractor
+        run: cargo binstall --no-confirm candid-extractor@0.1.4 && candid-extractor --version
+      - name: Generate bindings
+        run: npm run generate
+      - name: Check bindings
+        run: |
+          test -z "$(git status --porcelain)" || {
+                  echo "FIX: Please run npm run generate"
```
peterpeterparker, line 61: An sub-sub-sub-sub-sub-sub-important idea, maybe "Please execute npm run generate" to avoid the content "run npm run"

peterpeterparker (approved):
LGTM, thanks

scripts/generate.sh
bitdivine: Running a blanket `dfx generate`, as in the original `package.json`, will fail if some candid files are missing, like this:
```
Generating type declarations for canister icp_ledger:
Error: Failed while trying to generate type declarations for 'icp_ledger'.
Caused by: Failed while trying to generate type declarations for 'icp_ledger'.
  Candid file: /home/max/dfn/oisy-wallet/branches/check-generate-in-ci/target/ic/icp_ledger.did doesn't exist.
```
I guess one could generate only the canisters for which there are directories in `declarations`, like this:

```
for canister in $(ls src/declarations/) ; do echo "Generating bindings for $canister" ; dfx generate "$canister" ; done
```
Is that what you mean?  I think it's a good idea.  Less files to download => faster execution. :-)

scripts/generate.sh
bitdivine: Yes, can do.  Added to the `--help` and added a check.

.github/workflows/binding-checks.yml
bitdivine: Good idea.  Done.

scripts/generate.sh
peterpeterparker: I did not knew. I think we should just get rid of `dfx generate` and directly use `candid-extractor`.

That's what I do in Juno. e.g. https://github.com/junobuild/juno/blob/main/scripts/did.sh

I can provide a PR to do so once we have merged this PR.

bitdivine:
Merged but happy to revisit if any of the changes were not as imagined.

PR #1956, opened 2024-07-31
feat(ci): Shell formatting

bitdivine:
# Motivation
The repo currently lacks a shell formatter.  As sis typical when a formatter is missing this is causing spurious diffs.

# Changes
- Add the shell formatter used in the nns-dapp (and other GIX projects).

# Tests
- CI already runs the top-level format check.


scripts/format.sh.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
bitdivine, line 1: This is copied verbatim from nns-dapp.

.github/workflows/backend-checks.yml
```diff
@@ -29,6 +29,8 @@ jobs:
             ~/.cargo/git
             target
```
bitdivine, line 30: Most changes are just formatting.  The only other changes are this here and others highlighted with comments below.

scripts/format.sh
```diff
@@ -6,4 +6,5 @@ set -x
 time xargs -P8 -I{} bash -c "{}" <<EOF
```
bitdivine, line 6: Here we format shell scripts by default.

peterpeterparker (approved):
LGTM, thanks

Should we add a new GitHub Actions checks.yml that runs when a modification is applied in docker or scripts? `backend-checks.yml` runs only when a changes is applied to backend related things while most scripts are not necesserely related with the backend, more local development or build scripts. WDYT?

bitdivine:
@peterpeterparker Thanks for the review.  I agree.  A formatting checks are fast so I would even potentially just run the check every time.  KISS :-)

PR #1957, opened 2024-08-01
refactor(ci): Move the formatting check into a separate workflow

bitdivine:
# Motivation
As discussed in #1956, it makes sense to move the formatting check to a separate workflow to ensure that is run on every relevant change, not just when the backend changes.

# Changes
- Move the format check from the backend workflow to a separate workflow.
- Adapt the rules for which file changes cause the workflow to run.
- Use `cargo binstall` to use a pre-compiled `cargo sort` rather than compiling `cargo sort` on every run.

# Tests
See CI

Note that by far the slowest part of the workflow is compiling `cargo sort`.  This can be sped up by using `cargo binstall` instead of `cargo install`.  Done.


.github/workflows/formatting-checks.yml
```diff
@@ -0,0 +1,43 @@
+name: Formatting Checks
```
peterpeterparker, line 1: Should we name it, even if verbose, "Backend & Scripts Formatting Checks" to make obvious it is not meant to format frontend?

peterpeterparker (approved):
LGTM, thx

.github/workflows/formatting-checks.yml
bitdivine: Sure!  Done (well, a slightly shorter version! :slightly_smiling_face:  )

PR #1959, opened 2024-08-02
feat(frontend): use CoinGecko API to fetch ICRC prices

AntonioVentilii-DFINITY:
# Motivation

It was suggested to use CoinGecko API for ICRC tokens too. They are not all there, but [some are](https://www.coingecko.com/en/categories/internet-computer-ecosystem).

# Changes

- Create derived with enabled ICRC tokens ledger canister IDs.
- Create function to fetch ICRC prices from Coingecko (similar to ERC20).
- Adjust exchange services and workers to include ICRC tokens.
- Adapt derived of exchange store.

# Tests

Tested the final returns from the exchange derived by printing the results in console.

Screenshot 2024-08-02 at 10 34 43


anedos-dfinity (approved):
LGTM, I wonder if it makes sense to merge the ICP token in the ICRC category

src/frontend/src/icp/derived/icrc.derived.ts
```diff
@@ -130,3 +131,8 @@ export const enabledIcrcTokens: Readable<IcToken[]> = derived(
 		...$enabledIcrcCustomTokens
 	]
 );
+
+export const enabledIcrcTokensCanisterIds: Readable<LedgerCanisterIdText[]> = derived(
```
peterpeterparker, line 135: The name is too generic - it's not clear it's ledger canister ids.

src/frontend/src/lib/derived/exchange.derived.ts
```diff
@@ -24,9 +24,9 @@ export const exchanges: Readable<ExchangesData> = derived(
 			[SEPOLIA_TOKEN_ID]: ethPrice,
 			[ICP_TOKEN_ID]: icpPrice,
 			...Object.entries($exchangeStore ?? {}).reduce((acc, [key, currentPrice]) => {
-				const token = $erc20Tokens.find(
-					({ address }) => address.toLowerCase() === key.toLowerCase()
-				);
+				const token =
+					$erc20Tokens.find(({ address }) => address.toLowerCase() === key.toLowerCase()) ??
```
peterpeterparker, line 28: Does that work? Arent' the Icrc token IDs added twice to the object?

```
....
// your code
const token = $erc20To ?? $icrcTokens.find(({ ledgerCanisterId }) => ledgerCanisterId === key);

// later in the same module
...$icrcTokens.reduce((acc, token) => {
				const { id, exchangeCoinId } = token;
				
				etc.
```

src/frontend/src/lib/services/exchange.services.ts
```diff
@@ -36,10 +37,21 @@ export const exchangeRateERC20ToUsd = async (
 		include_market_cap: true
 	});
 
+export const exchangeRateICRCToUsd = async (
+	contractAddresses: LedgerCanisterIdText[]
+): Promise<CoingeckoSimpleTokenPriceResponse | null> =>
+	simpleTokenPrice({
+		id: 'internet-computer',
+		vs_currencies: 'usd',
+		contract_addresses: contractAddresses.map((address) => address.toLowerCase()),
```
peterpeterparker, line 46: ICRC token have addresses?

peterpeterparker (changes requested):

src/frontend/src/lib/types/coingecko.ts
```diff
@@ -6,7 +6,7 @@ export type CoingeckoCoinsId = 'ethereum' | 'bitcoin' | 'internet-computer';
 
 // We are interested in the ERC20 <> USD on Ethereum only, therefore not an exhaustive list.
```
peterpeterparker, line 7: The comments does not match anymore

src/frontend/src/lib/services/exchange.services.ts
AntonioVentilii-DFINITY: in this case I am keeping the coingecko API convention. They define this function with an input called contract address. But in the rest of the code, I changed to canisterId

src/frontend/src/lib/types/coingecko.ts
AntonioVentilii-DFINITY: i had updated it, I think i forgot to push the commit

src/frontend/src/icp/derived/icrc.derived.ts
AntonioVentilii-DFINITY: changed to enabledIcrcTokensLedgerCanisterIds

src/frontend/src/icp/derived/icrc.derived.ts
```diff
@@ -130,3 +131,8 @@ export const enabledIcrcTokens: Readable<IcToken[]> = derived(
 		...$enabledIcrcCustomTokens
 	]
 );
+
+export const enabledIcrcTokensLedgerCanisterIds: Readable<LedgerCanisterIdText[]> = derived(
```
peterpeterparker, line 135: `enabledIcrcLedgerCanisterIds` to shorten it a bit?

src/frontend/src/lib/derived/exchange.derived.ts
AntonioVentilii-DFINITY: From what I understood in the code, that part is done to set the prices of ckBTC and ckETH tokens (that otherwise we wouldn't have before), but even, for some reason that I am not sure about, to have a price in case of custom tokens (see function `loadMetadata`).

I didn't want to change that, BUT you are right, I was confused: if the deconstructed array comes after it will overwrite the values, and not the contrary. I will invert the position of those two parts of the return

src/frontend/src/icp/derived/icrc.derived.ts
AntonioVentilii-DFINITY: done

src/frontend/src/lib/services/exchange.services.ts
AntonioVentilii-DFINITY: yes, makes sense, changed

src/frontend/src/lib/derived/exchange.derived.ts
```diff
@@ -62,7 +43,26 @@ export const exchanges: Readable<ExchangesData> = derived(
 									? icpPrice
 									: undefined
 				};
-			}, {})
+			}, {}),
+			...Object.entries($exchangeStore ?? {}).reduce((acc, [key, currentPrice]) => {
+				const token =
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: well, we are iterating first on the keys and values of the exchange store record, so we would not know beforehand if it is erc20 or icrc, unless we change the keys of the exchange store to save this info too.

However, I think that would be better if we refactor everything, so that we loop only once in `erc20Tokens` and only once in `icrcTokens` (right now we do it twice for both).

If it is ok, I would do it in another PR, I prefer to separate the two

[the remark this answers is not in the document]
peterpeterparker: But we are already iterating on `...$icrcTokens.reduce((acc, token)`. I'm actually a bit confused about the proposed change here. I probably need to give it a try to understand better. Will do Monday and report.

src/frontend/src/lib/components/exchange/ExchangeWorker.svelte
```diff
@@ -22,12 +23,15 @@
 
 	const syncTimer = () => {
 		worker?.stopExchangeTimer();
-		worker?.startExchangeTimer({ erc20Addresses: $enabledMergedErc20TokensAddresses });
+		worker?.startExchangeTimer({
+			erc20Addresses: $enabledMergedErc20TokensAddresses,
+			icrcCanisterIds: $enabledIcrcLedgerCanisterIds
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Yes, better, I created a new derived `enabledIcrcLedgerCanisterIdsNoCk` but I am not convinced of the name

src/frontend/src/lib/derived/exchange.derived.ts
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: I think i understood now your point. have a look at my last commit:

```typescript
                            ...$icrcTokens.reduce((acc, token) => {
				const { id, ledgerCanisterId, exchangeCoinId } = token;

				const icrcPrice = $exchangeStore?.[ledgerCanisterId];

				if (nonNullish(icrcPrice)) {
					return {
						...acc,
						[id]: icrcPrice
					};
				}

				const { twinToken } = token as Partial<IcCkToken>;
                                 [rest of the code]
```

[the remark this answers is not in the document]
peterpeterparker: That's better for performance, thanks

src/frontend/src/lib/components/exchange/ExchangeWorker.svelte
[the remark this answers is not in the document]
peterpeterparker: Thanks. I don't really have a better idea of naming neither. We can iterate from there.

peterpeterparker (approved):
LGTM, thanks

Before merge can you also adapt the type `CoingeckoResponse<T>` 

From:

```
export type CoingeckoResponse<T> = Record<CoingeckoCoinsId | string, T>;
```

To:

```
export type CoingeckoResponse<T> = Record<
	CoingeckoCoinsId | LedgerCanisterIdText | ETH_ADDRESS | string,
	T
>;
```

AntonioVentilii-DFINITY:
Right! I will just wait for PR #1964 to be merged, so that `ETH_ADDRESS` is `EthAddress`

PR #1962, opened 2024-08-02
refactor(frontend): change visual name for ckERC20 tokens

AntonioVentilii-DFINITY:
# Motivation

We will mark the ckERC20 tokens with a tag for the "ck" initial part.

# Changes

- Create new prop `DisplayName` for type `Token`.
- Add new prop for ck tokens in `ICRC_TOKENS` list.
- Create `Tag` component.
- Adapt `TokenCard` with the new prop.


src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -17,7 +19,12 @@
 		aria-label={`Open the list of ${token.symbol} transactions`}
```
peterpeterparker, line 19: Unrelated: an i18n key is missing here

src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -17,7 +19,12 @@
...
 	>
 		<Card noMargin>
-			{token.name}
+			{#if isTokenCkLedger(token)}
+				<Tag>ck</Tag>
```
peterpeterparker, line 23: A screen reader will read this as "ckErc20" or will read it as "ck erc20"?

src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -17,7 +19,12 @@
...
+				{token.twinToken?.name ?? token.name.replace(/^ck/, '')}
```
peterpeterparker, line 24: I'm not a fan of adding an exception for ck tokens within a generic component. Is there a more generic, maintainable and future proof way to approach this rather than hardcoding an exception?

AntonioVentilii-DFINITY: we could create a `displayName` prop for the class `Token` and set it differently for the ck tokens, but my issue was with the Tag component

unless we make `displayName: string | SvelteComponent`

src/frontend/src/lib/components/tokens/TokenCard.svelte
AntonioVentilii-DFINITY: for now "ck erc20"... there is a whitespace since it is inside a `<span>` but we can remove it. I have no strong opinion on this.

What  is better?

For example "USD Coin": what is the difference in having either "ckUSD Coin" or "ck USD Coin"? Or, with USDT: "ckTether USD" or "ck Tether USD"? What is preferable?

src/frontend/src/lib/components/tokens/TokenCard.svelte
AntonioVentilii-DFINITY: it was fixed on the feature branch (PR #1863), so it will be fixed once merged

src/frontend/src/lib/components/tokens/TokenCard.svelte
peterpeterparker: I assume that showing "ck USD" is not accurate in term of accessibility.

AntonioVentilii-DFINITY: included in last commit, have a look please

AntonioVentilii-DFINITY:

src/frontend/src/env/networks.icrc.env.ts
```diff
@@ -414,7 +414,15 @@ export const ICRC_TOKENS: IcCkInterface[] = [
 	...(nonNullish(CKWBTC_IC_DATA) ? [CKWBTC_IC_DATA] : []),
 	...(nonNullish(CKUSDT_IC_DATA) ? [CKUSDT_IC_DATA] : []),
 	...(nonNullish(CKWSTETH_IC_DATA) ? [CKWSTETH_IC_DATA] : [])
-];
+].map((token) => ({
```
peterpeterparker, line 417: It might be a bit more verbose, but I feel like it would be cleaner not to apply the map here. By adding a map, we are mixing building a list of defined tokens with applying some logic, while the former is less prone to error and more portable.

src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -2,12 +2,23 @@
 	import Card from '$lib/components/ui/Card.svelte';
 	import TokenLogo from '$lib/components/tokens/TokenLogo.svelte';
 	import type { Token } from '$lib/types/token';
+	import Tag from '$lib/components/ui/Tag.svelte';
+	import { nonNullish } from '@dfinity/utils';
 
 	export let token: Token;
+
+	const ariaLabel = nonNullish(token.displayName)
```
peterpeterparker, line 10: isn't the `ariaLabel` just equals to the `name`?

src/frontend/src/lib/components/ui/Tag.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+	export let ariaHidden = false;
+</script>
+
+<span
+	aria-hidden={ariaHidden ? 'true' : 'false'}
```
peterpeterparker, line 6: Isn't Svelte mapping boolean to string?

`aria-hidden={true}` isn't parsed as `aria-hidden="true"` in the DOM?

src/frontend/src/lib/components/tokens/TokenCard.svelte
AntonioVentilii-DFINITY: ah! it really depends on what we would like: as I have written it, if we display for example `<Tag>ck</Tag> Tether USD`, the arialabel is "ckTether USD", otherwise it would be "ckUSDT"

what do you think? what is better?

src/frontend/src/lib/components/ui/Tag.svelte
AntonioVentilii-DFINITY: right! changed

src/frontend/src/env/networks.icrc.env.ts
AntonioVentilii-DFINITY: Yes, it makes sense. In the last commit I moved it to a specific mapping function outside that list

src/frontend/src/lib/components/tokens/TokenCard.svelte
peterpeterparker: I was thinking that we should someday actually use zod for the types and the displayName.prefix + displayName.name === token.name

src/frontend/src/lib/types/token.ts
```diff
@@ -20,7 +22,16 @@ export interface TokenMetadata {
...
+export interface TokenAppearance {
+	displayName?: TokenDisplayName;
+}
+
+export type TokenDisplayName = {
+	prefix: string | undefined;
+	name: string;
+};
+
+export type RequiredToken = RequiredExcept<Token, keyof TokenAppearance>;
```
peterpeterparker, line 34: Could we implement standard TypeScript in this pull request by using constructs like `Omit<Token, "displayName">`, and then, in a subsequent PR, doing a refactoring and introducing TypeScript utilities?

i.e. implementing the feature first and then refactoring, rather than doing everything within the same PR, given that we have already iterated on this solution multiple times.

AntonioVentilii-DFINITY: ah yes, sure!

src/frontend/src/lib/components/tokens/TokenCard.svelte
AntonioVentilii-DFINITY: eh, but in the ck-tokens' case it will be `displayName.prefix + displayName.name !== token.name` for now

P.S. agreed on Zod, but I imagine this is another refactoring in a separate PR

peterpeterparker: > displayName.prefix + displayName.name !== token.name

Why? The token name is the token name, period. This is given by its related ledger. Really not sure what we are doing here then?

> agreed on Zod

Definitely for the future, there is even a Jira task I think.

AntonioVentilii-DFINITY: Basically they asked that for ck-tokens, instead of having the normal name, that is "ckXXXX", they want ck + the name of the twin token. For example, for ckUSDT, instead of showing "ckUSDT" they want to see "ck" + "Tether USD", for "ckBTC" show "ck Bitcoin" (or "ckBitcoin")

At least, this is the last version they agreed upon.

Plus they liked the "ck" to be in a tag format

peterpeterparker: so here the name is `ckLINK`.

peterpeterparker: In other words, if one want to have a different name, then a proposal should be made to update the tokens name. 

Like you would not display "dallar" instead of "dollar" because all other platforms in the word display "dollar".

AntonioVentilii-DFINITY: yes, I think they are looking into that, but they are not sure they want to change the official name of the token. So, for now, they saw the mockup of this new style and they liked it, even with the wrong name. But as you said, they will probably change it again later on. I think they want to see iterations of it

src/frontend/src/lib/types/token.ts
```diff
@@ -20,7 +21,16 @@ export interface TokenMetadata {
 	icon?: string;
 }
 
-export type RequiredToken = Required<Token>;
+export interface TokenAppearance {
+	displayName?: TokenDisplayName;
+}
+
+export type TokenDisplayName = {
```
peterpeterparker, line 28: Let's rename this to emphasize that it's a custom choice made by the foundation. Without being to explicity, something like:

```
export type TokenCustomName = {
        prefix: string | undefined;
	customName: string;
};
```

or

```
export type TokenOisyName = {
        prefix: string | undefined;
	oisyName: string;
};
```

i.e. something that makes it obvious it's strictly for some mumbo jumbo

AntonioVentilii-DFINITY: i like the second one, it shows it is really specific to oisy

src/frontend/src/env/tokens.env.ts
```diff
@@ -61,7 +61,7 @@ export const ICP_SYMBOL = 'ICP';
 
 export const ICP_TOKEN_ID = Symbol(ICP_SYMBOL);
 
-export const ICP_TOKEN: Required<IcToken> = {
+export const ICP_TOKEN: Required<Omit<IcToken, keyof TokenAppearance>> & TokenAppearance = {
```
peterpeterparker, line 64: cf `Required<Omit<IcToken, keyof TokenAppearance>>` === nice 👍

You need `& TokenAppearance` here?

peterpeterparker (approved):
Technically looks good. Thx.

src/frontend/src/env/tokens.env.ts
AntonioVentilii-DFINITY: not really, but I would like to have the same "format" in all of the changes i did, so that it is easier to catch them all, once i do the refactoring

AntonioVentilii-DFINITY: - PR #1985

src/frontend/src/lib/types/token.ts
AntonioVentilii-DFINITY: - PR #1985

PR #1966, opened 2024-08-05
feat(frontend): include roadmap and ticket report in Menu

AntonioVentilii-DFINITY:
# Motivation

We include the links to the Oisy Roadmap and the Issue report into the `Menu` component. Furthermore, we re-arrange the order so that the source code is after the new items.


peterpeterparker (approved):
LGTM, thx

PR #1967, opened 2024-08-05
feat(backend): User data migration

bitdivine:
# Motivation
We would like to migrate user data to a new canister.  The groundwork for the migration code is already in place.  Now we need the actual migration code.

# Changes
- Populate the skeleton code for migrating data.
- Populate the skeleton e2e test.

# Tests
A `pocket-ic` test is included that exercises the migration from end to end.  The main difference from production code is that the `pocket-ic` code steps explicitly whereas normally the migration is driven by a timer.


src/backend/src/migrate.rs
```diff
@@ -67,8 +77,110 @@ pub fn bulk_up(data: &[u8]) {
...
+}
+
+/// Migrates a chunk of data.
+///
+/// # Returns
+/// The updated progress.
+///
+/// # Errors
+/// - Throws a `MigrationError::DataMigrationFailed` if the data transfer fails.
+macro_rules! migrate {
```
frederikrothenberger, line 163: Interesting. I would have approached that using generics, but this is probably cleaner. 👍

frederikrothenberger (approved):
LGTM!

src/backend/src/migrate.rs
bitdivine: I think generics are easier to debug than macros and more familiar to most people.  So I tend to avoid macros where possible.  But I am not sure it is possible to pass say `$chunk_variant` in a generic function, so I _think_ this is the only option.  I didn't spend a long time deciding though; I could be wrong.

PR #1970, opened 2024-08-06
feat(frontend): it is now possible to hide the logo name

AntonioVentilii-DFINITY:
# Motivation

We would like for the name in the main logo to disappear when the user is logged in and in a small screen.

# Changes

- Add property `hideName` to Oisy logo component.
- Use such property in the `Header`.
- Change the part of the logo that is the name to be dependent on `media-width` and `hideName`.
- Unrelated: David Aerne helped and simplified the logo component, removing redundant definitions and grouping the elements that have the same `fill` property.

# Tests

https://github.com/user-attachments/assets/54fe8054-5597-4a82-9034-362e0ba8cc50


src/frontend/src/lib/components/hero/Header.svelte
```diff
@@ -14,8 +14,17 @@
 
 	let back = false;
 	$: back = isSubRoute($page);
+
+	let hideName = false;
+	$: $authSignedIn, updateHideName();
+
+	const updateHideName = () => {
+		hideName = window.innerWidth < 768 && $authSignedIn;
```
peterpeterparker, line 22: Let's avoid the use of JavaScript and hardcoded size for display purpose. Please resolve this with CSS.

src/frontend/src/lib/components/hero/Header.svelte
```diff
@@ -14,8 +14,17 @@
 
 	let back = false;
 	$: back = isSubRoute($page);
+
+	let hideName = false;
+	$: $authSignedIn, updateHideName();
```
peterpeterparker, line 19: We can use css. If signedIn this class, else this class.

AntonioVentilii-DFINITY:
@peterpeterparker I was having trouble doing with just CSS, that is why I ended up using `window`.

However, David Aerne helped in cleaning the logo up, so now it is possible to use directly CSS.

Please review

src/frontend/src/lib/components/icons/OisyWalletLogo.svelte
```diff
@@ -1,1033 +1,51 @@
...
-		/>
-		<path d="M80.6306 12.8563H87.7606V31.1438H80.6306V12.8563Z" fill="white" />
-		<path
-			d="M66.5741 12.4072C73.4592 12.4072 79.0515 14.421 79.0515 21.3333V22.6667C79.0787 29.579 73.4455 31.5792 66.5741 31.5928C59.6891 31.5792 54.0559 29.579 54.0831 22.6667V21.3333C54.0559 14.4074 59.6891 12.4208 66.5741 12.4072ZM71.9216 22.2177V21.7687C71.9624 18.7888 69.486 18.2445 66.5741 18.2037C63.6487 18.2445 61.1722 18.7888 61.2131 21.7687V22.2177C61.1722 25.2112 63.6487 25.7555 66.5741 25.7963C69.486 25.7555 71.9624 25.2112 71.9216 22.2177Z"
-			fill="white"
-		/>
-		<path
-			fill-rule="evenodd"
+<script lang="ts">
+	export let hideName = false;
```
peterpeterparker, line 3: suggestion: rename to `hideTagline` maybe? We don't really hide the name but rather the "selling phrase".

peterpeterparker (approved):
LGTM, thanks

Minor suggestion.

src/frontend/src/lib/components/icons/OisyWalletLogo.svelte
AntonioVentilii-DFINITY: well, we are hiding both:

1) when logged out: we hide only the "selling phrase"
2) when logged in: we hide the "selling phrase" and the Oisy name; only the logo is left

maybe invert the varibale and call it `onlyLogo` or something similar?

PR #1971, opened 2024-08-06
fix(frontend): Send modal will provide info if there are no tokens in the list

AntonioVentilii-DFINITY:
# Motivation

The `Send` modal was providing no feedback when there are no tokens with a non-zero balance, so that the list is empty.

### DIFF
--- a/src/frontend/src/lib/components/send/SendTokensList.svelte
+++ b/src/frontend/src/lib/components/send/SendTokensList.svelte
@@ -26,6 +26,10 @@
 	{/each}
 </TokensSkeletons>
 
+{#if tokens.length === 0}
+	<p class="mt-4 text-dark opacity-50">{$i18n.tokens.manage.text.all_tokens_zero_balance}</p>
+{/if}
+
 <button class="secondary full center text-center" on:click={modalStore.close}>
 	{$i18n.core.text.close}
 </button>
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/send/SendTokensList.svelte
```diff
@@ -26,6 +26,10 @@
 	{/each}
 </TokensSkeletons>
 
+{#if tokens.length === 0}
+	<p class="mt-4 text-dark opacity-50">{$i18n.tokens.manage.text.all_tokens_zero_balance}</p>
```
peterpeterparker, line 30: Can you provide a screenshot? Is that visible? Do we have other text that are displayed with an `opacity-50` or that's an on-purpose choice?

AntonioVentilii-DFINITY: Sure, I included it in the comment (and I included a bit of margin at the bottom).

As per the style I am using the same as the `TokensSignedIn` list: the warning appear when the user hides zero balances and all the tokens have null balance.

peterpeterparker: Gotcha! Thanks

peterpeterparker (approved):
LGTM, thanks

PR #1973, opened 2024-08-06
refactor(backend): Use builder pattern for test env setup

bitdivine:
# Motivation
There are currently two functions for creating a test environment, called `setup_()` and `setup_with_custom_wasm()`.  For tests with the new separate signer canister I find the need for more variants.  Rather than creating many variants, it makes sense to create one builder.

# Changes
- Replace the test environment constructors with a builder.

# Tests
- See existing CI


src/backend/tests/it/upgrade/credentials_init_args.rs
```diff
@@ -1,7 +1,7 @@
 use crate::upgrade::constants::BACKEND_V0_0_25_WASM_PATH;
 use crate::upgrade::types::{ArgV0_0_25, InitArgV0_0_25};
```
bitdivine, line 2: Bulk change: Change calls to `setup*` to calls to the builder instead.

src/backend/tests/it/utils/pocketic.rs
```diff
@@ -15,35 +15,183 @@ const BACKEND_WASM: &str = "../../target/wasm32-unknown-unknown/release/backend.
 // Instead, we can use the master_ecdsa_public_key suffixed with the subnet ID. PocketID adds the suffix because it can have multiple subnets.
 const SUBNET_ID: &str = "fscpm-uiaaa-aaaaa-aaaap-yai";
```
bitdivine, line 17: This is the builder, the main contribution of this PR.

src/backend/tests/it/utils/pocketic.rs
```diff
@@ -68,6 +216,8 @@ pub fn upgrade_with_wasm(
 
     let arg = encoded_arg.unwrap_or(encode_one(&init_arg()).unwrap());
 
+    pic.advance_time(Duration::from_secs(100_000));
```
bitdivine, line 219: If we redeploy frequently without advancing the clock, `PocketIc` starts throwing errors about too many cycles being used.  I put this here rather than at the end of the builder deployment because (in theory) a canister can be redeployed many times.  In current tests we redeploy only once, as far as I can see, but this location seems safer.

src/backend/tests/it/upgrade/credentials_init_args.rs
```diff
@@ -15,7 +15,10 @@ fn test_upgrade_credential_init_args() {
         allowed_callers: allowed_callers.clone(),
     });
     let encoded_initial_arg = encode_one(initial_arg).unwrap();
-    let pic_setup = setup_with_custom_wasm(BACKEND_V0_0_25_WASM_PATH, Some(encoded_initial_arg));
+    let pic_setup = BackendBuilder::default()
+        .with_wasm(BACKEND_V0_0_25_WASM_PATH)
+        .with_arg(encoded_initial_arg)
```
AntonioVentilii-DFINITY, line 20: I imagine you already checked, but here it should not be optional, right? `Some(encoded_initial_arg)`

AntonioVentilii-DFINITY:
Is it worth to separate 

```rust
let pic_setup = BackendBuilder::default()
        .with_wasm(BACKEND_V0_0_19_WASM_PATH)
        .deploy();

```

to avoid repetition of code? or not really?

bitdivine:
@AntonioVentilii-DFINITY Do you mean, have a dedicated function for that?

src/backend/tests/it/utils/pocketic.rs
frederikrothenberger: How was the value chosen? Would 1 second also be enough?
Maybe leave a comment in the code as well, so that one does not need to dig up the PR to get to this insight. 😉

frederikrothenberger (approved):
LGTM, one small comment.

src/backend/tests/it/utils/pocketic.rs
bitdivine: I could not find the throttling period quickly so I chose a big number (this is just over a day).  I'll put a comment in saying that this is wildly conservative and can be reduced.

PR #1974, opened 2024-08-07
test(build): frontend reproducibility

peterpeterparker:
# Motivation

Reproducible builds are important if Oisy Wallet ever gets decentralized. That way, anyone can verify that a proposed version matches the content of the proposal and, by extension, what will be executed.

# Changes

Similarly to the NNS dapp, provide a script to test the reproducibility of the frontend builds (on the same machine) by running `npm run build` consecutively and comparing the list of files generated, their names, and their SHA values.

In addition, as in Juno, provide a GitHub Action to execute the script on demand or during a release.


AntonioVentilii-DFINITY (approved):
LGTM!

For my info, is it intentional to not provide an extension to the script file? For example ".sh"

peterpeterparker:
> For my info, is it intentional to not provide an extension to the script file? For example ".sh"

You are right it's more consistent to append the extension given that all our scripts do so. Thanks.

scripts/test.frontend.reproducibility.sh
```diff
@@ -0,0 +1,33 @@
...
+# Exit if a difference is detected.
+
+DIR=tmp
+
+if [ ! -d "$DIR" ]; then
+  mkdir "$DIR"
+fi
+
+for ((n = 0; n < 10; n++)); do
+  npm run build && find build -type f -exec sha256sum {} \; | awk '{print $2 " " $1}' | sort >"$DIR/sha256-batch${n}.txt"
```
bitdivine, line 13: Note that building several times in a row does not guarantee reproducibility.  For example, at one time Java used an implementation of HashMap that used a random seed set on boot.  This was a defence against targeted denial of service attacks that relied on knowing the HashMap buckets.  So as  long as a server was not rebooted, the elements of a hash map would be in the same order.  Reboot and they would be completely different.

This is still valuable, please just be aware that we might still have problems! :construction_worker:

bitdivine (approved):
Looks useful, thanks!

.github/workflows/frontend-reproducibility.yaml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/checkout@v4
+
+      - name: Prepare
+        uses: ./.github/actions/prepare
+
+      - name: Run reproducibility test
+        run: bash ./scripts/test.frontend.reproducibility.sh
+        shell: bash
+
+      - name: Check script exit status
```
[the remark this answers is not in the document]
peterpeterparker: Oh good idea! Done. Let me know if the implementation is correct now?

scripts/test.frontend.reproducibility.sh
peterpeterparker: I hope we do not have a random seed hardcoded in the frontend 😉

Noted

PR #1975, opened 2024-08-07
fix(build): frontend reproducibility

peterpeterparker:
# Motivation

By default, SvelteKit uses an environment variable `version`, assigned to the build timestamp, to ensure that the client always receives the latest chunks when a new version is deployed. This field is used for chunking purposes.

However, since we are particularly interested in reproducibility, this option undermines reproducibility.

Therefore, instead of using the default timestamp, we set the version field to the actual version of the release, as specified in `package.json`, which is constant for a same build but, different accross releases.

# Changes

- There was a misplacement of the `version` field in the Svelte configuration within `svelte.config`, which caused the build to ignore our settings.


AntonioVentilii-DFINITY (approved):

bitdivine:
Is the version field change every time you deploy?  I am worried about situations such as deploying repeatedly to localhost.

Some ideas:
- Use the git hash
- Use the git tree hash of `src/frontend` (this is basically the hash of the files and ignores the git history, so if there is say a  commit that updates the README, the tree hash is not updated).  This could potentially help avoid unnecessary rebuilds.
  ```
  git rev-parse HEAD:src/frontend
  ```
- Use the git hash IF git is clean, else the timestamp.  Otherwise updates in a dirty repo may not produce a clean rebuild.

bitdivine (approved):
Fix looks good to me.  Thanks!

peterpeterparker:
> Is the version field changed every time you deploy?

No. The `version` is set to the static version set in `package.json` - For example `v0.0.32`.

> use the git hash

We could potentially do this if we are interested to check reproducibility on every PR but, in this proposition, the idea is to check the reproducibility on release assuming that development does not alter this feature. Therefore, at least for a first iteration, I guess that using the version number is correct. Agree?

PR #1978, opened 2024-08-07
refactor(frontend): create component ReceiveButtonWithModal

AntonioVentilii-DFINITY:
# Motivation

We create a new component `ReceiveButtonWithModal` that incorporated the ReceiveButton with any specific modal required.

# Changes

- Create new component `ReceiveButtonWithModal` with a slot for the modal. Among the props there is the possibility of providing a specific `modalId` if required.
- Adapted usage throughout the code.

# Tests

Manual test on local replica showed no changes.


src/frontend/src/lib/components/receive/ReceiveButtonWithModal.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
+	import { modalStore } from '$lib/stores/modal.store';
+	import ReceiveButton from '$lib/components/receive/ReceiveButton.svelte';
+
+	export let openModal: (modalId: symbol) => void | Promise<void>;
```
peterpeterparker, line 5: Few naming ideas which feel like it could make things less verbose but, just sharing in case you like it.

```
export let openModal => export let open
export let specificModalId => export let modalId

const modalId => const definedModalId (or anything else)
```

src/frontend/src/lib/components/receive/ReceiveButtonWithModal.svelte
```diff
@@ -0,0 +1,16 @@
...
+	export let isOpen: boolean;
+	export let specificModalId: symbol | undefined = undefined;
+
+	const modalId = specificModalId ?? Symbol();
```
peterpeterparker, line 9: Given that the property is reactive, it's probably safer to make this assignement reactive as well.

peterpeterparker (commented):
Two suggestions. Feel free to ignore the naming ideas, the second feels like it makes more sense. WDYT?

src/frontend/src/lib/components/receive/ReceiveButtonWithModal.svelte
AntonioVentilii-DFINITY: yes please! i even asked chatGPT for better names but nothing good

I will use your suggestions

src/frontend/src/lib/components/receive/ReceiveButtonWithModal.svelte
AntonioVentilii-DFINITY: right!

peterpeterparker (approved):
LGTM, thx

PR #1980, opened 2024-08-07
feat(backend): Private stats API

bitdivine:
# Motivation
When migrating user data, it would be useful to be able to see how many users and other data is in the old and the new backend canister.

Data such as the number of users is not currently public, as far as I can tell, so I have kept this data restricted to "allowed" users.

# Changes
- Add a private `stats()` query endpoint to support checking migration progress.

# Tests
- `PocketIc` tests are included to ensure that the stats endpoint gives expected values and is accessible only to authorized users.


src/backend/src/lib.rs
```diff
@@ -112,6 +112,16 @@ pub struct State {
     user_profile_updated: UserProfileUpdatedMap,
 }
 
+impl From<&State> for Stats {
```
peterpeterparker, line 115: Should we move this implementation into `impls` module?

peterpeterparker (approved):
LGTM, thanks

Maybe in another PR, some day, we can refactor the test code to move the information that have been set to public to separate modules. Just an idea.

src/backend/src/lib.rs
bitdivine: Done

PR #1981, opened 2024-08-07
feat(backend): API to change guards

bitdivine:
# Motivation
To migrate from one backend canister to another, it is necessary to put some canister APIs into read-only mode or to disable them entirely, to guarantee that data is not changed during migration or in the old canister after migration.

# Changes
- Add an API method to change guards.

# Tests
- `PocketIc` tests are included that check whether the guards have the expected new values.


src/backend/tests/it/guard.rs
```diff
@@ -0,0 +1,70 @@
...
+            &pic_setup,
+            Principal::anonymous(),
+            "set_guards",
+            &guards
+        )
+        .is_err(),
+        "Anonymous user should not be able to set guards"
+    );
+    // A principal NOT listed in `BackendConfig::default_controllers()`.
+    const USER_1: &str = "7blps-itamd-lzszp-7lbda-4nngn-fev5u-2jvpn-6y3ap-eunp7-kz57e-fqe";
```
AntonioVentilii-DFINITY, line 31: is it useful to move this to `utils/mock`? if we are going to re-use it elsewhere as variable

AntonioVentilii-DFINITY (approved):
LGTM, tks

src/backend/tests/it/guard.rs
bitdivine: Sounds good.  Done.

PR #1982, opened 2024-08-07
feat(backend): Support multiple backends in tests

bitdivine:
# Motivation
The current tests each create a test environment consisting of the tuple `(PocketIc, Principal)` where the Principal is the backend canister ID.  This test environment is passed around as an argument to various utilities.  This approach has issues as soon as there are two backend canisters.  E.g. ownership of the PocketIc instance prevents us from having two such tuples.  We need a more flexible approach.

# Changes
- Create a struct that can take the place of the existing tuples.
  - That struct uses an `Arc` to allow several structs to share a single `PocketIc`.
- Copy (and adapt) utilities to these new structs.

Note: The bulk update of tests is NOT done in this PR in order to keep the change small and readable.

# Tests
These structs are as yet unused, due to breaking changes into small PRs.  The structs here are used in new tests in a feature branch; the test structures work there and are ergonomic.  A review of existing tests suggests that adapting them to use these structures instead of tuples should be fairly straightforward.


src/backend/tests/it/utils/pocketic.rs
```diff
@@ -309,3 +313,94 @@ where
...
+impl PicCanisterTrait for PicBackend {
+    fn pic(&self) -> Arc<PocketIc> {
+        self.pic.clone()
+    }
+    fn canister_id(&self) -> Principal {
+        self.canister_id.clone()
+    }
+}
+impl PicBackend {
+    /// Creates toy users with the given range of principals.
```
bitdivine, line 331: Note: This is copied (and adjusted) from https://github.com/dfinity/oisy-wallet/blob/cdf4246f4943215deb6eb40c16bf8cabdf5ee8e4/src/backend/tests/it/list_users.rs#L16

src/backend/tests/it/utils/pocketic.rs
```diff
@@ -309,3 +313,94 @@ where
...
+pub trait PicCanisterTrait {
+    /// A shared PocketIc instance.
+    ///
+    /// Note: `PocketIc` uses interior mutability for query and update calls.  No external mut annotation or locks appear to be necessary.
+    fn pic(&self) -> Arc<PocketIc>;
+
+    /// The ID of this canister.
+    fn canister_id(&self) -> Principal;
+
+    /// Makes an update call to the canister.
```
bitdivine, line 365: The query and update calls here are copied (and adjusted) from https://github.com/dfinity/oisy-wallet/blob/cdf4246f4943215deb6eb40c16bf8cabdf5ee8e4/src/backend/tests/it/utils/pocketic.rs#L259

AntonioVentilii-DFINITY (approved):

PR #1983, opened 2024-08-08
feat(lint): remove duplicates from svelte imports

AntonioVentilii-DFINITY:
# Motivation

The package [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import) provides some ESLint rules to format svelte imports. It is among the packages refered to in [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint/blob/f335c504bcf75623d2d671e2e784b047e5e186b9/docs/getting-started/linting/FAQ.md#eslint-plugin-import).

For this iteration we will apply first only the rule [no-duplicates](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md).

# Changes

- Install eslint-plugin-import among dev dependencies.
- Turn on the rule.
- Fix all issues raised by the new rule.

# Tests

The rule worked correctly when ran.


AntonioVentilii-DFINITY:
I am proposing to turn the parameter `prefer-inline` on, so that the type imports are merged with the normal imports. But since I have no strong opinion on it, I am happy to enforce the opposite (types separated from the rest).

❌ Invalid `["error", {"prefer-inline": true}]`

```js
import { CValue } from './papa-mia'
import type { CType } from './papa-mia'
```

✅ Valid with `["error", {"prefer-inline": true}]`

```js
import { CValue, type CType } from './papa-mia'
```

PR #1984, opened 2024-08-08
refactor(frontend): rename potentially risky stores

AntonioVentilii-DFINITY:
# Motivation

As suggestion and to be more aware of the code dependencies, we rename a few stores with the suffix `_POTENTIAL_RE_RENDER` to be conscious that they are mixing stores (tokens+network+balance+exchanges) and which potentially can imply quite some re-render.

The stores/deriveds renamed are:

- `sortedTokens`
- `enabledNetworkTokensUi`
- `sortedNetworkTokensUi`


peterpeterparker (commented):
The issue is not really a potential re-render, as all stores, by definition, re-render.

The problem we faced was an infinite loop of death.

Should we maybe highlight the fact that these stores are composed of multiple stores with various types of information, like `combined` or better naming?

Perhaps we can use naming conventions as follows:

```
src/stores => `somethingStore`
src/dervices => `somethingDerived` and `somethingCombined`
```

WDYT? A bit verbose, better idea welcome.

AntonioVentilii-DFINITY:
I like it, but maybe we can just add the "Combined" and avoid it for stores and for "normal" derived. For store in general, I would avoid: it is even a different type and for sure it cannot be combined. But for normal derived, no strong opinion.

As per naming: "combined" seems a good word (or "aggregated"). What do you think of having it after the "_"? I know it is not proper snake-case, but at least it is visually impactful (and it does not mix with the rest of the name).

peterpeterparker:
> Understood, but maybe we can just add the "Combined" and avoid it for stores and for "normal" derived.

It makes sense, works for me.

>  What do you think of having it after the "_"? I know it is not proper snake-case, but at least it is visually impactful

Not sure a snake-case separator makes that a difference. Personnally would avoid it but, not a strong opiniong

> combined

`somethingCombined` or `combinedSomething`?

AntonioVentilii-DFINITY:
what about a more verbose:

`combinedDerivedSomething`

peterpeterparker:
Sounds good!

peterpeterparker (approved):
LGTM, thanks

PR #1985, opened 2024-08-08
refactor(frontend): create RequiredExcept function for types

AntonioVentilii-DFINITY:
# Motivation

Create utility function `RequiredExcept` to be used for required tokens that still retain the state of specific props..


peterpeterparker (approved):
LGTM, thx

PR #1986, opened 2024-08-08
feat(backend): derive BTC address

AntonioVentilii-DFINITY:
# Motivation

As first step to request the Bitcoin address for the user's Principal, we created a custom call to [endpoint `ecdsa_public_key` of IC Management API](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-ecdsa_public_key).

To generate the addresses with threshold ECDSA, we refer to [IC Bitcoin Documentation](https://internetcomputer.org/docs/current/developer-docs/multi-chain/bitcoin/using-btc/generate-addresses#generating-addresses-with-threshold-ecdsa), using directly the crate [bitcoin](https://crates.io/crates/bitcoin) and its function.

Note: the logic is very similar to how we derive the Ethereum address.

# Changes

- Created util function to transform public key to a P2PKH address.
- Created function to call `ecdsa_public_key` and obtain the caller P2PKH address.

# Tests

1. Created local tests for the new functions.
2. I created a branch in [examples/basic_bitcoin](https://github.com/dfinity/examples/blob/ff95167f625d67494aab0b1bf3181dd28af92e44/rust/basic_bitcoin/src/basic_bitcoin/src/lib.rs#L215) to test the functions created in this PR: the results were the same as the functions that already exists there.


src/backend/src/bitcoin_utils.rs
```diff
@@ -0,0 +1,20 @@
+use bitcoin::{Address, Network, PublicKey};
+use ic_cdk::api::management_canister::bitcoin::BitcoinNetwork;
+
+fn transform_network(network: BitcoinNetwork) -> Network {
```
Sawchord, line 4: NIT: Since this is a 1-1 match, why not implement it as
```rust
impl From<BitcoinNetwork> for Network
```

AntonioVentilii-DFINITY: good suggestion!

AntonioVentilii-DFINITY: ah wait, `Network` is not in my crate, so I cannot `impl` it

AntonioVentilii-DFINITY: however, it could be useful to have this in the `ic_cdk`

```rust
impl BitcoinNetwork {
    pub fn to_network(self) -> Network {
        match self {
            BitcoinNetwork::Mainnet => Network::Bitcoin,
            BitcoinNetwork::Testnet => Network::Testnet,
            BitcoinNetwork::Regtest => Network::Regtest,
        }
    }
}


```

bitdivine (approved):
LGTM, thanks!

PR #1987, opened 2024-08-08
feat(backend): Add migration state

bitdivine:
# Motivation
The user data migration is performed by a timer, in multiple steps.  The timer ID and current progress need to be stored in the canister.

# Changes
- Define a data structure to store the migration state.
- Add an API to read the canister state.
- Tidy: In accordance with the convention of this repository, the `ApiEnabled` struct implementations have been moved into ` impls.rs`

# Tests
None, this data is not used yet.

### DIFF
--- a/src/shared/src/types.rs
+++ b/src/shared/src/types.rs
@@ -1,4 +1,5 @@
 use candid::{CandidType, Deserialize, Principal};
+use ic_cdk_timers::TimerId;
 use std::fmt::Debug;
 
 pub type Timestamp = u64;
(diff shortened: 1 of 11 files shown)


src/shared/src/types.rs
```diff
@@ -35,30 +36,6 @@ pub enum ApiEnabled {
     ReadOnly,
     Disabled,
 }
```
bitdivine, line 38: Moved to `impls.rs` according to the convention of this repository.

src/shared/src/types.rs
```diff
@@ -261,3 +238,50 @@ pub mod user_profile {
         pub custom_token_count: u64,
     }
 }
+
+/// The current state of progress of a user data migration.
+#[derive(CandidType, Deserialize, Copy, Clone, Eq, PartialEq, Debug, Default)]
+pub enum MigrationProgress {
+    // WARNING: The following are subject to change.  The migration has NOT been implemented yet.
+    /// Migration has been requested.
```
AntonioVentilii-DFINITY, line 246: Maybe add a TODO too here?

AntonioVentilii-DFINITY (approved):

src/shared/src/types.rs
bitdivine: Good idea.  Done

PR #1988, opened 2024-08-08
refactor(backend-tests): Use pic backend struct

bitdivine:
# Motivation
Historically the backend test environment consisted of a tuple `(PocketIc, Principal)`.  This worked well until we needed multiple backend canisters, as is the case now.  In #1982 a structure was introduced to replace the role of that tuple.  Now we need to use it.

# Changes
- Bulk change the pic_setup tuple to the replacement struct.
- Bulk change `canister_query(...)` to `pic_setup.query(...)` and the same for update.
- Miscellaneous other changes required for the switch.

# Tests
- Please see CI


src/backend/tests/it/custom_token.rs
```diff
@@ -1,6 +1,6 @@
 use crate::utils::assertion::{assert_custom_tokens_eq, assert_tokens_data_eq};
 use crate::utils::mock::CALLER;
-use crate::utils::pocketic::{query_call, setup, update_call};
+use crate::utils::pocketic::{setup, PicCanisterTrait};
```
AntonioVentilii-DFINITY, line 3: I cannot find the place in the code where `PicCanisterTrait` is used

src/backend/tests/it/guard.rs
```diff
@@ -5,7 +5,7 @@ use shared::types::{ApiEnabled, Config, Guards};
 
 use crate::utils::{
     mock::USER_1,
-    pocketic::{controller, setup},
+    pocketic::{controller, setup, PicCanisterTrait},
```
AntonioVentilii-DFINITY, line 8: ditto

src/backend/tests/it/address.rs
```diff
@@ -1,5 +1,5 @@
 use crate::utils::mock::{CALLER, CALLER_ETH_ADDRESS};
-use crate::utils::pocketic::{setup, update_call};
+use crate::utils::pocketic::{setup, PicCanisterTrait};
```
AntonioVentilii-DFINITY, line 2: I cannot find in the code the place where `PicCanisterTrait` is used

src/backend/tests/it/config.rs
```diff
@@ -1,4 +1,4 @@
-use crate::utils::pocketic::{controller, init_arg, setup, update_call};
+use crate::utils::pocketic::{controller, init_arg, setup, PicCanisterTrait};
```
AntonioVentilii-DFINITY, line 1: ditto

AntonioVentilii-DFINITY (approved):
LGTM, but only one doubt: I cannot find the place in the code where `PicCanisterTrait` is used (it is imported in all the modules)

bitdivine:
@AntonioVentilii-DFINITY  `PicCanisterTrait` defines methods such as `query()` and `update(..)`.  To use these methods in a file, it is necessary to import the trait that defines them.

bitdivine:
Thank you for the review!

AntonioVentilii-DFINITY:
Ah! thank you for the explanation!

PR #1989, opened 2024-08-08
style(frontend): adjust text style for AboutItem

AntonioVentilii-DFINITY:
# Motivation

When `AboutItem` is part of a menu, it does not behave as standard item in menu (no change in color when hovering for example). This PR adjusts it.


peterpeterparker (approved):
LGTM, thanks

PR #1990, opened 2024-08-08
chore(backend): Ensure that ic_cdk println is used

bitdivine:
# Motivation
Clippy rules can prevent errors, in particular they can prevent `std::println!()` being called accidentally instead of `ic_cdk::println!()`.  The first is black-holed while the second is written to the canister logs.

# Changes
- Add common clippy method restrictions, taken from `nns-dapp`
- Update code to match.

# Tests
See CI


clippy.toml
```diff
@@ -0,0 +1,10 @@
+disallowed-macros = [
+    { path = "std::print", reason = "Use ic_cdk::print instead" },
+    { path = "std::println", reason = "Use ic_cdk::println instead" },
+    { path = "std::eprint", reason = "Use ic_cdk::eprint instead" },
+    { path = "std::eprintln", reason = "Use ic_cdk::eprintln instead" },
+    { path = "std::assert_eq", reason = "Use pretty_assertions::assert_eq instead" },
+]
+disallowed-methods = [
+    { path = "dfn_core::api::print", reason = "Use ic_cdk::print or ic_cdk::println instead" },
+]
```
bitdivine, line 10: The above is taken from the nns-dapp: https://github.com/dfinity/nns-dapp/blob/main/clippy.toml

Sawchord (approved):
LGTM

PR #1991, opened 2024-08-08
refactor(frontend): create Link component

AntonioVentilii-DFINITY:
# Motivation

For readability and to simplify the code, we create `Link` component. Furthermore it is useful for `a11y` reasons. Note that we provide the prop `href` possibly as `null` in case the usage is by design without URL.

# Tests

Local replica behaves as expected as usual.

### DIFF
--- a/src/frontend/src/lib/components/core/Menu.svelte
+++ b/src/frontend/src/lib/components/core/Menu.svelte
@@ -11,13 +11,13 @@
 	import IconSettings from '$lib/components/icons/IconSettings.svelte';
 	import IconGitHub from '$lib/components/icons/IconGitHub.svelte';
 	import { i18n } from '$lib/stores/i18n.store';
-	import ButtonMenu from '$lib/components/ui/ButtonMenu.svelte';
 	import ButtonHero from '$lib/components/ui/ButtonHero.svelte';
 	import MenuWallet from '$lib/components/core/MenuWallet.svelte';
 	import { page } from '$app/stores';
 	import AboutHow from '$lib/components/hero/about/AboutHow.svelte';
 	import AboutWhat from '$lib/components/hero/about/AboutWhat.svelte';
 	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils.js';
+	import Link from '$lib/components/ui/Link.svelte';
 
 	let visible = false;
 	let button: HTMLButtonElement | undefined;
@@ -47,10 +47,12 @@
 		{/if}
 
 		{#if !settingsRoute}
-			<ButtonMenu ariaLabel={$i18n.navigation.alt.more_settings} on:click={gotoSettings}>
-				<IconSettings />
+			<!-- If the url is passed as href instead of click event, the pages refreshes. -->
+			<!-- After a few tries, it seems it is a behavior that happens only inside the popover. -->
+			<Link href={null} ariaLabel={$i18n.navigation.alt.more_settings} on:click={gotoSettings}>
+				<IconSettings slot="icon" />
 				{$i18n.settings.text.title}
-			</ButtonMenu>
+			</Link>
 
 			<Hr />
 		{/if}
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -47,10 +47,12 @@
 		{/if}
 
 		{#if !settingsRoute}
-			<ButtonMenu ariaLabel={$i18n.navigation.alt.more_settings} on:click={gotoSettings}>
-				<IconSettings />
+			<!-- If the URL is passed as href instead of click event, the pages refreshes. -->
+			<!-- After a few tries, it seems it is a behavior that happens only inside the popover. -->
+			<Link href={null} ariaLabel={$i18n.navigation.alt.more_settings} on:click={gotoSettings}>
```
peterpeterparker, line 52: That's an invalid HTML behavior I think.

From the HTML 5 [standard](https://html.spec.whatwg.org/#the-a-element)

> If the [a](https://html.spec.whatwg.org/#the-a-element) element has no [href](https://html.spec.whatwg.org/#attr-hyperlink-href) attribute, then the element [represents](https://html.spec.whatwg.org/#represents) a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element's contents.

src/frontend/src/lib/components/ui/Link.svelte
```diff
@@ -0,0 +1,46 @@
...
+	export let ariaLabel: string;
+	export let external = false;
+	export let iconVisible = true;
+	export let inline = false;
+	export let color: 'blue' | 'inherit' = 'inherit';
+	export let fullWidth = false;
+
+	const dispatch = createEventDispatcher();
+
+	// Custom click handler to guarantee that it prevents default browser behaviour (full page reload)
```
peterpeterparker, line 15: Isn't using `click` event on a HTML `a` element considered as a bad practice?

From MDN [documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a):

> Use a [<button>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) instead. **In general, you should only use a hyperlink for navigation to a real URL.**

Also isn't this hack having for side effect to push twice the targeted page in the history? once with the href and once with the click?

I think it's cleaner and safer to use link strictly for navigation and not try to hijack browser and web API behavior.

AntonioVentilii-DFINITY: The issue is that if we use it for internal navigation, and inside the popover, it will always refresh the page. remember you had said once that it should be `<a>`, but maybe in that case I shall leave a `<button>` element just for the settings menu item?

peterpeterparker: Yeah a link but without click 😅

peterpeterparker: Would using a link and closing the popover on mount in settings be a not to dirty option? Just thinking out loud

peterpeterparker: Or a `beforeNavigate`

AntonioVentilii-DFINITY: I am using `beforeNavigate` in the last commit, but the issue persist: it is refreshing the page when I click on it, because inside the popover the links are treated as native behavior, re-routing the page, and refreshing everything.

Shall I leave the `Settings` item as button, instead of link?

peterpeterparker: Unless we have a web standard compleant solution, I would keep the button yes.

AntonioVentilii-DFINITY: It was just for `a11y` reasons that you mentioned one. I will close this one

AntonioVentilii-DFINITY:
Since it is not possible to avoid refreshing the page inside the popover, I will close this PR

PR #1993, opened 2024-08-09
refactor(frontend): rename type to OptionEthAddress

AntonioVentilii-DFINITY:
# Motivation

To be more consistent (and expecting to integrate BTC too), we rename type `OptionAddress` to `OptionEthAddress`.


peterpeterparker (approved):
Oh good idea! Thanks

PR #1995, opened 2024-08-09
feat(frontend): create new type for all non-IC addresses

AntonioVentilii-DFINITY:
# Motivation

Since we would like to onboard further networks (starting from Bitcoin), we create a new type that collects all the derived addresses, that for now it is just the ETH one.
The idea is that they should ALL be checked at the same way to be certified or not and the behavior shall be the same if even one is not coherent.

# Changes

- Create new type `Address` with tokens as props.
- Adapt code to usage.

# Tests

No changes in local replica.


peterpeterparker (commented):
The idea is to have a single endpoint in the backend that derives all addresses or will the backend expose multiple endpoints? In other words will the frontend do one query+update call or multiple calls of those types to load the addresses?

AntonioVentilii-DFINITY:
At first the backend will expose multiple endpoints, but the frontend will call them sequentially, and not in parallel, because I think we prefer that if even one of them fails, we do not proceed, correct?

- Fetch ETH address --> fails --> do not fetch BTC address && process failed
- Fetch ETH address --> success --> fetch BTC address --> fails --> process failed
- Etcetera for future other derivations

For sure in the backend at some point we can do a single call for all of them, since there is even a small repetition of code, but for now I would like to separate BTC and ETH, just in case there is a need to change something on BTC.

Performance-wise, it slows down the frontend, but just for now: once we finish with the BTC feature, if I can merge the calls to the addresses I will do it

peterpeterparker:
> At first the backend will expose multiple endpoints

In that case this PR does not work. We need to know for each call if it was made as a query or update call. By modifying the state with an object but keep a single certified flag, we might ends with untrustable information.

AntonioVentilii-DFINITY:
you mean to have separate certified flags for each address, right? I can adapt it to that then, create an object `Address` with the certified flag (as it was before) and then `Addresses` becomes a collection of several Address types

WDYT?

peterpeterparker:
If there are multiple endpoints with related information, it is important not to combine all the data into a single entity stored together. Doing so could lead to data desynchronization. Instead, each address and its certification state should be maintained in a separate state when fetching data from multiple endpoints. To consolidate these addresses into a single structure, we can use a derived state.

AntonioVentilii-DFINITY:
Makes sense, I will work in that way then and close this one

PR #1996, opened 2024-08-09
feat(frontend): gather some anonymous analytics about transactions duration

peterpeterparker:
# Motivation

We discussed with @DSharifi if there was an easy way to gather analytics and insights about the duration of processes as perceived by users. As a result, Daniel had the idea to persist the duration of function executions as anonymous metadata in the events we save in Juno analytics. Basically could potentially be a quick win. This PR provides some of the first metrics.

# Changes

- Add two new functions in the analytics modules to initiate events and ultimately submit those once a function is executed
- Track those kind of events for ETH, IC, ICRC and ERC20 transactions


src/frontend/src/icp/components/send/IcSendTokenWizard.svelte
```diff
@@ -107,35 +129,41 @@
...
+			await Promise.allSettled([
+				trackTimedEventSuccess(timedEvent),
+				trackEvent({
+					name: isNetworkIdBitcoin(networkId)
+						? TRACK_COUNT_CONVERT_CKBTC_TO_BTC_SUCCESS
+						: isNetworkIdETH(networkId) && isTokenCkEthLedger($token)
+							? TRACK_COUNT_CONVERT_CKETH_TO_ETH_SUCCESS
+							: isNetworkIdETH(networkId) && isTokenCkErc20Ledger($token)
+								? TRACK_COUNT_CONVERT_CKERC20_TO_ERC20_SUCCESS
+								: TRACK_COUNT_IC_SEND_SUCCESS,
```
AntonioVentilii-DFINITY, line 141: wouldn't it more secure to check if the netwoork is IC as final condition? I know it does not happen, but imagine the network is ETH and it's neither ckETH ledger nor ckERC20 ledger.

peterpeterparker: I would not say "secure" given those are "anonymous analytics hints" but, yeah sure we can totally skip the events if it's an unknown analytics tracked network. I can do that in another PR.

AntonioVentilii-DFINITY (approved):
LGTM!

DSharifi:
Thanks for the quick work! @peterpeterparker

PR #1997, opened 2024-08-09
refactor(frontend): rename derived 'address' to 'ethAddress'

AntonioVentilii-DFINITY:
# Motivation

To be more consistent with the scope of the derived store, we rename it `ethAddress`.


peterpeterparker (approved):
LGTM, thanks

PR #2001, opened 2024-08-12
fix(frontend): signed out transactions text

peterpeterparker:
# Motivation

Only a subset of the tokens is loaded by default when the user is signed out. As a result, the text that should be displayed on the transactions page when not logged in is not necessarily displayed if the token in the URL is not one of those.

# Changes

- Display a generic non token related text.

# Tests

Capture d’écran 2024-08-12 à 06 50 13


AntonioVentilii-DFINITY (approved):
Good catch!

PR #2002, opened 2024-08-12
style(frontend): reduce skeleton cards in SendManage modal

AntonioVentilii-DFINITY:
# Motivation

During load, the `SendModal` was showing too many skeletons cards in Chain Fusion.

# Results

Screenshot 2024-08-12 at 08 50 41
Screenshot 2024-08-12 at 08 50 56


src/frontend/src/lib/components/tokens/TokensSignedIn.svelte
```diff
@@ -23,7 +24,7 @@
 	);
 </script>
 
-<TokensSkeletons>
+<TokensSkeletons rows={ERC20_CONTRACTS.length + 1}>
```
peterpeterparker, line 27: Actually I would set a smaller value. Maybe 3 or 5 too.

See for example this current screenshot of the loading experience:

Capture d’écran 2024-08-12 à 08 58 46

WDYT?

AntonioVentilii-DFINITY: ah yes, for sure, then I just default both of them to 5, it seems a good middle ground for both


Screenshot 2024-08-12 at 09 12 09
Screenshot 2024-08-12 at 09 12 24
Screenshot 2024-08-12 at 09 12 44
Screenshot 2024-08-12 at 09 12 53

peterpeterparker: Sounds good

AntonioVentilii-DFINITY: ready for another review

peterpeterparker (approved):
LGTM, thanks

PR #2007, opened 2024-08-12
feat(frontend): create specific BitcoinNetwork type

AntonioVentilii-DFINITY:
# Motivation

We create a specific type for the Bitcoin network, to include the environment corresponding to the backend mapping.

### DIFF
--- a/src/frontend/src/btc/types/network.ts
+++ b/src/frontend/src/btc/types/network.ts
@@ -1,0 +1,8 @@
+import type { BitcoinNetwork as BitcoinNetworkBackend } from '$declarations/backend/backend.did';
+import type { Network } from '$lib/types/network';
+
+interface BitcoinBackendData {
+	backendEnv: BitcoinNetworkBackend;
+}
+
+export type BitcoinNetwork = Network & BitcoinBackendData;
--- a/src/frontend/src/btc/types/network.ts
+++ b/src/frontend/src/btc/types/network.ts
@@ -1,5 +1,6 @@
-import type { BitcoinNetwork as BitcoinNetworkBackend } from '$declarations/backend/backend.did';
 import type { Network } from '$lib/types/network';
+
+type BitcoinNetworkBackend = { mainnet: null } | { regtest: null } | { testnet: null };
 
 interface BitcoinBackendData {
 	backendEnv: BitcoinNetworkBackend;
--- a/src/frontend/src/env/networks.env.ts
+++ b/src/frontend/src/env/networks.env.ts
@@ -1,3 +1,4 @@
+import type { BitcoinNetwork } from '$btc/types/network';
 import { ETHEREUM_EXPLORER_URL, SEPOLIA_EXPLORER_URL } from '$env/explorers.env';
 import { ETH_MAINNET_ENABLED } from '$env/networks.eth.env';
 import sepolia from '$eth/assets/sepolia.svg';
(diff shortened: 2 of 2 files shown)


src/frontend/src/btc/types/network.ts
```diff
@@ -0,0 +1,8 @@
+import type { BitcoinNetwork as BitcoinNetworkBackend } from '$declarations/backend/backend.did';
+import type { Network } from '$lib/types/network';
+
+interface BitcoinBackendData {
+	backendEnv: BitcoinNetworkBackend;
```
peterpeterparker, line 5: Should we declare a type and avoid relying on Candid variant for the UI?

Those variants are not really handy to manipulate when used for display purpose in components. Also we do not have yet union on candid types at the token/network level.

WDYT?

AntonioVentilii-DFINITY: yes, for sure, can do that. It will be a copy of the one from the backend, check last commit

peterpeterparker: Not exactly what I meant.

I meant something similar to `export type NetworkEnvironment = 'mainnet' | 'testnet';`

For example `export type BitcoinNetwork = 'mainnet' | 'testnet' | 'test;`

peterpeterparker: Actually, should we use `BitcoinNetwork` from `@dfinity/ckbtc`?

Should we also drop regtest, are we really using regtest?

AntonioVentilii-DFINITY: about regtest: I would not drop it already, because we could run the replica with a local regtest node, if we want. Or maybe I can remove it now and add it if and when I integrate the regtest node in the local replica?

about BitcoinNetwork: I understood now. But then we already have it. The `backendEnv` prop was just to adapt to the backend definition. But I actually thought of a different way to go, so I think I will remove this definition

peterpeterparker: Ok. Actually now I'm unsure what's the goal of `BitcoinBackendData`.

AntonioVentilii-DFINITY: At first I was going to use it to map each frontend `BitcoinNetwork` with the backend respective, to use with the call to the backend that requires the `network` parameter:

```typescript
const getBtcAddress = async ({
	identity,
	network
}: {
	identity: OptionIdentity;
	network: BitcoinNetwork;
}): Promise<EthAddress> => {
	const { caller_btc_address } = await getBackendActor({ identity });
	return caller_btc_address(network);
};
```

 But I changed idea, I am going to create a specific generic call for each network:
 
 ```typescript
export const getBtcAddressMainnet = async (identity: OptionIdentity): Promise<BtcAddress> =>
	getBtcAddress({ identity, network: { mainnet: null } });

export const getBtcAddressTestnet = async (identity: OptionIdentity): Promise<BtcAddress> =>
	getBtcAddress({ identity, network: { testnet: null } });
```

So, we don't need to associate each frontend `BitcoinNetwork` to the corresponding backend `BitcoinNetwork`.

For now it is a draft:
- PR #2013

Comment (AntonioVentilii-DFINITY) [reply src/frontend/src/btc/types/network.ts:6]: So, I will close this PR, since it is enough to use the type Network that already exists. At least for now

Ok?

Comment (peterpeterparker) [reply src/frontend/src/btc/types/network.ts:6]: Let's rather use the first solution and avoid duplicating code. 

```
const getBtcAddress = async ({
	identity,
	network
}: {
	identity: OptionIdentity;
	network: BitcoinNetwork;
}): Promise<EthAddress> => {
	const { caller_btc_address } = await getBackendActor({ identity });
	return caller_btc_address(network);
};
```

then in a service you path the appropriate parameters.

AntonioVentilii-DFINITY:
Closing this one as discussed

PR #2011, opened 2024-08-13
refactor(frontend): generalize idb store creation

AntonioVentilii-DFINITY:
# Motivation

To make it more re-usable, we generalize the function that creates the DB store of IDB.


peterpeterparker (commented):
Can you somehow split this in multiple PRs to ease the review?

peterpeterparker:
Also particularly given the fact that this does not require any UI or replica, should we take the time to first add a test and then apply the refactoring?

AntonioVentilii-DFINITY:
> Can you somehow split this in multiple PRs to ease the review?

Done

AntonioVentilii-DFINITY:
> Also particularly given the fact that this does not require any UI or replica, should we take the time to first add a test and then apply the refactoring?

good idea!

AntonioVentilii-DFINITY:
@peterpeterparker ready for a review from scratch

src/frontend/src/lib/api/idb.api.ts
```diff
@@ -1,13 +1,15 @@
...
 import type { Principal } from '@dfinity/principal';
 import { isNullish } from '@dfinity/utils';
 import { createStore, del, get, set, update, type UseStore } from 'idb-keyval';
 
 // There is no IndexedDB in SSG. Since this initialization occurs at the module's root, SvelteKit would encounter an error during the dapp bundling process, specifically a "ReferenceError [Error]: indexedDB is not defined". Therefore, the object for bundling on NodeJS side.
-const oisyEthAddressesStore = browser
-	? createStore('oisy-eth-addresses', 'eth-addresses')
-	: ({} as unknown as UseStore);
+const oisyAddressesStore = (storeName: string) =>
+	browser ? createStore(IDB_ADDRESS_STORE, storeName) : ({} as unknown as UseStore);
```
peterpeterparker, line 10: Are you sure this work? i.e. having a unique name here but different store name?

src/frontend/src/lib/api/idb.api.ts
```diff
@@ -1,13 +1,15 @@
...
 import type { IdbEthAddress } from '$lib/types/idb';
 import type { Principal } from '@dfinity/principal';
 import { isNullish } from '@dfinity/utils';
 import { createStore, del, get, set, update, type UseStore } from 'idb-keyval';
 
 // There is no IndexedDB in SSG. Since this initialization occurs at the module's root, SvelteKit would encounter an error during the dapp bundling process, specifically a "ReferenceError [Error]: indexedDB is not defined". Therefore, the object for bundling on NodeJS side.
-const oisyEthAddressesStore = browser
-	? createStore('oisy-eth-addresses', 'eth-addresses')
-	: ({} as unknown as UseStore);
+const oisyAddressesStore = (storeName: string) =>
```
peterpeterparker, line 9: I would add the term `Idb` in that function name to makes its purpose more obvious

src/frontend/src/lib/types/address.ts
```diff
@@ -1,3 +1,5 @@
-export type EthAddress = string;
+export type Address = string;
```
peterpeterparker, line 1: Isn't this unrelated to the goal of this PR? -> Separate PR?

src/frontend/src/lib/api/idb.api.ts
AntonioVentilii-DFINITY: done

src/frontend/src/lib/types/address.ts
AntonioVentilii-DFINITY: ah wow, sorry, different PR, removing it

src/frontend/src/lib/api/idb.api.ts
AntonioVentilii-DFINITY: I double-checked and you are right, I confused it with [IDB](https://www.npmjs.com/package/idb): in our case, [it won't work](https://github.com/jakearchibald/idb-keyval/blob/main/custom-stores.md#defining-a-custom-database--store-name).

I will adapt the code to have different DB names too

src/frontend/src/lib/api/idb.api.ts
```diff
@@ -1,21 +1,26 @@
...
 import { isNullish } from '@dfinity/utils';
 import { createStore, del, get, set, update, type UseStore } from 'idb-keyval';
 
 // There is no IndexedDB in SSG. Since this initialization occurs at the module's root, SvelteKit would encounter an error during the dapp bundling process, specifically a "ReferenceError [Error]: indexedDB is not defined". Therefore, the object for bundling on NodeJS side.
-const oisyEthAddressesStore = browser
-	? createStore('oisy-eth-addresses', 'eth-addresses')
-	: ({} as unknown as UseStore);
+export const idbAddressesStore = (tokenId: TokenId) =>
+	browser
+		? createStore(`oisy-${tokenId.description}-address`, `${tokenId.description}-address`)
```
peterpeterparker, line 12: Does it work if the description contains special charaters such as emoji or space?

AntonioVentilii-DFINITY: the first one can be a no-restricted string, but the second one is a keypath, so it should not have spaces (or unexpected special characters).

But we don't expect `tokenId` to have spaces and/or emojis, right? Otherwise it's usage as key in type `Record` should be revisited in a few parts of the code

peterpeterparker: We don't but we don't enforce it neither. Let's add it on our todo list or I guess it will be improved the day we migrate to zod.

One more question, why `tokenId` and not `networkId`?

AntonioVentilii-DFINITY: the same reason I commented [in the other PR](https://github.com/dfinity/oisy-wallet/pull/2012#discussion_r1715326867): right now, I am thinking that we want to associate it with the crypto behind the chain? but if you think that here it would be stronger/more correct `networkId`, I change it

AntonioVentilii-DFINITY: I slept on it, and makes sense that in this case it is the network. Changed a little the code, have a look

src/frontend/src/lib/api/idb.api.ts
```diff
@@ -1,21 +1,25 @@
...
 import type { IdbEthAddress } from '$lib/types/idb';
 import type { Principal } from '@dfinity/principal';
 import { isNullish } from '@dfinity/utils';
 import { createStore, del, get, set, update, type UseStore } from 'idb-keyval';
 
 // There is no IndexedDB in SSG. Since this initialization occurs at the module's root, SvelteKit would encounter an error during the dapp bundling process, specifically a "ReferenceError [Error]: indexedDB is not defined". Therefore, the object for bundling on NodeJS side.
-const oisyEthAddressesStore = browser
-	? createStore('oisy-eth-addresses', 'eth-addresses')
-	: ({} as unknown as UseStore);
+export const idbAddressesStore = (identifier: string) =>
```
peterpeterparker, line 9: nth:

- would you mind rename to `key` or something like that? `identitfier` automatically means `account identifier` for me. old habits.

- not sure, the function is `void` or returns a type?

src/frontend/src/lib/api/idb.api.ts
```diff
@@ -1,21 +1,25 @@
...
+	browser
+		? createStore(`oisy-${identifier}-address`, `${identifier}-address`)
```
peterpeterparker, line 11: Is there a reason for changing `-addresses` to `-address`? If we keep the first, aren't we backwards compatible?

src/frontend/src/lib/api/idb.api.ts
AntonioVentilii-DFINITY: right! done both!

src/frontend/src/lib/api/idb.api.ts
AntonioVentilii-DFINITY: it was an oversight on my side, correcting

peterpeterparker (approved):
LGTM, thx

PR #2012, opened 2024-08-13
refactor(frontend): use common initCertifiedSetterStore for addressStore

AntonioVentilii-DFINITY:
# Motivation

Since we want to have a series of derived addresses corresponding to their token (and right now we have just ETH), we change the `addressStore` to use `initCertifiedSetterStore`, that is a common store in the code basically returning a record, with certified data. We already use it for example for `balanceStore`.

# Tests

No changes in local replica.


src/frontend/src/lib/services/address.services.ts
```diff
@@ -12,20 +12,22 @@ import { assertNonNullish, isNullish } from '@dfinity/utils';
...
 		await saveEthAddressForFutureSignIn({ address, identity });
 	} catch (err: unknown) {
-		addressStore.reset();
+		addressStore.reset(tokenId);
 
 		toastsError({
 			msg: {
 				text: replacePlaceholders(get(i18n).init.error.loading_address, {
-					$symbol: ETHEREUM_NETWORK_SYMBOL
+					$symbol: tokenId.description ?? ''
```
peterpeterparker, line 30: We used to use the network symbol here, that was a mistake?

src/frontend/src/lib/components/guard/AddressGuard.svelte
```diff
@@ -3,19 +3,21 @@
 	import { isNullish } from '@dfinity/utils';
 	import { certifyAddress } from '$lib/services/address.services';
 	import { warnSignOut } from '$lib/services/auth.services';
+	import { addressCertified } from '$lib/derived/address.derived';
+	import { ETHEREUM_TOKEN_ID } from '$env/tokens.env';
 
 	const validateAddress = async () => {
-		if (isNullish($addressStore)) {
+		if (isNullish($addressStore?.[ETHEREUM_TOKEN_ID])) {
```
peterpeterparker, line 10: Can we use `$ethAddress` here? I'm not sure but, if possible, it would be a bit more readable.

src/frontend/src/lib/derived/address.derived.ts
```diff
@@ -1,19 +1,24 @@
...
+import { isNullish, nonNullish } from '@dfinity/utils';
 import { derived, type Readable } from 'svelte/store';
 
 export const addressNotLoaded: Readable<boolean> = derived([addressStore], ([$addressStore]) =>
 	isNullish($addressStore)
 );
 
 export const ethAddress: Readable<OptionEthAddress> = derived([addressStore], ([$addressStore]) =>
-	$addressStore === null ? null : $addressStore?.address
+	$addressStore?.[ETHEREUM_TOKEN_ID] === null ? null : $addressStore?.[ETHEREUM_TOKEN_ID].data
```
peterpeterparker, line 12: `$addressStore?.[ETHEREUM_TOKEN_ID].data` could not potentially through an error? i.e. `$addressStore?.[ETHEREUM_TOKEN_ID]` can potentially be `undefined`?

src/frontend/src/lib/derived/address.derived.ts
```diff
@@ -1,19 +1,24 @@
...
 );
 
 export const addressCertified: Readable<boolean> = derived(
 	[addressStore],
-	([$addressStore]) => $addressStore?.certified === true
+	([$addressStore]) =>
+		nonNullish($addressStore) &&
+		Object.values($addressStore).every(
```
peterpeterparker, line 19: This introduces a different behavior. I would stick to certfying only the Eth address for now. In another PR we can rename `addressCertified` to `ethAddressCertified`.

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: I am divided on that actually: basically, we will use that message to tell that there was an error loading the address.

In case of ETH, mainnet and testnet have the same address. For bitcoin no: mainnet address will be different from testnet. So, for consistency either I use the network symbol saying in a generic way that we are referring to the chain. Or we use the tokenId because in a generic way, that is the crypto behind...


Boh.. any opinion here?

src/frontend/src/lib/components/guard/AddressGuard.svelte
AntonioVentilii-DFINITY: right! I am just in doubt about possible infinite loops: `validateAddress` will update `addressStore`, that then updates `ethAddress`. The loop should be interrupted right away, but could there be a situation with infinite loops?

src/frontend/src/lib/derived/address.derived.ts
AntonioVentilii-DFINITY: indeed, but previously that was how it was treated too, meaning that when it was just `$addressStore` (referring to ETH address), it could have been `string | undefined | null`, and the derived was accepting undefined for the right-side of the condition too (as it is in our case now).

I didn't want to change the logic

src/frontend/src/lib/derived/address.derived.ts
AntonioVentilii-DFINITY: changed

src/frontend/src/lib/services/address.services.ts
peterpeterparker: No strong opinion here, I was more curious to know if the change was expected. Works for me, thanks for the explanation.

src/frontend/src/lib/derived/address.derived.ts
peterpeterparker: You did not understand my feedback: this `$addressStore?.[ETHEREUM_TOKEN_ID].data` can throw an error if `$addressStore?.[ETHEREUM_TOKEN_ID] === undefined`.

peterpeterparker: i.e. => isn't optional chaining missing `$addressStore?.[ETHEREUM_TOKEN_ID]?.data` ?

AntonioVentilii-DFINITY: ah! yes! you are right, but then why was not raised by lint? strange... in the store right after it raised it

corrected

src/frontend/src/lib/components/guard/AddressGuard.svelte
AntonioVentilii-DFINITY: but thinking better, this was happening before too

peterpeterparker: I don't think there was a loop. It's normal to update the story if query or update. Where was the loop?

AntonioVentilii-DFINITY: Not a loop, but a re-calling of the function. Below the original code:

```typescript
const validateAddress = async () => {
.................

// step 3: the recalled function exits here
if ($addressStore.certified === true) {
	// The address is certified, all good
	return;
}

// step 1: this modifies $addressStore
const { success, err } = await certifyAddress($addressStore.address);

...............
};

// step 2: this would trigger the function above again
$: $addressStore, (async () => await validateAddress())();
```

but it seems all good. So I will use `ethAddress`

peterpeterparker: > // step 2: this would trigger the function above again

It does trigger the function again but the function prevents loop by checking for the certification

```
if ($addressStore.certified === true) {
			// The address is certified, all good
			return;
		}
```

AntonioVentilii-DFINITY: perfect!

peterpeterparker (commented):
Oh, apologies, I see it now, indeed we have to stick to the `$addressStore`.

If we use `$ethAddress` and `$addressCertified` then there might be a slight delay in the population of the derived store, therefore we might trigger the certifyAddress function too many times.

We cannnot also use `$ethAddress` solely because it does not contains the certified information.

So yes you were right, again sorry, we have to use `$addressStore?.[ETHEREUM_TOKEN_ID]`

```
<script lang="ts">
	import { addressStore } from '$lib/stores/address.store';
	import { certifyAddress } from '$lib/services/address.services';
	import { warnSignOut } from '$lib/services/auth.services';
	import { ETHEREUM_TOKEN_ID } from '$env/tokens.env';

	const validateAddress = async () => {
		const ethAddressData = $addressStore?.[ETHEREUM_TOKEN_ID];

		if (isNullish(ethAddressData)) {
			// No address is loaded, we don't have to verify it
			return;
		}

		const { certified, data: ethAddress } = ethAddressData;

		if (certified === true) {
			// The address is certified, all good
			return;
		}

		const { success, err } = await certifyAddress(ethAddress);

		if (success) {
			// The address is valid
			return;
		}

	};

	$: $addressStore, (async () => await validateAddress())();
</script>

<slot />
```

AntonioVentilii-DFINITY:
No worries! Good that you double-checked, I had tested it and I didn't perceived the functions delay

Changing

AntonioVentilii-DFINITY:
@peterpeterparker changed! ready for another review

peterpeterparker (approved):
LGTM, thanks!

PR #2013, opened 2024-08-13
feat(frontend): add BTC address calls from backend

AntonioVentilii-DFINITY:
# Motivation

New functions to call the backend and generate Bitcoin addresses by network.


src/frontend/src/lib/api/backend.api.ts
```diff
@@ -8,12 +9,29 @@ import type {
...
+	network
+}: {
+	identity: OptionIdentity;
+	network: BitcoinNetwork;
+}): Promise<EthAddress> => {
+	const { caller_btc_address } = await getBackendActor({ identity });
+	return caller_btc_address(network);
+};
+
+export const getBtcAddressMainnet = async (identity: OptionIdentity): Promise<BtcAddress> =>
```
peterpeterparker, line 29: Somehow, I remember commenting that I prefer not to have two distinct functions at the API level, which, in my opinion, is only meant to serve as a sort of proxy. Also commenting that those kind of logic can be applied at the services level.

However, I cannot recall where or what your response was—specifically, why we have those two distinct functions. Could you remind me where this discussion or these comments took place?

peterpeterparker: > Maybe I understood wrong. You meant to have just one getBtcAddress and then we map the networks when we are using it?

Yes 👍

We can discuss it and I'm not 100% I was consistent but, I see the `api` layer as a dumb proxy layer. Ideally, I would even generate it automatically, like it get's value, get the actor, pass those to the actor and that's it. You know what I mean?

AntonioVentilii-DFINITY: yes, I think I understand eheh! I read through the entire module and indeed it's that. Maybe you wanted to reduce the repetition of code (each time call `getBackendActor`) or concentrate it somewhere, so that if something changed on the backend you could easily change it from there.

In the meanwhile, I will remove the two repeated functions.

peterpeterparker (approved):
LGTM, thanks!

PR #2017, opened 2024-08-14
chore(frontend): rename transaction label for ckTokens conversions

AntonioVentilii-DFINITY:
# Motivation

We were asked to rename the transaction label when there the user converts a ckToken to the twin token, and viceversa.


peterpeterparker (approved):
LGTM, thx

PR #2019, opened 2024-08-14
feat(dev-env): Add spellcheck config

bitdivine:
# Motivation
Any developer that uses `cargo spellcheck` will get warnings on every run from terms such as `ERC20`.  While not a big deal, this is annoying and simple to fix.

# Changes
- Add a spellcheck dictionary with terms such as `ERC20`.

# Tests
None.  This is for local use only.  We could check spelling in CI but that is not a priority now.


.config/spellcheck.dic
```diff
@@ -0,0 +1,5 @@
+1
+Ethereum
+ERC20
+Wasm
+P2PKH
```
AntonioVentilii-DFINITY, line 5: shall Bitcoin be included too?

AntonioVentilii-DFINITY (approved):

PR #2021, opened 2024-08-14
feat(backend): Update migration states

bitdivine:
# Motivation
We need to small changes to the migration states.  These states are not yet used in main, so this is just an update on a type that is unused.

# Changes
- Add some additional migration states.
- Update the autogenerated bindings.
- Add an iterator to the migration states.

# Tests
There is a unit test that verifies that the iterator behaves _almost_ like the `strum` iterator.


AntonioVentilii-DFINITY (approved):
LGTM!

Out of curiosity, why the timestamps are migrated after the tokens but before the profiles?

bitdivine:
@AntonioVentilii-DFINITY  Thank you for the review!  During the data migration, the user data APIs are locked so the order of the data transfer doesn't matter.  The order happened to be the order in which I listed the managed memory structures in my notes, but any order would be fine.

PR #2022, opened 2024-08-14
fix(backend): Deduplicate State

bitdivine:
# Motivation
The `State` struct definition was duplicated by accident, probably by a botched merge or something like that.

# Changes
- Remove the duplicate (and misplaced) struct definition.

# Tests
Existing CI should suffice.


src/shared/src/types.rs
```diff
@@ -230,13 +230,6 @@ pub mod user_profile {
...
         NotFound,
     }
-
-    #[derive(CandidType, Deserialize, Copy, Clone, Eq, PartialEq, Debug)]
-    pub struct Stats {
-        pub user_profile_count: u64,
-        pub user_token_count: u64,
-        pub custom_token_count: u64,
-    }
 }
```
bitdivine, line 233: Note: Teh second, identical, type declaration is further down in this file at line 284: https://github.com/dfinity/oisy-wallet/pull/2022/files#diff-1d1df10b095588937f588d42f421dba7d5fe034432bb42802862e576e1106016L284

AntonioVentilii-DFINITY (approved):
good catch!

PR #2023, opened 2024-08-14
feat(backend): API for bulk uploading data during a migration

bitdivine:
# Motivation
To migrate data, the underlying stable memory data (excluding configuration) needs to be transported from the old backend to the new one.  This PR implements the receiving end of this transfer.

# Changes
- Add an API for bulk uploading stable memory data.
- Add rust bindings for this new method (and other methods needed for the migration).

# Tests
The data migration code (coming) has tests that verify that the data in the old canister matches the data in the new.


src/backend/src/migrate.rs
```diff
@@ -0,0 +1,70 @@
...
+};
+use candid::{decode_one, CandidType, Principal};
+use serde::Deserialize;
+use shared::types::{
+    custom_token::CustomToken, token::UserToken, user_profile::StoredUserProfile, Timestamp,
+};
+
+/// A chunk of data to be migrated.
+///
+/// Note: Given that the migration moves data types that may be private, data is transferred with candid type `Vec<u8>`
```
frederikrothenberger, line 13: Have you considered using `cbor` instead? Cbor is much more efficient and given this is about bulk data, it might make a difference.

src/shared/src/backend_api.rs
```diff
@@ -0,0 +1,27 @@
+//! Selected backend canister API signatures.
+#![allow(clippy::missing_errors_doc)] // The code is auto-generated.  Baybe `didc bind` will preserve comments in future.
```
frederikrothenberger, line 2:
```suggestion
#![allow(clippy::missing_errors_doc)] // The code is auto-generated.  Maybe `didc bind` will preserve comments in future.
```

src/shared/src/backend_api.rs
```diff
@@ -0,0 +1,27 @@
...
+use crate::types::{Guards, Stats};
+use candid::{self, Principal};
+use ic_cdk::api::call::CallResult as Result;
+
+/// Client for the backend canister, implementing a subset of the API.
+///
+/// Note: The canister signatures are generated by: `didc bind --type rs src/backend/backend.did`
```
frederikrothenberger, line 9: Are you sure this is not available (at least as an experimental feature)? I remember Yan Chen giving a demo of rust code generated from candid with a lot of customization...

frederikrothenberger (approved):
LGTM, only questions 👍

src/backend/src/migrate.rs
bitdivine: The performance now is even worse than that:  When the data is read from the `BTreeMap` it is parsed from candid bytes into Rust structs and there is a corresponding serialization when the data is re-inserted.  I thought about getting the raw bytes from the `BTreeValue` and transmitting those; that would be a significant improvement.  But the bulk data is still quite small and downtime is not a priority, so I decided not to optimize.  Using CBOR is a good idea but in a similar way this type of optimization is not a priority.  I hope that makes sense :-)

src/shared/src/backend_api.rs
bitdivine: I'm not sure.  I would love that. But even if it is available, it would take time to learn how to use it and to reap the full benefits here.  I would postpone such a refactoring until after the migration.  Maybe the right thing is to make a Jira ticket and revisit it when we are in less of a rush.  Better: I'll want to learn about this anyway, so I'll  take some of my own time to become familiar with the latest state.  That is a better motivation than Jira.

PR #2026, opened 2024-08-15
chore(ci): remove frontend folder from backend workflows

AntonioVentilii-DFINITY:
# Motivation

There may be files in src/frontend with 'backend' in their name. They may eventually trigger the backend Github workflows, even if not necessary (for example, it happened with PR #2013 )


peterpeterparker (approved):
Thx

bitdivine (approved):
LGTM but you might want to test because exclusion filters are ignored in some contexts.  If it doesn't work it won't hurt so I don't mind merging, testing, and reverting if it does nothing.

AntonioVentilii-DFINITY:
> LGTM but you might want to test because exclusion filters are ignored in some contexts. If it doesn't work it won't hurt so I don't mind merging, testing, and reverting if it does nothing.

Tested with PR #2030 and it works

PR #2028, opened 2024-08-15
fix(frontend): avoid multiple balance refresh periodically

AntonioVentilii-DFINITY:
# Motivation

As spotted by @peterpeterparker , the ETH/ERC20 balance was being refreshed periodically and that could cause asynchronous information between the balance and the transactions.

The cause is most likely a cascade refresh caused by the periodic refresh of `exchanges` store:

`exchanges --> combinedDerivedSortedTokens --> networkTokens --> enabledNetworkTokens --> enabledErc20NetworkTokens --> LoaderBalances`

Since `enabledErc20NetworkTokens` is used only in `LoaderBalances`, the solution I thought would be to simply derive it directly from the `enabledErc20Tokens` and filter by network.

### DIFF
--- a/src/frontend/src/lib/derived/network-tokens.derived.ts
+++ b/src/frontend/src/lib/derived/network-tokens.derived.ts
@@ -30,7 +30,7 @@ export const enabledNetworkTokens: Readable<Token[]> = derived(
 );
 
 /**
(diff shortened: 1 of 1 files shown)


AntonioVentilii-DFINITY:
Forget about this PR, I will do a different refactor

PR #2029, opened 2024-08-15
fix(frontend): avoid multiple balance refresh periodically

AntonioVentilii-DFINITY:
# Motivation

As spotted by @peterpeterparker , the ETH/ERC20 balance was being refreshed periodically and that could cause incoherent and asynchronous information between the balance and the transactions.

The cause is most likely a cascade refresh caused by the periodic refresh of `exchanges` store:

`exchanges --> combinedDerivedSortedTokens --> networkTokens --> enabledNetworkTokens --> enabledErc20NetworkTokens --> LoaderBalances`

This PR tries to solve that issue removing the exchanges dependency from `enabledErc20NetworkTokens` dependencies.

To do that, we move `combinedDerivedSortedTokens`, so that the sorting happens after having filtered for `network` and `enabled`. This has the additional effect of potentially reducing the computational work for sorting.

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensSignedOut.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensSignedOut.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { enabledNetworkTokens } from '$lib/derived/network-tokens.derived.js';
+	import { combinedDerivedSortedNetworkTokens } from '$lib/derived/network-tokens.derived.js';
 	import TokenCardWithUrl from '$lib/components/tokens/TokenCardWithUrl.svelte';
 	import { onMount } from 'svelte';
 	import { unsafeLoadDefaultPublicIcrcTokens } from '$icp/services/icrc.services';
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/tokens/TokensSignedOut.svelte
```diff
@@ -8,7 +8,7 @@
 	onMount(unsafeLoadDefaultPublicIcrcTokens);
 </script>
 
-{#each $enabledNetworkTokens as token (token.id)}
+{#each $combinedDerivedSortedNetworkTokens as token (token.id)}
```
peterpeterparker, line 11: I'm not sure why you change this. It seems unrelated to the PR fix and it's also the list of tokens displayed when the user is not signed-in - i.e. no exchange values should be fetched at that time. What's the reason?

AntonioVentilii-DFINITY: before this PR, since `enabledNetworkTokens` derived, after some level, from `combinedDerivedSortedTokens`, that list was sorted according to the "sorting logic". Meaning that we were showing the signed-out list in a coherent order in respect to the signed in one (even if blurred).

After this PR that would have not be true anymore: `combinedDerivedSortedNetworkTokens` becomes derived from `enabledNetworkTokens`, with the only difference of the pinned tokens and the sorted ones.

peterpeterparker: Ah thanks for the explanation. Somehow I was suprised that the "Signed-out" tokens list used a store that relies on the exchanges values.

peterpeterparker (approved):
LGTM, thanks

PR #2032, opened 2024-08-16
ci(e2e): fix e2e tests pipeline

DenysKarmazynDFINITY:
# Motivation

The goal of this PR is to fix e2e tests pipeline.

# Changes

1. To prevent "out of space" error while building an e2e docker image, added a small command to remove cache folder.
2. Dockerfile.e2e was missing a few tools required to build and deploy backend canister.
3. To fix the issue described [here](https://dfinity.slack.com/archives/C072H2FMW1F/p1723737578374699), we need to create some canisters before running the deploy script.

### DIFF
--- a/docker/deploy
+++ b/docker/deploy
@@ -6,6 +6,7 @@ dfx start --background --quiet
 
 ./docker/wait-port "$DFX_PORT"
 
+# TODO: Remove these lines after the dependency problem is solved by ic-terraform
 dfx canister create internet_identity --specified-id rdmx6-jaaaa-aaaaa-aaadq-cai
 dfx canister create pouh_issuer --specified-id qbw6f-caaaa-aaaah-qdcwa-cai
 
--- a/docker/deploy
+++ b/docker/deploy
@@ -6,10 +6,6 @@ dfx start --background --quiet
 
 ./docker/wait-port "$DFX_PORT"
 
-# TODO: Remove these lines after the dependency problem is solved by ic-terraform
-dfx canister create internet_identity --specified-id rdmx6-jaaaa-aaaaa-aaadq-cai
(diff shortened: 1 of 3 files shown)


docker/deploy
```diff
@@ -6,6 +6,9 @@ dfx start --background --quiet
 
 ./docker/wait-port "$DFX_PORT"
 
+dfx canister create internet_identity --specified-id rdmx6-jaaaa-aaaaa-aaadq-cai
```
AntonioVentilii-DFINITY, line 9: Can you please add a to-do here saying that it is a temporary solution? For what I understood from Max, it will be solved by ic-terraform, even if it is not something so in the short-term. Right, @bitdivine ?

bitdivine: I have to temper your expectations.  `ic-terraform` is in very early stages of development and it will be months until we can use it extensively.  Maybe even years if we are kept very busy, with no time to improve the infrastructure.  But the sole purpose of `terraform` is to map the dependencies between infrastructure components and deploy them in the right order.  The above is an example of having to fulfil such a  dependency so, yes, that should be covered by `terraform`.  I am looking forward to it.

DenysKarmazynDFINITY: Added the comment. And thanks for the explanation 👍

AntonioVentilii-DFINITY (approved):
> 3\. To fix the issue described [here](https://dfinity.slack.com/archives/C072H2FMW1F/p1723737578374699), we need to create some canisters before running the deploy script.

As discussed, please put this in a separate PR. For the rest, all good

DenysKarmazynDFINITY:
The backend deployment fixed is moved to a separate PR - https://github.com/dfinity/oisy-wallet/pull/2035

PR #2033, opened 2024-08-16
test(e2e): homepage screenshot tests

DenysKarmazynDFINITY:
# Motivation

The goal of this PR is to add basic snapshot tests for homepage (user logged in and logged out states).

# Changes

1. Basic Playwright code.
2. To properly compare homepage screenshots, I deactivate the infinite Hero background animation before taking a snapshot.
3. Added data-tid to the related elements on the page.

# Tests

Please see the committed screenshots.


DenysKarmazynDFINITY:
To get the pipeline correctly running, https://github.com/dfinity/oisy-wallet/pull/2032 should be merged first.

DenysKarmazynDFINITY:
The pipeline is now green, so PR can be fully reviewed.

src/frontend/src/lib/constants/test-ids.constant.ts
```diff
@@ -0,0 +1,3 @@
+export const LOGIN_BUTTON = 'login-button';
+export const TOKENS_SKELETONS_INITIALIZED = 'tokens-skeletons-initialized';
+export const THREE_BACKGROUND_CANVAS = 'three-background-canvas';
```
AntonioVentilii-DFINITY, line 3: as side note: there are some `tid` around the code that could go there too

e2e/homepage.spec.ts
```diff
@@ -1,11 +1,49 @@
...
-	await page.goto(testUrl);
+const hideHeroAnimation = async (page: Page): Promise<void> => {
+	await page
+		.getByTestId(THREE_BACKGROUND_CANVAS)
+		.evaluate((element) => (element.style.display = 'none'));
+};
 
-	await page.getByTestId('login-button').waitFor();
+test.describe('logged out user', () => {
+	test.beforeEach(async ({ page }) => {
```
AntonioVentilii-DFINITY, line 18: curiosity for me: is it necessary to add `test.` before the functions? if not, can we remove it, so that it would be "cleaner"?

e2e/homepage.spec.ts
```diff
@@ -1,11 +1,49 @@
...
+		await hideHeroAnimation(page);
+
+		await expect(page).toHaveScreenshot({ fullPage: true });
+	});
+});
+
+testWithII.describe('logged in user', () => {
+	testWithII.beforeEach(async ({ page, iiPage }) => {
+		const url = 'http://127.0.0.1:4943';
+		const canisterId = 'rdmx6-jaaaa-aaaaa-aaadq-cai';
```
AntonioVentilii-DFINITY, line 34: for another side note: I am thinking that in another PR we should put these values somewhere to be re-used throughout the code...

e2e/homepage.spec.ts
DenysKarmazynDFINITY: The idea was to extend this `describe` with other tests related to Homepage and therefore to make use of `beforeEach`, but now as I started working on other tests in a parallel PR, I realised that the structure where 1 spec file = 1 test suite (or a few, in case applicable) will work better for us, similar to this repo - https://github.com/dfinity/gix-components/tree/main/e2e. Having said that, we don't need `beforeEach`, I'll remove it 👍

e2e/homepage.spec.ts
DenysKarmazynDFINITY: Definitely. I will move all consts and helpers methods (`hideHeroAnimation` in this case) to separate files, especially since we'll proceed with one-spec-per-file approach and hence we'll need to re-use those things.

e2e/homepage.spec.ts
AntonioVentilii-DFINITY: no wait, maybe you misunderstood my comment. My point is, instead of `test.describe`, can we just use `describe`? Or do we need the specific `describe` prop of the playwright test suite?
same question for `beforeEach`. It is a genuine doubt I have

e2e/homepage.spec.ts
AntonioVentilii-DFINITY: Actually, sorry, the canisterId you could take it directly from `canister_e2e_ids.json` in this PR, right?

e2e/homepage.spec.ts
DenysKarmazynDFINITY: Ah, I see. Default `describe` and `beforeEach` are declared in `vitest`. And in case of e2e tests, we want to make sure that we use the ones from Playwright. 
P.S.: we actually don't need `describe` here at all because of the same reason described above in this thread. I'll adjust it.

e2e/homepage.spec.ts
DenysKarmazynDFINITY: Good tip, will use that function.

e2e/homepage.spec.ts
DenysKarmazynDFINITY: This part is now updated. Resolving the thread.

e2e/homepage.spec.ts
DenysKarmazynDFINITY: I moved the constants/utils already in this PR. Though, for now, `getInternetIdentityCanisterId` returns a hardcoded string because in order to use `readCanisterIds` I need to update Dockerfile.e2e (at the moment, `canister_e2e_ids.json` is not a part of the e2e docker image). I will do it in a separate PR.

e2e/authentication.spec.ts
```diff
@@ -1,9 +1,8 @@
 import { testWithII } from '@dfinity/internet-identity-playwright';
-
-const testUrl = '/';
+import { HOMEPAGE_URL } from './shared/constants';
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: 1. Agreed. 
2. I double-checked it - those files are not part of the build.

e2e/homepage.spec.ts
```diff
@@ -1,11 +1,26 @@
...
 
-test('should display not logged in homepage', async ({ page }) => {
-	await page.goto(testUrl);
+	await hideHeroAnimation(page);
 
-	await page.getByTestId('login-button').waitFor();
+	await expect(page).toHaveScreenshot({ fullPage: true });
+});
+
+testWithII('should display homepage in logged in state', async ({ page, iiPage }) => {
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Yes, this PR is mostly for us to agree on the setup and the structure for all future tests.

e2e/homepage.spec.ts
```diff
@@ -1,11 +1,26 @@
+import { LOGIN_BUTTON, TOKENS_SKELETONS_INITIALIZED } from '$lib/constants/test-ids.constant';
+import { testWithII } from '@dfinity/internet-identity-playwright';
 import { expect, test } from '@playwright/test';
+import { DEFAULT_ROOT_URL, HOMEPAGE_URL } from './shared/constants';
+import { getInternetIdentityCanisterId, hideHeroAnimation } from './shared/utils';
 
-const testUrl = '/';
+test('should display homepage in logged out state', async ({ page }) => {
+	await page.goto(HOMEPAGE_URL);
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Really like the idea, code looks much neater now.

e2e/shared/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const HOMEPAGE_URL = '/';
+
+export const DEFAULT_ROOT_URL = 'http://127.0.0.1:4943';
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Agreed.

e2e/shared/utils.ts
```diff
@@ -0,0 +1,11 @@
...
+import type { Page } from '@playwright/test';
+
+export const hideHeroAnimation = async (page: Page): Promise<void> => {
+	await page
+		.getByTestId(THREE_BACKGROUND_CANVAS)
+		.evaluate((element) => (element.style.display = 'none'));
+};
+
+// TODO: use readCanisterIds to get the II canister ID
+export const getInternetIdentityCanisterId = (): string => 'rdmx6-jaaaa-aaaaa-aaadq-cai';
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Didn't work when I just tried to import it, so I guess I'll need to use the function from `vite.utils`. In any case, I'll do it separately, if that's ok with you.

playwright.config.ts
```diff
@@ -9,6 +9,7 @@ export default defineConfig({
 		port: DEV ? 5173 : 4173
 	},
 	testDir: 'e2e',
+	timeout: 180000,
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: With the default timeout (30s), the tests felt a bit flaky - looks like sometime the wallet initialisation wasn't finishing in time when running tests inside a docker in CI. To be on the safer side, I increased the value.

src/frontend/src/lib/constants/test-ids.constant.ts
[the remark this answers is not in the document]
DenysKarmazynDFINITY: I was sticking to component names, but agree with your point - renamed to `HERO_ANIMATION_CANVAS`

src/frontend/src/lib/constants/test-ids.constant.ts
```diff
@@ -0,0 +1,3 @@
+export const LOGIN_BUTTON = 'login-button';
+export const TOKENS_SKELETONS_INITIALIZED = 'tokens-skeletons-initialized';
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: `TokensSkeleton` has 2 different states - when `erc20UserTokensNotInitialized` is not initialised, it renders skeleton cards, when it is initialised - the real tokens data. So, I added `INITIALIZED` just to make an emphasis on the state of the component we're interested in.

DenysKarmazynDFINITY:
@peterpeterparker @AntonioVentilii-DFINITY thanks for the reviews! I applied all your comments, so the code is ready for another go.

e2e/homepage.spec.ts
```diff
@@ -1,11 +1,19 @@
...
-const testUrl = '/';
+test('should display homepage in logged out state', async ({ page }) => {
+	const homepage = new Homepage(page);
 
-test('should display not logged in homepage', async ({ page }) => {
-	await page.goto(testUrl);
+	await homepage.waitForLoggedOut();
 
-	await page.getByTestId('login-button').waitFor();
+	await expect(homepage.page).toHaveScreenshot({ fullPage: true });
```
peterpeterparker, line 10: I would not leak the `page`, feels a bit like an anti pattern but, that's maybe an opiniated answer. I would use the page available in the function or defer the screenshot with the page object.

e2e/utils/pages/homepage.page.ts
```diff
@@ -0,0 +1,64 @@
...
+
+	get page(): Page {
+		return this._page;
+	}
+
+	async goto(): Promise<void> {
+		await this._page.goto(HOMEPAGE_URL);
+	}
+
+	async waitForLoggedOut(): Promise<void> {
```
peterpeterparker, line 31: I'm not sure so just asking, shouldn't this function finds place in a child not in the parent class?

e2e/utils/pages/homepage.page.ts
```diff
@@ -0,0 +1,64 @@
...
+		await this.goto();
+		await this._page.getByTestId(LOGIN_BUTTON).waitFor();
+		await this.hideHeroAnimation();
+	}
+}
+
+export class HomepageLoggedIn extends Homepage {
+	private readonly _iiPage: InternetIdentityPage;
```
peterpeterparker, line 39: I assumed you prefixed `_page` with `_` because it is `protected`, so no I'm confuse what `_` means?

If you want to distinguish protected and private you can also use `#` for the latest.

```
readonly #iiPage: InternetIdentityPage;
```

If it means "internal", given that we are using Typescript, I'm not sure it adds much value.

But again, maybe I am just not understanding the purpose of the prefix.

e2e/utils/pages/homepage.page.ts
```diff
@@ -0,0 +1,64 @@
...
+
+	constructor(page: Page, iiPage: InternetIdentityPage) {
```
peterpeterparker, line 41: Can we / should weuse an object as parameter for the constructor? For consistency reasons.

playwright.config.ts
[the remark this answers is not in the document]
peterpeterparker: If the tests are flaky, it's better to address the root cause rather than just extending the timeout, as this could lead to future failures. Could you please revert the change and provide a PR that ensures the tests run reliably in the CI environment?

src/frontend/src/lib/constants/test-ids.constant.ts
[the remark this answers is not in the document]
peterpeterparker: Gotcha. Thanks for the explanation.

e2e/utils/pages/homepage.page.ts
```diff
@@ -0,0 +1,64 @@
...
+		super(page);
+
+		this._iiPage = iiPage;
+	}
+
+	async signInWithNewIdentity(): Promise<void> {
+		await this._iiPage.waitReady({
+			url: LOCAL_REPLICA_URL,
+			// TODO: take this value from vite.utils or FE constants
```
peterpeterparker, line 50: And if you try to access the vite env import or process env? Reading `vite.utils` is not a correct solution, this utils should only be use by the vite configuration.

playwright.config.ts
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Removed the timeout param. I initially added it before I started using `iiPage.waitReady`, so it could be one of the reasons why CI threw a timeout error initially. I will obviously keep my eyes on CI and come back to the topic if this issue appears again.

e2e/homepage.spec.ts
DenysKarmazynDFINITY: Using the original `page` to take a screenshot sounds clean to me. Updated.

e2e/utils/pages/homepage.page.ts
DenysKarmazynDFINITY: I consider `Homepage` class as a `HomepageLoggedOut`. That's why I kept `waitForLoggedOut` there and have `waitForLoggedIn` in `HomepageLoggedIn`. Now that you asked it, I think it makes sense to rename in the code to keep the intention clear. Let me know what you think.

e2e/utils/pages/homepage.page.ts
DenysKarmazynDFINITY: That was a mixup from my side. Now, since we don't access neither `page` nor `iiPage` outside of the classes, I simplified it by just defining them as `readonly`.

e2e/utils/pages/homepage.page.ts
DenysKarmazynDFINITY: Done.

e2e/utils/pages/homepage.page.ts
DenysKarmazynDFINITY: Ok, got it regarding the `vite.utils`. Looking at the moment what would a better solution here.

e2e/utils/pages/homepage.page.ts
peterpeterparker: It's a bit confusing to me but, I'm also maybe just picky 😅. Why not having `Homepage` as an `abstract class` and then two implementation `HomepageLoggedOut` and `HomepageLoggedIn`?

e2e/utils/pages/homepage.page.ts
```diff
@@ -0,0 +1,68 @@
...
+type HomepageParams = {
+	page: Page;
+};
+
+type HomepageLoggedInParams = {
+	iiPage: InternetIdentityPage;
+} & HomepageParams;
+
+export class Homepage {
+	readonly page: Page;
```
peterpeterparker, line 19: I guess those (I said those because same for `iiPage` in the other class) can be private? 

```
private readonly page: Page;

// or

readonly #page: Page;
```

DenysKarmazynDFINITY: Yes, now I made them both private.

e2e/utils/pages/homepage.page.ts
DenysKarmazynDFINITY: Yep, was confusing indeed - I now update it the way you suggested.

e2e/utils/pages/homepage.page.ts
```diff
@@ -0,0 +1,84 @@
...
+
+	abstract waitForReady(): Promise<void>;
+}
+
+export class HomepageLoggedOut extends Homepage {
+	constructor({ page }: HomepageParams) {
+		super(page);
+	}
+
+	async waitForReady(): Promise<void> {
```
peterpeterparker, line 53: Just sharing in case you like it: you can mark those functions as being the override of the abstract. Don't know if that's the standard though. Really just sharing the idea.

```
/**
  * @override
*/
async waitForReady(): Promise<void> {
  await this.waitForHomepageReady();
}
```

peterpeterparker (approved):
LGTM, thanks

PR #2034, opened 2024-08-16
feat(frontend): IDB functions for Bitcoin address

AntonioVentilii-DFINITY:
# Motivation

We replicate the IDB functions for Bitcoin Mainnet address, as they are used for Ethereum.

### DIFF
--- a/src/frontend/src/lib/types/idb.ts
+++ b/src/frontend/src/lib/types/idb.ts
@@ -1,4 +1,10 @@
-import type { EthAddress } from '$lib/types/address';
+import type { BtcAddress, EthAddress } from '$lib/types/address';
+
+export interface IdbBtcAddress {
+	address: BtcAddress;
+	createdAtTimestamp: number;
+	lastUsedTimestamp: number;
+}
 
 export interface IdbEthAddress {
 	address: EthAddress;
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/types/idb.ts
```diff
@@ -1,4 +1,10 @@
-import type { EthAddress } from '$lib/types/address';
+import type { BtcAddress, EthAddress } from '$lib/types/address';
+
+export interface IdbBtcAddress {
+	address: BtcAddress;
+	createdAtTimestamp: number;
```
peterpeterparker, line 5: should we have a common interface `IdbAddress` that declares the timestamps and implements it for Btc and Eth?

peterpeterparker (approved):
LGTm, thx

src/frontend/src/lib/types/idb.ts
AntonioVentilii-DFINITY: Yes! Agreed, something like:

```typescript
interface IdbAddress<T> {
	address: T;
	createdAtTimestamp: number;
	lastUsedTimestamp: number;
}

export type IdbBtcAddress = IdbAddress<BtcAddress>;

export type IdbEthAddress = IdbAddress<EthAddress>;
```

peterpeterparker: Nice

PR #2036, opened 2024-08-16
feat(frontend): animate header - I dont like my solution

peterpeterparker:
# Motivation

Just a PR to showcase @AntonioVentilii-DFINITY what I tried to implement. I don't really like it because the solution does not cover 100%, there is a glich when the toggle between the two observers exactly match the distance between the two.


peterpeterparker:
This is not the solution. We should explore modify the layout to adapt to the requirements.

PR #2040, opened 2024-08-17
feat(frontend): add BTC address loader services

AntonioVentilii-DFINITY:
# Motivation

We add the Bitcoin address loader among the service (we will call it in a separate PR).

# Changes

- Refactor address load services to be more generic, in order to avoid duplication of code.
- Adjust the existing ETH address services.
- Create the functions for BTC.

# Tests

No changes on local replica when loading the ETH address.


src/frontend/src/lib/services/address.services.ts
```diff
@@ -39,19 +55,51 @@ export const loadAddress = async (): Promise<{ success: boolean }> => {
...
+const saveTokenAddressForFutureSignIn = async <T extends Address>({
 	identity,
-	address
+	address,
+	setIdbAddress
 }: {
 	identity: OptionIdentity;
-	address: EthAddress;
+	address: T;
+	setIdbAddress: (params: { address: IdbAddress<T>; principal: Principal }) => Promise<void>;
```
peterpeterparker, line 95: Can the type of `params` be shortened to `SetIdbAddressParams` maybe?

src/frontend/src/lib/services/address.services.ts
```diff
@@ -39,19 +55,51 @@ export const loadAddress = async (): Promise<{ success: boolean }> => {
 	return { success: true };
 };
 
-const saveEthAddressForFutureSignIn = async ({
+const loadBtcAddress = async ({
+	tokenId,
+	network
+}: {
+	tokenId: TokenId;
```
peterpeterparker, line 62: Can the type here be something like `typeof BTC_MAINNET_TOKEN_ID`?

src/frontend/src/lib/services/address.services.ts
```diff
@@ -39,19 +55,51 @@ export const loadAddress = async (): Promise<{ success: boolean }> => {
...
+	network: BitcoinNetwork;
+}): Promise<{ success: boolean }> =>
+	loadTokenAddress<BtcAddress>({
+		tokenId,
+		getAddress: (identity) =>
+			getBtcAddress({
+				identity,
+				network: network === 'mainnet' ? { mainnet: null } : { testnet: null }
```
peterpeterparker, line 70: What's more important, loading mainnet or testnet? If mainnet, should we invert the if? i.e. default fallback

Note: Strictly a question, no preferences from my side

peterpeterparker (approved):
LGTM, minor questions and observations therefore already approving. Thanks.

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: good catch! done

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: I think I understand your point. That in any case, the default fallback should be mainnet. Make sense, changed

PR #2041, opened 2024-08-17
feat(frontend): add IDB BTC address service loader

AntonioVentilii-DFINITY:
# Motivation

We add the IDB Bitcoin address loader among the services (we will call it in a separate PR).

# Changes

- Refactor IDB address load services to be more generic, in order to avoid duplication of code.
- Adjust the existing ETH address services.
- Create the functions for BTC.

# Tests

No changes on local replica when loading the ETH address.


src/frontend/src/lib/services/address.services.ts
```diff
@@ -121,20 +130,20 @@ export const loadIdbAddress = async (): Promise<{ success: boolean }> => {
...
+		const { address } = idbAddress;
 		addressStore.set({ tokenId, data: { data: address, certified: false } });
 
-		await updateIdbEthAddressLastUsage(identity.getPrincipal());
+		await updateIdbAddressLastUsage(identity.getPrincipal());
 	} catch (err: unknown) {
 		// We silence the error as the dapp will proceed with a standard lookup of the address.
 		console.error(
-			'Error encountered while searching for a locally stored public address in the browser.'
+			`Error encountered while searching for locally stored ${tokenId.description} public address in the browser.`
```
peterpeterparker, line 146: i18n key in another pR?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: correct, there are a lot more around address services, address guards and address loader, so, once finished, I will do them all

PR #2043, opened 2024-08-19
feat(frontend,devOps): Split backend

bitdivine:
# Motivation
Here we wish to make the frontend make all its signing API calls to the signing canister.

# Changes
- Create a signing agent.
- Move all signing API calls from the backend agent to the signer agent.

# Tests
- Existing CI should pass.


src/frontend/src/eth/services/send.services.ts
```diff
@@ -19,7 +19,7 @@ import {
 	toCkErc20HelperContractAddress,
 	toCkEthHelperContractAddress
 } from '$icp-eth/utils/cketh.utils';
-import { signTransaction } from '$lib/api/backend.api';
+import { signTransaction } from '$lib/api/signer.api';
```
bitdivine, line 22: This API method is now available from the signer canister, not the backend.

src/frontend/src/eth/services/wallet-connect.services.ts
```diff
@@ -7,7 +7,7 @@ import {
 	getSignParamsMessageTypedDataV4Hash
 } from '$eth/utils/wallet-connect.utils';
 import { assertCkEthMinterInfoLoaded } from '$icp-eth/services/cketh.services';
-import { signMessage as signMessageApi, signPrehash } from '$lib/api/backend.api';
+import { signMessage as signMessageApi, signPrehash } from '$lib/api/signer.api';
```
bitdivine, line 10: This API method is now available from the signer canister, not the backend.

src/frontend/src/lib/services/address.services.ts
```diff
@@ -9,6 +8,7 @@ import {
 	updateIdbBtcAddressMainnetLastUsage,
 	updateIdbEthAddressLastUsage
 } from '$lib/api/idb.api';
+import { getBtcAddress, getEthAddress } from '$lib/api/signer.api';
```
bitdivine, line 11: This API method is now available from the signer canister, not the backend.

src/frontend/src/lib/constants/app.constants.ts
```diff
@@ -43,6 +43,12 @@ export const BACKEND_CANISTER_ID = LOCAL
 		? import.meta.env.VITE_STAGING_BACKEND_CANISTER_ID
 		: import.meta.env.VITE_IC_BACKEND_CANISTER_ID;
 
+export const SIGNER_CANISTER_ID = LOCAL
+	? import.meta.env.VITE_LOCAL_SIGNER_CANISTER_ID
+	: STAGING
+		? import.meta.env.VITE_STAGING_SIGNER_CANISTER_ID
+		: import.meta.env.VITE_IC_SIGNER_CANISTER_ID;
```
bitdivine, line 50: The logic for getting the signer canister ID is identical to that for the backend.  I guess it might diverge in future though.

src/frontend/src/lib/api/backend.api.ts
```diff
@@ -1,69 +1,17 @@
...
-	SignRequest,
 	UserProfile,
 	UserToken
 } from '$declarations/backend/backend.did';
 import { getBackendActor } from '$lib/actors/actors.ic';
-import type { EthAddress } from '$lib/types/address';
 import type { OptionIdentity } from '$lib/types/identity';
 import type { Identity } from '@dfinity/agent';
 import type { Principal } from '@dfinity/principal';
 import { toNullable, type QueryParams } from '@dfinity/utils';
```
bitdivine, line 14: The public key crypto methods below are moved to the signing canister.

src/frontend/src/lib/actors/actors.ic.ts
```diff
@@ -42,6 +43,36 @@ export const getBackendActor = async ({
 	return backend;
 };
```
bitdivine, line 44: Here we create a new Signer actor.  The code is identical to the backend actor code except that it refers to the signer canister ID instead.  Dependencies such as internet identity can be stripped out from this copy soon.

src/frontend/src/lib/actors/actors.ic.ts
```diff
@@ -42,6 +43,36 @@ export const getBackendActor = async ({
...
+}): Promise<SignerActor> => {
+	// TODO: Delete this dependency once the signer has been stripped down.
+	assertNonNullish(identity, get(i18n).auth.error.no_internet_identity);
+
+	const { signer } = actors ?? { signer: undefined };
+
+	if (isNullish(signer)) {
+		const actor = await createActor<SignerActor>({
+			canisterId: SIGNER_CANISTER_ID,
+			idlFactory: certified ? idlCertifiedFactoryBackend : idlFactoryBackend,
```
AntonioVentilii-DFINITY, line 61: shouldn't here go the signer's idlFactories?

src/frontend/src/lib/actors/actors.ic.ts
AntonioVentilii-DFINITY: eheh! here we fall in the same topic: should we have a single function that takes as input the respective differentiations between `backend` and `signer`, such as canisterId, idlFactories, etc, to avoid repetition of code?

AntonioVentilii-DFINITY:
@bitdivine I took the liberty of updating with main branch because I just wanted to check if a few recent PRs of mine would be affected by the migration of the actors functions

src/frontend/src/lib/actors/actors.ic.ts
bitdivine: I believe that the similarity in code is again likely to be short-lived.  For example, the internet identity check on line 54 will disappear for the signer.  That said, if you have a concrete suggestion for how to do this, I would be interested to learn.  In Rust I know how to pass a type such as `{Backend,Signer}Actor`, but not in typescript.

src/frontend/src/lib/actors/actors.ic.ts
bitdivine: Good catch!  Done.

bitdivine: Done

AntonioVentilii-DFINITY (approved):

src/frontend/src/lib/actors/actors.ic.ts
AntonioVentilii-DFINITY: indeed! it was just an observation; since it is all short-term, I would leave it like that and we adjust after, as you said

PR #2045, opened 2024-08-19
feat(devops): Add signer canister

bitdivine:
# Motivation
The backend will be split into a signer and a user data canister.  The first step in this process, according to [the plan](https://www.notion.so/dfinityorg/Oisy-frontend-uses-separate-backend-and-signer-6a84a876f6ec434891eafcb56756937c), is to have separate canister names, even though they initially refer to the same canister.

# Changes
- Add a `signer` canister to `dfx.json` and `canister_ids.json`.

# Tests
- Existing CI should pass.
- Reviewers should manually verify that, at the time of writing, the canister IDs are identical for the `backend` and `signer`.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -1,6 +1,13 @@
 {
 	"dfx": "0.20.0",
 	"canisters": {
+		"signer": {
+			"candid": "src/backend/backend.did",
+			"package": "backend",
+			"type": "rust",
+			"optimize": "cycles",
+			"gzip": true
+		},
 		"backend": {
 			"candid": "src/backend/backend.did",
 			"package": "backend",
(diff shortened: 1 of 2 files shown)


dfx.json
```diff
@@ -1,6 +1,13 @@
 {
 	"dfx": "0.20.0",
 	"canisters": {
+		"signer": {
+			"candid": "src/backend/backend.did",
+			"package": "backend",
```
peterpeterparker, line 6: I assume it's correct that "the signer" points to "backend".

peterpeterparker (approved):
LGTM, thx

dfx.json
bitdivine: Yes, it is!  For now it uses the exact same Wasm.

PR #2048, opened 2024-08-19
fix(frontend): ICP Erc20 to native ICP destination account identifier

peterpeterparker:
# Motivation

To enable the conversion of "ICP ERC20" tokens to native ICP tokens, users should be able to input an "old" ICP account identifier in hexadecimal format. Previously, this function worked as expected. However, due to an assertion introduced in early 2024, the destination field in the ETH "Send" wizard now requires a properly formatted ETH address. To accommodate this specific use case, we need to modify the validation process to recognize and accept ICP account identifiers as valid inputs in the destination field.

# Notes

There might be some more abstracted ways to solve this but, to prevent issues, I rather liked to bit a bit verbose given that it is a really particular use case. 

# Changes

- Improve `isInvalidDestination` to throw if the address is an invalid account identifier on "ICP Erc20"
- Do not map to Eth address prefixed with `0x` on send
- Do not map to Eth address prefixed with `0x` when calculating fee

### DIFF
--- a/src/frontend/src/eth/components/fee/FeeContext.svelte
+++ b/src/frontend/src/eth/components/fee/FeeContext.svelte
@@ -3,7 +3,7 @@
 	import type { Erc20Token } from '$eth/types/erc20';
 	import { ethAddress } from '$lib/derived/address.derived';
 	import { toastsError, toastsHide } from '$lib/stores/toasts.store';
-	import { debounce } from '@dfinity/utils';
+	import { debounce, nonNullish } from '@dfinity/utils';
 	import { initMinedTransactionsListener } from '$eth/services/eth-listener.services';
 	import { getContext, onDestroy, onMount } from 'svelte';
 	import { FEE_CONTEXT_KEY, type FeeContext } from '$eth/stores/fee.store';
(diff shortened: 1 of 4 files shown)


src/frontend/src/eth/components/fee/FeeContext.svelte
```diff
@@ -97,7 +98,13 @@
 				...(await getFeeData()),
 				gas: await getErc20FeeData({
 					...erc20GasFeeParams,
-					targetNetwork
+					targetNetwork,
+					to:
+						// When converting "ICP Erc20" to native ICP, the destination address is an "old" ICP hex account identifier.
+						// Therefore, it should not be prefixed with 0x.
+						nonNullish(targetNetwork) && isNetworkICP(targetNetwork)
```
AntonioVentilii-DFINITY, line 105: just as side note, we could actually have that `isNetworkXXX` accepts `undefined` at some point, no?

peterpeterparker: Definitely! Can you do that afterwards?

AntonioVentilii-DFINITY (approved):

src/frontend/src/eth/components/fee/FeeContext.svelte
AntonioVentilii-DFINITY: Done:

- PR #2057

PR #2056, opened 2024-08-19
feat(devops): Signer deploy script

bitdivine:
# Motivation
Step 2 of [the frontend migration plan](https://www.notion.so/dfinityorg/Oisy-frontend-uses-separate-backend-and-signer-6a84a876f6ec434891eafcb56756937c) is to make a deploy script for the signer canister.

This PR implements that step, although the script has no use (yet):
- `deploy.sh` does not deploy the signer separately, and probably never will as signer deployments will (soon) be from another repository.
- Likewise the `HACKING.md` file does not need instructions for deploying the signer (yet).

# Changes
- Add a deploy script for the `signer` canister.

# Tests
The script is a copy of `deploy.backend.sh` so most code branches will already be checked there.

I have run:
```
ENV=local WALLET=bnz7o-iuaaa-aaaaa-qaaaa-cai bash -x scripts/deploy.signer.sh
```
Note: `WALLET` needs to be set explicitly when deploying to `local`

### DIFF
--- a/scripts/deploy.signer.sh
+++ b/scripts/deploy.signer.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+
+II_CANISTER_ID="$(dfx canister id internet_identity --network "${ENV:-local}")"
+POUH_ISSUER_CANISTER_ID="$(dfx canister id pouh_issuer --network "${ENV:-local}")"
+
+case $ENV in
+"staging")
+  ECDSA_KEY_NAME="test_key_1"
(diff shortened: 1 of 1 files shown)


AntonioVentilii-DFINITY (approved):
LGTM! BTW, It would be ideal to have a single script that can be called either with 'signer' or 'backend' parameter, or something similar,

bitdivine:
@AntonioVentilii-DFINITY Well, the script is the same now, but it is expected to diverge fairly rapidly. :-)

PR #2057, opened 2024-08-19
refactor(frontend): ICP network checker accepts undefined

AntonioVentilii-DFINITY:
# Motivation

Since we could accept an `undefined` network (Chain Fusion), we adapt the function `isNetworkICP`.


src/frontend/src/eth/services/fee.services.ts
```diff
@@ -42,10 +41,11 @@ export const getErc20FeeData = async ({
...
 	try {
-		const { getFeeData: fn } =
-			nonNullish(targetNetwork) && isNetworkICP(targetNetwork)
-				? infuraErc20IcpProviders(targetNetwork.id)
-				: infuraErc20Providers(targetNetwork?.id ?? sourceNetworkId);
+		const targetNetworkId: NetworkId | undefined = targetNetwork?.id;
+
+		const { getFeeData: fn } = isNetworkIdICP(targetNetworkId)
+			? infuraErc20IcpProviders(targetNetworkId)
+			: infuraErc20Providers(targetNetworkId ?? sourceNetworkId);
```
AntonioVentilii-DFINITY, line 48: here, I refactored using a different function, because the alternative was to remove the `nonNullish` check, BUT adding a fallback in `infuraErc20IcpProviders`:

```typescript
const { getFeeData: fn } = isNetworkICP(targetNetwork)
			? infuraErc20IcpProviders(targetNetwork?.id ?? XXXXXXX)
			: infuraErc20Providers(targetNetwork?.id ?? sourceNetworkId);
```

because Typescript does not read `targetNetwork` as `NonNullable` on the `true` branch of the condition. I tried to force the `NonNullable`, but then it reads `targetNetwork` as `undefined` on the `false` branch of the condition always

peterpeterparker (approved):
LGTM; thanks

PR #2058, opened 2024-08-19
feat(backend): Step through migration states

bitdivine:
# Motivation
User data migration to another canister is triggered by an API call.  This PR introduces a skeleton API call that, ghostlike, steps through the motions of the migration without doing anything.

# Changes
- Add an update method that starts a user data migration.
- Implement a skeleton migration that steps through the stages of the migration.

# Tests
- A test is included that verifies that the migration state does indeed advance as expected.


src/backend/tests/it/migration.rs
```diff
@@ -0,0 +1,279 @@
...
+    fn migration_state(&self) -> Option<MigrationReport> {
+        self.old_backend
+            .query::<Option<MigrationReport>>(controller(), "migration", ())
+            .expect("Failed to get migration report")
+    }
+    /// Steps the migration.
+    fn step_migration(&self) {
+        self.old_backend
+            .update::<()>(controller(), "step_migration", ())
+            .expect("Failed to stop migration tmer")
```
frederikrothenberger, line 106: This seems to be a misleading `expect` message.

src/backend/tests/it/migration.rs
```diff
@@ -0,0 +1,279 @@
...
+                to: pic_setup.new_backend.canister_id(),
+                progress: shared::types::MigrationProgress::Pending
+            }),
+        );
+        // Stop the timer so that we can control the migration.
+        assert_eq!(
+            pic_setup
+                .old_backend
+                .update::<Result<(), String>>(controller(), "migration_stop_timer", ())
+                .expect("Failed to stop migration tmer"),
```
frederikrothenberger, line 188:
```suggestion
                .expect("Failed to stop migration timer"),
```

frederikrothenberger (approved):
LGTM, only found a typo. 👍

PR #2059, opened 2024-08-19
refactor(frontend): generalize certify function for address

AntonioVentilii-DFINITY:
# Motivation

We generalize the `certifyAddress` function to be re-used for BTC.


src/frontend/src/lib/services/address.services.ts
```diff
@@ -17,6 +18,8 @@ import type { OptionIdentity } from '$lib/types/identity';
 import type { TokenId } from '$lib/types/token';
 import { replacePlaceholders } from '$lib/utils/i18n.utils';
 import type { BitcoinNetwork } from '@dfinity/ckbtc';
+import type { address } from '@dfinity/ckbtc/dist/candid/bitcoin';
```
peterpeterparker, line 21: This type is lowercase?

src/frontend/src/lib/services/address.services.ts
```diff
@@ -143,11 +146,17 @@ export const loadIdbAddress = async (): Promise<{ success: boolean }> => {
 	return { success: true };
 };
 
-export const certifyAddress = async (
-	address: string
-): Promise<{ success: boolean; err?: string }> => {
-	const tokenId = ETHEREUM_TOKEN_ID;
-
+const certifyAddress = async <T extends address>({
```
peterpeterparker, line 149: Not sure why `extends address` if this function is generic? Eth address does not extends adress or this is just for BTC?

src/frontend/src/lib/services/address.services.ts
```diff
@@ -157,23 +166,47 @@ export const certifyAddress = async (
...
 		}
 
-		const certifiedAddress = await getEthAddress(identity);
+		const certifiedAddress = await getAddress(identity);
 
 		if (address.toLowerCase() !== certifiedAddress.toLowerCase()) {
 			return {
 				success: false,
-				err: 'The address used to load the data did not match your actual wallet address, which is why your session was ended. Please sign in again to reload your own data.'
+				err: `The address used to load the data did not match your actual ${tokenId.description} wallet address, which is why your session was ended. Please sign in again to reload your own data.`
```
peterpeterparker, line 174: I guess it's already the plan but, just in case, could you extract this i18n in another PR?

src/frontend/src/lib/services/address.services.ts
```diff
@@ -157,23 +166,47 @@ export const certifyAddress = async (
...
 
 		addressStore.set({ tokenId, data: { data: address, certified: true } });
 
-		await updateIdbEthAddressLastUsage(identity.getPrincipal());
+		await updateIdbAddressLastUsage(identity.getPrincipal());
 	} catch (err: unknown) {
 		addressStore.reset(tokenId);
 
-		return { success: false, err: 'Error while loading the ETH address.' };
+		return { success: false, err: `Error while loading the ${tokenId.description} address.` };
```
peterpeterparker, line 184: same i18n key for error msg

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: no, you are right, it can be without extending, since it is generic

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: indeed I was going to do it in another PR, since there are actually more eheh

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: ditto

src/frontend/src/lib/services/address.services.ts
```diff
@@ -179,23 +187,43 @@ export const certifyAddress = async (address: string): Promise<ResultSuccess<str
...
 	}
 
 	return { success: true };
 };
+
+export const certifyBtcAddressMainnet = async (address: string): Promise<ResultSuccess<string>> =>
```
peterpeterparker, line 211: It's not type `Address`? or that's in another PR?

src/frontend/src/lib/services/address.services.ts
```diff
@@ -179,23 +187,43 @@ export const certifyAddress = async (address: string): Promise<ResultSuccess<str
...
+	certifyAddress({
+		tokenId: BTC_MAINNET_TOKEN_ID,
+		address,
+		getAddress: (identity) =>
```
peterpeterparker, line 215: `identity` has no type here? Typescript inherits it?

peterpeterparker (approved):
LGTM, thanks

Minor observations

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: right! good catch!

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: indeed it inherits it, that is why it didn't raise an error. But I will put it explicits just to be sure

PR #2060, opened 2024-08-19
refactor(frontend): create common return type for success

AntonioVentilii-DFINITY:
# Motivation

Just to avoid repetition of code and to optimize a little, we create a common type for the return of a successful (or not) function.

### DIFF
--- a/src/frontend/src/eth/services/balance.services.ts
+++ b/src/frontend/src/eth/services/balance.services.ts
@@ -10,11 +10,12 @@ import { toastsError } from '$lib/stores/toasts.store';
 import type { OptionEthAddress } from '$lib/types/address';
 import type { NetworkId } from '$lib/types/network';
 import type { Token, TokenId } from '$lib/types/token';
+import type { SuccessOrNot } from '$lib/types/utils';
 import { replacePlaceholders } from '$lib/utils/i18n.utils';
 import { isNullish } from '@dfinity/utils';
 import { get } from 'svelte/store';
 
-export const reloadBalance = async (token: Token): Promise<{ success: boolean }> => {
+export const reloadBalance = async (token: Token): Promise<SuccessOrNot> => {
 	if (isSupportedEthTokenId(token.id)) {
 		return loadBalance({ networkId: token.network.id, tokenId: token.id });
 	}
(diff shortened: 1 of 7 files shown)


src/frontend/src/lib/types/utils.ts
```diff
@@ -1 +1,3 @@
 export type RequiredExcept<T, K extends keyof T> = Required<Omit<T, K>> & Pick<T, K>;
+
+export type SuccessOrNot<T = unknown> = { success: boolean; err?: T };
```
peterpeterparker, line 3: There are few types that were defined on purpose with no `err`. Can we have two types?

For the types if `err?: unknown`, in another PR, maybe thinking that we can improve the type like `{success: true} | {success: false, err: unknown}` or something?

peterpeterparker (commented):
Cool idea. Two questions.

src/frontend/src/lib/types/utils.ts
peterpeterparker: Nvm my comment.

src/frontend/src/eth/services/balance.services.ts
```diff
@@ -10,11 +10,12 @@ import { toastsError } from '$lib/stores/toasts.store';
 import type { OptionEthAddress } from '$lib/types/address';
 import type { NetworkId } from '$lib/types/network';
 import type { Token, TokenId } from '$lib/types/token';
+import type { SuccessOrNot } from '$lib/types/utils';
 import { replacePlaceholders } from '$lib/utils/i18n.utils';
 import { isNullish } from '@dfinity/utils';
 import { get } from 'svelte/store';
 
-export const reloadBalance = async (token: Token): Promise<{ success: boolean }> => {
+export const reloadBalance = async (token: Token): Promise<SuccessOrNot> => {
```
peterpeterparker, line 18: Just an idea if you like it, instead of `SuccessOrNot` what about calling it `SomethingResult` or `ResultSomething`? Like `ResultSuccess`.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/eth/services/balance.services.ts
AntonioVentilii-DFINITY: naming! ahahaha! I like `ResultSuccess`, I will use that

PR #2063, opened 2024-08-19
feat(frontend): add BTC loaders to UX component

AntonioVentilii-DFINITY:
# Motivation

We join the BTC address loader with the ETH address loader.

# Changes

- Create function to load all IDB addresses.
- Create function to get addresses from backend by token.
- Make the `Loader` component await **ALL** loading promises of the above functions.
- Adjust the text and the comments.
- Unrelated: remove the useless `export` properties.

# Tests

Bitcoin address is correctly loaded (and/or re-loaded) during login.

### DIFF
--- a/src/frontend/src/lib/components/core/Loader.svelte
+++ b/src/frontend/src/lib/components/core/Loader.svelte
@@ -62,7 +62,7 @@
 	let progressModal = false;
 
 	onMount(async () => {
-		const { success: addressIdbSuccess, err: tokenIdsNotLoaded } = await loadIdbAddresses();
+		const { success: addressIdbSuccess, err } = await loadIdbAddresses();
 
 		if (addressIdbSuccess) {
 			loading.set(false);
@@ -75,7 +75,9 @@
 		// We are loading the BTC and ETH address from the backend. Consequently, we aim to animate this operation and offer the user an explanation of what is happening. To achieve this, we will present this information within a modal.
 		progressModal = true;
 
-		const { success: addressSuccess } = await loadAddresses(tokenIdsNotLoaded ?? []);
+		const { success: addressSuccess } = await loadAddresses(
+			err?.map(({ tokenId }) => tokenId) ?? []
+		);
 
 		if (!addressSuccess) {
 			await signOut();
--- a/src/frontend/src/lib/components/core/Loader.svelte
+++ b/src/frontend/src/lib/components/core/Loader.svelte
@@ -69,12 +69,24 @@
 		}
 	};
 
+	let idbBtcMainnetSuccess = false;
+	let idbEthSuccess = false;
+
+	$: {
+		if ($loading && idbBtcMainnetSuccess && idbEthSuccess) {
+			loading.set(false);
+		}
+	}
+
 	onMount(async () => {
 		const results = await Promise.all([
 			NETWORK_BITCOIN_ENABLED
-				? safeLoadBtcAddressMainnet(displayProgressModal)
+				? safeLoadBtcAddressMainnet({
+						displayProgressModal,
+						onIdbSuccess: () => (idbBtcMainnetSuccess = true)
+					})
 				: Promise.resolve({ success: true }),
-			safeLoadEthAddress(displayProgressModal)
+			safeLoadEthAddress({ displayProgressModal, onIdbSuccess: () => (idbEthSuccess = true) })
 		]);
 
 		if (results.every(({ success }) => success)) {
--- a/src/frontend/src/lib/components/core/Loader.svelte
+++ b/src/frontend/src/lib/components/core/Loader.svelte
@@ -63,14 +63,18 @@
 	let progressModal = false;
 
 	// We are loading the BTC and ETH address from the backend. Consequently, we aim to animate this operation and offer the user an explanation of what is happening. To achieve this, we will present this information within a modal.
-	const setProgressModal = () => (progressModal = true);
+	const displayProgressModal = () => {
+		if (!progressModal) {
+			progressModal = true;
+		}
+	};
 
 	onMount(async () => {
 		const results = await Promise.all([
 			NETWORK_BITCOIN_ENABLED
-				? safeLoadBtcAddressMainnet(setProgressModal)
+				? safeLoadBtcAddressMainnet(displayProgressModal)
 				: Promise.resolve({ success: true }),
-			safeLoadEthAddress(setProgressModal)
+			safeLoadEthAddress(displayProgressModal)
 		]);
 
 		if (results.every(({ success }) => success)) {
--- a/src/frontend/src/lib/components/core/Loader.svelte
+++ b/src/frontend/src/lib/components/core/Loader.svelte
@@ -2,7 +2,7 @@
 	import { Modal, type ProgressStep } from '@dfinity/gix-components';
 	import InProgress from '$lib/components/ui/InProgress.svelte';
 	import { onMount } from 'svelte';
-	import { loadEthAddress, loadIdbEthAddress } from '$lib/services/address.services';
+	import { safeLoadBtcAddressMainnet, safeLoadEthAddress } from '$lib/services/address.services';
 	import { fade } from 'svelte/transition';
 	import { signOut } from '$lib/services/auth.services';
 	import { loadErc20Tokens } from '$eth/services/erc20.services';
@@ -13,8 +13,9 @@
 	import { loadIcrcTokens } from '$icp/services/icrc.services';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { authStore } from '$lib/stores/auth.store';
+	import { NETWORK_BITCOIN_ENABLED } from '$env/networks.btc.env';
 
-	let progressStep: string = ProgressStepsLoader.ETH_ADDRESS;
+	let progressStep: string = ProgressStepsLoader.BTC_ETH_ADDRESS;
 
 	let steps: [ProgressStep, ...ProgressStep[]];
 	$: steps = [
@@ -24,8 +25,8 @@
 			state: 'completed'
 		} as ProgressStep,
 		{
-			step: ProgressStepsLoader.ETH_ADDRESS,
-			text: $i18n.init.text.retrieving_eth_key,
+			step: ProgressStepsLoader.BTC_ETH_ADDRESS,
+			text: $i18n.init.text.retrieving_public_keys,
 			state: 'in_progress'
 		} as ProgressStep
 	];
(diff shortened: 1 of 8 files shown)


peterpeterparker (commented):
I might be wrong but, at first glance it seems that two important pieces of the UX have been removed in this PR.

Currently, in `Loader`, then the addresses are read from Idb `onMount`, then `loading.set(false);` is called to notify Oisy is ready.

If not read from idb, then `progressModal = true;` to display the loading modal.

Was this behavior on purpose removed and if yes, can you describe what should be the new login experience for new and returning user?

AntonioVentilii-DFINITY:
- the `progressModal = true` will happen in the same case as before: when one of the two addresses is not loaded from IDB, but needs to be re-fetched from backend. I put it in the prop that I called `onIdbNotSuccessful`. That triggers the "progress" modal that says that we are retrieving the pub keys.

- the `loading.set(false)` was happening as soon as the address was loaded from IDB. Or with a 1 second lag in case we calculate from backend. I thought to make the less verbose change and remove the first one, and make it by default that it disappears with 1 second lag. Otherwise I can create a `onIdbSuccessful` and treat the case there. WDYT?

peterpeterparker:
> the progressModal = true will happen in the same case as before: when one of the two addresses is not loaded from IDB, but needs to be re-fetched from backend. I put it in the prop that I called onIdbNotSuccessful. That triggers the "progress" modal that says that we are retrieving the pub keys.

1. If I'm not wrong, if both Eth and Btc addresses are not loaded from Idb then in your implementation `progressModal = true` will be called twice. While visually speaking this might be not noticable. If that's correct, it would be cleaner to only call it once.

2. NTH: if would be nice to rename `setProgressModal` to e.g. `displayProgressModal` for readability reasons. Likewise I would preserve the name in the services, it's not easy to follow if the naming use in the component changes when pass in the services.

peterpeterparker:
> the loading.set(false) was happening as soon as the address was loaded from IDB. Or with a 1 second lag in case we calculate from backend. I thought to make the less verbose change and remove the first one, and make it by default that it disappears with 1 second lag. Otherwise I can create a onIdbSuccessful and treat the case there. WDYT?

I believe we should maintain the current implementation since it works fine. This PR should not change the existing behavior but simply introduce loading BTC in addition to ETH. Any potential improvements or modifications to the loading scheme should be submitted through separate PRs. This approach feels safer to me if you don't mind?

AntonioVentilii-DFINITY:
1. will correct that as such:

```typescript
	const setProgressModal = () => {
		if (!progressModal) {
			progressModal = true;
		}
	};
```

peterpeterparker:
> 1. and 2. will correct that as such:
> 	const displayProgressModal = () => {
>		if (!progressModal) {
>			progressModal = true;
>		}
>	};

Mmmh this feels like a shorthand I have to say. I rather like to have a clear expressive logic.

```
const allAddressesLoadedFromIdb = loadIdbAddress()

if (allAddressesLoadedFromIdb)
   return
   
// Explanation
progressModal = true
```

AntonioVentilii-DFINITY:

src/frontend/src/lib/components/core/Loader.svelte
```diff
@@ -62,7 +62,7 @@
 	let progressModal = false;
 
 	onMount(async () => {
-		const { success: addressIdbSuccess } = await loadIdbEthAddress();
+		const { success: addressIdbSuccess, err: tokenIdsNotLoaded } = await loadIdbAddresses();
```
AntonioVentilii-DFINITY, line 65: maybe `tokenAddressesNotLoaded` ?

src/frontend/src/lib/services/address.services.ts
```diff
@@ -120,21 +132,21 @@ const loadIdbTokenAddress = async <T extends Address>({
...
+}): Promise<ResultSuccess<TokenId>> => {
 	try {
 		const { identity } = get(authStore);
 
 		// Should not happen given the current layout and guards.
 		assertNonNullish(identity, 'Cannot continue without an identity.');
 
 		if (identity.getPrincipal().isAnonymous()) {
-			return { success: false };
+			return { success: false, err: tokenId };
```
peterpeterparker, line 143: Even though it is type as `unknown`, if you wish to use the `err` to return the token id, I would then suggest to create a class that extends `Error` and return it as it's content. This will make the usage more clear, particularly as the error is consumed and interpreted in the component, and also better typed when transpiled to JavaScript as well.

src/frontend/src/lib/services/address.services.ts
```diff
@@ -147,26 +159,51 @@ const loadIdbTokenAddress = async <T extends Address>({
...
+export const loadIdbAddresses = async (): Promise<ResultSuccess<TokenId[]>> => {
+	const results = await Promise.all([
+		NETWORK_BITCOIN_ENABLED ? loadIdbBtcAddressMainnet() : { success: true, err: null },
+		loadIdbEthAddress()
+	]);
+
+	let success = true;
+	const err: TokenId[] = [];
+
+	results.map(({ success: s, err: e }) => {
```
peterpeterparker, line 191: Can you resolve this in a functional non imperative way?

AntonioVentilii-DFINITY: right! better!

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: maybe I understood what you mean: please check last commit

src/frontend/src/lib/types/errors.ts
```diff
@@ -1 +1,13 @@
+import type { TokenId } from '$lib/types/token';
+
 export class UserProfileNotFoundError extends Error {}
+
+export class AddressError extends Error {
+	tokenId: TokenId;
+
+	constructor(tokenId: TokenId) {
+		super();
+		Object.setPrototypeOf(this, AddressError.prototype);
```
peterpeterparker, line 10: I honestly don't know what's the goal of `Object.setPrototypeOf...`. What's the oucome if I may ask for my knowledges?

src/frontend/src/lib/types/errors.ts
```diff
@@ -1 +1,13 @@
+import type { TokenId } from '$lib/types/token';
+
 export class UserProfileNotFoundError extends Error {}
+
+export class AddressError extends Error {
+	tokenId: TokenId;
+
+	constructor(tokenId: TokenId) {
```
peterpeterparker, line 8: Can it be simplied by just declaring it private?

```
export class AddressError extends Error {
   constructor(private readonly tokenId: TokenId) {
      super()
```

src/frontend/src/lib/services/address.services.ts
```diff
@@ -147,26 +160,43 @@ const loadIdbTokenAddress = async <T extends Address>({
...
 
-export const loadIdbEthAddress = async (): Promise<ResultSuccess> =>
+const loadIdbEthAddress = async (): Promise<ResultSuccess<AddressError>> =>
 	loadIdbTokenAddress<EthAddress>({
 		tokenId: ETHEREUM_TOKEN_ID,
 		getIdbAddress: getIdbEthAddress,
 		updateIdbAddressLastUsage: updateIdbEthAddressLastUsage
 	});
 
+export const loadIdbAddresses = async (): Promise<ResultSuccess<TokenId[]>> => {
```
peterpeterparker, line 183: Shouldn't we bubble `ResultSuccess<AddressError[]>` to the component? That way in the component `error` has also a strict type?

src/frontend/src/lib/types/errors.ts
```diff
@@ -1 +1,13 @@
+import type { TokenId } from '$lib/types/token';
+
 export class UserProfileNotFoundError extends Error {}
+
+export class AddressError extends Error {
```
peterpeterparker, line 5: Should we call this `LoadIdbAddressError`?

src/frontend/src/lib/types/errors.ts
AntonioVentilii-DFINITY: usually it is suggested to use it right after a `super()` statement: I have found a sort of technical explanation [here](https://stackoverflow.com/a/76851585), but reading better it says that it is indeed not needed for instances of `Error`

So I am removing it

src/frontend/src/lib/types/errors.ts
AntonioVentilii-DFINITY: yes, but then I need to create a getter to read it, right?


```typescript
	public getTokenId(): TokenId {
		return this.tokenId;
	}
```

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: for sure!

src/frontend/src/lib/types/errors.ts
AntonioVentilii-DFINITY: yes! better eheh!

src/frontend/src/lib/types/errors.ts
peterpeterparker: Ah gotcha. While a bit more verbose, it's still nice to have it as `readonly`? Maybe just adding that keyword then?

src/frontend/src/lib/types/errors.ts
peterpeterparker: Thanks for the explanation, did not knew. Somehow always expected the bundler to take care of those kind of `setPropotyOf` thingy. Good to know!

src/frontend/src/lib/types/errors.ts
AntonioVentilii-DFINITY: yes, agreed! changed

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/lib/services/address.services.ts
```diff
@@ -75,19 +77,30 @@ const loadBtcAddress = async ({
...
 
-export const loadEthAddress = async (): Promise<ResultSuccess> =>
+const loadEthAddress = async (): Promise<ResultSuccess> =>
 	loadTokenAddress<EthAddress>({
 		tokenId: ETHEREUM_TOKEN_ID,
 		getAddress: getEthAddress,
 		setIdbAddress: setIdbEthAddress
 	});
 
+export const loadAddresses = async (tokenIds: TokenId[]): Promise<ResultSuccess> => {
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: correct, because in this case we are not using the err propriety, so we don't define the type of err, and automatically it goes to `unknown`

src/frontend/src/lib/services/address.services.ts
```diff
@@ -147,26 +160,45 @@ const loadIdbTokenAddress = async <T extends Address>({
...
 
+export const loadIdbAddresses = async (): Promise<ResultSuccess<LoadIdbAddressError[]>> => {
+	const results = await Promise.all([
+		NETWORK_BITCOIN_ENABLED
+			? loadIdbBtcAddressMainnet()
+			: Promise.resolve<ResultSuccess<LoadIdbAddressError>>({ success: true }),
+		loadIdbEthAddress()
+	]);
+
+	const { success, err } = results.reduce(
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: yes, understood, make sense! will do some refactoring in another PR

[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Doing it in PR https://github.com/dfinity/oisy-wallet/pull/2120

PR #2066, opened 2024-08-20
refactor(frontend): common function to get address from data

AntonioVentilii-DFINITY:
# Motivation

We use a common function to get the addresses from the `addressStore`, to avoid duplication of code.


src/frontend/src/lib/utils/address.utils.ts
```diff
@@ -0,0 +1,13 @@
+import type { AddressData } from '$lib/stores/address.store';
+import type { CertifiedStoreData } from '$lib/stores/certified.store';
+import type { Address, OptionAddress } from '$lib/types/address';
+import type { TokenId } from '$lib/types/token';
+
+export const getNullableAddress = ({
```
peterpeterparker, line 6: The purpose of the function is not really to get a nullable address but rather to map the address regardless if certified or not certified. Do you have a better naming idea? Maybe just `mapAdress` or `addressData`?

AntonioVentilii-DFINITY: to keep consistent with similar ones (for example the ones related to tokens), I would call it `mapAddress` as you said (or `mapAddressData`)

peterpeterparker: Let's do that

AntonioVentilii-DFINITY: done

peterpeterparker (approved):
LGTM, thanks

PR #2067, opened 2024-08-20
feat(frontend): add BTC verification to AddressGuard

AntonioVentilii-DFINITY:
# Motivation

We include the BTC address verification in the `AddressGuard` component.


src/frontend/src/lib/components/guard/AddressGuard.svelte
```diff
@@ -1,33 +1,43 @@
...
+		ethAddress,
+		ethAddressNotCertified
+	} from '$lib/derived/address.derived';
+	import { NETWORK_BITCOIN_ENABLED } from '$env/networks.btc.env.js';
 
 	const validateAddress = async () => {
-		const ethAddressData = $addressStore?.[ETHEREUM_TOKEN_ID];
-
-		if (isNullish(ethAddressData)) {
-			// No address is loaded, we don't have to verify it
```
peterpeterparker, line 12: This check was removed?

src/frontend/src/lib/derived/address.derived.ts
```diff
@@ -9,24 +9,49 @@ export const addressNotLoaded: Readable<boolean> = derived([addressStore], ([$ad
...
+	[btcAddressMainnetData],
+	([$btcAddressMainnetData]) => $btcAddressMainnetData?.certified === true
+);
+
+export const btcAddressMainnetNotCertified: Readable<boolean> = derived(
+	[btcAddressMainnetCertified],
+	([$btcAddressMainnetCertified]) => !$btcAddressMainnetCertified
+);
+
+const ethAddressCertified: Readable<boolean> = derived(
```
peterpeterparker, line 43: Do we need global derived stores? According current implementation aren't we only checking for certification in a single component? If we do need one, can you provide the refactoring in a separate PR?

src/frontend/src/lib/components/guard/AddressGuard.svelte
```diff
@@ -1,33 +1,43 @@
...
+				? certifyBtcAddressMainnet($btcAddressMainnet)
+				: Promise.resolve({ success: true, err: null }),
+			nonNullish($ethAddress) && $ethAddressNotCertified
+				? certifyEthAddress($ethAddress)
+				: Promise.resolve({ success: true, err: null })
+		]);
+
+		let err: string | undefined = undefined;
+
+		results.map(({ success, err: e }) => {
```
peterpeterparker, line 32: Same as in previous PR, can we use a non imperative approach?

src/frontend/src/lib/components/guard/AddressGuard.svelte
```diff
@@ -1,33 +1,43 @@
...
 
-		const { success, err } = await certifyEthAddress(ethAddress);
-
-		if (success) {
-			// The address is valid
-			return;
+		const results = await Promise.all([
+			NETWORK_BITCOIN_ENABLED && nonNullish($btcAddressMainnet) && $btcAddressMainnetNotCertified
+				? certifyBtcAddressMainnet($btcAddressMainnet)
+				: Promise.resolve({ success: true, err: null }),
```
peterpeterparker, line 24: `err` is optional, specifying `null` is required for which reason?

src/frontend/src/lib/components/guard/AddressGuard.svelte
AntonioVentilii-DFINITY: i included it in the rest of the code, but inverted `if (nonnullish(....`

src/frontend/src/lib/derived/address.derived.ts
AntonioVentilii-DFINITY: not really needed to be honest... I just wanted to keep it more immediately clear, than just having the notCertified store. But it makes sense to reduce the stores. Correcting

src/frontend/src/lib/components/guard/AddressGuard.svelte
AntonioVentilii-DFINITY: on it

src/frontend/src/lib/components/guard/AddressGuard.svelte
AntonioVentilii-DFINITY: it is required when i deconstruct the promises results afterwards `{ success, err }`: if I remove it, the interpreter raises an error that `err` may be missing.

The alternative is to cast it:

`... : Promise.resolve({ success: true } as ResultSuccess<string>)`

src/frontend/src/lib/components/guard/AddressGuard.svelte
AntonioVentilii-DFINITY: actually, let me leave it there

src/frontend/src/lib/components/guard/AddressGuard.svelte
AntonioVentilii-DFINITY: perfect! thank you!

AntonioVentilii-DFINITY:

src/frontend/src/lib/components/guard/AddressGuard.svelte
```diff
@@ -1,29 +1,49 @@
...
+		const results = await Promise.all([
+			NETWORK_BITCOIN_ENABLED && nonNullish($btcAddressMainnet) && $btcAddressMainnetNotCertified
+				? certifyBtcAddressMainnet($btcAddressMainnet)
+				: Promise.resolve<ResultSuccess<never>>({ success: true }),
+			nonNullish($ethAddress) && $ethAddressNotCertified
+				? certifyEthAddress($ethAddress)
+				: Promise.resolve<ResultSuccess<never>>({ success: true })
+		]);
+
+		const { success, err } = results.reduce(
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: we use it after in case of non success:

```typescript
if (success) {
	// The addresses are valid
	return;
}

```

src/frontend/src/lib/derived/address.derived.ts
```diff
@@ -9,24 +9,33 @@ export const addressNotLoaded: Readable<boolean> = derived([addressStore], ([$ad
 	isNullish($addressStore)
 );
 
-export const btcAddressMainnet: Readable<OptionBtcAddress> = derived(
+export const btcAddressMainnetData: Readable<AddressData | null | undefined> = derived(
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: will do!

src/frontend/src/lib/components/guard/AddressGuard.svelte
```diff
@@ -1,29 +1,49 @@
...
+		if (isNullish($btcAddressMainnetData) && isNullish($ethAddressData)) {
 			// No address is loaded, we don't have to verify it
 			return;
 		}
 
-		const { certified, data: ethAddress } = ethAddressData;
-
-		if (certified === true) {
-			// The address is certified, all good
+		if (!$btcAddressMainnetNotCertified && !$ethAddressNotCertified) {
```
peterpeterparker, line 23: Oh! Should we consider the flag `NETWORK_BITCOIN_ENABLED` here or we gonna have an issue until enabled?

src/frontend/src/lib/components/guard/AddressGuard.svelte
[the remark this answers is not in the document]
peterpeterparker: Ok. In this or another PR, can we have a function for this? I think there are two similar reduce over the results in recent PRs.

src/frontend/src/lib/components/guard/AddressGuard.svelte
AntonioVentilii-DFINITY: absolutely! good catch! thankyou

src/frontend/src/lib/components/guard/AddressGuard.svelte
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Will do!

src/frontend/src/lib/components/guard/AddressGuard.svelte
```diff
@@ -1,29 +1,49 @@
...
+		if (isNullish($btcAddressMainnetData) && isNullish($ethAddressData)) {
 			// No address is loaded, we don't have to verify it
 			return;
 		}
 
-		const { certified, data: ethAddress } = ethAddressData;
-
-		if (certified === true) {
-			// The address is certified, all good
+		if ((!NETWORK_BITCOIN_ENABLED || !$btcAddressMainnetNotCertified) && !$ethAddressNotCertified) {
```
peterpeterparker, line 23: this hurts my brain 🤪. in another PR maybe we should transform it into a "positive flag check" assertion?

src/frontend/src/lib/components/guard/AddressGuard.svelte
```diff
@@ -1,29 +1,49 @@
...
+			// The addresses are certified, all good
 			return;
 		}
 
-		const { success, err } = await certifyEthAddress(ethAddress);
+		const results = await Promise.all([
```
peterpeterparker, line 28: Reading the code again. I'm a bit concerned about triggering too many update calls. Isn't there a race condition here in which we might trigger twice for example `certifyEthAddress`? If yes, would it be maybe safer to split both checks?

```
$: $ethAddressSomething, (async () => await validateEthAddress())();
$: $btcAddressSomething, (async () => await validateBtcAddress())();
```

Of course it there is no such issue we do not need it. I'm just being really precaucious here.

AntonioVentilii-DFINITY: ah! understood you point, make sense! however, I put all addresses in the same store: `ethAddressSomething` and `btcAddressSomething` will always be derived from there and triggered together for any change in the addresses store...

Shall I change then and create different stores for different addresses? I can put this PR on hold and do that first in another PR

src/frontend/src/lib/components/guard/AddressGuard.svelte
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: doing in PR https://github.com/dfinity/oisy-wallet/pull/2120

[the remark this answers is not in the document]
AntonioVentilii-DFINITY: adapted

src/frontend/src/lib/components/guard/AddressGuard.svelte
AntonioVentilii-DFINITY: @peterpeterparker WDYT of my comment above?

peterpeterparker: > will always be derived from there and triggered together for any change in the addresses store...

I actually have a doubt. If you have a store and two derived states, derivedA and derivedB, are they both triggered when the store changes, or only when their values change?

peterpeterparker: Ah but we do not have a condition here that would trigger only derivedA or derivedB.

peterpeterparker: AH in any case, looking at the code again, there is something else that does not work neither.

`validateAddress` is triggered when `$addressStore` changes but within the function derived store are used. Therefore there is a chance that the derived store are not populated yet with the value derived. It's an almost impossible race condition but, it's one.

peterpeterparker: So to answer your question, I feel like you are right, we might need different stores. I don't see right now another solution but, if you have an idea, I would be really happy to hear it!!!!

Regarding the race condition, we can then do something as following:

```
const validateAddress = async (data: OptionAddressData) => {
     // Same logic of the function as on main but using data parameter
}

$: $btcAddressMainnetData, (async () => await validateAddress($btcAddressMainnetData))();
$: $ethAddressMainnetData, (async () => await validateAddress($ethAddressMainnetData))();
```

AntonioVentilii-DFINITY: Yes, I don't see any other solution honestly. I will split that in several PR:

- PR #2207

AntonioVentilii-DFINITY:
As per @peterpeterparker suggestion, we split the store into different ones (see PRs list in comments above). To have a clean history, this PR will be superseded by PR #

PR #2068, opened 2024-08-20
style(frontend): change conversion icon

AntonioVentilii-DFINITY:
# Motivation

Updated conversion icon.

Screenshot 2024-08-20 at 12 09 00


peterpeterparker (approved):
LGTM, thanks

The `IconBurn` is still use somewhere?

AntonioVentilii-DFINITY:
> The IconBurn is still use somewhere?

For now it is used in the transaction list, when the conversion is pending, but I think we are going to remove it and use this one too. However, I wanted to leave it there even when not used, so that I can verify that the remove-unused-components script works correctly during the weekend.

PR #2072, opened 2024-08-20
fix(backend): Rm erroneous expect

bitdivine:
# Motivation
When "prettifying" a recent PR before merge I added `.expect()` lines after data insertion.  In fact, the `.insert()` returns not a result but an option containing the previous value, so all the (still to be merged) tests that use these functions fail.  A bit embarrassing!

# Changes
- Don't expect() on the returned options.


frederikrothenberger (approved):
It happens to the best of us 😉

PR #2073, opened 2024-08-20
style(frontend): use IconConvert with pending conversion transactions

AntonioVentilii-DFINITY:
# Motivation

We prefer to display `IconConvert` while the conversion transaction is pending.


src/frontend/src/icp/components/transactions/IcTransaction.svelte
```diff
@@ -36,9 +37,13 @@
 
 	let icon: ComponentType;
 	$: icon = ['burn', 'approve'].includes(transactionType)
-		? IconBurn
+		? pending
```
peterpeterparker, line 40: Instead of duplicating `pending` shouldn't it be the first criteria?

```
$: icon = pending ? IconConvert : same_as_on_main...
```

AntonioVentilii-DFINITY: yes and no, since it should be only for `burn`, `approve` and `mint` (and not `send` and `receive`). So it would be something like

```typescript
	$: icon =
		['burn', 'approve', 'mint'].includes(transactionType) && pending
			? IconConvert
			: ['burn', 'approve'].includes(transactionType)
				? IconBurn
				: transactionType === 'mint'
					? IconMint
					: incoming === false
						? IconSend
						: IconReceive;
```

WDYT?

peterpeterparker: Sounds good.

peterpeterparker (approved):
LGTM, thx

peterpeterparker (approved):

PR #2074, opened 2024-08-20
feat(backend): Update migration states

bitdivine:
# Motivation
Testing migrations showed that:
- Some naming was potentially confusing.
- When testing error conditions, it would be useful to have a formal error variant for the migration state.

# Changes
- Modify the (as yet unused in main) migration state enum:
  - Some steps have been renamed or reorganized for clarity.
  - A failed state has been added, with various reasons for failure.

- Updated code to match

# Tests
Tests have been updated to match the changes.

### DIFF
--- a/src/backend/backend.did
+++ b/src/backend/backend.did
@@ -79,17 +79,30 @@ type ListUsersResponse = record {
   users : vec OisyUser;
   matches_max_length : nat64;
 };
+type MigrationError = variant {
+  TargetLockFailed;
+  TargetUnlockFailed;
+  CouldNotGetTargetPostStats;
+  CouldNotGetTargetPriorStats;
+  DataMigrationFailed;
+  TargetStatsMismatch : record { Stats; Stats };
+  Unknown;
+  TargetCanisterNotEmpty : Stats;
+  NoMigrationInProgress;
+};
 type MigrationProgress = variant {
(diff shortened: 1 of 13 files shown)


src/backend/backend.did
```diff
@@ -79,17 +79,30 @@ type ListUsersResponse = record {
   users : vec OisyUser;
   matches_max_length : nat64;
```
bitdivine, line 80: This is the updated list of migration progress states, including failure reasons.

src/backend/src/migrate.rs
```diff
@@ -77,54 +77,61 @@ pub async fn step_migration() {
         });
     }
```
bitdivine, line 78: Updated the match statement to reflect the new enum variants.

Note that this is placeholder code, waiting for the actual migration code.

Also factored out the repeated `set_progress()`.  The `next()` cannot be factored out because it will be different in some branches.

src/backend/tests/it/migration.rs
```diff
@@ -194,24 +194,19 @@ fn test_migration() {
         // Migration should be in progress.
```
bitdivine, line 194: Updated the test that checks the migration proceeds as planned.

AntonioVentilii-DFINITY (approved):

PR #2075, opened 2024-08-20
feat(backend): add BTC balance call

AntonioVentilii-DFINITY:
# Motivation

We add the backend call to fetch the BTC balance of a generic address, based on the [IC documentation](https://internetcomputer.org/docs/current/developer-docs/multi-chain/bitcoin/using-btc/read-state#reading-current-balance).

### DIFF
--- a/src/backend/src/bitcoin_lib.rs
+++ b/src/backend/src/bitcoin_lib.rs
@@ -5,7 +5,7 @@ use ic_cdk::api::management_canister::bitcoin::{
 /// Returns the balance of the given bitcoin address.
 ///
 /// Relies on the `bitcoin_get_balance` endpoint.
-/// See https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_get_balance
+/// See [IC Bitcoin Documentation](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_get_balance)
 pub async fn get_balance(network: BitcoinNetwork, address: String) -> u64 {
     let balance_res = bitcoin_get_balance(GetBalanceRequest {
         address,
--- a/src/backend/src/bitcoin_lib.rs
+++ b/src/backend/src/bitcoin_lib.rs
@@ -1,0 +1,18 @@
+use ic_cdk::api::management_canister::bitcoin::{
+    bitcoin_get_balance, BitcoinNetwork, GetBalanceRequest,
+};
+
+/// Returns the balance of the given bitcoin address.
+///
+/// Relies on the `bitcoin_get_balance` endpoint.
+/// See https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_get_balance
+pub async fn get_balance(network: BitcoinNetwork, address: String) -> u64 {
+    let balance_res = bitcoin_get_balance(GetBalanceRequest {
+        address,
+        network,
+        min_confirmations: None,
+    })
+    .await;
+
+    balance_res.unwrap().0
+}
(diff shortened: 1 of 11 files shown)


src/backend/src/lib.rs
```diff
@@ -265,6 +266,12 @@ async fn caller_btc_address(network: BitcoinNetwork) -> String {
     public_key_to_p2pkh_address(network, &ecdsa_pubkey_of(&ic_cdk::caller()).await)
 }
 
+/// Returns the balance of the given Bitcoin address.
+#[query]
```
peterpeterparker, line 270: Public is fine?

src/backend/src/lib.rs
```diff
@@ -265,6 +266,12 @@ async fn caller_btc_address(network: BitcoinNetwork) -> String {
...
+async fn btc_balance(address: String, network: BitcoinNetwork) -> u64 {
+    bitcoin_lib::get_balance(network, address).await
```
peterpeterparker, line 272: nth: just `get_balance` and declare the `use` at the top for consistency?

src/backend/src/bitcoin_lib.rs
```diff
@@ -0,0 +1,18 @@
...
+/// See [IC Bitcoin Documentation](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_get_balance)
+pub async fn get_balance(network: BitcoinNetwork, address: String) -> u64 {
+    let balance_res = bitcoin_get_balance(GetBalanceRequest {
+        address,
+        network,
+        min_confirmations: None,
+    })
+    .await;
+
+    balance_res.unwrap().0
```
peterpeterparker, line 17: Shouldn't we properly catch the error or bubble the error to the caller?

src/backend/src/bitcoin_lib.rs
```diff
@@ -0,0 +1,18 @@
...
+
+/// Returns the balance of the given bitcoin address.
+///
+/// Relies on the `bitcoin_get_balance` endpoint.
+/// See [IC Bitcoin Documentation](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_get_balance)
+pub async fn get_balance(network: BitcoinNetwork, address: String) -> u64 {
+    let balance_res = bitcoin_get_balance(GetBalanceRequest {
+        address,
+        network,
+        min_confirmations: None,
```
peterpeterparker, line 13: None is correct?

src/backend/src/lib.rs
peterpeterparker, line 270: Maybe a stupid question: Query is correct? `bitcoin_get_balance` isn't an inter-canister update call?

peterpeterparker (commented):
OH! Another important question maybe: why do we need that call in the backend of Oisy? Couldn't we just query the balance from the frontend as we already do for utxos in `bitcoin.api.ts`?

src/backend/src/lib.rs
AntonioVentilii-DFINITY: yes, agreed!

peterpeterparker (changes requested):
As discussed with @AntonioVentilii-DFINITY, we do not need this PR as the calls can be made from the frontend. Antonio has started implementing this in ic-js (https://github.com/dfinity/ic-js/pull/701). So, just blocking the PR so that Max do not take time to review it as well.

AntonioVentilii-DFINITY:
closing (see @peterpeterparker comment above)

peterpeterparker:
Turns out this was incorrect. Query can indeed be made from the frontend but, update calls requires a backend call.

AntonioVentilii-DFINITY:
Re-opened with PR https://github.com/dfinity/oisy-wallet/pull/2184

PR #2076, opened 2024-08-20
style(frontend): remove space between fee label and value

AntonioVentilii-DFINITY:
# Motivation

There was a gap between the label and the value of component `EthereumEstimatedFee` that was not disappearing when the fee was loaded.

### Before

Screenshot 2024-08-20 at 19 37 59

### After

Screenshot 2024-08-20 at 19 38 19


peterpeterparker (approved):
LGTM, thx

PR #2077, opened 2024-08-21
test(e2e): add canister ids as env vars for e2e tests

DenysKarmazynDFINITY:
# Motivation

The goal is to access canister_e2e_ids.json in e2e tests as env vars.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -1,11 +1,11 @@
 import { defineConfig, devices } from '@playwright/test';
 import dotenv, { type DotenvPopulateInput } from 'dotenv';
 import { join } from 'node:path';
-import { readCanisterIdsFromJSONFile } from './shared.utils';
+import { readCanisterIds } from './env.utils';
 
 dotenv.populate(
 	process.env as DotenvPopulateInput,
-	readCanisterIdsFromJSONFile({
+	readCanisterIds({
 		filePath: join(process.cwd(), 'canister_e2e_ids.json'),
 		prefix: 'E2E_'
 	})
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -1,4 +1,15 @@
 import { defineConfig, devices } from '@playwright/test';
+import dotenv, { type DotenvPopulateInput } from 'dotenv';
+import { join } from 'node:path';
+import { readCanisterIdsFromJSONFile } from './shared.utils';
+
+dotenv.populate(
+	process.env as DotenvPopulateInput,
+	readCanisterIdsFromJSONFile({
+		filePath: join(process.cwd(), 'canister_e2e_ids.json'),
+		prefix: 'E2E_'
+	})
+);
 
 const DEV = (process.env.NODE_ENV ?? 'production') === 'development';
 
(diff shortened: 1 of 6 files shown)


AntonioVentilii-DFINITY (approved):
LGTM!

tsconfig.eslint.json
```diff
@@ -9,6 +9,7 @@
 		"./src/**/*.js",
 		"./src/**/*.ts",
 		"./src/**/*.svelte",
+		"./shared.utils.ts",
```
peterpeterparker, line 12: `shared.utils` feels a bit like reading `utils.utils` 😄. What about `env.utils.ts` or `canister-ids.utils.ts`?

playwright.config.ts
```diff
@@ -1,4 +1,15 @@
 import { defineConfig, devices } from '@playwright/test';
+import dotenv, { type DotenvPopulateInput } from 'dotenv';
+import { join } from 'node:path';
+import { readCanisterIdsFromJSONFile } from './shared.utils';
+
+dotenv.populate(
+	process.env as DotenvPopulateInput,
```
peterpeterparker, line 7: the type cast is a must?

shared.utils.ts
```diff
@@ -0,0 +1,45 @@
+import { readFileSync } from 'node:fs';
+
+/**
+ * Read a JSON file with canister IDs and transform it into a dictionary. Optionally add prefix to all dictionary keys.
+ */
+export const readCanisterIdsFromJSONFile = ({
```
peterpeterparker, line 6: nth: can be shortened `readCanisterIds`. From IC dev perspective, it's the standard to read ids from a file

shared.utils.ts
```diff
@@ -0,0 +1,45 @@
...
+	filePath,
+	prefix
+}: {
+	filePath: string;
+	prefix?: string;
+}): Record<string, string> => {
+	try {
```
peterpeterparker, line 13: Nothing changed in the logic of this function correct? Simply refactoring and it does need a review?

tsconfig.eslint.json
DenysKarmazynDFINITY: `env.utils.ts` it is 👍

shared.utils.ts
DenysKarmazynDFINITY: Agreed.

shared.utils.ts
DenysKarmazynDFINITY: Yep, I only extracted this function into a separate file.

playwright.config.ts
DenysKarmazynDFINITY: This cast is needed to align global env type (where each value is `string | undefined`) with the dotenv.populate function param (`DotenvPopulateInput`, where each value is `string`). It should be safe to cast, as even if some var doesn't have a value in an env file, it still resolves to an empty string in `process.env`.

peterpeterparker: Alright. Thanks for the feedback.

peterpeterparker (approved):
LGTM, thx

PR #2085, opened 2024-08-21
test(e2e): about modals tests

DenysKarmazynDFINITY:
# Motivation

About-how and about-what snapshots.


e2e/about-how-modal.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+	await homepageLoggedOut.setViewportSize({
+		width: ABOUT_MODALS_VIEWPORT_WIDTH,
+		height: ABOUT_HOW_MODAL_VIEWPORT_HEIGHT
+	});
+	await homepageLoggedOut.waitForReady();
+
+	const aboutHowModal = await homepageLoggedOut.waitForAboutHowModal();
+
+	await expect(aboutHowModal).toHaveScreenshot();
+});
```
AntonioVentilii-DFINITY, line 21: is it possible to create a common test that takes as input width, height and waitForAboutXXXModal? to try and reduce the code

e2e/utils/constants/e2e.constants.ts
```diff
@@ -1,3 +1,7 @@
 export const HOMEPAGE_URL = '/';
 
 export const LOCAL_REPLICA_URL = 'http://127.0.0.1:4943';
+
+export const ABOUT_MODALS_VIEWPORT_WIDTH = 800;
+export const ABOUT_WHAT_MODAL_VIEWPORT_HEIGHT = 930;
+export const ABOUT_HOW_MODAL_VIEWPORT_HEIGHT = 1600;
```
AntonioVentilii-DFINITY, line 7: out of curiosity, why did you define `ABOUT_WHAT_MODAL_VIEWPORT_HEIGHT` and `ABOUT_HOW_MODAL_VIEWPORT_HEIGHT` as external constants and not directly in the test? will they be re-used in the future? or are already used somewhere I missed?

DenysKarmazynDFINITY: Indeed, moved the height values to the related files. Though, the width const is still in `e2e.constants.ts` just so it can be re-used by both modals.

e2e/about-how-modal.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { test } from '@playwright/test';
+import {
+	ABOUT_HOW_MODAL,
+	ABOUT_HOW_MODAL_OPEN_BUTTON
+} from '../src/frontend/src/lib/constants/test-ids.constant';
+import { ABOUT_MODALS_VIEWPORT_WIDTH } from './utils/constants/e2e.constants';
+import { HomepageLoggedOut } from './utils/pages/homepage.page';
+
+export const ABOUT_HOW_MODAL_VIEWPORT_HEIGHT = 1600;
```
AntonioVentilii-DFINITY, line 9: maybe remove the `export` here?

e2e/about-how-modal.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+
+test('should display about-how modal', async ({ page }) => {
+	const homepageLoggedOut = new HomepageLoggedOut({ page });
+
+	await homepageLoggedOut.takeModalSnapshot({
```
AntonioVentilii-DFINITY, line 14: since it is already doing the test in it, I would call it something like `testModalSnapshot`, WDYT?

BTW, I was thinking something different with my suggestion: instead of being a function of `HomepageLoggedOut`, it would have been a common test, that takes as input the page too. But you can refactor in another PR if you want, or still this one? WDYT?

e2e/utils/pages/homepage.page.ts
```diff
@@ -42,6 +66,25 @@ abstract class Homepage {
 		await this.#page.getByTestId(TOKENS_SKELETONS_INITIALIZED).waitFor();
 	}
 
+	async takeModalSnapshot({
+		viewportSize,
+		modalOpenButtonTestId,
+		modalTestId
+	}: TakeModalSnapshotParams): Promise<void> {
+		if (viewportSize) {
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Nice, will be using them

e2e/about-how-modal.spec.ts
DenysKarmazynDFINITY: Agree, `testModalSnapshot` makes more sense in this case, renamed.

Regarding your second point - as soon as we start creating tests for other pages (e.g. `SettingsPage`), we'll need to transform the current `Homepage` class into something even more general, so all future pages can inherit it. The modal functionality will also go into this parent class. If we want to separate Modal from Page completely, it would mean we need to pass not only e.g. `HomepageLoggedOut` instance (so we have access to its custom methods like `waitFor`) but also playwright `Page` object (so we can access DOM-related methods, e.g. `getByTestId`). 

So, I suggest we keep the current structure for now and update it if needed as the e2e tests codebase grows.

AntonioVentilii-DFINITY: perfect! agreed!

AntonioVentilii-DFINITY (approved):

DenysKarmazynDFINITY:
The pipeline fails due to a missing signer canister id value (the thread in Slack: https://dfinity.slack.com/archives/C05M7P123HR/p1724313699569529). I will merge the code so I can re-use it in my other branches and update the snapshots in a separate PR after the backend deployment works w/o issues.

PR #2087, opened 2024-08-21
feat(frontend): add BTC price to exchange store

AntonioVentilii-DFINITY:
# Motivation

Since it is already being fetched, the Bitcoin Mainnet price will be added to the `exchanges` store.


peterpeterparker (approved):
LGTM, thanks

PR #2069, opened 2024-08-20
fix(frontend): pending ck conversions displayed under incorrect token

peterpeterparker:
# Motivation

It was reported that pending ck conversions were displayed under the wrong token. e.g. one user converted LINK to ckLINK and then browsed ckUSDC and the frontend was displaying "Pending USDC to ckUSDC". This seems to happens because the pending transactions resolved with logs are not filtered for the selected token.

To solve it, we refine the logs topic including the ERC20 contract address among the parameters.

We can see a practical example [here](https://etherscan.io/tx/0xddbf4aebe292c3d3b33240496db12a1a5622c4edb63017ebcd12ffa32fa5b924#eventlog).

Screenshot 2024-08-20 at 16 14 30

PR #2091, opened 2024-08-21
fix(frontend): revert PR #2069

AntonioVentilii-DFINITY:
Reverts dfinity/oisy-wallet#2069

It causes an issue:

Screenshot 2024-08-21 at 14 54 28


peterpeterparker (approved):
LGTM, thanks

We tested this on staging on which we were not able the issue that was reported anymore.

PR #2094, opened 2024-08-22

sa-github-api:


peterpeterparker:
It's false-positive. TRAX is out of cycles again. It's been 10 days since I tried to reach them out.

I ran the script locally but, it ends in error. Not sure why here it updated the json. I gonna close the PR and if it ever happens again try to dig deeper.

PR #2098, opened 2024-08-22
ci(e2e): improve e2e workflow

DenysKarmazynDFINITY:
# Motivation

The goal is to avoid the creation of an e2e docker image on every pull request. Instead, we can re-use artifacts between the workflows. 

The next steps are:
1. Update snapshots inside a PR in case e2e failed with related error.
2. Enable e2e tests on all PRs that contain changed frontend files.
3. Enable e2e tests for all PRs with rebuild of the artifact if backend files were changed.

### DIFF
--- a/.github/workflows/e2e-tests-build-docker-image.yml
+++ b/.github/workflows/e2e-tests-build-docker-image.yml
@@ -1,0 +1,16 @@
+name: Docker Image Builder For E2E Tests
+
+on:
+  workflow_dispatch:
+
+jobs:
+  docker-build-e2e:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Build E2E Docker Image
+        uses: ./.github/actions/docker-build-e2e
+        with:
+          name: oisy-e2e-image
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -8,32 +8,19 @@ on:
 
 jobs:
 
-  docker-build-e2e:
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Delete unnecessary tools folder to prevent "out of space" error
-        run: rm -rf /opt/hostedtoolcache
-
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Build E2E Docker Image
-        uses: ./.github/actions/docker-build-e2e
-        with:
-          name: oisy-e2e-image
-
   e2e:
     runs-on: ubuntu-latest
-    needs: docker-build-e2e
     steps:
       - name: Checkout
         uses: actions/checkout@v4
 
       - name: Download E2E Docker image artifact
-        uses: actions/download-artifact@v4
+        id: download-artifact
+        uses: dawidd6/action-download-artifact@v6
         with:
-          name: oisy-e2e-image
-          path: .
+          workflow: e2e-tests-build-docker-image.yml
+          branch: main
+          if_no_artifact_found: fail
 
       - name: Load Docker image
         run: |
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -19,6 +19,7 @@ jobs:
         uses: dawidd6/action-download-artifact@v6
         with:
           workflow: e2e-tests-build-docker-image.yml
+          name: oisy-e2e-image
           branch: main
           if_no_artifact_found: fail
 
--- a/.github/workflows/update-snapshots.yml
+++ b/.github/workflows/update-snapshots.yml
@@ -5,26 +5,6 @@ on:
 
 jobs:
 
-  docker-build-e2e:
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Fail if branch is main
-        if: ${{ github.ref == 'refs/heads/main' }}
-        run: |
-          echo "This workflow should not be triggered with workflow_dispatch on main"
-          exit 1
-
-      - name: Delete unnecessary tools folder to prevent "out of space" error
-        run: rm -rf /opt/hostedtoolcache
-
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Build E2E Docker Image
-        uses: ./.github/actions/docker-build-e2e
-        with:
-          name: oisy-e2e-image
-
   update_snapshots:
     runs-on: ubuntu-latest
     needs: docker-build-e2e
@@ -39,10 +19,11 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Download E2E Docker image artifact
-        uses: actions/download-artifact@v4
+        uses: dawidd6/action-download-artifact@v6
         with:
-          name: oisy-e2e-image
-          path: .
+          workflow: e2e-tests-build-docker-image.yml
+          branch: main
+          if_no_artifact_found: fail
 
       - name: Load Docker image
         run: |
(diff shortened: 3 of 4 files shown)


DenysKarmazynDFINITY:
Hey @bitdivine @AntonioVentilii-DFINITY . I’ve been looking today into how to share cached docker images between jobs, but it seems that it’s not possible (source: [docker docs](https://docs.docker.com/build/ci/github-actions/share-image-jobs/)). So instead, I tried a different approach: sharing artifacts. In this PR, you can see my idea. We would have a separate workflow (`e2e-tests-build-docker-image.yml`) responsible for building e2e docker images and exposing them as an artifact. As the first step, this workflow is triggered only on `workflow_dispatch`, but later it can/will be extended to being launched on `pull_request` when backend files are changed. The updated `e2e_tests.yml` workflow is therefore now only responsible for downloading the artifact and running e2e tests against it. WDYT?

.github/workflows/e2e-tests.yml
```diff
@@ -8,32 +8,20 @@ on:
 
 jobs:
 
-  docker-build-e2e:
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Delete unnecessary tools folder to prevent "out of space" error
-        run: rm -rf /opt/hostedtoolcache
```
AntonioVentilii-DFINITY, line 15: this step is deprecated? if yes, why? I see that you didn't put it in the new workflow

.github/workflows/update-snapshots.yml
```diff
@@ -5,26 +5,6 @@ on:
...
-    steps:
-      - name: Fail if branch is main
-        if: ${{ github.ref == 'refs/heads/main' }}
-        run: |
-          echo "This workflow should not be triggered with workflow_dispatch on main"
-          exit 1
-
-      - name: Delete unnecessary tools folder to prevent "out of space" error
-        run: rm -rf /opt/hostedtoolcache
-
```
AntonioVentilii-DFINITY, line 19: the 2 steps above are deprecated? if yes, why?

.github/workflows/e2e-tests-build-docker-image.yml
```diff
@@ -0,0 +1,16 @@
...
+  docker-build-e2e:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Build E2E Docker Image
+        uses: ./.github/actions/docker-build-e2e
+        with:
+          name: oisy-e2e-image
```
AntonioVentilii-DFINITY, line 16: isn't the upload of the image missing here? or am i missing something? honest question

.github/workflows/e2e-tests.yml
```diff
@@ -8,32 +8,20 @@ on:
...
     runs-on: ubuntu-latest
-    needs: docker-build-e2e
     steps:
       - name: Checkout
         uses: actions/checkout@v4
 
       - name: Download E2E Docker image artifact
-        uses: actions/download-artifact@v4
+        id: download-artifact
+        uses: dawidd6/action-download-artifact@v6
```
AntonioVentilii-DFINITY, line 19: the difference between `actions/download-artifact@v4` and `dawidd6/action-download-artifact@v6`, for what I understand [in the docs](https://github.com/dawidd6/action-download-artifact), is the possibility to pass an uploaded artifacts from a workflow to the other, correct? isn't there any "official" workflow to rely upon?

I would be careful with using "unofficial" packages/workflows; I will remit to @bitdivine and @peterpeterparker regarding this; I am not really sure what is our policy here, since it is the first time for me too

.github/workflows/e2e-tests.yml
DenysKarmazynDFINITY: Good catch, I will move this step to the docker-build-e2e workflow.

.github/workflows/e2e-tests-build-docker-image.yml
DenysKarmazynDFINITY: This is where an artifact (an archive with e2e docker image) is uploaded - https://github.com/dfinity/oisy-wallet/blob/ci(e2e)/separate-e2e-workflows/.github/actions/docker-build-e2e/action.yml#L44

.github/workflows/e2e-tests.yml
DenysKarmazynDFINITY: Yes, `dawidd6/action-download-artifact@v6` allows to download artifacts created in separate workflows (with `actions/download-artifact@v4` you can only do it within a single workflow run). Unfortunately, there is no official action for that, GitHub suggests to use its REST API to do so. So, in the end `dawidd6/action-download-artifact@v6` is just a wrapper around official API endpoint.

We already use at least one unofficial action (https://github.com/dfinity/oisy-wallet/blob/ci(e2e)/separate-e2e-workflows/.github/actions/create-pr/action.yml#L23), but in case you have doubts in `dawidd6/action-download-artifact@v6` please let me know.

.github/workflows/e2e-tests-build-docker-image.yml
AntonioVentilii-DFINITY: ah! thank you for the explanation! I had missed that part

.github/workflows/e2e-tests.yml
AntonioVentilii-DFINITY: I sincerely have not strong opinion on using it or not, so, I would wait for Max and David to give their input

.github/workflows/update-snapshots.yml
AntonioVentilii-DFINITY: what about the two steps above? `Fail if branch is main` and `Delete unnecessary...`

DenysKarmazynDFINITY: `Delete unnecessary...`  - itis now a part of `docker-build-e2e`, so we don't need to have it here.
`Fail if branch is main` - there is the same check in `update_snapshots` job (a bit below in this file), so the `update-snapshots` is still disabled for `main` as it was before

AntonioVentilii-DFINITY: perfect! thank you

DenysKarmazynDFINITY: With this particular PR, I don't touch the screenshot updating part. IMO we first need to agree on the structure re. how and when we build e2e docker image, then it will be clear how we introduce `update_snapshots` workflow into the process. I added this step as the next one to tackle after we complete the current PR.

.github/workflows/e2e-tests.yml
bitdivine: Note that standard caches are valid across workflows and across different runs.  That is why [the binding check](https://github.com/dfinity/oisy-wallet/actions/workflows/binding-checks.yml) usually completes in about 1m20s for PRs that have only frontend changes.

I would recommend making a GitHub action in `.github/actions/oisy-backend` that does nothing but provide the Oisy backend Wasm, ideally from a cached build but building it if necessary.  The runtime of that action will be strictly less than the binding check, so less than 1m20s for frontend-only PRs and about 4-5m for a typical backend change.  That action can be used in many places, for example in the e2e tests, the binding tests and elsewhere.  You can use whatever tools you want to build that action; you don't need custom tools but if you  find them super-performant and the implementation looks secure, go ahead!

I would recommend doing the same for the frontend.  So, making an action that provides the `/build/` directory populated with all the frontend assets.  This action would have to take network as an argument.  This way the frontend doesn't need to be rebuilt if there are no frontend changes.

For the e2e tests themselves, note that you can request a bigger runner with more CPU cores, so you can run more tests in parallel.  Alternatively, if that is likely to cause concurrency issues, you can manually split the e2e tests into shards, [as we do for the nns-dapp](https://github.com/dfinity/nns-dapp/blob/main/.github/workflows/build.yml#L77).

DenysKarmazynDFINITY: Thanks for the very useful insights @bitdivine. Your suggestion does sound good, I'm already working on implementing it. Related PRs will follow. 

For now, closing this PR.

PR #2103, opened 2024-08-22
fix(frontend): pending ck conversions displayed under incorrect token

AntonioVentilii-DFINITY:
# Motivation

It was reported that pending ck conversions were displayed under the wrong token. e.g. one user converted LINK to ckLINK and then browsed ckUSDC and the frontend was displaying "Pending USDC to ckUSDC". This seems to happens because the pending transactions resolved with logs are not filtered for the selected token.

To solve it, we refine the logs topic including the ERC20 contract address among the parameters (encoding the token address).

We can see a practical example [here](https://etherscan.io/tx/0xddbf4aebe292c3d3b33240496db12a1a5622c4edb63017ebcd12ffa32fa5b924#eventlog).

Screenshot 2024-08-20 at 16 14 30

# Note

This is the second attempt to solve the issue, after the PR #2069, that was ultimately reverted.


src/frontend/src/eth/utils/token.utils.ts
```diff
@@ -10,3 +10,6 @@ export const isSupportedErc20TwinTokenId = (tokenId: TokenId): boolean =>
 
 export const isNotSupportedErc20TwinTokenId = (tokenId: TokenId): boolean =>
 	!isSupportedErc20TwinTokenId(tokenId);
+
+export const tokenAddressToHex = (address: string): string =>
```
peterpeterparker, line 14: - For `0x` we can use `mapAddressStartsWith0x`

- For converting to hex:
   - Isn't there already an EtherJS or existing ic-js function we can reuse?
   - If none, can you provide some tests.

peterpeterparker (commented):
Solution looks good, thx! Tested on staging with conversion for ckUSDC and ckLINK.

src/frontend/src/eth/utils/token.utils.ts
AntonioVentilii-DFINITY: for sure, we can use `hexZeroPad` from ethers (doc: https://docs.ethers.org/v4/api-utils.html#hex-strings); it does everything

changed

src/frontend/src/eth/utils/token.utils.ts
```diff
@@ -10,3 +11,6 @@ export const isSupportedErc20TwinTokenId = (tokenId: TokenId): boolean =>
...
+	ethers.utils.hexZeroPad(address.toLowerCase(), 32);
```
peterpeterparker, line 16: nth: I'm not familiar so as a nice to have, I would not be a comment about why `32`

peterpeterparker (approved):
LGTM, thanks

Tested on staging after a new re-redeploy. Made a conversion from USDC to ckUSDC and LINK to ckLINK and then opened related ck tokens. Both conversion, minus the UI glitch, were correctly displayed - i.e. one on going conversion per token.

Capture d’écran 2024-08-26 à 09 07 57
Capture d’écran 2024-08-26 à 09 08 01

src/frontend/src/eth/utils/token.utils.ts
AntonioVentilii-DFINITY: right, included comment

AntonioVentilii-DFINITY:
UI glitch being fixed in PR https://github.com/dfinity/oisy-wallet/pull/2124

PR #2104, opened 2024-08-22
feat(frontend): add warning banner in conversion modal for ck tokens

AntonioVentilii-DFINITY:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/eth/components/send/SendForm.svelte
+++ b/src/frontend/src/eth/components/send/SendForm.svelte
@@ -7,13 +7,14 @@
 	import { ethAddress } from '$lib/derived/address.derived';
 	import type { Network } from '$lib/types/network';
 	import SendAmount from '$eth/components/send/SendAmount.svelte';
-	import { isNullish } from '@dfinity/utils';
+	import { isNullish, nonNullish } from '@dfinity/utils';
 	import SendDestination from '$eth/components/send/SendDestination.svelte';
 	import { SEND_CONTEXT_KEY, type SendContext } from '$icp-eth/stores/send.store';
 	import { i18n } from '$lib/stores/i18n.store';
 	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
 	import type { EthereumNetwork } from '$eth/types/network';
 	import type { Token } from '$lib/types/token';
+	import Warning from '$lib/components/ui/Warning.svelte';
 
 	export let destination = '';
 	export let network: Network | undefined = undefined;
@@ -22,6 +23,7 @@
 	export let nativeEthereumToken: Token;
 	// TODO: to be removed once minterInfo breaking changes have been executed on mainnet
 	export let sourceNetwork: EthereumNetwork;
+	export let warning: string | undefined;
 
 	let insufficientFunds: boolean;
 	let invalidDestination: boolean;
@@ -54,6 +56,10 @@
 		<SendSource token={$sendToken} balance={$sendBalance} source={$ethAddress ?? ''} />
 
 		<FeeDisplay />
+
+		{#if nonNullish(warning)}
+			<Warning><p>{warning}</p></Warning>
+		{/if}
 	</div>
 
 	<ButtonGroup>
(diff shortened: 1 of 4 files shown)


AntonioVentilii-DFINITY:
Discussed with Artem and he is ok with multiple lines

src/frontend/src/lib/components/ui/Warning.svelte
```diff
@@ -1,10 +1,16 @@
 <script lang="ts">
-	import { IconWarning } from '@dfinity/gix-components';
+	import { IconInfo, IconWarning } from '@dfinity/gix-components';
+	import type { ComponentType } from 'svelte';
+
+	export let info = false;
+
+	let icon: ComponentType;
+	$: icon = info ? IconInfo : IconWarning;
```
peterpeterparker, line 8: Feels weird to have a component named `Warning` that can be an `info`. What's the plan?

peterpeterparker (commented):
Unsollicited UI/UX feedback:

- Is it correct that product want to display this "warning" on the "Form" step and not the "Review" step?

- Is it correct that design want to display this "warning" as last element of the list and not for example first?

src/frontend/src/lib/components/ui/Warning.svelte
AntonioVentilii-DFINITY: I did that to avoid duplicating code. What would be a good name for that component? `AlertDialog` or `MessageDialog` or `AlertLabel`?

peterpeterparker: We can have two separate components. `Warning.svelte` and `Info.svelte`.

AntonioVentilii-DFINITY: yes, maybe better eheh! separated

src/frontend/src/eth/components/send/SendForm.svelte
```diff
@@ -32,7 +35,17 @@
 
 	const dispatch = createEventDispatcher();
 
-	const { sendToken, sendBalance } = getContext<SendContext>(SEND_CONTEXT_KEY);
+	const { sendToken, sendBalance, sendPurpose } = getContext<SendContext>(SEND_CONTEXT_KEY);
+
+	let info: string | undefined = undefined;
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Agreed! So that we can re-use it easily

Done

src/frontend/src/eth/components/send/SendInfo.svelte
```diff
@@ -0,0 +1,30 @@
...
+</script>
+
+{#if displayInfo}
+	<Info>
+		<p>
+			{#if sendPurpose === 'convert-eth-to-cketh'}
+				{$i18n.convert.text.cketh_conversions_may_take}
+			{:else}
+				{replacePlaceholders($i18n.convert.text.ckerc20_conversions_may_take, {
+					$ckErc20: sendTokenAsErc20?.twinTokenSymbol ?? 'ckETH'
```
peterpeterparker, line 25: - Instead of ckETH should'nt we use something like `nativeEthereumToken.symbol` or something?

- Or maybe "ckETH" is fundamentally an incorrect fallback and we need something else?

AntonioVentilii-DFINITY: Theoretically, that fallback is just a proxy to make the type-interpreter ok: we would never arrive at that else condition if `twinTokenSymbol` is `undefined`, because we arrive at that part if we are in an ERC20 token with a ckERC20 counterparty.

The extreme correct way to do this, is to assert first if `twinTokenSymbol` is `undefined` and raise an error saying something like "inconsistent data, you should not have this screen, etc etc"

peterpeterparker: I see your point but, even if edge case, potentially displaying "ckETH" is incorrect no? Should we just fallback to `''` or have two sentences?

AntonioVentilii-DFINITY: i prefer to fall back to `''`

however, can I do it in another PR, after? There is another place where it should be changed too

peterpeterparker: I prefer that we fallback this one already to `''` as it feels incorrect and it's a new piece of code if you don't mind?

peterpeterparker (approved):
LGTM, thx

PR #2105, opened 2024-08-23
style(frontend): update authenticate button style

AntonioVentilii-DFINITY:
# Motivation

New style for the `ButtonAuthenticate`.

https://github.com/user-attachments/assets/4f309ae0-8bfb-4fc3-b208-0c8976de9b69

https://github.com/user-attachments/assets/0cb88f97-9554-4e36-84b6-b6fa30bd3234


src/frontend/src/lib/styles/global/button.scss
```diff
@@ -144,4 +144,39 @@ button {
...
 		border: 1px solid rgba(255, 255, 255, 0.3);
 	}
+
+	&.authenticate {
+		position: relative;
+		z-index: 0;
+	}
+
+	&.authenticate::before {
+		border-radius: 9999px;
```
peterpeterparker, line 154: `border-radius: 50%`?

src/frontend/src/lib/styles/global/button.scss
```diff
@@ -144,4 +144,39 @@ button {
...
+		bottom: 0;
+		left: 0;
+		z-index: -1;
+		background: radial-gradient(
+			49.28% 49.28% at 50% 100%,
+			var(--color-turquoise) 0%,
+			var(--color-crayola-blue) 100%
+		);
+		opacity: 0.6;
+		box-shadow:
```
peterpeterparker, line 168:
```
&.authenticate::before {
  box-shadow: 0 2px 8px 0 rgba(255, 255, 255, var(--box-shadow-opacity, 0.35)) inset, 0 4px 16px 0 rgba(0, 0, 0, 0.25);
}

&.authenticate:hover::before {
   --box-shadow-opacity: 0.75;
}
```

peterpeterparker (approved):
LGTM, minor suggestions

src/frontend/src/lib/styles/global/button.scss
AntonioVentilii-DFINITY: thank you! yes!

src/frontend/src/lib/styles/global/button.scss
AntonioVentilii-DFINITY: I tried that, but it is not working... I think it is because there is no width, so it becomes like this

Screenshot 2024-08-23 at 11 33 06

peterpeterparker: Gotcha

PR #2107, opened 2024-08-23
fix(frontend): do not show BTC text before onboard

AntonioVentilii-DFINITY:
# Motivation

It was reported that the address loading modal was writing Bitcoin even if it is still not onboarded.

image


peterpeterparker (commented):
Should we just say "Retrieving your public keys" ?

AntonioVentilii-DFINITY:
smarter! agreed! I will close this one and do a new one, to be cleaner

PR #2110, opened 2024-08-23
ci(e2e): issue with missing signer canister id

DenysKarmazynDFINITY:
# Motivation

Fix issue with missing signer canister id for e2e docker image.

### DIFF
--- a/canister_e2e_ids.json
+++ b/canister_e2e_ids.json
@@ -2,6 +2,9 @@
 	"backend": {
 		"local": "tdxud-2yaaa-aaaad-aadiq-cai"
 	},
+	"signer": {
+		"local": "tdxud-2yaaa-aaaad-aadiq-cai"
+	},
 	"ckbtc_index": {
 		"local": "mm444-5iaaa-aaaar-qaabq-cai"
 	},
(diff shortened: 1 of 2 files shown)


e2e/authentication.spec.ts
```diff
@@ -3,6 +3,5 @@ import { HomepageLoggedIn } from './utils/pages/homepage.page';
 
 testWithII('should sign-in', async ({ page, iiPage }) => {
 	const homepageLoggedIn = new HomepageLoggedIn({ page, iiPage });
-
```
DenysKarmazynDFINITY, line 6: I'll revert this change, it was used only to trigger e2e workflow.

canister_e2e_ids.json
```diff
@@ -2,6 +2,9 @@
 	"backend": {
 		"local": "tdxud-2yaaa-aaaad-aadiq-cai"
```
AntonioVentilii-DFINITY, line 3: should the backend be updated too? it changed?

DenysKarmazynDFINITY: I can be wrong, but I think this is kind of a temporary solution - same as locally, to make backend work, we need to manually update the signer canister id value in `.dfx/local/canister_ids.json` with the same id as `backend` canister has. My assumption is that when the old backend is completely removed, new backend and singer will have separate ids.

AntonioVentilii-DFINITY (approved):

PR #2111, opened 2024-08-23
refactor(frontend): generalize text when loading addresses from backend

AntonioVentilii-DFINITY:
# Motivation

Generic text when loading addresses from backend.


peterpeterparker (approved):
LGTM, thx

PR #2119, opened 2024-08-25
refactor(frontend): rename ETH address derived stores

AntonioVentilii-DFINITY:
# Motivation

Since we will most probably have derived store about BTC too, we rename the ETH derived stores about certified address, to avoid misunderstandings.


peterpeterparker (approved):
LGTM, thx

PR #2120, opened 2024-08-25
refactor(frontend): create function to reduce ResultSuccess

AntonioVentilii-DFINITY:
# Motivation

We create a function to reduce results given by type `ResultSuccess`.

# Changes

- Create new type `ResultSuccessReduced`.
- Create function `reduceResults`.
- Create test.

# Tests

Test worked correctly.


src/frontend/src/lib/utils/results.utils.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const reduceResults = <T = unknown>(
+	results: [ResultSuccess<T>, ...ResultSuccess<T>[]]
+): ResultSuccessReduced<T> =>
+	results.reduce(
+		(acc, { success: s, err: e }) => ({
+			success: acc.success && s,
+			err: nonNullish(e) ? [...acc.err, e] : acc.err
+		}),
+		{ success: true as boolean, err: [] as T[] }
```
peterpeterparker, line 12: - why `as boolean`?
- if you try `results.reduce<ResultSuccess<T>>(acc ...`, do you maybe not need any cast?

peterpeterparker: Thanks. I learned about the second thanks to an eslint rule which is set by default in the `oisy-wallet-signer` project. Maybe we can set the same rule in this repo if you like it too?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/utils/results.utils.ts
AntonioVentilii-DFINITY: yes! always in favour of adding eslint rules eheh! will do in a separate PR

peterpeterparker: > just as info, is it "standard-with-typescript" ruleset?

As you feel its best

AntonioVentilii-DFINITY: Done with:

- https://github.com/dfinity/oisy-wallet/pull/2343

PR #2124, opened 2024-08-26
style(frontend): fix overflowing IconConvert component

AntonioVentilii-DFINITY:
# Motivation

It was reported that there is a UI glitch for pending ck conversion transactions:

Capture d’écran 2024-08-26 à 09 07 57

That is caused by incorrect size and missing class, using the `IconConvert` component: in PR #2068, I introduced the component with fixed size `28px`, since that it is how it was used at the time. However, in PR #2073, we started using it for the transactions too, that require `24px` and a specific class.

I adapted the component to be similar to its peers `IconBurn` and `IconMint`: it will have an optional size prop that defaults to `24px` and an optional prop `styleClass` that defaults to `undefined`.


peterpeterparker (commented):
While this is an improvement I think it is not the solution. Similarly to `IconSend`, `IconReceive` or `IconBurn`, the `IconConvert` should also contains a property `styleClass: string | undefined` otherwise it cannot inherit the class `inset-center `.

AntonioVentilii-DFINITY:
Right! Thank you! Adjusted

peterpeterparker (approved):
LGTM, thx

PR #2128, opened 2024-08-26
style(frontend): change ERC20 ck conversion message

AntonioVentilii-DFINITY:
# Motivation

Request to change the message.

### DIFF
--- a/src/frontend/src/lib/i18n/en.json
+++ b/src/frontend/src/lib/i18n/en.json
@@ -434,7 +434,7 @@
 		},
 		"ethereum": {
 			"title": "Convert to $ckToken",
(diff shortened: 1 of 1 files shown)


peterpeterparker (commented):
Is the sentence used for ckETH?

Just thinking that it can potentially displays:

> With a few steps, convert your native Ethereum Ethereum to its twinned ICP ckEthereum directly within the Oisy

So wondering if that would be the case and if it would be the expectation?

AntonioVentilii-DFINITY:
It should show

> convert your native Ethereum ETH to its twinned ICP ckETH

but I agree that it seems redundant. Maybe I remove the network just when it is ETH and mainnet?

peterpeterparker (approved):
LGTM, thx

PR #2130, opened 2024-08-26
fix(frontend): pass env to mapCkEthereumTransaction

AntonioVentilii-DFINITY:
# Motivation

The `env` param was not being passed to `mapCkEthereumTransaction`: the variable `ckETHExplorerUrl` inside was automatically `undefined` in this way.


peterpeterparker (commented):
Should the call `mapCkEthereumTransaction` in `ic-transactions.utils` be fixed as well? 
If yes, should we fix it in this PR as well?

AntonioVentilii-DFINITY:
Good catch! Yes, included

peterpeterparker (approved):
Cool, thanks

PR #2131, opened 2024-08-26
test(e2e): "receive tokens" modal snapshot

DenysKarmazynDFINITY:
# Motivation

1. Adjust modal-related methods to make it available for both signed-in and signed-out homepages/
2. "Receive tokens" modal snapshot.

Next steps are to test copy and QR button in this modal.


peterpeterparker:
> "Receive tokens" modal snapshot. Note: token addresses have to be masked since they have dynamic values and would therefore fail screenshot comparison.

We need tests to validate the addresses, either here or elsewhere. Do you have a plan for that?

Speaking of address comparison, I didn’t look at the test code. Do you also plan to scan the QR code image and compare it with the displayed textual address to ensure that both are displaying same values?

DenysKarmazynDFINITY:
@peterpeterparker the plan is to do the QR code tests in a separate PR. I updated the respective JIRA ticket with the test case you suggested. 

P.S.: please feel free to add any test cases you want me to tackle here - https://dfinity.atlassian.net/browse/GIX-2843.

peterpeterparker:
Thanks, great to hear about the QR code. 👍

I read the JIRA task but, regarding my first question, I'm still unsure if you plan to assert the addresses as well or not?

DenysKarmazynDFINITY:
@peterpeterparker regarding the addresses, the plan is the following:
1. Test that browser clipboard value equals to the address displayed in the modal after "copy" button has been clicked.
2. Test that addresses displayed in the list view equal to the values in the QR code view.

Anything else you'd like me to validate?

DenysKarmazynDFINITY:
@peterpeterparker yes, agree, we can and should use e2e tests layer for that. My current approach is to start with "low-hanging fruits" and cover with snapshots and smaller UI/UX tests all major part of the app. After that, I would switch to business logic, advanced test scenarios and data integrity..

peterpeterparker:
Sure. That said, I would suggest to not spend too much time on covering all UI/UX and only start validating business logic and data too far away in the future. In my humble opinion, the latter is the most important. But for sure, starting with low handing fruits sounds like a good way to start. 👍

DenysKarmazynDFINITY:
@peterpeterparker agree with your point. I'll try to find the golden middle between adding UI/UX and business logic tests 👍

peterpeterparker:
Super!

PR #2133, opened 2024-08-26
style(frontend): avoid showing empty tokens list on loading

AntonioVentilii-DFINITY:
# Motivation

We aim to show a non-empty tokens list as first, and avoid showing an empty list just because it is loading, thus reducing glitches.

### Before

https://github.com/user-attachments/assets/b67eacb5-4553-4882-9f98-4bd847afaaa0

### After

https://github.com/user-attachments/assets/6937aa12-702f-49ab-8431-2cb31561f79a

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensSignedIn.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensSignedIn.svelte
@@ -3,7 +3,7 @@
 	import TokensSkeletons from '$lib/components/tokens/TokensSkeletons.svelte';
 	import { combinedDerivedSortedNetworkTokensUi } from '$lib/derived/network-tokens.derived';
 	import { i18n } from '$lib/stores/i18n.store';
-	import type { TokenUi } from '$lib/types/token';
+	import type { Token, TokenUi } from '$lib/types/token';
 	import { hideZeroBalancesStore } from '$lib/stores/settings.store';
 	import { fade } from 'svelte/transition';
 	import { modalManageTokens } from '$lib/derived/modal.derived';
@@ -13,7 +13,7 @@
 	import TokenCardContent from '$lib/components/tokens/TokenCardContent.svelte';
 	import { balancesStore } from '$lib/stores/balances.store';
 	import { BigNumber } from '@ethersproject/bignumber';
-	import { debounce } from '@dfinity/utils';
+	import { debounce, isNullish } from '@dfinity/utils';
 
 	let displayZeroBalance: boolean;
 	$: displayZeroBalance = $hideZeroBalancesStore?.enabled !== true;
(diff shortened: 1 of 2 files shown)


peterpeterparker (commented):
I don't understand what's the root issue? Can you share some screenshots or demo or what problem this PR aims to address?

AntonioVentilii-DFINITY:
I rephrased a little and included some videos done in staging. Basically I am breaking down PR #1796 into two main parts:

1. Avoid glitches but showing some list (even with balance and exchanges loading)
2. Disable sorting during pointer events

src/frontend/src/lib/components/tokens/TokensSignedIn.svelte
```diff
@@ -24,6 +24,15 @@
 			($balancesStore?.[tokenId]?.data ?? BigNumber.from(0n)).gt(0n) || displayZeroBalance
 	);
 
+	// We start it as undefined to avoid showing an empty list before the first update.
+	let tokensToDisplay: Token[] | undefined = undefined;
```
peterpeterparker, line 28: Is there a particular reason why you are introducing a new state? Why not debouncing the filtering that leads to assing `tokens`?

src/frontend/src/lib/components/tokens/TokensSkeletons.svelte
```diff
@@ -3,9 +3,11 @@
 	import { fade } from 'svelte/transition';
 	import SkeletonCards from '$lib/components/ui/SkeletonCards.svelte';
 	import { TOKENS_SKELETONS_INITIALIZED } from '$lib/constants/test-ids.constants';
+
+	export let loading = false;
 </script>
 
-{#if $erc20UserTokensNotInitialized}
+{#if $erc20UserTokensNotInitialized || loading}
```
peterpeterparker, line 10: If we introduce a property named `loading` then I would move `$erc20UserTokensNotInitialized` outside of the components and used it in the composition of the `loading` assignment.

Alternativey, I would not add a property loading but, a property "tokens" and add the logic "isNullish" here.

The first solution being probably cleaner but, I suggest to not mix logic between inner and outter sources.

src/frontend/src/lib/components/tokens/TokensSignedIn.svelte
AntonioVentilii-DFINITY: No particular reason: I just wanted to separate what would be the tokens list "raw" and the one that we would display, just for clarity (even if there is not final difference in values in the end). But I can for sure merge, if you want

src/frontend/src/lib/components/tokens/TokensSkeletons.svelte
AntonioVentilii-DFINITY: Yes, agreed! But maybe better to put it in another PR, to be cleaner?

https://github.com/dfinity/oisy-wallet/pull/2136

src/frontend/src/lib/components/tokens/TokensSignedIn.svelte
peterpeterparker: It's maybe cleaner but, not introducing a new state has the advantages to improve the performances as debouncing will reduce the number of filtering.

AntonioVentilii-DFINITY:
this PR became obsolete, closing

PR #2134, opened 2024-08-26
fix(backend): Remove gzip timestamp to make build reproducible

Sawchord:
# Motivation

This PR fixes a reproducibility bug in the backend.
`gzip` by default includes a timestamp. This PR removes it.

# Tests

Manually tested by running Docker build of `main` on both, my Mac(Intel) and remote Linux and hashes match


peterpeterparker (approved):
Good catch, thanks!

I double checked II, they do `gzip --no-name --force "$canister.wasm"` so it was something missing here.

PR #2136, opened 2024-08-27
refactor(frontend): make loading a prop for TokensSkeleton

AntonioVentilii-DFINITY:
# Motivation

Component `TokensSkeleton` will have a new prop `loading`, since we expect such a value to be dependent on more factors.

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensSignedIn.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensSignedIn.svelte
@@ -44,7 +44,7 @@
 	}, 250);
 
 	let loading: boolean;
-	$: $erc20UserTokensNotInitialized, (loading = $erc20UserTokensNotInitialized);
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/tokens/TokensSignedIn.svelte
```diff
@@ -41,9 +42,12 @@
 			handleAnimationEnd();
 		}
 	}, 250);
+
+	let loading: boolean;
+	$: $erc20UserTokensNotInitialized, (loading = $erc20UserTokensNotInitialized);
```
peterpeterparker, line 47: Can be simplified to `$: loading = $erc20UserTokensNotInitialized;`

AntonioVentilii-DFINITY: corrected

peterpeterparker (approved):
LGTM, thx

PR #2137, opened 2024-08-27
refactor(frontend): derived store for ETH address data

AntonioVentilii-DFINITY:
# Motivation

To avoid repeating code and simplify readability, there is a new derived store that determines the ETH address certified data.

### DIFF
--- a/src/frontend/src/lib/derived/address.derived.ts
+++ b/src/frontend/src/lib/derived/address.derived.ts
@@ -1,12 +1,17 @@
 import { BTC_MAINNET_TOKEN_ID } from '$env/tokens.btc.env';
 import { ETHEREUM_TOKEN_ID } from '$env/tokens.env';
-import { addressStore } from '$lib/stores/address.store';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/derived/address.derived.ts
```diff
@@ -9,6 +9,11 @@ export const addressNotLoaded: Readable<boolean> = derived([addressStore], ([$ad
 	isNullish($addressStore)
 );
 
+const ethAddressData: Readable<AddressData | null | undefined> = derived(
```
peterpeterparker, line 12: Do we maybe already have a type that equals `AddressData | null | undefined`?

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/lib/derived/address.derived.ts
AntonioVentilii-DFINITY: Included

PR #2138, opened 2024-08-27
refactor(frontend): move InfoEthereum to icp-eth folder

AntonioVentilii-DFINITY:
# Motivation

Since we aim to use it for both ETH and IC tokens, we move `InfoEthereum` to folder `icp-eth`.


peterpeterparker (approved):
LGTM, thanks

PR #2139, opened 2024-08-27
fix(frontend): correct fallback string when converting ckERC20

AntonioVentilii-DFINITY:
# Motivation

As discussed in PR #2104 ([comment](https://github.com/dfinity/oisy-wallet/pull/2104#discussion_r1732623091)), it is better to make the token symbol to fallback on an empty string `''`.

Unrelated: there is a slight mispelling in one of the variables.


peterpeterparker (approved):
Thanks!

PR #2140, opened 2024-08-27
fix(frontend): tokens not showing when loading

AntonioVentilii-DFINITY:
# Motivation

There is a concurrency issue: when starting the `TokensSkeleton`, component `TokensDisplayHandler` is not mounted on loading, meaning that `tokens` variable is not updated, and it remains `undefined`. That affects `loading` itself and it keeps being `true`. 

# Changes

Moved `TokensDisplayHandler` above `TokensSkeletons`.

# Tests

### Before

Screen was stuck to this:

Screenshot 2024-08-27 at 13 59 05

### After

It updates correctly:

Screenshot 2024-08-27 at 14 03 57


peterpeterparker (approved):
LGTM, thanks

PR #2150, opened 2024-08-27
feat(frontend): add balance to TokenUi

AntonioVentilii-DFINITY:
# Motivation

The balance store is called a few times in the code, related to derived store `combinedDerivedEnabledNetworkTokensUi`. So, to simplify the code, we add the prop `balance` to type `TokenUi`.
After this PR, we will make use of the new prop throughout the code.

# Changes

- Add `balance` to `TokenUi`.
- Calculate balance in store `combinedDerivedEnabledNetworkTokensUi`.

# Tests

No changes in local replica, just more loading time (that should be compensated once we use the new prop).

### DIFF
--- a/src/frontend/src/lib/types/token.ts
+++ b/src/frontend/src/lib/types/token.ts
@@ -40,6 +40,7 @@ export type OptionTokenStandard = TokenStandard | undefined | null;
 export type TokenToPin = Pick<Token, 'id'> & { network: Pick<Token['network'], 'id'> };
 
 interface TokenFinancialData {
+	balance?: number;
 	usdBalance?: number;
 }
 
--- a/src/frontend/src/lib/types/token.ts
+++ b/src/frontend/src/lib/types/token.ts
@@ -1,5 +1,6 @@
 import type { Network } from '$lib/types/network';
 import type { RequiredExcept } from '$lib/types/utils';
+import type { BigNumber } from '@ethersproject/bignumber';
 
 export type TokenId = symbol;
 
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/types/token.ts
```diff
@@ -40,6 +40,7 @@ export type OptionTokenStandard = TokenStandard | undefined | null;
 export type TokenToPin = Pick<Token, 'id'> & { network: Pick<Token['network'], 'id'> };
 
 interface TokenFinancialData {
+	balance?: number;
```
peterpeterparker, line 43: I would keep non formatted values in stores, in that case `BigNumber` which feels more future prone. If it's more handy to hold a formatted value in the store then I would name the variable accordingly, like `formattedBalance` or something.

AntonioVentilii-DFINITY: make sense! however is more useful to have the formatted one: i will rename it, but if you prefer i can add even the `BigNumber` prop, having two new values: `balance` and `formattedBalance`

src/frontend/src/lib/types/token.ts
```diff
@@ -40,6 +40,7 @@ export type OptionTokenStandard = TokenStandard | undefined | null;
 export type TokenToPin = Pick<Token, 'id'> & { network: Pick<Token['network'], 'id'> };
 
 interface TokenFinancialData {
+	formattedBalance?: number;
```
peterpeterparker, line 43: If we want to hold the formatted balance, shouldn't it be a `string` then or is there a use case to convert it to number?

src/frontend/src/lib/types/token.ts
peterpeterparker: Sounds good

src/frontend/src/lib/types/token.ts
AntonioVentilii-DFINITY: Yes, it is even more consistent with the name

peterpeterparker (approved):
LGTM, thx

PR #2151, opened 2024-08-27
refactor(frontend): use existing balance prop to filter tokens

AntonioVentilii-DFINITY:
# Motivation

Instead of calling again the balances store, we can use directly the `usdBalance` prop of type `TokenUi`.

# Tests

Local replica unchanged:

All balances

Screenshot 2024-08-28 at 07 36 02

Only non-zero balances
Screenshot 2024-08-28 at 07 36 10

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
@@ -5,8 +5,6 @@
 	import { debounce } from '@dfinity/utils';
 	import { hideZeroBalancesStore } from '$lib/stores/settings.store';
 	import { combinedDerivedSortedNetworkTokensUi } from '$lib/derived/network-tokens.derived';
-	import { balancesStore } from '$lib/stores/balances.store';
-	import { BigNumber } from '@ethersproject/bignumber';
 
 	// We start it as undefined to avoid showing an empty list before the first update.
 	export let tokens: Token[] | undefined = undefined;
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
```diff
@@ -16,8 +14,7 @@
 
 	let sortedTokens: TokenUi[];
 	$: sortedTokens = $combinedDerivedSortedNetworkTokensUi.filter(
-		({ id: tokenId }) =>
-			($balancesStore?.[tokenId]?.data ?? BigNumber.from(0n)).gt(0n) || displayZeroBalance
+		({ usdBalance }) => (usdBalance ?? 0) || displayZeroBalance
```
peterpeterparker, line 17: It's not the same logic anymore - i.e. there is no comparison that the balance is greater than zero - is that expected?

AntonioVentilii-DFINITY: The value `0` is typically a `falsy` value, meaning that we can use it as direct condition (and the balance will never be negative, but even if it is, we should probably show it either way)

in any case, to have the scope more clear in the code, i can put the `>0` condition, if you prefer

peterpeterparker: If the behavior remains intact and the change is excepted for sure it's fine.

peterpeterparker (approved):
LGTM, thanks

PR #2152, opened 2024-08-27
feat(frontend): create store for tokens to be displayed

AntonioVentilii-DFINITY:
# Motivation

The tokens to be displayed in `TokensSignedIn` are "re-calculated" every time that the page mounts, meaning that if an user moves back and forth between pages, it will always see the skeleton first. That is not a good user experience.

So, we create a derived store to manage the tokens to be displayed. Note that the logic of enable and disabling sorting is kept the same as before.

# Changes

- Create store for list current token keys that defines the current order.
- Move the defining logic for the tokens to be displayed in a separate util function.
- Create tests for such function.
- Create derived store `combinedDerivedTokensToDisplay`.
- Adapt component `TokensDisplayHandler`.

# Tests

Manual test in local replica:

https://github.com/user-attachments/assets/fc715556-9033-4e27-ab56-312d4508b8e2


src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
```diff
@@ -1,55 +1,18 @@
 <script lang="ts">
-	import type { Token, TokenUi } from '$lib/types/token';
-	import { pointerEventStore } from '$lib/stores/events.store';
+	import type { Token } from '$lib/types/token';
 	import { pointerEventsHandler } from '$lib/utils/events.utils';
-	import { debounce } from '@dfinity/utils';
 	import { hideZeroBalancesStore } from '$lib/stores/settings.store';
-	import { combinedDerivedSortedNetworkTokensUi } from '$lib/derived/network-tokens.derived';
+	import { combinedDerivedTokensToDisplay } from '$lib/derived/tokens-to-display.derived';
```
peterpeterparker, line 5: `combinedDerivedTokensUi`? Just an idea but given we have `TokenUi` maybe it makes sense?

src/frontend/src/lib/derived/tokens-to-display.derived.ts
```diff
@@ -0,0 +1,11 @@
+import { combinedDerivedSortedNetworkTokensUi } from '$lib/derived/network-tokens.derived';
+import { pointerEventStore } from '$lib/stores/events.store';
+import type { TokenUi } from '$lib/types/token';
+import { defineTokensToDisplay } from '$lib/utils/tokens-to-display.utils';
+import { derived, type Readable } from 'svelte/store';
+
+export const combinedDerivedTokensToDisplay: Readable<TokenUi[]> = derived(
+	[combinedDerivedSortedNetworkTokensUi, pointerEventStore],
```
peterpeterparker, line 8: should we call this modules `tokens-ui.derived` and in another PR moved `combinedDerivedSortedNetworkTokensUi` here?

src/frontend/src/lib/stores/tokens-to-display.store.ts
```diff
@@ -0,0 +1,3 @@
+import { writable } from 'svelte/store';
+
+export const currentTokensKeysStore = writable<string[]>([]);
```
peterpeterparker, line 3: what's this? do we need it? what's string?

src/frontend/src/lib/utils/tokens-to-display.utils.ts
```diff
@@ -0,0 +1,45 @@
...
+
+	if (!$pointerEventStore) {
+		// No pointer events, so we are not worried about possible glitches on user's interaction.
+		currentTokensKeysStore.set(sortedTokensKeys);
+		return $sortedTokens;
+	}
+
+	// If there are pointer events, we need to avoid visually re-sorting the tokens, to prevent glitches on user's interaction.
+
+	// We are statically getting the current tokens keys, as it should not trigger any reactivity.
```
peterpeterparker, line 26: statically? you mean imperatively?

src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
AntonioVentilii-DFINITY: yes, better!

in a separate PR after this, i will move all these derived UI to a separate module, to concentrate them

src/frontend/src/lib/derived/tokens-to-display.derived.ts
AntonioVentilii-DFINITY: yes, agreed!

src/frontend/src/lib/stores/tokens-to-display.store.ts
AntonioVentilii-DFINITY: yes, we are using it in the `defineTokensToDisplay` function: it "stores" the last sorting order of the token list. And we update it ONLY when we update the tokens to display. 

In case of pointer events, it is used to compare if the previous order is the same as the new one.

I added some comments in `defineTokensToDisplay`. Shall I add here too?

src/frontend/src/lib/utils/tokens-to-display.utils.ts
AntonioVentilii-DFINITY: right! corrected

src/frontend/src/lib/stores/tokens-to-display.store.ts
peterpeterparker: I don't quite understand—why does it need to be global? Is it to save the sorting order when navigating back and forth?

Following is few questions that crossed my mind right now, what do you think of those?

1. Should we consider providing this in a separate PR?

2. Is it possible to improve the type definition, perhaps by creating a specific type for `string`? This could make the code more readable.

3. Instead of `currentTokensKeyStore`, maybe it could be renamed to something more generic, like `tokensSortingStore`, and contain an object that includes the `string[]`. This way, it becomes more descriptive, and if additional sort options are added in the future, the structure would already accommodate them. I generally don't like adding things in advance but maybe here it's interesting?

4. Similarly to the "hide zero balance" filter, should this list of keys be saved in the browsers' cache or those have to be recaculated each time the app boots?

5. Speaking of, should we even think at having the "filter zero balance: boolean" in the same store? Actually not sure it makes sense to have a single object that contains a filter.

Again, just few things that crossed my mind.

AntonioVentilii-DFINITY: > Is it to save the sorting order when navigating back and forth?

it is to save the sorting order between re-sorting: this store "stores" the last state of the order of the token list displayed

1. I can separate in another, but I put this one in draft, since it should come before

2. yes, will do, something like `TokenIndexKey` or simply `TokenKey`?

3. agreed!

4. for this one, I would do it in a separate PR, after having discussed it better

5. make sense! will do it in a separate PR

peterpeterparker: Cool!

4. Note again that I'm really unsure if it makes sense or not to save those in idb, just something that crossed my mind.

5. We have to pay super attention to not create an endless loop of hell when if we do this.

AntonioVentilii-DFINITY:
This PR was split into:

- PR #2164
- PR #2171

So, I will be closing it

PR #2154, opened 2024-08-28
style(frontend): pulsating token balance when loading in list

AntonioVentilii-DFINITY:
…in list

# Motivation

# Changes

# Tests

I slowed down the initialization to test:

https://github.com/user-attachments/assets/4bd287b3-a24d-4c0c-a3dc-410454305394


AntonioVentilii-DFINITY:
As discussed with Artem, this is not a good solution

PR #2155, opened 2024-08-28
refactor(frontend): use prop balance to sort tokens

AntonioVentilii-DFINITY:
# Motivation

The dependency of `balanceStore` is removed from `pinTokensWithBalanceAtTop` since we can use directly the `formattedBalance` prop of TokenUi introduced in PR #2150.

# Changes

# Tests

Adapted tests worked as expected and no changes in local replica.

### DIFF
--- a/src/frontend/src/lib/utils/tokens.utils.ts
+++ b/src/frontend/src/lib/utils/tokens.utils.ts
@@ -1,10 +1,6 @@
-import type { BalancesData } from '$lib/stores/balances.store';
-import type { CertifiedStoreData } from '$lib/stores/certified.store';
 import type { ExchangesData } from '$lib/types/exchange';
-import type { Token, TokenToPin, TokenUi, TokenWithBalance } from '$lib/types/token';
-import { formatToken } from '$lib/utils/format.utils';
+import type { Token, TokenToPin, TokenUi } from '$lib/types/token';
 import { nonNullish } from '@dfinity/utils';
-import { BigNumber } from '@ethersproject/bignumber';
 
 export const pinTokensAtTop = ({
 	$tokens,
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -61,37 +57,15 @@ export const sortTokens = ({
...
-	);
-
-	const [positiveBalances, nonPositiveBalances] = tokensWithBalance.reduce<
-		[TokenWithBalance[], TokenWithBalance[]]
-	>(
+export const pinTokensWithBalanceAtTop = ($tokens: TokenUi[]): TokenUi[] => {
+	const [positiveBalances, nonPositiveBalances] = $tokens.reduce<[TokenUi[], TokenUi[]]>(
 		(acc, token) => (
-			(token.usdBalance ?? 0) > 0 || token.balance > 0 ? acc[0].push(token) : acc[1].push(token),
+			(token.usdBalance ?? 0) > 0 || Number(token.formattedBalance ?? 0) > 0
```
peterpeterparker, line 66: Why not use `balance`?

AntonioVentilii-DFINITY: no particular reason: before it was a number comparison, so I kept the same logic, but it is indifferent for me, no strong opinion

What would you prefer?

peterpeterparker: I prefer avoiding casting as I'm always afraid to ends on `NaN` or other similar issue in some ways.

AntonioVentilii-DFINITY: perfect, changed!

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -101,10 +76,10 @@ export const pinTokensWithBalanceAtTop = ({
 		...positiveBalances.sort(
 			(a, b) =>
 				(b.usdBalance ?? 0) - (a.usdBalance ?? 0) ||
-				b.balance - a.balance ||
+				Number(b.formattedBalance ?? 0) - Number(a.formattedBalance ?? 0) ||
```
peterpeterparker, line 79: Can we use balance here as well?

AntonioVentilii-DFINITY: I tried here too, but it becomes quite verbose, since I have to use both `gt` and `lt` for the bignumbers. That part of the condition would become

```
(b.balance ?? BigNumber.from(0)).gt(a.balance ?? BigNumber.from(0)) - 
        (a.balance ?? BigNumber.from(0)).gt(b.balance ?? BigNumber.from(0))
```

but if all ok, I will add it there

peterpeterparker: You can probably have a constant somewhere `const ZERO = BigNumber.from(0)` (or else) that should make things less verbose.

peterpeterparker (approved):
LGTM, thanks

PR #2159, opened 2024-08-28
feat(frontend): constant for BigNumber zero

AntonioVentilii-DFINITY:
# Motivation

To make the code less verbose, we set a constant for `BigNumber` 0.


src/frontend/src/lib/constants/utils.constants.ts
```diff
@@ -0,0 +1,3 @@
+import { BigNumber } from '@ethersproject/bignumber';
+
+export const ZERO = BigNumber.from(0n);
```
peterpeterparker, line 3: Maybe move it to `app.constants`? `utils.constants` feel like a constant that is only used in the "utils" modules. A bit an opiniated feedback, hope it makes sense?

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/lib/constants/utils.constants.ts
AntonioVentilii-DFINITY: agreed, changed

PR #2160, opened 2024-08-28
refactor(frontend): use prop balance for TokenBalance in list

AntonioVentilii-DFINITY:
# Motivation

Since both usage of `TokenCardContent` provide type `TokenUi`, we adapt the components to use the `balance` prop instead of the store.


peterpeterparker (approved):
LGTM, thx

PR #2163, opened 2024-08-28
build(frontend): use ic-js next

AntonioVentilii-DFINITY:
# Motivation

We need to bump `@dfinity/ckbtc` to the next version to use the `getBalance` function of `BitcoinCanister` object.


peterpeterparker (approved):
LGTM, thanks

PR #2164, opened 2024-08-28
feat(frontend): create store for tokens keys

AntonioVentilii-DFINITY:
# Motivation

This PR aims at saving the last order of the tokens list that is displayed: we are using this state internally in the component, however, it can become useful to transform into a store to be called externally.

Either way, it is set only during the process to update the tokens to be displayed.

Just to clarify, the scope of this variable is to compare if the new order is different from the old one, after re-sorting.

# Changes

- Create new type for the keys of the list.
- New store for generic `TokenUi[]` data, including the list of tokens keys.
- Adapt component `TokensDisplayHandler` to use and set the new store.

# Tests

No changes in local replica.

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
@@ -1,10 +1,11 @@
 <script lang="ts">
-	import type { Token, TokenUi } from '$lib/types/token';
+	import type { Token, TokenIndexKey, TokenUi } from '$lib/types/token';
 	import { pointerEventStore } from '$lib/stores/events.store';
 	import { pointerEventsHandler } from '$lib/utils/events.utils';
 	import { debounce } from '@dfinity/utils';
 	import { hideZeroBalancesStore } from '$lib/stores/settings.store';
 	import { combinedDerivedSortedNetworkTokensUi } from '$lib/derived/network-tokens.derived';
+	import { tokensUiStore } from '$lib/stores/tokens-ui.store';
 
 	// We start it as undefined to avoid showing an empty list before the first update.
 	export let tokens: Token[] | undefined = undefined;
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
```diff
@@ -19,22 +20,23 @@
...
-
-	let sortedTokensKeys: string;
-	$: sortedTokensKeys = sortedTokens?.map(parseTokenKey).join(',') ?? '';
+	let sortedTokensKeys: TokenIndexKey[];
+	$: sortedTokensKeys = sortedTokens.map(parseTokenKey);
 
 	const updateTokensToDisplay = () => {
 		if (!$pointerEventStore) {
 			// No pointer events, so we are not worried about possible glitches on user's interaction.
+			tokensUiStore.set({ tokensKeys: sortedTokensKeys });
```
peterpeterparker, line 29: Is there a particular reason why you implemented the assignement in an imperative way instead of keeping the original reactive assignement?

i.e. why not something like?

```
$: tokensUiStore.set({ tokensKeys: tokens?.map(parseTokenKey)
```

AntonioVentilii-DFINITY: no particular reason: i am reading it imperatively, since i wanted to avoid recursive loops; so, I wanted to set it imperatively too, to have all concentrated in this function.

But no strong opinion. Maybe like this, there will be slightly less assignment iterations (it is not assigned for every change of tokens, but only on re-sorting)?

If you prefer I put it reactively

peterpeterparker: Thing is we are doing two things in this PR: introducing a store plus modifying the behavior. Given that we do not have test around this, I would stick to existing logic.

In addition, we are setting back to back two different states in the function - `tokensUiStore.set` and `tokens = sortedTokens;` - which make me ask myself if it's good or not.

Long story short, this might good be an improvement but I would split the feature in two and when implementing this change, do some manual testing and reporting to check if it behaves the same, better or worse.

Makes sense?

AntonioVentilii-DFINITY: Let's do like this:

Since this was already a split of PR #2152, I will first do a PR for the other part that involves adding tests to the function that updates the tokens to display: PR  https://github.com/dfinity/oisy-wallet/pull/2171

Maybe that would be enough, so that this one is not needed anymore

AntonioVentilii-DFINITY:
This PR is deprecated, since for now PR https://github.com/dfinity/oisy-wallet/pull/2171 seems to be enough

PR #2165, opened 2024-08-28
refactor(frontend): invert update condition in EthereumFeeContext

AntonioVentilii-DFINITY:
# Motivation

Just to simplify the code, we invert the condition in `updateContext` function of `EthereumFeeContext` component.


peterpeterparker (approved):
LGTM, thanks

PR #2166, opened 2024-08-28
refactor(frontend): use ckETH fee context for ckERC20 conversion approval

AntonioVentilii-DFINITY:
# Motivation

The `ckETH` fee for `ckERC20 --> ERC20` conversion approval (referring to the first step of the [withdrawal scheme](https://github.com/dfinity/ic/blob/master/rs/ethereum/cketh/docs/ckerc20.adoc#withdrawal-ckerc20-to-erc20)) was hardcoded. We aim to use the actual fee defined by the context `EthereumFeeContext`.
The scope is that the fee amount showed to the user is exactly the same as the one approved by the `ckETH` ledger.

# Changes

- Capture `ckEthMaxTransactionFee` from the context. This value is not `undefined` only when there is a ck-conversion (`ckETH` or `ckERC20`).
- Pass the value to be used only by the `convertCkErc20ToErc20` function in the first approval step.

# Tests

In `staging` env `ckSepoliaUSDC --> SepoliaUSDC` worked correctly

### IC tx [346](https://dashboard.internetcomputer.org/ethereum/yfumr-cyaaa-aaaar-qaela-cai/transaction/346)

Screenshot 2024-08-30 at 07 03 27

### Sepolia tx [0x00548b62f1fc42d5aacf98b734eb54863c24fbee3e06751a0cd57af76da70bb9](https://sepolia.etherscan.io/tx/0x00548b62f1fc42d5aacf98b734eb54863c24fbee3e06751a0cd57af76da70bb9)

Screenshot 2024-08-30 at 07 03 10

### Transaction flow

https://github.com/user-attachments/assets/f9f38543-2243-4955-a06b-8ac291667fc9


src/frontend/src/icp/components/send/IcSendTokenWizard.svelte
```diff
@@ -71,6 +71,12 @@
 
 	const dispatch = createEventDispatcher();
 
+	const { store: ethereumFeeStore } = getContext<EthereumFeeContext>(ETHEREUM_FEE_CONTEXT_KEY);
+
+	// In case we are converting ERC20 to ckERC20, we need to include ckETH fees in the transaction.
+	let ckEthMaxTransactionFee: bigint | undefined = undefined;
+	$: ckEthMaxTransactionFee = $ethereumFeeStore.maxTransactionFee;
```
AntonioVentilii-DFINITY, line 78: @peterpeterparker the doubt I asked you about is:

this value is updated by the child `EthereumFeeContext`, so, will it be fetched here (in the parent) correctly?

src/frontend/src/icp/components/send/IcSendTokenWizard.svelte
```diff
@@ -190,6 +191,12 @@
 		store: initEthereumFeeStore()
 	});
 
+	const { store: ethereumFeeStore } = getContext<EthereumFeeContext>(ETHEREUM_FEE_CONTEXT_KEY);
+
+	// In case we are converting ERC20 to ckERC20, we need to include ckETH fees in the transaction.
+	let ckEthMaxTransactionFee: bigint | undefined = undefined;
+	$: ckEthMaxTransactionFee = $ethereumFeeStore?.maxTransactionFee;
```
AntonioVentilii-DFINITY, line 198: @peterpeterparker the doubt I asked you about is:

this value is updated by the child `EthereumFeeContext`, so, will it be fetched here (in the parent) correctly?

src/frontend/src/icp/services/ck.services.ts
```diff
@@ -69,12 +69,17 @@ export const convertCkErc20ToErc20 = async ({
 
 	// 1. Approve fees on ckETH Ledger for minter
 
+	assertNonNullish(
```
peterpeterparker, line 72: NTH: for readability reasons I would insert the assertion before above comment `// 1. Approve ...`

peterpeterparker (commented):
Actually, would it be more accurate to not pass the `maxTransactionFee` as a parameter and instead, as step 0. in the `convertCkErc20ToErc20`, fetch the fee with a call at this point again?

The fee we pass as a parameter might be up to 30 seconds old, so adding a call might provide a more accurate value, though I'm not sure if it's necessary. It's also an additional call.

Have you considered this option? What do you think?

Strictly a question.

src/frontend/src/icp/services/ck.services.ts
AntonioVentilii-DFINITY: done

AntonioVentilii-DFINITY:
Ah wait, forget my last part: it is indeed dynamic in the Review step too. The user would see that fee amount right before clicking send, and the time that passes between the click and the first step of the conversion should enough, I think, no?

WDYT?

peterpeterparker:
Yes it's dynamic also in the review step because it's an information for any Ethereum and Erc20 transfer.

peterpeterparker:
But sounds good then, if the fees displayed are those to be paid indeed the implementation matchs this requirement.

Thanks for the feedback.

src/frontend/src/icp/types/ic-send.ts
```diff
@@ -5,6 +5,7 @@ import type { TransferParams } from '$lib/types/send';
 export type IcTransferParams = Pick<TransferParams, 'amount' | 'to'> & {
 	identity: OptionIdentity;
 	progress: (step: ProgressStepsSendIc) => void;
+	ckEthMaxTransactionFee?: bigint;
```
peterpeterparker, line 8: I suggest to rename this parameter for something more meaningful about it's usage - i.e. the fee are used as amount of the tranfer approval in the conversion of ckErc20. That way the parameter name would already gives us a hint of its usage.

peterpeterparker: You know what I mean?

AntonioVentilii-DFINITY: I think I understood! It shouls say: max ckETH fees burned to cover Ethereum fees in case of ckERC20

What about `ckErc20MaxEthereumFee`? Or very verbose "ckErc20EthereumFeeToBeApproved"?

peterpeterparker: Or `convertCkErc20ToErc20Fees` or else?

What I mean is also making clear that this variable is particularly use only for conversion of ckErc20 to Erc20

`ckErc20...` might also be interpreted as some fee for a simplem ckErc20 transfer

Does that make sense?

AntonioVentilii-DFINITY: ah! right! 

`ckErc20ToErc20EthereumFees`?

I think it is important to specify somewhere that they are ckETH or ETH-related? or not? if not, what you suggested is ok `convertCkErc20ToErc20Fees`

or even `convertCkErc20ToErc20MaxFees`

peterpeterparker: Works for me

AntonioVentilii-DFINITY: done

src/frontend/src/icp/components/send/IcSendTokenWizard.svelte
```diff
@@ -190,6 +191,15 @@
 		store: initEthereumFeeStore()
 	});
 
+	const { store: ethereumFeeStore } = getContext<EthereumFeeContext>(ETHEREUM_FEE_CONTEXT_KEY);
+
+	// In case we are converting ckERC20 to ERC20, we need to include ckETH related fees in the transaction.
+	let ckErc20ToErc20MaxCkEthFees: bigint | undefined = undefined;
```
peterpeterparker, line 197: Is there any particular reason why you are assigning a state value? We do not need this for the UI right?

If the above is correct and given the fact that the transfer is executed in an imperative way, I would maybe also assign `ckErc20ToErc20MaxCkEthFees` imperatively as we already do for example with `$tokenAsIcToken`. WDYT?

```
// In case....
const ckErc20ToErc20MaxCkEthFees = isNetworkIdETH(networkId) && $tokenCkErc20Ledger ? $ethereumFeeStore?.maxTransactionFee : undefined;

const params: IcTransferParams = {
				to: destination,
				amount: parseToken({
					value: `${amount}`,
					unitName: $token.decimals
				}),
				identity: $authStore.identity,
				progress: (step: ProgressStepsSendIc) => (sendProgressStep = step),
				ckErc20ToErc20MaxCkEthFees
			};

await sendIc({
				...params,
				token: $tokenAsIcToken,
				targetNetworkId: networkId
			});
```

AntonioVentilii-DFINITY: good point! changed

peterpeterparker (approved):
LGTm, thx

PR #2172, opened 2024-08-28
feat(frontend): extract getBalance from Bitcoin API

AntonioVentilii-DFINITY:
# Motivation

After having bumped `ic-js` to latest version, we can extract the new `getBalance` feature of object `BitcoinCanister`.

It will be used for the BTC integration.


src/frontend/src/icp/api/bitcoin.api.ts
```diff
@@ -6,19 +6,21 @@ import { BitcoinCanister, type BitcoinNetwork, type get_utxos_response } from '@
 import { Principal } from '@dfinity/principal';
 import { assertNonNullish } from '@dfinity/utils';
 
+interface BitcoinCanisterParams {
```
peterpeterparker, line 9: If you introduce an interface you can also use it in `const bitcoinCanister = async ({`

`Pick<BitcoinCanisterParams, "identity"...`

peterpeterparker (approved):
LGTM, thx

src/frontend/src/icp/api/bitcoin.api.ts
AntonioVentilii-DFINITY: Only for `bitcoinCanisterId`, because `identity` in `bitcoinCanister` function is `Identity` and not `OptionIdentity` (like the interface).

We assert the `nonNullish` identity in each extraction. Maybe I can do a separate PR where I assert it inside the function `bitcoinCanister`.

peterpeterparker: Ah right, then it's not worth it. Nvm my comment. Thanks for the feedback.

PR #2173, opened 2024-08-28
feat(frontend): remove estimated fee from ckETH conversions

AntonioVentilii-DFINITY:
# Motivation

During a `ckETH --> ETH` conversion, the estimated `ETH` fee was being showed. We remove it in this PR.

### Before

694d01b5-b684-4cd6-9eee-e3e9881a7cf3

### After

Screenshot 2024-08-28 at 21 42 27

### DIFF
--- a/src/frontend/src/icp/components/fee/EthereumFeeContext.svelte
+++ b/src/frontend/src/icp/components/fee/EthereumFeeContext.svelte
@@ -20,9 +20,6 @@
 
 	export let networkId: NetworkId | undefined = undefined;
 
-	let ckETH = false;
-	$: ckETH = isTokenCkEthLedger($tokenWithFallbackAsIcToken);
-
 	let ckErc20 = false;
 	$: ckErc20 = isTokenCkErc20Ledger($tokenWithFallbackAsIcToken);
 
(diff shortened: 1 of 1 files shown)


peterpeterparker (commented):
Is this correct? According [documentation](https://github.com/dfinity/ic/blob/master/rs/ethereum/cketh/docs/cketh.adoc#cost-of-a-withdrawal) the amount the receiver is getting has Ethereum fees deducted.

AntonioVentilii-DFINITY:
Indeed, that is correct! But there are a few points on that, that I have in plan of possibly solving in other PRs if needed:

When we are converting `ckETH` to `ETH`, in the code we ask for approval for a certain amount and then we withdraw that amount. However, that certain amount does not include the `ckETH` related to `ETH` fees. That means, for example, if the user request 10 `ckETH` to convert into ETH, the transaction will work, but the user will receive less than 10 `ETH`, because a part of `ckETH` was burned for `ETH` fees.

That is coherent with what we are showing to the user up until now: we are telling the user that he will exchange 10 `ckETH` to `ETH`, and we tell him that we will subtract a certain amount of `ETH` as fees to cover, so, he will see less that 10 ETH in the account. Completely correct.

But my understanding of the requirements is that we want now show that the fees are all on `ckETH` and not `ETH`, to have not separate the two of them.
I will inquiry again

peterpeterparker:
Thanks for the explanation.

> we want now show that the fees are all on ckETH and not ETH

But, that's incorrect no? ckETH fees === IC network, ETH fees = Ethereum network.

AntonioVentilii-DFINITY:
As discussed with @anedos-dfinity , I will close this PR and create a new one, that is the requirement

PR #2174, opened 2024-08-28
feat(frontend): ckETH fee for ckERC20 conversions is calculated exactly

AntonioVentilii-DFINITY:
# Motivation

The amount of `ckETH` required for a `ckERC20 --> ERC20` conversion was calculated quite conservatively, adding a specific big amount. That is not an issue, since it defines the approved `ckETH` amount that could be burned to cover the `ETH` fees (as explained in the [withdrawal scheme](https://github.com/dfinity/ic/blob/master/rs/ethereum/cketh/docs/ckerc20.adoc#withdrawal-ckerc20-to-erc20)).
Ultimately, the real amount burned is less than that, so the user is not strictly negatively impacted by unexpected costs.

Talking with cross-chain team, they explained that the `ETH` fee required for a specific `ERC20` token withdrawal can be fetched from the minter using the `eip_1559_transaction_price` call. That amount is already stored (and fetched) in Oisy code.

Additionally, the transaction cost of `ckETH` of the ckETH ledger should be included, to pay for the first transaction step of the aforementioned withdrawal scheme.

# Note from cross-chain team

The `max_transaction_fee` fetched with call `eip_1559_transaction_price` from the minter is not real-time: it is re-calculated everytime there is a new withdrawal request to the minter. The same call provides the last update timestamp.

However, after it is calculated, the minter conservatively doubles the value before providing it to the `eip_1559_transaction_price` call.
The scope is the same as mentioned before: that amount should be the approved `ckETH` amount that could be burned to cover the `ETH` fees, but ultimately it most probably be less in reality.

In any case, the minter rejects the transaction in the eventuality that the approved `ckETH` amount to be burned is below a newly-refreshed `ETH` fee.

# Changes

- Removed dependency from hard-coded conservative offset.
- For ckETH, transaction fee is the fetched amount from `eip_1559_transaction_price`.
- For ckERC20, apart from the above, we add the ckETH ledger fee.
- Include docstring documentation and explanations.

# Tests

In `staging` env, `ckSepoliaUSDC --> SepoliaUSDC` worked correctly

### IC tx [348](https://dashboard.internetcomputer.org/ethereum/yfumr-cyaaa-aaaar-qaela-cai/transaction/348)

Screenshot 2024-08-30 at 07 16 25
Screenshot 2024-08-30 at 07 19 22

### Sepolia tx [0x8c6a7ac99ea4fd95dcc4028fa67eb97352c7281b6644991bbee9c8b77ba73fe6](https://sepolia.etherscan.io/tx/0x8c6a7ac99ea4fd95dcc4028fa67eb97352c7281b6644991bbee9c8b77ba73fe6)

Screenshot 2024-08-30 at 07 19 51

### Transaction flow

https://github.com/user-attachments/assets/865a0df9-649c-4266-812a-abfa51b8f805


peterpeterparker (approved):
LGTM, thanks

We discussed this offline.

PR #2175, opened 2024-08-29
feat(frontend): Update ckErc20 Tokens

sa-github-api:
Modifications have been made to the list of ckErc20 tokens deployed by the Orchestrator and/or to their icons.


peterpeterparker:
@AntonioVentilii-DFINITY We don't use those two icons. For CK tokens, they are fetched at runtime, and for ERC20 tokens, we use the official logo.

Should we merge these into the codebase but exclude them from being used, or should we enhance the script that imports the tokens to only populate icons based on specific rules? However the rules can't be fully automated, that would be more a manual list of icons to import or not.

AntonioVentilii-DFINITY:
Good question... For what I understood, these will not be the last tokens to have these rules, correct? If so, I would prefer to make a list in the script, something like "tokens that do not need icons to be fetched", and expand that.

But if only these two are the ones in question, forever (but I am not sure of it), then better to have them and write some comment/exception in the usage

peterpeterparker:
Both sounds good to me at this point, which one you prefer?

AntonioVentilii-DFINITY:
I am more on the first one... it gives more flexibility, if something changes in the future and we have more exceptions

peterpeterparker:
Let's do that. Can you take care of it?

We can close this automatic PR then.

AntonioVentilii-DFINITY:
Done with PR https://github.com/dfinity/oisy-wallet/pull/2178

(there was already a place to put the tokens we want to exclude)

PR #2182, opened 2024-08-29
fix(devops): Reproducible docker

bitdivine:
# Motivation
In the Dockerfile, the platform is not specified, so a different CPU can mean that a different Ubuntu is used and the build is not reproducible.

# Changes
Specify the Linux CPU type.

# Tests


AntonioVentilii-DFINITY (approved):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04 as deps
+FROM --platform=linux/amd64 ubuntu:22.04 AS deps
```
peterpeterparker, line 1: In II:

```
FROM --platform=linux/amd64 ubuntu@sha256:bbf3d1baa208b7649d1d0264ef7d522e1dc0deeeaaf6085bf8e4618867f03494 as deps
```

https://github.com/dfinity/internet-identity/blob/4bae280aaaf6f1de3bd330b3f8baf1f0dd0466a4/Dockerfile#L8C1-L8C115

bitdivine: Good idea.

PR #2184, opened 2024-08-30
feat(backend): add Bitcoin balance call

AntonioVentilii-DFINITY:
# Motivation

We add the backend call to fetch the BTC balance of a generic address, based on the [IC documentation](https://internetcomputer.org/docs/current/developer-docs/multi-chain/bitcoin/using-btc/read-state#reading-current-balance).


src/backend/src/lib.rs
```diff
@@ -266,6 +268,20 @@ async fn caller_btc_address(network: BitcoinNetwork) -> String {
     public_key_to_p2pkh_address(network, &ecdsa_pubkey_of(&ic_cdk::caller()).await)
 }
 
+/// Returns the balance of the given bitcoin address.
+///
+/// Relies on the `bitcoin_get_balance` endpoint.
+/// See [IC Bitcoin Documentation](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_get_balance)
+#[update]
+async fn btc_balance(request: GetBalanceRequest) -> Result<u64, String> {
```
peterpeterparker, line 276: We don't have the list yet, but given that `bitcoin_get_balance` can only be called by a backend canister as it requires cycles—since it is not free—shouldn't we add a guard to this function to allow only Oisy's users? Otherwise, we potentially introduce a way for any dapp to use this proxy to call get balance for free.

peterpeterparker (commented):
In addition which can potential opens quite some questions, we also needs test for backend PR but first, let's discuss my question.

src/backend/src/lib.rs
AntonioVentilii-DFINITY: There is the guard `caller_is_allowed`, but I think it is used for migration scopes, right @bitdivine ?

lmuntaner: Another idea is to give the balance of the caller and not expect an address in the params, only the network. This way we also avoid Oisy users from checking balances of other users with this endpoint.

I think `caller_is_allowed` checks that the caller is in the config. We probably need a new guard for this.

AntonioVentilii-DFINITY:
As discussed with @lmuntaner , this feature will be reviewed. So for now, I am closing this PR

PR #2185, opened 2024-08-30
feat(frontend): show all ck-conversions fees in ck-denominations

AntonioVentilii-DFINITY:
# Motivation

Since the ck-conversions fees related to Ethereum transaction fees are always paid in `ckETH`, we change the component `EthereumEstimatedFee` to reflect that. The change is visible happens only in `ckETH --> ETH` conversion modals.

### NOTE

Both situations below end in the same place and results for the user, nothing changes in final amounts, only the dialectics on how the Ethereum fees are paid by the user, that will reflect [what is happening in the background effectively](https://github.com/dfinity/ic/blob/master/rs/ethereum/cketh/docs/cketh.adoc#cost-of-a-withdrawal), that is how it is effectively already implemented in the Oisy code.

### Before:

The Ethereum fees were showed in `ETH`, telling the users that it is the amount that it would be subtracted from the final `ETH` amount that they would expect, after having converted the `ckETH` input by the users into `ETH`.
_Example of user experience:_
1. User input `10 ckETH` to be converted.
2. Oisy shows a fee of `1 ETH`.
3. `10 ckETH` --> `10 ETH`.
4. Ethereum fees are paid: `10 ETH - 1 ETH = 9 ETH`.
5. User expects `9 ETH`.

### After:

The Ethereum fees will be showed in `ckETH`, telling the users that it is the amount that it would be subtracted from the `ckETH` input amount, before converting to `ETH`.
_Example of user experience:_
1. User input `10 ckETH` to be converted.
2. Oisy shows a fee of `1 ckETH`.
3. Ethereum fees are paid: `10 ckETH - 1 ckETH = 9 ckETH`.
3. `9 ckETH` --> `9 ETH`.
4. User expects `9 ETH`.

# Changes

- Modify `ethereumFeeTokenCkEth` derived store to return `ckETH` as Fee Token for `ckETH`.
- Adapt `EthereumEstimatedFee` to show always the ck-denomination.

The above changes are ok, since `ethereumFeeTokenCkEth` is used only in two circumstances for now: showing the denomination of the Ethereum Fee and asserting the `ckETH` balance in case of `ckERC20 --> ERC20` conversions (not affected by the changes). Furthermore, `EthereumEstimatedFee` is showed only for ck-conversions.

# Tests

### Before

Screenshot 2024-08-29 at 16 32 13

## After

Screenshot 2024-08-30 at 07 25 15

### DIFF
--- a/src/frontend/src/icp/components/fee/EthereumEstimatedFee.svelte
+++ b/src/frontend/src/icp/components/fee/EthereumEstimatedFee.svelte
@@ -10,22 +10,17 @@
 		ETHEREUM_FEE_CONTEXT_KEY,
 		type EthereumFeeContext
 	} from '$icp/stores/ethereum-fee.store';
-	import { isTokenCkErc20Ledger } from '$icp/utils/ic-send.utils';
 	import { ethereumFeeTokenCkEth } from '$icp/derived/ethereum-fee.derived';
-	import { tokenWithFallbackAsIcToken } from '$icp/derived/ic-token.derived';
 	import type { Token } from '$lib/types/token';
 	import FeeAmountDisplay from '$icp-eth/components/fee/FeeAmountDisplay.svelte';
-
-	let ckEr20 = false;
-	$: ckEr20 = isTokenCkErc20Ledger($tokenWithFallbackAsIcToken);
 
 	const { store } = getContext<EthereumFeeContext>(ETHEREUM_FEE_CONTEXT_KEY);
 
 	let feeToken: Token;
-	$: feeToken = ckEr20 ? $ethereumFeeTokenCkEth ?? $ckEthereumNativeToken : $ckEthereumNativeToken;
+	$: feeToken = $ethereumFeeTokenCkEth ?? $ckEthereumNativeToken;
 
 	let maxTransactionFee: bigint | undefined | null = undefined;
-	$: maxTransactionFee = $store?.maxTransactionFee;
+	$: maxTransactionFee = $store?.maxTransactionFee ?? 100_000_000_000_000_000n;
 </script>
 
 {#if nonNullish($store)}
--- a/src/frontend/src/icp/components/fee/EthereumEstimatedFee.svelte
+++ b/src/frontend/src/icp/components/fee/EthereumEstimatedFee.svelte
@@ -20,7 +20,7 @@
 	$: feeToken = $ethereumFeeTokenCkEth ?? $ckEthereumNativeToken;
 
 	let maxTransactionFee: bigint | undefined | null = undefined;
-	$: maxTransactionFee = $store?.maxTransactionFee ?? 100_000_000_000_000_000n;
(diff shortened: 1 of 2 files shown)


peterpeterparker:
I've got two questions regarding your last screenshot:

1. Should we rename "Estimated Ethereum Fee (Updated 30 seconds" into "Estimated Fee (Updated 30 seconds") - i.e. remove the keyword "Ethereum" from the title

2. In your screenshot the balance is 0.000000077 ckETH but in the message it's 0.00000008 ckETH. I guess it's rounding. Should we use the same rounding as in the display of the balance in the error message?

image

src/frontend/src/icp/components/fee/EthereumEstimatedFee.svelte
```diff
@@ -10,22 +10,17 @@
...
-
 	const { store } = getContext<EthereumFeeContext>(ETHEREUM_FEE_CONTEXT_KEY);
 
 	let feeToken: Token;
-	$: feeToken = ckEr20 ? $ethereumFeeTokenCkEth ?? $ckEthereumNativeToken : $ckEthereumNativeToken;
+	$: feeToken = $ethereumFeeTokenCkEth ?? $ckEthereumNativeToken;
 
 	let maxTransactionFee: bigint | undefined | null = undefined;
-	$: maxTransactionFee = $store?.maxTransactionFee;
+	$: maxTransactionFee = $store?.maxTransactionFee ?? 100_000_000_000_000_000n;
```
peterpeterparker, line 23: What's `100_000_000_000_000_000n`?

AntonioVentilii-DFINITY: oh wow, sorry, it was just a test I was doing, shouldn't be there

AntonioVentilii-DFINITY:
1. I am looking with Artem how to re-visit the wording there (or even adding some info). But if you prefer I can already remove that word. The point is that we should tell in some way that those fees are related to the Ethereum fees

AntonioVentilii-DFINITY:
2. agreed! will do in another PR

src/frontend/src/icp/components/fee/EthereumEstimatedFee.svelte
AntonioVentilii-DFINITY: I updated with a Staging screenshot

peterpeterparker (approved):
LGTM, thanks

anedos-dfinity:
@AntonioVentilii-DFINITY In your after xshot, the Ethereum fee is `0.01...` ckSepoliaETH, which is still really high. Is this amount reflective of how it will be in mainnet, i.e., will the user see an amount close to `0.01..` of ckETH ?

AntonioVentilii-DFINITY:
@anedos-dfinity indeed, that was just a test value in local replica; I updated the screenshot with a staging version

PR #2186, opened 2024-08-30
style(frontend): show same decimals as balance in fee

AntonioVentilii-DFINITY:
# Motivation

When there is an error for insufficient balance, we would like to show the same number as decimals as the balance already showed.

# Changes

- Add `feeDecimals` to Fee Context.
- Set `feeDecimals` throughout the code using already existing token values used by `feeSymbol` and `feeTokenId`.
- Add prop `feeDecimals` to `FeeAmountDisplay`.

# Tests

### Before

image

### After

Screenshot 2024-08-30 at 07 50 19


peterpeterparker (approved):
LGTM, thx

PR #2187, opened 2024-08-30
feat(frontend): show fee last updated time

AntonioVentilii-DFINITY:
# Motivation

# Changes

# Tests

https://github.com/user-attachments/assets/479c555a-cd85-4a7d-abd8-6c796f0b6eb9


AntonioVentilii-DFINITY:
I discussed it with Artem and it does not really make sense to show the user the update time if it is that long ago.

I discussed with cross-chain team about possible solutions, maybe making it so that it is not older that 1 hour (or less). But the discussion will take longer.

For now, I will close this PR.

PR #2192, opened 2024-08-31
refactor(frontend): create checks to see if converting ck tokens

AntonioVentilii-DFINITY:
# Motivation

Throughout the code, we often check if we are dealing with a ck-conversion (either `ckETH` or `ckERC20`). The scope of this PR is to centralize the logic in a single place and re-use.


src/frontend/src/icp/components/fee/EthereumFeeContext.svelte
```diff
@@ -16,18 +15,19 @@
...
 
 	export let networkId: NetworkId | undefined = undefined;
 
-	let ckETH = false;
-	$: ckETH = isTokenCkEthLedger($tokenWithFallbackAsIcToken);
-
-	let ckErc20 = false;
-	$: ckErc20 = isTokenCkErc20Ledger($tokenWithFallbackAsIcToken);
+	let ckEthConvert = false;
+	$: ckEthConvert = isConvertCkEthToEth({ token: $tokenAsIcToken, networkId });
```
peterpeterparker, line 27: The logic is changing here no, are you sure the `networkId` passed as property is the same as the one derived in `$tokenWithFallbackAsIcToken`? Wouldn't it be more safe to stick to `$tokenWithFallbackAsIcToken`?

AntonioVentilii-DFINITY: in this specific case, networkId is given as prop to the component:

	`export let networkId: NetworkId | undefined = undefined;`
	
I wanted to change from `tokenWithFallbackAsIcToken` to `tokenAsIcToken` because the only difference between the two of them is that the first one is never `undefined` (that is not needed anymore with the new check-function).

But I agree that it could lead to change in logic, so I can change it back if you want

peterpeterparker: Yes, please do.

AntonioVentilii-DFINITY: done

peterpeterparker (approved):
LGTM, thx

PR #2193, opened 2024-09-01
feat(frontend): remove tab selector when signed out

AntonioVentilii-DFINITY:
# Motivation

When signed out it was possible to select the blurred tokens list using `Tab` clicks. We remove this possibility.

Screenshot 2024-09-01 at 08 32 00

# Tests

In local replica, clicking in loop `Tab`, will never reach the tokens list.

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokenCardWithUrl.svelte
+++ b/src/frontend/src/lib/components/tokens/TokenCardWithUrl.svelte
@@ -5,6 +5,7 @@
 	import { i18n } from '$lib/stores/i18n.store';
 
 	export let token: Token;
+	export let disableTabSelector: boolean = false;
 
 	let url: string;
 	$: url = transactionsUrl({ token });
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/tokens/TokenCardWithUrl.svelte
```diff
@@ -5,6 +5,7 @@
 	import { i18n } from '$lib/stores/i18n.store';
 
 	export let token: Token;
+	export let disableTabSelector: boolean = false;
```
peterpeterparker, line 8: Linter does not complain about `: boolean` being not necessary?

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/lib/components/tokens/TokenCardWithUrl.svelte
AntonioVentilii-DFINITY: No, but I can introduce the ESlint rule

peterpeterparker: Weird, I asked because I thought it was the default eslint configuration of SvelteKit projects.

PR #2194, opened 2024-09-01
refactor(frontend): simplify code to calculate USD balance

AntonioVentilii-DFINITY:
# Motivation

To reduce repetition of codes and optimize processing, function `usdValue` will accepts direct values instead of stores.

### DIFF
--- a/src/frontend/src/lib/utils/exchange.utils.ts
+++ b/src/frontend/src/lib/utils/exchange.utils.ts
@@ -1,23 +1,21 @@
 import { ZERO } from '$lib/constants/app.constants';
-import type { BalancesData } from '$lib/stores/balances.store';
-import type { CertifiedStoreData } from '$lib/stores/certified.store';
-import type { ExchangesData } from '$lib/types/exchange';
 import type { Token } from '$lib/types/token';
 import { formatToken } from '$lib/utils/format.utils';
+import type { BigNumber } from '@ethersproject/bignumber';
 
 export const usdValue = ({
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/utils/exchange.utils.ts
```diff
@@ -1,23 +1,21 @@
...
 }): number =>
 	Number(
 		formatToken({
-			value: balances?.[id]?.data ?? ZERO,
+			value: balance ?? ZERO,
 			unitName: decimals,
 			displayDecimals: decimals
 		})
-	) * (exchanges?.[id]?.usd ?? 0);
+	) * exchangeRate;
```
peterpeterparker, line 21: The function used to fallback to `0` when no exchange rate was fetched. Is this behavior preserved? Do we need to preserve it? Was it modifed when we introduce the usd balance within `TokenUI`?

AntonioVentilii-DFINITY: that was a redundant check just because we would pass `ExchangesData` instead of `number`. But in effectivity, that value would have never been undefined, since we were checking before using the function:

`usdBalance: nonNullish($exchanges?.[token.id]?.usd) ? ....`

and that is the only instance where we use the function

peterpeterparker: Oki doki, thanks for the feedback.

peterpeterparker (approved):
LGTM, thanks

PR #2195, opened 2024-09-01
feat(frontend): revert unused prop formattedBalance

AntonioVentilii-DFINITY:
# Motivation

Prop `formattedBalance` (introduced in PR #2150 ) of `TokenUI` is not used, since there is prop `balance` that is the most straightforward way of getting a token balance value. Plus, we slightly reduce computational time.


peterpeterparker (approved):
LGTM, thanks

However, why is it not used anymore?

AntonioVentilii-DFINITY:
@peterpeterparker alas, it was never used... the plan was to provide it to the calculation of USD balance, and elsewhere too, but the prop `balance` is more than enough

PR #2197, opened 2024-09-02
refactor(frontend): homologate list of tokens and total balance

AntonioVentilii-DFINITY:
# Motivation

To have more consistent, we set the calculation of `ExchangeBalance` with the store `combinedDerivedSortedNetworkTokensUi`, since that is the list of tokens used in `TokensSignedIn`.

The aim is to avoid possible small lags between the balance loading and the tokens loading.

The only difference between the old one used and the new one is the sorting by balance.


peterpeterparker (approved):
LGTM, thx

PR #2201, opened 2024-09-02
refactor(frontend): merge sorting util with pinning utils for tokens

AntonioVentilii-DFINITY:
# Motivation

To reduce the computational time, we merge two of the loops needed for the derived store `combinedDerivedSortedNetworkTokensUi`:
1. First loop to create props `balance` and `usdBalance`, done in store `combinedDerivedEnabledNetworkTokensUi`.
2. Second loop to separate the tokens with balance, from the one that do not have it.

# Changes

- Add the `balance` and `usdBalance` calculation in function `pinTokensWithBalanceAtTop`.
- Adapt tests to the new function, providing a simpler mocked `usdValue` function.
- Adapt derived store `combinedDerivedSortedNetworkTokensUi` to feed the correct store stores the new function.
- Removed deprecated unused store `combinedDerivedEnabledNetworkTokensUi`.
- Unrelated: we renamed a few mock values for module `tokens.utils.spec`, to make the code less verbose.

# Tests

No changes in local replica.

Screenshot 2024-09-02 at 13 45 34


src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -55,17 +59,45 @@ export const sortTokens = ({
...
+					})
+				: undefined;
+
+			const tokenUI: TokenUi = {
+				...token,
+				balance,
+				usdBalance
+			};
+
+			(usdBalance ?? 0) > 0 || (balance ?? ZERO).gt(0)
```
peterpeterparker, line 95: Not necessarely for this PR and a bit an feedback but, given that the function isn't a single line anymore, do we have chance here to introduce a `if` and transform the `push` into a more functional programming approach?

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: i will do it in this one

AntonioVentilii-DFINITY: @peterpeterparker before I merge, have a look, if that what you meant

peterpeterparker: Almost there, still not fully functional. We can keep the push in this PR and try to do that later or not.

PR #2205, opened 2024-09-02
ci(e2e): re-use oisy backend

DenysKarmazynDFINITY:
# Motivation

The idea is to cache backend WASM and expose it as an artifact in order to fasten e2e and other pipelines. The cache will be invalidated on backend files change. In this PR, I'm using the generated `backend.wasm.gz` in Dockerfile.e2e by updating `dfx.json` to use `custom` build instead of `rust`. The time needed to run FE e2e tests is now decreased from ~22mins to ~8mins.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     paths:
       - 'e2e/**'
-      - 'src/frontend/**'
+
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -12,9 +12,6 @@ jobs:
   oisy-backend-wasm:
     runs-on: ubuntu-20.04
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
       - name: Build oisy-backend WASM
         uses: ./.github/actions/oisy-backend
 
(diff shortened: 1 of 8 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -4,12 +4,23 @@ on:
...
+
   workflow_dispatch:
 
 jobs:
 
+  oisy-backend-wasm:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
AntonioVentilii-DFINITY, line 16: Do you need this one? I see you do it already in `actions/oisy-backend`

.github/workflows/update-snapshots.yml
```diff
@@ -5,8 +5,18 @@ on:
 
 jobs:
 
+  oisy-backend-wasm:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
AntonioVentilii-DFINITY, line 12: same as previous comment

AntonioVentilii-DFINITY (approved):
LGTM! Just a question

.github/workflows/e2e-tests.yml
DenysKarmazynDFINITY: Yes, can be removed. Done.

DenysKarmazynDFINITY:
Thanks @AntonioVentilii-DFINITY for checking the PR! 

@bitdivine @peterpeterparker could you please take a look, too? 

* Example of the pipeline where `backend.wasm.gz` was not available in the cache - https://github.com/dfinity/oisy-wallet/actions/runs/10680149182
* Example of the pipeline where `backend.wasm.gz` was available in the cache - https://github.com/dfinity/oisy-wallet/actions/runs/10682468976

PR #2206, opened 2024-09-02
feat(frontend): Set up Bitcoin transactions page

lmuntaner:
# Motivation

The transactions page for Bitcoin was not showing balance, nor logo, nor nice background.

The error was that it didn't find any token in the pageToken derived store and that the btc didn't have any assigned background.

# Changes

* Add the bitcoin tokens as the tokens to search the pageToken.
* Set up a BtcTransactions component.
* Add a background in the Hero for the Bitcoin network. (To be determined with design which one, but for now, same as ICP).
* Render BtcTransactions if network is BTC.
* New derived store `networkBitcoin`.

# Tests

Tested manually that Hero appears as expected.

### DIFF
--- a/src/frontend/src/btc/components/BtcTransactions.svelte
+++ b/src/frontend/src/btc/components/BtcTransactions.svelte
@@ -1,0 +1,10 @@
+<script lang="ts">
+	import Header from '$lib/components/ui/Header.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+</script>
+
+<Header>
+	<h2 class="text-base">{$i18n.transactions.text.title}</h2>
+</Header>
+
+<!-- TODO: Implement https://dfinity.atlassian.net/browse/GIX-2883 -->
--- a/src/frontend/src/lib/components/hero/Hero.svelte
+++ b/src/frontend/src/lib/components/hero/Hero.svelte
@@ -49,6 +49,11 @@
 			background: radial-gradient(66.11% 97.11% at 50% 115.28%, #300097 0%, #1f005e 100%);
 		}
 
+		&.btc {
+			// TODO: Choose background for bitcoin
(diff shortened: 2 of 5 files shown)


lmuntaner:
Before:

Screenshot 2024-09-02 at 15 32 23

This PR:

Screenshot 2024-09-02 at 15 30 31

peterpeterparker (commented):
Can you provide this in two separate PRs please? e.g. stores part followed by UI usage

lmuntaner:
> Can you provide this in two separate PRs please? e.g. stores part followed by UI usage

Sure, I love small PRs. First one: https://github.com/dfinity/oisy-wallet/pull/2214

What is the rule to know when a PR is the right size? A PR can only touch one layer like api, services, components, etc? Or is it common sense?

peterpeterparker:
> What is the rule to know when a PR is the right size? A PR can only touch one layer like api, services, components, etc? Or is it common sense?

I'm not sure what the informal rule is. It’s a bit like grouping small PRs and related changes together. In this particular PR, the concern is with the stores, as we've recently had issues with an endless loop involving derived stores.

Perhaps @AntonioVentilii-DFINITY has a better definition?

AntonioVentilii-DFINITY:
I usually base them more on single revertable (or modifiable) changes.  For example, in this case, the `Hero` background would go in a different PR, even if small. Just because we could revert or change it after merging, and it would have a simpler history. 
I'm not really sure if I explained correctly. 

P.S. for me, it is even easier to follow: I submit 5 small PRs at the same time, and if there is a review in one of them, the others are not blocked. Plus, I can change them independently and really verify that one does not affect the other, unrelated (that always means good)

lmuntaner:
Closing in favor of #2223 and #2214

PR #2207, opened 2024-09-02
feat(frontend): split address store [revert PR #2012]

AntonioVentilii-DFINITY:
# Motivation

As discussed in PR #2067 , we would actually need to split the `addressStore` into separate stores for each address. That was how it was done before PR #2012 . So, as first step we partially revert that PR, creating a specific `initAddressStore` function and using it for two separate stores for BTC mainnet and ETH.

They will be used in another PR, where we substitute the usage of `addressStore`.


src/frontend/src/lib/stores/address.store.ts
```diff
@@ -1,10 +1,32 @@
...
+import { writable, type Readable } from 'svelte/store';
 
-export type AddressData = CertifiedData<Address>;
+export type AddressData<T extends Address = Address> = CertifiedData<T>;
 
 export type OptionAddressData = StorageStoreData<AddressData>;
 
 export const addressStore = initCertifiedSetterStore<AddressData>();
+
+export type OptionCertifiedAddressData<T extends Address> = AddressData<T> | undefined | null;
```
peterpeterparker, line 13: `OptionCertifiedAddressData` and `AddressStore` needs to be exposed (`export`)?

peterpeterparker (approved):
LGTM, thanks

PS Was wondering if you want to mark `addressStore` as deprecated but, I guess you'll remove it soon so not worth it.

src/frontend/src/lib/stores/address.store.ts
AntonioVentilii-DFINITY: i will un-expose in this PR, and check when I will need to expose them in the other one

PR #2209, opened 2024-09-03
refactor(frontend): use split address stores

AntonioVentilii-DFINITY:
# Motivation

As discussed in PR #2067 and after PR #2207, we are going to use the split address stores for BTC mainnet and ETH .

# Changes

- Change derived store to use the split stores directly: `ethAddressStore` and `btcAddressMainnetStore`.
- In `AddressGuard`, use the `ethAddressStore` directly.
- Adjust address services to receive the specific store to set for each token.
- Remove deprecated types and derived stores.

# Tests

No change in local replica.


src/frontend/src/lib/derived/address.derived.ts
```diff
@@ -1,38 +1,32 @@
...
 import { derived, type Readable } from 'svelte/store';
 
-export const addressNotLoaded: Readable<boolean> = derived([addressStore], ([$addressStore]) =>
-	isNullish($addressStore)
-);
-
-const btcAddressData: Readable<OptionAddressData> = derived(
-	[addressStore],
-	([$addressStore]) => $addressStore?.[BTC_MAINNET_TOKEN_ID]
+export const addressNotLoaded: Readable<boolean> = derived(
```
peterpeterparker, line 12: You rename it right after this PR or we should add a TODO? ah or maybe it gonna become `addressesNotLoad` with both eth and bitcoin, that's possible.

src/frontend/src/lib/services/address.services.ts
```diff
@@ -30,21 +34,23 @@ import { get } from 'svelte/store';
 const loadTokenAddress = async <T extends Address>({
 	tokenId,
 	getAddress,
-	setIdbAddress
+	setIdbAddress,
+	addressStore
 }: {
 	tokenId: TokenId;
```
peterpeterparker, line 40: We still need the tokenId here?

src/frontend/src/lib/stores/address.store.ts
```diff
@@ -1,24 +1,18 @@
...
-import type { StorageStoreData } from '$lib/stores/storage.store';
 import type { Address, BtcAddress, EthAddress } from '$lib/types/address';
 import type { CertifiedData } from '$lib/types/store';
 import { writable, type Readable } from 'svelte/store';
 
-export type AddressData<T extends Address = Address> = CertifiedData<T>;
+type AddressData<T extends Address> = CertifiedData<T>;
 
-export type OptionAddressData = StorageStoreData<AddressData>;
+export type OptionAddressData<T extends Address> = AddressData<T> | undefined | null;
```
peterpeterparker, line 7: Can you provide this change in a separate PR?

src/frontend/src/lib/derived/address.derived.ts
AntonioVentilii-DFINITY: i rename it in the next PR, but i will add a TODO either way

src/frontend/src/lib/services/address.services.ts
AntonioVentilii-DFINITY: yes, but only for the error message

```typescript
toastsError({
	msg: {
			$symbol: tokenId.description ?? ''
		})
	},
	err
});
```

i would prefer not to generalize the error message, but WDYT?

src/frontend/src/lib/stores/address.store.ts
AntonioVentilii-DFINITY: will do

src/frontend/src/lib/services/address.services.ts
peterpeterparker: Makes sense. Maybe we can pass the `tokenSymbol` as parameter in another PR or something, just thinking out at loud. When I read `tokenId` passed to a function I automatically think that it will be use for some feature but, that's maybe just me. All good. Thanks for the feedback.

src/frontend/src/lib/stores/address.store.ts
AntonioVentilii-DFINITY: PR https://github.com/dfinity/oisy-wallet/pull/2213

peterpeterparker: Did you?

AntonioVentilii-DFINITY: ah, I meant, I will do after this one

peterpeterparker: Oki doki

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/stores/address.store.ts
AntonioVentilii-DFINITY: Done in PR #2220

PR #2210, opened 2024-09-03
refactor(frontend): wallets loader for all enabled ICRC tokens

AntonioVentilii-DFINITY:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/icp/components/core/IcLoaderWallets.svelte
+++ b/src/frontend/src/icp/components/core/IcLoaderWallets.svelte
@@ -1,0 +1,55 @@
+<script lang="ts">
+	import { onDestroy } from 'svelte';
+	import { initIcrcWalletWorker } from '$icp/services/worker.icrc-wallet.services';
+	import { initIcpWalletWorker } from '$icp/services/worker.icp-wallet.services';
+	import type { WalletWorker } from '$icp/types/ic-listener';
+	import { enabledIcNetworkTokens } from '$lib/derived/network-tokens.derived';
+	import type { TokenId } from '$lib/types/token';
+	import { debounce, nonNullish } from '@dfinity/utils';
+
+	let workers: Map<TokenId, WalletWorker> = new Map<TokenId, WalletWorker>();
+
+	const manageWorkers = async () => {
+		const unusedWorkers: TokenId[] = Array.from(workers.keys()).filter(
+			(tokenId) => !$enabledIcNetworkTokens.some(({ id: icTokenId }) => icTokenId === tokenId)
+		);
+
+		unusedWorkers.forEach((tokenId) => {
+			const worker = workers.get(tokenId);
+
+			if (nonNullish(worker)) {
+				worker.stop();
+				workers.delete(tokenId);
+			}
+		});
+
+		await Promise.all(
+			$enabledIcNetworkTokens.map(async (token) => {
+				if (!workers.has(token.id)) {
+					const worker = await (token.standard === 'icrc'
+						? initIcrcWalletWorker(token)
+						: initIcpWalletWorker());
+
+					worker.start();
+					workers.set(token.id, worker);
+				}
+			})
+		);
+	};
+
+	const debounceManageWorkers = debounce(manageWorkers, 500);
+
+	$: $enabledIcNetworkTokens, debounceManageWorkers();
+
+	onDestroy(() => {
+		console.log('onDestroy');
+		workers.forEach((worker) => worker.stop());
+		workers.clear();
+	});
+
+	const triggerTimer = () => workers.forEach((worker) => worker.trigger());
+</script>
+
+<svelte:window on:oisyTriggerWallet={triggerTimer} />
+
+<slot />
--- a/src/frontend/src/lib/components/core/Listener.svelte
+++ b/src/frontend/src/lib/components/core/Listener.svelte
@@ -2,7 +2,6 @@
 	import type { ComponentType } from 'svelte';
 	import type { OptionToken } from '$lib/types/token';
 	import EthListener from '$eth/components/core/EthListener.svelte';
-	import IcListener from '$icp/components/core/IcListener.svelte';
 	import { isNetworkIdBitcoin, isNetworkIdICP } from '$lib/utils/network.utils';
 	import { isNullish } from '@dfinity/utils';
 	import NoListener from '$lib/components/core/NoListener.svelte';
@@ -16,7 +15,7 @@
 		isNullish(token) || !$authSignedIn
 			? NoListener
 			: isNetworkIdICP(token.network.id)
-				? IcListener
+				? NoListener
 				: isNetworkIdBitcoin(token.network.id)
 					? BitcoinListener
 					: EthListener;
--- a/src/frontend/src/lib/derived/network-tokens.derived.ts
+++ b/src/frontend/src/lib/derived/network-tokens.derived.ts
@@ -1,4 +1,5 @@
 import type { Erc20Token } from '$eth/types/erc20';
+import type { IcToken } from '$icp/types/ic';
 import { exchanges } from '$lib/derived/exchange.derived';
 import { pseudoNetworkChainFusion, selectedNetwork } from '$lib/derived/network.derived';
 import { tokens, tokensToPin } from '$lib/derived/tokens.derived';
@@ -30,6 +31,14 @@ export const enabledErc20NetworkTokens: Readable<Erc20Token[]> = derived(
 	[enabledNetworkTokens],
 	([$enabledNetworkTokens]) =>
 		$enabledNetworkTokens.filter(({ standard }) => standard === 'erc20') as Erc20Token[]
+);
+
+export const enabledIcNetworkTokens: Readable<IcToken[]> = derived(
+	[enabledNetworkTokens],
+	([$enabledNetworkTokens]) =>
+		$enabledNetworkTokens.filter(
+			({ standard }) => standard === 'icp' || standard === 'icrc'
+		) as IcToken[]
 );
 
 /**
--- a/src/frontend/src/lib/components/core/Loaders.svelte
+++ b/src/frontend/src/lib/components/core/Loaders.svelte
@@ -4,16 +4,19 @@
 	import ExchangeWorker from '$lib/components/exchange/ExchangeWorker.svelte';
 	import Loader from '$lib/components/core/Loader.svelte';
 	import LoaderMetamask from '$lib/components/core/LoaderMetamask.svelte';
+	import IcLoaderWallets from '$icp/components/core/IcLoaderWallets.svelte';
 </script>
 
 <AddressGuard>
 	<Loader>
 		<LoaderBalances>
-			<ExchangeWorker>
-				<LoaderMetamask>
-					<slot />
-				</LoaderMetamask>
-			</ExchangeWorker>
+			<IcLoaderWallets>
+				<ExchangeWorker>
+					<LoaderMetamask>
+						<slot />
+					</LoaderMetamask>
+				</ExchangeWorker>
+			</IcLoaderWallets>
 		</LoaderBalances>
 	</Loader>
 </AddressGuard>


src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -32,6 +33,14 @@ export const enabledErc20NetworkTokens: Readable<Erc20Token[]> = derived(
 		$enabledNetworkTokens.filter(({ standard }) => standard === 'erc20') as Erc20Token[]
 );
 
+export const enabledIcNetworkTokens: Readable<IcToken[]> = derived(
+	[enabledNetworkTokens],
+	([$enabledNetworkTokens]) =>
+		$enabledNetworkTokens.filter(
+			({ standard }) => standard === 'icp' || standard === 'icrc'
```
AntonioVentilii-DFINITY, line 40: here i prefer this instead of

`['icp', 'icrc'].includes(....`

because this is type checked, i.e. it would give an error if there was a mistype in the right-hand side of the comparison that does not match one of the possible values of `TokenStandard`.

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
```diff
@@ -0,0 +1,71 @@
...
+	import { initIcpWalletWorker } from '$icp/services/worker.icp-wallet.services';
+	import type { WalletWorker } from '$icp/types/ic-listener';
+	import { enabledIcNetworkTokens } from '$lib/derived/network-tokens.derived';
+	import type { TokenId } from '$lib/types/token';
+	import { debounce, nonNullish } from '@dfinity/utils';
+	import { toastsError } from '$lib/stores/toasts.store';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+
+	let workers: Map<TokenId, WalletWorker> = new Map<TokenId, WalletWorker>();
```
peterpeterparker, line 13: `const`?

src/frontend/src/lib/components/core/Listener.svelte
```diff
@@ -16,7 +15,7 @@
 		isNullish(token) || !$authSignedIn
 			? NoListener
 			: isNetworkIdICP(token.network.id)
-				? IcListener
```
peterpeterparker, line 19: The module/file `IcListener` should no be deleted?

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
```diff
@@ -0,0 +1,71 @@
...
+	const debounceManageWorkers = debounce(manageWorkers, 500);
+
+	$: $enabledIcNetworkTokens, debounceManageWorkers();
+
+	onDestroy(() => {
+		workers.forEach((worker) => worker.stop());
+		workers.clear();
+	});
+
+	const triggerTimer = () => workers.forEach((worker) => worker.trigger());
```
peterpeterparker, line 66: Only the selected worker should be triggered no?

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
```diff
@@ -0,0 +1,71 @@
...
+							},
+							err
+						});
+					}
+				}
+			})
+		);
+	};
+
+	const debounceManageWorkers = debounce(manageWorkers, 500);
```
peterpeterparker, line 57: Why debouncing? Why 500ms? Or generally speaking, there isn't a way to not require debounce? I always see debounce as a last resort.

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
```diff
@@ -0,0 +1,71 @@
...
+
+	const manageWorkers = async () => {
+		const unusedWorkers: TokenId[] = Array.from(workers.keys()).filter(
+			(tokenId) => !$enabledIcNetworkTokens.some(({ id: icTokenId }) => icTokenId === tokenId)
+		);
+
+		unusedWorkers.forEach((tokenId) => {
+			const worker = workers.get(tokenId);
+
+			if (nonNullish(worker)) {
```
peterpeterparker, line 23: Wouldn't it work with optional chaining too?

```
worker?.stop();
workers.delete(tokenId);
```

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
```diff
@@ -0,0 +1,71 @@
...
+				worker.stop();
+				workers.delete(tokenId);
+			}
+		});
+
+		await Promise.all(
```
peterpeterparker, line 29: `allSettled`?

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
```diff
@@ -0,0 +1,71 @@
...
+	import type { TokenId } from '$lib/types/token';
+	import { debounce, nonNullish } from '@dfinity/utils';
+	import { toastsError } from '$lib/stores/toasts.store';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+
+	let workers: Map<TokenId, WalletWorker> = new Map<TokenId, WalletWorker>();
+
+	const manageWorkers = async () => {
+		const unusedWorkers: TokenId[] = Array.from(workers.keys()).filter(
```
peterpeterparker, line 16: maybe extract this step, cleaning up worker for tokens that are not active anymore, in a dedicated function? For readability and maintanability reasons

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
```diff
@@ -0,0 +1,71 @@
...
+			const worker = workers.get(tokenId);
+
+			if (nonNullish(worker)) {
+				worker.stop();
+				workers.delete(tokenId);
+			}
+		});
+
+		await Promise.all(
+			$enabledIcNetworkTokens.map(async (token) => {
```
peterpeterparker, line 30: for readability maybe extract a function?

```
const loadWorker = async (token) => {...}
await Promise.all($enabledIcNetworkTokens.map(loadWorker))
```

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
```diff
@@ -0,0 +1,71 @@
...
+				if (!workers.has(token.id)) {
```
peterpeterparker, line 31: That's a performance improvement in comparison to existing solution right?

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
```diff
@@ -0,0 +1,71 @@
...
+							? initIcrcWalletWorker(token)
+							: initIcpWalletWorker());
+
+						worker.stop();
+						worker.start();
+
+						workers.set(token.id, worker);
+					} catch (err: unknown) {
+						// We show the error to the user, but we keep loading the rest of the workers
+						toastsError({
```
peterpeterparker, line 43: I'm a bit afraid that users ends with a wall of error toasts. Would it be more user friendly to already think at potential issue and display a single error toast if one or more worker did not load?

src/frontend/src/lib/derived/network-tokens.derived.ts
peterpeterparker: If we ends up using this derived store, maybe provide it in a separate PR?

src/frontend/src/lib/components/core/Listener.svelte
AntonioVentilii-DFINITY: yes, but I still want to test that workflow that deletes unused components

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: No particular reason to be honest. We are doing the same for the balances of ERC20 tokens. so I thought we would want to keep the same pattern. But more than happy to remove it, I see it in the same way you do

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: yes, for sure, changed

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: right! tks!

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: done!

src/frontend/src/lib/derived/network-tokens.derived.ts
AntonioVentilii-DFINITY: Done:

- PR https://github.com/dfinity/oisy-wallet/pull/2213

peterpeterparker:
Ah, one more thing, can you rename the title of this PR? This is not a refactoring, maybe a feature, for sure a fix.

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: done! extracted

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: done, created util function

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: Up until now the event `oisyTriggerWallet` is emitted in a few cases but without any specific token associated.

In case it is emitted and the workers is mounted by `IcListener` (meaning that we are in the token list page), it would have triggered all of the workers, because all the `IcListener` are mounted in that page and each one build a different worker.

The same behavior will happen after this PR: the workers of `IcLoaderWallets` are all mounted and dismounted by the token page.

I didn't want to change any behavior or logic in this PR, but I will revisit the necessity of having the manual trigger in the token list page.

NOTE: I am not affecting the workers that are in the `transaction` page.

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: Not really sure what you mean: in sense of speed to load?

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: Before this PR, they were not treated in the component actually, the user didn't have any feedback at component level.

The idea here was only to make the promises continue without fail, but using `allSettled` should be enough, so I can remove the error if you prefer.

Or do you prefer to show the error either way (in that case yes, I will do a single one)

src/frontend/src/icp/utils/ic-wallet.utils.ts
```diff
@@ -14,3 +23,60 @@ export const waitAndTriggerWallet = async () => {
...
+	tokens
+}: {
+	workers: Map<TokenId, WalletWorker>;
+	tokens: IcToken[];
+}) =>
+	Array.from(workers.keys())
+		.filter((tokenId) => !tokens.some(({ id: icTokenId }) => icTokenId === tokenId))
+		.forEach((tokenId) => {
+			workers.get(tokenId)?.stop();
+			workers.delete(tokenId);
```
peterpeterparker, line 44: I'm not against a more functional programming approach. Deleting entries from a list which is the same list that is used to loop the entries isn't my favorite pattern. This can lead to some unexpected behavior generally speaking.

WDYT?

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
peterpeterparker: We have of course to double check that it isn't called multiple times but if it isn't, then the debounce is here for no reason and should be removed.

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
peterpeterparker: Agree. Maybe something we can improve later.

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
peterpeterparker: Currently the workers are stop / starter without any logic on "if already mounted or not". Therefore felt like an on purpose performance improvement which is a good thing. Just wanted to ask if it what meant on purpose.

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
peterpeterparker: Before this PR, if I'm not wrong, no error was thrown if a worker was not loaded correctly therefore there wasn't a risk of "wall of toasts" or am I missing something?

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: ah understood! yes, it was on purpose: i had tested it debugging and that was indeed called like so. So it makes sense to make a sort of pre-check to stop/start

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: correct! so, if we want to keep the same behavior I can remove the toastError. Shall I remove it?

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: During the session it is not called several times, but at loading time it may happen that it is called a few times consecutively: the store `enabledIcNetworkTokens` (that we use for this) is purely a filter of enabled IC tokens done on store `networkTokens`, that is a filter of all tokens in the selected network.

That means that the function `manageWorkers` is triggered even by other things like ETH/ERC20(/BTC) tokens at loading, when everything is being created.

I checked and the same happens in the ERC20 balance loaders.

I can try to improve the filter-stores in a separate PR

AntonioVentilii-DFINITY: In the equivalent store for ERC20 tokens `enabledErc20NetworkTokens` there is this comment

```typescript
/**
 * It isn't performant to post filter again the Erc20 tokens that are enabled for the specific selected network or no network selected but, it's code wise convenient to avoid duplication of logic.
 */
```

peterpeterparker: Let's add a comment and TODO. Generally speaking, we should avoid adding debounce when it is not absolutely required.  This can lead to some unexpected errors. So if there is a chance to not use a debounce we should try but, let's postpone.

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: romved

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
AntonioVentilii-DFINITY: done

src/frontend/src/icp/utils/ic-wallet.utils.ts
AntonioVentilii-DFINITY: maybe `cleanWorkers` can return a new map with the remaining `workers`?

```typescript
export const cleanWorkers = ({
	workers,
	tokens
}: {
	workers: Map<TokenId, WalletWorker>;
	tokens: IcToken[];
}): Map<TokenId, WalletWorker> =>
	new Map(
		Array.from(workers.entries()).filter(([tokenId, worker]) => {
			const isStillUsed = tokens.some(({ id: icTokenId }) => icTokenId === tokenId);
			if (!isStillUsed) {
				worker.stop();
			}
			return isStillUsed;
		})
	);
```

peterpeterparker: Maybe add a todo for our future self?

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/icp/utils/ic-wallet.utils.ts
AntonioVentilii-DFINITY: done

PR #2211, opened 2024-09-03
chore(frontend): removed unused wstETH logo

AntonioVentilii-DFINITY:
# Motivation

The token `wstETH` already uses the logo that was set in folder `icp-eth/assets`. So we remove the one that was initially used in `eth/assets`.


peterpeterparker (approved):
LGTM, thx

PR #2213, opened 2024-09-03
feat(frontend): enabled IC network tokens

AntonioVentilii-DFINITY:
# Motivation

To load balances for all enabled IC tokens, it will be useful to have a specific derived store, similar to what we have for ERC20 tokens.

Note that we are not using `['icp', 'icrc'].includes(....` because this is type checked, i.e. it would give an error if there was a mistype in the right-hand side of the comparison that does not match one of the possible values of TokenStandard.

### DIFF
--- a/src/frontend/src/lib/derived/network-tokens.derived.ts
+++ b/src/frontend/src/lib/derived/network-tokens.derived.ts
@@ -1,4 +1,5 @@
 import type { Erc20Token } from '$eth/types/erc20';
+import type { IcToken } from '$icp/types/ic';
 import { exchanges } from '$lib/derived/exchange.derived';
 import { pseudoNetworkChainFusion, selectedNetwork } from '$lib/derived/network.derived';
 import { tokens, tokensToPin } from '$lib/derived/tokens.derived';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -32,6 +33,14 @@ export const enabledErc20NetworkTokens: Readable<Erc20Token[]> = derived(
 		$enabledNetworkTokens.filter(({ standard }) => standard === 'erc20') as Erc20Token[]
 );
 
+export const enabledIcNetworkTokens: Readable<IcToken[]> = derived(
```
peterpeterparker, line 36: Can you add a comment on why we have this derived store? We have comments on other stores of this particular module

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/lib/derived/network-tokens.derived.ts
AntonioVentilii-DFINITY: done!

PR #2214, opened 2024-09-03
feat(frontend): Add missing bitcoin in stores

lmuntaner:
# Motivation

The transactions page for Bitcoin is not showing balance, nor logo, nor nice background.

One error is that it doesn't find any token in the pageToken derived store and it's rendering the default ETH transactions component in `TransactionsSignedIn`.

In this PR, I add a new derived store to be used in `TransactionsSignedIn` and fix the pageToken derived store to include bitcoin tokens.

# Changes

* New derived store `networkBitcoin`.
* Add bitcoin tokens to `pageToken` derived store.

# Tests

Tested that the logo appears in the transactions page.


lmuntaner:

This is a subset of #2206

src/frontend/src/lib/derived/network.derived.ts
```diff
@@ -25,6 +25,10 @@ export const networkICP: Readable<boolean> = derived([networkId], ([$networkId])
 	isNetworkIdICP($networkId)
 );
 
+export const networkBitcoin: Readable<boolean> = derived([networkId], ([$networkId]) =>
+	isNetworkIdBitcoin($networkId)
```
peterpeterparker, line 29: maybe should we prevent this to be enabled if `VITE_BITCOIN_MAINNET` is not true for now?

@AntonioVentilii-DFINITY, since you introduced the env variable, wdyt?

AntonioVentilii-DFINITY: theoretically, any network that could be submitted to this function comes from a place that already filters using that flag (for example `enabledBitcoinTokens` or `networks` stores).

No strong opinion to add it here: it would just be another layer of insurance, but I am almost sure it is not needed.

@lmuntaner you decide

AntonioVentilii-DFINITY (approved):
Apart from that small comment about the flag, LGTM!

src/frontend/src/lib/derived/network.derived.ts
peterpeterparker: If already filtered I think we do not have to duplicate the flag. Thanks for the feedback Antonio.

peterpeterparker (approved):

PR #2215, opened 2024-09-03
feat(frontend): util for tokens total USD balance calculation

DenysKarmazynDFINITY:
# Motivation

Step 1 of updating the chain selector with aggregate USD balance values: re-use util for calculation tokens total USD balance. In the next steps, it will be used to calculate this value for different networks (as shown on the screenshot).

Screenshot 2024-09-03 at 13 39 05

### DIFF
--- a/src/frontend/src/lib/components/exchange/ExchangeBalance.svelte
+++ b/src/frontend/src/lib/components/exchange/ExchangeBalance.svelte
@@ -2,10 +2,10 @@
 	import { formatUSD } from '$lib/utils/format.utils';
 	import { exchangeInitialized } from '$lib/derived/exchange.derived';
 	import { combinedDerivedSortedNetworkTokensUi } from '$lib/derived/network-tokens.derived';
-	import { getTokensTotalUsdBalance } from '$lib/utils/tokens.utils';
+	import { sumTokensUsdBalance } from '$lib/utils/tokens.utils';
 
 	let totalUsd: number;
-	$: totalUsd = getTokensTotalUsdBalance($combinedDerivedSortedNetworkTokensUi);
+	$: totalUsd = sumTokensUsdBalance($combinedDerivedSortedNetworkTokensUi);
 </script>
 
 <span class="text-off-white block">
--- a/src/frontend/src/lib/components/exchange/ExchangeBalance.svelte
+++ b/src/frontend/src/lib/components/exchange/ExchangeBalance.svelte
@@ -2,12 +2,10 @@
 	import { formatUSD } from '$lib/utils/format.utils';
 	import { exchangeInitialized } from '$lib/derived/exchange.derived';
 	import { combinedDerivedSortedNetworkTokensUi } from '$lib/derived/network-tokens.derived';
+	import { getTokensTotalUsdBalance } from '$lib/utils/tokens.utils';
 
 	let totalUsd: number;
-	$: totalUsd = $combinedDerivedSortedNetworkTokensUi.reduce(
-		(acc, token) => acc + (token.usdBalance ?? 0),
-		0
-	);
+	$: totalUsd = getTokensTotalUsdBalance($combinedDerivedSortedNetworkTokensUi);
 </script>
 
 <span class="text-off-white block">
--- a/src/frontend/src/lib/utils/tokens.utils.ts
+++ b/src/frontend/src/lib/utils/tokens.utils.ts
@@ -84,8 +84,8 @@ export const pinTokensWithBalanceAtTop = ($tokens: TokenUi[]): TokenUi[] => {
 /**
  * Calculates total USD balance of the provided tokens list.
  *
- * @param $tokens - The list of tokens for total USD balance calculation.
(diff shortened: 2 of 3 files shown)


src/frontend/src/lib/components/exchange/ExchangeBalance.svelte
```diff
@@ -2,12 +2,10 @@
...
 	import { exchangeInitialized } from '$lib/derived/exchange.derived';
 	import { combinedDerivedSortedNetworkTokensUi } from '$lib/derived/network-tokens.derived';
+	import { getTokensTotalUsdBalance } from '$lib/utils/tokens.utils';
 
 	let totalUsd: number;
-	$: totalUsd = $combinedDerivedSortedNetworkTokensUi.reduce(
-		(acc, token) => acc + (token.usdBalance ?? 0),
-		0
-	);
+	$: totalUsd = getTokensTotalUsdBalance($combinedDerivedSortedNetworkTokensUi);
```
peterpeterparker, line 8: it's not really a getter per sé, maybe we can rename it to something like `sumTotalUsdBalance` or better idea?

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -80,3 +80,11 @@ export const pinTokensWithBalanceAtTop = ($tokens: TokenUi[]): TokenUi[] => {
 		...nonPositiveBalances
 	];
 };
+
+/**
+ * Calculates total USD balance of the provided tokens list.
+ *
+ * @param $tokens - The list of tokens for total USD balance calculation.
```
peterpeterparker, line 87: If you add the documentation then maybe also document the return type?

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -80,3 +80,11 @@ export const pinTokensWithBalanceAtTop = ($tokens: TokenUi[]): TokenUi[] => {
...
+ */
+export const getTokensTotalUsdBalance = ($tokens: TokenUi[]): number =>
```
peterpeterparker, line 89: I think @AntonioVentilii-DFINITY uses the `$` pattern to express the fact that the variable is a store. In this case it isn't one so I would remove this prefix but, Antonio can confirm?

src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
```diff
@@ -156,3 +160,16 @@ describe('pinTokensWithBalanceAtTop', () => {
 		]);
 	});
 });
+
+describe('getTokensTotalUsdBalance', () => {
+	it('should correctly calculate tokens USD total balance', () => {
```
peterpeterparker, line 165: Maybe add a test if the previded list is empty?
And another if a token has no usdBalance?

peterpeterparker (commented):
Jira is not public, so not really useful in the content of open source. Could you maybe edit the PR description to remove this reference?

src/frontend/src/lib/components/exchange/ExchangeBalance.svelte
DenysKarmazynDFINITY: Agreed. 
I renamed it to `sumTokensUsdBalance` to highlight that a list of tokens is expected as param and also removed the `Total` part since `sum` already means that there will be a total. WDYT?

src/frontend/src/lib/utils/tokens.utils.ts
DenysKarmazynDFINITY: Done.

src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
DenysKarmazynDFINITY: Sounds good. Added 2 additional test cases (empty list + list with tokens w/o `usdBalance`).

src/frontend/src/lib/utils/tokens.utils.ts
DenysKarmazynDFINITY: True, this helper can possible called for any list of tokens, not necessarily a store value from `network-tokens.derived`. @AntonioVentilii-DFINITY fine with you if I remove `$` here?

src/frontend/src/lib/components/exchange/ExchangeBalance.svelte
peterpeterparker: Coolio, thanks.

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: If the scope is not purely for stores or drives, I agree, you can remove the dollar sign

peterpeterparker (approved):
LGTM, thanks

Already approving, that way you can move on once the `$` removed as discussed.

src/frontend/src/lib/utils/tokens.utils.ts
DenysKarmazynDFINITY: Renamed 👍

PR #2216, opened 2024-09-03
refactor(frontend): move token.services to lib folder

AntonioVentilii-DFINITY:
# Motivation

Since `loadTokenAndRun` is used both by components in `eth`, `icp` and `lib` folder, we move it to the `lib` folder.

Unrelated: we remove unused `runAndResetToken` function.


peterpeterparker (approved):
LGTM, thanks

PR #2217, opened 2024-09-03
refactor(frontend): remove redundant address data type

AntonioVentilii-DFINITY:
# Motivation

`OptionAddressData` and `OptionCertifiedAddressData` are basically the same thing, so we adapt `OptionAddressData` to "fill the shoes" of `OptionCertifiedAddressData`.

### DIFF
--- a/src/frontend/src/lib/stores/address.store.ts
+++ b/src/frontend/src/lib/stores/address.store.ts
@@ -3,19 +3,17 @@ import type { Address, BtcAddress, EthAddress } from '$lib/types/address';
 import type { CertifiedData } from '$lib/types/store';
 import { writable, type Readable } from 'svelte/store';
 
-type AddressData<T extends Address = Address> = CertifiedData<T>;
+type AddressData<T extends Address> = CertifiedData<T>;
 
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/stores/address.store.ts
```diff
@@ -3,19 +3,17 @@ import type { Address, BtcAddress, EthAddress } from '$lib/types/address';
 import type { CertifiedData } from '$lib/types/store';
 import { writable, type Readable } from 'svelte/store';
 
-type AddressData<T extends Address = Address> = CertifiedData<T>;
+type AddressData<T extends Address> = CertifiedData<T>;
```
peterpeterparker, line 6: Verbose but, should this be actually renamed tp `CertifiedAddressData` and the other option as `StorageAddressData`?

Somehow I'm a bit confuse about which types is what. WDYT?

AntonioVentilii-DFINITY: yes, make sense. doing it in this PR

peterpeterparker (approved):
LGTM, thx

PR #2218, opened 2024-09-03
refactor(frontend): rename ETH address store

AntonioVentilii-DFINITY:
# Motivation

We rename the derived store specific for ETH address not loaded.


peterpeterparker (approved):
LGTM, thanks

NTH: it's not a "rename" per sé, not sure exactly what it is in a single word though

AntonioVentilii-DFINITY:
theoretically "refactor" is the right word eheh

PR #2219, opened 2024-09-03
feat(frontend): UI for chain selector with USD values

DenysKarmazynDFINITY:
# Motivation

Step 1 of updating the chain selector with aggregate USD balance values: implement UI component to display text, description and logo according to the provided design. Next step is to calculate and pass balance values to that component for all non-test chains.

Note: potentially, we could extend `TextWithLogo` to accept `description`, but I decided against it in order to avoid even more conditional things inside that component. Please let me know if you think otherwise.

Browser:
web

Figma:
Figma


src/frontend/src/lib/components/ui/TextWithDescriptionAndLogo.svelte
```diff
@@ -0,0 +1,19 @@
+<script lang="ts">
```
peterpeterparker, line 1: I'm not a fund of component's code duplication I have to say. Wouldn't it be cleaner to extend `TextWithLogo` either by:

- Adding the property `description` directly within that component

or

- Adding a `slot` in `TextWithLogo` and then using this component in `TextWithDescriptionAndLogo`

WDYT?

src/frontend/src/lib/components/ui/TextWithDescriptionAndLogo.svelte
```diff
@@ -0,0 +1,19 @@
...
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+	import { i18n } from '$lib/stores/i18n.store';
+	import Logo from '$lib/components/ui/Logo.svelte';
+
+	export let name: string;
+	export let description: string | undefined;
```
peterpeterparker, line 7: You can assign a default value `undefined` that way you are not forced to use the property in consumers.

```
// before

<NetworkButton {id} {name} {icon} on:icSelected description={undefined} />

// after


<NetworkButton {id} {name} {icon} on:icSelected />
```

src/frontend/src/lib/components/ui/TextWithDescriptionAndLogo.svelte
DenysKarmazynDFINITY: Yes, in the end re-using `TextWithLogo` doesn't end up in a lot of `if`s as I'm initially was afraid of, just a few adjustments and now this component can accept `description` as a prop. Thanks for the idea 👍

src/frontend/src/lib/components/ui/TextWithDescriptionAndLogo.svelte
DenysKarmazynDFINITY: Done 👍

src/frontend/src/lib/components/networks/NetworkButton.svelte
```diff
@@ -26,8 +27,8 @@
 	};
 </script>
 
-<button class="w-full flex justify-between items-center" on:click={onClick}>
-	<TextWithLogo {name} {icon} logo="start" />
+<button class="w-full flex justify-between items-start" on:click={onClick}>
```
peterpeterparker, line 30: `items-start` is maybe the default? Would removing this class as the same result?

src/frontend/src/lib/components/ui/TextWithLogo.svelte
```diff
@@ -6,14 +6,23 @@
...
 	class:flex-row-reverse={logo === 'start'}
 >
-	<span>{name}</span>
+	{#if description}
+		<span class="flex flex-col gap-0.5">
+			<span class="leading-5">{name}</span>
+			<span class="leading-none text-xs text-misty-rose text-left">{description}</span>
+		</span>
+	{:else}
+		<span>{name}</span>
```
peterpeterparker, line 25: To avoid duplicating the usage of `{name}`, what do you think of following?

```
<script>
  let withDescription = false;
  $: withDescription = nonNullish(description);
</script>

<span class="flex flex-col gap-0.5">
   <span class:leading-5={withDescription}>{name}</span>
   
   {#if withDescription}
      <span class="leading-none text-xs text-misty-rose text-left">{description}</span>
		</span>
   {/else}
```

peterpeterparker: Alternatively, can potentially `leading-5` be use for `name` without `description` as well? If so:

```
<span class="flex flex-col gap-0.5">
   <span class="leading-5">{name}</span>
   
   {#if nonNullish(description)}
      <span class="leading-none text-xs text-misty-rose text-left">{description}</span>
		</span>
   {/else}
```

src/frontend/src/lib/components/networks/NetworkButton.svelte
DenysKarmazynDFINITY: Yes, forgot to remove this one. It's done now.

src/frontend/src/lib/components/ui/TextWithLogo.svelte
DenysKarmazynDFINITY: I used your second suggestion, and code-wise the component looks much neater now. Having `leading-5` for `name` affected by a few pixels the `TextWithLogo` component appearance, but it's barely noticeable. 

Local:
local

Staging:
staging

peterpeterparker (approved):
LGTM, thx

PR #2220, opened 2024-09-03
fix(frontend): use dynamical type for certify address function

AntonioVentilii-DFINITY:
# Motivation

To be more coherent, function `certifyAddress` should accept a dynamic type that extends `Address`.


peterpeterparker (approved):
LGTM, thx

PR #2222, opened 2024-09-03
refactor(frontend): extract validateAddress util

AntonioVentilii-DFINITY:
# Motivation

We extract the util function that checks and validate a specific address. It will be re-used for BTC address.


peterpeterparker:
I know it's a draft but, if refactored, note that it should not be a util but rather a service.

peterpeterparker (approved):
LGTM, thanks

PR #2226, opened 2024-09-04
feat(frontend): add BTC verification to AddressGuard

AntonioVentilii-DFINITY:
# Motivation

We include the BTC address verification in the `AddressGuard` component.

### DIFF
--- a/src/frontend/src/lib/services/address.services.ts
+++ b/src/frontend/src/lib/services/address.services.ts
@@ -302,6 +302,12 @@ const validateAddress = async <T extends Address>({
 	await warnSignOut(err ?? 'Error while certifying your address');
 };
 
+export const validateBtcAddressMainnet = async ($addressStore: StorageAddressData<BtcAddress>) =>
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/services/address.services.ts
```diff
@@ -302,6 +302,12 @@ const validateAddress = async <T extends Address>({
 	await warnSignOut(err ?? 'Error while certifying your address');
```
peterpeterparker, line 302: Do you think we should pay a special attention to the fact that potentially both functions fail and call `warnSignOut`? I guess the first win given that sign-out reload the screen but, just wanted to know what you think.

peterpeterparker (approved):
LGTM, thanks

Add a question indirectly related to the PR

src/frontend/src/lib/services/address.services.ts
peterpeterparker: Glad to hear you share the same understanding. Let's keep it as it then, sounds good!

PR #2227, opened 2024-09-04
feat(frontend): track analytics when IC send completed

peterpeterparker:
# Motivation

There is a 2s timeout after an IC send transactions is complemented to optimistically fetch the transactions. Given that we are interested to get to know how long does a send least, it's not really correct to count that timeout within the overall send. Therefore we can track the anonymous metrics when the send is over.

# Changes

- Pass the analytics as callback for `sendIc` and collect analytics when send is done.

### DIFF
--- a/src/frontend/src/icp/components/send/IcSendTokenWizard.svelte
+++ b/src/frontend/src/icp/components/send/IcSendTokenWizard.svelte
@@ -131,27 +131,30 @@
 				ckErc20ToErc20MaxCkEthFees
 			};
 
+			const trackAnalyticsOnSendComplete = async () => {
+				await Promise.allSettled([
+					trackTimedEventSuccess(timedEvent),
+					trackEvent({
+						name: isNetworkIdBitcoin(networkId)
+							? TRACK_COUNT_CONVERT_CKBTC_TO_BTC_SUCCESS
+							: isConvertCkEthToEth({ token: $token, networkId })
+								? TRACK_COUNT_CONVERT_CKETH_TO_ETH_SUCCESS
+								: isConvertCkErc20ToErc20({ token: $token, networkId })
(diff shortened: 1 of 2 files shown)


peterpeterparker:
cc @DSharifi

src/frontend/src/icp/components/send/IcSendTokenWizard.svelte
```diff
@@ -131,28 +131,31 @@
 				ckErc20ToErc20MaxCkEthFees
 			};
 
+			const trackAnalyticsOnSendComplete = async () => {
```
AntonioVentilii-DFINITY, line 134: didn't Lint asked you to remove the `{}` parenthesis? strange...

AntonioVentilii-DFINITY (approved):
LGTM!

src/frontend/src/icp/components/send/IcSendTokenWizard.svelte
peterpeterparker: No it didn't. I added those because in the library I got a different setup. We should maybe align the rules.

Which one is the particular eslint rule? That way I can copy it?

peterpeterparker: Ah it's probably because it's in a Svelte component script tag?

peterpeterparker: Anyway I need the `{}` because otherwise the function does not return `Promise<void>` but the result of `allSettled` here.

AntonioVentilii-DFINITY: all good! no worries! it is indeed because you are not returning it, it is void, so the linter worked. [The rule](https://eslint.org/docs/latest/rules/arrow-body-style#as-needed) is

	`'arrow-body-style': ['warn', 'as-needed']`

PR #2229, opened 2024-09-04
feat(lint): add rule prefer-template

AntonioVentilii-DFINITY:
# Motivation

We include the rule [prefer-template](https://eslint.org/docs/latest/rules/prefer-template) to avoid string concatenations,  and we fix issues found.


peterpeterparker (approved):
Cool, thanks

PR #2230, opened 2024-09-04
feat(ci): add release template

AntonioVentilii-DFINITY:
# Motivation

We would like to standardize the release notes.

### DIFF
--- a/.github/release_template.md
+++ b/.github/release_template.md
@@ -1,0 +1,45 @@
+# Frontend
+
+## Features
+
+- The list of new Frontend features/enhancements/fixes goes here.
+
+## Fixes
+
+- The list of bugs fixed on Frontend goes here.
+
+## Style
+
+- Adjustments in the UI style or layout.
+
+## Refactoring
+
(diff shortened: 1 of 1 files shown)


peterpeterparker:
Wait, does that work? I've been looking for years for such a solution and always referred to this SO thread which says there is no such feature https://stackoverflow.com/questions/56798253/release-template-for-github.

AntonioVentilii-DFINITY:
@peterpeterparker this PR would be just to structure the notes. My plan is to have it in a workflow/action that is triggered for new releases

bitdivine:
Interesting link, @peterpeterparker .  The GitHub functionality it references is documented here: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes

So it looks as if, if we were to label PRs, we could automatically populate release notes.  E.g. we could make a section called `# Frontend` and populate it with commits tagged with frontend.  I have a lot of questions about how this works but it looks like something worth exploring.  At some time.  When we have time. :laughing:

PR #2231, opened 2024-09-04
feat(lint): add lint rule to sort svelte imports

AntonioVentilii-DFINITY:
# Motivation

It seems like a good point to start: we would like to introduce the [sorting import rule](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/order.md) for svelte component. However, done it drastically for the whole project requires a few hundreds of fixing. So, we start doing it folder by folder.

Gradually, we will include the entire project, at which point we move the rule in the main `rules` prop.


peterpeterparker (commented):
Does organize auto import sort those import? i.e. does `npm run format` sorts the import in Svelte cmp?

AntonioVentilii-DFINITY:
yes, however, just for the first time we are running, it must be ran a few times, because it does not fix all the rules in loop:

if the first fix changes some imports to merge for example, it will not sort them, and vice versa; so for now it needs to run a few times, until it says that it is all fixed.

but ultimately, yes, `npm run format` solves this rule.

peterpeterparker (approved):
LGTM, thx

PR #2232, opened 2024-09-04
refactor(frontend): move invalidIcpAddress to lib folder

AntonioVentilii-DFINITY:
# Motivation

Since it is being used both by ETH and ICP components/functions, we move `invalidIcpAddress` to folder `lib`. This is the same practice that other functions in the same module have: for example `isEthAddress` is used both by ETH and ICP modules.

As side effect of this, we have one less possible loop dependency inside `icp` folder.


peterpeterparker (approved):
LGTM, thanks

PR #2233, opened 2024-09-04
feat(frontend): update network-tokens to enabled tokens balance

DenysKarmazynDFINITY:
# Motivation

Step 3 of updating the chain selector with aggregate USD balance values: adjust `network-tokens.derived` to load enabled tokens balance regardless if a network selected by user or not. We discussed with @AntonioVentilii-DFINITY the performance aspect of this change, and since it will be basically the amount of data loaded (and hence time needed to load it) as if we were in Chain Fusion, it should not affect the user experience and performance.

### DIFF
--- a/src/frontend/src/icp/components/core/IcLoaderWallets.svelte
+++ b/src/frontend/src/icp/components/core/IcLoaderWallets.svelte
@@ -4,25 +4,25 @@
 	import type { IcToken } from '$icp/types/ic';
 	import type { WalletWorker } from '$icp/types/ic-listener';
 	import { cleanWorkers, loadWorker } from '$icp/utils/ic-wallet.utils';
-	import { enabledIcNetworkTokens } from '$lib/derived/network-tokens.derived';
+	import { enabledIcTokens } from '$lib/derived/tokens.derived';
 	import type { TokenId } from '$lib/types/token';
 
 	const workers: Map<TokenId, WalletWorker> = new Map<TokenId, WalletWorker>();
 
 	const manageWorkers = async () => {
-		cleanWorkers({ workers, tokens: $enabledIcNetworkTokens });
+		cleanWorkers({ workers, tokens: $enabledIcTokens });
 
 		await Promise.allSettled(
-			$enabledIcNetworkTokens.map(async (token: IcToken) => await loadWorker({ workers, token }))
+			$enabledIcTokens.map(async (token: IcToken) => await loadWorker({ workers, token }))
 		);
 	};
 
 	const debounceManageWorkers = debounce(manageWorkers, 500);
 
 	// TODO: here we debounce the manageWorkers function to avoid multiple calls in a short period
-	//  of time due to the several dependencies of enabledIcNetworkTokens, that are not strictly only IC tokens.
+	//  of time due to the several dependencies of enabledIcTokens, that are not strictly only IC tokens.
 	//  This is a temporary solution, and we should find a better way to handle this, improving the store.
-	$: $enabledIcNetworkTokens, debounceManageWorkers();
+	$: $enabledIcTokens, debounceManageWorkers();
 
 	onDestroy(() => {
 		workers.forEach((worker) => worker.stop());
--- a/src/frontend/src/lib/derived/network-tokens.derived.ts
+++ b/src/frontend/src/lib/derived/network-tokens.derived.ts
@@ -7,40 +7,32 @@ import { balancesStore } from '$lib/stores/balances.store';
 import type { Token, TokenUi } from '$lib/types/token';
 import { filterTokensForSelectedNetwork } from '$lib/utils/network.utils';
 import {
-	filterEnabledNetworkTokens,
+	filterEnabledTokens,
 	pinTokensWithBalanceAtTop,
 	sortTokens
 } from '$lib/utils/tokens.utils';
 import { derived, type Readable } from 'svelte/store';
 
 /**
+ * All enabled by user tokens.
+ */
+const enabledTokens: Readable<Token[]> = derived([tokens], filterEnabledTokens);
+
+/**
  * All tokens matching the selected network or chain fusion, regardless if they are enabled by the user or not.
  */
-const selectedNetworkTokens: Readable<Token[]> = derived(
-	[tokens, selectedNetwork, pseudoNetworkChainFusion],
+const enabledNetworkTokens: Readable<Token[]> = derived(
+	[enabledTokens, selectedNetwork, pseudoNetworkChainFusion],
 	filterTokensForSelectedNetwork
-);
-
-/**
- * All enabled network tokens, regardless if they are selected by the user or not.
- */
-const enabledNetworkTokens: Readable<Token[]> = derived([tokens], filterEnabledNetworkTokens);
-
-/**
- * All enabled network tokens that matching the selected network or chain fusion and are enabled by user
- */
-const enabledSelectedNetworkTokens: Readable<Token[]> = derived(
-	[selectedNetworkTokens],
-	filterEnabledNetworkTokens
 );
 
 /**
  * It isn't performant to post filter again the Erc20 tokens that are enabled but, it's code wise convenient to avoid duplication of logic.
  */
 export const enabledErc20NetworkTokens: Readable<Erc20Token[]> = derived(
-	[enabledNetworkTokens],
-	([$enabledNetworkTokens]) =>
-		$enabledNetworkTokens.filter(({ standard }) => standard === 'erc20') as Erc20Token[]
+	[enabledTokens],
+	([$enabledTokens]) =>
+		$enabledTokens.filter(({ standard }) => standard === 'erc20') as Erc20Token[]
 );
 
 /**
@@ -49,18 +41,16 @@ export const enabledErc20NetworkTokens: Readable<Erc20Token[]> = derived(
 // TODO: The several dependencies of enabledIcNetworkTokens are not strictly only IC tokens, but other tokens too.
 //  We should find a better way to handle this, improving the store.
 export const enabledIcNetworkTokens: Readable<IcToken[]> = derived(
-	[enabledNetworkTokens],
-	([$enabledNetworkTokens]) =>
-		$enabledNetworkTokens.filter(
-			({ standard }) => standard === 'icp' || standard === 'icrc'
-		) as IcToken[]
+	[enabledTokens],
+	([$enabledTokens]) =>
+		$enabledTokens.filter(({ standard }) => standard === 'icp' || standard === 'icrc') as IcToken[]
 );
 
 /**
  * Network tokens sorted by market cap, with the ones to pin at the top of the list.
  */
 export const combinedDerivedSortedNetworkTokens: Readable<Token[]> = derived(
-	[enabledSelectedNetworkTokens, tokensToPin, exchanges],
+	[enabledNetworkTokens, tokensToPin, exchanges],
 	([$tokens, $tokensToPin, $exchanges]) => sortTokens({ $tokens, $exchanges, $tokensToPin })
 );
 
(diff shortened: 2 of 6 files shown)


src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -6,23 +6,36 @@ import { tokens, tokensToPin } from '$lib/derived/tokens.derived';
...
+	pinTokensWithBalanceAtTop,
+	sortTokens
+} from '$lib/utils/tokens.utils';
 import { derived, type Readable } from 'svelte/store';
 
 /**
  * All tokens matching the selected network or chain fusion, regardless if they are enabled by the user or not.
  */
-const networkTokens: Readable<Token[]> = derived(
+const selectedNetworkTokens: Readable<Token[]> = derived(
```
AntonioVentilii-DFINITY, line 19: may I ask, why renaming it? I think the old name still fits the scope, no? (and less verbose?)

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -125,3 +125,12 @@ export const pinTokensWithBalanceAtTop = ({
...
 export const sumTokensUsdBalance = (tokens: TokenUi[]): number =>
 	tokens.reduce((acc, token) => acc + (token.usdBalance ?? 0), 0);
+
+/**
+ * Filters and returns a list of "enabled" by user network tokens
+ *
+ * @param $tokens - The list of network tokens.
+ * @returns The list of "enabled" network tokens.
+ */
+export const filterEnabledNetworkTokens = ([$tokens]: [$tokens: Token[]]): Token[] =>
```
AntonioVentilii-DFINITY, line 135: i would rename it just `filterEnabledTokens`

src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -6,23 +6,36 @@ import { tokens, tokensToPin } from '$lib/derived/tokens.derived';
...
 	[tokens, selectedNetwork, pseudoNetworkChainFusion],
 	filterTokensForSelectedNetwork
 );
 
-const enabledNetworkTokens: Readable<Token[]> = derived([networkTokens], ([$networkTokens]) =>
-	$networkTokens.filter((token) => ('enabled' in token ? token.enabled : true))
+/**
+ * All enabled network tokens, regardless if they are selected by the user or not.
+ */
+const enabledNetworkTokens: Readable<Token[]> = derived([tokens], filterEnabledNetworkTokens);
```
AntonioVentilii-DFINITY, line 27: i would rename this `enabledTokens`

src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -6,23 +6,36 @@ import { tokens, tokensToPin } from '$lib/derived/tokens.derived';
...
+
+/**
+ * All enabled network tokens that matching the selected network or chain fusion and are enabled by user
+ */
+const enabledSelectedNetworkTokens: Readable<Token[]> = derived(
```
AntonioVentilii-DFINITY, line 32: and this would be the old `enabledNetworkTokens` still

src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -6,23 +6,36 @@ import { tokens, tokensToPin } from '$lib/derived/tokens.derived';
...
+	[selectedNetworkTokens],
+	filterEnabledNetworkTokens
 );
 
 /**
- * It isn't performant to post filter again the Erc20 tokens that are enabled for the specific selected network or no network selected but, it's code wise convenient to avoid duplication of logic.
+ * It isn't performant to post filter again the Erc20 tokens that are enabled but, it's code wise convenient to avoid duplication of logic.
  */
 export const enabledErc20NetworkTokens: Readable<Erc20Token[]> = derived(
 	[enabledNetworkTokens],
```
AntonioVentilii-DFINITY, line 41: then here and below I would use the `enabledTokens`

src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -6,28 +6,33 @@ import { tokens, tokensToPin } from '$lib/derived/tokens.derived';
...
 	filterTokensForSelectedNetwork
 );
 
-const enabledNetworkTokens: Readable<Token[]> = derived([networkTokens], ([$networkTokens]) =>
-	$networkTokens.filter((token) => ('enabled' in token ? token.enabled : true))
-);
-
 /**
- * It isn't performant to post filter again the Erc20 tokens that are enabled for the specific selected network or no network selected but, it's code wise convenient to avoid duplication of logic.
+ * It isn't performant to post filter again the Erc20 tokens that are enabled but, it's code wise convenient to avoid duplication of logic.
```
AntonioVentilii-DFINITY, line 30:
```suggestion
 * It isn't performant to post filter again the Erc20 tokens that are enabled but it's code wise convenient to avoid duplication of logic.
```

AntonioVentilii-DFINITY (approved):
LGTM! Tks!

(small suggestion in a docstring)

src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -6,28 +6,33 @@ import { tokens, tokensToPin } from '$lib/derived/tokens.derived';
...
+} from '$lib/utils/tokens.utils';
 import { derived, type Readable } from 'svelte/store';
 
+/**
+ * All enabled by user tokens.
+ */
+const enabledTokens: Readable<Token[]> = derived([tokens], filterEnabledTokens);
+
 /**
  * All tokens matching the selected network or chain fusion, regardless if they are enabled by the user or not.
```
peterpeterparker, line 22: This description isn't accurate anymore

src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -6,28 +6,33 @@ import { tokens, tokensToPin } from '$lib/derived/tokens.derived';
...
+	filterEnabledTokens,
+	pinTokensWithBalanceAtTop,
+	sortTokens
+} from '$lib/utils/tokens.utils';
 import { derived, type Readable } from 'svelte/store';
 
+/**
+ * All enabled by user tokens.
+ */
+const enabledTokens: Readable<Token[]> = derived([tokens], filterEnabledTokens);
```
peterpeterparker, line 19: Should this derived store find place in related `tokens.derived` module?

src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -6,28 +6,33 @@ import { tokens, tokensToPin } from '$lib/derived/tokens.derived';
...
-
 /**
- * It isn't performant to post filter again the Erc20 tokens that are enabled for the specific selected network or no network selected but, it's code wise convenient to avoid duplication of logic.
+ * It isn't performant to post filter again the Erc20 tokens that are enabled but, it's code wise convenient to avoid duplication of logic.
  */
 export const enabledErc20NetworkTokens: Readable<Erc20Token[]> = derived(
-	[enabledNetworkTokens],
-	([$enabledNetworkTokens]) =>
-		$enabledNetworkTokens.filter(({ standard }) => standard === 'erc20') as Erc20Token[]
+	[enabledTokens],
```
peterpeterparker, line 33: Is this correct and expected? Shouldn't this be renamed and shouldn't it be moved to the `tokens.derived` module?

src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -36,11 +41,9 @@ export const enabledErc20NetworkTokens: Readable<Erc20Token[]> = derived(
 // TODO: The several dependencies of enabledIcNetworkTokens are not strictly only IC tokens, but other tokens too.
 //  We should find a better way to handle this, improving the store.
 export const enabledIcNetworkTokens: Readable<IcToken[]> = derived(
```
peterpeterparker, line 43: Should this derived store maybe be removed, it isn't used anymore?

If yes, not in this PR.

src/frontend/src/lib/derived/network-tokens.derived.ts
DenysKarmazynDFINITY: The change is expected - this derived store is used to load balances in `LoaderBalances`, and the idea is that now we fetch all enabled tokens balances (as if Chain Fusion were selected in `NetworksSwitcher`).

And yes, It makes sense to move it to `tokens.derived`. If that's okay, I would do it in a separate PR, as I would also move the IC tokens store, as well as rename both: `enabledErc20NetworkTokens` -> `enabledErc20Tokens`, `enabledIcNetworkTokens` -> `enabledIcTokens`.

src/frontend/src/lib/derived/network-tokens.derived.ts
DenysKarmazynDFINITY: After I move all the things we discussed above, `network-tokens.derived` will have 3 stores left: `enabledIcNetworkTokens` (internal, not exported), `combinedDerivedSortedNetworkTokens` (exported) and `combinedDerivedSortedNetworkTokensUi` (exported). Those ones are filtered with `filterTokensForSelectedNetwork` so we can potentially keep them in that file, unless you think that even with that filtering they belong to `tokens.derived`.

src/frontend/src/lib/derived/network-tokens.derived.ts
DenysKarmazynDFINITY: Good point, moved.

src/frontend/src/lib/derived/network-tokens.derived.ts
DenysKarmazynDFINITY: Updated.

src/frontend/src/lib/derived/network-tokens.derived.ts
DenysKarmazynDFINITY: Updated.

peterpeterparker (commented):
LGTM, thanks.

Before approving I would be happy if @AntonioVentilii-DFINITY can have another look given that we implement quite some changes after his review and also because I think we should be super precaucious given our recent history when we modify those core derived stores.

src/frontend/src/icp/components/core/IcLoaderWallets.svelte
```diff
@@ -22,7 +22,7 @@
 	// TODO: here we debounce the manageWorkers function to avoid multiple calls in a short period
 	//  of time due to the several dependencies of enabledIcNetworkTokens, that are not strictly only IC tokens.
```
AntonioVentilii-DFINITY, line 23:
```suggestion
	//  of time due to the several dependencies of enabledIcTokens, that are not strictly only IC tokens.
```

src/frontend/src/lib/derived/tokens.derived.ts
```diff
@@ -27,3 +30,28 @@ export const tokensToPin: Readable<TokenToPin[]> = derived(
...
+export const enabledErc20Tokens: Readable<Erc20Token[]> = derived(
+	[enabledTokens],
+	([$enabledTokens]) =>
+		$enabledTokens.filter(({ standard }) => standard === 'erc20') as Erc20Token[]
+);
+
+/**
+ * The following store is use as reference for the list of WalletWorkers that are started/stopped in the main token page.
+ */
+// TODO: The several dependencies of enabledIcNetworkTokens are not strictly only IC tokens, but other tokens too.
```
AntonioVentilii-DFINITY, line 51:
```suggestion
// TODO: The several dependencies of enabledIcTokens are not strictly only IC tokens, but other tokens too.
```

AntonioVentilii-DFINITY (approved):
Changes LGTM! Tks! (small changes in docstrings)

Tks @peterpeterparker for the heads-up! I tested the branch checking the calls for balances (that is the affected part of the flow) during the session for the balances and they are called correctly: ONLY on loading the tokens AND on enabling/disabling tokens

Screenshot 2024-09-05 at 12 29 07

peterpeterparker (approved):

PR #2238, opened 2024-09-05
refactor(frontend): Remove token global store in Hero

lmuntaner:
# Motivation

There is a global store `token` that is deprecated.

In this PR, I remove the usage of this store in the `HeroContent` code path starting from the layout.

# Changes

* Add a prop `token` to `Hero` and `HeroContent` and use it instead of relying in the global store `token`.

# Tests

I tested that everything is still working as expected.


peterpeterparker (commented):
The `$token` is indeed deprecated, but we have not yet decided what and how we will ultimately replace it fully. Therefore, I would keep the current code as is for now.

lmuntaner:
Ok, closing it then. I

> The `$token` is indeed deprecated, but we have not yet decided what and how we will ultimately replace it fully. Therefore, I would keep the current code as is for now.

Ah, ok, my bad. I'll close this PR then.

peterpeterparker:
> Ah, ok, my bad. I'll close this PR then.

Well our bad, it's not like it's explicitely documented...

PR #2240, opened 2024-09-05
chore(prettier): add "prettier-plugin-tailwindcss"

DenysKarmazynDFINITY:
# Motivation

I'm not sure if this has already been decided against, but I would like to suggest using `prettier-plugin-tailwindcss` to keep Tailwind class names organized in a consistent order. The plugin also removes duplicate classes, unexpected whitespace, etc. You can find the detailed documentation [here](https://github.com/tailwindlabs/prettier-plugin-tailwindcss).

I've prettified the `Card` component as an example of what the plugin does.

If we decide to use it, I can apply it to all components. My only concern is that most of the pending PRs with UI changes might get merge conflicts.

@peterpeterparker @AntonioVentilii-DFINITY WDYT?

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -3,6 +3,6 @@
 	"singleQuote": true,
 	"trailingComma": "none",
 	"printWidth": 100,
-	"plugins": ["prettier-plugin-svelte", "prettier-plugin-organize-imports"],
+	"plugins": ["prettier-plugin-svelte", "prettier-plugin-organize-imports", "prettier-plugin-tailwindcss"],
 	"overrides": [{ "files": "*.svelte", "options": { "parser": "svelte" } }]
 }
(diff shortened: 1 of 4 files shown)


AntonioVentilii-DFINITY:
Two questions:

1. If you change all the components, how many of those are to be corrected? I imagine a good amount
2. I am doing something similar with sorting component imports, but I am doing folder-by-folder, to avoid possible conflicts with PRs, or simply to control the changes better. Maybe you can think of something similar?

DenysKarmazynDFINITY:
> I like the suggestion! Two questions:
> 
> 1. If you change all the components, how many of those are to be corrected? I imagine a good amount
> 2. I am doing something similar with sorting component imports, but I am doing folder-by-folder, to avoid possible conflicts with PRs, or simply to control the changes better. Maybe you can think of something similar?

1. We can assume the worst-case scenario - all Svelte components with more than 2 Tailwind classes are gonna be updated.
2. I think your strategy is pretty much applicable to this plugin, too. Therefore, we can also apply it folder-by-folder.

DenysKarmazynDFINITY:
The plugin usage will be gradually integrated into the code base (separate PRs). Closing this one.

PR #2244, opened 2024-09-05
build(ci): fix node version to 20

AntonioVentilii-DFINITY:
# Motivation

We fix the Node version to `LTS` (i.e. `^20.x`), to avoid possible incompatibility issues when running the scripts.

For example, with Node ^22, `npm run i18n` gives an issue possibly related to [this proposal](https://github.com/tc39/proposal-import-attributes?tab=readme-ov-file#synopsis).


peterpeterparker (approved):
LGTM, thx

PR #2245, opened 2024-09-05
refactor(frontend): calculate --border-radius-xs

AntonioVentilii-DFINITY:
# Motivation

Just for coherence, we calculate `--border-radius-xs` CSS variable from `--border-radius-sm`.


peterpeterparker (approved):
LGTM, thx

PR #2247, opened 2024-09-05
feat(btc): add Btc Regtest network and token

DenysKarmazynDFINITY:
# Motivation

The goal is to create a new test network and token (`BTC (Regtest` / `Bitcoin (Regtest)`) that are gonna be used for local BTC development.

### DIFF
--- a/src/frontend/src/btc/derived/networks.derived.ts
+++ b/src/frontend/src/btc/derived/networks.derived.ts
@@ -1,5 +1,6 @@
 import { BTC_MAINNET_ENABLED, NETWORK_BITCOIN_ENABLED } from '$env/networks.btc.env';
 import { BTC_MAINNET_NETWORK, BTC_REGTEST_NETWORK, BTC_TESTNET_NETWORK } from '$env/networks.env';
+import { LOCAL } from '$lib/constants/app.constants';
 import { testnets } from '$lib/derived/testnets.derived';
 import type { Network } from '$lib/types/network';
 import { derived, type Readable } from 'svelte/store';
@@ -8,7 +9,7 @@ export const enabledBitcoinNetworks: Readable<Network[]> = derived([testnets], (
 	NETWORK_BITCOIN_ENABLED
 		? [
 				...(BTC_MAINNET_ENABLED ? [BTC_MAINNET_NETWORK] : []),
-				...($testnets ? [BTC_TESTNET_NETWORK, BTC_REGTEST_NETWORK] : [])
+				...($testnets ? [BTC_TESTNET_NETWORK, ...(LOCAL ? [BTC_REGTEST_NETWORK] : [])] : [])
 			]
 		: []
 );
--- a/src/frontend/src/btc/derived/networks.derived.ts
+++ b/src/frontend/src/btc/derived/networks.derived.ts
@@ -1,5 +1,5 @@
 import { BTC_MAINNET_ENABLED, NETWORK_BITCOIN_ENABLED } from '$env/networks.btc.env';
-import { BTC_MAINNET_NETWORK, BTC_TESTNET_NETWORK } from '$env/networks.env';
+import { BTC_MAINNET_NETWORK, BTC_REGTEST_NETWORK, BTC_TESTNET_NETWORK } from '$env/networks.env';
 import { testnets } from '$lib/derived/testnets.derived';
 import type { Network } from '$lib/types/network';
 import { derived, type Readable } from 'svelte/store';
@@ -8,7 +8,7 @@ export const enabledBitcoinNetworks: Readable<Network[]> = derived([testnets], (
 	NETWORK_BITCOIN_ENABLED
 		? [
 				...(BTC_MAINNET_ENABLED ? [BTC_MAINNET_NETWORK] : []),
-				...($testnets ? [BTC_TESTNET_NETWORK] : [])
+				...($testnets ? [BTC_TESTNET_NETWORK, BTC_REGTEST_NETWORK] : [])
 			]
 		: []
 );
--- a/src/frontend/src/btc/derived/tokens.derived.ts
+++ b/src/frontend/src/btc/derived/tokens.derived.ts
@@ -1,5 +1,6 @@
 import { BTC_MAINNET_ENABLED, NETWORK_BITCOIN_ENABLED } from '$env/networks.btc.env';
 import { BTC_MAINNET_TOKEN, BTC_REGTEST_TOKEN, BTC_TESTNET_TOKEN } from '$env/tokens.btc.env';
+import { LOCAL } from '$lib/constants/app.constants';
 import { testnets } from '$lib/derived/testnets.derived';
 import type { Token } from '$lib/types/token';
 import { derived, type Readable } from 'svelte/store';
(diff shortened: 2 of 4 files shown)


src/frontend/src/env/tokens.btc.env.ts
```diff
@@ -15,7 +15,7 @@ export const BTC_MAINNET_TOKEN: Token = {
 	standard: 'bitcoin',
 	category: 'default',
 	name: 'Bitcoin',
-	symbol: 'BTC',
+	symbol: BTC_MAINNET_SYMBOL,
```
DenysKarmazynDFINITY, line 18: While creating a new token, I noticed that `BTC_MAINNET_TOKEN` doesn't use `BTC_MAINNET_SYMBOL` as a value for `symbol`. Same with `BTC_TESTNET_TOKEN` below. I update it, but please let me know if there was a specific reason for that.

src/frontend/src/btc/derived/tokens.derived.ts
```diff
@@ -8,7 +8,7 @@ export const enabledBitcoinTokens: Readable<Token[]> = derived([testnets], ([$te
 	NETWORK_BITCOIN_ENABLED
 		? [
 				...(BTC_MAINNET_ENABLED ? [BTC_MAINNET_TOKEN] : []),
-				...($testnets ? [BTC_TESTNET_TOKEN] : [])
+				...($testnets ? [BTC_TESTNET_TOKEN, BTC_REGSTEST_TOKEN] : [])
```
AntonioVentilii-DFINITY, line 11: Shoudn't Regtest be enabled only for devs (only if running env is LOCAL)? Or the idea is that any user can decide to run the Regtest node locally if they want?

src/frontend/src/env/tokens.btc.env.ts
```diff
@@ -30,7 +30,22 @@ export const BTC_TESTNET_TOKEN: Token = {
...
 	name: 'Bitcoin (Testnet)',
-	symbol: 'BTC (Testnet)',
+	symbol: BTC_TESTNET_SYMBOL,
+	decimals: BTC_DECIMALS,
+	icon: bitcoinTestnet
+};
+
+export const BTC_REGSTEST_SYMBOL = 'BTC (Regtest)';
+
+export const BTC_REGSTEST_TOKEN_ID: unique symbol = Symbol(BTC_REGSTEST_SYMBOL);
```
AntonioVentilii-DFINITY, line 40:
```suggestion
export const BTC_REGTEST_TOKEN_ID: unique symbol = Symbol(BTC_REGTEST_SYMBOL);
```

src/frontend/src/env/tokens.btc.env.ts
```diff
@@ -30,7 +30,22 @@ export const BTC_TESTNET_TOKEN: Token = {
...
+
+export const BTC_REGSTEST_TOKEN: Token = {
+	id: BTC_REGSTEST_TOKEN_ID,
```
AntonioVentilii-DFINITY, line 43:
```suggestion
	id: BTC_REGTEST_TOKEN_ID,
```

src/frontend/src/btc/derived/networks.derived.ts
```diff
@@ -8,7 +8,7 @@ export const enabledBitcoinNetworks: Readable<Network[]> = derived([testnets], (
 	NETWORK_BITCOIN_ENABLED
 		? [
 				...(BTC_MAINNET_ENABLED ? [BTC_MAINNET_NETWORK] : []),
-				...($testnets ? [BTC_TESTNET_NETWORK] : [])
+				...($testnets ? [BTC_TESTNET_NETWORK, BTC_REGTEST_NETWORK] : [])
```
AntonioVentilii-DFINITY, line 11: same as other comment

src/frontend/src/btc/derived/tokens.derived.ts
DenysKarmazynDFINITY: @lmuntaner can you please take a look at Antonio's question?

src/frontend/src/env/tokens.btc.env.ts
```diff
@@ -30,7 +30,22 @@ export const BTC_TESTNET_TOKEN: Token = {
 	standard: 'bitcoin',
 	category: 'default',
 	name: 'Bitcoin (Testnet)',
-	symbol: 'BTC (Testnet)',
+	symbol: BTC_TESTNET_SYMBOL,
+	decimals: BTC_DECIMALS,
+	icon: bitcoinTestnet
```
peterpeterparker, line 35: Regtest and testnet do not have different logo?

src/frontend/src/env/tokens.btc.env.ts
AntonioVentilii-DFINITY: Good catch! I saw you corrected both in `mainnet` and `testnet`, thank you!

AntonioVentilii-DFINITY: Strange that it is done in the same way for `ETH` and `SepoliaETH`... maybe @peterpeterparker has a better insight?

src/frontend/src/btc/derived/tokens.derived.ts
lmuntaner: Yes, that's makes a lot of sense, I agree.

src/frontend/src/env/tokens.btc.env.ts
lmuntaner: We don't have a logo for regtest yet and we thought we leave the same as bitcoin testnet.

Regtest is only available for local development, no user will see it.

peterpeterparker: Does it needs to be the same logo then?

If you don't specify a logo then it default to a color generate based on the name of the token.

lmuntaner: Maybe we should remove the logo and use the default. Wdyt @DenysKarmazynDFINITY ?

src/frontend/src/btc/derived/tokens.derived.ts
DenysKarmazynDFINITY: Agreed, it's now added to the array only in the local env

src/frontend/src/env/tokens.btc.env.ts
DenysKarmazynDFINITY: Totally, I removed the icon property

DenysKarmazynDFINITY: Since it doesn't look too pretty w/o having icons for the network and the token at the same time (see the first screenshot), I used `bitcoinTestnet` for the BTC Regtest token (see the second screenshot). Please let me know if you have other ideas here. 

Screenshot 2024-09-09 at 10 55 43

Screenshot 2024-09-09 at 11 00 19

lmuntaner: I don't have a strong opinion on one or the other. @peterpeterparker your call.

peterpeterparker: Sounds good.

AntonioVentilii-DFINITY (approved):
LGTM!

PR #2249, opened 2024-09-05
feat(ci): auto-update package-lock file

AntonioVentilii-DFINITY:
# Motivation

it would be useful to have a check/action that auto-updates the `package-lock.json` file if there is a need for it if `package.json` has changed.


peterpeterparker:
I would say not until we got tests. There might be often updates and each of those will need tests.

AntonioVentilii-DFINITY:
@peterpeterparker then I will leave as draft for now

AntonioVentilii-DFINITY:
yes yes, agreed, makes sense. I will just leave it here as draft otherwise I end up forgetting it

AntonioVentilii:
Since it was my old account, it is superseded by:

- PR #

PR #2251, opened 2024-09-05
style(frontend): change modal title for conversions

AntonioVentilii-DFINITY:
# Motivation

We already change the title of the modal for ck-conversions on the first step: so we adapt the last step too, changing the title from `Sending` to `Converting`.


peterpeterparker (approved):
LGTM, thx

PR #2255, opened 2024-09-06
refactor(frontend): remove redundant dependency from network filter util

AntonioVentilii-DFINITY:
# Motivation

The store `pseudoNetworkChainFusion` already uses the store `selectedNetwork`, so we simplify the logic re-using it.


peterpeterparker (commented):
For readability reasons, I would stick to using `pseudoNetworkChainFusion`.

peterpeterparker:
It's also future prone in case the rules for `pseudoNetworkChainFusion` ever changes.

PR #2260, opened 2024-09-06
feat(tokens): re-use code needed to calculate token USD balance

DenysKarmazynDFINITY:
# Motivation

The idea is to re-use code needed to calculate token USD balance.

### DIFF
--- a/src/frontend/src/lib/utils/tokens.utils.ts
+++ b/src/frontend/src/lib/utils/tokens.utils.ts
@@ -50,6 +50,36 @@ export const sortTokens = ({
 };
 
 /**
+ * Calculates USD balance for the provided token.
+ *
+ * @param token - Token for which USD balance will be calculated.
+ * @param $balancesStore - The balances data for the tokens.
+ * @param $exchanges - The exchange rates data for the tokens.
+ * @returns The USD balance or undefined in case the number cannot be calculated.
+ *
+ */
+export const calculateTokenUsdBalance = ({
+	token,
+	$balances,
+	$exchanges
+}: {
+	token: Token;
+	$balances: CertifiedStoreData<BalancesData>;
+	$exchanges: ExchangesData;
+}): number | undefined => {
+	const balance: BigNumber | undefined = $balances?.[token.id]?.data;
+	const exchangeRate: number | undefined = $exchanges?.[token.id]?.usd;
+
+	return nonNullish(exchangeRate)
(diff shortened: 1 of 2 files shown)


src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
AntonioVentilii-DFINITY, line 1: Can you change the tests on `pinTokensWithBalanceAtTop` too, so that now `calculateTokenUsdBalance` can basically be a mocked function?

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -76,15 +106,8 @@ export const pinTokensWithBalanceAtTop = ({
...
-			const exchangeRate: number | undefined = $exchanges?.[token.id]?.usd;
 
-			const usdBalance: number | undefined = nonNullish(exchangeRate)
-				? usdValue({
-						token,
-						balance,
-						exchangeRate
-					})
-				: undefined;
+			const usdBalance = calculateTokenUsdBalance({ token, $balances, $exchanges });
```
AntonioVentilii-DFINITY, line 110: just for readability, i would leave the type here too

`: number | undefined`

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -49,6 +49,36 @@ export const sortTokens = ({
...
+	$balances: CertifiedStoreData<BalancesData>;
+	$exchanges: ExchangesData;
+}): number | undefined => {
+	const balance: BigNumber | undefined = $balances?.[token.id]?.data;
+	const exchangeRate: number | undefined = $exchanges?.[token.id]?.usd;
+
+	return nonNullish(exchangeRate)
+		? usdValue({
+				token,
+				balance,
```
AntonioVentilii-DFINITY, line 76: maybe you don't need to create the `balance` as `const` and pass it here directly?

DenysKarmazynDFINITY: done

src/frontend/src/lib/utils/tokens.utils.ts
DenysKarmazynDFINITY: done

AntonioVentilii-DFINITY (approved):
LGTM! tks

PR #2261, opened 2024-09-06
feat(tokens): implement function to sum mainnet tokens USD value

DenysKarmazynDFINITY:
# Motivation

The idea is to create function for calculating Mainnet tokens total USD.


src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
```diff
@@ -368,3 +369,65 @@ describe('calculateTokenUsdBalance', () => {
...
+		const result = sumMainnetTokensUsdBalance({ $tokens: tokens, $balances: balances, $exchanges });
+		expect(result).toEqual(0);
+	});
+
+	it('should return 0 if no tokens are provided', () => {
+		const result = sumMainnetTokensUsdBalance({ $tokens: [], $balances, $exchanges });
+		expect(result).toEqual(0);
+	});
+
+	it('should return 0 if exchange rate is not available', () => {
```
AntonioVentilii-DFINITY, line 424: i think the last two tests are not needed, since they refer to testing `calculateTokenUsdBalance` basically, right?

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -152,6 +152,34 @@ export const pinTokensWithBalanceAtTop = ({
...
+	$tokens,
+	$balances,
+	$exchanges
+}: {
+	$tokens: Token[];
+	$balances: CertifiedStoreData<BalancesData>;
+	$exchanges: ExchangesData;
+}) =>
+	$tokens.reduce((acc, token) => {
+		if (token.network.env === 'mainnet') {
```
AntonioVentilii-DFINITY, line 174: maybe doing a one-line definition?

`token.network.env === 'mainnet' ? acc + ..... : acc`

or something similar

DenysKarmazynDFINITY: Agreed, I shortened it accordingly

src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
DenysKarmazynDFINITY: You're right, no need to test these 2 cases for this function. Removed.

AntonioVentilii-DFINITY (approved):
LGTM tks

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -152,6 +152,33 @@ export const pinTokensWithBalanceAtTop = ({
...
+/**
+ * Calculates total USD balance of mainnet tokens from the provided tokens list.
+ *
+ * @param $tokens - The list of tokens for filtering by network env and total USD balance calculation.
+ * @param $balancesStore - The balances data for the tokens.
+ * @param $exchanges - The exchange rates data for the tokens.
+ * @returns The sum of mainnet tokens USD balance.
+ *
+ */
+export const sumMainnetTokensUsdBalance = ({
```
peterpeterparker, line 164: 1. Can you type the return type

2. Is it correct to default to zero as result if there is no mainnet token? To some extend how to you know if the sum is 0 because there is a zero balance on mainnet or because there is no tokens related to mainnet?

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -152,6 +152,33 @@ export const pinTokensWithBalanceAtTop = ({
...
+	$tokens,
+	$balances,
+	$exchanges
+}: {
+	$tokens: Token[];
+	$balances: CertifiedStoreData<BalancesData>;
+	$exchanges: ExchangesData;
+}) =>
+	$tokens.reduce(
```
peterpeterparker, line 173: Is there a use case in the app in which this function would be called with an empty exchanges or balances? If yes, it would be more performant to not iterate over the tokens in those cases.

peterpeterparker:
Please see comments @DenysKarmazynDFINITY, I was reviewing the PR when you merged it.

src/frontend/src/lib/utils/tokens.utils.ts
DenysKarmazynDFINITY: Just to be sure, I tested it, and we can indeed add a check to not go through tokens if some of the required data is not yet available.

src/frontend/src/lib/utils/tokens.utils.ts
DenysKarmazynDFINITY: 1. Sure.
2. ICP and ETH mainnet tokens are always on by default. If we assume that no-mainnet-tokens case can indeed happen, for me it becomes a product/design question - do we still display "$0" in NetworkSwitcher, do we avoid showing description completely, do we show a small text message with a hint that no tokens available, etc. Do you think we need to be prepared for this case now or is it okay to handle it if it is decided to remove pre-selected tokens feature?

src/frontend/src/lib/utils/tokens.utils.ts
DenysKarmazynDFINITY: I applied this comment as well as added return type here - https://github.com/dfinity/oisy-wallet/pull/2264

PR #2262, opened 2024-09-06
feat(tokens): create derived stores for tokens total USD balance

DenysKarmazynDFINITY:
# Motivation

The idea is to create derived stores for tokens total USD balance.

### DIFF
--- a/src/frontend/src/lib/derived/tokens.derived.ts
+++ b/src/frontend/src/lib/derived/tokens.derived.ts
@@ -8,10 +8,11 @@ import { icrcChainFusionDefaultTokens, sortedIcrcTokens } from '$icp/derived/icr
 import type { IcToken } from '$icp/types/ic';
 import { exchanges } from '$lib/derived/exchange.derived';
 import { balancesStore } from '$lib/stores/balances.store';
-import type { Token, TokenToPin, TokensTotalUsdBalancesPerNetwork } from '$lib/types/token';
+import type { Token, TokenToPin } from '$lib/types/token';
+import type { TokensTotalUsdBalancePerNetwork } from '$lib/types/token-balance';
 import {
-	calculateMainnetTokensUsdBalancesPerNetwork,
-	filterEnabledTokens
+	filterEnabledTokens,
+	sumMainnetTokensUsdBalancesPerNetwork
 } from '$lib/utils/tokens.utils';
 import { derived, type Readable } from 'svelte/store';
 
@@ -64,9 +65,9 @@ export const enabledIcTokens: Readable<IcToken[]> = derived(
 /**
  * A store with NetworkId-number dictionary with total USD balance of mainnet tokens per network.
  */
-export const enabledMainnetTokensUsdBalancesPerNetwork: Readable<TokensTotalUsdBalancesPerNetwork> =
+export const enabledMainnetTokensUsdBalancesPerNetwork: Readable<TokensTotalUsdBalancePerNetwork> =
 	derived([enabledTokens, balancesStore, exchanges], ([$enabledTokens, $balances, $exchanges]) =>
-		calculateMainnetTokensUsdBalancesPerNetwork({
+		sumMainnetTokensUsdBalancesPerNetwork({
 			$tokens: $enabledTokens,
 			$balances,
 			$exchanges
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/derived/tokens.derived.ts
```diff
@@ -55,3 +57,20 @@ export const enabledIcTokens: Readable<IcToken[]> = derived(
 	([$enabledTokens]) =>
 		$enabledTokens.filter(({ standard }) => standard === 'icp' || standard === 'icrc') as IcToken[]
 );
+
+export const enabledMainnetErc20TokensTotalUsd: Readable<number> = derived(
```
peterpeterparker, line 61: Does the use case require different derived stores, or is the information used for the same feature? If it's the latter, it would be more efficient to iterate over the tokens once to calculate the sum.

DenysKarmazynDFINITY: I'd say they have to be implemented separately, as all 3 stores will be used in the following PR where I add UI components for this feature.

peterpeterparker: Actually, regardless of how many derived stores you need, would it be more efficient to derive an object containing the sums - e.g. `{total: number, ic: number, erc20: number}` - by iterating over the tokens only once. Afterward, whether you use the entire object directly or access specific properties, the performance cost will be minimal.

WDYT?

DenysKarmazynDFINITY: This is a great idea 👍 This way, if I use network id as a key for this dictionary, I can easily access this data in `Network` component, without manually checking `isNetworkIdICP`, `isNetworkIdEthereum` and so on.

DenysKarmazynDFINITY: Done. The store is now `Record<NetworkId, number>`. The total value can now be calculated just by adding all values.

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -153,33 +158,41 @@ export const sumTokensUiUsdBalance = (tokens: TokenUi[]): number =>
...
+		? $tokens.reduce<TokensTotalUsdBalancesPerNetwork>((acc, token) => {
+				if (token.network.env === 'mainnet') {
+					if (isNullish(acc[token.network.id])) {
+						acc[token.network.id] = 0;
+					}
+
+					return {
+						...acc,
+						[token.network.id]:
+							acc[token.network.id] +
```
AntonioVentilii-DFINITY, line 188: any loss in performance in doing here `(acc[token.network.id] ?? 0)`

DenysKarmazynDFINITY: Totally! It's a one-liner now

src/frontend/src/lib/types/token.ts
```diff
@@ -48,3 +48,5 @@ interface TokenFinancialData {
 export type TokenUi = Token & TokenFinancialData;
 
 export type TokenIndexKey = string;
+
+export type TokensTotalUsdBalancesPerNetwork = Record<NetworkId, number>;
```
peterpeterparker, line 52: I would move this to a dedicated type module like `token-something.ts`.

`token.ts` currently define a single entity, maybe opiniated but just feels like it would be a bit cleaner what do you think?

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -124,33 +129,37 @@ export const sumTokensUiUsdBalance = (tokens: TokenUi[]): number =>
...
  *
  * @param $tokens - The list of tokens for filtering by network env and total USD balance calculation.
  * @param $balancesStore - The balances data for the tokens.
  * @param $exchanges - The exchange rates data for the tokens.
- * @returns The sum of mainnet tokens USD balance if balances and exchanges stores are available.
+ * @returns A NetworkId-number dictionary with total USD balance of mainnet tokens per network.
  *
  */
-export const sumMainnetTokensUsdBalance = ({
+export const calculateMainnetTokensUsdBalancesPerNetwork = ({
```
peterpeterparker, line 140: `sumMainnetTokensUsdBalancesPerNetwork` would not have been correct too?

peterpeterparker (approved):
LGTM, thanks

Minor suggestion and question.

src/frontend/src/lib/types/token.ts
DenysKarmazynDFINITY: Agree! Moved to `token-balance.ts`.

src/frontend/src/lib/utils/tokens.utils.ts
DenysKarmazynDFINITY: Yes, renamed.

PR #2263, opened 2024-09-06
refactor(frontend): move test mock values to specific folders

AntonioVentilii-DFINITY:
# Motivation

For readability and reusability we move the mocked values of tokens, balances and exchanges stores in their specific modules.

### DIFF
--- a/src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
+++ b/src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
@@ -14,29 +14,10 @@ import {
 	sumMainnetTokensUsdBalance,
 	sumTokensUiUsdBalance
 } from '$lib/utils/tokens.utils';
-import { BigNumber } from 'alchemy-sdk';
 import { describe, expect, it, type MockedFunction } from 'vitest';
-
-const usd = 1;
-
-const certified = true;
-
-const bn1 = BigNumber.from(1n);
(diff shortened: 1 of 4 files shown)


src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
```diff
@@ -14,29 +14,10 @@ import {
...
-	[ETHEREUM_TOKEN.id]: { data: bn3, certified }
-};
-
-const $exchanges: ExchangesData = $tokens.reduce<ExchangesData>((acc, token) => {
-	acc[token.id] = { usd };
-	return acc;
-}, {});
+import { $balances, bn1, bn2, bn3, certified } from '../../mocks/balances.mock';
+import { $exchanges, usd } from '../../mocks/exchanges.mock';
+import { $tokens } from '../../mocks/tokens.mock';
```
DenysKarmazynDFINITY, line 20: The import aliases do not work for `test` folder?

DenysKarmazynDFINITY (approved):

src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
AntonioVentilii-DFINITY: we don't have an alias for the mock folder, but we can add it for another PR

PR #2266, opened 2024-09-06
refactor(frontend): move util function calculateTokenUsdBalance

AntonioVentilii-DFINITY:
# Motivation

Just for coherence, we move the function `calculateTokenUsdBalance` to the `token.util` module.


DenysKarmazynDFINITY (approved):
Thanks for moving it!

PR #2267, opened 2024-09-06
feat(frontend): create map for TokenUi

AntonioVentilii-DFINITY:
# Motivation

For better readability and reusability, we create a function to map an object `Token` to an object `TokenUi`.

### DIFF
--- a/src/frontend/src/lib/utils/token.utils.ts
+++ b/src/frontend/src/lib/utils/token.utils.ts
@@ -3,7 +3,7 @@ import type { BalancesData } from '$lib/stores/balances.store';
 import type { CertifiedStoreData } from '$lib/stores/certified.store';
 import type { CanisterIdText } from '$lib/types/canister';
 import type { ExchangesData } from '$lib/types/exchange';
-import type { Token, TokenStandard } from '$lib/types/token';
+import type { Token, TokenStandard, TokenUi } from '$lib/types/token';
 import type { TokenToggleable } from '$lib/types/token-toggleable';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -88,3 +88,28 @@ export const calculateTokenUsdBalance = ({
...
 };
+
+/** Maps a Token object to a TokenUi object, meaning it adds the balance and the USD balance to the token.
+ *
+ * @param token - The given token.
+ * @param $balancesStore - The balances data for the tokens.
+ * @param $exchanges - The exchange rates data for the tokens.
+ * @returns The token UI.
+ */
+export const mapTokenUi = ({
```
peterpeterparker, line 99: Should we add a test since it's a utils?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/utils/token.utils.ts
AntonioVentilii-DFINITY: i thought of that, but it is actually just a map, in the sense that it actually has no logic inside, just an aggregation of other functions (that have tests). Maybe it is not an util? is there a map folder?

peterpeterparker: If has some logic, it calls `calculateTokenUsdBalance` and map a balance only if the related balance is defined.

AntonioVentilii-DFINITY: right! included

PR #2272, opened 2024-09-06
fix(frontend): toggle should show all non-zero balances

AntonioVentilii-DFINITY:
# Motivation

The "zero balance" toggle should show non-zero balances even for token that have no exchange rate.


peterpeterparker (approved):
LGTM, thx

PR #2274, opened 2024-09-06
feat(frontend): avoid sorting if no balances store

AntonioVentilii-DFINITY:
# Motivation

It does not make sense to run the sorting by balance if the balance store is not initiated.

### DIFF
--- a/src/frontend/src/lib/derived/network-tokens.derived.ts
+++ b/src/frontend/src/lib/derived/network-tokens.derived.ts
@@ -37,5 +37,5 @@ export const combinedDerivedSortedNetworkTokensUi: Readable<TokenUi[]> = derived
 					$balances,
 					$exchanges
 				})
-			: $enabledNetworkTokens
+			: ($enabledNetworkTokens as TokenUi[])
 );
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -25,13 +26,16 @@ export const combinedDerivedSortedNetworkTokens: Readable<Token[]> = derived(
...
-			$balances,
-			$exchanges
-		})
+		nonNullish($balances)
+			? pinTokensWithBalanceAtTop({
+					$tokens: $enabledNetworkTokens,
+					$balances,
+					$exchanges
+				})
+			: ($enabledNetworkTokens as TokenUi[])
```
peterpeterparker, line 40: Why this cast? Shouldn't we map the tokens? What if we add fields in TokenUI in the future?

AntonioVentilii-DFINITY: Right! Moved internally using the `mapTokenUi` functions

peterpeterparker:
It's a feature not refactoring. Updated the title.

peterpeterparker (commented):
LGTM, thanks

Note that I assume you added the `.map` for the best performances, that's cool. To some extension, we could also return the result of the `.reduce()` which maps as well.

```
if (isNullish($balance)) {
  return [...positiveBalances, ...nonPositiveBalances]
}
```

AntonioVentilii-DFINITY:
Note that I assume you added the .map for the best performances, that's cool.

Indeed, I could have return the `nonPositiveBalances` directly after checking balance data, but tecnically it is one if condition less (the one that splits into positive and negative balances.

peterpeterparker (approved):
Sorry, forgot to click "approve" when I reviewed it earlier today. Just added a comment but, it was already fine.

PR #2278, opened 2024-09-07
feat(frontend): avoid calculating USD value if no balance store

AntonioVentilii-DFINITY:
# Motivation

There is no need to calculate the USD value if the balance store is nullish.

### DIFF
--- a/src/frontend/src/lib/utils/exchange.utils.ts
+++ b/src/frontend/src/lib/utils/exchange.utils.ts
@@ -1,6 +1,6 @@
-import { ZERO } from '$lib/constants/app.constants';
 import type { Token } from '$lib/types/token';
 import { formatToken } from '$lib/utils/format.utils';
+import { nonNullish } from '@dfinity/utils';
 import type { BigNumber } from '@ethersproject/bignumber';
 
 export const usdValue = ({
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/utils/exchange.utils.ts
```diff
@@ -12,10 +12,12 @@ export const usdValue = ({
...
-	) * exchangeRate;
+	nonNullish(balance)
+		? Number(
+				formatToken({
+					value: balance,
+					unitName: decimals,
+					displayDecimals: decimals
+				})
+			) * exchangeRate
+		: 0;
```
peterpeterparker, line 23: not sure if it makes sense to use ZERO constant here?

peterpeterparker (approved):
LGTM, thx

peterpeterparker:
It's a feature, not refactoring. Updated the title.

src/frontend/src/lib/utils/exchange.utils.ts
AntonioVentilii-DFINITY: you mean inside `Number(ZERO)` ? For sure, if you prefer. I did like that so that we have less computational work, simply `0`, eheh

peterpeterparker: Yeah something like `ZERO.toNumber()`. I was just thinking that maybe it would be convenient to search everywhere in the app where we use zero for a feature? Not sure it's worth it, just a though. That's why the PR is already approved ;)

AntonioVentilii-DFINITY: ah! makes sense indeed! it could be useful at some point.

PR #2281, opened 2024-09-08
fix(frontend): token balance could be null or undefined

AntonioVentilii-DFINITY:
# Motivation

There is a difference between an `undefined` balance and a `null` balance. The balance is `undefined` if the balance store is not initiated or the specific balance loader for the token is not initiated. But, if the balance loader was initiated at some point, it will either contain data or be `null`, but not `undefined`.

We distinguish the two cases to make it easier to understand when a Token could still need to wait for the balance loader to exists, or if the balance loader is simply not working (disconnected/stopped/error/etc).


src/frontend/src/lib/types/token.ts
```diff
@@ -41,7 +41,7 @@ export type OptionTokenStandard = TokenStandard | undefined | null;
 export type TokenToPin = Pick<Token, 'id'> & { network: Pick<Token['network'], 'id'> };
 
 interface TokenFinancialData {
-	balance?: BigNumber;
+	balance?: BigNumber | null;
```
peterpeterparker, line 44: Just an idea that crossed my mind and not for this PR: maybe we can create a TypeScript util for those nullable types?

peterpeterparker (approved):
LGTM, thanks

Nitpick: Regarding title of the PR, isn't this more a fix than a feature?

src/frontend/src/lib/types/token.ts
AntonioVentilii-DFINITY: similar to `mapAddress` you mean? generalize it?

```
export const mapAddress = <T extends Address>(
	$addressStore: StorageAddressData<T>
): OptionAddress<Address> => ($addressStore === null ? null : $addressStore?.data);
```

peterpeterparker: No I meant a type, more like `RequiredExcept` or `ResultSuccessReduced`.

AntonioVentilii-DFINITY: Started working on it with:

https://github.com/dfinity/oisy-wallet/pull/2326

PR #2282, opened 2024-09-08
style(frontend): improve glitching when changing networks

AntonioVentilii-DFINITY:
# Motivation

It was reported that the `animate:flip` feature when changing networks or settings was quite glitchy, since it would not be correctly managed together with `transition:fade`, causing overlapping of token cards and components.
Based on [this suggestion from StackOverflow](https://stackoverflow.com/questions/75154309/how-to-animateflip-with-in-outfade/75156370#75156370) on a similar topic, I changed the list to be `flex`. The explanation on StackOverflow is:

> Looking at the CSS of the vanishing elements they have a `transform: translate()` property during the animation inside the flex container and it's missing in the other case.

### Before

https://github.com/user-attachments/assets/ddc76fd9-d226-4746-b99a-6d099211936f

### After

https://github.com/user-attachments/assets/def85c8e-8f40-49cb-b755-63c6748a5c17


peterpeterparker (approved):
LGTM, thx

PR #2288, opened 2024-09-09
feat(lint): sort svelte imports

AntonioVentilii-DFINITY:
# Motivation

As final step of sorting Svelte imports, we make the rule include all the `.svelte` files.


peterpeterparker (approved):
Thanks!

PR #2289, opened 2024-09-09
chore(frontend): Upgrade ckbtc

lmuntaner:
# Motivation

I want to use `"regtest"` from `BitcoinNetwork` type. The new entry was added in https://github.com/dfinity/ic-js/pull/713 and it hasn't been published in any version but next yet.

# Changes

* Upgrade ckbtc to `3.0.0-next-2024-09-09`

# Tests

I checked that the new entry is accessible in node_modules.


peterpeterparker (approved):
LGTM, thx

PR #2290, opened 2024-09-09
feat(frontend): show skeleton when starting balance loader

AntonioVentilii-DFINITY:
# Motivation

Since the `undefined` balance was introduced with PR #2281 for tokens that still don't have the balance loader working, we can show such information to the user visually: we provide a skeleton text until the balance is either some data or `null`, but, either way, until the balance loader exists.

# Changes

- Create component for token balance skeleton.
- Add it to the Token card component.

# Tests

### Normal speed

https://github.com/user-attachments/assets/11b817af-76e1-41c4-b9bc-7d17ecb2032e

### Balance loader initiator is slowed down

https://github.com/user-attachments/assets/527b0b3f-2b14-43de-a4c6-5166fbbd1a00


AntonioVentilii-DFINITY:
@artkorotkikh-dfinity have a look please

src/frontend/src/lib/components/tokens/SkeletonTokenBalance.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
```
peterpeterparker, line 1: For consistency, probably `TokenBalanceSkeleton` instead of `SkeletonTokenBalance`

src/frontend/src/lib/components/tokens/SkeletonTokenBalance.svelte
```diff
@@ -0,0 +1,12 @@
...
+	import SkeletonText from '$lib/components/ui/SkeletonText.svelte';
+	import type { TokenUi } from '$lib/types/token';
+
+	export let token: TokenUi;
+</script>
+
+{#if token.balance === undefined}
+	<span class="w-full max-w-[100px] mt-2"><SkeletonText /></span>
+{:else}
+	<slot />
```
peterpeterparker, line 11: If balance equals null we display zero?

AntonioVentilii-DFINITY: correct, at least for now: I checked the code, and when it is set to `null` is because there was an error with the balance loader and the user receive a toast error warning.

It would make sense to give some more permanent feedback there in such case, I will check with @artkorotkikh-dfinity , but, in any case, I would leave for another PR, WDYT?

src/frontend/src/lib/components/tokens/SkeletonTokenBalance.svelte
AntonioVentilii-DFINITY: funny enough: i was undecided between the two and I tossed a coin eheh!

Changed

src/frontend/src/lib/components/tokens/SkeletonTokenBalance.svelte
peterpeterparker: Sounds good

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/components/tokens/SkeletonTokenBalance.svelte
artkorotkikh-dfinity: Would it be better to display the same `$-` symbols that we show for SNSes without prices if the balance is null? I know this shouldn’t happen, but displaying 0 might be misleading from the user’s perspective.

image

artkorotkikh-dfinity:
@AntonioVentilii-DFINITY could we please show a skeleton for dollar values as well?

src/frontend/src/lib/components/tokens/SkeletonTokenBalance.svelte
AntonioVentilii-DFINITY: I like it: we already do that if there is no exchange rate, so for me is cool

@peterpeterparker WDYT?

AntonioVentilii-DFINITY:
@artkorotkikh-dfinity , doing with https://github.com/dfinity/oisy-wallet/pull/2342

src/frontend/src/lib/components/tokens/SkeletonTokenBalance.svelte
peterpeterparker: Makes sense

AntonioVentilii-DFINITY: Done with https://github.com/dfinity/oisy-wallet/pull/2352

PR #2292, opened 2024-09-09
feat(frontend): auto-enable ERC20 tokens when converting from ckERC20

AntonioVentilii-DFINITY:
# Motivation

When the user converts ckERC20 to ERC20, it is useful that the ERC20 token is enabled if it is not. We use the same flow as per the Receive button: when the user click the Convert button if the token is not enabled, it will be.
Thanks to @peterpeterparker for the suggestion.

# Tests

In local replica, the following test worked:

- Disable Sepolia USDC.
- Convert ckSepoliaUSD to Sepolia USDC.
- Check if Sepolia USDC was enabled.

https://github.com/user-attachments/assets/7a3a5ea3-f91f-4182-b181-7481350fb545


peterpeterparker (commented):
I'm not sure why  this PR. We already have `autoLoadUserToken`, what is different?

If anything is different, can we implement the feature the same way as the existing pattern? Or at least with the same flow?

Also reloading the list of tokens and handling errors seems to be missing in the implementation.

AntonioVentilii-DFINITY:
> I'm not sure why this PR. We already have autoLoadUserToken, what is different?

Sorry, I didn't know we had it already, let me check

AntonioVentilii-DFINITY:
@peterpeterparker thank you! that was way easier! I used the same exact flow that we use for `Receive`

peterpeterparker (approved):
LGTM, thx

PR #2294, opened 2024-09-09
test(frontend): increase fake ck Sepolia ETH sent to local replica

AntonioVentilii-DFINITY:
# Motivation

The amount of ckSepoliaETH sent by default was very small to be used for testing.


peterpeterparker (approved):
LGTM, thx

PR #2291, opened 2024-09-09
style(frontend): update modals style

AntonioVentilii-DFINITY:
# Motivation

Modal style updated as per requirements.

### Before

Screenshot 2024-09-09 at 14 10 40
Screenshot 2024-09-09 at 14 10 47
Screenshot 2024-09-09 at 14 10 53

### After

Screenshot 2024-09-09 at 14 09 37
Screenshot 2024-09-09 at 14 09 44
Screenshot 2024-09-09 at 14 09 48

### Mobile

Screenshot 2024-09-09 at 14 20 02
Screenshot 2024-09-09 at 14 20 28

PR #2296, opened 2024-09-09
style(frontend): adjust button class for Receive modal

AntonioVentilii-DFINITY:
# Motivation

With the new modal style introduced with PR #2291  , we adjust the button of the Receive modal to match the rest.

### Before

Screenshot 2024-09-09 at 17 04 05

### After

Screenshot 2024-09-09 at 17 03 53

### DIFF
--- a/src/frontend/src/lib/components/receive/ReceiveAddresses.svelte
+++ b/src/frontend/src/lib/components/receive/ReceiveAddresses.svelte
@@ -99,6 +99,6 @@
 	</ReceiveAddressWithLogo>
 </div>
 
-<button class="primary full center text-center" on:click={modalStore.close}
+<button class="secondary full center text-center" on:click={modalStore.close}
 	>{$i18n.core.text.done}</button
 >


src/frontend/src/lib/components/receive/ReceiveAddresses.svelte
```diff
@@ -99,6 +99,6 @@
 	</ReceiveAddressWithLogo>
 </div>
 
-<button class="primary full center text-center" on:click={modalStore.close}
+<button class="secondary full center text-center" on:click={modalStore.close}
```
peterpeterparker, line 102: Sounds wrong to have a unique primary action called secondary.

AntonioVentilii-DFINITY: yes, it seemed off, but I am not really sure how to call it. That class will be used:
1. By secondary buttons (Close, Cancel, etc) when there is a primary
2. By buttons that are alone in a modal as "definitive" action (Done, Ok, Close, etc)

You think I shall create a new class for the second case?

peterpeterparker: UX wise this seems incorrect no? Same design for "cancel" on certain screen but, used for "confirmation" on other screen? Is that correct @ artkorotkikh-dfinity?

PR #2297, opened 2024-09-09
fix(frontend): Fix issue with sending erc20 tokens

DenysKarmazynDFINITY:
# Motivation

This PR fixes the issue with clicking "Max" button while sending erc20 tokens (e.g. PEPE) - for big values, it was showing the "Insufficient funds for amount" error even though in the UI the balance and "Amount" values were completely the same (see screenshot). 

544cdd22-4321-4d6f-bbb5-7d1938f7b031

The problem was reproducable with the following steps:
1. User clicks the "Max" button.
2. For erc20 tokens, the fee is deducted from ETH balance, therefore `getMaxTransactionAmount` takes the whole balance (e.g. `100000009000099049545454`), parses it according to `token.decimals` (in case of PEPE - 16), and returns the value (`100000009000099050000000` in this case).
3. `customValidate` is triggered with the value from the step 2.
4. `userAmount.gt(parsedSendBalance ?? ZERO)` returns `true` because it compares parsed value with raw balance (`100000009000099050000000` with `100000009000099049545454`).
5. `insufficient_funds_for_amount` is shown.

To fix the issue, I'm formatting `sendBalance` to be compatible with the value from the input. As the result, clicking the "Max" button in both `Send` and `Convert to ckToken` works correctly (screenshot below).

Screenshot 2024-09-09 at 18 57 36


AntonioVentilii-DFINITY:
@DenysKarmazynDFINITY renamed the PR title as per convention

src/frontend/src/tests/lib/utils/format.utils.spec.ts
```diff
@@ -0,0 +1,70 @@
+import { ZERO } from '$lib/constants/app.constants';
+import { formatBigNumberish } from '$lib/utils/format.utils';
+import { describe, expect } from 'vitest';
+
+describe('formatBigNumberish', () => {
+	it('formats bigint correctly when the value needs to be rounded', () => {
+		const value = 100000009000099049545454n;
+		const result = 10000000.900009904;
```
AntonioVentilii-DFINITY, line 8: I have to say that mathematically, this seems wrong...  that number, when rounded, should be `10000000.900009905`, no?

src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -86,3 +86,18 @@ export const formatUSD = (
...
+ * @param decimals The decimals to format with.
+ * @returns Formated and rounded number.
+ */
+export const formatBigNumberish = ({
+	value,
+	decimals
+}: {
+	value: BigNumberish;
+	decimals: number;
+}): number => Number(value) / 10 ** decimals;
```
AntonioVentilii-DFINITY, line 103: I would suggest to leave it in the universe of `BigNumber` and only in the final converting to `Number`, using something like `formatUnits` from `ethers` (I think we have a `parseToken` function that does something similar) and then `toFixed` if you want to round it

```
// pseudo-code
Utils.formatUnits(value, unitName:decimals).toNumber().toFixed(decimals)
```

src/frontend/src/tests/lib/utils/format.utils.spec.ts
DenysKarmazynDFINITY: I'm looking for explanation why `100000009000099049545454n` results in `10000000.900009904` (not rounded), and `100000009000099059545454n` -> `10000000.900009906` (last 5 becomes 6). It's the same result no matter whether I do the div operation with numbers or BigNumbers. Any ideas?

src/frontend/src/lib/utils/format.utils.ts
DenysKarmazynDFINITY: Just wondering what would be the benefit here? Do you think it will be more performant? It does this operation with numbers in `main`, I only extracted it in a separate function.

src/frontend/src/tests/lib/utils/format.utils.spec.ts
AntonioVentilii-DFINITY: That is pretty strange... what if you use `formatUnits`?

src/frontend/src/lib/utils/format.utils.ts
AntonioVentilii-DFINITY: Indeed, I think it was wrong from the beginning, it shouldn't have been done in `number`, it should have stayed in `BigNumberish` and then parsed

It is always safer, I think, since Typescript native `Number` type (which is a 64-bit floating-point number) cannot safely handle very large integers or perform operations on them without losing precision

DenysKarmazynDFINITY: I changed the implementation to use BigNumbers instead of numbers. Tested it for small and big values and decimals, working fine. Please take a look.

src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -86,3 +86,18 @@ export const formatUSD = (
...
 };
+
+/**
+ * Formats and if needed rounds in an Upward direction provided BigNumberish
+ *
+ * @param value BigNumberish to format.
+ * @param decimals The decimals to format with.
+ * @returns Formated and rounded number.
+ */
+export const formatBigNumberish = ({
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: As we switched to using BigNumbers, I don't need any additional utils. Therefore, removed this one.

src/frontend/src/tests/lib/utils/format.utils.spec.ts
DenysKarmazynDFINITY: Yes, it's the same no matter what type or util is used to make the calculations. I think we are fine as long as both values we use for comparisons (e.g. `userAmount` and `sendBalance`) are formatted in the same way.

src/frontend/src/eth/components/send/SendAmount.svelte
```diff
@@ -35,19 +37,31 @@
 			return;
 		}
 
+		// We should align the $sendBalance and userAmount to avoid issues caused by comparing formatted and unformatted BN
+		const parsedSendBalance = $sendBalance
```
AntonioVentilii-DFINITY, line 41: `nonNullish()` ?

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -19,17 +22,28 @@ import { nonNullish } from '@dfinity/utils';
  * @returns {number} The maximum amount for the transaction.
  */
 export const getMaxTransactionAmount = ({
```
AntonioVentilii-DFINITY, line 24: the docstring should be changed too right? from `bigint` to `BigNumber`

AntonioVentilii-DFINITY (approved):
LGTM! Just a small change for the docstring

src/frontend/src/lib/utils/token.utils.ts
DenysKarmazynDFINITY: Good catch, updated.

PR #2298, opened 2024-09-09
feat(frontend): Add other bitcoin addresses

lmuntaner:
# Motivation

# Changes

* One more index db store for the testnet address (no need to store the regtest).
* New service `listenTestnetNetworksToAddBtcTestnetAddresses` to listen to changes in the testnets store.
* Load the testnet address by listening to the testnet store with `listenTestnetNetworksToAddBtcTestnetAddresses` used in `Loader`.
* Two new derived stores to get the address for Bitcoin Testnet and Regtest.
* Expand `loadBtcAddress` to support all Bitcoin networks.
* Two new addresses to store Bitcoin testnet addresses.

# Tests

Tested locally.

### DIFF
--- a/src/frontend/src/lib/api/idb.api.ts
+++ b/src/frontend/src/lib/api/idb.api.ts
@@ -1,6 +1,6 @@
 import { browser } from '$app/environment';
 import { ETHEREUM_NETWORK_SYMBOL } from '$env/networks.env';
-import { BTC_MAINNET_SYMBOL } from '$env/tokens.btc.env';
+import { BTC_MAINNET_SYMBOL, BTC_TESTNET_SYMBOL } from '$env/tokens.btc.env';
 import type { BtcAddress, EthAddress } from '$lib/types/address';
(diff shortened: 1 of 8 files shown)


lmuntaner:
New UI:

Screenshot 2024-09-09 at 19 49 49

lmuntaner:
@peterpeterparker @AntonioVentilii-DFINITY the code is ready, but I set it to draft because I want to split it, but I'm not sure how to do so.

How should I split it? All the pieces are related, but in isolation none works.

Maybe I can put the loader of addresses in another PR? What do you think of this approach. I wasn't sure it's the best suited, but I didn't want to do it on init.

peterpeterparker:
> How should I split it?

First the store and services, then the loader and finally the UI.

Regarding fetching the addresses from the backend, we need an update call per network?
Have we though about fetching multiple networks addresses in a single call, would that be performance wise interesting?

lmuntaner:
This PR has been splitted with different PRs and the last one is already in review: https://github.com/dfinity/oisy-wallet/pull/2323

PR #2299, opened 2024-09-10
feat(frontend): derive auth identity

peterpeterparker:
# Motivation

Instead of accessing `$authStore.identity` in our components we can derive the optional identity `$authIdentity`. 

No big reasons, just a bit cuter to read.

# Changes

- Derive identity
- Use derived information instead of accessing store


AntonioVentilii-DFINITY (approved):
Tks!

PR #2300, opened 2024-09-10
feat(frontned): Add stores and change services for other bitcoin addresses

lmuntaner:
# Motivation

In this PR, I introduce stores for the testnet addresses and adapt current services to them.

# Changes

* One more index db store for the testnet address (no need to store the regtest).
* Two new stores to save Bitcoin testnet addresses.
* Two new derived stores to get the address for Bitcoin Testnet and Regtest.
* Expand loadBtcAddress to support all Bitcoin networks.

# Tests

This was tested along the UI code in #2298

### DIFF
--- a/src/frontend/src/lib/services/address.services.ts
+++ b/src/frontend/src/lib/services/address.services.ts
@@ -73,20 +73,20 @@ const loadTokenAddress = async <T extends Address>({
 };
 
 // We use the Testnet address for Regtest.
-type BtcTokenIdsWithAddresses = typeof BTC_MAINNET_TOKEN_ID | typeof BTC_TESTNET_TOKEN_ID;
-type BtcNetworkWithAddresses = Exclude<BitcoinNetwork, 'regtest'>;
-const bitcoinStoreMapper: Record<BtcNetworkWithAddresses, AddressStore<BtcAddress>> = {
+type TokenIdBtcPublicNetwork = typeof BTC_MAINNET_TOKEN_ID | typeof BTC_TESTNET_TOKEN_ID;
+type BtcPublicNetwork = Exclude<BitcoinNetwork, 'regtest'>;
+const bitcoinStoreMapper: Record<BtcPublicNetwork, AddressStore<BtcAddress>> = {
 	mainnet: btcAddressMainnetStore,
 	testnet: btcAddressTestnetStore
 };
 
-const bitcoinNetworkMapper: Record<BtcNetworkWithAddresses, SignerBitcoinNetwork> = {
+const bitcoinNetworkMapper: Record<BtcPublicNetwork, SignerBitcoinNetwork> = {
 	mainnet: { mainnet: null },
 	testnet: { testnet: null }
 };
 
 const idbBtcAddressMapper: Record<
-	BtcNetworkWithAddresses,
+	BtcPublicNetwork,
 	(params: SetIdbAddressParams<BtcAddress>) => Promise<void>
 > = {
 	mainnet: setIdbBtcAddressMainnet,
@@ -97,8 +97,8 @@ const loadBtcAddress = async ({
 	tokenId,
 	network
 }: {
-	tokenId: BtcTokenIdsWithAddresses;
-	network: BtcNetworkWithAddresses;
+	tokenId: TokenIdBtcPublicNetwork;
+	network: BtcPublicNetwork;
 }): Promise<ResultSuccess> =>
 	loadTokenAddress<BtcAddress>({
 		tokenId,
--- a/src/frontend/src/lib/services/address.services.ts
+++ b/src/frontend/src/lib/services/address.services.ts
@@ -1,6 +1,6 @@
 import type { BitcoinNetwork as SignerBitcoinNetwork } from '$declarations/signer/signer.did';
 import { NETWORK_BITCOIN_ENABLED } from '$env/networks.btc.env';
-import { BTC_MAINNET_TOKEN_ID, type BTC_SYMBOL } from '$env/tokens.btc.env';
+import { BTC_MAINNET_TOKEN_ID, BTC_TESTNET_TOKEN_ID } from '$env/tokens.btc.env';
 import { ETHEREUM_TOKEN_ID } from '$env/tokens.env';
 import {
 	getIdbBtcAddressMainnet,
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/stores/address.store.ts
```diff
@@ -25,3 +25,5 @@ const initAddressStore = <T extends Address>(): AddressStore<T> => {
 export const btcAddressMainnetStore = initAddressStore<BtcAddress>();
 
 export const ethAddressStore = initAddressStore<EthAddress>();
+export const btcAddressTestnetStore = initAddressStore<BtcAddress>();
+export const btcAddressRegtestStore = initAddressStore<BtcAddress>();
```
AntonioVentilii-DFINITY, line 29: just for readability, can you move it below the mainnet store?

AntonioVentilii-DFINITY (commented):
In general, I would prefer to avoid putting `regtest` logic inside IDB flow (even if `undefined`).

According to documentation for `regtest`, one can simply use the `testnet` address and query the local node, so it does not seem necessary to have conditions on  IDB for `regtest`.

Can we make it so the `regtest` address store refers to the `testnet` address (and whatever it entails too, like fetching, setting, etc)?

WDYT?

lmuntaner:
> In general, I would prefer to avoid putting `regtest` logic inside IDB flow (even if `undefined`).
> 
> According to documentation for `regtest`, one can simply use the `testnet` address and query the local node, so it does not seem necessary to have conditions on IDB for `regtest`.
> 
> Can we make it so the `regtest` address store refers to the `testnet` address (and whatever it entails too, like fetching, setting, etc)?
> 
> WDYT?

Sure, I remove the regtest store and also from the accepted networks and token ids when loading addresses. Is that what you meant?

lmuntaner:
@AntonioVentilii-DFINITY ready for another review

AntonioVentilii-DFINITY:
> Is that what you meant?

I think so, yes. It does make sense to you, right? Either we use the `testnet` address for `regtest`, or we create a complete set of IDB flow for `regtest` (but in that case it will be the exact same values as `testnet`). I am thinking that this could spare computational time and calls to backend. WDYT?

We can always change it, if at some point `regtest` starts to differ from `testnet`

src/frontend/src/lib/stores/address.store.ts
```diff
@@ -22,6 +22,8 @@ const initAddressStore = <T extends Address>(): AddressStore<T> => {
 	};
 };
 
+// We use the Testenet address for Regtest.
```
AntonioVentilii-DFINITY, line 25:
```suggestion
// We use the Testnet address for Regtest.
```

src/frontend/src/lib/services/address.services.ts
```diff
@@ -69,22 +72,43 @@ const loadTokenAddress = async <T extends Address>({
 	return { success: true };
 };
 
+// We use the Testenet address for Regtest.
```
AntonioVentilii-DFINITY, line 75:
```suggestion
// We use the Testnet address for Regtest.
```

src/frontend/src/lib/services/address.services.ts
```diff
@@ -69,22 +72,43 @@ const loadTokenAddress = async <T extends Address>({
...
+type BtcTokenIdsWithAddresses = typeof BTC_MAINNET_TOKEN_ID | typeof BTC_TESTNET_TOKEN_ID;
+type BtcTokenIdsStringWithAddresses = Exclude<BitcoinNetwork, 'regtest'>;
```
AntonioVentilii-DFINITY, line 77: why `-WithAddresses` suffix in the two types above? in the sense that this type is "to be used with addresses"?

BTW, shouldn't the second one be something like `BtcNetworkIdsWithAddresses`?

lmuntaner: I wanted to indicate that those are the btc network that have a different address. Regtest doesn't have it's unique address. Naming is hard...

Yes, second makes more sense `BtcNetworksWithAddresses`, done!

AntonioVentilii-DFINITY: Ahahahah agreed! Naming IS hard!

src/frontend/src/lib/services/address.services.ts
```diff
@@ -69,22 +72,43 @@ const loadTokenAddress = async <T extends Address>({
 	return { success: true };
 };
 
+// We use the Testnet address for Regtest.
+type BtcTokenIdsWithAddresses = typeof BTC_MAINNET_TOKEN_ID | typeof BTC_TESTNET_TOKEN_ID;
+type BtcNetworkWithAddresses = Exclude<BitcoinNetwork, 'regtest'>;
```
AntonioVentilii-DFINITY, line 77: For the naming, I would ask suggestions to @peterpeterparker , usually he gets them

AntonioVentilii-DFINITY (approved):
LGTM, tks!

src/frontend/src/lib/services/address.services.ts
lmuntaner: Done!

src/frontend/src/lib/services/address.services.ts
```diff
@@ -69,22 +72,43 @@ const loadTokenAddress = async <T extends Address>({
...
 		tokenId,
 		getAddress: (identity: OptionIdentity) =>
 			getBtcAddress({
 				identity,
-				network: network === 'testnet' ? { testnet: null } : { mainnet: null }
+				network: bitcoinNetworkMapper[network]
 			}),
-		setIdbAddress: setIdbBtcAddressMainnet,
-		addressStore: btcAddressMainnetStore
+		setIdbAddress: idbBtcAddressMapper[network],
```
peterpeterparker, line 110: Would it be interesting to merge `idbBtcAddressMapper` and `bitcoinStoreMapper` together?

Code wise we are fine with current implementation as we are using `network` to index the two but, technically there is no restriction so someday it might happens that one use `idbBtcAddressMapper` with `mainnet` and `bitcoinStoreMapper` with `testnet` or the contrary. You know what I mean?

```
...something[network]

something = {[network]: {addressMapper, storeMapper}}
```

lmuntaner: I think that if it starts to differ so much, maybe it's worth not using the same `loadBtcToken` function and create different ones.

peterpeterparker: I'm no sure I conveyed my idea clearly and therefore not sure you understood my feedback so here's a PR: https://github.com/dfinity/oisy-wallet/pull/2302

PR #2301, opened 2024-09-10
feat(frontend): Load testnet btc address

lmuntaner:
# Motivation

In this PR, we create a listener of the testnets store to load the btc testnet address in the store and set it up in the `Loader` component.

# Changes

* New service `listenTestnetNetworksToAddBtcTestnetAddresses` to listen to changes in the testnets store.
* Load the testnet address by listening to the testnet store with listenTestnetNetworksToAddBtcTestnetAddresses used in Loader.

# Tests

Tested together with the rest of the changes in #2298

### DIFF
--- a/src/frontend/src/lib/components/core/Loader.svelte
+++ b/src/frontend/src/lib/components/core/Loader.svelte
@@ -9,7 +9,11 @@
 	import InProgress from '$lib/components/ui/InProgress.svelte';
 	import { authIdentity } from '$lib/derived/auth.derived';
 	import { ProgressStepsLoader } from '$lib/enums/progress-steps';
-	import { loadAddresses, loadIdbAddresses } from '$lib/services/address.services';
+	import {
+		listenTestnetNetworksToAddBtcTestnetAddresses,
+		loadAddresses,
+		loadIdbAddresses
+	} from '$lib/services/address.services';
 	import { signOut } from '$lib/services/auth.services';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { loading } from '$lib/stores/loader.store';
@@ -63,6 +67,10 @@
 
 	onMount(async () => {
 		const { success: addressIdbSuccess, err } = await loadIdbAddresses();
+
+		// Listen to the user changing the testnets flag to load testnet btc addresses.
+		// We don't load them on init to avoid unnecessary requests.
+		listenTestnetNetworksToAddBtcTestnetAddresses();
 
 		if (addressIdbSuccess) {
 			loading.set(false);
--- a/src/frontend/src/lib/components/core/Loader.svelte
+++ b/src/frontend/src/lib/components/core/Loader.svelte
@@ -68,11 +68,11 @@
 
 	let progressModal = false;
 
-	const debounceLadBtcAddressTestnet = debounce(loadBtcAddressTestnet);
+	const debounceLoadBtcAddressTestnet = debounce(loadBtcAddressTestnet);
 
 	$: {
 		if ($testnets && nonNullish($btcAddressTestnet)) {
-			debounceLadBtcAddressTestnet();
+			debounceLoadBtcAddressTestnet();
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/services/address.services.ts
```diff
@@ -36,6 +37,17 @@ import type { Principal } from '@dfinity/principal';
 import { assertNonNullish, isNullish } from '@dfinity/utils';
 import { get } from 'svelte/store';
 
+export const listenTestnetNetworksToAddBtcTestnetAddresses = async (): Promise<void> => {
+	testnetsStore.subscribe(async (data) => {
```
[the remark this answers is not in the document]
lmuntaner: * Good point, I changed the approach. It's simpler, I don't know why I didn't think of it before.
* Also good point, done in the new approach.
* This could be done but at the UI and store level (in another PR). The store should indicate that the address is loading. Does the store have the meaning of "loading"? I see the type is `T | null | undefined`. Does one of them mean loading?

lmuntaner:

src/frontend/src/lib/services/address.services.ts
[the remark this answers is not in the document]
peterpeterparker: > This could be done but at the UI and store level (in another PR). The store should indicate that the address is loading. Does the store have the meaning of "loading"? I see the type is T | null | undefined. Does one of them mean loading?

Good point, `undefined`. Maybe not stricto sensu "loading", rather "not yet loaded or loading" but, yes!

src/frontend/src/lib/components/core/Loader.svelte
```diff
@@ -61,6 +67,12 @@
 
 	let progressModal = false;
 
+	$: {
+		if ($testnets && !$btcAddressTestnet) {
```
peterpeterparker, line 71: - `nonNullish`
- should we debounce the loader in case the user have fun toggling off and on the testnet repeatitively?
- while there is no eslint for it, in Oisy we generally `await` promises unless there is a significant reason not too. So I think for consistency reason we can do the same here.

src/frontend/src/lib/services/address.services.ts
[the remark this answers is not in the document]
lmuntaner: I can use `undefined` to show a spinner when I do the UI then.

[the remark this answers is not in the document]
peterpeterparker: Yeah. Altenartively, a skeleton given that the toggle for testnet find place in a popover menu and in the settings page.

src/frontend/src/lib/components/core/Loader.svelte
lmuntaner: - Done!
- Done!
- The is in the top level component where I can't use `await`. I can do a function that does an await inside of the loader, but I won't be able to await that function either. How woudl you do it?

peterpeterparker: > The is in the top level component where I can't use await

Sounds good now with the debounce, we have the same pattern.

You've got a typo though in `debounceLadBtcAddressTestnet` => "Lad", I guess "Load"?

src/frontend/src/lib/components/core/Loader.svelte
```diff
@@ -61,6 +68,14 @@
 
 	let progressModal = false;
 
+	const debounceLadBtcAddressTestnet = debounce(loadBtcAddressTestnet);
```
peterpeterparker, line 71: Not sure you have seen my last comment given that you ask me for review again, so repeating here: is that a typo?

`debounceLadBtcAddressTestnet`, `...Lad...`?

lmuntaner: Yes, that's a typo, thanks!

src/frontend/src/lib/components/core/Loader.svelte
lmuntaner: Fixed, thanks!

peterpeterparker (approved):
LGTM, thx

PR #2302, opened 2024-09-10
feat(frontend): single BTC mapper to load address

peterpeterparker:
# Motivation

Using two different BTC mappers accepting both a network as key is prone to misuage where one or the other is use with another network.

# Changes

- Merge `idbBtcAddressMapper` and `bitcoinStoreMapper`
- Extract a type for the parameters of `loadTokenAddress` to avoid code duplication in the types


AntonioVentilii-DFINITY (approved):

lmuntaner (approved):
No, I didn't understand it, now I see. Thanks!

PR #2303, opened 2024-09-10
ci(e2e): fix update-snapshots workflow

DenysKarmazynDFINITY:
# Motivation

To fix the `update-snapshots` workflow as well as to make sure that `e2e-tests` always works as expected, we need to add back the checkout step before building oisy-backend (even though the `oisy-backend` action also has the checkout step inside).

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -12,6 +12,9 @@ jobs:
   oisy-backend-wasm:
     runs-on: ubuntu-20.04
     steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
       - name: Build oisy-backend WASM
         uses: ./.github/actions/oisy-backend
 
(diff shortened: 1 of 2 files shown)


AntonioVentilii-DFINITY:
Is there a technical reason for that? Because, as you said,` oisy-backend `does it too

DenysKarmazynDFINITY:
> Is there a technical reason for that? Because, as you said,`oisy-backend`does it too

The workflow fails with the following error - https://github.com/dfinity/oisy-wallet/actions/runs/10790664021/job/29926355383 . Also, it is the case in all other workflow/action combinations, e.g. [backend-tests](https://github.com/dfinity/oisy-wallet/blob/main/.github/workflows/backend-tests.yml#L26) -> [docker-build-base](https://github.com/dfinity/oisy-wallet/blob/main/.github/actions/docker-build-base/action.yml#L10)

PR #2304, opened 2024-09-10
refactor(frontend): create type for Transaction type

AntonioVentilii-DFINITY:
# Motivation

Created common type for Transaction type.


src/frontend/src/lib/types/transaction.ts
```diff
@@ -12,3 +12,5 @@ export type Transaction = Omit<EthTransaction, 'data'> &
 export type TransactionFeeData = Pick<FeeData, 'maxFeePerGas' | 'maxPriorityFeePerGas'> & {
 	gas: BigNumber;
 };
+
+export type TransactionType = 'send' | 'receive';
```
peterpeterparker, line 16: There is already `IcTransactionType` so this should probably be scoped to `EthTransactionType`. Or we introduce `TransactionType` equals to send and receive and then `IcTransactionType` extends it or else.

peterpeterparker (approved):
LGTM, thx

PR #2305, opened 2024-09-10
feat(devops): Deploy signer from pre-built release

bitdivine:
# Motivation
The signer is now hosted in a separate repository and is available as a pre-built Wasm

# Changes
* Move the signer canister ids from `canister_ids.json` to `dfx.json` as befits an external canister.
* Simplify:
  * The signer now has only signing APIs, so there are fewer bindings.
  * The deploy args are simplified:
    * Some fields are no longer in the deploy args.
    * We can _probably_ reduce the ECDSA keys to 1
  * Deploying is now just `dfx deploy`

# Tests
Existing CI  should suffice


.nvmrc
lmuntaner, line 1: Could we add this in another PR?

I think they prefer that each PR has one single topic.

dfx.json
```diff
@@ -2,11 +2,19 @@
...
 		"signer": {
-			"candid": "src/backend/backend.did",
-			"package": "backend",
-			"type": "rust",
-			"optimize": "cycles",
-			"gzip": true
+			"type": "custom",
+			"candid": "https://raw.githubusercontent.com/dfinity/chain-fusion-signer/v0.1.1/src/signer/signer.did",
+			"wasm": "https://github.com/dfinity/chain-fusion-signer/releases/download/v0.1.1/signer.wasm.gz",
+			"init_arg": "(variant { Init = record { ecdsa_key_name = \"key_1\"; ic_root_key_der = null;}})",
```
lmuntaner, line 8: `"key_1"` is only for local development, right? And is `ic_root_key_der` different for local?

src/declarations/signer/signer.did
lmuntaner, line 1: Is this copied over with `dfx generate`?

lmuntaner (approved):
Thanks!!!

Just a couple of comments, but already approving.

dfx.json
bitdivine: key_1 is the prod value.  I was hoping that we could use only prod values here.  I think it's unclear whether that would work so I'll update the PR to use the same keys as before.

.nvmrc
bitdivine: Good catch.  Committed accidentally.

dfx.json
bitdivine: Done.  Please see `scripts/build.signer.sh`

src/declarations/signer/signer.did
bitdivine: I guess so.  Let me check.  It is created by `npm run generate` which is `./scripts/genmerate.sh`.  If I delete the generated files and regenerate, it is created:
```
$ rm -fr src/declarations/signer/
$ mkdir src/declarations/signer/
$ scripts/generate.sh
$ ls src/declarations/signer/signer.did
src/declarations/signer/signer.did
```

Let's check whether just `dfx generate signer` will also create it:
```
$ rm -fr src/declarations/signer
$ dfx generate signer
$ ls src/declarations/signer/
index.d.ts       index.js         signer.did       signer.did.d.ts  signer.did.js    
$ ls src/declarations/signer/signer.did
src/declarations/signer/signer.did
```
So, yes, but the `npm` command then modifies the files created by `dfx generate`.

src/declarations/signer/signer.did
```diff
@@ -124,66 +42,14 @@ type SignRequest = record {
...
-  chain_id : nat64;
-  contract_address : text;
-  symbol : opt text;
-};
-type UserTokenId = record { chain_id : nat64; contract_address : text };
 service : (Arg) -> {
-  add_user_credential : (AddUserCredentialRequest) -> (Result);
-  bulk_up : (blob) -> ();
   caller_btc_address : (BitcoinNetwork) -> (text);
   caller_eth_address : () -> (text);
```
lmuntaner, line 47: Why isn't this did file include `caller_btc_balance`? When did you make the release?

lmuntaner: I saw that the release used is from 2 weeks ago. That answers my question, sorry for the noise!

bitdivine: Two weeks ago.  I can make a new release if wanted?

src/declarations/signer/signer.did
bitdivine: The generation is actually really annoying.  It's very tempting to do just 'dfx deploy && dfx generate'  But in another PR.

src/declarations/signer/signer.did
lmuntaner: Yes, but not for this PR. Let's get this in first.

src/declarations/signer/signer.did
lmuntaner: I can believe that...

src/declarations/signer/signer.did
lmuntaner: Great thanks!

PR #2308, opened 2024-09-10
feat(frontend): pin enabled tokens at top in ManageTokens

AntonioVentilii-DFINITY:
# Motivation

To make it easier, we pin the enabled tokens on top of the manage modal.

# Changes

- Create util function to pin enabled tokens on top.
- Create tests.
- Change type of token list in `ManageTokens` to be toggleable.
- Use new util after normal sorting.

# Tests

### Before

Screenshot 2024-09-10 at 21 29 06

### After

Screenshot 2024-09-10 at 21 28 47

### DIFF
--- a/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
@@ -33,10 +33,11 @@
 	import { i18n } from '$lib/stores/i18n.store';
 	import type { ExchangesData } from '$lib/types/exchange';
 	import type { Token } from '$lib/types/token';
+	import type { TokenToggleable } from '$lib/types/token-toggleable';
 	import { replacePlaceholders } from '$lib/utils/i18n.utils';
 	import { isNullishOrEmpty } from '$lib/utils/input.utils';
 	import { filterTokensForSelectedNetwork } from '$lib/utils/network.utils';
-	import { sortTokens } from '$lib/utils/tokens.utils';
+	import { pinEnabledTokensAtTop, sortTokens } from '$lib/utils/tokens.utils';
 
 	const dispatch = createEventDispatcher();
 
@@ -82,7 +83,7 @@
 	$: manageEthereumTokens = $pseudoNetworkChainFusion || $networkEthereum;
 
 	// TODO: Bitcoin tokens ($enabledBitcoinTokens) are not included yet.
-	let allTokens: Token[] = [];
+	let allTokens: TokenToggleable[] = [];
 	$: allTokens = filterTokensForSelectedNetwork([
 		[
 			{
@@ -99,11 +100,13 @@
 
 	let allTokensSorted: Token[] = [];
 	$: allTokensSorted = nonNullish(exchangesStaticData)
-		? sortTokens({
-				$tokens: allTokens,
-				$exchanges: exchangesStaticData,
-				$tokensToPin: $tokensToPin
-			})
+		? pinEnabledTokensAtTop(
+				sortTokens({
+					$tokens: allTokens,
+					$exchanges: exchangesStaticData,
+					$tokensToPin: $tokensToPin
+				})
+			)
 		: [];
 
 	let filterTokens = '';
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -155,3 +156,17 @@ export const sumMainnetTokensUsdBalancesPerNetwork = ({
...
+/** Pins enabled tokens at the top of the list, preserving the order of the parts.
+ *
+ * @param $tokens - The list of tokens.
+ * @returns The list of tokens with enabled tokens at the top.
+ * */
+export const pinEnabledTokensAtTop = <T extends Token>(
+	$tokens: TokenToggleable<T>[]
+): TokenToggleable<T>[] =>
+	$tokens
+		.reduce<
```
peterpeterparker, line 169: Why not just a `.sort`?

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
```diff
@@ -99,11 +100,13 @@
 
 	let allTokensSorted: Token[] = [];
 	$: allTokensSorted = nonNullish(exchangesStaticData)
-		? sortTokens({
-				$tokens: allTokens,
-				$exchanges: exchangesStaticData,
-				$tokensToPin: $tokensToPin
-			})
+		? pinEnabledTokensAtTop(
```
peterpeterparker, line 103: We are sorting the all list of displayed tokens, will ICP and Ethereum  always be listed first?

AntonioVentilii-DFINITY: yes, I maintained the idea that the ones that we decided to pin always (like ETH, ICP, BTC, etc) are always on top, even if disabled (that is not the case for now, since we force the user to have them enabled, without a choice to disable)

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: too used to use `reduce` eheh! simplified with `sort`

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
peterpeterparker: You might want to ask Artem but, as a user, if the enabled token are pined at the top, I would except the same logic as currently but, with enabled first and disabled then - i.e. the list of enabled tokens following the sorting rules and then the disabled tokens sorted by the rules.

But again, if that's an interesting feedback, better ask Artem.

AntonioVentilii-DFINITY: I asked Artem and he says that for now that is ok. In the future, when and if it is decided that those tokens can be disabled, we can check which decision to take:

- Leave them above the others
- Mix them with the others

So, for now, I would say, this is the simplest solution to have the enabled ones above the disabled.

peterpeterparker: Oki doki. Thanks for double checking.

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -155,3 +156,13 @@ export const sumMainnetTokensUsdBalancesPerNetwork = ({
...
+
+/** Pins enabled tokens at the top of the list, preserving the order of the parts.
+ *
+ * @param $tokens - The list of tokens.
+ * @returns The list of tokens with enabled tokens at the top.
+ * */
+export const pinEnabledTokensAtTop = <T extends Token>(
+	$tokens: TokenToggleable<T>[]
+): TokenToggleable<T>[] =>
+	$tokens.sort((a, b) => (a.enabled === b.enabled ? 0 : a.enabled ? -1 : 1));
```
peterpeterparker, line 168: Just an idea: `.sort(({enabled: a, {enabled: b}}) => Number(b) - Number(a))`

Note: note sure about b - a or a - b ;)

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/utils/tokens.utils.ts
AntonioVentilii-DFINITY: that is correct! changed

PR #2309, opened 2024-09-10
feat(frontend): explicitly label conversions in Ethereum tokens

AntonioVentilii-DFINITY:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/eth/components/transactions/Transaction.svelte
+++ b/src/frontend/src/eth/components/transactions/Transaction.svelte
@@ -2,8 +2,14 @@
 	import { nonNullish } from '@dfinity/utils';
 	import { BigNumber } from '@ethersproject/bignumber';
 	import type { ComponentType } from 'svelte';
+	import type { Erc20Token } from '$eth/types/erc20';
 	import type { EthTransactionType } from '$eth/types/eth-transaction';
+	import { isSupportedEthTokenId } from '$eth/utils/eth.utils';
 	import { isTransactionPending } from '$eth/utils/transactions.utils';
+	import { ckHelpersContractAddresses } from '$icp-eth/derived/cketh.derived';
+	import IconBurn from '$lib/components/icons/IconBurn.svelte';
+	import IconConvert from '$lib/components/icons/IconConvert.svelte';
+	import IconMint from '$lib/components/icons/IconMint.svelte';
 	import IconReceive from '$lib/components/icons/IconReceive.svelte';
 	import IconSend from '$lib/components/icons/IconSend.svelte';
 	import TransactionPending from '$lib/components/transactions/TransactionPending.svelte';
@@ -11,10 +17,12 @@
 	import Card from '$lib/components/ui/Card.svelte';
 	import RoundedIcon from '$lib/components/ui/RoundedIcon.svelte';
 	import { ethAddress } from '$lib/derived/address.derived';
+	import { tokenId, tokenWithFallback } from '$lib/derived/token.derived';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { modalStore } from '$lib/stores/modal.store';
 	import type { Transaction } from '$lib/types/transaction';
 	import { formatSecondsToDate } from '$lib/utils/format.utils';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
 
 	export let transaction: Transaction;
 
@@ -23,19 +31,66 @@
 	let timestamp: number | undefined;
 	let displayTimestamp: number | undefined;
 
-	$: ({ from, value, timestamp, displayTimestamp } = transaction);
-
-	let type: EthTransactionType;
-	$: type = from?.toLowerCase() === $ethAddress?.toLowerCase() ? 'send' : 'receive';
-
-	let icon: ComponentType;
-	$: icon = type === 'send' ? IconSend : IconReceive;
-
-	let amount: BigNumber;
-	$: amount = type == 'send' ? value.mul(BigNumber.from(-1)) : value;
-
 	let pending: boolean;
 	$: pending = isTransactionPending(transaction);
+
+	$: ({ from, to, value, timestamp, displayTimestamp } = transaction);
+
+	let type: EthTransactionType;
+	$: type = $ckHelpersContractAddresses.includes(from.toLowerCase())
+		? 'withdraw'
+		: $ckHelpersContractAddresses.includes(to?.toLowerCase())
+			? 'deposit'
+			: from?.toLowerCase() === $ethAddress?.toLowerCase()
+				? 'send'
+				: 'receive';
+
+	let twinTokenSymbol: string;
+	$: twinTokenSymbol =
+		nonNullish($tokenId) && isSupportedEthTokenId($tokenId)
+			? 'ckETH'
+			: ($tokenWithFallback as Erc20Token).twinTokenSymbol ?? '';
+
+	let label: string;
+	$: label =
+		type === 'withdraw'
+			? replacePlaceholders(
+					pending
+						? $i18n.transaction.label.converting_ck_token
+						: $i18n.transaction.label.ck_token_converted,
+					{
+						$twinToken: $tokenWithFallback.symbol,
+						$ckToken: twinTokenSymbol
+					}
+				)
+			: type === 'deposit'
+				? replacePlaceholders(
+						pending
+							? $i18n.transaction.label.converting_twin_token
+							: $i18n.transaction.label.ck_token_sent,
+						{
+							$token: $tokenWithFallback.symbol,
+							$ckToken: twinTokenSymbol
+						}
+					)
+				: type === 'send'
+					? $i18n.send.text.send
+					: $i18n.receive.text.receive;
+
+	let icon: ComponentType;
+	$: icon =
+		(type === 'withdraw' || type === 'deposit') && pending
+			? IconConvert
+			: type === 'withdraw'
+				? IconMint
+				: type === 'deposit'
+					? IconBurn
+					: type === 'send'
+						? IconSend
+						: IconReceive;
+
+	let amount: BigNumber;
+	$: amount = type === 'send' || type === 'deposit' ? value.mul(BigNumber.from(-1)) : value;
 
 	let transactionDate: number | undefined;
 	$: transactionDate = timestamp ?? displayTimestamp;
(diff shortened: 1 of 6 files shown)


src/frontend/src/icp-eth/derived/cketh.derived.ts
```diff
@@ -117,3 +118,20 @@ export const ckErc20HelperContractAddress: Readable<OptionEthAddress> = derived(
 	([$ckEthMinterInfoStore, $ethereumTokenId]) =>
 		toCkErc20HelperContractAddress($ckEthMinterInfoStore?.[$ethereumTokenId])
 );
+
+/**
+ * The minter address.
+ */
+const ckMinterAddress: Readable<OptionEthAddress> = derived(
```
peterpeterparker, line 125: I'm not convinced we should introduce a new global store for this scoped feature. On one hand, our idea is to deprecate tokenWillFallback; on the other, adding more global stores can impact performance. So unless this is truly used globally and not just in one or two functions, I don't think it should be a global store. WDYT?

peterpeterparker (commented):
Just one feedback, did not reviewed the PR yet.

src/frontend/src/icp-eth/derived/cketh.derived.ts
AntonioVentilii-DFINITY: It is not using tokenWillFallback (or a derived), I think. But in any case, I understand what you mean with not adding more stores. Please check last commit, I moved the store logic inside the component.

src/frontend/src/eth/components/transactions/Transaction.svelte
```diff
@@ -23,22 +38,84 @@
 	let timestamp: number | undefined;
 	let displayTimestamp: number | undefined;
 
+	let ckMinterInfoAddresses: (EthAddress | undefined)[];
+	$: $ckEthMinterInfoStore,
+		$ethereumToken,
+		(() => {
+			const ckMinterInfoData = $ckEthMinterInfoStore?.[$ethereumToken.id];
```
peterpeterparker, line 45: This feels like it doesn't belong here.

- We do not want to map and iterate over the minter addresses on each transaction render for performance reasons.

- I understand now why you primarily created a global store, but even with such a store, we would still need to map here.

- However, the component should remain as naive as possible, and we are doing the opposite by adding logic here. "It's just rendering".

Spontaneously, I feel like we need something similar to IC side; we need some kind of mapper.

Basically:

sortedTransactions + ckMinterAddress.recude => map in a utils = upgraded list => <Transaction />

Do you know what I mean? What do you think?

AntonioVentilii-DFINITY: Ah, you mean to add a way to add the prop `type` directly to the transactions list?

peterpeterparker: - Either the transaction component calls a map or the all combined list goes through a mapper.

- `ckMinterAddress.recude` should be performed only once.

AntonioVentilii-DFINITY: Maybe have a look at last commit, is that it?

peterpeterparker: Yeah exactly, I think it's better. That way the component is naive and the heavy load is done has few times as possible without have global performance impact.

Can you add tests for the utils?

AntonioVentilii-DFINITY: Yes, makes sense!

I was waiting to create tests, until I confirmed this was the change

peterpeterparker: mmmm you have broke something, a bunch of code you removed previously is back in the component

peterpeterparker: nvm maybe I oversee something. let me have another look

src/frontend/src/eth/components/transactions/Transaction.svelte
```diff
@@ -2,43 +2,84 @@
...
 
-	$: ({ from, value, timestamp, displayTimestamp } = transaction);
+	$: ({ value, timestamp, displayTimestamp, uiType: type } = transaction);
 
-	let type: EthTransactionType;
-	$: type = from?.toLowerCase() === $ethAddress?.toLowerCase() ? 'send' : 'receive';
+	let ckTokenSymbol: string;
+	$: ckTokenSymbol =
+		nonNullish($tokenId) && isSupportedEthTokenId($tokenId)
+			? `ck${$tokenSymbol}`
```
peterpeterparker, line 40: Not sure why this `ck${$tokenSymbol}` mapping, can you briefly explain what information we are missing?

src/frontend/src/eth/components/transactions/Transaction.svelte
```diff
@@ -2,43 +2,84 @@
...
+			: ($token as Erc20Token).twinTokenSymbol ?? '';
```
peterpeterparker, line 41: If `$tokenId` can potentially be nullish, couldn't `$token` be also potentially nullish?

src/frontend/src/eth/components/transactions/Transaction.svelte
```diff
@@ -2,43 +2,84 @@
...
+					: $i18n.receive.text.receive;
 
 	let icon: ComponentType;
-	$: icon = type === 'send' ? IconSend : IconReceive;
+	$: icon =
+		(type === 'withdraw' || type === 'deposit') && pending
+			? IconConvert
+			: type === 'withdraw'
+				? IconMint
+				: type === 'deposit'
```
peterpeterparker, line 75: It's correct, deposit = burn?

src/frontend/src/eth/components/transactions/Transaction.svelte
```diff
@@ -2,43 +2,84 @@
...
+					? IconBurn
+					: type === 'send'
+						? IconSend
+						: IconReceive;
 
 	let amount: BigNumber;
-	$: amount = type == 'send' ? value.mul(BigNumber.from(-1)) : value;
```
peterpeterparker, line 41: Oopsie

src/frontend/src/eth/components/transactions/Transaction.svelte
peterpeterparker: Added few minor questions as I had another look. Hope you don't mind. But this is good, thanks a ton!

src/frontend/src/eth/components/transactions/Transaction.svelte
AntonioVentilii-DFINITY: yes, they represent the same "scope" in the ck-conversions, but I am using the naming of the [documentation](https://github.com/dfinity/ic/blob/master/rs/ethereum/cketh/docs/cketh.adoc) for the transaction types. For the icons, I was planning already to change the name of the icons in another PR, so that we call them in another way, representing this

src/frontend/src/eth/components/transactions/Transaction.svelte
AntonioVentilii-DFINITY: yes, that `nonNullish` is just because `isSupportedEthTokenId` does not accept `undefined`... what would you suggest?

src/frontend/src/eth/components/transactions/Transaction.svelte
peterpeterparker: I think there is something, like not a link but there is already a piece of code somewhere that does something similar....

peterpeterparker: I don't remember where so not a blocker but, ....

src/frontend/src/eth/components/transactions/Transaction.svelte
peterpeterparker: If the `$token` is potentially `undefined` then the following cast is maybe more accurate:

```
($token as Erc20Token | undefined)?.twinTokenSymbol ?? '';
```

peterpeterparker: But if it cannot be undefined, it's fine.

src/frontend/src/eth/components/transactions/Transaction.svelte
peterpeterparker: Thanks for the explanation.

src/frontend/src/eth/components/transactions/Transaction.svelte
AntonioVentilii-DFINITY: I will use your suggestion, but I add a TODO about changing it when `$token` will become something "better" eheh

src/frontend/src/eth/components/transactions/Transaction.svelte
AntonioVentilii-DFINITY: I searched and I cannot find anything, do you have some clue on what to search? Or maybe about which flow?

P.S. the alternative would be to make an array of ckETH data (prod, staging and local) and use `.find` to compare against their `twinToken` property. That would be the completly correct way

AntonioVentilii-DFINITY:
```typescript
// PSEUDO CODE

export const CKETH_DATA: IcCkInterface[] = [
	...(nonNullish(CKETH_LOCAL_DATA) ? [CKETH_LOCAL_DATA] : []),
	...(nonNullish(CKETH_STAGING_DATA) ? [CKETH_STAGING_DATA] : []),
	...(nonNullish(CKETH_IC_DATA) ? [CKETH_IC_DATA] : [])
];

CKETH_DATA.find(({ twinToken: { id } }) => id === $tokenId)?.symbol;
```

peterpeterparker: Ah, couldn't we use `twinTokenSymbol` or `Erc20Contract`?

AntonioVentilii-DFINITY: for ETH and SepoliaETH we don't have the property `twinTokenSymbol`, we never assign it. Shall I add it (in another PR)?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/eth/components/transactions/Transaction.svelte
AntonioVentilii-DFINITY: Changed icons name with https://github.com/dfinity/oisy-wallet/pull/2347

src/frontend/src/eth/components/transactions/Transaction.svelte
AntonioVentilii-DFINITY: Doing with PRs:

- https://github.com/dfinity/oisy-wallet/pull/2348
- https://github.com/dfinity/oisy-wallet/pull/2350
-

PR #2310, opened 2024-09-10
refactor(frontend): extract label in transaction component

AntonioVentilii-DFINITY:
# Motivation

For better readability, we extract the logic of the label in the `Transaction` component. That will come useful for a future PR.

For the same reason, unrelated, we move the `pending` variable definition above.


peterpeterparker (approved):
LGTM, thx

PR #2311, opened 2024-09-10
feat(frontend): store with ck twin token for Ethereum tokens

AntonioVentilii-DFINITY:
# Motivation

It will come useful to have a derived store that provides the possible ck-twin-token for the current selected token.


peterpeterparker (commented):
The ultimate goal is to remove the deprecated `tokenWithFallback` so, I'm not sure we should add more derived store at the top. WDYT?

AntonioVentilii-DFINITY:
Yes, agreed! But shall I use `token` store instead?

peterpeterparker:
If you are agree, then I would say we don't need this PR. If this PR was meant to avoid code duplication for future PR, can you describe which features?

AntonioVentilii-DFINITY:
Let me revisit then, closing for now

PR #2312, opened 2024-09-10
chore(frontend): remove unnecessary undefined condition

AntonioVentilii-DFINITY:
# Motivation

According to the type, it is unnecessary to fallback `ckEthereumTwinTokenNetwork` since the property `network` of type `Token` cannot be `undefined`.


peterpeterparker (approved):
LGTM, thx

PR #2314, opened 2024-09-10
feat(frontend): stores for ck Ethereum native token helpers

AntonioVentilii-DFINITY:
# Motivation

To reduce repetition of code, we create derived stores for the helpers of `ckEthereumNativeToken`.

### DIFF
--- a/src/frontend/src/icp-eth/derived/cketh.derived.ts
+++ b/src/frontend/src/icp-eth/derived/cketh.derived.ts
@@ -93,9 +93,36 @@ export const ckEthereumNativeTokenId: Readable<TokenId> = derived(
 	([{ id }]) => id
 );
 
+export const ckEthereumNativeTokenNetworkId: Readable<NetworkId> = derived(
+	[ckEthereumNativeToken],
+	([
+		{
+			network: { id }
+		}
+	]) => id
+);
+
 export const ckEthereumNativeTokenBalance: Readable<BigNumber | undefined | null> = derived(
 	[balancesStore, ckEthereumNativeToken],
 	([$balanceStore, { id }]) => $balanceStore?.[id]?.data
+);
+
+/**
+ * The contract helper used to convert ETH -> ckETH for ckEthereumNativeToken.
+ */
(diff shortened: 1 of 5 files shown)


src/frontend/src/icp-eth/derived/cketh.derived.ts
```diff
@@ -93,11 +93,38 @@ export const ckEthereumNativeTokenId: Readable<TokenId> = derived(
...
+
 export const ckEthereumNativeTokenBalance: Readable<BigNumber | undefined | null> = derived(
 	[balancesStore, ckEthereumNativeToken],
 	([$balanceStore, { id }]) => $balanceStore?.[id]?.data
 );
 
+/**
+ * The contract helper used to convert ETH -> ckETH for ckEthereumNativeToken.
+ */
+export const ckEthereumNativeTokenEthHelperContractAddress: Readable<OptionEthAddress> = derived(
```
peterpeterparker, line 113: I'm not convinced by the need of having a global store for those two addresses. Their usages is scoped within two particular functions moreover it feels like a particular logic rather than something that have to be derived. Is there a will to reuse the information elsewhere?

If not, should can maybe use a component or function to reduce code duplication instead?

AntonioVentilii-DFINITY: Sorry, forgot to mention, that this was a split for https://github.com/dfinity/oisy-wallet/pull/2309

But maybe I can submit that to review and check directly there

I will move to draft this one

peterpeterparker: Mmm still not convinced I have to say. Given that we want to move away of having a "global token", I'm a bit reluctant to adding new global derived store that dervies from it. In addition, the issue you are linking seems to be a scoped feature.

Can you list the use cases for having those new global stores?

AntonioVentilii-DFINITY: actually it is deprecated: in the PR I mentioned I changed the code a little bit, so this is useless for now

PR #2315, opened 2024-09-11
feat(frontend): save selected network in local storage

DenysKarmazynDFINITY:
# Motivation

The goal is to save selected network in local storage and update related URL param if it's needed. 

A few use-cases when it's gonna be used:
1. User selects a network -> user navigates to a token transactions page -> user refreshes the browser tab -> user goes back to the home page -> pre-selected network from local storage will be used (previously: network selection was always reset to Chain Fusion).
2. User selects a network -> user closes the browser tab -> user comes back to Oisy -> pre-selected network from local storage will be used (previously: Chain Fusion was always selected as the default option).

Same as with other `settings` from local storage, the selected network value will be reset on logout.

The UX aspect of these changes is discussed and agreed with Artem.

### DIFF
--- a/src/frontend/src/lib/components/networks/NetworkButton.svelte
+++ b/src/frontend/src/lib/components/networks/NetworkButton.svelte
@@ -9,6 +9,7 @@
 	import type { NetworkId } from '$lib/types/network';
 	import { formatUSD } from '$lib/utils/format.utils';
 	import { gotoReplaceRoot, isRouteTransactions, switchNetwork } from '$lib/utils/nav.utils';
+	import { saveSelectedNetwork } from '$lib/utils/network.utils';
 
 	export let id: NetworkId | undefined;
 	export let name: string;
@@ -18,6 +19,7 @@
 	const dispatch = createEventDispatcher();
 
 	const onClick = async () => {
+		saveSelectedNetwork(id);
 		await switchNetwork(id);
 
 		if (isRouteTransactions($page)) {
--- a/src/frontend/src/lib/services/auth.services.ts
+++ b/src/frontend/src/lib/services/auth.services.ts
@@ -8,7 +8,7 @@ import { trackEvent } from '$lib/services/analytics.services';
 import { authStore, type AuthSignInParams } from '$lib/stores/auth.store';
 import { busy } from '$lib/stores/busy.store';
 import { i18n } from '$lib/stores/i18n.store';
-import { testnetsStore } from '$lib/stores/settings.store';
+import { selectedNetworkStore, testnetsStore } from '$lib/stores/settings.store';
 import { toastsClean, toastsError, toastsShow } from '$lib/stores/toasts.store';
 import type { ToastMsg } from '$lib/types/toast';
 import { replaceHistory } from '$lib/utils/route.utils';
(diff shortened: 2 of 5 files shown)


peterpeterparker (commented):
Unsolicited feedback: What is the reasoning behind this change? Can it be challenged?

Personally, I think we should not save the selected network anywhere other than specifying it through the URL. Otherwise, this could introduce some weird behavior, particularly on mobile devices if the dapp is pinned.

DenysKarmazynDFINITY:
@peterpeterparker the scenario 1 from the PR description was reported as a bug (see the screen recording below). I think it still can be challenges as according to Artem (who reported it) he doesn’t have a very strong opinion here. On desktop I didn't experience issues while testing it, but if you foresee unexpected behaviour, I can update Artem on our decision.

https://github.com/user-attachments/assets/2c71c2d2-49f4-4ea9-a4f3-4e9e5bffe4fa

peterpeterparker:
Thanks for the video. Then why not fixing the bug?

The issue seems to be the query parameters that goes lost when clicking 'Back'. I don’t see why it would require the introduction of some storage value?

DenysKarmazynDFINITY:
The problem from the video is that user refreshes the browser tab before clicking the "back" button. [According to the `back` function implementation](https://github.com/dfinity/oisy-wallet/blob/main/src/frontend/src/lib/utils/nav.utils.ts#L38), it will just go to `/` as `pop` param will be always `false` after the page refresh. If we remove the condition and always use `history.back();` it won't work as expected, as we cannot guarantee that user will stay inside Oisy after clicking the "back" button. Having said that, I'm not sure if I can achieve the wanted result without persisting previously selected network (or previous URL with query params) via local storage.

Instead of doing it on the page component level, I can reduce the usage of the network value from local storage only to the `back" method (pseudocode: `await goto(savedValue ? '/?network={savedValue}' : '/');`. WDYT?

AntonioVentilii-DFINITY:
If I can give my opinion: if that issue happens when the user refreshes, then I think it is correct behavior that we just go back to Chain Fusion.

It is like you just copy the entire address for the transaction page, you open in a new tab and then click `back`: you cannot expect to go back to the original network you had before.

If we treat the network switcher as a "filter" feature, we should expect that if there is "no history", the filter is reset

peterpeterparker:
```
const {location: {seach}} = window;
await goto('/${search ?? ""}');
```

DenysKarmazynDFINITY:
The token transaction page always has `network` as a query param. It means, even if user had Chain Fusion option selected previously, we would overwrite it with the viewed token network (e.g. if it was transaction page of Ethereum token, the selected network would become ETH).

DenysKarmazynDFINITY:
Alright, just to come to a conclusion here, I suggest we call it as "expected behaviour" that users loose browser history (and hence selected network query param) after they refresh the browser tab. Instead, we can investigate why the transaction are loading slowly/not loading which forces users to refreshing Oisy app.

👍 / 👎 ?

AntonioVentilii-DFINITY:
I agree with @DenysKarmazynDFINITY  (I mentioned in the Jira ticket too)

About the transactions not loading, I have a ticket open that I am working on

PR #2318, opened 2024-09-11
fix(frontend): BTC testnet addresss was not loaded

lmuntaner:
# Motivation

Render the BTC test addresses.

There was a bug in when the address was loaded. It was loaded only when it was present, which doesn't make sense. It should be loaded when not present.

# Changes

* Change the check to load the address when it's not in the store.

# Tests

Tested that it's loaded properly now.

### DIFF
--- a/src/frontend/src/lib/components/core/Loader.svelte
+++ b/src/frontend/src/lib/components/core/Loader.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { Modal, type ProgressStep } from '@dfinity/gix-components';
-	import { debounce, nonNullish } from '@dfinity/utils';
+	import { debounce, isNullish } from '@dfinity/utils';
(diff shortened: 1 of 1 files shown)


lmuntaner:

AntonioVentilii-DFINITY (approved):
Good catch!

src/frontend/src/lib/components/core/Loader.svelte
```diff
@@ -71,7 +71,7 @@
 	const debounceLoadBtcAddressTestnet = debounce(loadBtcAddressTestnet);
 
 	$: {
-		if ($testnets && nonNullish($btcAddressTestnet)) {
+		if ($testnets && isNullish($btcAddressTestnet)) {
```
peterpeterparker, line 74: Oh I missed it! Good catch

PR #2319, opened 2024-09-11
feat(frontend): Support no address in ReceiveAddressWithLogo

lmuntaner:
# Motivation

We want to show a loading indicator when an address is being loaded.

In thir PR, `ReceiveAddressWithLogo` accepts `undefined | null` in `address` prop and shows a skeleton in that case.

# Changes

* Changes in `ReceiveAddressWithLogo` to accept no address and show skeleton.
* Do not pass empty string in ReceiveAddresses for bitcoin address.

# Tests

Tested locally.


lmuntaner:
New UX:

Screenshot 2024-09-11 at 14 54 24

AntonioVentilii-DFINITY:
@lmuntaner I was checking the code and it seems that undefined is when it is loading and null when there is some sort of error mid-time, or not? In any case, you want to achieve the same behavior, correct?

AntonioVentilii-DFINITY (approved):
Thank you for the explanation! It makes sense that the error is treated elsewhere if the address is not loaded for some reason

PR #2321, opened 2024-09-11
chore(signer): Bump signer release

bitdivine:
# Motivation
The signer has a new method "caller_btc_balance" that we would like to use.

# Changes
* Update the downloaded signer version from 0.1.1 to 0.1.2.

Note: This is NOT available in production yet, but is expected to be soon.

# Tests
See CI


bitdivine:
@lmuntaner Corerct, with `npm run generate`.

PR #2322, opened 2024-09-11
feat(frontend): Add BTC to manage tokens

lmuntaner:
# Motivation

Add BTC to ManageTokens modal.

# Changes

* New component for toggle of BTC tokens. It's a dummy component which renders a disabled toggle.
* Btc token util to identify bitcoin tokens.
* Add `enabledBitcoinTokens` in the list of tokens in ManageToken and manage the scenario in the toggle for bitcoin tokens.

# Tests

Tested locally.

### DIFF
--- a/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
@@ -85,7 +85,6 @@
 	let manageEthereumTokens = false;
 	$: manageEthereumTokens = $pseudoNetworkChainFusion || $networkEthereum;
 
-	// TODO: Bitcoin tokens ($enabledBitcoinTokens) are not included yet.
 	let allTokens: TokenToggleable<Token>[] = [];
 	$: allTokens = filterTokensForSelectedNetwork([
 		[
--- a/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
+++ b/src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
@@ -38,6 +38,9 @@
 	import { isNullishOrEmpty } from '$lib/utils/input.utils';
 	import { filterTokensForSelectedNetwork } from '$lib/utils/network.utils';
 	import { pinEnabledTokensAtTop, sortTokens } from '$lib/utils/tokens.utils';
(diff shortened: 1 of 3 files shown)


lmuntaner:
UI:

Chain fusion network:
Screenshot 2024-09-11 at 15 46 32

Btc testnet network:
Screenshot 2024-09-11 at 15 46 41

Btc regtest network:
Screenshot 2024-09-11 at 15 46 49

AntonioVentilii-DFINITY (approved):
LGTM!

P.S. may you please remove the TODO comment right above the list?

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
```diff
@@ -3,6 +3,9 @@
 	import { debounce, nonNullish } from '@dfinity/utils';
 	import { createEventDispatcher, onMount } from 'svelte';
 	import { fade } from 'svelte/transition';
+	import BtcManageTokenToggle from '$btc/components/BtcManageTokenToggle.svelte';
```
peterpeterparker, line 6: I'm not sure anything finding place in `$icp-eth` should reference `$btc` or we should rename the package.

I tend to think that we need a new package or this should find place in `lib`. Not sure what's best.

lmuntaner:
> LGTM!
> 
> P.S. may you please remove the TODO comment right above the list?

Good catch! I removed it but with rebasing it sneaked in again.

src/frontend/src/icp-eth/components/tokens/ManageTokens.svelte
lmuntaner: That's a good point, I also saw some other files within `$icp-eth` that seem to belong more to `lib`.

I believe that the idea behind `icp-eth` was about what ICP and ETH share because those where the only two networks. But with Bitcoin, most of the stuff here is susceptible to be reused.

Maybe rename to `icp-eth-btc`? Or `shared`? Or move everything to `lib`. I don't think there will be many things that are only ICP or ETH...

peterpeterparker: Let's move it to `$lib`, at least for now. https://github.com/dfinity/oisy-wallet/pull/2332

PR #2324, opened 2024-09-11
chore(frontend): remove unused bubble event in IcSend

AntonioVentilii-DFINITY:
# Motivation

I checked and the `nnsClose` event is not used in the parent components where `IcSend` is used, so we remove it.


peterpeterparker (approved):
LGTM, thx

PR #2325, opened 2024-09-11
feat(frontend): auto-enable ckERC20 tokens when converting from ERC20

AntonioVentilii-DFINITY:
# Motivation

Same as PR #2292 , when the user converts ERC20 to ckERC20, it is useful that the ckERC20 token is enabled if it is not. We use the same flow as the one used for `ckERC20 --> ERC20`: when the user click the Convert button if the token is not enabled, it will be.

# Tests

In local replica, the following test worked:

- Disable ckSepoliaUSD.
- Convert Sepolia USDC to ckSepoliaUSD.
- Check if ckSepoliaUSD was enabled.

https://github.com/user-attachments/assets/fbf58751-2533-4f8b-b6e6-22fdbaa5a055


src/frontend/src/icp-eth/services/custom-token.services.ts
```diff
@@ -0,0 +1,107 @@
...
+ * As a result, the UI might lack necessary information.
+ * Therefore, this function aims to enable the CK token if the user has only enabled the counterpart token.
+ *
+ * @param {Object} params - The parameters for the function.
+ * @param {Erc20UserToken[]} params.icrcCustomTokens - The list of user's ICRC tokens.
+ * @param {Token} params.sendToken - The token to be sent.
+ * @param {OptionIdentity} params.identity - The user's identity.
+ * @returns {Promise<{ result: 'loaded' | 'skipped' | 'error' }>} The result of the operation.
+ */
+export const autoLoadCustomToken = async ({
```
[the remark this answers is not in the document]
peterpeterparker: Is there some potential of refactoring to avoid logic duplication?

src/frontend/src/icp-eth/services/custom-token.services.ts
```diff
@@ -0,0 +1,107 @@
...
+	identity
+}: {
+	icrcCustomTokens: IcrcCustomToken[];
+	sendToken: Token;
+	identity: OptionIdentity;
+}): Promise<{ result: 'loaded' | 'skipped' | 'error' }> => {
+	if (sendToken.standard !== 'erc20') {
+		return { result: 'skipped' };
+	}
+
```
peterpeterparker, line 39: In the other loader, here a test is finding place:

```
const twinToken = (sendToken as OptionIcCkToken)?.twinToken;

	if (twinToken?.standard !== 'erc20') {
		return { result: 'skipped' };
	}

```

Shouldn't we mirror it for icrc?

peterpeterparker (approved):
LGTM, thanks

Minor questions

src/frontend/src/icp-eth/services/custom-token.services.ts
AntonioVentilii-DFINITY: Not really, sure, I need your opinion:

We are not getting directly the twin token (like for ERC20), but the twin token symbol. Then we search for the ck-token among ICRC tokens. So, theoretically, they will always have the correct standard, no? Or shall I check either way for good measure?

AntonioVentilii-DFINITY: Better to put it... included

src/frontend/src/icp-eth/services/custom-token.services.ts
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Done with PRs:

- #2338 
- #2339 
- #2340

PR #2328, opened 2024-09-11
refactor(frontend): common type for optional-nullable CanisterId

AntonioVentilii-DFINITY:
# Motivation

To avoid repetition of code, we create a common type for optional-nullable type `CanisterIdText`.

### DIFF
--- a/src/frontend/src/lib/types/canister.ts
+++ b/src/frontend/src/lib/types/canister.ts
@@ -1,1 +1,3 @@
 export type CanisterIdText = string;
+
+export type OptionalNullableCanisterIdText = CanisterIdText | null | undefined;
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/types/canister.ts
```diff
@@ -1 +1,3 @@
 export type CanisterIdText = string;
+
+export type OptionalNullableCanisterIdText = CanisterIdText | null | undefined;
```
peterpeterparker, line 3: `OptionCanisterIdText`

AntonioVentilii-DFINITY: Changed

peterpeterparker (approved):
LGTM, thx

PR #2329, opened 2024-09-11
refactor(frontend): common type for balance as BigNumber

AntonioVentilii-DFINITY:
# Motivation

To avoid repetition of code, we create a common type for balance when is defined as `BigNumber`.

### DIFF
--- a/src/frontend/src/lib/types/balance.ts
+++ b/src/frontend/src/lib/types/balance.ts
@@ -2,6 +2,4 @@ import { BigNumber } from '@ethersproject/bignumber';
 
 export type Balance = BigNumber;
 
-export type NullableBalance = Balance | null;
-
-export type OptionalNullableBalance = NullableBalance | undefined;
+export type OptionBalance = Balance | null | undefined;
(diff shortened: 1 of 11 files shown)


src/frontend/src/lib/types/balance.ts
```diff
@@ -0,0 +1,7 @@
+import { BigNumber } from '@ethersproject/bignumber';
+
+export type Balance = BigNumber;
+
+export type NullableBalance = Balance | null;
+
+export type OptionalNullableBalance = NullableBalance | undefined;
```
peterpeterparker, line 7: Have you try to have a single type `export type OptionBalance = Balance | null | undefined`? How does that look?

A bit an opiniated feedback but I needed few reads to process types like `Exclude<OptionalNullableBalance, null>` I have to say. It's early morning. 😆

AntonioVentilii-DFINITY: Ahahahahah have a look at last commit, I removed the `NullableBalance` type

peterpeterparker (approved):
LGTM, thx

PR #2330, opened 2024-09-11
refactor(frontend): common type for optional-nullable string

AntonioVentilii-DFINITY:
# Motivation

To avoid repetition of code, we create a common type for optional-nullable type `string`.

### DIFF
--- a/src/frontend/src/lib/types/string.ts
+++ b/src/frontend/src/lib/types/string.ts
@@ -1,0 +1,1 @@
+export type OptionalNullableString = string | null | undefined;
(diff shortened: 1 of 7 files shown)


src/frontend/src/lib/types/string.ts
```diff
@@ -0,0 +1 @@
+export type OptionalNullableString = string | null | undefined;
```
peterpeterparker, line 1: `OptionString` ?

AntonioVentilii-DFINITY: Changed!

peterpeterparker (approved):
LGTM, thx

PR #2332, opened 2024-09-12
refactor(frontend): move manage tokens to $lib

peterpeterparker:
# Motivation

"Manage tokens" related components has started using BTC therefore it cannot remain located under `$icp-eth`. I don't really have a better idea so let's move it to `$lib`.


AntonioVentilii-DFINITY (approved):
LGTM! 

My opinion: I think it is correct to move it there, since ultimately, it will treat all the tokens that we onboard

AntonioVentilii-DFINITY:
@peterpeterparker renamed the title as per convention

PR #2333, opened 2024-09-12
fix(devops): Fix shell quoting in signer args file

bitdivine:
# Motivation
The command that creates the signer args file leaves unintended characters in the argument file.  The code was moved from a shell variable to a heredoc, thereby introducing these errors.

# Changes
- Remove quotes and escapes from the signer args heredoc.

# Tests
Running `scripts/build.signer.sh` now creates a valid candid file.


DenysKarmazynDFINITY (approved):
LGTM, thanks!

PR #2334, opened 2024-09-12
fix(devops): Deploying the signer without env was failing

lmuntaner:
# Motivation

Running `npm run deploy` wasn't installing the signer canister.

The reason was that when using `--network` you can't pass an empty string.

# Changes

* Add `local` as default value when deploying the signer.

# Tests

Tested locally.


lmuntaner:
@DenysKarmazynDFINITY @bitdivine please review

DenysKarmazynDFINITY (approved):

PR #2335, opened 2024-09-12
fix(frontend): Fix local bitcoin node setup

lmuntaner:
# Motivation

New minted blocks were not adding balance to the address.

The reason seemed to be a config parameter when setting up the local node.

# Changes

* Remove the parameter (it was there to reduce the amount of memory, which is a premature optimization).

# Tests

Tested locally.


lmuntaner:
@DenysKarmazynDFINITY please review

DenysKarmazynDFINITY (approved):
thanks!

PR #2338, opened 2024-09-12
refactor(frontend): mapping util to convert to UserToken

AntonioVentilii-DFINITY:
# Motivation

To avoid repetition of code, we create a function to convert a `SaveUserToken` type to `UserToken`.


peterpeterparker (approved):
LGTM, thx

PR #2339, opened 2024-09-12
refactor(frontend): mapping util to convert to CustomToken

AntonioVentilii-DFINITY:
# Motivation

To avoid repetition of code, we create a function to convert a `SaveCustomToken` type to `CustomToken`.


peterpeterparker (approved):
LGTM, thx

PR #2340, opened 2024-09-12
refactor(frontend): common service to auto-load token

AntonioVentilii-DFINITY:
# Motivation

The functions that auto-load tokens (both for ERC20 and ICRC) seem to have the same patterns. This PR aims on reducing code repetition creating a common function to be re-used
A few logic changes happens to achieve that:
- When the common function finds the counter-token, it will not check the `standard`, since it is implied that the find function will return the token that it is proper.
- We provide the sendToken directly casted as the expected type. That is not necessary an issue, since the first check is specific on the `standard` of such a token.


src/frontend/src/icp-eth/services/custom-token.services.ts
```diff
@@ -1,25 +1,35 @@
...
+
 /**
  * When a user converts a ckERC20 token to an ERC20 twin token, the UI needs information about the counterpart token (ckERC20).
  * For example, the user might have an ERC20 token (e.g., USDC) enabled, but its token counterpart (e.g., ckUSDC) disabled.
  * As a result, the UI might lack necessary information.
  * Therefore, this function aims to enable the CK token if the user has only enabled the counterpart token.
  *
  * @param {Object} params - The parameters for the function.
- * @param {Erc20UserToken[]} params.icrcCustomTokens - The list of user's ICRC tokens.
+ * @param {IcrcCustomToken[]} params.icrcCustomTokens - The list of user's ICRC tokens.
```
AntonioVentilii-DFINITY, line 32: this is just a small correction that I fix together here

src/frontend/src/lib/i18n/en.json
```diff
@@ -118,7 +118,7 @@
 			"icrc_canisters": "Error while loading the ICRC canisters.",
 			"erc20_user_tokens": "Error while loading the ERC20 user tokens.",
 			"erc20_user_token": "Error while enabling the ERC20 twin token.",
-			"icrc_custom_tokens": "Error while loading the ICRC custom tokens.",
+			"icrc_custom_token": "Error while loading the ICRC custom token.",
```
AntonioVentilii-DFINITY, line 121: this is just a small correction that I fix together here

peterpeterparker (approved):
LGTM, thx

PR #2342, opened 2024-09-12
style(frontend): skeleton text for loading exchange value

AntonioVentilii-DFINITY:
# Motivation

Similar to the balance, we want to show a skeleton for the exchange value if there is no balance loader started.

Screenshot 2024-09-12 at 23 43 22


peterpeterparker (approved):
LGTM, thx

PR #2343, opened 2024-09-12
feat(lint): add rule prefer-reduce-type-parameter

AntonioVentilii-DFINITY:
# Motivation

We include the rule [prefer-reduce-type-parameter](https://typescript-eslint.io/rules/prefer-reduce-type-parameter) for `.reduce`,  and we fix issues found.


peterpeterparker (approved):
Thx

PR #2346, opened 2024-09-13
feat(frontend): listen to ETH mined transactions instead of pending ones [GIX-2909]

AntonioVentilii-DFINITY:
# Motivation

Listening to pending transactions will only provide pending transactions in the mempool in a prompt manner. That is a limitative since we would like to display the user's ETH (and SepoliaETH) transactions as soon as they are not-pending (excluding non-approved transactions).

This PR changes the transaction listener to listen to mined transactions instead of pending ones.

Furthermore, we add a `slide` transition for the list of transactions.


src/frontend/src/eth/components/transactions/Transactions.svelte
```diff
@@ -83,8 +84,10 @@
 <Header>{$i18n.transactions.text.title}</Header>
 
 <TransactionsSkeletons>
-	{#each sortedTransactionsUi as transaction, index (`${transaction.hash}-${index}`)}
-		<Transaction {transaction} />
+	{#each sortedTransactionsUi as transaction (transaction.hash)}
+		<div transition:slide={{ duration: 250 }}>
```
peterpeterparker, line 88: That's not related to the PR description?

src/frontend/src/eth/services/eth-listener.services.ts
```diff
@@ -22,11 +19,11 @@ export const initTransactionsListener = ({
 	address: EthAddress;
 }): WebSocketListener => {
 	if (isSupportedEthTokenId(token.id)) {
-		return initEthPendingTransactionsListenerProvider({
+		return initMinedTransactionsListenerProvider({
```
peterpeterparker, line 22: By doing so, aren't we removing the display of pending transaction that are not yet mined?

peterpeterparker (approved):
LGTM, thx

PR #2348, opened 2024-09-13
refactor(frontend): use common type for RequiredToken

AntonioVentilii-DFINITY:
# Motivation

To amass all the necessary logic and avoid repetition of code, we change type `RequiredToken` to be re-used instead of `RequiredExcept` util.


peterpeterparker (approved):
LGTM, thx

PR #2349, opened 2024-09-13
refactor(frontend): common zod type for IC tokens in env folder

AntonioVentilii-DFINITY:
# Motivation

We make a new `zod` object with base data for IC tokens and we move `envTokenSymbol` to another module, to be re-used.


peterpeterparker (approved):
LGTM, thx

PR #2350, opened 2024-09-13
feat(frontend): add twinTokenSymbol to ETH native tokens

AntonioVentilii-DFINITY:
# Motivation

We want to add the prop `twinTokenSymbol` to ETH, mainnet and testnet. So, we create a specific type for this kind of tokens, moving such property closer to type `Token`.


peterpeterparker (commented):
1. I'm a bit afraid that the interface named `TwinToken` lead to some confusion with reading the actual `twinToken` on IC side. Do you have another naming id or should we add it to `TokenAppearance`?

2. On the `IcToken` we should maybe enforce excluding this information?

```
export type IcToken = Omit<Token, "twinTokenSymbol"> & IcFee & IcInterface; 
```

AntonioVentilii-DFINITY:
1. Maybe `TokenLinkedData` or `TokenLinkedAssets` ? We already have similar `IcCkLinkedAssets` for ck-token

2. Right! It makes sense, changed!

peterpeterparker:
Sounds good.

For 1., let's go for `TokenLinkedData` instead of `TwinToken`.

peterpeterparker (approved):
LGTM, thx

PR #2351, opened 2024-09-13
refactor(frontend): ICRC_TOKENS are generic IcInterface tokens [GIX-2928]

AntonioVentilii-DFINITY:
# Motivation

We would like to add more IC tokens to the list `ICRC_TOKENS`, so we redefine them as IcInterface.

To do that, we need to change the util `mapCkTokenOisyName` too, to accept tokens that may possibly be not ck-tokens.


src/frontend/src/icp/utils/icrc.utils.ts
```diff
@@ -135,12 +135,17 @@ export const buildIcrcCustomTokenMetadataPseudoResponse = ({
 export const icTokenIcrcCustomToken = (token: Partial<IcrcCustomToken>): token is IcrcCustomToken =>
 	(token.standard === 'icp' || token.standard === 'icrc') && 'enabled' in token;
 
-export const mapCkTokenOisyName = (token: IcCkInterface): IcCkInterface => ({
+const icCkToken = (token: IcInterface): token is IcCkInterface =>
+	'minterCanisterId' in token && 'twinToken' in token;
+
+export const mapCkTokenOisyName = (token: IcInterface): IcInterface => ({
```
peterpeterparker, line 141: Should we rename the function as well? i.e. remove the "Ck"?

Or if it's still used for `IcCkInterface`, whouldn't it be still using those interfaces?

AntonioVentilii-DFINITY: no, it makes sense to change the name, otherwise i should do something like `ifCkToken ? mapCkToken... : token` directly in the `.map` function of the array

peterpeterparker (approved):
LGTM, thx

PR #2352, opened 2024-09-13
style(frontend): show a dash for null balances

AntonioVentilii-DFINITY:
# Motivation

In case the `balance` prop of the token is null (meaning that the loader is there but there is an issue/error, or it was stopped), we would like to show a dash `-` instead of `0`, both for the balance and for the USD balance.

# Results

Simulating with random tokens with `null` balance.

Screenshot 2024-09-13 at 17 03 38


peterpeterparker (approved):
LGTM, thx

PR #2354, opened 2024-09-13
feat(frontend): remove sorting dependency on pointer events [GIX-2944]

AntonioVentilii-DFINITY:
# Motivation

The token list was giving issues due to wrong consistency between pointer events and re-sorting. We remove then the logic of re-sorting ONLY when there is no pointer event, and we leave the pointer events disabled when animating.

# Changes

- Removed logic on resorting dependent on pointer events from `TokenDisplayHandler`.
- Removed all deprecated utils.

# Test

On local replica and in staging, now there is the correct behavior.


peterpeterparker (approved):
LGTM, thx

PR #2355, opened 2024-09-13
feat(lint): created local rule to use isNullish and nonNullish

AntonioVentilii-DFINITY:
# Motivation

We would like to enforce the usage of functions `isNullish` and `nonNullish` in the code whenever possible. As first step we look for instances of comparing a variable with `undefined` and/or `null`.

# Changes

- Created custom local lint rule for `isNullish` usage.
- Created custom local lint rule for `nonNullish` usage.
- Create exceptions where necessary.
- Fix issues created by new rules.

# Tests

It raised issues, correctly, and we fixed it.


AntonioVentilii-DFINITY:
Clean PR:

- #2553

PR #2357, opened 2024-09-14
feat(frontend): UI component for content with buttons

AntonioVentilii-DFINITY:
# Motivation

We create an UI component that organize the content of the modals that have a button at the end, to concentrate the code.

### DIFF
--- a/src/frontend/src/lib/components/ui/ContentWithButtons.svelte
+++ b/src/frontend/src/lib/components/ui/ContentWithButtons.svelte
@@ -1,0 +1,3 @@
+<slot class="stretch" />
+
+<slot name="buttons" />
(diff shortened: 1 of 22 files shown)


src/frontend/src/lib/components/ui/ContentWithButtons.svelte
```diff
@@ -0,0 +1,3 @@
+<slot class="stretch" />
+
+<slot name="buttons" />
```
peterpeterparker, line 3: What do you think about naming this slot `toolbar` or something similar? If yes, don't know if the cmp should reflects it?

Just an idea, already approving the PR.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/components/ui/ContentWithButtons.svelte
AntonioVentilii-DFINITY: So, if it's `toolbar` the component will be `ContentWithToolBar`, correct?

peterpeterparker: Ah why not, sounds cool no?

AntonioVentilii-DFINITY: Yes, better! Eheh! Changed

PR #2358, opened 2024-09-15
refactor(frontend): remove fallback from ic-token derived stores

AntonioVentilii-DFINITY:
# Motivation

The derived store `tokenWithFallback` falls back on ETH (or SepoliaETH) and consequently even the store `tokenWithFallbackAsIcToken`.

However, it is unnecessary for the derived stores that checks if a particular `token` is a ck-token, since ETH will never have a prop `ledgerCanisterId` (and the util functions checks for a non-nullish token).

So we remove that particular usage from the code substituting `tokenWithFallbackAsIcToken` with `tokenAsIcToken`.

### DIFF
--- a/src/frontend/src/icp-eth/derived/cketh.derived.ts
+++ b/src/frontend/src/icp-eth/derived/cketh.derived.ts
@@ -8,12 +8,13 @@ import {
 	toCkErc20HelperContractAddress,
 	toCkEthHelperContractAddress
 } from '$icp-eth/utils/cketh.utils';
-import { tokenWithFallbackAsIcToken } from '$icp/derived/ic-token.derived';
+import { tokenAsIcToken } from '$icp/derived/ic-token.derived';
 import type { IcCkToken } from '$icp/types/ic';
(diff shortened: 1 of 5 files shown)


peterpeterparker (commented):
Are you sure it works in every use case and feature of the dApp? Have you thoroughly tested this? `tokenAsIcToken` is actually cast as `IcToken`, but after reviewing the code, it seems to be a shorthand, as it can actually be null or undefined.

On another note, could you first provide a test in a separate PR for the existing derived stores modified in this PR? It would be great to start testing those stores, given that this information is highly sensitive. WDYT?

AntonioVentilii-DFINITY:
> On another note, could you first provide a test in a separate PR for the existing derived stores modified in this PR?

You mean the utils inside the stores? Or the whole logic of the derived stores?

peterpeterparker:
> Or the whole logic of the derived stores?

I meant the all logic - like starting testing derived stores

AntonioVentilii:
This PR is from my old account and it is not actual anymore.

PR #2359, opened 2024-09-15
feat(frontend): add multi-token send modal to networks

AntonioVentilii-DFINITY:
# Motivation

When the user selects a specific network, it would be useful that the Send button shows an initial step to choose the token from (like it is done in Chain Fusion).
The generic Send button already provides this feature with a small change: remove the initial step to choose the token, when we are already in the page of a token.

# Tests

https://github.com/user-attachments/assets/d15539fd-1430-4248-a9e1-9478d5632bb9

### DIFF
--- a/src/frontend/src/lib/components/hero/Actions.svelte
+++ b/src/frontend/src/lib/components/hero/Actions.svelte
@@ -2,12 +2,10 @@
 	import EthReceive from '$eth/components/receive/EthReceive.svelte';
 	import ConvertToCkERC20 from '$eth/components/send/ConvertToCkERC20.svelte';
 	import ConvertToCkETH from '$eth/components/send/ConvertToCkETH.svelte';
-	import EthSend from '$eth/components/send/EthSend.svelte';
 	import { erc20UserTokensInitialized } from '$eth/derived/erc20.derived';
 	import ConvertToBTC from '$icp/components/convert/ConvertToBTC.svelte';
 	import ConvertToEthereum from '$icp/components/convert/ConvertToEthereum.svelte';
 	import IcReceive from '$icp/components/receive/IcReceive.svelte';
-	import IcSend from '$icp/components/send/IcSend.svelte';
 	import { tokenCkBtcLedger } from '$icp/derived/ic-token.derived';
 	import { erc20ToCkErc20Enabled, ethToCkETHEnabled } from '$icp-eth/derived/cketh.derived';
 	import ContextMenu from '$lib/components/hero/ContextMenu.svelte';
@@ -41,13 +39,7 @@
 		<Receive />
 	{/if}
 
-	{#if $networkICP}
-		<IcSend token={$tokenWithFallback} />
-	{:else if $networkEthereum}
-		<EthSend token={$tokenWithFallback} />
-	{:else if $pseudoNetworkChainFusion}
-		<Send />
-	{/if}
+	<Send />
 
 	{#if convertEth}
 		{#if $networkICP}
--- a/src/frontend/src/lib/components/send/SendModal.svelte
+++ b/src/frontend/src/lib/components/send/SendModal.svelte
@@ -1,10 +1,12 @@
 <script lang="ts">
 	import { WizardModal, type WizardStep, type WizardSteps } from '@dfinity/gix-components';
+	import { isNullish } from '@dfinity/utils';
 	import { createEventDispatcher } from 'svelte';
 	import SendTokensList from '$lib/components/send/SendTokensList.svelte';
 	import SendWizard from '$lib/components/send/SendWizard.svelte';
-	import { sendWizardStepsComplete } from '$lib/config/send.config';
+	import { sendWizardStepsComplete, sendWizardStepsWithQrCodeScan } from '$lib/config/send.config';
 	import { ethAddressNotLoaded } from '$lib/derived/address.derived';
+	import { pageToken } from '$lib/derived/page-token.derived';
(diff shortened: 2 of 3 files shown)


src/frontend/src/lib/components/send/SendModal.svelte
```diff
@@ -25,7 +27,9 @@
 	let sendProgressStep: string = ProgressStepsSend.INITIALIZATION;
 
 	let steps: WizardSteps;
-	$: steps = sendWizardStepsComplete({ i18n: $i18n });
+	$: steps = isNullish($pageToken)
```
peterpeterparker, line 30: Why `isNullish($pageToken)`?

src/frontend/src/lib/components/send/SendModal.svelte
```diff
@@ -25,7 +27,9 @@
...
+		? sendWizardStepsComplete({ i18n: $i18n })
```
peterpeterparker, line 31: What does "StepsComplete" mean?

src/frontend/src/lib/components/send/SendModal.svelte
```diff
@@ -82,6 +86,7 @@
 			bind:targetNetwork
 			bind:amount
 			bind:sendProgressStep
+			formCancelAction={isNullish($pageToken) ? 'back' : 'close'}
```
peterpeterparker, line 89: `isNullish($pageToken)` is repeated, please use a local state

src/frontend/src/lib/components/send/SendModal.svelte
AntonioVentilii-DFINITY: It is just to differentiate if we are in the token transactions page already (in this case skips the "Select token" step) or if any other page

src/frontend/src/lib/components/send/SendModal.svelte
AntonioVentilii-DFINITY: we have 3 different lists:

- `sendWizardSteps` --> `SEND` + `REVIEW` + `SENDING`
- `sendWizardStepsWithQrCodeScan` --> the above + `QR CODE`
- `sendWizardStepsComplete` --> the above + `SELECT TOKEN` as additional first step 

if you want I can make `sendWizardSteps` have a parameter that says if it should include the `SELECT TOKEN` step or not, adapt the usage through the code, and then deprecate `sendWizardStepsComplete`

src/frontend/src/lib/components/send/SendModal.svelte
AntonioVentilii-DFINITY: done!

src/frontend/src/lib/components/send/SendModal.svelte
AntonioVentilii-DFINITY: OR

Since sendWizardStepsComplete is used only there, i can make in the condition something like

`[ STEP_SEELCT_TOKEN, ... sendWizardStepsWithQrCodeScan]`

src/frontend/src/lib/components/send/SendModal.svelte
peterpeterparker: - I would suggest to not have a route related logic within this component. I would use a property and set that particular value in its consumer, in the here

- Instead of `isNullish($pageToken)` I would use an explicit logic such as `isRouteTokens`

src/frontend/src/lib/components/send/SendModal.svelte
peterpeterparker: So "complete" means "all steps"? Maybe just renaming `sendWizardStepsComplete` to `allSendWizardSteps` or something in that direction?

AntonioVentilii-DFINITY: yes... makes more sense as name! this PR or another?

peterpeterparker: As you rather like

src/frontend/src/lib/components/send/SendModal.svelte
AntonioVentilii-DFINITY: Have a look at the last commit please? Is that what you meant?

src/frontend/src/lib/components/send/SendModal.svelte
AntonioVentilii-DFINITY: Done with PR https://github.com/dfinity/oisy-wallet/pull/2366

src/frontend/src/lib/components/hero/Actions.svelte
```diff
@@ -30,6 +30,9 @@
 
 	let convertBtc = false;
 	$: convertBtc = $tokenCkBtcLedger && $erc20UserTokensInitialized;
+
+	let isTokenPage = false;
+	$: isTokenPage = isRouteTransactions($page);
```
peterpeterparker, line 35: Maybe a naming to review? 

isToken... vs is...Transaction

Or `isRouteTransactions` is actually incorrect?

src/frontend/src/lib/components/send/SendModal.svelte
peterpeterparker: Sounds good, thx

src/frontend/src/lib/components/hero/Actions.svelte
AntonioVentilii-DFINITY: right! changed to `isTransactionsPage` (even if ultimately probably it will become a token page that page... but it's the story for another time)

peterpeterparker (approved):
LGTM, thx

PR #2365, opened 2024-09-16
chore(frontend): remove redundant opacity class [GIX-2929]

AntonioVentilii-DFINITY:
# Motivation

With the introduction of a determined opacity for class `disabled` of `primary` and `secondary` buttons (PR #2345 ), we can remove the `opacity` class in the components, since they will be superseded by the CSS definition.


peterpeterparker (approved):
LGTM, thx

PR #2367, opened 2024-09-16
feat(frontend): create env flag for OnRamper features

AntonioVentilii-DFINITY:
# Motivation

We will start adding features to buy from OnRamper. So, until it is finished, we will use the env flag `VITE_ONRAMPER_ENABLED` to show/hide such feature, for development scopes.


peterpeterparker (approved):
Thx

PR #2369, opened 2024-09-16
chore(backend): Remove signing apis

bitdivine:
# Motivation
The backend canister signing APIs are now provided by the chain fusion signer.

# Changes
- Remove all signing and threshold signature APIs from the backend canister.
- Remove the utilities and tests for these methods.
- Fixes:
  - Get the `SignRequest` typescript data type from the signer bindings rather than the backend bindings.  This was an oversight but has no consequence as the types are the same.

Not removed:
- Anything beyond the bulk removal of methods and immediate dependencies.  For example:
  - The migration methods, or if we wish to keep the migration methods, the threshold signing aspects of migration.
  - Removal of dependencies in `Crates.toml` that may be used only for signing.
  - I have not made a detailed scan for other small things that might be removed or simplified.
  - Any frontend simplification possible now that the backend and signer are definitely distinct.

# Tests
- Existing CI should cover the remaining functionality.
- The chain fusion signer CI should cover the signing methods comprehensively.


AntonioVentilii-DFINITY (approved):
LGTM, tks!

PR #2379, opened 2024-09-16
style(frontend): update social image [GIX-2949]

AntonioVentilii-DFINITY:
# Motivation

New social image updated. Note that we gave `-v2` suffix to avoid conflicts with cached files.


peterpeterparker (approved):
LGTM, thx

PR #2381, opened 2024-09-17
build(frontend): bump dfinity dependencies (for signer)

peterpeterparker:
# Motivation

For the signer we will need a new version of gix-cmp (which contains a `Markdown` cmp) and agent-js 2.1.1 (which contains fixes). So this PR bumps the `next` and released libraries of DFINITY.

Since I was already there I also bumped verifiable-credentials to its released version.


AntonioVentilii-DFINITY (approved):
LGTM!

PR #2385, opened 2024-09-17
refactor(frontend): common util to check ck-tokens ledgers

AntonioVentilii-DFINITY:
# Motivation

Small refactoring to avoid repetition of code.


src/frontend/src/icp/utils/ic-send.utils.ts
```diff
@@ -30,14 +31,19 @@ export const isBtcAddress = (address: BtcAddress | undefined): boolean => {
 export const invalidBtcAddress = (address: BtcAddress | undefined): boolean =>
 	!isBtcAddress(address);
 
+const isTokenCkLedger = (
```
peterpeterparker, line 34: - `isTokenLedger`?

- Can you use an object as parameter?

peterpeterparker: I prefer

```
export const isTokenCkBtcLedger = (token: Partial<IcToken>): boolean =>
	isTokenLedger({token, ledgerIds: CKBTC_LEDGER_CANISTER_IDS});
```

peterpeterparker: Btw. if you are up to setup an eslint rules when you got a bit of time, that would be super.

AntonioVentilii-DFINITY: About this one? what would you put as rule?

peterpeterparker: > Always pass an object unless one single parameter?

I would say so, what do you think?

AntonioVentilii-DFINITY: I like it! Even because I didn't know what was the convention eheh

AntonioVentilii-DFINITY: In the meanwhile, corrected this one, have a look please

peterpeterparker (approved):
LGTm, thx

src/frontend/src/icp/utils/ic-send.utils.ts
AntonioVentilii-DFINITY: Done with PR:

- #2416

PR #2386, opened 2024-09-17
feat(frontend): add OnRamper ID to networks [GIX-2854]

AntonioVentilii-DFINITY:
# Motivation

We add the OnRamper Network ID to all the networks that we would like to have supported with it. The list is [here](https://docs.onramper.com/docs/network-support).


src/frontend/src/lib/types/onramper.ts
```diff
@@ -0,0 +1,3 @@
+// The list of networks that are supported by OnRamper can be found here:
+// https://docs.onramper.com/docs/network-support
+export type OnRamperNetworkId = string;
```
peterpeterparker, line 3: Can we list only the subset of those we are using instead of `string`?

src/frontend/src/lib/types/network.ts
```diff
@@ -1,10 +1,16 @@
...
-export interface Network {
+export type Network = {
 	id: NetworkId;
 	env: NetworkEnvironment;
 	name: string;
 	icon?: string;
+} & NetworkLinkedData;
+
+export interface NetworkLinkedData {
+	onRamperId?: OnRamperNetworkId;
```
peterpeterparker, line 15: Should we forseen that we may support more links to crypto exchange than OnRamper?

src/frontend/src/lib/types/onramper.ts
AntonioVentilii-DFINITY: Better! Changed

src/frontend/src/lib/types/network.ts
AntonioVentilii-DFINITY: you mean to create a sub interface inside NetworkLinkedData?

```
interface OnRamper {
   onRamperId...
}
```

peterpeterparker: I just meant what if we soon add another crypto exchange XYZ? then we just add xyzId next to onRamperId?

peterpeterparker: We already use exchange to get the conversion Token / USD. Should we convert this also as exchange or those are two different things?

AntonioVentilii-DFINITY: well, this one is specific for networks. For token I would have merged with `exchange`

peterpeterparker: So not using exchange here? Like `NetworkOnrampData` or `NetworkFiatToCryptoData` or something?

I think I gonna be confuse if there are exchange and exchange, if this makes sense?

AntonioVentilii-DFINITY: Sorry, I think I got lost: you would then create a new interface specifically for OnRamper like `NetworkOnramperData`? And put it directly to the `Network` type, or create another interface that collects all the "exchange" data?

peterpeterparker: Something that does not contains the word "exchange", for example:

```
export type Network = {
   ...existingFields
   buy?: NetworkBuyCrypto
}

export type NetworkBuyCrypto {
   onRamperId: OnRamperNetworkId;
}
```

AntonioVentilii-DFINITY: Check my last commit please (even if I am not sure if I should create a middle interface or not)

src/frontend/src/lib/types/network.ts
```diff
@@ -1,10 +1,18 @@
+import type { OnRamperNetworkId } from '$lib/types/onramper';
+import type { AtLeastOne } from '$lib/types/utils';
+
 export type NetworkId = symbol;
 
 export type NetworkEnvironment = 'mainnet' | 'testnet';
 
-export interface Network {
+export type Network = {
```
peterpeterparker, line 8: Leftover change or opiniated change?

Not against btw. maybe we should define an eslint rule for this but, in the oisy wallet signer library I think I enforced interface.

src/frontend/src/lib/types/utils.ts
```diff
@@ -7,3 +7,6 @@ export type ResultSuccessReduced<T = unknown> = ResultSuccess<T[]>;
 // We disable the eslint rule here because this is the utility type that we use to define such rule.
 // eslint-disable-next-line local-rules/use-option-type-wrapper
 export type Option<T> = T | null | undefined;
+
+export type AtLeastOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<keyof T, Keys>> &
```
peterpeterparker, line 11: Nice!

src/frontend/src/lib/types/network.ts
peterpeterparker: Looks good. Hope it works for you too?

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/lib/types/network.ts
AntonioVentilii-DFINITY: yes yes! i was afraid on naming, but then all good!

src/frontend/src/lib/types/network.ts
AntonioVentilii-DFINITY: ah! thank you! another thing I was always in doubt with! I will enforce `interface` then in another PR

AntonioVentilii-DFINITY: Done with https://github.com/dfinity/oisy-wallet/pull/2467

PR #2390, opened 2024-09-17
refactor(frontend): move listener types to lib

DenysKarmazynDFINITY:
# Motivation

Move listener types from `icp` to `lib` to re-use them for both BTC and IC wallet workers.


peterpeterparker (approved):
LGTM, thx

PR #2391, opened 2024-09-17
refactor(frontend): move wallet utils to lib

DenysKarmazynDFINITY:
# Motivation

Move wallet utils from `icp` to `lib` to re-use them for both BTC and IC wallet workers.

### DIFF
--- a/src/frontend/src/icp/utils/wallet.utils.ts
+++ b/src/frontend/src/icp/utils/wallet.utils.ts
@@ -1,8 +1,7 @@
 import { initIcpWalletWorker } from '$icp/services/worker.icp-wallet.services';
 import { initIcrcWalletWorker } from '$icp/services/worker.icrc-wallet.services';
 import type { IcToken } from '$icp/types/ic';
-import type { WalletWorker } from '$lib/types/listener';
-import type { Token } from '$lib/types/token';
+import type { InitWallerWorkerParams, WalletWorker } from '$lib/types/listener';
 
-export const initWalletWorker = async ({ token }: { token: Token }): Promise<WalletWorker> =>
+export const initWalletWorker = async ({ token }: InitWallerWorkerParams): Promise<WalletWorker> =>
 	token.standard === 'icrc' ? initIcrcWalletWorker(token as IcToken) : initIcpWalletWorker();
--- a/src/frontend/src/icp/utils/wallet.utils.ts
+++ b/src/frontend/src/icp/utils/wallet.utils.ts
@@ -1,7 +1,7 @@
 import { initIcpWalletWorker } from '$icp/services/worker.icp-wallet.services';
 import { initIcrcWalletWorker } from '$icp/services/worker.icrc-wallet.services';
 import type { IcToken } from '$icp/types/ic';
-import type { InitWallerWorkerParams, WalletWorker } from '$lib/types/listener';
+import type { InitWalletWorkerFn } from '$lib/types/listener';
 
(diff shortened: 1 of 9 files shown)


src/frontend/src/lib/utils/wallet.utils.ts
```diff
@@ -45,18 +45,18 @@ export const cleanWorkers = ({
...
 	token
 }: {
 	workers: Map<TokenId, WalletWorker>;
-	token: IcToken;
+	token: Token;
 }) => {
 	if (!workers.has(token.id)) {
 		const worker = await (token.standard === 'icrc'
-			? initIcrcWalletWorker(token)
+			? initIcrcWalletWorker(token as IcToken)
```
peterpeterparker, line 59: This should remains in `$icp` in my opinion as it is solely related to the IC network. I mean the all function `loadWorker` or just those two `initIc...Worker` calls.

`wallet.utils.ts` should ideally not have references on `$icp`.

DenysKarmazynDFINITY: Updated the function to hide the ICRC/ICP wallet workers initialisation logic. Though, `wallet.utils` still has one reference on `$icp` (`initWalletWorker`) and will have one on `$btc`, as my idea is to re-use `loadWorker` for BTC tokens, too. The next step would be to update the `IcLoaderWallet` component: it will be renamed to `LoaderWallet`, moved to `$lib/components/core` and is gonna accept tokens as a prop.

peterpeterparker: We should pass the function to init as a parameter, that way it won't have any references.

DenysKarmazynDFINITY: Sounds good, done.

src/frontend/src/icp/utils/wallet.utils.ts
```diff
@@ -0,0 +1,7 @@
+import { initIcpWalletWorker } from '$icp/services/worker.icp-wallet.services';
+import { initIcrcWalletWorker } from '$icp/services/worker.icrc-wallet.services';
+import type { IcToken } from '$icp/types/ic';
+import type { InitWallerWorkerParams, WalletWorker } from '$lib/types/listener';
+
+export const initWalletWorker = async ({ token }: InitWallerWorkerParams): Promise<WalletWorker> =>
```
peterpeterparker, line 6: nitpick: function does not need to be marked as `async`

src/frontend/src/lib/types/listener.ts
```diff
@@ -1,5 +1,9 @@
+import type { Token } from '$lib/types/token';
+
 export interface WalletWorker {
 	start: () => void;
 	stop: () => void;
 	trigger: () => void;
 }
+
+export type InitWallerWorkerParams = { token: Token };
```
peterpeterparker, line 9: nitpick: you can type the all function type, not just the params

```
type InitWalletWorkerFn = (params: {token: Token}) => Promise<WalletWorker>;

export const initWalletWorker: InitWalletWorkerFn = ({token}): : Promise<WalletWorker> => ...
```

peterpeterparker (approved):
LGTM, thanks

Added two nitpick suggestions.

src/frontend/src/icp/utils/wallet.utils.ts
DenysKarmazynDFINITY: Indeed!

src/frontend/src/lib/types/listener.ts
DenysKarmazynDFINITY: True, thanks 👍

PR #2398, opened 2024-09-18
feat(frontend): create signer canister

DenysKarmazynDFINITY:
# Motivation

The idea is to make `signer` endpoints (specifically `caller_btc_balance`) available as a canister to use in a web worker.

### DIFF
--- a/src/frontend/src/lib/api/signer.api.ts
+++ b/src/frontend/src/lib/api/signer.api.ts
@@ -57,11 +57,9 @@ export const signPrehash = async ({
 	return sign_prehash(hash);
 };
 
-export const signerCanister = async ({
-	identity
-}: {
-	identity: Identity;
-}): Promise<SignerCanister> => {
+// TODO: implement updateBtcBalance method
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const signerCanister = async ({ identity }: { identity: Identity }): Promise<SignerCanister> => {
 	const agent = await getAgent({ identity });
 
 	return SignerCanister.create({
--- a/src/frontend/src/lib/api/signer.api.ts
+++ b/src/frontend/src/lib/api/signer.api.ts
@@ -1,7 +1,12 @@
 import type { BitcoinNetwork, SignRequest } from '$declarations/signer/signer.did';
 import { getSignerActor } from '$lib/actors/actors.ic';
+import { getAgent } from '$lib/actors/agents.ic';
+import { SignerCanister } from '$lib/canisters/signer.canister';
+import { SIGNER_CANISTER_ID } from '$lib/constants/app.constants';
 import type { EthAddress } from '$lib/types/address';
 import type { OptionIdentity } from '$lib/types/identity';
+import type { Identity } from '@dfinity/agent';
+import { Principal } from '@dfinity/principal';
 
 export const getBtcAddress = async ({
 	identity,
@@ -50,4 +55,17 @@ export const signPrehash = async ({
 }): Promise<string> => {
 	const { sign_prehash } = await getSignerActor({ identity });
 	return sign_prehash(hash);
+};
+
+export const signerCanister = async ({
+	identity
+}: {
+	identity: Identity;
+}): Promise<SignerCanister> => {
+	const agent = await getAgent({ identity });
+
+	return SignerCanister.create({
+		agent,
+		canisterId: Principal.fromText(SIGNER_CANISTER_ID)
+	});
 };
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/api/signer.api.ts
```diff
@@ -51,3 +56,16 @@ export const signPrehash = async ({
 	const { sign_prehash } = await getSignerActor({ identity });
 	return sign_prehash(hash);
 };
+
+export const signerCanister = async ({
+	identity
+}: {
+	identity: Identity;
+}): Promise<SignerCanister> => {
```
peterpeterparker, line 64: The api should not leak the entire canister but solely expose the function required by the consumer.

For example `getTransactions` or `getKnownUtxos` for web worker.

DenysKarmazynDFINITY: My bad, seeing `export` in this place is indeed confusing - it was there as a temporary fix to satisfy the `@typescript-eslint/no-unused-vars` rule.  For clarity, I updated it to use `eslint-disable-next-line` instead. It will be of course removed later.

The plan is to add `updateBtcBalance` function that can be used by a worker in the next PR. Why not in this one - I would also need to create (or re-use the one from `address.store`) a mapper between bitcoin networks values (e.g. `mainnet` -> `{mainnet: null}`), so don't want to overload this branch.

src/frontend/src/lib/canisters/signer.canister.ts
```diff
@@ -0,0 +1,29 @@
...
+import { idlFactory as idlCertifiedFactorySigner } from '$declarations/signer/signer.factory.certified.did';
+import { idlFactory as idlFactorySigner } from '$declarations/signer/signer.factory.did';
+import { Principal } from '@dfinity/principal';
+import { Canister, createServices, type CanisterOptions } from '@dfinity/utils';
+
+interface SignerCanisterOptions<T> extends Omit<CanisterOptions<T>, 'canisterId'> {
+	canisterId: Principal;
+}
+
+export class SignerCanister extends Canister<SignerService> {
```
peterpeterparker, line 11: Is there a particular reason why you developed this with a class and specific pattern? Are you planing to re-implement  the module`actors.ic` later on?

If not, we can also resolve it with a function and expose `createActor`.

```
export const getSignerWorkerActor = async ({
	identity,
	signerCanisterId,
	certified = true
}: {
	identity: OptionIdentity;
	certified?: boolean;
	signerCanisterId: CanisterIdText;
}): Promise<SignerActor> => {
	assertNonNullish(identity);

	return await createActor<SignerActor>({
		canisterId: signerCanisterId,
		idlFactory: certified ? idlCertifiedFactorySigner : idlFactorySigner,
		identity
	});
};
```
 
But not against the class pattern, we should just be consistent and rework existing actors, might be a good idea.

DenysKarmazynDFINITY: No particular reasons, I was mostly inspired by how canisters implemented in `ic-js`. In general, I think it would be nice if we follow the same "class" pattern for actors in oisy. It doesn't involve a lot of refactorings as, from what I see, we'll only need to rework the backend actor as well as to extend the one that I implemented for signer with other methods that are already in use. WDYT?

peterpeterparker: That’s generally not the approach we've taken in the NNS dapp or Oisy; we used this pattern in ic-js for historical reasons.

That said, I'm not opposed to it, and I can see the value if we want to expand further by introducing scoped caching.

If you’d like to go down this route, we can start by adapting the existing code and transforming `getSignerActor` into a Class-based approach in a separate PR, sSince the current PR adds new code without modifying existing functionality and also since we discussed scope in comments. Note that the pattern should account for multiple actors (for UI or worker contexts) and maintain the caching features. It may also be worthwhile to add tests if we decide to proceed too I would say.

DenysKarmazynDFINITY: Sounds like a plan! Shall we then merge this branch so I can use the SignerCanister class for the further refactorings we discussed + for the next BTC integration PRs?

peterpeterparker: >  Shall we then merge this branch

I rather like not. I think we should not introduce the new pattern this way but, cleanly implement it.

DenysKarmazynDFINITY: I created a PR based on this one where I used the canister approach for all the existing signer API functions - https://github.com/dfinity/oisy-wallet/pull/2425.

DenysKarmazynDFINITY: The code from this PR is now a part of https://github.com/dfinity/oisy-wallet/pull/2425. We can continue the discussions there.

PR #2401, opened 2024-09-18
style(frontend): move "Back" button and context menu [GIX-2973]

AntonioVentilii-DFINITY:
# Motivation

We move the `Back` link and the context menu on the sides of the token logo (except for Settings page).

NOTE: the context menu icon will be changed after.

# Changes

- Set the `Back` element to appear in the header only for Settings page.
- Remove `ContextMenu` from the `Actions`.
- Include both elements in the Hero content on the sides of the token logo and the balance.

# Tests

Screenshot 2024-09-18 at 11 41 13
Screenshot 2024-09-18 at 11 41 18
Screenshot 2024-09-18 at 11 41 36
Screenshot 2024-09-18 at 11 41 43
Screenshot 2024-09-18 at 11 41 53


peterpeterparker:
Unsolicited opiniated feedback: Are you sure? Following is really the expected UI/UX? That look super weird, I don't think I have any mobile apps installed on my phone that have that behavior.

image

AntonioVentilii-DFINITY:
> Unsolicited opiniated feedback: Are you sure? Following is really the expected UI/UX? That look super weird, I don't think I have any mobile apps installed on my phone that have that behavior.

Completly agreed, but it is just temporary: that button will change after this PR (i didn't want to put all together). If you prefer I can put it here

peterpeterparker:
Ah ok, that wasn't clear for me when I read the description. Let's follow your path then, let me review the PR.

src/frontend/src/lib/components/hero/Header.svelte
```diff
@@ -9,10 +9,10 @@
 	import OisyWalletLogo from '$lib/components/icons/OisyWalletLogo.svelte';
 	import { authSignedIn } from '$lib/derived/auth.derived';
 	import { modalAboutHow, modalAboutWhat } from '$lib/derived/modal.derived';
-	import { isSubRoute } from '$lib/utils/nav.utils';
+	import { isRouteSettings } from '$lib/utils/nav.utils';
 
 	let back = false;
-	$: back = isSubRoute($page);
+	$: back = isRouteSettings($page);
```
peterpeterparker, line 15: Not sure about testing solely for `settings` within this `Header` component, feels like something we might forget if we ever add more pages. Would it be worth having a prop?

Also `transactions` page has no back button anymore?

AntonioVentilii-DFINITY: - yes, makes sense! add a back `prop` that is given directly in `+layout`?

- the `transactions` page will have it automatically because it is included in `summary` that is defined as `summary={route === 'transactions'}` on `+layout`. My idea is to change this after: make it so summary is not passed down, but it is calculated directly in the component as `summary=isTransactionPage(`... do you want me to do it in a PR before this one (it will be in `main`)?

peterpeterparker: >  add a back prop that is given directly in +layout?

yeah for example

> My idea is to change this afte

Coolio

> do you want me to do it in a PR before

Another PR sounds good

AntonioVentilii-DFINITY: @peterpeterparker have a look at the last commit please

peterpeterparker: Did you push?

AntonioVentilii-DFINITY: I think so yes, should be this commit

https://github.com/dfinity/oisy-wallet/pull/2401/commits/0610a1b471074d40ca74fffc818605dfd97e0817

peterpeterparker: I see it now, thx, but, shouldn't we use it here as well?

```
let back = false;
$: back = isRouteSettings($page);
```

AntonioVentilii-DFINITY: I refactored it a bit, have a look please

src/frontend/src/lib/components/hero/Hero.svelte
```diff
@@ -8,15 +8,16 @@
 	export let usdTotal = false;
 	export let summary = false;
 	export let actions = true;
-	export let more = false;
+	export let headerBack = false;
```
peterpeterparker, line 11: Is that temporary or definitive?

If temporary, can we add a TODO.

If definitive, is it possible to have two backs buttons in the UI?
If not, can we have a single property?

```
export let back: "header" | "hero" | undefined
```

AntonioVentilii-DFINITY: i am not really sure it is temporary, not that i know of...

but I would prefer to have a single component, and a prop with the selection (last commit)

peterpeterparker (approved):
LGTM, thx

PR #2394, opened 2024-09-18
build(frontend): bump sass

AntonioVentilii-DFINITY:
# Motivation

Bumping `sass`.

# Note

Sass `1.77.8` introduces a deprecation warning:

```
DEPRECATION WARNING: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.

More info: https://sass-lang.com/d/mixed-decls

   ╷
19 │           aspect-ratio: 4 / 3;
   │           ^^^^^^^^^^^^^^^^^^^ declaration
   ╵
   ╷
15 │ ┌         @media only screen and (hover: none) and (pointer: coarse) {
16 │ │             aspect-ratio: 1 / 1;
17 │ │         }
   │ └─── nested rule
   ╵
    - 19:3  root stylesheet
```

# Changes

- Update Sass.
- Resolve warnings my moving declarations above rule.

PR #2402, opened 2024-09-18
fix(frontend): missing fixes after bumping sass

AntonioVentilii-DFINITY:
# Motivation

There are some missing fixes that appears only during building, due to the bumped sass (PR #2394 ).


DenysKarmazynDFINITY (approved):
LGTM, thanks!

PR #2407, opened 2024-09-18
feat(frontend): make LoaderWallets component reusable

DenysKarmazynDFINITY:
# Motivation

The idea is to make LoaderWallets component reusable for BTC.

### DIFF
--- a/src/frontend/src/lib/components/core/LoaderWallets.svelte
+++ b/src/frontend/src/lib/components/core/LoaderWallets.svelte
@@ -34,3 +34,5 @@
 </script>
 
 <svelte:window on:oisyTriggerWallet={triggerTimer} />
+
+<slot />
--- a/src/frontend/src/lib/components/core/LoaderWallets.svelte
+++ b/src/frontend/src/lib/components/core/LoaderWallets.svelte
@@ -1,0 +1,36 @@
+<script lang="ts">
+	import { debounce } from '@dfinity/utils';
+	import { onDestroy } from 'svelte';
+	import type { InitWalletWorkerFn, WalletWorker } from '$lib/types/listener';
+	import type { Token, TokenId } from '$lib/types/token';
+	import { cleanWorkers, loadWorker } from '$lib/utils/wallet.utils';
+
+	export let tokens: Token[];
+	export let initWalletWorker: InitWalletWorkerFn;
+
+	const workers: Map<TokenId, WalletWorker> = new Map<TokenId, WalletWorker>();
+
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/core/LoaderWallets.svelte
```diff
@@ -0,0 +1,36 @@
...
+
+	onDestroy(() => {
+		workers.forEach((worker) => worker.stop());
+		workers.clear();
+	});
+
+	const triggerTimer = () => workers.forEach((worker) => worker.trigger());
+</script>
+
+<svelte:window on:oisyTriggerWallet={triggerTimer} />
```
peterpeterparker, line 36: It should not contains a `<slot />`?

i.e. shouldn't it be use

```
<LoaderWallets tokens={$enabledIcTokens} {initWalletWorker} >
  <slot />
</LoaderWallets />
```

Or is it safe if the children are outside? If it's safe, shouldn't it rather be a `service`?

DenysKarmazynDFINITY: Good point. The answer to your question is yes, it's safe to have children outside. This is also valid for other loaders too (`LoaderBalances`, `ExchangeWorker`, `LoaderMetamask`). In case of `LoaderWallet`, we can justify having it as a component because of attaching a global event [here](https://github.com/dfinity/oisy-wallet/pull/2407/files#diff-b0c3d5389316fc4e867ef469a0ffcf9037d1a584ca59827279acf8f6be5071c3R36). But you're right, all those things can be considered as services. For the consistency, I'd propose to keep `LoaderWallet` as component, and update all loaders separately if we decide to proceed with the refactoring. WDYT?

peterpeterparker: I personally prefer using `slot`, as it makes the loaders similar to a React or Svelte context pattern. It also potentially allows for animating the UI while the worker is being initialized or destroyed.

That said, if this approach doesn't make much sense, I'm certainly not against removing it.

DenysKarmazynDFINITY: It does, I checked animation implementation in the `Loader` component, something like this can be potentially done in other loaders, too. 

I wrapped `slot` in `IcLoaderWallet` (and will do the same in `BtcLoaderWallet`) so in case we want to implement some enter/exit effects it can be done in the shared `WalletLoader` component.

peterpeterparker (approved):
LGTM, thx

PR #2408, opened 2024-09-18
feat(frontend): Reproducible frontend build 

Sawchord:
# Motivation

If we want to deploy the frontend via `dfx-orbit`, we need to be able to build it reproducibly.
This PR adds frontend building to the Dockerfile. 

# Changes

Modifies the Dockerfile to install NVM and then build the frontend

# Tests

Reproducibly has been tested manually by building the artifacts multiple times on differenc machines (MAC and Linux)


Dockerfile
```diff
@@ -57,5 +69,17 @@ RUN ./docker/build --backend
 
 RUN sha256sum /backend.wasm.gz
 
+
+ARG network="staging"
```
frederikrothenberger, line 73: Would it make sense to introduce a new build stage for the front-end build? As is, this is part of the `build_backend` stage, which now is probably no longer named accurately?

Dockerfile
```diff
@@ -19,6 +19,18 @@ RUN DEBIAN_FRONTEND=noninteractive apt update && apt install -y \
 
 RUN curl --proto '=https' --tlsv1.2 -L https://github.com/mikefarah/yq/releases/download/v4.33.3/yq_linux_amd64 | install -m 755 /dev/stdin /bin/yq && yq --version | grep yq
 
+# Install node
+RUN curl --fail -sSf https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash
```
frederikrothenberger, line 23: Do the front-end builds and back-end builds depend on each other in any way? Maybe it makes sense to _not_ share a docker file if they are separate, leading to faster builds and smaller set of dependencies.

Dockerfile
```diff
@@ -57,5 +69,17 @@ RUN ./docker/build --backend
...
+ENV ENV=$network
+ENV DFX_NETWORK=$network
+
+RUN echo $ENV
+RUN echo $DFX_NETWORK
+
+RUN npm ci
+RUN npm run build
+
 FROM scratch AS scratch_backend
```
frederikrothenberger, line 83: Imho this should be renamed or split into a separate stage, see above.

frederikrothenberger (commented):
I have a few questions. Note however, that I'm not familiar with the Oisy build at all. @bitdivine is probably more qualified to review this change.

Dockerfile
Sawchord: I think that should be possible. I just wanted to get something going fast, so I decided to go this route. I think, separating the two shouldn't be too difficult. But maybe I should just give this a try now, since this would resolve all your comments below, about the different stages etc.

.node-version
```diff
@@ -0,0 +1 @@
+20.17.0
```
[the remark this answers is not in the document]
Sawchord: To have reproducibility, we need to specify the version down to the patch version, otherwise different runs might download different minor/patch versions. I don't know exactly where `.nvmrc` is used, but if we could specify this down to the patch version, then I think we could also use that file.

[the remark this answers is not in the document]
peterpeterparker: Let's try that. Having it specified in two files is already too many in my opinion.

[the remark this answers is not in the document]
Sawchord: Agreed. I will give it a try.

Sawchord (commented):
I split the frontend build of into two different dockerfiles. This should resolve @frederikrothenberger s remarks about the inappropriately named stages as well.

Dockerfile
frederikrothenberger: Yes, thanks!

Dockerfile
Sawchord: Its in a separate dockerfile now, so this no longer applies

Dockerfile
Sawchord: Its in a different file now 👍

.node-version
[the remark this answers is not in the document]
Sawchord: It seems to work 👍

Dockerfile
```diff
@@ -19,6 +19,18 @@ RUN DEBIAN_FRONTEND=noninteractive apt update && apt install -y \
 
 RUN curl --proto '=https' --tlsv1.2 -L https://github.com/mikefarah/yq/releases/download/v4.33.3/yq_linux_amd64 | install -m 755 /dev/stdin /bin/yq && yq --version | grep yq
 
+# Install node
```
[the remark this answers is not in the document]
Sawchord: The code that I used was copied from the Internet Idenitity repository.

I assume your approach should also work. However, I have not tested the reproducibility of this approach, and I don't think it makes sense to investigate this right now. We can come back to this maybe at a later time, if it is ok with you.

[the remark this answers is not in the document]
peterpeterparker: Sure

peterpeterparker (approved):
LGTM, thanks!

Dockerfile.frontend
```diff
@@ -0,0 +1,38 @@
...
+RUN DEBIAN_FRONTEND=noninteractive apt update && apt install -y \
+    curl \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install node
+RUN curl --fail -sSf https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash
+ENV NVM_DIR=/root/.nvm
+COPY .nvmrc .node-version
+RUN cat .node-version
+RUN . "$NVM_DIR/nvm.sh" && nvm install "$(cat .node-version)"
```
bitdivine, line 14:
```
nvm install
```
will pick up the version in the `.nvmrc` and set it to the current version, so you can simplify the line above and probably delete the line below.

bitdivine (commented):
It looks to me as if the code could be simplified.  But otherwise it looks good to me!

PR #2412, opened 2024-09-18
ci(dependabot): add devDependencies auto-updates

AntonioVentilii-DFINITY:
# Motivation

We would like to auto-update devDependencies, with the due exceptions (like `typescript` and `lint`).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,16 +3,17 @@
 
 version: 2
 updates:
-  - package-ecosystem: "cargo"
+  - package-ecosystem: cargo
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: weekly
 
-  - package-ecosystem: "npm"
+  - package-ecosystem: npm
     directory: "/"
     schedule:
-      interval: "weekly"
-    target: "dev"
+      interval: weekly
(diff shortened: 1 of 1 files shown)


AntonioVentilii-DFINITY:
@peterpeterparker please have a look if there are packages to be included/excluded among the exceptions

.github/dependabot.yml
```diff
@@ -6,4 +6,20 @@ updates:
...
+    schedule:
+      interval: "weekly"
+    target: "dev"
+    ignore:
+      - dependency-name: "@dfinity/*"
+      - dependency-name: "@types/*"
+      - dependency-name: "@typescript-eslint/*"
+      - dependency-name: "eslint"
+      - dependency-name: "eslint-*"
+      - dependency-name: "sass"
```
peterpeterparker, line 22: I think we can try to not ignore sass, svelte, and vite. WDYT?

peterpeterparker (approved):
Coolio, thanks. Already approving, just a question.

.github/dependabot.yml
AntonioVentilii-DFINITY: very brave eheh! agreed and done

PR #2416, opened 2024-09-18
feat(lint): custom rule to prefer object parameters

AntonioVentilii-DFINITY:
# Motivation

We would like to introduce a new custom lint rule that enforces the usage of an object when there are two or more parameters as function input.

### Incorrect
```typescript
const fun = (param1: string, param2: number) => {...
```

### Correct
```typescript
const fun2 = ({ param1, param2 }: { param1: string; param2: number }) => {...
```

There are some exceptions:
- loop methods (`map`, `reduce`, `filter`, `forEach`, `sort`, `replace`);
- new `Promise` definitions;
- `Array.from`;
- `JSON.stringify`;
- class constructors.

Furthermore, in this PR, we will not solve all the new issues raised by the new rule, otherwise it would become too big: we add the exceptions to the rule with a `TODO`, for all the functions that we need to fix in other PRs.

# Changes

- Create new local rule excluding the nodes mentioned in the list of exceptions.
- Add specific exceptions to custom functions that serve the same purpose as loop methods (mapping functions, sorting functions).
- Fix a few rules.
- Add temporary exceptions to the rule with a `TODO` associated.

# Tests

The rule worked well: it raises new issues that we could fix.


peterpeterparker (approved):
Really nice, thx!

PR #2421, opened 2024-09-19
style(frontend): new style for Alpha warning [GIX-2981]

AntonioVentilii-DFINITY:
# Motivation

We update the style for the Alpha warning.

### Before

Screenshot 2024-09-19 at 08 03 03

### After

Screenshot 2024-09-19 at 08 02 33

### DIFF
--- a/src/frontend/src/lib/i18n/en.json
+++ b/src/frontend/src/lib/i18n/en.json
@@ -125,7 +125,7 @@
 	},
 	"hero": {
 		"text": {
-			"use_with_caution": "Learn more about this alpha version. Use with caution.",
+			"use_with_caution": "This is alpha version. Use with caution.",
 			"learn_more_about_erc20_icp": "Learn more about ERC20 ICP on Ethereum."
 		}
 	},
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/i18n/en.json
```diff
@@ -125,7 +125,7 @@
 	},
 	"hero": {
 		"text": {
-			"use_with_caution": "Learn more about this alpha version. Use with caution.",
+			"use_with_caution": "This is alpha version. Use with caution.",
```
peterpeterparker, line 128: Isn't a "an" missing in the sentence? "This is an alpha" version or something?

peterpeterparker (approved):
Way better

src/frontend/src/lib/i18n/en.json
AntonioVentilii-DFINITY: I think so to, and in fact I wanted to have even "This is THE Alpha version..."; let me ask Max (native speaker eheh)

AntonioVentilii-DFINITY: Checked with Max: as you suggested, `This is an alpha version...` is the better version

PR #2422, opened 2024-09-19
style(frontend): new class for tertiary buttons

AntonioVentilii-DFINITY:
# Motivation

There is a new class for `tertiary` buttons (settings, manage, options). We apply it to the button "Manage your tokens".

### Before

Screenshot 2024-09-19 at 10 50 55

### After

Screenshot 2024-09-19 at 10 50 35
Screenshot 2024-09-19 at 10 51 21

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -34,7 +34,8 @@ export default {
 			platinum: '#e4e4e4',
 			'mountain-meadow': '#30af91',
 			'dodger-blue': '#1e90ff',
-			turquoise: '#59e7d6'
+			turquoise: '#59e7d6',
+			'alice-blue': '#ecf3fb'
 		}
 	},
 	plugins: []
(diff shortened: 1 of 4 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks!

tailwind.config.js
```diff
@@ -34,7 +34,8 @@ export default {
 			platinum: '#e4e4e4',
 			'mountain-meadow': '#30af91',
 			'dodger-blue': '#1e90ff',
-			turquoise: '#59e7d6'
+			turquoise: '#59e7d6',
+			'alice-blue': '#ecf3fb'
```
peterpeterparker, line 38: Unrelated: At some point we should have a look if all colors are still used. It feels like there will be some that can be cleaned.

peterpeterparker (approved):
LGTM, thx

tailwind.config.js
AntonioVentilii-DFINITY: Yes, I was having a look if there is some lint-ish rule that takes care of it for CSS, but the closest one is not that good. I will have a better look after the remodelling.

In the meanwhile, I can go just searching usage eheh

PR #2424, opened 2024-09-19
style(frontend): change progress color pallet [GIX-2983]

AntonioVentilii-DFINITY:
# Motivation

New color for step in progress.

### Before

Screenshot 2024-09-19 at 11 49 08

### After

Screenshot 2024-09-19 at 11 41 33


src/frontend/src/lib/styles/global/colors.scss
```diff
@@ -30,4 +30,5 @@
 
 	// Theme
 	--color-primary: var(--color-blue-ribbon);
+	--color-primary-rgb: 0, 102, 255;
```
peterpeterparker, line 33: Should we rather define the tailwind color in rgb that way we don't hardcode the equivalent in css?

peterpeterparker: Or maybe there is a way to derive the rgb?

AntonioVentilii-DFINITY: there is a way 

`from var(--color-blue-ribbon) r g b;`

but I have to check if it's compatible all over. Let me check and come back here

tailwind.config.js
```diff
@@ -19,6 +19,7 @@ export default {
 			'light-blue': '#ede7fb',
 			blue: '#3b00b9',
 			'blue-ribbon': '#0066ff',
+			'blue-ribbon-rgb': '0, 102, 255',
```
peterpeterparker, line 22: Just sharing an idea, would following work?

1. Remove `blue-ribbon`
2. Then in CSS:

```
--color-blue-ribbon: rgb(theme(colors.blue-ribbon-rgb));
```

AntonioVentilii-DFINITY: not only it works, but even better:

```scss
	--color-primary-rgb: var(--color-blue-ribbon-rgb);
	--color-primary: rgb(var(--color-primary-rgb));
```

peterpeterparker (approved):
LGTM, thx

PR #2425, opened 2024-09-19
feat(frontend): use signer API via canister

DenysKarmazynDFINITY:
# Motivation

The idea is to create a signer canister class (instead of using actors) and refactor `signer.api` methods to use it.

### DIFF
--- a/src/frontend/src/lib/api/signer.api.ts
+++ b/src/frontend/src/lib/api/signer.api.ts
@@ -57,11 +57,9 @@ export const signPrehash = async ({
 	return sign_prehash(hash);
 };
 
-export const signerCanister = async ({
-	identity
-}: {
-	identity: Identity;
-}): Promise<SignerCanister> => {
+// TODO: implement updateBtcBalance method
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const signerCanister = async ({ identity }: { identity: Identity }): Promise<SignerCanister> => {
 	const agent = await getAgent({ identity });
 
 	return SignerCanister.create({
--- a/src/frontend/src/lib/api/signer.api.ts
+++ b/src/frontend/src/lib/api/signer.api.ts
@@ -81,7 +81,6 @@ const signerCanister = async ({
 }: {
 	identity: OptionIdentity;
 }): Promise<SignerCanister> => {
-	// TODO: Delete this dependency once the signer has been stripped down.
 	assertNonNullish(identity);
 
 	if (isNullish(canister)) {
--- a/src/frontend/src/lib/api/signer.api.ts
+++ b/src/frontend/src/lib/api/signer.api.ts
@@ -1,5 +1,4 @@
 import type { BitcoinNetwork, SignRequest } from '$declarations/signer/signer.did';
-import { getAgent } from '$lib/actors/agents.ic';
 import { SignerCanister } from '$lib/canisters/signer.canister';
 import { SIGNER_CANISTER_ID } from '$lib/constants/app.constants';
 import type { BtcAddress, EthAddress } from '$lib/types/address';
@@ -84,10 +83,8 @@ const signerCanister = async ({
 	assertNonNullish(identity);
 
 	if (isNullish(canister)) {
-		const agent = await getAgent({ identity });
-
-		canister = SignerCanister.create({
-			agent,
+		canister = await SignerCanister.create({
+			identity,
 			canisterId: Principal.fromText(SIGNER_CANISTER_ID)
 		});
 	}
--- a/src/frontend/src/lib/api/signer.api.ts
+++ b/src/frontend/src/lib/api/signer.api.ts
@@ -20,19 +20,6 @@ export const getBtcAddress = async ({
 	return getBtcAddress({ network });
 };
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-export const updateBtcBalance = async ({
-	identity,
-	network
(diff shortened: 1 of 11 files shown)


DenysKarmazynDFINITY:
@peterpeterparker I added tests for the signer canister, so the PR is ready for another review.

src/frontend/src/lib/api/signer.api.ts
```diff
@@ -1,22 +1,42 @@
...
-	const { caller_btc_address } = await getSignerActor({ identity });
-	return caller_btc_address(network);
+}): Promise<BtcAddress> => {
+	const { getBtcAddress } = await signerCanister({ identity });
+
+	return getBtcAddress({ network });
+};
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const updateBtcBalance = async ({
```
peterpeterparker, line 24: can you provide this in another PR. Let's stay focus on the canister part in this PR.

Plus it will need to use another instance of the canister if you want to use the call in a web worker.

src/frontend/src/lib/api/signer.api.ts
```diff
@@ -48,6 +70,24 @@ export const signPrehash = async ({
...
+
+	return signPrehash({ hash });
+};
+
+const signerCanister = async ({
+	identity
+}: {
+	identity: OptionIdentity;
+}): Promise<SignerCanister> => {
+	assertNonNullish(identity);
```
peterpeterparker, line 83: any particular reson why you removed the i18n key?

src/frontend/src/tests/lib/canisters/signer.canister.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+			identity: mockIdentity,
+			...services
+		});
+	const service = mock<ActorSubclass<SignerService>>();
+
+	beforeEach(() => {
+		vi.clearAllMocks();
+	});
+
+	it('returns correct BTC balance', async () => {
```
peterpeterparker, line 31: should we also add some assertions about bubbling the errors in case the service throw an error?

peterpeterparker (commented):
Looks good, thx. Just a couple of minor questions.

src/frontend/src/lib/api/signer.api.ts
DenysKarmazynDFINITY: I didn't use the i18 store due to `no-svelte-store-in-api` (and hence future usage of the signer canister in a web worker). I can move `assertNonNullish(identity)` up to the places where signer API methods are called from, if we need to keep that message. WDYT?

src/frontend/src/tests/lib/canisters/signer.canister.spec.ts
DenysKarmazynDFINITY: Good idea, added them.

src/frontend/src/lib/api/signer.api.ts
peterpeterparker: Ah right, good idea. Thx!

src/frontend/src/lib/canisters/signer.canister.ts
```diff
@@ -0,0 +1,98 @@
...
+				agent
+			},
+			idlFactory: idlFactorySigner,
+			certifiedIdlFactory: idlCertifiedFactorySigner
+		});
+
+		return new SignerCanister(canisterId, service, certifiedService);
+	}
+
+	getBtcAddress = ({
```
peterpeterparker, line 39: Actually none of those methods, `getBtcAddress` , `getEthAddress` etc. are available as query. They are all always `certified = true`. Should we simplify their interfaces accordingly?

peterpeterparker (commented):
Looks great. I had a very final look and just spotted two last questions. What do you think of those?

After that I think we will be good to merge it.

src/frontend/src/lib/canisters/signer.canister.ts
DenysKarmazynDFINITY: Yes, sounds good! Adjusted.

src/frontend/src/lib/canisters/signer.canister.ts
```diff
@@ -0,0 +1,98 @@
...
+} from '$declarations/signer/signer.did';
+import { idlFactory as idlCertifiedFactorySigner } from '$declarations/signer/signer.factory.certified.did';
+import { idlFactory as idlFactorySigner } from '$declarations/signer/signer.factory.did';
+import { getAgent } from '$lib/actors/agents.ic';
+import type { BtcAddress, EthAddress } from '$lib/types/address';
+import type { Identity } from '@dfinity/agent';
+import { Principal } from '@dfinity/principal';
+import { Canister, createServices, type CanisterOptions } from '@dfinity/utils';
+
+export interface SignerCanisterOptions<T> extends Omit<CanisterOptions<T>, 'canisterId' | 'agent'> {
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: It's not needed indeed, removed.

src/frontend/src/lib/api/signer.api.ts
DenysKarmazynDFINITY: Done.

peterpeterparker (approved):
LGTM, thanks

Maybe let's be sure the release v0.0.40 is deployed before merging this. Just being precautious to avoid that we ship this change in production without testing it thourghouly.

PR #2426, opened 2024-09-19
style(frontend): add new background [GIX-2984]

AntonioVentilii-DFINITY:
# Motivation

We introduce a new background, but only for non-small screens (not for mobile).

# Changes

- Make general background transparent.
- Create `Background` component totally white but with an additional fixed image at the top, centered.
- Add condition for just small screens.

# Tests

### Before

Screenshot 2024-09-19 at 15 43 41
Screenshot 2024-09-19 at 15 43 47

### After

https://github.com/user-attachments/assets/e855d980-c5b8-41b3-8b13-7991bcd19a28


src/frontend/src/lib/components/core/Background.svelte
```diff
@@ -0,0 +1,14 @@
...
+	import Img from '$lib/components/ui/Img.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+</script>
+
+<div class="fixed inset-0 z-[-1] overflow-hidden bg-white">
+	<Img
+		src={background}
+		alt={replaceOisyPlaceholders($i18n.core.alt.background)}
+		extraClass="min-w-[1280px] absolute left-1/2 top-0 transform -translate-x-1/2 sm:block hidden"
```
AntonioVentilii-DFINITY, line 12: Not sure if I should use `srcset` to hide the image or the condition `sm:block hidden` is ok

peterpeterparker: `srcset` in my opinion. The wallet is potentially used everything in the word and in some countries there are more phones than desktop, plus where data are not unlimited - i.e. those consumers are probably happy not to have to download an unused asset image. Makes sense?

AntonioVentilii-DFINITY: right! otherwise, they would have to load the background image always.... check last commit please

src/frontend/src/lib/i18n/en.json
```diff
@@ -19,7 +19,8 @@
 			"test_banner": "For testing purposes only!"
 		},
 		"alt": {
-			"logo": "$name logo"
+			"logo": "$name logo",
+			"background": "$oisy_name background"
```
peterpeterparker, line 23: Not an a11y expert but maybe the description can be a bit more descriptive?

src/frontend/src/lib/components/core/Background.svelte
```diff
@@ -0,0 +1,16 @@
...
+	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+</script>
+
+<div class="fixed inset-0 z-[-1] overflow-hidden bg-white">
+	<picture
+		class="absolute left-1/2 top-0 min-w-[1280px] -translate-x-1/2 transform"
+		aria-label={replaceOisyPlaceholders($i18n.core.alt.background)}
+	>
+		<source srcset={background} media="(min-width: 641px)" />
+		<Img src="" />
```
peterpeterparker, line 14: Does it throw an error in the console or it's valid to path `""` as `src`?

src/frontend/src/lib/i18n/en.json
peterpeterparker: For example yes thanks

src/frontend/src/lib/components/core/Background.svelte
peterpeterparker: Just to be sure, did you test in Firefox and Safari as well?
In worth case we can add an empty inlined base64 image I guess. Also had a look to the spec indeed not much we can do otherwise.

src/frontend/src/lib/i18n/en.json
AntonioVentilii-DFINITY: done!

src/frontend/src/lib/components/core/Background.svelte
AntonioVentilii-DFINITY: tested right now both of them and the behavior went as expected

But, I can do your suggestion and pass an empty image (I actually tried that too in one of the tests). I thought just that theoretically it is unnecessary, to have a new empty image just to pass it here, but no strong opinion, so WDYT?

peterpeterparker: Let's try with this version

peterpeterparker (approved):
LGTM, thx

PR #2431, opened 2024-09-20
refactor(frontend): fixed width for logo

AntonioVentilii-DFINITY:
# Motivation

For good form (and to avoid strange behavior during screen re-sizing), we set the width of the logo to a fixed value (that is the same as the SVG that defines it).


peterpeterparker (approved):
LGTM, thx

PR #2435, opened 2024-09-20
refactor(frontend): separate button style for icons only

AntonioVentilii-DFINITY:
# Motivation

Since we are going to change `ButtonHero` to have a different style, we create a new similar button for its usage outside strict `Hero` functions: that is typically for a button that is just an icon (for example: About menu or User menu).


src/frontend/src/lib/components/ui/ButtonAsIcon.svelte
AntonioVentilii-DFINITY, line 1: This component is just a simplified copy of `ButtonHero`

src/frontend/src/lib/components/ui/ButtonAsIcon.svelte
```diff
@@ -0,0 +1,17 @@
...
+	export let button: HTMLButtonElement | undefined = undefined;
+	export let testId: string | undefined = undefined;
+	export let ariaLabel: string;
+</script>
+
+<button
+	class="w-16 flex flex-col text-primary text-xs font-normal text-center"
+	bind:this={button}
+	on:click
+	aria-label={ariaLabel}
```
peterpeterparker, line 11: In terms of accessibility I suggest to replicate the pattern I use for example in Juno: https://github.com/junobuild/juno/blob/main/src/frontend/src/lib/components/ui/ButtonIcon.svelte

AntonioVentilii-DFINITY: you mean, passing the label slot anyway but hide it? if that is the case, should the aria-label stay?

peterpeterparker: Yes `<span class="visually-hidden"><slot /></span>`

Not sure I found it in this blog post but, few details about it: https://www.sarasoueidan.com/blog/accessible-icon-buttons/#technique-%231%3A-accessible-visually-hidden-text

AntonioVentilii-DFINITY: I think I understood! I copied the same schema as Juno (just in case we will use that `visually-hidden` class somehwere else), have a look please

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/lib/components/ui/ButtonAsIcon.svelte
peterpeterparker: Coolio. I don't think there is something similar in Tailwind out-of-the-box so makes sense to me.

PR #2436, opened 2024-09-20
test(e2e): fix current issues with e2e

DenysKarmazynDFINITY:
# Motivation

Fix issue that appeared after the `jimp` package got updated. Also, temporary increased timeout value due to tokens not being fully loaded with the default value (30s).


AntonioVentilii-DFINITY (approved):
LGTM tks!

PR #2437, opened 2024-09-20
style(frontend): new style for Authenticate button [GIX-2958]

AntonioVentilii-DFINITY:
# Motivation

Authenticate button receives a new style.

### Before

Screenshot 2024-09-20 at 10 35 02
Screenshot 2024-09-20 at 10 35 16

### After

Screenshot 2024-09-20 at 11 18 14
Screenshot 2024-09-20 at 11 18 21

### DIFF
--- a/src/frontend/src/lib/styles/global/background.scss
+++ b/src/frontend/src/lib/styles/global/background.scss
@@ -1,3 +1,0 @@
-.bg-primary {
-	background-color: var(--color-primary);
-}
--- a/src/frontend/src/lib/styles/global/background.scss
+++ b/src/frontend/src/lib/styles/global/background.scss
@@ -1,0 +1,3 @@
+.bg-primary {
+	background-color: var(--color-primary);
+}
--- a/src/frontend/src/lib/styles/global/button.scss
+++ b/src/frontend/src/lib/styles/global/button.scss
@@ -1,4 +1,5 @@
 @use '../mixins/media';
+@use '../mixins/button';
 
 button {
 	display: flex;
@@ -134,5 +135,9 @@ button {
 				}
 			}
 		}
+	}
+  
+  	&.icon {
+		@include button.icon;
 	}
 }
--- a/src/frontend/src/lib/styles/global/button.scss
+++ b/src/frontend/src/lib/styles/global/button.scss
@@ -136,8 +136,8 @@ button {
 			}
 		}
 	}
-  
-  	&.icon {
+
+	&.icon {
 		@include button.icon;
 	}
 }
--- a/src/frontend/src/lib/styles/global/button.scss
+++ b/src/frontend/src/lib/styles/global/button.scss
@@ -135,37 +135,4 @@ button {
 			}
 		}
 	}
-
-	&.authenticate {
-		position: relative;
-		z-index: 0;
-	}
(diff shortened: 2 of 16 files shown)


src/frontend/src/lib/styles/global/button.scss
```diff
@@ -135,37 +135,4 @@ button {
...
-		position: absolute;
-		content: '';
-		top: 0;
-		right: 0;
-		bottom: 0;
-		left: 0;
-		z-index: -1;
-		background: radial-gradient(
-			49.28% 49.28% at 50% 100%,
-			var(--color-turquoise) 0%,
```
peterpeterparker, line 155: Colors can be cleaned in Tailwind config?

src/frontend/src/lib/styles/global/button.scss
```diff
@@ -135,37 +135,4 @@ button {
...
-			var(--color-crayola-blue) 100%
-		);
-		opacity: 0.6;
-		box-shadow:
-			0 2px 8px 0 rgba(255, 255, 255, var(--box-shadow-opacity, 0.35)) inset,
```
peterpeterparker, line 160: Ditto for shadow variable?

src/frontend/src/lib/styles/global/background.scss
```diff
@@ -0,0 +1,3 @@
+.bg-primary {
```
peterpeterparker, line 1: Do we need a class? Can't it be resolved with tailwind? Or this will be reused?

AntonioVentilii-DFINITY: two things for this:

1. it will be re-used
2. tailwind does not understand `bg-primary` automatically and I would prefer to leave it like this if we are going to change the primary color at some point

src/frontend/src/lib/styles/global/button.scss
AntonioVentilii-DFINITY: done

src/frontend/src/lib/styles/global/button.scss
AntonioVentilii-DFINITY: this one I think it is built-in: i could not find it anywhere

src/frontend/src/lib/styles/global/background.scss
peterpeterparker: 1. Will we re-used the background or will we reused  or the button style?

2. `bg-[var(--color-primary))]`?

AntonioVentilii-DFINITY: 1. the background with primary color
2. yes, initially i did that, but I thought that maybe we would prefer have a specific class; but if we don't, I for sure can do that

peterpeterparker: I prefer not adding class. There are few that are still here for historical reason but, given that we are using tailwind, I kind of feel that we should not interfer with global style unless it's configuration. You know what I mean?

AntonioVentilii-DFINITY: Ah, then I will use your suggestion: done!

peterpeterparker (approved):
LGTM, thx

PR #2438, opened 2024-09-20
chore(frontend): correct exception to lint rule about object param

AntonioVentilii-DFINITY:
# Motivation

Correction on an "temporary" exception create by to the new custom ES lint rule introduced by PR https://github.com/dfinity/oisy-wallet/pull/2416: instead of being temporary, it should be a real exception since the function needs to have sequential arguments.


peterpeterparker (approved):
LGTM; thx

PR #2440, opened 2024-09-20
test(e2e): update waitForTokensInitialization implementation

DenysKarmazynDFINITY:
# Motivation

After the tokens list implementation changed, we need to adjust the way of how we wait for its initialisation.


AntonioVentilii-DFINITY (approved):
LGTM! Thank you for the fix!

PR #2442, opened 2024-09-20
feat(frontend): Oisy logo redirects to Home Page

AntonioVentilii-DFINITY:
# Motivation

We wrap the Oisy Logo into a button that redirects to the Home Page.

# Tests

Local replica worked as expected.


src/frontend/src/lib/components/hero/Header.svelte
```diff
@@ -20,9 +27,13 @@
 	{#if back}
 		<Back />
 	{:else}
-		<div class="flex p-4 items-center">
+		<button
```
peterpeterparker, line 30: Can we rather use `a` or is there is some issues that prevent us to use a link?

AntonioVentilii-DFINITY: ah right! i forgot that it is preferred: have a look

peterpeterparker (approved):
LGTM, thx

PR #2445, opened 2024-09-20
refactor(frontend): correct media condition in srcset

AntonioVentilii-DFINITY:
# Motivation

To be more precise and coherent with the Tailwind definition of small-screens, we adjust the `srcset` conditions to have:

- min width from `641` to `640`
- max width from `640` to `639`


peterpeterparker (approved):
LGTM, thanks

If we starting having more responsive images we should think about generalizing the approach. Just thinking out at loud for the future.

PR #2446, opened 2024-09-20
style(frontend): change background of transaction card icon

AntonioVentilii-DFINITY:
# Motivation

New background color for the icon in the transaction card. Since RoundedIcon component is used only in the transactions card, we change directly the default prop.

### Before

Screenshot 2024-09-20 at 23 47 43

### After

Screenshot 2024-09-20 at 23 47 31


peterpeterparker (approved):
LGTM, thanks

PR #2447, opened 2024-09-21
style(frontend): expand background for larger screens [GIX-2984]

AntonioVentilii-DFINITY:
# Motivation

There should not be a restriction of width for larger screens, so we remove the fixed width from the SVG file and set it as minimum width for the picture.


src/frontend/src/lib/components/core/Background.svelte
```diff
@@ -7,7 +7,7 @@
 
 <div class="fixed inset-0 z-[-1] overflow-hidden bg-white">
 	<picture
-		class="absolute left-1/2 top-0 w-full min-w-[1280px] -translate-x-1/2 transform"
+		class="absolute left-1/2 top-0 w-full min-w-[1728px] -translate-x-1/2 transform"
```
peterpeterparker, line 10: We cannot use something like min width 100vw or something?

AntonioVentilii-DFINITY: the idea is that they don't want it to re-size below the inherent width of the SVG (1728px): in those cases it should just hide in overflow. However, it can enlarge for bigger screens.

peterpeterparker: Should we create a new breakpoint in the tailwind configuration for it that way we don't have an hardcoded value?

AntonioVentilii-DFINITY: what if instead of a breakpoint, I make a custom `minWidth`? 

```
// tailwind config
extend: {
	minWidth: {
		'2_5xl': '1728px'
	}
}

// code
... w-full min-w-2_5xl ...
```

peterpeterparker: sounds good too!

AntonioVentilii-DFINITY: done!

peterpeterparker (approved):
LGTm, thx

PR #2448, opened 2024-09-21
chore(frontend): remove unused CSS color variables

AntonioVentilii-DFINITY:
# Motivation

Using a manual search, we remove the unused CSS color variables. However, they could have been overwriting the default values that are used by third-party library components (for example GIX components).

Navigating around in the local replica, there are no changes, so we will proceed.


src/frontend/src/lib/styles/global/colors.scss
```diff
@@ -4,22 +4,16 @@
 	--color-white: theme(colors.white);
 	--color-grey: theme(colors.grey);
 	--color-dust: theme(colors.dust);
-	--color-light-blue: theme(colors.light-blue);
```
AntonioVentilii-DFINITY, line 7: this one is not removes from Tailwind config because it is used as `bg-light-blue` in the code.

peterpeterparker (approved):
LGTM, thx

PR #2449, opened 2024-09-21
fix(frontend): reactive string for Alpha warning position

AntonioVentilii-DFINITY:
# Motivation

Some browsers do not evaluate correctly the string at runtime: `${mw}` is treated as a string interpolation rather than a logic-based evaluation. So, when `mw` changes, the template doesn't correctly update the classes. Sometimes the dynamic classes are even ignored.

To correct it, we transform `mw` into a derived of `authSignedIn`. The alternative would be to put the reactive statement in each of those classes.

### DIFF
--- a/src/frontend/src/lib/components/hero/Header.svelte
+++ b/src/frontend/src/lib/components/hero/Header.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+	import { derived, type Readable } from 'svelte/store';
 	import WalletConnect from '$eth/components/wallet-connect/WalletConnect.svelte';
 	import Alpha from '$lib/components/core/Alpha.svelte';
 	import Back from '$lib/components/core/Back.svelte';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/hero/Header.svelte
```diff
@@ -11,13 +11,10 @@
...
 
 	export let back = false;
-
-	let mw: 'sm' | 'xl';
-	$: mw = $authSignedIn ? 'sm' : 'xl';
 </script>
 
 <header
-	class={`grid grid-cols-2 ${mw}:grid-cols-[1fr_auto_1fr] items-center md:px-4 relative z-10 pointer-events-none`}
+	class={`grid grid-cols-2 ${$authSignedIn ? 'sm:grid-cols-[1fr_auto_1fr]' : 'xl:grid-cols-[1fr_auto_1fr]'} items-center md:px-4 relative z-10 pointer-events-none`}
```
peterpeterparker, line 17: I would use conditional class then.

```
<header
   class="grid grid-cols-2..."
   class:sm:grid-cols-[1fr_auto_1fr]={$authSignedIn}
   class:xl:grid-cols-[1fr_auto_1fr]={$authNotSignedIn}
```

WDYT?

AntonioVentilii-DFINITY: ah yes! maybe a bit verbose but more organized

peterpeterparker: yeah. ultimately tailwind is bit verbose anyway...

AntonioVentilii-DFINITY: done

peterpeterparker (approved):
LGTM, thx

PR #2462, opened 2024-09-23
chore(build): remove three from chunk-able dependencies

AntonioVentilii-DFINITY:
# Motivation

Dependency `three` was deprecated so we don't need to chunk it anymore.


peterpeterparker (approved):
LGTM, thx

PR #2463, opened 2024-09-23
feat(frontend): add updateBtcBalance function

DenysKarmazynDFINITY:
# Motivation

Implement a function for getting certified BTC balance. The implementation on the Signer side - https://github.com/dfinity/chain-fusion-signer/blob/0d611dd8e18d2d83a6e8ae1577ba762cc500c081/src/signer/canister/src/lib.rs#L140

### DIFF
--- a/src/frontend/src/lib/api/signer.api.ts
+++ b/src/frontend/src/lib/api/signer.api.ts
@@ -2,6 +2,7 @@ import type { BitcoinNetwork, SignRequest } from '$declarations/signer/signer.di
 import { SignerCanister } from '$lib/canisters/signer.canister';
 import { SIGNER_CANISTER_ID } from '$lib/constants/app.constants';
 import type { BtcAddress, EthAddress } from '$lib/types/address';
+import type { CanisterIdText } from '$lib/types/canister';
 import type { OptionIdentity } from '$lib/types/identity';
 import { Principal } from '@dfinity/principal';
 import { assertNonNullish, isNullish } from '@dfinity/utils';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/api/signer.api.ts
```diff
@@ -30,6 +30,17 @@ export const getEthAddress = async ({ identity }: CommonParams): Promise<EthAddr
 	return getEthAddress();
 };
 
+export const updateBtcBalance = async ({
+	identity,
+	network
+}: CommonParams<{
+	network: BitcoinNetwork;
+}>): Promise<bigint> => {
+	const { updateBtcBalance } = await signerCanister({ identity });
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Good point. I added an optional `signerCanisterId` param to the API methods. It defaults to the value from `import.meta.env` so we can only pass it from the functions that are gonna be used in a worker (for now, only BTC balance method).

src/frontend/src/lib/api/signer.api.ts
```diff
@@ -30,6 +32,18 @@ export const getEthAddress = async ({ identity }: CommonParams): Promise<EthAddr
 	return getEthAddress();
 };
 
+export const updateBtcBalance = async ({
```
peterpeterparker, line 35: Just noticing it now, sorry did not picked it on first review, why is it prefixed with `update...`? why not just `btcBalance` or `getBtcBalance`?

DenysKarmazynDFINITY: The idea was to highlight that it's a certified balance (on the contrary to the `query` balance method available in `bitcoin.api`). However, now that I think about it again, `getBtcBalance` makes much more sense since the "certified" part is already implied by the use of the signer canister. Renaming.

peterpeterparker (approved):
LGTM, thx

PR #2464, opened 2024-09-23
refactor(devops): Get signer.did path from release

bitdivine:
# Motivation
The `signer.did` file is included in new cf signer releases, as it should be.  It can be added manually to old releases.  Previously the candid file was taken from within the codebase but that breaks across versions (when the file moves).  Also the release files should contain all files needed to work with that version; users should not have to search through the codebase to find the candid.

# Changes
- Update the download URL for the `signer.did`.

# Tests
See CI


AntonioVentilii-DFINITY (approved):
LGTM! tks

PR #2466, opened 2024-09-23
style(frontend): define new secondary color

AntonioVentilii-DFINITY:
# Motivation

We deprecate `--color-dark` in favor of `--color-black`, and we define it as `secondary` color.

Visually there is not much difference since we change from a very dark blue to black.

### Before

Screenshot 2024-09-23 at 20 18 57
Screenshot 2024-09-23 at 20 19 05
Screenshot 2024-09-23 at 20 19 29
Screenshot 2024-09-23 at 20 19 40
Screenshot 2024-09-23 at 20 20 03

### After

Screenshot 2024-09-23 at 20 20 56
Screenshot 2024-09-23 at 20 21 05
Screenshot 2024-09-23 at 20 21 23
Screenshot 2024-09-23 at 20 21 30
Screenshot 2024-09-23 at 20 21 37
Screenshot 2024-09-23 at 20 21 44
Screenshot 2024-09-23 at 20 21 50


peterpeterparker:
Can you share a screenshot (before/after) for a generally understanding?

AntonioVentilii-DFINITY:
@peterpeterparker you are right, I forgot to add screenshots! Included in the description with a comment (basically, the two colors are very very similar)

peterpeterparker (commented):
I give it some thought and actually not sure "black" should become "secondary"

1. black is not a color per sé
2. black is a base color - i.e. `text-black` is standard Tailwind

So I was wondering if instead of adding "text-secondary" we should not rather just replace `text-dark` by `text-black`. WDYT?

AntonioVentilii-DFINITY:
The purpose would be to make it easy to change it when we will have theme, since it was decided that in the pallet, black is the secondary color for the default theme. We can always substitute later if you want

peterpeterparker (approved):
LGTM; thanks

PR #2470, opened 2024-09-23
feat(frontend): add Onramper ID for tokens

AntonioVentilii-DFINITY:
# Motivation

Another step to onboard Onramper is to map the correspondent Crypto ID to each token based [on this list](https://docs.onramper.com/docs/crypto-asset-support).

For now, we will set it just for ICP and ETH, but, after having tested more, map the rest.

# Changes

- Create new type to define the mapping to Crypto ID.
- Adjust RequiredToken type to excepts new type TokenBuyable.
- Set values for ICP and ETH.

### DIFF
--- a/src/frontend/src/lib/types/onramper.ts
+++ b/src/frontend/src/lib/types/onramper.ts
@@ -1,3 +1,7 @@
 // The list of networks that are supported by OnRamper can be found here:
 // https://docs.onramper.com/docs/network-support
 export type OnRamperNetworkId = 'icp' | 'bitcoin' | 'ethereum';
+
+// The list of cryptocurrencies that are supported by Onramper can be found here:
+// https://docs.onramper.com/docs/crypto-asset-support
+export type OnramperCryptoId = string;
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/types/onramper.ts
```diff
@@ -1,3 +1,7 @@
 // The list of networks that are supported by OnRamper can be found here:
 // https://docs.onramper.com/docs/network-support
 export type OnRamperNetworkId = 'icp' | 'bitcoin' | 'ethereum';
+
+// The list of cryptocurrencies that are supported by Onramper can be found here:
+// https://docs.onramper.com/docs/crypto-asset-support
+export type OnramperCryptoId = string;
```
peterpeterparker, line 7: What about just `OnramperId` given that's the term used in the linked documentation?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/types/onramper.ts
AntonioVentilii-DFINITY: We will have a `OnRamperFiatId` too, I wanted to distinguish, even if they are both strings in the end, WDYT?

peterpeterparker: Honestly? When I open the link provided in the comment, https://docs.onramper.com/docs/crypto-asset-support, I started looking for the keyword "crypto" to find the ID

AntonioVentilii-DFINITY: ah! yes.. make sense, changed!

peterpeterparker (approved):
LGTM, thx

PR #2455, opened 2024-09-23
style(frontend): adjust spacing in layout

AntonioVentilii-DFINITY:
# Motivation

The spacing was a bit odd with the new UI, so we adjust the margins and the padding according to design. Furthermore, we limit the header width to a maximum, to avoid expanding too much on larger screens.

### Logged out

Screenshot 2024-09-23 at 08 35 03
Screenshot 2024-09-23 at 08 35 13
Screenshot 2024-09-23 at 08 35 22
Screenshot 2024-09-23 at 08 35 30
Screenshot 2024-09-23 at 08 35 46

### Logged in

Screenshot 2024-09-23 at 08 36 10
Screenshot 2024-09-23 at 08 36 21
Screenshot 2024-09-23 at 08 36 32
Screenshot 2024-09-23 at 08 36 43
Screenshot 2024-09-23 at 08 36 52

### Transaction page

Screenshot 2024-09-23 at 08 37 06
Screenshot 2024-09-23 at 08 37 17
Screenshot 2024-09-23 at 08 37 26
Screenshot 2024-09-23 at 08 37 49
Screenshot 2024-09-23 at 08 37 58

### With WalletConnect

Screenshot 2024-09-23 at 08 51 11
Screenshot 2024-09-23 at 08 51 23
Screenshot 2024-09-23 at 08 51 34
Screenshot 2024-09-23 at 08 51 43
Screenshot 2024-09-23 at 08 51 55

PR #2471, opened 2024-09-23
fix(frontend): backdrop should have no margin

AntonioVentilii-DFINITY:
# Motivation

With PR #2455 , there was a glitch with the backdrop of the modals, that would inherit the spacing forced by the `Hero` component.

### Before

Screenshot 2024-09-23 at 22 10 49

### After

Screenshot 2024-09-23 at 22 10 36


src/frontend/src/lib/components/hero/Hero.svelte
```diff
@@ -14,11 +14,11 @@
 	$: heroContent = usdTotal || summary;
 </script>
 
-<div class="pt-6 space-y-10">
+<div class="pt-6">
 	<Header back={back === 'header'} />
 
 	<article
-		class="flex flex-col rounded-lg pb-6 relative main 2xl:mt-[-70px] items-center"
+		class="flex flex-col rounded-lg pt-10 pb-6 relative main 2xl:mt-[-70px] items-center"
```
peterpeterparker, line 21: Have you double check that modifying this does not lead to any conflicts with everything single other modals and overlays used in the app?

AntonioVentilii-DFINITY: yes yes, checked a few modals; but even because I introduced that `space-y-10`  and removed `pt-` from `article`, in PR #2455 , and it was all good before

peterpeterparker: ok

peterpeterparker (approved):
LGTm, thx

PR #2473, opened 2024-09-23
refactor(frontend): blurred transactions page when signed out

AntonioVentilii-DFINITY:
# Motivation

The transactions page should be blurred like the main page when the user is not signed in.

# Changes

- Created new component to blur a slot.
- Use the component for `Transactions`.
- Replace and use it in the code of `Tokens`, to avoid repetition.

# Tests

### Before

Screenshot 2024-09-23 at 22 42 05
Screenshot 2024-09-23 at 22 42 12
Screenshot 2024-09-23 at 22 42 18

### After

Screenshot 2024-09-23 at 22 41 38
Screenshot 2024-09-23 at 22 41 46
Screenshot 2024-09-23 at 22 41 53

### DIFF
--- a/src/frontend/src/lib/components/tokens/Tokens.svelte
+++ b/src/frontend/src/lib/components/tokens/Tokens.svelte
@@ -5,10 +5,11 @@
 	import TokensSignedIn from '$lib/components/tokens/TokensSignedIn.svelte';
 	import TokensSignedOut from '$lib/components/tokens/TokensSignedOut.svelte';
 	import Header from '$lib/components/ui/Header.svelte';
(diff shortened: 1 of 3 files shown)


peterpeterparker (commented):
I don't think that's a good idea. The blurred text says "Sign-in to see your transactions". If we blur it, the text does not make any sense anymore.

AntonioVentilii-DFINITY:
Let me talk it again with design team

AntonioVentilii-DFINITY:
As discussed with Dan, we will have a specific "empty-state" component as placeholder for these cases, so for now we leave it as it is. Closing

PR #2477, opened 2024-09-24
feat(frontend): move scheduler to lib

DenysKarmazynDFINITY:
# Motivation

The idea is to move the SchedulerTimer class to lib so it can be reused by BTC worker.


AntonioVentilii-DFINITY (approved):
LGTM, tks!

PR #2478, opened 2024-09-24
fix(frontend): solve glitch when token logo slides out [GIX-2973]

AntonioVentilii-DFINITY:
# Motivation

There was a glitch introduced by the new UI: when the user goes from the transactions page back to the main page, the logo would show a placeholder, since the `token` store would be set to `null`.

My solution is to render the logo ONLY if `token` store is non-nullish, so that instead there is no placeholder but a simple logo skeleton.

### Before

https://github.com/user-attachments/assets/ac9ec094-4d3c-484c-9672-0d42c30ea498

### After

https://github.com/user-attachments/assets/05dc7d02-d7eb-456d-a38f-68c010be2794


peterpeterparker (approved):
LGTM, thx

PR #2484, opened 2024-09-24
feat(frontend): rename syncWalletStatus to syncWalletIcStatus

DenysKarmazynDFINITY:
# Motivation

Since we are gonna have 2 WalletWorkers, we need to adjust the existing `PostMessageResponseStatus.syncWalletStatus` to make it specific for the IC wallet. The new one for BTC will be called `syncBtcWalletStatus`.


AntonioVentilii-DFINITY (approved):
LGTM!

PR #2485, opened 2024-09-24
feat(frontend): reuse PostMessageWalletData interface for BTC and IC

DenysKarmazynDFINITY:
# Motivation

The idea is to re-use wallet worker interfaces as much as possible. For now, BTC will have a separate type, but after I start working on adding transactions to the BTC worker, `PostMessageDataResponseBtcWallet` can be deleted.


AntonioVentilii-DFINITY (approved):
LGTM!

peterpeterparker:
Unrelated but, somehow related: we have to convert the post message types to Zod and add validations

DenysKarmazynDFINITY:
@peterpeterparker good idea, noted 👍

PR #2488, opened 2024-09-24
feat(frontend): Onramper util to create the widget source link [GIX-2854]

AntonioVentilii-DFINITY:
# Motivation

We create an util to generate the source link to provide to the Onramper widget, given a set of parameters. Documentation can be found [here](https://docs.onramper.com/docs/supported-widget-parameters).

For now we define a small set of FIAT currencies, just for the scope of testing. The list can be expanded or improved after.

# Changes

- Create ENV variables for Onramper API keys.
- Add types for params of the new util.
- Create utils to build the source code and concatenate all params into a single string.
- Create tests.

# Tests

Create some tests to verify the correct widget link: the links were manually tested in the [Onramper sandbox](https://docs.onramper.com/docs/customise-the-ux) and simply opening them in the browser with the test API key.


peterpeterparker (commented):
Did not reviewed the PR, got a question before it:

Is there an environment where we have both onramper dev and prod? It's not one or the other?

AntonioVentilii-DFINITY:
> It's not one or the other?

In staging and local, at some point, I would like to have the choice to have both, because we should test Oisy with both Onramper env, I think. The dev env of Onramper has more feature enabled that the prod (for example more payment solutions, different rates providers, etc).

That is why I decided to have both keys in `.env` file.

You think not necessary and we stick just with one key, until we really need to test Onramper PROD?

peterpeterparker:
Oki doki works for me

peterpeterparker (approved):
LGTM, thx

PR #2489, opened 2024-09-24
feat(frontend): BTC wallet services

DenysKarmazynDFINITY:
# Motivation

The goal is to implement BTC wallet services. Worker and scheduler will be merged in a separate PR. 

In the final PR of this feature, `BtcLoaderWallets` will be added to the list of all loaders.

### DIFF
--- a/src/frontend/src/btc/services/worker.btc-wallet.services.ts
+++ b/src/frontend/src/btc/services/worker.btc-wallet.services.ts
@@ -1,5 +1,5 @@
 import { syncWallet } from '$btc/services/btc-listener.services';
-import { isNetworkIdBTCMainnet, isNetworkIdBTCTestnet } from '$icp/utils/ic-send.utils';
+import { isNetworkIdBTCRegtest, isNetworkIdBTCTestnet } from '$icp/utils/ic-send.utils';
 import type { WalletWorker } from '$lib/types/listener';
 import type { PostMessage, PostMessageDataResponseBtcWallet } from '$lib/types/post-message';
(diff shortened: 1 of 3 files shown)


peterpeterparker (commented):
Is there a way to split things in two PRs?

DenysKarmazynDFINITY:
@peterpeterparker let's start reviewing scheduler + worker (https://github.com/dfinity/oisy-wallet/pull/2493), and then come back to this PR with related services.

DenysKarmazynDFINITY:
@peterpeterparker this one is now ready for review.

src/frontend/src/btc/services/worker.btc-wallet.services.ts
```diff
@@ -0,0 +1,56 @@
...
+		}
+	};
+
+	return {
+		start: () => {
+			worker.postMessage({
+				msg: 'startBtcWalletTimer',
+				data: {
+					bitcoinNetwork: mapToSignerBitcoinNetwork({
+						network: isNetworkIdBTCMainnet(networkId)
```
peterpeterparker, line 36: I always feel like it's safer to default to mainnet / production rather than testnet. Always fear that something goes sideways in the build and we endup having test on prod. WDYT?

DenysKarmazynDFINITY: Yes, it does make sense. I adjusted it accordingly.

peterpeterparker (approved):
LGTm, thx

PR #2490, opened 2024-09-24
style(frontend): change scrollbar style [GIX-3019]

AntonioVentilii-DFINITY:
# Motivation

New scrollbar style.

### Before

https://github.com/user-attachments/assets/87c7491b-c779-49bf-aec7-a4bfee3d656f

### After

https://github.com/user-attachments/assets/343a5994-41ac-4b1b-a95b-86f5f6c2cfbe

### DIFF
--- a/src/frontend/src/lib/styles/global/colors.scss
+++ b/src/frontend/src/lib/styles/global/colors.scss
@@ -2,6 +2,7 @@
 	// Figma color names
 	--color-off-white: theme(colors.off-white);
 	--color-white: theme(colors.white);
+	--color-white-rgb: theme(colors.white-rgb);
 	--color-black-rgb: theme(colors.black-rgb);
 	--color-grey: theme(colors.grey);
 	--color-dust: theme(colors.dust);
--- a/src/frontend/src/lib/styles/global/theme.scss
+++ b/src/frontend/src/lib/styles/global/theme.scss
@@ -121,6 +121,7 @@ figcaption {
 	color: var(--color-off-white);
 }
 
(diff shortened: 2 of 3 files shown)


peterpeterparker (commented):
Did you test in Chrome, Safari and Firefox?

AntonioVentilii-DFINITY:
- Chrome: yes, it's the video in the description
- Safari: i reduced a bit the border opacity, jut to make it more immediate (it is not so clear as before, but it is noticable)
- Firefox: we don't set the scrollbar color, to do that we need to change the native one:

```
html {
    scrollbar-width: ...
    scrollbar-color: ....
}
```

shall I do it here or in another PR?

peterpeterparker:
Let's do everything here. Thanks for double checking.

AntonioVentilii-DFINITY:
@peterpeterparker I am having an issue setting it for Foirefox (and I think that is the reason why we don't do it):

according to [documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scrollbars_styling), there is not much you can change with the scrollbar: in our case only color and width. However the width has fixed parameter, and if I set it and the color, it will overwrite the rest:

```css
::-webkit-scrollbar-thumb {
	background-color: var(--color-misty-rose);
}

::-webkit-scrollbar-track {
	border-radius: 0;
	background: none;
}

::-webkit-scrollbar {
	width: 0.45rem;
}
```

any idea? I tried a few things, but it seems a know behavior

peterpeterparker:
Thanks for trying. Let's just implement the best we can for Firefox (or nothing if nothing is posible) and we review it later if necessary.

AntonioVentilii-DFINITY:
@peterpeterparker perfect, then in this case, I will leave it as it is, but I added a TODO for the future

peterpeterparker (approved):
LGTM, thx

PR #2492, opened 2024-09-24
feat(frontend): extend exchange.derived with BTC regtest

DenysKarmazynDFINITY:
# Motivation

Extend exchange.derived with BTC testnet and regtest in order to calculate `usdBalance` for those tokens.


AntonioVentilii-DFINITY (commented):
So, I agree for `regtest`, but I am not really sure that is completely correct for `testnet`. IIRC, there was a market for testnet BTC at some point: is it still the case?

DenysKarmazynDFINITY:
I can see that there is indeed a market for testnet, but I somehow can't find it in coingecko API. In any case, let's use this PR for regtest only, will do the same for testnet separately.

AntonioVentilii-DFINITY (approved):
LGTM!

PR #2493, opened 2024-09-24
feat(frontend): BTC wallet scheduler + worker

DenysKarmazynDFINITY:
# Motivation

PR with the BTC wallet scheduler and worker. Services will be merged separately.

### DIFF
--- a/src/frontend/src/btc/schedulers/wallet.scheduler.ts
+++ b/src/frontend/src/btc/schedulers/wallet.scheduler.ts
@@ -1,0 +1,58 @@
+import { getBtcBalance } from '$lib/api/signer.api';
+import { WALLET_TIMER_INTERVAL_MILLIS } from '$lib/constants/app.constants';
+import { SchedulerTimer, type Scheduler, type SchedulerJobData } from '$lib/schedulers/scheduler';
+import type {
+	PostMessageDataRequestBtc,
+	PostMessageDataResponseBtcWallet
+} from '$lib/types/post-message';
+import { assertNonNullish } from '@dfinity/utils';
+
+export class WalletScheduler implements Scheduler<PostMessageDataRequestBtc> {
+	private timer = new SchedulerTimer('syncBtcWalletStatus');
+
+	stop() {
+		this.timer.stop();
+	}
+
+	async start(data: PostMessageDataRequestBtc | undefined) {
+		await this.timer.start<PostMessageDataRequestBtc>({
+			interval: WALLET_TIMER_INTERVAL_MILLIS,
+			job: this.syncWallet,
+			data
+		});
+	}
+
+	async trigger(data: PostMessageDataRequestBtc | undefined) {
+		await this.timer.trigger<PostMessageDataRequestBtc>({
+			job: this.syncWallet,
+			data
+		});
+	}
+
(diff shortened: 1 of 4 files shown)


src/frontend/src/btc/schedulers/wallet.scheduler.ts
```diff
@@ -0,0 +1,58 @@
+import { getBtcBalance } from '$lib/api/signer.api';
+import { WALLET_TIMER_INTERVAL_MILLIS } from '$lib/constants/app.constants';
+import { SchedulerTimer, type Scheduler, type SchedulerJobData } from '$lib/schedulers/scheduler';
+import type {
+	PostMessageDataRequestBtc,
+	PostMessageDataResponseBtcWallet
+} from '$lib/types/post-message';
+import { assertNonNullish } from '@dfinity/utils';
+
+export class WalletScheduler implements Scheduler<PostMessageDataRequestBtc> {
```
peterpeterparker, line 10: `btc-wallet.scheduler.ts` and `BtcWalletScheduler` maybe? Can be handy when we lookup for the module

src/frontend/src/btc/schedulers/wallet.scheduler.ts
```diff
@@ -0,0 +1,58 @@
...
+	private syncWallet = async ({ identity, data }: SchedulerJobData<PostMessageDataRequestBtc>) => {
+		const bitcoinNetwork = data?.bitcoinNetwork;
+
+		assertNonNullish(bitcoinNetwork, 'No BTC network provided to get certified balance.');
+
+		const balance = await getBtcBalance({
+			identity,
+			network: bitcoinNetwork
+		});
+
```
peterpeterparker, line 41: Can you add some TODOs or comments for the missing logic part? Assuming we also want to fetch transactions, etc., that would be handy to review the PR and ensure it will support what we ultimately implement.

DenysKarmazynDFINITY: Added a todo comment for fetching transactions (next steps after balance is done) and querying uncertified balance (will be done after the main functionality is implemented).

src/frontend/src/btc/schedulers/wallet.scheduler.ts
DenysKarmazynDFINITY: https://github.com/dfinity/oisy-wallet/pull/2498

src/frontend/src/btc/schedulers/wallet.scheduler.ts
peterpeterparker: Thanks.

What’s the plan? Are we going to ensure that we always have the balance and transactions in sync with certification, or do we plan to keep them separate? For example, displaying a certified balance but uncertified transactions? If in sync, are we going to ensure this at the JS level or through a single endpoint on the backend side? One or the other or if it's unclear, should we also add a TODO for this particular topic?

DenysKarmazynDFINITY: The first step would be to get uncertified transactions via bitcoin transactions API. It's still an open question how are we gonna get them certified (most likely, a BE endpoint). I will extend my TODO comment with this information.

peterpeterparker (approved):
LGTM, thx

PR #2495, opened 2024-09-24
feat(backend): Fund use of paid signer API

bitdivine:
# Motivation
The chain fusion signer has paid endpoints.  To use these, users need to get an ICRC-2 approval from the backend.

# Changes
- Add a method to call ICRC-2 approve for a given user.

# Tests
This flow is tested extensively in the PAPI repository: https://github.com/dfinity/papi/blob/main/src/example/paid_service/tests/it/patron_pays_icrc2_cycles.rs#L29

Once the frontend calls this method and a paid endpoint, we can also have an integration/e2e test.


src/backend/src/lib.rs
```diff
@@ -398,6 +400,11 @@ fn get_user_profile() -> Result<UserProfile, GetUserProfileError> {
     })
 }
 
+#[query(guard = "may_read_user_data")]
+async fn allow_signing() -> Result<(), AllowSigningError> {
```
lmuntaner, line 404: Will this be used only for signing?

We also use the CFS to get the address and balance for example.

src/backend/src/signer.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/// Enables the user to sign transactions.
+///
+/// Signing costs cycles.  Managing that cycle payment can be painful so we take care of that.
+pub async fn allow_signing() -> Result<(), AllowSigningError> {
+    let cycles_ledger: Principal = *CYCLES_LEDGER;
+    let signer: Principal = *SIGNER;
+    let caller = ic_cdk::caller();
+    let expected_fee = 1_000_000_000u64;
```
lmuntaner, line 22: Can we create it as a const and add a reference to why is it `1_000_000_000`?

Or is this even the fee or the units of TCycles?

src/backend/src/signer.rs
```diff
@@ -0,0 +1,60 @@
...
+        .await
+        .map_err(|_| AllowSigningError::FailedToContactCyclesLedger)?
+        .0
+        .map_err(AllowSigningError::ApproveError)?;
+    Ok(())
+}
+
+const SUB_ACCOUNT_ZERO: Subaccount = Subaccount([0; 32]);
+#[must_use]
+pub fn principal2account(principal: &Principal) -> ByteBuf {
```
lmuntaner, line 47: Why is it public?

src/backend/src/signer.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// Enables the user to sign transactions.
+///
+/// Signing costs cycles.  Managing that cycle payment can be painful so we take care of that.
+pub async fn allow_signing() -> Result<(), AllowSigningError> {
+    let cycles_ledger: Principal = *CYCLES_LEDGER;
+    let signer: Principal = *SIGNER;
+    let caller = ic_cdk::caller();
+    let expected_fee = 1_000_000_000u64;
+    let amount = Nat::from(100 * expected_fee);
```
lmuntaner, line 23: Why 100? Why multiplied the expected fee, is the expected_fee the conversion to TCycles?

src/backend/src/lib.rs
bitdivine: Done.

src/backend/src/signer.rs
bitdivine: Done

src/backend/src/signer.rs
bitdivine: Updated the signer fee to the value it will most likely be when th ecfs goes live.

Updated the number of ops to a new arbitrary value of 30, in view of how expensive the operations are.

src/backend/src/signer.rs
bitdivine: If it is public, the editor will make it obvious if there are two copies of the same function in the codebase.

Note: I merged a PR upstream that provides an account identifier as bytes directly.  There is a Jira ticket to replace all of these `principal2account()` calls.  That can be done after the next ic-cdk release.

lmuntaner: Thanks for the explanation

src/backend/src/signer.rs
```diff
@@ -0,0 +1,84 @@
...
+const SIGNER_FEE: u64 = 80_000_000_000;
+/// A reasonable number of signing operations per user per login.
+///
+/// Projected uses:
+/// - Getting Ethereum address (1x per login)
+/// - Getting Bitcoin address (1x per login)
+/// - Signing operations (10x per login)
+///
+/// Margin of error: 3x (given  that the signer fee is subject to change in the next few days and weeks)
+const SIGNING_OPS_PER_LOGIN: u64 = 36;
```
lmuntaner, line 33: I guess we could move this to the config in another PR. Definitely not in this one.

lmuntaner (approved):
Thanks for the extra comments!!

PR #2496, opened 2024-09-25
feat(frontend): enable pointer events in header

AntonioVentilii-DFINITY:
# Motivation

The header had disabled pointer events (PR #121 ), however it seems like now we can enable again. This even fixes the issue that the Alpha warning was not clickable.


peterpeterparker (approved):
LGTM, thanks

PR #2500, opened 2024-09-25
feat(frontend): BTC loader wallets component

DenysKarmazynDFINITY:
# Motivation

The goal of this PR is to initialise BTC wallet workers. Should be merged after https://github.com/dfinity/oisy-wallet/pull/2489.

### DIFF
--- a/src/frontend/src/btc/components/core/BtcLoaderWallets.svelte
+++ b/src/frontend/src/btc/components/core/BtcLoaderWallets.svelte
@@ -1,0 +1,21 @@
+<script lang="ts">
+	import { initBtcWalletWorker } from '$btc/services/worker.btc-wallet.services';
+	import { BTC_REGTEST_NETWORK_ID } from '$env/networks.env';
+	import LoaderWallets from '$lib/components/core/LoaderWallets.svelte';
+	import { LOCAL } from '$lib/constants/app.constants';
+	import { enabledBtcTokens } from '$lib/derived/tokens.derived';
+	import type { InitWalletWorkerFn } from '$lib/types/listener';
+
+	// Locally, only the Regtest worker has to be launched, in all other envs - testnet and mainnet
+	const tokens = $enabledBtcTokens.filter((token) =>
+		LOCAL
+			? token.network.id === BTC_REGTEST_NETWORK_ID
+			: token.network.id !== BTC_REGTEST_NETWORK_ID
+	);
+
+	const initWalletWorker: InitWalletWorkerFn = ({ token }) => initBtcWalletWorker(token);
+</script>
+
+<LoaderWallets {tokens} {initWalletWorker}>
+	<slot />
+</LoaderWallets>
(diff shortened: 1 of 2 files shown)


src/frontend/src/btc/components/core/BtcLoaderWallets.svelte
```diff
@@ -0,0 +1,21 @@
+<script lang="ts">
+	import { initBtcWalletWorker } from '$btc/services/worker.btc-wallet.services';
+	import { BTC_REGTEST_NETWORK_ID } from '$env/networks.env';
+	import LoaderWallets from '$lib/components/core/LoaderWallets.svelte';
+	import { LOCAL } from '$lib/constants/app.constants';
+	import { enabledBtcTokens } from '$lib/derived/tokens.derived';
+	import type { InitWalletWorkerFn } from '$lib/types/listener';
+
+	// Locally, only the Regtest worker has to be launched, in all other envs - testnet and mainnet
+	const tokens = $enabledBtcTokens.filter((token) =>
```
peterpeterparker, line 10: shouldn't `tokens` be reactive? i.e. each time `$enabledBtcTokens` changes, the worker need to be disconnected, tokens filtered and then worker mounted?

src/frontend/src/lib/components/core/Loaders.svelte
```diff
@@ -11,11 +12,13 @@
 	<Loader>
 		<LoaderBalances>
 			<IcLoaderWallets>
-				<ExchangeWorker>
-					<LoaderMetamask>
-						<slot />
-					</LoaderMetamask>
-				</ExchangeWorker>
+				<BtcLoaderWallets>
```
peterpeterparker, line 15: Maybe, in a future PR, we can extract a component `LoaderWallets` in which we use `<IcLoaderWallets>` and `<BtcLoaderWallets>` that way here we consume just one component?

DenysKarmazynDFINITY: Sounds good, added to my todo list.

src/frontend/src/btc/components/core/BtcLoaderWallets.svelte
DenysKarmazynDFINITY: In theory - yes, we should do it exactly like you described. The thing with `enabledBtcTokens` is that they cannot be disabled by users. 

Having said that, I realised using `filterEnabledTokens` for `enabledBtcTokens` derived is actually redundant. So I can either use `btc/tokens.derived/enabledBitcoinTokens` directly or simplify the value from `lib/tokens.derived (to be consistent with how it's done for the IC wallet):
```
export const enabledBtcTokens: Readable<Token[]> = derived(
	[enabledBitcoinTokens],
	([$enabledBitcoinTokens]) => $enabledBitcoinTokens
);
```

WDYT?

peterpeterparker: I'm not sure my feedback was clear. Let me try to rephrase:

My point is that there's a potential race condition if the component is mounted before the derived store is populated.

So instead of `const tokens =`, I think we need a reactive variable `let token: Something; $: tokens = ...`

DenysKarmazynDFINITY: My bad, your first comment was clear. The `token` list is now reactive.

peterpeterparker (approved):
LGTM, thx

PR #2502, opened 2024-09-25
style(frontend): new style for modals [GIX-3021]

AntonioVentilii-DFINITY:
# Motivation

The modal style changes: we remove the dark frame and we adjust the layout a bit. Furthermore, for mobile version we leave a bit of margin to sho the backdrop.

UNRELATED: since we are creating a new CSS variable for `--padding-2_5x`, we use it in another similar instance.

### Before

Screenshot 2024-09-25 at 13 29 26
Screenshot 2024-09-25 at 13 27 48
Screenshot 2024-09-25 at 13 28 27

### After

Desktop
Screenshot 2024-09-25 at 13 16 45
Screenshot 2024-09-25 at 13 17 10
Screenshot 2024-09-25 at 13 17 17

Mobile with large screen
Screenshot 2024-09-25 at 13 17 33
Screenshot 2024-09-25 at 13 17 43

Mobile with small screen
Screenshot 2024-09-25 at 13 18 05
Screenshot 2024-09-25 at 15 01 50

### DIFF
--- a/src/frontend/src/lib/styles/global/gix.scss
+++ b/src/frontend/src/lib/styles/global/gix.scss
@@ -127,6 +127,7 @@ div.modal {
 	}
 
 	div.wrapper.dialog div.header {
+		// TODO: Improve padding definition when the Modal of GIX Components has an updated way of setting it and not being hard-coded (https://github.com/dfinity/gix-components/blob/1c4ab390f9cab1d0e3ec73a23384e045679eb6b8/src/lib/components/Modal.svelte#L195)
 		--padding-3x: 0;
 		--padding: var(--padding-2x);
 		--dialog-padding-y: 0;
--- a/src/frontend/src/lib/styles/global/gix.scss
+++ b/src/frontend/src/lib/styles/global/gix.scss
@@ -141,7 +141,6 @@ div.modal {
 
 		div.container {
 			border-radius: 0;
-			padding-right: var(--scrollbar-width);
 
 			div.content {
 				padding-inline: var(--padding-2_5x);
--- a/src/frontend/src/lib/styles/global/gix.scss
+++ b/src/frontend/src/lib/styles/global/gix.scss
@@ -131,11 +131,10 @@ div.modal {
 		--padding: var(--padding-2x);
 		--dialog-padding-y: 0;
 
-		margin-inline:  calc( 3 / 2 * var(--padding));
+		margin-inline: calc(3 / 2 * var(--padding));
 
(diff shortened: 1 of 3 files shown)


peterpeterparker:
Looks really good expect to one issue in my opinion. Can we have the dropdown on the side of the modal not within? You might want to double check with the designer, I guess it's also an opiniated feedback.

370691115-0547c4b7-a26a-44ac-abfa-6afad3ac061e

AntonioVentilii-DFINITY:
@peterpeterparker good feedback, already mentioned to Artem and he agreed too: changed (and updated the screenshots)

src/frontend/src/lib/styles/global/gix.scss
```diff
@@ -126,14 +121,32 @@ div.modal {
...
 
 	div.wrapper.dialog {
-		color: var(--color-white);
+		color: var(--color-secondary);
+		padding-block: var(--padding-3x);
+		padding-inline: var(--padding-0_5x);
+	}
+
+	div.wrapper.dialog div.header {
+		--padding-3x: 0;
```
peterpeterparker, line 130: That sounds "dangerous". Any other way to solve the requirement without resetting a core variable?

AntonioVentilii-DFINITY: the issue is that in the `Modal` component ([see source code](https://github.com/dfinity/gix-components/blob/1c4ab390f9cab1d0e3ec73a23384e045679eb6b8/src/lib/components/Modal.svelte#L195)), they are hard-coded like that...

either I change it in gix-component and await a release, or I do it like this now. I understand that it is something that we should deploy quickly, so what about this?

- we do it like this PR
- I add a TODO: adjust better when Gix component is updated
- I change GIX component
- we await release and fix this

peterpeterparker: Setting the `padding` does not work?

```
div.wrapper.dialog div.header {
   padding: something
}
```

AntonioVentilii-DFINITY: nope, i tried it and it doesn't work... the CSS inside the component will override the one in the main definition file

peterpeterparker: Alright, let's add a TODO

AntonioVentilii-DFINITY: cool, done

peterpeterparker (approved):
Looks great! Thanks

PR #2503, opened 2024-09-25
fix(frontend): pass class stretch to content of ContentWithToolbar

AntonioVentilii-DFINITY:
# Motivation

On mobile (when the screen is longer) it is noticeable that the buttons are not at the bottom of the modal, as they should.

That is because the `stretch` class is not correctly set in component `ContentWithToolbar`: it was passed to `<slot>` that does not accept it.

### Before

Screenshot 2024-09-25 at 13 33 06

### After

Screenshot 2024-09-25 at 13 32 57


peterpeterparker (approved):
LGTM, thx

PR #2506, opened 2024-09-25
chore(devops): Update signer release version

lmuntaner:
# Motivation

We introduced breaking changes and new endpoints in CFS.

In this PR, I updated the release version of the signer canister which in turn updates the declaration files.

I also fixed the breaking changes with minimal edits. Proper error management will be added in subsequent PRs.

# Changes

## Manual changes

* Change signer release to `0.2.1`.
* Fix breaking changes in the declarations in `canisters/signer.canister.ts`.

## Automatic changes

All changes within `src/declarations/signer/*` were created by executing `npm run generate` after running `./scripts/build.signer.sh` with the new release.

# Tests

I deployed the signer to the local replica with `./scrtips/deploy.signer.sh` using `--mode reinstall` because I already had one signer installed and I had to force the reinstallation.

After reinstallation I was able to run Oisy locally.

### DIFF
--- a/scripts/build.signer.sh
+++ b/scripts/build.signer.sh
@@ -20,7 +20,7 @@ print_help() {
 
 DFX_NETWORK="${DFX_NETWORK:-local}"
 
-SIGNER_RELEASE="v0.1.2"
+SIGNER_RELEASE="v0.2.1"
 SIGNER_RELEASE_URL="https://github.com/dfinity/chain-fusion-signer/releases/download/${SIGNER_RELEASE}"
 CANDID_URL="${SIGNER_RELEASE_URL}/signer.did"
 WASM_URL="${SIGNER_RELEASE_URL}/signer.wasm.gz"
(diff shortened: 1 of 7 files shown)


lmuntaner:

@DenysKarmazynDFINITY this PR includes the changes after the breaking changes to the signer canister. You'd need to redeploy the signer canister to have the latest version.

AntonioVentilii-DFINITY (approved):
LGTM!

Do you think we should already release a version with this, since it contains adaptation to breaking changes?

lmuntaner:
> LGTM!
> 
> Do you think we should already release a version with this, since it contains adaptation to breaking changes?

The breaking changes of the CFS have not been deployed yet to mainnet. They are only used when btc flat is enable, so it doesn't matter.

This doesn't stop us nor puts pressure on a next release. But we shouldn't wait too long before we deploy the new CFS to prod and also Oisy. I'll coordinate with Max.

PR #2507, opened 2024-09-25
feat(frontend): implement service for communicating with Blockchain API

DenysKarmazynDFINITY:
# Motivation

The goal is to implement service for communicating with Blockchain API. Initially, it will be used for fetching uncertified BTC transactions.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,7 @@ VITE_OISY_URL=https://oisy.com
 VITE_ERC20_ICP_SYMBOL=ckICP
 VITE_ERC20_ICP_REPO_URL=https://github.com/dfinity/erc20-icp
 VITE_COINGECKO_API_URL=https://api.coingecko.com/api/v3
+VITE_BLOCKCHAIN_API_URL=https://blockchain.info
 VITE_EXCHANGE_DISABLED=true
 VITE_COINGECKO_API_KEY=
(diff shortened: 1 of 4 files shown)


.env.example
```diff
@@ -8,6 +8,7 @@ VITE_OISY_URL=https://oisy.com
 VITE_ERC20_ICP_SYMBOL=ckICP
 VITE_ERC20_ICP_REPO_URL=https://github.com/dfinity/erc20-icp
 VITE_COINGECKO_API_URL=https://api.coingecko.com/api/v3
+VITE_BLOCKCHAIN_API_URL=https://blockchain.info
```
peterpeterparker, line 11: That's correct? Their api endpoint is the same URL as their website?

peterpeterparker (approved):
LGMT, thanks

Code looks good, so already approving but, checkout the comment about the env file please before merge.

.env.example
DenysKarmazynDFINITY: Surprisingly, yes, it's the same URL.

peterpeterparker: Weird there they say it's `api.blockchain.com` 🤷‍♂️
https://api.blockchain.com/v3/#authentication

Maybe I'm looking at the wrong documentation. Thanks for the feedback.

PR #2508, opened 2024-09-25
feat(frontend): add exchange rate for BTC testnet

DenysKarmazynDFINITY:
# Motivation

Add exchange rate for BTC testnet. For now, we will assume that 1 BTC mainnet = 1 BTC testnet.


AntonioVentilii-DFINITY (approved):
LGTM! Tks! (discussed outside)

PR #2509, opened 2024-09-25
feat(frontend): Services and stores for regtest address

lmuntaner:
# Motivation

With the new P2WPKH address type, the address in regtest and testnet are not the same.

Therefore, I added a new store and services to load the new address.

# Changes

* New address store `btcAddressRegtestStore`.
* Add regtest in `bitcoinMapper` in the address services mapper.
* New service `loadBtcAddressRegtest` to load regtest the address and store it.
* New derived store `btcAddressRegtest` to get the regtest address.

# Tests

Tested locally using the store in components and load it.


lmuntaner:

src/frontend/src/lib/stores/address.store.ts
```diff
@@ -22,7 +22,7 @@ const initAddressStore = <T extends Address>(): AddressStore<T> => {
 	};
 };
 
-// We use the Testnet address for Regtest.
+export const btcAddressRegtestStore = initAddressStore<BtcAddress>();
```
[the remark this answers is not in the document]
lmuntaner: With P2WPKH there are different addresses per mainnet, testnet and regtest.

https://bitcoin.stackexchange.com/a/106285

Where did you get the concept that it was the same for testnet and regtest?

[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Ah! I was using `P2PKH`. Your are right, for `P2WPKH` are different, sorry, I didn't know

src/frontend/src/lib/services/address.services.ts
```diff
@@ -89,6 +98,11 @@ const bitcoinMapper: Record<
 	testnet: {
 		addressStore: btcAddressTestnetStore,
 		setIdbAddress: setIdbBtcAddressTestnet
+	},
+	regtest: {
+		addressStore: btcAddressRegtestStore,
+		// No need to store the regtest in the local storage because it's only used locally.
+		setIdbAddress: undefined
```
AntonioVentilii-DFINITY, line 105: Considering how often you are going to test this locally, maybe it will be worth to save it at some point, to spare calls? Or not necessary for the effort?

src/frontend/src/lib/stores/address.store.ts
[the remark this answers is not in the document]
lmuntaner: I also didn't know they were different until I tried 😂

AntonioVentilii-DFINITY (approved):
LGTM, tks!

src/frontend/src/lib/services/address.services.ts
lmuntaner: I wanted to skip storing things in local state.

I also assume we will be storing it in the user profile at some point as discussed in the daily. The extra code is code that we won't have to remove later.

If you have a strong opinion I can add it. i thought it simplified the PR. If I can avoid code I'll avoid it 😄

AntonioVentilii-DFINITY: Nono, no strong opinion, just wanted to suggest eheh

peterpeterparker, line 105: If it's a choice, `setIdbAddress` shouldn't be set to `null` instead of `undefined`?

src/frontend/src/lib/services/address.services.ts
```diff
@@ -55,7 +60,9 @@ const loadTokenAddress = async <T extends Address>({
 		const address = await getAddress(identity);
 		addressStore.set({ data: address, certified: true });
 
-		await saveTokenAddressForFutureSignIn({ address, identity, setIdbAddress });
+		if (nonNullish(setIdbAddress)) {
```
peterpeterparker, line 63: Should we have a test for this? If a bug is introduced this can potentially have quite an impact on the user experience.

lmuntaner: I agree with tests, but since there are no tests for these services, I didn't think it necessary.

We should start by having tests.

peterpeterparker: We have E2E tests for the login now no @DenysKarmazynDFINITY ? Do we also test re-entering the app with a valid identity ? i.e. are we testing that when the use returns the modal to load the addresses is not displayed?

DenysKarmazynDFINITY: We have tests for the login and the logout actions but not for re-entering. Would be cool to have it though (we have it listed in backlog).

peterpeterparker: Given this change, if you don't mind, let's then priorize this particular test in the backlog

DenysKarmazynDFINITY: Agreed 👍

PR #2510, opened 2024-09-25
feat(frontend): change Buy modal title based on Onramper env [GIX-3024]

AntonioVentilii-DFINITY:
# Motivation

We would prefer to have a clear indication that Onramper is in DEV mode when an user clicks on the Buy button: we adapt the modal title based on the ENV.


peterpeterparker (approved):
LGTM, thx

PR #2512, opened 2024-09-25
style(frontend): set theme color from blue to white

AntonioVentilii-DFINITY:
# Motivation

The theme color was giving a bad scenario outside the screen. So we change it to white (browser deafult)

Thanks @peterpeterparker for spotting it!

Oisy Wallet


src/frontend/static/manifest.webmanifest
```diff
@@ -3,7 +3,7 @@
 	"short_name": "{{OISY_NAME}}",
 	"start_url": ".",
 	"display": "standalone",
-	"theme_color": "#016DFC",
+	"theme_color": "#FFFFFF",
 	"background_color": "#016DFC",
```
peterpeterparker, line 7: Shouldn't the background color be white as well?

AntonioVentilii-DFINITY: the background color is the one visible for the "splash screen" for example, or during bootstraping, or for filling the screen behind the app icon and the loading animation. I think this one is safe to leave it as it is no? We can always change it if we don't like the results?

peterpeterparker: Isn't the logo displayed over the splashscreen on Android? "Blue on blue"

AntonioVentilii-DFINITY: It is a stronger blue on a lighter blue, but I agree, I will revisit it with Artem

peterpeterparker: Discussed offline, we go for white for now at least and we do some testing. Thanks.

peterpeterparker (approved):
LGTM, thx

PR #2514, opened 2024-09-25
style(frontend): change secondary button to have lower contrast [GIX-3021]

AntonioVentilii-DFINITY:
# Motivation

We want to change the secondary buttons to have the style of what we were calling tertiary. However, as discussed with the Design team, we keep the "old" secondary button (now called secondary-alt) to be re-used too.

# Changes

- Rename button class `secondary` to `secondary-alt`
- Rename button class `tertiary` to `secondary`
- Adjust all instances of `secondary-alt` that should use the new `secondary`

# Results

### New class

Screenshot 2024-09-25 at 18 57 22
Screenshot 2024-09-25 at 18 57 28
Screenshot 2024-09-25 at 18 57 35
Screenshot 2024-09-25 at 18 57 42

### Old class

Screenshot 2024-09-25 at 18 57 12

### DIFF
--- a/src/frontend/src/lib/styles/global/button.scss
+++ b/src/frontend/src/lib/styles/global/button.scss
@@ -25,20 +25,20 @@ button {
 
 	&.primary,
 	&.secondary,
-	&.tertiary {
+	&.secondary-alt {
 		justify-content: flex-start;
 
 		padding: var(--button-padding);
 	}
 
-	&.secondary,
+	&.secondary-alt,
 	&.user {
 		border: 1px solid var(--color-blue);
 	}
 
 	&.primary,
 	&.secondary,
-	&.tertiary {
+	&.secondary-alt {
 		font-weight: var(--font-weight-bold);
 
 		&[disabled],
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/styles/global/button.scss
```diff
@@ -59,15 +59,15 @@ button {
...
-		background: var(--color-cetacean-blue);
-		color: var(--color-white);
-	}
-
-	&.tertiary {
 		background: var(--color-alice-blue);
 		color: var(--color-primary);
 	}
 
+	&.secondary-alt {
```
peterpeterparker, line 66: A comment "TODO: to be removed"?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/styles/global/button.scss
AntonioVentilii-DFINITY: i will put it as "TODO: check with design team if it is stiil in use", because for what I undersstood we will have some usecase

PR #2515, opened 2024-09-25
refactor(frontend): extract common component for primary buttons

AntonioVentilii-DFINITY:
# Motivation

We want to organize better the button classes, to have better control over global changes. So, as first step, we extract a common button component that represents buttons (primary and secondary).

# A few notes

1. There are more possible instances of its usage that we have not yet replaced: it will be done in another PR, since it can become quite big.

2. The type submit is the default one for the <button> element according to documentation (I tested the behavior too...)
Screenshot 2024-09-25 at 19 25 56


src/frontend/src/eth/components/send/SendForm.svelte
```diff
@@ -61,9 +62,9 @@
 
 		<ButtonGroup slot="toolbar">
 			<slot name="cancel" />
-			<button class="primary block flex-1" type="submit" disabled={invalid}>
+			<ButtonPrimary disabled={invalid}>
```
peterpeterparker, line 65: What about calling it `Button` and having it's ui behavior, primary or secondary, as property?

src/frontend/src/lib/components/ui/ButtonPrimary.svelte
```diff
@@ -0,0 +1,17 @@
+<script lang="ts">
+	import { createEventDispatcher } from 'svelte';
+
+	export let disabled = false;
+	export let onClick: (() => void) | undefined = undefined;
+
+	const dispatch = createEventDispatcher();
+
+	let handleOnClick = () => {
```
peterpeterparker, line 9: Let's rather just bubble `on:click`

src/frontend/src/lib/components/ui/ButtonPrimary.svelte
```diff
@@ -0,0 +1,17 @@
...
+		onClick?.();
+		dispatch('click');
+	};
+</script>
+
+<button class="primary block flex-1" {disabled} on:click={handleOnClick}>
```
peterpeterparker, line 15: The type is missing, some buttons where `button` and some `submit`

src/frontend/src/eth/components/send/SendForm.svelte
AntonioVentilii-DFINITY: Yes!!!! I was planning of having ButtonSecondary, but that is way better

src/frontend/src/lib/components/ui/ButtonPrimary.svelte
AntonioVentilii-DFINITY: cool! done

src/frontend/src/lib/components/ui/ButtonPrimary.svelte
AntonioVentilii-DFINITY: theoretically, that is already the default. You mean to explicitly show that we are going with the default?

peterpeterparker: You are right and that was badly explained. Ultimately we will need both so let's already add the type as a property

```
export let type: "submit" | "button" = "submit";
```

AntonioVentilii-DFINITY:

src/frontend/src/lib/components/ui/Button.svelte
```diff
@@ -0,0 +1,37 @@
...
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
+	type $$Props = HierarchyOptions & {
+		type?: 'submit' | 'reset' | 'button';
+		disabled?: boolean;
+	};
+
+	export let type: 'submit' | 'reset' | 'button' = 'submit';
+	export let disabled = false;
+
+	// We guarantee that, if either primary or secondary is set, they mutually exclude each other
```
peterpeterparker, line 15: Not familliar with this pattern. What happens if not, does it fails at build time or runtime?

Build time sounds good. If runtime, how's the error communicated to the user?

AntonioVentilii-DFINITY: For now it is just type-notation-safe, it doesn't fail at runtime or buildtime. At run time, we fallback on defining both props if none is provided, or on primary if both provided (and the user ignored the type-warning).

So the cases are, user passes in a non-type-safe way:

1. `primary` provided --> `primary` is used and `secondary` is the opposite
2. `secondary` provided --> `secondary` is used and `primary` is the opposite
3. both provided --> `primary` takes precedence and we follow rules 1
4. none provided --> `primary` and `secondary` are false, so it's a normal button

peterpeterparker: It's a (really) nice pattern but, if it's "just" for type safety kind of feel like it's an over-complication.

```
export let color: "primary" | "secondary" = "primary";
```

(or another) would do the job as well no?

AntonioVentilii-DFINITY: ah yes, agreed! I did like that because I thought we wanted a very generic `Button` component, but if it's only `primary` or `secondary`, then yes, for sure! Shall I do your suggestion then?

peterpeterparker: Yes I would say so.

PS: the pattern is really cool! Just thinking we can keep it simple at least for now.

AntonioVentilii-DFINITY: yes yes! agreed! changed!

and BTW, the pattern is in the history now, so all good if we need it at some point eheh

peterpeterparker: Nice!

src/frontend/src/lib/components/ui/Button.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+	export let color: 'primary' | 'secondary' = 'primary';
+	export let type: 'submit' | 'reset' | 'button' = 'submit';
+	export let disabled = false;
+</script>
+
+<button class={`${color} block flex-1`} {type} {disabled} on:click>
```
peterpeterparker, line 7: If you have a better idea than `color`, go for it. Not good at naming today.

peterpeterparker (approved):
LGTM, thanks

src/frontend/src/lib/components/ui/Button.svelte
AntonioVentilii-DFINITY: ahahah honestly not really, i search around and it is mostly called either `type` (but we already use it) or `hierarchy` (and I don't like it)

for now, i change it to `style`, if we really don't like it, we change it after

PR #2517, opened 2024-09-25
feat(frontend): Load and render regtest address

lmuntaner:
# Motivation

Regtest and Testnet addresses are not the same for P2WPKH address type.

We already introduced the stores and services for the regtest address. In this PR, I use them.

# Changes

* Load the regtest address if local and testnets in `Loader`.
* User regtest derived store in ReceiveAddresses instead of the testnet.

# Tests

Tested locally.

### DIFF
--- a/src/frontend/src/lib/components/core/Loader.svelte
+++ b/src/frontend/src/lib/components/core/Loader.svelte
@@ -15,12 +15,14 @@
 	import { ProgressStepsLoader } from '$lib/enums/progress-steps';
 	import {
 		loadAddresses,
+		loadBtcAddressRegtest,
 		loadBtcAddressTestnet,
 		loadIdbAddresses
 	} from '$lib/services/address.services';
 	import { signOut } from '$lib/services/auth.services';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { loading } from '$lib/stores/loader.store';
+	import { LOCAL } from '$lib/constants/app.constants';
 
 	let progressStep: string = ProgressStepsLoader.ADDRESSES;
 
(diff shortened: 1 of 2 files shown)


lmuntaner:
Different addresses:

Screenshot 2024-09-25 at 20 03 33

lmuntaner:

This is a follow-up PR from #2509

src/frontend/src/lib/components/core/Loader.svelte
```diff
@@ -69,10 +71,14 @@
 	let progressModal = false;
 
 	const debounceLoadBtcAddressTestnet = debounce(loadBtcAddressTestnet);
+	const debounceLoadBtcAddressRegtest = debounce(loadBtcAddressRegtest);
 
 	$: {
 		if ($testnets && isNullish($btcAddressTestnet)) {
 			debounceLoadBtcAddressTestnet();
+			if (LOCAL) {
```
peterpeterparker, line 79: Shoul we have a single debounced function and within that function add the logic of which addresses to load?

Or it's better two debounce multiple functions for this case?

lmuntaner: No strong opinion, I didn't think this would matter. Because it was called loadAddressTestnet I thought of using two. If it would have been called loadAddressTest I would have put it inside.

peterpeterparker: The questions is probably about what happens if one or the other does not load. Like is it fine if one fail and the other not?

If the former, then it is indeed opiniated but, if the latter, we need to express it.

In both case it's probably more readable to have `await Promise.all` or `allSettled`?

Kind of a half opiniated / half meaningful feedback 😄

lmuntaner: It's fine if one fails and not the other.

We were not doing anything here with the outcome before, nor we do anything now.

I can change the logic though. How do you want me to change it?

peterpeterparker: > It's fine if one fails and not the other.

Great. Then it's maybe a bit an opiniated answer so I gonna approve the PR and let you decide if you want to adapt or not but, I would maybe do following which kind of reflect the other function of the component called `loadData`:

```
const loadBtcAddress = async () = {
   await Promise.allSettled([
			loadBtcAddressTestnet(),
			...(LOCAL ? [loadBtcAddressRegtest()] : [])
		]);
}
const loadBtcAddress = debounce(loadBtcAddress);

$: {
		if ($testnets && isNullish($btcAddressTestnet)) {
		     loadBtcAddress();
		}
	}
```

peterpeterparker (approved):
LGTM, thx

PR #2519, opened 2024-09-26
build(frontend): support for different prerendered routes CSS style

peterpeterparker:
# Motivation

The designer wants to use different background colors for specific routes, such as using white in the wallet and light blue on the signer (`/sign`) route.

We want to prerender the background color because, if we don’t, the user will experience a "glitchy" loading effect where the dapp initially loads with a white background before applying the correct color.

This PR introduces a build setup to pre-render custom styles per route.

The idea is to provide a custom stylesheet next to the Svelte page, which will be automatically prerendered.

For example, if I wanted to add a route `/hello` with a red background, I would add the following files in `src`:

```
src/routes/(group)/hello/+page.svelte
src/routes/(group)/hello/+oisy.page.css
```

and in the css

```
:root {
  background: red;
}
```

# Notes

This requires a custom solution because SvelteKit or the static adapter does not support this feature. They support setting head properties, but those require JS to be loaded first.

# Changes

- Added a placeholder `<!-- ROUTE_STYLE -->` in `app.html`
- Created a script `build.style.mjs` which parses the style (or cleans up the placeholder if no style is provided).
- Set up a Rollup exception to treat the custom `+oisy.page.css` files as `external` resources. Otherwise, Rollup bundles their CSS, causing the specific route styles to be treated as global styles


AntonioVentilii-DFINITY (approved):
LGTM! Tks!

Shall this go somewhere in some documentation for the developers?

peterpeterparker:
Sure. Documentation provided in #2526.

PR #2521, opened 2024-09-26
refactor(frontend): extract replica host

peterpeterparker:
# Motivation

Given that we have a constant file, it's a bit cleaner to also have the DFX replica host set there rather than at the agent level.

# Changes

- Extract `REPLICA_HOST` constant


AntonioVentilii-DFINITY (approved):
LGTM! tks

PR #2522, opened 2024-09-26
feat(frontend): prerender background

peterpeterparker:
# Motivation

We want to improve the user experience when loading the dapp by pre-rendering the background. This way, users won't encounter a UI glitch where the background appears only after the CSS and JS are loaded.

# Notes

Moving the background into `app.html` increases the file size to 4KB instead of 2KB gzipped. According to tests, this is acceptable for this particular background. However, if in the future a more complex background is required by the foundation, we will need to be cautious about file size.

# Trade-off

The background used to be integrated as an image to avoid loading it on mobile devices. Now that we are inlining the background within the HTML file, users will always have to download it. Unfortunately, it's a trade-off.

# Changes

- Integrate background SVG into `app.html`
- Change the color of the prerendered loading spinner to match the background color
- Remove the `Background` component, `background.svg`, and related i18n key


AntonioVentilii-DFINITY (commented):
Does it make sense to leave it as separate element and just use an "import"? Does it work?

```html
<body data-sveltekit-preload-data="hover">
	<img src="background.svg" alt=.... />
</body>
```

peterpeterparker:
> separate element 

I intentionally did not do so. I'm not 100% sure, but if we do, the browser will load and render the HTML page first, then fetch the image separately. By inlining the SVG, we ensure the browser renders it simultaneously with the elements of the HTML page. I know it's few ko but given that the goal is to display the background as soon as possible, this approach feels like the most render blocking solution. WDYT?

AntonioVentilii-DFINITY (approved):
Thank you for the explanation, it makes total sense!

LGTM!

PR #2525, opened 2024-09-26
feat(frontend): icon astronaut empty component

peterpeterparker:
# Motivation

We might need to reuse the astronaut icon elsewhere in the app as in the signer flow. That's why this PR extract an icon.

# Note

The icon is prefixed with `IconAstronaut...` as recently discussed. Not sure about suffix `Empty` but I did not had better idea. Open to suggestion!

> Update: We ultimately used `IconAstronautHelmet` (way better than empty)

# Changes

- Extract Oisy logo "empty" in a dedicated component


AntonioVentilii-DFINITY (approved):
LGTM! Tks!

For the name: maybe `IconAstronautHelmet` (that is an helmet right? eheh)

AntonioVentilii-DFINITY (approved):

PR #2527, opened 2024-09-26
refactor(frontend): extract secondary buttons like Back and Cancel

AntonioVentilii-DFINITY:
# Motivation

Another step to "concentrate" all button style in less places: we create two new components ButtonBack and ButtonCancel, that are secondary buttons re-used several times in the code.

Furthermore, to avoid any possible conflict with built-in props of `<button>` we rename property `style` to `colorStyle`.

# Changes

- Change property `style` to `colorStyle`.
- Create new button components.
- Adapt usage in the code.
- Use secondary buttons in other needed instances.

# Tests

No changes in local replica.


src/frontend/src/eth/components/send/EthSendTokenWizard.svelte
```diff
@@ -289,13 +291,9 @@
 		>
 			<svelte:fragment slot="cancel">
 				{#if formCancelAction === 'back'}
-					<button type="button" class="secondary block flex-1" on:click={back}
-						>{$i18n.core.text.back}</button
-					>
+					<ButtonBack type="button" on:click={back} />
```
peterpeterparker, line 294: Is there a use case where the back or cancel button `submit` a form? We can maybe set those per default to `button` ?

AntonioVentilii-DFINITY: there is no case, BUT, the default is `submit` when it is not given, for the built-in` <button>` component. So we can either keep the same behaviour of the built-in, or I can check the code for the usage (but I am very afraid I can lose track of something)

peterpeterparker: Not sure we are speaking about the same idea. I meant for example modifying `ButtonCancel` as following:

```
<script lang="ts">
	import Button from '$lib/components/ui/Button.svelte';
	
	// REMOVE THIS
	export let type: 'submit' | 'reset' | 'button' = 'submit';
	
	export let disabled = false;
</script>

// MAKE type = button here
<Button colorStyle="secondary" type="button" {disabled} on:click>
	{$i18n.core.text.cancel}
</Button>
```

AntonioVentilii-DFINITY: Yes yes, indeed, but my point is that the instances where it was used without `type` provided, for example:

```html
<button class="secondary block flex-1" on:click={() => dispatch('icCancel')}>
  {$i18n.core.text.cancel}
</button>
```

 means that they were automatically `submit` type (that is the default if nothing is provided).
 
So to maintain the same behavior and avoid any disruptions, we need to omit it too, let it default to `submit`.
 
Otherwise, I should check all instances mentioned to check if one of them MUST explicitly have `submit` type. I can do that too, no worries

peterpeterparker: Good point. Let's pass this PR and then modify the buttons in another PR? I feel like "cancel" and "back" should not be "submit" maybe?

AntonioVentilii-DFINITY: right! will do!

AntonioVentilii-DFINITY: Done with:

- PR #2544

PR #2530, opened 2024-09-26
chore(frontend): Remove the "^" from the dependencies we need a next version

lmuntaner:
# Motivation

I realized that the BitcoinNetwork from ckbtc didn't have the latest changes. Specifically, it didn't include "regtest" in the Bitcoin networks.

I didn't find the npm documentation, but this is what ChatGPT says:

> The reason it ignores “next” or pre-releases (like 2.0.0-beta or 1.3.0-next) is that the caret range is designed to only allow non-breaking changes according to semantic versioning rules.
>
> Since pre-releases or “next” releases may not be stable or fully backward-compatible, they are typically excluded from the versions matched by the caret (^) unless they are explicitly specified.

# Changes

* Remove "^" from package.json in all the dependencies using a "next" version.
* Change in package.lock.json after running `npm i` after change package.json.

# Tests

Everything still works and the new bitcoin network "regtest" is there.


lmuntaner:

peterpeterparker (commented):
Not sure that's worth removing. Next time we upgrade the libs with `npm rm && npm i` those will be back.

lmuntaner:
> Not sure that's worth removing. Next time we upgrade the libs with `npm rm && npm i` those will be back.

How do you upgrade with `npm rm && npm i`?

How do you propose we use the next versions then?

peterpeterparker:
I'm not sure I understand the problem. Wouldn't "just" updating `npm rm @dfinity/ckbtc && npm i @dfinity/ckbtc@next` resolve your issue?

lmuntaner:
> I'm not sure I understand the problem. Wouldn't "just" updating `npm rm @dfinity/ckbtc && npm i @dfinity/ckbtc@next` resolve your issue?

I discovered that the problem was not actually the caret but package.lock which was pointing to 3.0.0 without the next version.

I updated that in #2534 

Closing this issue.

Thanks for the comments @peterpeterparker

PR #2532, opened 2024-09-26
style(frontend): fix excessive margin below the Card description

AntonioVentilii-DFINITY:
# Motivation

The spacing of label and description in the Card component was off: that is because the `<p>` element is rendered either way with a certain margin. So we transform it into a `<span>`

### Before

Screenshot 2024-09-26 at 12 08 26
Screenshot 2024-09-26 at 12 08 22
Screenshot 2024-09-26 at 12 08 40
Screenshot 2024-09-26 at 12 08 47

### After

Screenshot 2024-09-26 at 12 06 43
Screenshot 2024-09-26 at 12 06 47
Screenshot 2024-09-26 at 12 07 02
Screenshot 2024-09-26 at 12 07 12


peterpeterparker (approved):
LGTM, thx

Assuming the slot description is never passed as `p`.

PR #2533, opened 2024-09-26
feat(frontend): Add regtest to IcSendBtcNetwork

lmuntaner:
# Motivation

The component displaying the network name wasn't taking into account regtest.

# Changes

* Add a condition to check testnet instead of falling back to testnet if not mainnet.

# Tests

Works locally in #2528


lmuntaner:

AntonioVentilii-DFINITY (approved):
LGTM, tks!

I think for ETH we have some util similar to this, like `isEthereumTestnet`, or something. Maybe you can create a similar one at some point? I imagine you will use this comparison quite a lot

lmuntaner:
> LGTM, tks!
> 
> I think for ETH we have some util similar to this, like `isEthereumTestnet`, or something. Maybe you can create a similar one at some point? I imagine you will use this comparison quite a lot

I'll create it when I need it then, thanks!

PR #2535, opened 2024-09-26
fix(frontend): Pass targetNetwork to SendModal

lmuntaner:
# Motivation

The `networkId` was undefined within the Send modal even though there was a network and token selected.

I realized that the `networkId` comes from `targetNetwork` which was never passed, therefore, it was never set.

I wonder whether there is a rule to check that a prop is never passed when using the component. According to ChatGPT there is no such rule.

# Changes

* Pass `targetNetwork` to `SendModal` in the `Send` component using the global store `token`.

# Tests

This was tested in #2528

### DIFF
--- a/src/frontend/src/lib/components/send/Send.svelte
+++ b/src/frontend/src/lib/components/send/Send.svelte
@@ -3,10 +3,11 @@
 	import SendModal from '$lib/components/send/SendModal.svelte';
 	import { modalSend } from '$lib/derived/modal.derived';
 	import { modalStore } from '$lib/stores/modal.store';
+	import { token } from '$lib/stores/token.store';
 
 	export let isTransactionsPage: boolean;
 </script>
 
 <SendButtonWithModal open={modalStore.openSend} isOpen={$modalSend}>
-	<SendModal {isTransactionsPage} on:nnsClose slot="modal" />
+	<SendModal {isTransactionsPage} on:nnsClose slot="modal" targetNetwork={$token?.network} />
 </SendButtonWithModal>


lmuntaner:

src/frontend/src/lib/components/send/Send.svelte
```diff
@@ -3,10 +3,11 @@
...
 	import { modalSend } from '$lib/derived/modal.derived';
 	import { modalStore } from '$lib/stores/modal.store';
+	import { token } from '$lib/stores/token.store';
 
 	export let isTransactionsPage: boolean;
 </script>
 
 <SendButtonWithModal open={modalStore.openSend} isOpen={$modalSend}>
-	<SendModal {isTransactionsPage} on:nnsClose slot="modal" />
+	<SendModal {isTransactionsPage} on:nnsClose slot="modal" targetNetwork={$token?.network} />
```
peterpeterparker, line 12: Did you faced an issue? In which use case?

Spontaneously, I would think that we might omit `targetNetwork` here on purpose but, of course I might be wrong therefore asking about the issue.

lmuntaner: The difference is to have two networks or one.

Without:
Screenshot 2024-09-26 at 14 46 41

With:
Screenshot 2024-09-26 at 14 48 17

Rethinking about it, I wonder whether just one makes more sense.

Should I remove the prop which is not used anyway? It seems useful, but maybe more for conversions and maybe they use another modal now?

peterpeterparker: I get the feeling that the destination network is only displayed or provided if it's different, like ckETH -> ETH.

That's why the first screenshot says "Network" while the second says "Source / Destination Network".

Given that we have no tests, we should be extra cautious if we modify anything. However, if you've encountered an error (not cosmetic), we definitely need to investigate.

lmuntaner: No error. I'll close this PR.

lmuntaner:
We prefer the UX without the prop.

PR #2543, opened 2024-09-26
style(frontend): ManageTokensButton is shown only when signed in [GIX-3015]

AntonioVentilii-DFINITY:
# Motivation

The ManageTokensButton will be shown only when signed in, since it is not much useful visually when signed out.

### Before

Screenshot 2024-09-26 at 18 26 18

### After

Screenshot 2024-09-26 at 18 24 29
Screenshot 2024-09-26 at 18 24 46


src/frontend/src/lib/components/tokens/Tokens.svelte
```diff
@@ -17,9 +17,9 @@
 
 	{#if $authSignedIn}
 		<TokensSignedIn />
+
+		<ManageTokensButton />
```
peterpeterparker, line 21: Does it needs a `fade` effect?

AntonioVentilii-DFINITY: Good one! Added `fade` transition (same as token list)


https://github.com/user-attachments/assets/cd8d3fe4-3ab5-4dc1-9642-7e48f1812ba4

src/frontend/src/lib/components/tokens/Tokens.svelte
```diff
@@ -17,9 +18,11 @@
 
 	{#if $authSignedIn}
 		<TokensSignedIn />
+
+		<div transition:fade>
```
peterpeterparker, line 22: nitpick: `in:` is probably enough

peterpeterparker (approved):
LGTM, thx

PR #2544, opened 2024-09-26
refactor(frontend): default type of Cancel and Back buttons is not submit

AntonioVentilii-DFINITY:
# Motivation

As mentioned by @peterpeterparker , it doesn't make sense that the `ButtonCancel` and the `ButtonBack` components have `submit` type as default. Moreover, the few instances where they were used in a form, they were set as type `button`.

So, we default both of them as type `button`.


peterpeterparker (approved):
Cool, thx

PR #2550, opened 2024-09-26
feat(frontend): use common canister/api types for signer

DenysKarmazynDFINITY:
# Motivation

Follow-up PR for using common canister/api types for Signer.


peterpeterparker (approved):
LGTM, thx

Just sharing an idea if interesting: maybe `CommonCanisterApiFunctionParams` can be shortened to `CanisterApiFunctionParams`?

To some extent "Canister" is "Common" so it can be interpreted as "CommonCommon" 😉

DenysKarmazynDFINITY:
Makes sense, that type is renamed 👍

PR #2556, opened 2024-09-27
style(frontend): new header title when signed off [GIX-3015]

AntonioVentilii-DFINITY:
# Motivation

New header title when the user is not signed in.

Screenshot 2024-09-27 at 10 45 31
Screenshot 2024-09-27 at 10 45 49
Screenshot 2024-09-27 at 10 45 56

### DIFF
--- a/src/frontend/src/lib/i18n/en.json
+++ b/src/frontend/src/lib/i18n/en.json
@@ -43,7 +43,7 @@
 	"auth": {
 		"text": {
 			"title_part_1": "Multi-chain wallet.",
-			"title_part_2": "100% onchain",
+			"title_part_2": "100% onchain.",
 			"authenticate": "Authenticate",
 			"logout": "Logout"
 		},
--- a/src/frontend/src/lib/i18n/en.json
+++ b/src/frontend/src/lib/i18n/en.json
@@ -42,8 +42,8 @@
 	},
 	"auth": {
 		"text": {
-			"title_part_1": "A wallet that lives on-chain for",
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/i18n/en.json
```diff
@@ -42,8 +42,8 @@
 	},
 	"auth": {
 		"text": {
-			"title_part_1": "A wallet that lives on-chain for",
-			"title_part_2": "100% decentralization",
+			"title_part_1": "Multi-chain wallet.",
+			"title_part_2": "100% onchain",
```
DenysKarmazynDFINITY, line 46: On the screenshot that Andronikos shared in Slack, the second part also has a dot (`100% on chain.`).

DenysKarmazynDFINITY (approved):
The code LGTM, just one comment re. the translation.

src/frontend/src/lib/i18n/en.json
peterpeterparker: The foundation wants "onchain" I understand.

See for example this recent PR applied to the portal: https://github.com/dfinity/portal/pull/3530

DenysKarmazynDFINITY: Oh true, it should be `100% onchain.`. The question about the dot is still pending though.

AntonioVentilii-DFINITY: > On the screenshot that Andronikos shared in Slack, the second part also has a dot (100% on chain.).

Good catch! Corrected

PR #2557, opened 2024-09-27
feat(frontend): add exporerUrl to BTC networks [GIX-2795]

DenysKarmazynDFINITY:
# Motivation

The goal is to extend BTC networks with already defined explorer URLs. It will be used in the BTC transaction detail modal.

### DIFF
--- a/src/frontend/src/btc/types/network.ts
+++ b/src/frontend/src/btc/types/network.ts
@@ -1,0 +1,7 @@
+import type { Network } from '$lib/types/network';
+
+export interface BitcoinAppMetadata {
+	explorerUrl?: string;
+}
+
+export type BitcoinNetwork = Network & BitcoinAppMetadata;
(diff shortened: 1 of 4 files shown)


AntonioVentilii-DFINITY (commented):
Just to simplify the code, we have already `EthereumAppMetadata`: can we "merge" the two types and have something like `NetworkAppMetadata`? (that should include the original optionality/non-optionality of both)

DenysKarmazynDFINITY:
@AntonioVentilii-DFINITY good idea, it's done. `explorerUrl` is required in `NetworkMetadataApp` but I made it optional just for BTC to be compatible with regtest network type (no explorer there).

src/frontend/src/btc/types/network.ts
```diff
@@ -0,0 +1,3 @@
+import type { Network, NetworkAppMetadata } from '$lib/types/network';
+
+export type BitcoinNetwork = Network & Partial<Pick<NetworkAppMetadata, 'explorerUrl'>>;
```
AntonioVentilii-DFINITY, line 3: For now, I think it is safe to remove the Pick part, right? Just an observation, unless you plan already on adding more things there that will be non-optional

AntonioVentilii-DFINITY (approved):
LGTM! Small observation

PR #2558, opened 2024-09-27
fix(frontend): remove unnecessary BTC network type

DenysKarmazynDFINITY:
# Motivation

Since the `BitcoinNetowrk` type from `@dfinity/ckbtc` now includes `regtest`, the type that I created previously is redundant.


AntonioVentilii-DFINITY (approved):
LGTM!

PR #2559, opened 2024-09-27
Upgrade pocket-ic and install bitcoin canister

lmuntaner:
# Motivation

We added some endpoints that talk with bitcoin nodes.

If we want to add integration tests for those, we first need to upgrade pocket-ic.

# Changes

* Change pocket-ic versions in binary and cargo.toml.

# Tests


lmuntaner:
Closing. It got split into smaller PRs, the last one #2575

PR #2560, opened 2024-09-27
feat(frontend): Onramper integration [GIX-2854]

AntonioVentilii-DFINITY:
# Motivation

As discussed offline, we merge this PR providing only the feature of buying tokens, without feedback on pending transactions to the user. That will come in a second moment.

Note: we remove the feature-flag for OnRamper


peterpeterparker (approved):
LGTM, thx

Confirmed offline: This PR is the sum ups of PRs that have been reviewed separatly.

PR #2561, opened 2024-09-27
chore(test): Upgrade pocket-ic

lmuntaner:
# Motivation

If we wanto to add integration tests interacting with bitcoin API we need to upgrade pocket-ic.

This is a first step upgrading pocket-ic to the latest release.

# Changes

* Upgrade Rust and binary versions of pocket-ic according to the [compatibility table](https://docs.google.com/document/d/1VYmHUTjrgbzRHtsAyRrI5cj-gWGs7ktTnutPvUMJioU/edit#heading=h.5wf28dvt742x)
* Create Oisy backend in fiduciary subnet directly.
* Install with II subneet to have access to ecdsa keys.
* Change `ecdsa_key_name` init arg to the test one.

# Tests

I had to move two timestamps because the tests weren't passing. Not sure what changed the time between versions, but it makes sense to move it after the update call.

### DIFF
--- a/src/backend/tests/it/list_users.rs
+++ b/src/backend/tests/it/list_users.rs
@@ -77,13 +77,6 @@ fn test_list_users_returns_filtered_users_by_updated() {
 
     // Advance time before updating one of the users
     pic_setup.pic().advance_time(Duration::new(10, 0));
-    let timestamp_nanos_2 = pic_setup
-        .pic()
-        .get_time()
-        .duration_since(UNIX_EPOCH)
-        .expect("Time went backwards")
-        .as_nanos();
-
     // Update one of the users created before timestamp
     let add_user_cred_arg = AddUserCredentialRequest {
         credential_jwt: VP_JWT.to_string(),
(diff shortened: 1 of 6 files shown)


lmuntaner:
@bitdivine @peterpeterparker please review

src/backend/tests/it/list_users.rs
bitdivine, line 1: It is not clear to me why these timestamps need to be moved.  Can you shed some light on this, please?

bitdivine (approved):

src/backend/tests/it/list_users.rs
lmuntaner: It also surprised me and I'd like to dig deeper, but time is scarce.

Context: This timestamp is used to compare the created with the expected user.

Before, I was getting the timestamp right after advancing the time but before doing the update call. I was expecting the user to have the time right after the advanced time. But now, it seems that making the update call takes longer and the user has a a higher timestamp. Specifically it was always + 1 the one after advancing time. But I also found out that it was always the time of the replica after the update call. Therefore, I chose to move getting the timestamp of the replica after the update call.

PR #2563, opened 2024-09-27
feat(frontend): util for parsing BTC transactions

DenysKarmazynDFINITY:
# Motivation

A parser for BTC transactions. It will later extended with additional field: "confirmations".

### DIFF
--- a/src/frontend/src/btc/types/btc.ts
+++ b/src/frontend/src/btc/types/btc.ts
@@ -1,0 +1,20 @@
+import type { TransactionType } from '$lib/types/transaction';
+
+export type BtcTransactionStatus = 'confirmed' | 'pending';
+
+export interface BtcTransactionUi {
+	hash: string;
+	timestamp: number;
+	value: bigint;
+	type: TransactionType;
+	status: BtcTransactionStatus;
+	from: string;
+	to: string;
+	// The block number available only if a transaction has been confirmed
+	blockNumber?: number;
+
+	/* TODO: add one more field "confirmations", a number that represents the acceptance of a new block by the blockchain network.
+	 1. Use https://blockchain.info/latestblock to get info about the latest block height.
+	 2. Calculate confirmations: confirmations = currentBlockHeight - transactionBlockHeight + 1
+	 */
+}
--- a/src/frontend/src/btc/types/btc.ts
+++ b/src/frontend/src/btc/types/btc.ts
@@ -3,13 +3,13 @@ import type { TransactionType } from '$lib/types/transaction';
 export type BtcTransactionStatus = 'confirmed' | 'pending';
 
 export interface BtcTransactionUi {
-	hash: string;
+	id: string;
 	timestamp: number;
 	value: bigint;
 	type: TransactionType;
 	status: BtcTransactionStatus;
 	from: string;
-	to: string;
+	to?: string;
 	// The block number available only if a transaction has been confirmed
 	blockNumber?: number;
 
--- a/src/frontend/src/btc/utils/btc-transactions.utils.ts
+++ b/src/frontend/src/btc/utils/btc-transactions.utils.ts
@@ -1,59 +1,29 @@
-import type { BtcTransactionStatus, BtcTransactionUi } from '$btc/types/btc';
+import type { BtcTransactionUi } from '$btc/types/btc';
 import type { BtcAddress } from '$lib/types/address';
 import type { BitcoinOutput, BitcoinTransaction } from '$lib/types/blockchain';
-import type { TransactionType } from '$lib/types/transaction';
 import { nonNullish } from '@dfinity/utils';
 
 export const mapBtcTransaction = ({
-	transaction,
+	transaction: { inputs, block_index, out, hash, time },
 	btcAddress
 }: {
 	transaction: BitcoinTransaction;
 	btcAddress: BtcAddress;
 }): BtcTransactionUi => {
-	const type: TransactionType = transaction.inputs.some(
-		(input) => input.prev_out.addr === btcAddress
-	)
-		? 'send'
-		: 'receive';
-	let amount = 0;
-	let senderAddress = '';
-	let receiverAddress = '';
-	let blockNumber;
-	let output: BitcoinOutput | undefined;
-	let status: BtcTransactionStatus = 'pending';
-
-	if (nonNullish(transaction.block_index)) {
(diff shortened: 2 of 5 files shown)


src/frontend/src/btc/types/btc.ts
```diff
@@ -0,0 +1,20 @@
+import type { TransactionType } from '$lib/types/transaction';
+
+export type BtcTransactionStatus = 'confirmed' | 'pending';
+
+export interface BtcTransactionUi {
```
peterpeterparker, line 5: Shouldn't we first implement the effective loading of the transactions before implementing this?

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -0,0 +1,59 @@
+import type { BtcTransactionStatus, BtcTransactionUi } from '$btc/types/btc';
+import type { BtcAddress } from '$lib/types/address';
+import type { BitcoinOutput, BitcoinTransaction } from '$lib/types/blockchain';
+import type { TransactionType } from '$lib/types/transaction';
+import { nonNullish } from '@dfinity/utils';
+
+export const mapBtcTransaction = ({
```
peterpeterparker, line 7: Should we have a test for this or we will have an E2E tests?

Asking because generally now we provide tests when we add utils.

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -0,0 +1,59 @@
...
+	transaction,
+	btcAddress
+}: {
+	transaction: BitcoinTransaction;
+	btcAddress: BtcAddress;
+}): BtcTransactionUi => {
+	const type: TransactionType = transaction.inputs.some(
```
peterpeterparker, line 14: nitpick: Our code style leverages destructing objects. If you could follow it, it would be nice

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -0,0 +1,59 @@
...
+		(input) => input.prev_out.addr === btcAddress
+	)
+		? 'send'
+		: 'receive';
+	let amount = 0;
```
peterpeterparker, line 19: We generally follow a functional approach and are about to add an eslint rule that prevent the usage of `let` in favor of `const`. Can you refactor the code to amend this?

peterpeterparker (commented):
A first round of comments.

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: I will add unit tests 👍

DenysKarmazynDFINITY: Tests have been added.

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: Sure, done.

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: Done.

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -0,0 +1,29 @@
...
+	const isTypeSend = inputs.some(({ prev_out }) => prev_out.addr === btcAddress);
+	const blockIndexAvailable = nonNullish(block_index);
+	const output: BitcoinOutput | undefined = out.find(({ addr }) =>
+		isTypeSend ? addr !== btcAddress : addr === btcAddress
+	);
+
+	return {
+		id: hash,
+		timestamp: time,
+		value: BigInt(output?.value ?? 0),
```
peterpeterparker, line 22: Is it correct to default to zero?

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -0,0 +1,29 @@
...
+		status: blockIndexAvailable ? 'confirmed' : 'pending',
+		blockNumber: block_index ?? undefined,
```
peterpeterparker, line 24: You default to `undefined` because it can be `null` or `blockNumber: block_index` would be the same?

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -0,0 +1,29 @@
...
+		type: isTypeSend ? 'send' : 'receive',
+		from: isTypeSend ? btcAddress : inputs[0].prev_out.addr,
```
peterpeterparker, line 26: `inputs[0]` and `inputs[0].prev_out` are never undefined?

peterpeterparker (commented):
Looks good, few "minor" remaining questions

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: Yes, the Blockchain API returns either `null` or `number`, and for the UI type I want to have `number | undefined`.

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: According to the API, those values should always be there.

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: It's a good question. I think yes, but I can adjust it in case I face something that speaks against it.

peterpeterparker: After double checking the `IcTransaction`, I think should maybe handle the value as potentially `undefined`?

To some extent, I interpret no value displayed differently than zero - i.e. no value displayed might be a bug while zero would lead me to ask myself "WHAT Zero??? Where is my money?" 😉.

DenysKarmazynDFINITY: Ok, agreed 🙂  I have already merged this PR to my feature branch, so I will update it there.

PR #2566, opened 2024-09-27
feat(frontend): load BTC transactions

DenysKarmazynDFINITY:
# Motivation

The goal is to update the BTC wallet worker to fetch transactions. Similar to other store, we only load a BTC address if it hasn't been already fetched by FE.

### DIFF
--- a/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
+++ b/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
@@ -1,15 +1,22 @@
-import { getBtcBalance } from '$lib/api/signer.api';
+import type { BitcoinNetwork } from '$declarations/signer/signer.did';
+import { getBtcAddress, getBtcBalance } from '$lib/api/signer.api';
 import { WALLET_TIMER_INTERVAL_MILLIS } from '$lib/constants/app.constants';
+import { btcAddressData } from '$lib/rest/blockchain.rest';
 import { SchedulerTimer, type Scheduler, type SchedulerJobData } from '$lib/schedulers/scheduler';
+import type { OptionBtcAddress } from '$lib/types/address';
 import type { BitcoinTransaction } from '$lib/types/blockchain';
+import type { OptionIdentity } from '$lib/types/identity';
 import type {
 	PostMessageDataRequestBtc,
 	PostMessageDataResponseWallet
 } from '$lib/types/post-message';
+import type { CertifiedData } from '$lib/types/store';
 import { assertNonNullish, jsonReplacer } from '@dfinity/utils';
 
 export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> {
 	private timer = new SchedulerTimer('syncBtcWalletStatus');
+
+	private btcAddress: OptionBtcAddress;
 
 	stop() {
 		this.timer.stop();
@@ -30,6 +37,19 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> 
 		});
 	}
 
+	private async loadBtcAddress({
+		identity,
+		bitcoinNetwork
+	}: {
+		identity: OptionIdentity;
+		bitcoinNetwork: BitcoinNetwork;
+	}): Promise<string> {
+		// Save address for next timer
+		this.btcAddress = await getBtcAddress({ identity, network: bitcoinNetwork });
+
+		return this.btcAddress;
+	}
+
 	/* TODO: The following steps need to be done:
 	 * 1. Fetch uncertified transactions via BTC transaction API.
 	 * 2. Query uncertified balance in oder to improve UX (signer.getBtcBalance takes ~5s to complete).
@@ -45,8 +65,22 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> 
 			network: bitcoinNetwork
 		});
 
-		// TODO: Fetch and parse transactions
-		const transactions: BitcoinTransaction[] = [];
+		let uncertifiedTransactions: CertifiedData<BitcoinTransaction>[] = [];
+
+		if (data?.shouldFetchTransactions) {
+			const btcAddress =
+				data?.btcAddress ??
+				this.btcAddress ??
+				(await this.loadBtcAddress({ identity, bitcoinNetwork }));
+
+			const transactions = (await btcAddressData({ btcAddress })).txs;
+
+			uncertifiedTransactions = transactions.map((transaction) => ({
+				// TODO: Parse transactions to BtcTransactionUi type
+				data: transaction,
+				certified: false
+			}));
+		}
 
 		this.postMessageWallet({
 			wallet: {
@@ -54,7 +88,7 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> 
 					data: balance,
 					certified: true
 				},
-				newTransactions: JSON.stringify(transactions, jsonReplacer)
+				newTransactions: JSON.stringify(uncertifiedTransactions, jsonReplacer)
 			}
 		});
 	};
--- a/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
+++ b/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
@@ -51,9 +51,10 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> 
 	}
 
 	/* TODO: The following steps need to be done:
-	 * 1. Fetch uncertified transactions via BTC transaction API.
-	 * 2. Query uncertified balance in oder to improve UX (signer.getBtcBalance takes ~5s to complete).
-	 * 3. Fetch certified transactions via BE endpoint (to be discussed).
+	 * 1. [Required] Fetch uncertified transactions via BTC transaction API.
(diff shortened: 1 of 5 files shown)


src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -30,6 +37,19 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
 
+	private async loadBtcAddress({
+		identity,
+		bitcoinNetwork
+	}: {
+		identity: OptionIdentity;
+		bitcoinNetwork: BitcoinNetwork;
+	}): Promise<string> {
+		// Save address for next timer
+		this.btcAddress = await getBtcAddress({ identity, network: bitcoinNetwork });
```
peterpeterparker, line 48: Why is `this.btcAddress` needed?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -45,16 +65,30 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
 			network: bitcoinNetwork
 		});
 
-		// TODO: Fetch and parse transactions
-		const transactions: BitcoinTransaction[] = [];
+		let uncertifiedTransactions: CertifiedData<BitcoinTransaction>[] = [];
+
+		if (data?.shouldFetchTransactions) {
+			const btcAddress =
+				data?.btcAddress ??
```
peterpeterparker, line 72: What happens if the the new requested address is different than the address in the worker?
Is there any comparison of the certification, what happens if one address is certified and the other no? Or are addresses always certified? Do we have an assertion to ensure the address is certified?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -45,16 +65,30 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
+				this.btcAddress ??
+				(await this.loadBtcAddress({ identity, bitcoinNetwork }));
+
+			const transactions = (await btcAddressData({ btcAddress })).txs;
```
peterpeterparker, line 76: - nitpick to align our code style: `const {txs: transactions} = ...`

- `btcAddressData` returns the transactions? isn't the naming of the function a bit misleading?

src/frontend/src/btc/services/worker.btc-wallet.services.ts
```diff
@@ -27,16 +37,25 @@ export const initBtcWalletWorker = async ({
...
 					bitcoinNetwork: mapToSignerBitcoinNetwork({
-						network: isNetworkIdBTCTestnet(networkId)
-							? 'testnet'
-							: isNetworkIdBTCRegtest(networkId)
-								? 'regtest'
-								: 'mainnet'
-					})
+						network: isNetworkTestnet ? 'testnet' : isNetworkRegtest ? 'regtest' : 'mainnet'
+					}),
+					// only mainnet transactions can be fetched via Blockchain API
```
peterpeterparker, line 57: How do we develop locally and test on staging?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: `btcAddress` is needed to fetch BTC address data (`loadBtcAddress`). And `this.btcAddress` stores the value so it can be reused in the next timer runs.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: * Ok
* that function returns not only txs data, but for now and in this particular case, I only need `btcAddressData.txs`. If by the end of the implementation we end up only needing transactions info, I can refactor this method to ignore the rest of the data and rename it accordingly.

src/frontend/src/btc/services/worker.btc-wallet.services.ts
DenysKarmazynDFINITY: Unfortunately, at the moment I don't have a decent way of testing it locally - for now, I'm hardcoding my mainnet BTC address to test the transactions. For the future, we can potentially use mocked data to have some BTC txs displayed locally.

On staging though, we should be able to test it properly for mainnet addresses.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: `this.btcAddress` is always certified. After double-checking, it appeared that `data?.btcAddress` can indeed be uncertified if it's loaded from IDB and not canister directly.  

I updated the code to always fetch certified BTC address on the first worker run. After I'm done with the main functionality, I'll come back to the approach of trying to get the address from FE (while keeping in mind the questions you raised). For that, I added an additional TODO comment.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
peterpeterparker: I see, thanks for the explanation. Should/can we implement a similar pattern as the other workers by using a store?

For example in `ic-wallet.scheduler`.

```
private store: IcWalletStore<T> = {
		balance: undefined,
		transactions: {}
	};
```

Should we also need to store the transactions?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -45,16 +66,28 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
 			network: bitcoinNetwork
 		});
 
-		// TODO: Fetch and parse transactions
-		const transactions: BitcoinTransaction[] = [];
+		let uncertifiedTransactions: CertifiedData<BitcoinTransaction>[] = [];
+
+		if (data?.shouldFetchTransactions) {
```
peterpeterparker, line 71: Maybe we can extract a function for this? Like a private method?

src/frontend/src/btc/services/worker.btc-wallet.services.ts
peterpeterparker: Gotcha. Yeah then makes sense. I'm just wondering now if this is the best place for documenting that there is no transactions locally? You know what I mean? Do we have a way to make this easier to find?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
peterpeterparker: > I updated the code to always fetch certified BTC address on the first worker run. 

This will also be the call that will certify that the address in IDB is certified or that will be an extra call?

Context: we need update call for security to ensure data are valid but, we also need to spare update call because it loads the network. 🤪

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
peterpeterparker: Sounds good

src/frontend/src/btc/services/worker.btc-wallet.services.ts
DenysKarmazynDFINITY: I added a new section to `HACKING`, as it seems to be the best place to put this info. I'll make sure to keep it up-to-date as we proceed with the development.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: Done. Also did the same for loading balance.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: 1. There is no guarantee that FE loads the address before BTC wallet worker is launched. At least locally, I had an impression that it never happens. 
2. FE loads the address from IDB, and only after that certifies it. 

So, in order not to wait for those steps done by FE, I'm loading BTC address in the worker. Later, we can improve it by syncing these 2 processes (whoever loads it first, notifies the other one), similar as it's done in the IC wallet worker,

src/frontend/src/btc/services/worker.btc-wallet.services.ts
```diff
@@ -27,16 +37,24 @@ export const initBtcWalletWorker = async ({
 
 	return {
 		start: () => {
+			const isTestnetNetwork = isNetworkIdBTCTestnet(networkId);
+			const isRegtestNetwork = isNetworkIdBTCRegtest(networkId);
+
 			worker.postMessage({
 				msg: 'startBtcWalletTimer',
 				data: {
+					btcAddress: get(
```
peterpeterparker, line 46: I would not get the address imperatively, that can lead to some surprises. We should probably stop/start the worker each time the address change and pass it as a variable?

If you want to tackle this in another PR right after this one, we can just add a TODO for now that way we move on? WDYT?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
peterpeterparker: Thanks!

peterpeterparker (approved):
LGTM, thanks

PR #2569, opened 2024-09-27
refactor(frontend): replace internal IconMenu with GIX Components one

AntonioVentilii-DFINITY:
# Motivation

The `IconMenu` component is practically identical to the `IconMenu` component of [library](https://gix.design/icons) `@dfinity/gix-components` ([source code](https://github.com/dfinity/gix-components/blob/main/src/lib/icons/IconMenu.svelte)).

So we replace it.

Note: the default size of icons in the library is already `20px`.


peterpeterparker (approved):
LGTM, thx

PR #2570, opened 2024-09-27
refactor(frontend): replace internal IconMore with GIX Components equivalent

AntonioVentilii-DFINITY:
# Motivation

The `IconMore` component is practically identical to the `IconDots` component of [library](https://gix.design/icons) `@dfinity/gix-components` ([source code](https://github.com/dfinity/gix-components/blob/main/src/lib/icons/IconDots.svelte)).

Screenshot 2024-09-27 at 18 42 11

So we replace it.


peterpeterparker (approved):
LGTm, thx

PR #2571, opened 2024-09-29
chore(frontend): use aliases for relative imports

AntonioVentilii-DFINITY:
# Motivation

For consistency, we replace the relative imports with aliases.


peterpeterparker (approved):
LGTM, thanks

PS: should we have an eslint rule? If yes, then probably just one for this repo because in other repo where the eslint lib is used, alias are not used?

PR #2573, opened 2024-09-30
feat(rust): Rust bindings for the cycles ledger

bitdivine:
# Motivation
The Oisy backend canister will soon have to make API calls to the cycles ledger.

# Changes
- Add Rust bindings for calling the cycles ledger.

# Tests
This PR defines type signatures, so the relevant test is `cargo check` passing.

# TODO
The bindings were generated using `didc` and then derive traits such as `Eq` and `PartialEq` were added to particular types that will need them.  This derivation can probably be automated with the shiny new `didc` configuration options.  [Tracking issue.](https://dfinity.atlassian.net/browse/GIX-3032)


Cargo.toml
```diff
@@ -9,6 +10,8 @@ resolver = "2"
 ic-cdk = "0.15.1"
 ic-cdk-macros = "0.15.0"
 ic-cdk-timers = "0.9.0"
+ic-cycles-ledger-client = { path = "src/cycles_ledger_client" }
+ic-ledger-types = "0.13.0"
```
lmuntaner, line 14: Where is this used? Of will it be used in an upcoming PR?

src/cycles_ledger_client/src/lib.rs
lmuntaner, line 1: Is this file created with a script or manually?

Can we add which script was used if so?

lmuntaner (approved):
A couple of comments, but already approving.

Cargo.toml
bitdivine: It will be used in an upcoming PR.

src/cycles_ledger_client/src/lib.rs
bitdivine: Manually.  There is a TODO section in the MR description with details.

lmuntaner: Thanks!

PR #2574, opened 2024-09-30
feat(frontend): Improve error mgt signer canister

lmuntaner:
# Motivation

Have clearer understanding of the errors coming from the signer canister.

This PR doesn't improve UX, but only the logging of the errors. But now, improving the UX could be done by checking the error thrown.

# Changes

* Create a new helper `signerCanisterError` and two new errors `CanisterSignerPaymentError` and `CanisterSignerInternalError`.
* Use the new helper in the btc endpoints to get balance and address.

# Tests

* Change and add tests to prove that the helpers will throw as expected.


lmuntaner:

@DenysKarmazynDFINITY could we redepoy to staging with btc enabled after this PR? This should help us understand the error better.

DenysKarmazynDFINITY:
@lmuntaner yes, let's enable BTC on staging again right after this PR is merged.

peterpeterparker (approved):
LGTM, thx

PR #2575, opened 2024-09-30
feat(test): Install bitcoin canister to pocket-ic

lmuntaner:
# Motivation

Bitcoin canister is needed to perform integration tests in the backend that interact with the bitcoin api.

In this PR, the bitcoin canister is installed as part of the pocket ic setup.

# Changes

* Download the bitcoin canister if not present during test setup.
* Ignore bitcoin canister wasm.
* Refactor `deploy_to` with `deploy_backend` so that we can only install the backend which is needed in the migration test.
* Add bitcoin wasm and canister during the pocket-ic setup.

# Tests

Only test changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ internet_identity.did
 codes.txt
 
 backend.wasm.gz
+ic-btc-canister.wasm.gz
 backend-v*.wasm.gz
 /test-results/
 /playwright-report/
--- a/scripts/test.backend.sh
+++ b/scripts/test.backend.sh
@@ -2,6 +2,8 @@
 
 POCKET_IC_SERVER_VERSION=6.0.0
 OISY_UPGRADE_VERSIONS="v0.0.13,v0.0.19"
+BITCOIN_CANISTER_RELEASE="2024-08-30"
+BITCON_CANISTER_WASM="ic-btc-canister.wasm.gz"
 
 # If a backend wasm file exists at the root, it will be used for the tests.
 
@@ -14,6 +16,15 @@ else
   echo "Building backend canister."
   cargo build --locked --target wasm32-unknown-unknown --release -p backend
 fi
+
+if [ -f "./$BITCON_CANISTER_WASM" ]; then
+  echo "Use existing $BITCON_CANISTER_WASM canister."
+else
+  echo "Downloading bitcoin_canister canister."
+  curl -sSL "https://github.com/dfinity/bitcoin-canister/releases/download/release%2F$BITCOIN_CANISTER_RELEASE/ic-btc-canister.wasm.gz" -o $BITCON_CANISTER_WASM
+fi
+# Setting the environment variable that will be used in the test to load that particular file relative to the cargo workspace.
+export BITCOIN_CANISTER_WASM_FILE="../../$BITCON_CANISTER_WASM"
 
 # We use a previous version of the release to ensure upgradability
 
--- a/src/backend/tests/it/utils/pocketic.rs
+++ b/src/backend/tests/it/utils/pocketic.rs
@@ -1,5 +1,6 @@
 use crate::utils::mock::CALLER;
 use candid::{decode_one, encode_one, CandidType, Principal};
+use ic_cdk::api::management_canister::bitcoin::BitcoinNetwork;
 use pocket_ic::{CallError, PocketIc, PocketIcBuilder, WasmResult};
 use serde::Deserialize;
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -41,6 +41,7 @@ internet_identity.did
 codes.txt
 
 backend.wasm.gz
+ic-btc-canister.wasm.gz
 backend-v*.wasm.gz
```
bitdivine, line 45: At some time someone will replace this list with `*.wasm.gz` :laughing:

lmuntaner:
@bitdivine please review

scripts/test.backend.sh
```diff
@@ -15,6 +17,15 @@ else
...
 fi
 
+if [ -f "./$BITCON_CANISTER_WASM" ]; then
+  echo "Use existing $BITCON_CANISTER_WASM canister."
+else
+  echo "Downloading bitcoin_canister canister."
+  curl -sSL "https://github.com/dfinity/bitcoin-canister/releases/download/release%2F$BITCOIN_CANISTER_RELEASE/ic-btc-canister.wasm.gz" -o $BITCON_CANISTER_WASM
+fi
+# Setting the environment variable that will be used in the test to load that particular file relative to the cargo workspace.
+export BITCOIN_CANISTER_WASM_FILE="../../$BITCON_CANISTER_WASM"
```
bitdivine, line 27: Note that if you put code such as the above in a script, you can do things such as have `dfx deploy` "just work".  See e.g. the cycles depositor script here: https://github.com/dfinity/chain-fusion-signer/blob/main/scripts/build.cycles_depositor.sh  Running `dfx build cycles_depositor` runs that script.

src/backend/tests/it/utils/pocketic.rs
```diff
@@ -143,26 +199,38 @@ impl BackendBuilder {
...
         pic.install_canister(canister_id, wasm_bytes, arg, None);
     }
+    fn install_bitcoin(&mut self, pic: &PocketIc) {
+        let canister_id =
+            Principal::from_text(BITCOIN_CANISTER_ID).expect("Unexpected bitcoin canister id");
+        pic.create_canister_with_id(None, None, canister_id)
+            .expect("Failed creating bitcoin canister");
+        let wasm_bytes = self.bitcoin_wasm_bytes();
+        pic.install_canister(canister_id, wasm_bytes, Self::default_bitcoin_arg(), None);
+    }
```
bitdivine, line 215: This should technically be in a different builder, as there should be one such builder per canister.  But we can resolve tech debt later.

src/backend/tests/it/utils/pocketic.rs
```diff
@@ -50,6 +72,8 @@ pub struct BackendBuilder {
     cycles: u128,
     /// Path to the backend wasm file.
     wasm_path: String,
+    /// Path to the bitcoin canister wasm file.
+    bitcoin_wasm_path: String,
```
bitdivine, line 76: Each canister should really have its own builder.  If this works, let's run with it for now and refactor later.  There should be a test environment containing canisters.  This here is the builder for the backend canister in that environment and there should be a different canister struct in the environment.  But I realize that we are under time pressure and this can be marked up as tech debt, to be cleaned up later.

bitdivine (approved):

.gitignore
lmuntaner: I hope 😂

scripts/test.backend.sh
lmuntaner: I am not sure I want the bitcoin canister in dfx.json. The bitcoin canister should not be installed in the local replica. To have bitcoin in the local replica we need to use a flag from the command line. Then dfx installs the bitcoin canister.

src/backend/tests/it/utils/pocketic.rs
lmuntaner: I followed the same pattern as in the signer canister.

Let's improve it there as well.

src/backend/tests/it/utils/pocketic.rs
lmuntaner: Ditto, I followed the same pattern as the signer canister.

But yes, a different builder makes sense.

src/backend/tests/it/utils/pocketic.rs
lmuntaner: I created the ticket: https://dfinity.atlassian.net/browse/GIX-3033

PR #2565, opened 2024-09-27
style(frontend): change authentication button label [GIX-3015]

AntonioVentilii-DFINITY:
# Motivation

The Authenticate button label will change from `Authenticate` to `Create | Open`.

Screenshot 2024-09-27 at 15 43 47
Screenshot 2024-09-27 at 15 43 59
Screenshot 2024-09-27 at 15 44 06

PR #2577, opened 2024-09-30
refactor(frontend): remove usage of special whitespace character [GIX-3015]

AntonioVentilii-DFINITY:
# Motivation

In PR #2565, I rashly used the special character `&nbsp;` for whitespace. This PR cleans it.

Screenshot 2024-09-30 at 09 39 32
Screenshot 2024-09-30 at 09 39 44
Screenshot 2024-09-30 at 09 39 57


src/frontend/src/lib/components/hero/HeroSignIn.svelte
```diff
@@ -19,5 +18,7 @@
 </div>
 
 <ButtonAuthenticate on:click={async () => await signIn({})}>
-	<Html text={$i18n.auth.text.authenticate} />
+	{$i18n.core.text.open}
```
peterpeterparker, line 21: We do not need a style `gap` somewhere if we remove the `nbsp;`?

AntonioVentilii-DFINITY: I put the `|` character in a span with margins, and it seems to be working well. But it is the same, as you said: i can put them inside something with a `gap` instead of this. What do you prefer? No preference from me

peterpeterparker: Gotcha. All good.

peterpeterparker (approved):
LGTM, thx

PR #2585, opened 2024-09-30
feat(frontend): use Send context for IcSend flow too

AntonioVentilii-DFINITY:
# Motivation

We would like to slowly decommission the `token` store. As one of the first step, we provide a context in the entire Send flow. That means, including the Send context used by Ethereum into the IC Send wizard.

Note: for this PR we just change the immediate child that is `IcSendModal`, but in future PR we will slowly go deeper.

# Changes

- Add `sendTokenSymbol` to the Send context.
- Move `SendTokenContext` component one level up in the `SendWizard` (to include IC flow).
- Add `SendTokenContext` in `IcSendModal` too, for consistency.
- Modify `IcSendWizard` to deprecate token stores and use Send token context.

# Tests

I tried a few transactions in the local replica and they worked.

https://github.com/user-attachments/assets/d1f810ba-82b1-4ae7-9df4-618bc1d78edf

https://github.com/user-attachments/assets/d85c345c-3b84-43e7-b60c-5fe16dfdaed8


peterpeterparker (approved):
LGTM, thx

PR #2587, opened 2024-09-30
feat(frontend): re-use ReceiveModal component

DenysKarmazynDFINITY:
# Motivation

The goal is to move  the`ReceiveModal` component to lib so it can be reused by both ETH and BTC.

### DIFF
--- a/src/frontend/src/eth/components/receive/EthReceive.svelte
+++ b/src/frontend/src/eth/components/receive/EthReceive.svelte
@@ -25,7 +25,7 @@
 </script>
 
 <ReceiveButtonWithModal open={openReceive} isOpen={$modalEthReceive}>
-	<ReceiveModal slot="modal" address={$networkAddress ?? undefined}>
+	<ReceiveModal slot="modal" address={$networkAddress}>
 		<svelte:fragment slot="content">
 			{#if $networkEthereum}
 				<EthReceiveMetamask />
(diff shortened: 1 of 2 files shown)


src/frontend/src/eth/components/receive/EthReceive.svelte
```diff
@@ -23,5 +25,11 @@
 </script>
 
 <ReceiveButtonWithModal open={openReceive} isOpen={$modalEthReceive}>
-	<EthReceiveModal slot="modal" />
+	<ReceiveModal slot="modal" address={$networkAddress ?? undefined}>
```
AntonioVentilii-DFINITY, line 28: the `??` check is not needed here right?

src/frontend/src/eth/components/receive/EthReceive.svelte
```diff
@@ -23,5 +25,11 @@
...
+		<svelte:fragment slot="content">
+			{#if $networkEthereum}
+				<EthReceiveMetamask />
```
AntonioVentilii-DFINITY, line 31: out of curiosity: why the `slot` definition is not here but outside the `if` condition?

DenysKarmazynDFINITY: This is required by Svelte (`Svelte: <svelte:fragment> must be a child of a component`). More discussions here - https://github.com/sveltejs/svelte/issues/5604.

src/frontend/src/eth/components/receive/EthReceive.svelte
DenysKarmazynDFINITY: I adjusted the type of address prop so `??` is now gone.

src/frontend/src/eth/components/receive/EthReceive.svelte
AntonioVentilii-DFINITY: wow! thank you!

AntonioVentilii-DFINITY (approved):
LGTM!

PR #2590, opened 2024-10-01
chore(backend): Rm new_backend

bitdivine:
# Motivation
The backend canister migration is complete but some of the configuration used during the migration still needs to be cleaned away.  In particular, we had `backend` and `new_backend`.  Now the two are the same and we use only `backend`.

# Changes
- Delete `new_backend` from `dfx.json` and everywhere that depends on this.

# Tests
See CI


AntonioVentilii-DFINITY (approved):
LGTM! Tks!

P.S. does that means that the migration scripts will be removed too? Or we leave them?

bitdivine:
Re migration scripts: I'm not sure.

PR #2591, opened 2024-10-01
build(frontend): update agent-js patch v2.1.2

peterpeterparker:
# Motivation

There is a potential issue that was reverted in agent-js [v2.1.2](https://github.com/dfinity/agent-js/releases/tag/v2.1.2).

# Changes

- Update agent-js latest
- Update released ic-js (which use latest agent-js)
- Add two scripts to repeat updates in package.json


AntonioVentilii-DFINITY (approved):
LGTM!

Shall I maybe include the scripts in some CI?

peterpeterparker:
> Shall I maybe include the scripts in some CI?

We can potentially add a manually triggered CI but, I would not automate it because there is no clear pattern one when and how. Sometimes we need this, sometimes that etc. it's handy to have the all upgrade in scripts that way we can either update all or have a reference script.

PR #2592, opened 2024-10-01
chore(signer): Update signer

bitdivine:
# Motivation
A new version of the signer is available.  It features a slightly newer implementation of the PAPI API.  (same API, updated implementation).

# Changes
- Update the signer to version 0.2.2

# Tests
See CI


AntonioVentilii-DFINITY (approved):
LGTM! TKs

PR #2594, opened 2024-10-01
feat(frontend): incorporate text in button authenticate

peterpeterparker:
# Motivation

For simplicity and to avoid the issues we've encountered when updating the sign-in text on the main page but not in the signer, I suggest introducing the authentication call-to-action text with the button labeled "Authenticate."

To some extent, it's not ideal to have text in the `ui` package, but on the other hand, we've already faced these issues twice.

# Changes

- Incorporate text into `ButtonAuthenticate`


AntonioVentilii-DFINITY (approved):
Good one! Tks

PR #2595, opened 2024-10-01
fix(devops): Specify canister IDs

bitdivine:
# Motivation
The signer uses the cycles ledger for payment.  For this to work with pre-built Wasms, the cycles ledger needs to be deployed to the correct canister ID.

# Changes
- Fill in the canister dependencies:
  - the backend depends on the signer and the cycles ledger
  - the signer depends on the cycles ledger
- Specify which canister ID the signer and cycles ledger should be deployed to, so that pre-built Wasms will interact correctly.

# Tests


scripts/generate.sh
```diff
@@ -32,6 +32,7 @@ function install_did_files() {
       canister_name="${array[0]}"
       source="${array[1]}"
       filename="${source##*/}"
+      filename="${filename//-/_}"
```
bitdivine, line 35: The cycles-ledger has a hyphen in the did file name.  dfx uses underscores instead.

lmuntaner: Can you add this comment in the code?

lmuntaner (approved):
LGTM! Just a minor comment.

scripts/generate.sh
bitdivine: Done

PR #2597, opened 2024-10-01
refactor(frontend): use Send context instead of token store in IcSendWizard

AntonioVentilii-DFINITY:
# Motivation

We replace the token store (and its derived) in all the child components of IcSendWizard that are used ONLY in its tree.

NOTE: in some instances, we could have used the Send context derived (for example `$sendToken.decimals` could be `$sendTokenDecimals`), but to avoid any possible disruption or unexpected behavior I kept the same logic.

# Tests

Local replica send flow worked as expected.


peterpeterparker (approved):
LGTM, thx

PR #2598, opened 2024-10-01
feat(backend): Class to manage btc pending transactions

lmuntaner:
# Motivation

We want to store pending btc transactions to disable users from making transactions while there is one that hasn't finished.

In this PR, I introduce the module to manage the pending transactions. It adds functionality to add, read and prune the pending transaction.

This struc will then be added to the heap, (hence that I put it inside a directory `heap_state`) and used in the backend endpoints.

# Changes

* New class `BtcUserPendingTransactions` in module `heap_state/btc_user_pending_tx_state`.
* Add new module to lib.

# Tests

* Test class functionality.

### DIFF
--- a/src/backend/src/heap_state/btc_user_pending_tx_state.rs
+++ b/src/backend/src/heap_state/btc_user_pending_tx_state.rs
@@ -27,7 +27,7 @@ type PendingTransactionsMap = HashMap<BitcoinAddress, Vec<StoredPendingTransacti
 pub struct BtcUserPendingTransactions {
     /// Map of `user_principal` to `PendingTransactionsMap`;
     pending_transactions_map: HashMap<Principal, PendingTransactionsMap>,
-    /// Maximum number of transactions stored per principal.
+    /// Maximum number of transactions that will be stored per `(principal, address)` tuple.
     max_pending_transactions: usize,
     /// Maxumum number of addresses per user.
     max_addresses_per_user: usize,
@@ -127,7 +127,7 @@ impl BtcUserPendingTransactions {
                     *transactions = pruned_list;
                 }
             }
-            for address in addresses_to_remove.iter() {
+            for address in &addresses_to_remove {
                 address_map.remove(address.as_str());
             }
         }
--- a/src/backend/src/heap_state/btc_user_pending_tx_state.rs
+++ b/src/backend/src/heap_state/btc_user_pending_tx_state.rs
@@ -2,6 +2,7 @@ use candid::Principal;
 use ic_cdk::api::management_canister::bitcoin::Utxo;
 use std::collections::HashMap;
 
+#[allow(dead_code)]
 const MAX_PENDING_TRANSACTIONS: usize = 1000;
 #[allow(dead_code)]
 const DAY_IN_NS: u64 = 24 * 60 * 60 * 1_000_000_000;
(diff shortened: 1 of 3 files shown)


lmuntaner:
@bitdivine please review

src/backend/src/heap_state/btc_user_pending_tx_state.rs
```diff
@@ -0,0 +1,423 @@
...
+impl BtcUserPendingTransactions {
+    #[allow(dead_code)]
+    pub fn new(max_pending_txs: Option<usize>) -> Self {
+        Self {
+            pending_transactions_map: HashMap::new(),
+            max_pending_transactions: max_pending_txs.unwrap_or(MAX_PENDING_TRANSACTIONS),
+        }
+    }
+
+    /// Returns the pending transactions of a specific principal per address.
```
bitdivine, line 37: Can multiple users share the same Bitcoin address?  Not if the Bitcoin address is derived from the principal, I guess, but maybe in future.  My concern is that if one user makes a transaction, it may not appear for others sharing the same address.  This is not going to happen if we commit to not sharing addresses.

src/backend/src/heap_state/btc_user_pending_tx_state.rs
```diff
@@ -0,0 +1,423 @@
...
+    ///   Partial presence could happen if the utxos of a pending transaction were not really used in the transaction.
+    ///   We don't remove in partial presence because, in the end, partial presence will be temporary for one day.
+    #[allow(dead_code)]
+    pub fn prune_pending_transactions(
+        &mut self,
+        principal: Principal,
+        current_utxos: &[Utxo],
+        now_ns: u64,
+    ) {
+        if let Some(address_map) = self.pending_transactions_map.get_mut(&principal) {
```
bitdivine, line 96: Do I see correctly that the number of addresses per principal is unbounded?

bitdivine (approved):
Approving, with an incoming change to limit the number of accounts.

src/backend/src/heap_state/btc_user_pending_tx_state.rs
lmuntaner: At the moment, we don't share addreses.

This would be a problem not only for sharing, but if multiple users can make transactions from the same address. Which at the moment is not supported in CFS either.

But it's something to keep in mind for the future. I'll add a comment.

src/backend/src/heap_state/btc_user_pending_tx_state.rs
```diff
@@ -0,0 +1,423 @@
...
+}
+
+type BitcoinAddress = String;
+type PendingTransactionsMap = HashMap<BitcoinAddress, Vec<StoredPendingTransaction>>;
+
+#[allow(dead_code)]
+pub struct BtcUserPendingTransactions {
+    /// Map of `user_principal` to `PendingTransactionsMap`;
+    pending_transactions_map: HashMap<Principal, PendingTransactionsMap>,
+    /// Maximum number of transactions stored per principal.
```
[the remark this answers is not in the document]
lmuntaner: Good point!

src/backend/src/heap_state/btc_user_pending_tx_state.rs
lmuntaner: Fixed!

PR #2603, opened 2024-10-01
chore(build): Do not exclude env files from docker

Sawchord:
# Motivation

To be able to build the Docker frontend files reproducibly, we need to be able to set the environment variables via env files, just like when we are not building inside Docker,

# Changes

Removes the exclusion of `.env` files from `.dockerignore`


AntonioVentilii-DFINITY (approved):
LGTM!

NOTE: discussed offline that at some point we will use some other DOCKER tool to integrate env files

PR #2604, opened 2024-10-01
fix(ci): Run the bindings check when build commands are updated

bitdivine:
# Motivation
The version of the signer was updated in #2592 but CI did not check the bindings.

# Changes
- Include the build scripts in the list of files that cause the bindings check to run.  The build scripts include the external canister version number.

Alternative considered:
- We could add the canister release number in `dfx.json`.  This would be outside the schema, but dfx has never objected to additional useful data being stored in `dfx.json`.  Theoretically the release number and the location of the candid file are all that are needed to specify the candid exactly and if those are both in `dfx.json` we can keep the list of triggers small, leading to faster CI.
  - I leave it up to the reviewer which they prefer.  In the PR I have gone for "conservative but slow" but would be very happy to use this instead.

# Tests

### DIFF
--- a/.github/workflows/binding-checks.yml
+++ b/.github/workflows/binding-checks.yml
@@ -15,6 +15,8 @@ on:
       - # Rust files such as Cargo.lock, Cargo.toml and rust-toolchain.toml
         '**/Cargo*'
       - '**/*rust*'
+      - # Scripts that download external Candid files, with false positives.
+        'scripts/build.*.sh'
       - # This workflow
         '.github/workflows/binding-checks.yml'
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Thanks! This makes sense to me for now.

.github/workflows/binding-checks.yml
```diff
@@ -2,46 +2,57 @@ name: Binding Checks
...
+              - # The backend source code
+                'src/backend/**'
+              - 'src/shared/**'
+              - # Rust files such as Cargo.lock, Cargo.toml and rust-toolchain.toml
+                '**/Cargo*'
+              - '**/*rust*'
+              - # Scripts that download external Candid files, with false positives.
+                'scripts/build.*.sh'
+              - # This workflow
+                '.github/workflows/binding-checks.yml'
```
bitdivine, line 31: Note: Path filters moved here so that we can make this workflow required for merge, but not have workflows hang when a relevant file hasn't changed.

PR #2605, opened 2024-10-01
feat(backend): Get canister IDs

bitdivine:
# Motivation
The backend will need to  call the cycles ledger.  The PR is a bit large so here are a couple of small  preparation steps we can take:

- We will need the signer and cycles ledger canister IDs in the backend.
- We will need a couple of dependencies.

# Changes
- Add the parsed canister principals as lazy static variables.
- Add dependencies in cargo.toml

# Tests
None; these variables are still unused.

### DIFF
--- a/src/backend/src/state.rs
+++ b/src/backend/src/state.rs
@@ -4,6 +4,11 @@ use lazy_static::lazy_static;
 const MAINNET_CYCLES_LEDGER_CANISTER_ID: &str = "um5iw-rqaaa-aaaaq-qaaba-cai";
 const MAINNET_SIGNER_CANISTER_ID: &str = "grghe-syaaa-aaaar-qabyq-cai";
 
+// This gets canister IDs:
+// - `dfx` sets an environment variable with the canister ID.  If this is available, we use it.
+//    See: https://internetcomputer.org/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-envars#canister_id_canistername
+// -  If that variable is not set for any reason, e.g. because we are not building with dfx, we default to the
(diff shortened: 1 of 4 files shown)


AntonioVentilii-DFINITY (approved):
LGTM!

src/backend/src/state.rs
```diff
@@ -0,0 +1,12 @@
...
+use lazy_static::lazy_static;
+
+const MAINNET_CYCLES_LEDGER_CANISTER_ID: &str = "um5iw-rqaaa-aaaaq-qaaba-cai";
+const MAINNET_SIGNER_CANISTER_ID: &str = "grghe-syaaa-aaaar-qabyq-cai";
+
+lazy_static! {
+    pub static ref CYCLES_LEDGER: Principal = Principal::from_text(option_env!("CANISTER_ID_CYCLES_LEDGER").unwrap_or(MAINNET_CYCLES_LEDGER_CANISTER_ID))
+        .unwrap_or_else(|e| unreachable!("The cycles canister ID from DFX and mainnet are valid and should have been parsed.  Is this being compiled in some strange way? {e}"));
+    pub static ref SIGNER: Principal = Principal::from_text(option_env!("CANISTER_ID_SIGNER").unwrap_or(MAINNET_SIGNER_CANISTER_ID))
+        .unwrap_or_else(|e| unreachable!("The signer canister ID from mainnet or dfx valid and should have been parsed.  Is this being compiled in some strange way? {e}"));
```
AntonioVentilii-DFINITY, line 11: Small comment here: should the error string be the same as the above?

bitdivine: It should be the same except for signer/cycles_ledger.  Ah, I have just "cycles".  Updating...

PR #2608, opened 2024-10-01
feat(backend): Endpoint to add and get pending transactions

lmuntaner:
# Motivation

This is the follow up to the PR that added the class to manage pending transactions.

In this PR, I put the class in the heap and add two endpoints that interact with it. One to add pending transactions and one to get them.

Both prune the pending transactions before performing their logic.

# Changes

## Manual changes

* Change the min_confirmations filter in get_all_utxos to use 1 if the network is Regtest. Otherwise, the integration test isn't working.
* Expose the fields in `StoredPendingTransaction` to create an instance before adding it.
* Add state to the heap in `heap_state/state.rs`.
* Add two new endpoints `btc_add_pending_transaction` and `btc_get_pending_transactions` which interact with the class to prune and then add or read the pending transactions.
* New types for new endpoints.

## Automatic Changes

* `backend.did` using `./scripts/did.sh`.
* `declarations/backend/*` using `npm run generate`.

# Tests

* Add an integration tests using both new endpoints one after the other. The test doesn't work as expected because of the pruning pending transactions. I have added some comments to better explain the limitations. I hope we can improve this by mocking calls to the management canister.


lmuntaner:
@bitdivine please review

bitdivine (approved):
Looks good, thank you!

PR #2609, opened 2024-10-02
refactor(tests): Move PicCanisterTrait

bitdivine:
# Motivation
I would like to add more methods to the `PicCanisterTrait` but the file it is in is already large and quite cluttered.

# Changes
- Move the trait into a dedicated file.

# Tests
See CI


lmuntaner (approved):
Thanks!

PR #2610, opened 2024-10-02
feat(frontend): send BTC txs to FE as CertifiedData

DenysKarmazynDFINITY:
# Motivation

BTC txs were sent to FE without "certified flag.


peterpeterparker (approved):
LGTM, thanks

We really gonna need to introduce Zod for those postMessages as we already discussed.

PR #2611, opened 2024-10-02
feat(frontend): create and use "status" translation in txs modals

DenysKarmazynDFINITY:
# Motivation

Create and use "status" transaltion.


peterpeterparker (approved):
Thx!

PR #2612, opened 2024-10-02
fix(frontend): add missing on:click handler

DenysKarmazynDFINITY:
# Motivation

There should an onclick handler in the Transaction component' button.

### DIFF
--- a/src/frontend/src/lib/components/transactions/Transaction.svelte
+++ b/src/frontend/src/lib/components/transactions/Transaction.svelte
@@ -24,7 +24,7 @@
 	$: icon = type === 'send' ? IconSend : IconReceive;
 </script>
 
-<button class="contents">
+<button class="contents" on:click>
 	<Card>
 		<span class="inline-block first-letter:capitalize">{label}</span>
 


src/frontend/src/lib/components/transactions/Transaction.svelte
```diff
@@ -24,7 +24,7 @@
 	$: icon = type === 'send' ? IconSend : IconReceive;
 </script>
 
-<button class="contents">
+<button class="contents" on:click>
```
peterpeterparker, line 27: It works with the class `contents`?

In `IcTransaction`n the style is `class="block w-full border-0"`

DenysKarmazynDFINITY: Yes, it works with just `contents` - in prod, it's already used for ETH transactions. Also, I tried it for IC locally - also looking good. Here is a screenshot of a BTC tx:

Screenshot 2024-10-02 at 08 59 26

I will still triple-check it when creating a PR for using `Transaction.svelte` by all networks just to be on the safe side.

peterpeterparker (approved):
LGTM, thx

PR #2614, opened 2024-10-02
style(frontend): increase max width of header [GIX-3044]

AntonioVentilii-DFINITY:
# Motivation

As part of new requirements, we increase the max width of the header. We use Tailwind, with a custom screen size `2.5xl = 1728px`

NOTE: he previous custom `minWidth` in tailwind config was not used.


src/frontend/src/lib/styles/mixins/_media.scss
```diff
@@ -3,6 +3,7 @@ $breakpoint-small: 576px;
 $breakpoint-medium: 768px;
 $breakpoint-large: 1024px;
 $breakpoint-extra-large: 1300px;
+$breakpoint-extra-extra-large: 1728px;
```
peterpeterparker, line 6: In tailwind they use the term `2xl`, should we use that convention?

Or instead of `max-width: media.$breakpoint-extra-extra-large;` can you actually use a tailwind class?

The layout use css for historical reason but, we can migrate things now to Tailwind.

AntonioVentilii-DFINITY: ah yes, for sure! i thought we wanted CSS, but I can use that for sure

done!

peterpeterparker (approved):
LGTM, thx

PR #2615, opened 2024-10-02
style(frontend): change landing page layout to have two columns [GIX-3044]

AntonioVentilii-DFINITY:
# Motivation

This PR changes the main layout of the landing page when the user is not logged in. Before this PR, when the user is logged out, there was an `HeroSignIn` component that would stay inside the `Hero`, with the rest of the page in a "logged-out" state.

However, now, we make it so every page will have an authorization guard that does not load the page components, but it "simply" show the landing page.

Such page is composed of two parts: the "main" content and a preview image. For large screens they are in a row, while for smaller they move one above the other.

# Changes

- Create `AddressGuard` component in `+layout` to avoid loading the rest of the page when the user is logged out.
- Remove `HeroSignIn` from the `Hero` component.
- Add preview images (NOTE: we tried to reduce the memory size as much as possible).
- Create main layout for the landing page with shifting sub-components, based on screen width.

# Tests

NOTE: in the end of the video it is shown as the rest stayed the same

Screenshot 2024-10-02 at 15 30 06
Screenshot 2024-10-02 at 15 30 18
Screenshot 2024-10-02 at 15 30 34
Screenshot 2024-10-02 at 15 30 42
Screenshot 2024-10-02 at 15 30 48
Screenshot 2024-10-02 at 15 32 14
Screenshot 2024-10-02 at 15 32 28

https://github.com/user-attachments/assets/6d4459db-a3aa-4875-be12-199589648b9b


src/frontend/src/routes/(app)/+layout.svelte
```diff
@@ -24,10 +25,12 @@
 	back={route === 'settings' ? 'header' : route === 'transactions' ? 'hero' : undefined}
 />
 
-<main class="pb-5 pt-8 sm:pb-12">
-	<LoadersGuard>
-		<slot />
-	</LoadersGuard>
-</main>
+<AuthGuard>
```
peterpeterparker, line 28: We want the same login experience on both "Tokens" page as on any other pages ("Settings" and "Transactions"). Correct?

src/frontend/src/lib/components/auth/LandingPage.svelte
```diff
@@ -0,0 +1,30 @@
...
+	<div
+		class="flex w-full flex-1 flex-col items-center text-center md:items-start md:pl-8 md:pt-12 md:text-left"
+	>
+		<div class="md:absolute md:top-1/3">
+			<HeroSignIn />
+		</div>
+	</div>
+
+	<div class=" min-w-[1127px] flex-1 md:pt-12">
+		<picture aria-label={ariaLabel} class="w-full">
```
peterpeterparker, line 25: Should we add a `role=presentation` on the `picture`?

src/frontend/src/lib/components/auth/AuthGuard.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+	import LandingPage from '$lib/components/auth/LandingPage.svelte';
+	import { authNotSignedIn } from '$lib/derived/auth.derived';
+</script>
+
+{#if $authNotSignedIn}
+	<LandingPage />
+{:else}
+	<slot />
```
peterpeterparker, line 9: Is the transition smooth enough? Or should we add a `div in:fade`?

peterpeterparker (commented):
Regarding the screenshots:

- Have you tried modern extensions like `webp`? If not, would it help improve compression a bit?

- This is more of a product question than a technical one: Is it correct that the foundation wants to display a static image on the front page featuring only "Kong Swap" as the sole app? Shouldn't it showcase an imaginary dApp instead? And if "Kong Swap", I think there may be a type, "KongSwap", and I'm not sure it's their official logo.

src/frontend/src/routes/(app)/+layout.svelte
AntonioVentilii-DFINITY: Correct, on all pages. It becomes really a "Landing Page", less of "App Login"

src/frontend/src/lib/components/auth/AuthGuard.svelte
AntonioVentilii-DFINITY: so, i tried with a few transitions but maybe it is too fast, or the first elements are too fast...

I am adding it, just to be smooth in any case!

src/frontend/src/lib/components/auth/LandingPage.svelte
AntonioVentilii-DFINITY: good point! included!

AntonioVentilii-DFINITY:
> Is it correct that the foundation wants to display a static image....

@peterpeterparker  I will inquiry more, but as per my understanding, for now, we have on the pipeline (almost-confirmed) only KongSwap (will check name and logo). So I am not really sure if we should display more dapps, it may be "inappropriate"

src/frontend/src/lib/components/auth/AuthGuard.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+	import { fade } from 'svelte/transition';
+	import LandingPage from '$lib/components/auth/LandingPage.svelte';
+	import { authNotSignedIn } from '$lib/derived/auth.derived';
+</script>
+
+{#if $authNotSignedIn}
+	<LandingPage />
+{:else}
+	<div transition:fade>
```
peterpeterparker, line 10: You may just need `in` here because on logout we reload the all page anyway

src/frontend/src/lib/components/auth/LandingPage.svelte
```diff
@@ -0,0 +1,30 @@
+<script lang="ts">
+	import PreviewLarge from '$lib/assets/preview-large.webp';
```
peterpeterparker, line 2: for consistency I would suggest to name the imported image with lowercase. If I'm not wrong, didnt double checked the code, it's what we generally do when we import `$lib/assets...`

i.e. only component with capital letter.

src/frontend/src/lib/components/auth/LandingPage.svelte
```diff
@@ -0,0 +1,30 @@
...
+>
+	<div
+		class="flex w-full flex-1 flex-col items-center text-center md:items-start md:pl-8 md:pt-12 md:text-left"
+	>
+		<div class="md:absolute md:top-1/3">
+			<HeroSignIn />
+		</div>
+	</div>
+
+	<div class=" min-w-[1127px] flex-1 md:pt-12">
```
peterpeterparker, line 24: I guess there isn't another value that this exotic `1127px`? That's really specific for this page right?

peterpeterparker (approved):
LGTM, thx

Three minor comments if you can have a look before merge please.

src/frontend/src/lib/components/auth/AuthGuard.svelte
AntonioVentilii-DFINITY: Right! Done!

src/frontend/src/lib/components/auth/LandingPage.svelte
AntonioVentilii-DFINITY: You are right, corrected!

src/frontend/src/lib/components/auth/LandingPage.svelte
AntonioVentilii-DFINITY: Correct, at least for now, but I am adding a TODO

PR #2618, opened 2024-10-02
style(frontend): change authenticate button label [GIX-3044]

AntonioVentilii-DFINITY:
# Motivation

New label for `ButtonAuthenticate`.

Screenshot 2024-10-02 at 16 28 30
Screenshot 2024-10-02 at 16 28 39


peterpeterparker (approved):
We should have a generator that change the label every week 😉

Joke beside, I like it better, thanks 👍

PR #2621, opened 2024-10-02
style(frontend): smaller screens will have the same background as larger ones [GIX-3067]

AntonioVentilii-DFINITY:
# Motivation

With this PR we allow smaller screens to have the same background as bigger ones.

Screenshot 2024-10-02 at 18 39 27
Screenshot 2024-10-02 at 18 39 43
Screenshot 2024-10-02 at 18 39 55

### DIFF
--- a/src/frontend/src/app.html
+++ b/src/frontend/src/app.html
@@ -186,8 +186,6 @@
 				width: 100%;
 				min-width: 1728px;
 
-				display: block;
-
 				z-index: -1;
 			}
 		</style>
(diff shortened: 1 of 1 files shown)


src/frontend/src/app.html
```diff
@@ -186,16 +186,10 @@
 				width: 100%;
 				min-width: 1728px;
 
-				display: none;
+				display: block;
```
peterpeterparker, line 189: Given we set it to `blobk`, is this property still required?

AntonioVentilii-DFINITY: right! no, it is not, removed

peterpeterparker (approved):
LGTM, thx

PR #2622, opened 2024-10-02
fix(frontend): prefix btc-transactions derived stores

DenysKarmazynDFINITY:
# Motivation

The goal is to prefix all BTC tx derived stores with "btc".


peterpeterparker (approved):
LGTM, thx

PR #2623, opened 2024-10-02
style(frontend): remove deprecated button class secondary-alt

AntonioVentilii-DFINITY:
# Motivation

The `secondary-alt` button class could be substituted by `primary` class, according to the design team (the difference is just the background color). So, that class is deprecated.

# Results

Here the only two buttons that still had the deprecated class.

Screenshot 2024-10-02 at 18 49 21
Screenshot 2024-10-02 at 18 50 04


peterpeterparker (approved):
Thx

PR #2624, opened 2024-10-02
style(frontend): new tertiary-link class for buttons [GIX-3048]

AntonioVentilii-DFINITY:
# Motivation

We introduce a new class of buttons called `tertiary-link`. Its first usage is for the `ButtonIcon` components. 

### Normal state

Screenshot 2024-10-02 at 19 09 28

### Hover state

Screenshot 2024-10-02 at 19 09 35

### Usage with all states

https://github.com/user-attachments/assets/284bf65d-f9cc-4ecb-b230-cb3654c70bea

### DIFF
--- a/src/frontend/src/lib/components/ui/ButtonIcon.svelte
+++ b/src/frontend/src/lib/components/ui/ButtonIcon.svelte
@@ -5,7 +5,7 @@
 </script>
 
 <button
-	class="tertiary-link icon flex flex-col text-center text-xs font-normal"
+	class="tertiary link icon flex flex-col text-center text-xs font-normal"
 	bind:this={button}
 	on:click
 	aria-label={ariaLabel}
--- a/src/frontend/src/lib/components/ui/ButtonIcon.svelte
+++ b/src/frontend/src/lib/components/ui/ButtonIcon.svelte
@@ -5,14 +5,12 @@
 </script>
 
 <button
-	class="icon text-primary flex flex-col text-center text-xs font-normal"
+	class="tertiary-link icon flex flex-col text-center text-xs font-normal"
 	bind:this={button}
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/ui/ButtonIcon.svelte
```diff
@@ -5,14 +5,12 @@
 </script>
 
 <button
-	class="icon text-primary flex flex-col text-center text-xs font-normal"
+	class="tertiary-link icon flex flex-col text-center text-xs font-normal"
```
peterpeterparker, line 8: Curiosity question, why not just calling it "tertiary"? Why "-link"?

AntonioVentilii-DFINITY: According to plans and Figma, there is a "tertiary" class

peterpeterparker: Different than `tertiary-link`? I know it's minimal, just feel like that adding a suffix here feels like already adding an exception you know what I mean?

AntonioVentilii-DFINITY: understood! well the difference is really the color set ultimately

Do you prefer that I create a class `tertiary` with a sub class `link`

peterpeterparker: Oh yes, I would rather like that if it makes sense to you too?

AntonioVentilii-DFINITY: yes yes., done! have a look

peterpeterparker (approved):
LGTM, thx

PR #2625, opened 2024-10-02
style(frontend): remove focus state if button is clicked

AntonioVentilii-DFINITY:
# Motivation

Opinionated PR: when a button is clicked, it should not show the `focus` state colors. The `focus` state is visible only when the button is selection interaction.

### Before

https://github.com/user-attachments/assets/c08c79a8-ee4e-4ea7-bf88-2e3cbbcb2d97

### After

https://github.com/user-attachments/assets/290770dd-ace8-4b25-9886-f1f6431ad212

### DIFF
--- a/src/frontend/src/lib/styles/global/button.scss
+++ b/src/frontend/src/lib/styles/global/button.scss
@@ -63,6 +63,20 @@ button {
 				0 0 0 2px var(--color-white),
 				0 0 0 4px var(--color-blue-ribbon);
 		}
+
+		&:focus:not(:focus-visible) {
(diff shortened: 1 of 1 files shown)


peterpeterparker (commented):
Agree with you. Just to be sure, when you select the button keyboard "Tab", the focus is still visible right?

AntonioVentilii-DFINITY:
Correct! The last part of the video is when i use the keyboard (I know, it is not much obvious eheh)

peterpeterparker (approved):
Super! Thx

PR #2627, opened 2024-10-03
feat(backend): Check pending transaction before selecting utxos

lmuntaner:
# Motivation

Avoid users double spending their utxos.

In this PR, before we select the utxos for a transaction, we check whether the user has any pending utxos. If that is the case, then the endpoint fails.

Ideally, we would ignore the pending utxos and use other utxos. But this is an improvement we can do later.

# Changes

* Add one entry to enum `SelectedUtxosFeeError`: `PendingTransactions`.
* Prune pending utxos and check whether there are any within `btc_select_user_utxos_fee`.

# Tests

* I added a test to check the return value of the new error. For this, I checked that if we don't prune, the error is returned.


lmuntaner:
@bitdivine please review

bitdivine (approved):
LGTM, thanks!

PR #2628, opened 2024-10-03
feat(tests): Deploy the cycles ledger & depositor

bitdivine:
# Motivation
The cycles ledger is not deployed into test environments.  We will need it there.

# Changes
- Add `dfx deploy cycles_{ledger,depositor}` to `scripts/deploy.sh`

Alternatives considered:
- Replace the whole contents of `deploy.sh` with `dfx deploy` to deploy everything in one stroke.
  - There are probably details why this is not currently possible.  Maybe something to look into after launch.

# Tests
See existing CI.  Presumably `deplpoy.sh` is exercised and will fail if the new deployments fail.


AntonioVentilii-DFINITY (approved):
LGTM! Tks

PR #2629, opened 2024-10-03
feat(frontend): handle backend and signer canisters errors

DenysKarmazynDFINITY:
# Motivation

The goal is to handle Backend canister errors, for now only for BTC endpoints.

### DIFF
--- a/src/frontend/src/lib/canisters/backend.errors.ts
+++ b/src/frontend/src/lib/canisters/backend.errors.ts
@@ -4,15 +4,15 @@ import type {
 } from '$declarations/backend/backend.did';
 import { CanisterInternalError } from '$lib/canisters/errors';
 
-export const btcPendingTransactionError = (response: BtcAddPendingTransactionError) => {
+export const mapBtcPendingTransactionError = (response: BtcAddPendingTransactionError) => {
 	if ('InternalError' in response) {
 		return new CanisterInternalError(response.InternalError.msg);
 	}
 
-	return new Error('Unknown BtcAddPendingTransactionError');
+	return new CanisterInternalError('Unknown BtcAddPendingTransactionError');
 };
 
-export const btcSelectUserUtxosFeeError = (response: SelectedUtxosFeeError) => {
+export const mapBtcSelectUserUtxosFeeError = (response: SelectedUtxosFeeError) => {
 	if ('InternalError' in response) {
 		return new CanisterInternalError(response.InternalError.msg);
 	}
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/canisters/backend.errors.ts
```diff
@@ -0,0 +1,27 @@
+import type {
+	BtcAddPendingTransactionError,
+	SelectedUtxosFeeError
+} from '$declarations/backend/backend.did';
+import { CanisterInternalError } from '$lib/canisters/errors';
+
+export const btcPendingTransactionError = (response: BtcAddPendingTransactionError) => {
```
DenysKarmazynDFINITY, line 7: Note: in some cases, it makes sense to have one function per several related endpoints, e.g. - `btcAddPendingTransaction` and `btcGetPendingTransaction` share a common error type on the BE side, therefore 2 error parsers are not needed.

src/frontend/src/lib/canisters/backend.errors.ts
```diff
@@ -0,0 +1,27 @@
...
+	if ('InternalError' in response) {
+		return new CanisterInternalError(response.InternalError.msg);
+	}
+
+	return new Error('Unknown BtcAddPendingTransactionError');
```
peterpeterparker, line 12: I would throw a particular type of error here as well. It can be `CanisterInternalError` or another type but, not using a generic `Error` can be interesting at some point. Also reflects the pattern we have in ic-js.

src/frontend/src/lib/canisters/backend.errors.ts
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export const btcSelectUserUtxosFeeError = (response: SelectedUtxosFeeError) => {
```
peterpeterparker, line 15: In `ic-js` we often prefix those function to map error with the keyword, `map`. Should we do the same here?

e.g. `mapBtcSelectUserUtxosFeeError`

src/frontend/src/lib/canisters/backend.errors.ts
DenysKarmazynDFINITY: Agreed 👍

src/frontend/src/lib/canisters/backend.errors.ts
DenysKarmazynDFINITY: Sounds good, done 👍

peterpeterparker (approved):
LGTM, thx

PR #2632, opened 2024-10-03
feat(frontend): init signer allowance before validating addresses

peterpeterparker:
# Motivation

We aim to introduce a new API that utilizes the `Signer` to load BTC and ETH addresses. This requires the frontend to request, at most once per session, that the signer provide an allowance for the client to derive the addresses.

This pattern will need to be implemented for two use cases in Oisy, as we differentiate between a fresh sign-in (with a blank client state) and returning users who have cached addresses on their side (in IndexedDB).

The first use case has been implemented in PR #2650.

This PR addresses the second use case. When the user signs in to Oisy with existing addresses saved in their browsers' chache. Adresses which need to be validate with update calls in the background to ensure those are still valid and matching users' principals.

# Schema

image

# Changes

- New global event `oisyValidateAddresses`. We do not need to save the fact that the addresses need to be validated, therefore we do not need a store. Plus, using an event allow some flexibility and is relatively readable.
- Emit `oisyValidateAddresses` from the `Loader` when user is signed-in without "splash modal" - i.e. when IDB addresses are used.
- Integrate `initSignerAllowance` and window listener for the event in to `AddressGuard`.
- Update `AddressGuard` to prevent any race condition, i.e. validate addresses when stores are updated and when the event is emitted. Note that the services that validate the addresses takes care of performing a unique update call when required - i.e. the related services can be called multiple times.


src/frontend/src/tests/lib/components/guard/AdressGuard.spec.ts
AntonioVentilii-DFINITY, line 1: Small observation: the name of this test should have two DD ? `AddressGuard`?

src/frontend/src/lib/components/guard/AddressGuard.svelte
```diff
@@ -1,15 +1,38 @@
...
+			return;
+		}
+
+		await Promise.allSettled([
+			validateEthAddress($ethAddressStore),
+			...(NETWORK_BITCOIN_ENABLED ? [validateBtcAddressMainnet($btcAddressMainnetStore)] : [])
+		]);
+	};
+
+	$: $btcAddressMainnetStore, $ethAddressStore, (async () => await validateAddresses())();
```
AntonioVentilii-DFINITY, line 33: shouldn't `signerAllowanceLoaded` be among the reactive dependencies?

src/frontend/src/tests/lib/components/guard/AdressGuard.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+import * as authServices from '$lib/services/auth.services';
+import * as loaderServices from '$lib/services/loader.services';
+import { btcAddressMainnetStore, ethAddressStore } from '$lib/stores/address.store';
+import { authStore } from '$lib/stores/auth.store';
+import { emit } from '$lib/utils/events.utils';
+import { Ed25519KeyIdentity } from '@dfinity/identity';
+import { render } from '@testing-library/svelte';
+import { expect, type MockInstance } from 'vitest';
+
+describe('AdressGuard', () => {
```
AntonioVentilii-DFINITY, line 15: same for the double D: `AddressGuard`

src/frontend/src/tests/lib/components/guard/AdressGuard.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+			});
+		});
+	});
+
+	describe('Validate addresses', () => {
+		const mockEthAddress = '0x1d638414860ed08dd31fae848e527264f20512fa75d7d63cea9bbb372f020000';
+		const mockBtcAddress = '17VZNX1SN5NtKa8UQFxwQbFeFc3iqRYhem';
+
+		beforeEach(() => {
+			vi.spyOn(btcEnv, 'NETWORK_BITCOIN_ENABLED', 'get').mockReturnValue(true);
```
AntonioVentilii-DFINITY, line 78: Since this is not identified as a variable, when we are going to remove this flag from the project, will this be "flagged" as issue, or as unnecessary? if not, may you please add a TODO, just to keep track?

AntonioVentilii-DFINITY (approved):
LGTM! 

Just a few small observations please before merging?

src/frontend/src/tests/lib/components/guard/AdressGuard.spec.ts
peterpeterparker: Good catch thx

src/frontend/src/lib/components/guard/AddressGuard.svelte
peterpeterparker: Discussed offline. With current pattern we save one call to validateAddresses

src/frontend/src/tests/lib/components/guard/AdressGuard.spec.ts
peterpeterparker: Sure, added a TODO. Thx.

src/frontend/src/tests/lib/components/guard/AdressGuard.spec.ts
peterpeterparker: Done

PR #2633, opened 2024-10-03
refactor(frontend): Move source address in Send modal

lmuntaner:
# Motivation

We want to reuse the `SendModal` component for Bitcoin transactions.

There is one part that is specific to IC, which is the address shown in the form and review steps.

In this PR, I move this part up the chain so that we can pass later the bitcoin address instead of the IC one.

This will also help me find pending transactions, which will be keyed by address in a store and disable the form if any is present. This way, the idea of pending transactions doesn't need to be specific to BTC.

# Changes

* Add prop `source` to `IcSendReview` and `IcSendForm` and use it instead of `icrcAccountIdentifierText` store.
* Add prop `IcSendTokenWizard` and `SendWizard` to pass it down.
* Create local variable `source` using `icrcAccountIdentifierText` store in `SendModal` and `IcSendModal` to then pass it down.

# Tests

Still working locally.

### DIFF
--- a/src/frontend/src/lib/components/send/SendModal.svelte
+++ b/src/frontend/src/lib/components/send/SendModal.svelte
@@ -14,6 +14,7 @@
 	import type { Token } from '$lib/types/token';
 	import { closeModal } from '$lib/utils/modal.utils';
 	import { goToWizardSendStep } from '$lib/utils/wizard-modal.utils';
+	import { icrcAccountIdentifierText } from '$icp/derived/ic.derived';
 
 	export let destination = '';
 	export let targetNetwork: Network | undefined = undefined;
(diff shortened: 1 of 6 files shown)


lmuntaner:

@DenysKarmazynDFINITY FYI

src/frontend/src/lib/components/send/SendModal.svelte
```diff
@@ -64,6 +65,9 @@
 		};
 		await loadTokenAndRun({ token, callback });
 	};
+
+	let source: string;
+	$: source = $icrcAccountIdentifierText ?? '';
```
peterpeterparker, line 70: We should add a TODO here right as the source won't remain used as `$icrcAccountIdentifierText` in this component? Do I get it right.

lmuntaner: That is right. @DenysKarmazynDFINITY should take care of it when implementing sending bitcoin. I wanted to keep the same logic in this refactor. I added a comment

lmuntaner:

peterpeterparker (approved):
LGTM, thx

PR #2635, opened 2024-10-03
feat(frontend): Add btc pending transactions store

lmuntaner:
# Motivation

We will disable sending BTC tokens while the user has pending transaction. Pending transactions will come from Oisy backend.

In this PR, I created a store to save the pending transactions per address.

# Changes

* New store `pendingTransactionsStore`.

# Tests

* Test new store.

### DIFF
--- a/src/frontend/src/btc/stores/btc-pending-transactions.store.ts
+++ b/src/frontend/src/btc/stores/btc-pending-transactions.store.ts
@@ -1,6 +1,7 @@
 import type { PendingTransaction } from '$declarations/backend/backend.did';
 import { writable, type Readable } from 'svelte/store';
 
+// The endpoint can't be called with a query. Therefore, the information is always certified with an update call.
 type Address = string;
 type BtcPendingTransactionsStoreData = Record<Address, Array<PendingTransaction>>;
 
--- a/src/frontend/src/btc/stores/btc-pending-transactions.store.ts
+++ b/src/frontend/src/btc/stores/btc-pending-transactions.store.ts
@@ -1,9 +1,9 @@
 import type { PendingTransaction } from '$declarations/backend/backend.did';
+import type { CertifiedData } from '$lib/types/store';
 import { writable, type Readable } from 'svelte/store';
 
-// The endpoint can't be called with a query. Therefore, the information is always certified with an update call.
 type Address = string;
-type BtcPendingTransactionsStoreData = Record<Address, Array<PendingTransaction>>;
+type BtcPendingTransactionsStoreData = Record<Address, CertifiedData<Array<PendingTransaction>>>;
 
 interface BtcPendingTransactionsStore extends Readable<BtcPendingTransactionsStoreData> {
 	setPendingTransactions: (params: {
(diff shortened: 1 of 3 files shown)


lmuntaner:

src/frontend/src/btc/stores/btc-pending-transactions.store.ts
```diff
@@ -0,0 +1,45 @@
+import type { PendingTransaction } from '$declarations/backend/backend.did';
+import { writable, type Readable } from 'svelte/store';
+
+type Address = string;
+type BtcPendingTransactionsStoreData = Record<Address, Array<PendingTransaction>>;
+
+interface BtcPendingTransactionsStore extends Readable<BtcPendingTransactionsStoreData> {
```
peterpeterparker, line 7: Should the data hold in the store contains the information if those are certified or not? extending for example `CertifiedData` and `CertifiedStore`

lmuntaner: It will always be certified. This endpoint can't be called with a query. The endpoint also prunes outdated pending transactions.

This way I didn't need to implement a background tasks updating them.

I added a comment because it's not obvious. Good point.

lmuntaner:

src/frontend/src/btc/stores/btc-pending-transactions.store.ts
peterpeterparker: Instead of a comment, what do you think of using the `CertifiedData` type, excluding `certified` and replacing it with `certified: false` as type. That way even code wise it's specified that it's always certified?

lmuntaner: Do you mean setting `certified: true` always?

Sure I can do that as well.

lmuntaner: Done

lmuntaner:

src/frontend/src/btc/stores/btc-pending-transactions.store.ts
peterpeterparker: > Do you mean setting certified: true always?

Yes sorry, stupid typo.

peterpeterparker: Cool. So what I meant is also the following:

```
type AlwaysCertifiedData<T> = Omit<CertifiedData<T>, "certified"> & {certified: true};
type BtcPendingTransactionsStoreData = Record<Address, AlwaysCertifiedData<Array<PendingTransaction>>>;
```

Maybe `AlwaysCertifiedData` can even find place in the same module as `CertifiedData`?

lmuntaner: Ah, sorry, for the confusion. Done!

peterpeterparker: I mean, my bad, I started with a typo ;)

peterpeterparker (approved):
LGTM, thanks

PR #2637, opened 2024-10-03
chore(frontend): add TODO for ERC20 token conversion issue for approval

AntonioVentilii-DFINITY:
# Context

# Example

  1. `approve` transaction for 10 USDT --> successful
  2. `transfer_from` transaction for 10 USDT --> failed (for whatever reason)
  1. `approve` transaction for 20 USDT --> failed (the previous approved amount is still there)
  2. `transfer_from` transaction for 20 USDT --> failed (not enough approved amount)
  1. `approve` transaction for 7 USDT --> failed (the previous approved amount is still there)
  2. `transfer_from` transaction for 7 USDT --> successful (remaining approved amount is 3 USDT)

# Practical example

### DIFF
--- a/src/frontend/src/eth/services/send.services.ts
+++ b/src/frontend/src/eth/services/send.services.ts
@@ -264,6 +264,14 @@ export const send = async ({
 
 	const nonce = await getTransactionCount(from);
 
+	// TODO: We may need to add an approve transaction that resets the approved amount to zero, in case of conversion from ERC20.
+	// 			CONTEXT:
+	// 			It may happen that after the approval transaction is done, the transfer transaction may fail for whatever reasons.
+	//			In this case, the next approval transaction will be rejected, since there is already an approved amount.
+	//			So, the next transfer transaction will work only if the amount is below the one already approved.
+	//			Otherwise, the user is blocked until it consumes all the approved amount.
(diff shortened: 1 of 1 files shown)


src/frontend/src/eth/services/send.services.ts
```diff
@@ -264,6 +264,14 @@ export const send = async ({
 
 	const nonce = await getTransactionCount(from);
 
+	// TODO: We may need to add an approve transaction that resets the approved amount to zero, in case of conversion from ERC20.
```
peterpeterparker, line 267: Instead of resetting, couldn't we potentially know if the spender is already entitled to spend the amount and skip the approval if it's the case?

I guess this has been discussed with cross-chain team?

AntonioVentilii-DFINITY: yes, correct, we could query the allowance endpoint of the ERC20 to check. For example:

### Case 1:
- query allowance
- no amount approved --> ask for approval

### Case 2:
- query allowance
- old amount approved --> reset amount
- ask for new approval

That what you mean?

peterpeterparker: No, I mean:

# Case 3:

- query allowance
- is allowance still valid?
- yes, ask for new approval
- no, request allowance, ask for new approval

Would that be possible ? I really don't know, just a question.

PR #2641, opened 2024-10-03
fix(frontend): background was not hiding overflow

AntonioVentilii-DFINITY:
# Motivation

After PR #2621 , there is an error in the background. Since the big background is used for smaller screen, there is no overflow control and the screen ends up being larger and longer.

To solve it we wrap the background in a container that will manage the overflow (and the general position), while the background image keeps its dimensions.

### Before

https://github.com/user-attachments/assets/56c4d4b1-e37a-467c-8c0d-3bf8876294d1

### After

https://github.com/user-attachments/assets/9d37af36-f159-45ad-9172-7af762c21b7b

### DIFF
--- a/src/frontend/src/app.html
+++ b/src/frontend/src/app.html
@@ -176,7 +176,7 @@
 				}
 			}
 
-			#app-background {
+			#app-background-container {
 				position: absolute;
 				top: 0;
 				left: 50%;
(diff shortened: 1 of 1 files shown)


peterpeterparker (commented):
Looks good but, I'm now just wondering how does it looks like in Figma. Is it correct that only the top left part of the background should be displayed on small device?

AntonioVentilii-DFINITY:
@peterpeterparker for now yes, but we are going to change the background soon enough (in that case it will not be an issue to have it stick to top-left, according to design)

peterpeterparker (approved):
LGTM, thx

PR #2642, opened 2024-10-03
fix(frontend): modals where out of viewport in mobile Safari

AntonioVentilii-DFINITY:
# Motivation

In mobile Safari, the modals were going off-screen. To solve it on mobile, we use the same `-webkit-fill-available` that we were using before to control the height of smaller screens

### Before

Screenshot 2024-10-03 at 21 52 59
Screenshot 2024-10-03 at 21 53 08

### After

Screenshot 2024-10-03 at 21 53 18
Screenshot 2024-10-03 at 21 53 23
Screenshot 2024-10-03 at 21 53 28

### Other browsers

Screenshot 2024-10-03 at 21 53 51
Screenshot 2024-10-03 at 21 54 04


peterpeterparker (approved):
Nice, thanks!

PR #2643, opened 2024-10-04
chore(frontend): remove error constructor

peterpeterparker:
# Motivation

Overriding the constructor in `CanisterInternalError` is not needed given that it overrides the default with the same declaration.


AntonioVentilii-DFINITY (approved):
Good one!

PR #2644, opened 2024-10-04
chore(frontend): improve canister error code

peterpeterparker:
# Changes

- `response` is of type error so rename to `err`
- `SignerCanisterBtcError` type is not needed and can be inlined in its import
- Return types are missing
- `SignerCanisterPaymentError` should extend `CanisterInternalError`


AntonioVentilii-DFINITY (approved):
LGTM!

PR #2645, opened 2024-10-04
feat(frontend): handle allow signing errors

peterpeterparker:
# Motivation

If `allow_signing` throws error we want to map to errors to internal types and bubble those from the canister.

# Changes

- Add mapper and throw error at the "canister" level.

# Notes

The test regarding `ApproveError` is simplified, meaning that not all use cases are tested, as explained in the comment within the test code.


src/frontend/src/tests/lib/canisters/backend.canister.spec.ts
```diff
@@ -583,5 +585,62 @@ describe('backend.canister', () => {
 
 			await expect(res).rejects.toThrow(mockResponseError);
 		});
+
+		// We do not test all types of ApproveError:
+		// - for simplicity
+		// - because the utility mapIcrc2ApproveError, we are using to map those error types, is already covered by tests in ic-js
+		// - and we do not differentiate the error in Oisy anyway
```
DenysKarmazynDFINITY, line 592: 👍

DenysKarmazynDFINITY (approved):
LGTM, thanks!

PR #2647, opened 2024-10-04
fix(frontend): loading banner images layout shift

peterpeterparker:
# Motivation

This PR addresses a glitch when loading "banner images" on mainnet, where loading assets takes some time, causing a layout shift if no minimal height is set.

# Notes

I don't like adding unnecessary elements to the DOM. This issue might have been resolved by simply applying some styling to the `Img` component. However, since the glitch primarily affects banners, I felt this solution was less likely to introduce unexpected errors, especially as the application does not yet have sufficient E2E snapshot testing coverage.

# Changes

- Create new `ImgBanner` component that contains a component with a `min-height` and includes the img
- Use component in "What can I do with my tokens" and "How Oisy Wallet works"
- Use component in loader modal. Note that here there used to be a minimal height but I think it become broken with last changes


AntonioVentilii-DFINITY (approved):
Approved, since I imagine you already tested the results.

Just for posterity, could you please add some screenshots in the description?

peterpeterparker:
> some screenshots in the description?

I did not provided screenshots because, it still look the same. I would rather need to record GIFs on mainnet and deploy a version on mainnet to show the difference. Or faking the loading with some timeout locally. Just felt like it was quite some efforts that I could spare. Hope it makes sense.

PR #2649, opened 2024-10-04
feat(frontend): sign-out with error

peterpeterparker:
# Motivation

There is a new use case that will requires to force a sign-out, as we already do with warning, but with displaying an error. That's why this PR adds a function to do so.

# Changes

- new `errorSignOut` similar to existing `warnSignOut` but with error toast type

# Screenshot

Capture d’écran 2024-10-04 à 09 29 53


AntonioVentilii-DFINITY (approved):
LGTM!

As discussed offline, it will be used in another PR

PR #2650, opened 2024-10-04
feat(frontend): init signer allowance on sign-in before getting addresses

peterpeterparker:
# Motivation

We aim to introduce a new API that utilizes the `Signer` to load BTC and ETH addresses. This requires the frontend to request, at most once per session, that the signer provide an allowance for the client to derive the addresses.

This pattern will need to be implemented for two use cases in Oisy, as we differentiate between a fresh sign-in (with a blank client state) and returning users who have cached addresses on their side (in IndexedDB).

This PR addresses the first use case. When the user signs in and the 'Loading' modal is displayed, we should initiate the signer allowance before fetching the mainnet addresses, as we already do.

# Schema

image

# Notes

This PR provides both services and usage examples for a comprehensive review, given that the usage itself does not require many lines of code.

# Changes

- Create function `initSignerAllowance` which initialize the allowance and sign-out the user in case of any error
- Consumes the new function in the `Loader` component

# Screenshot

Capture d’écran 2024-10-04 à 09 55 28


src/frontend/src/lib/components/core/Loader.svelte
AntonioVentilii-DFINITY, line 1: just for a future TODO: maybe we should extract this entire loading function at some point, and maybe create some test

src/frontend/src/lib/i18n/en.json
```diff
@@ -127,7 +127,8 @@
 			"erc20_user_tokens": "Error while loading the ERC20 user tokens.",
 			"erc20_user_token": "Error while enabling the ERC20 twin token.",
 			"icrc_custom_token": "Error while loading the ICRC custom token.",
-			"loading_wallet_timeout": "Your wallet requires some initial data which has not been loaded yet. Please try again."
+			"loading_wallet_timeout": "Your wallet requires some initial data which has not been loaded yet. Please try again.",
+			"allow_signing": "An issue occurred while setting up your wallet for signing. You've been signed out. Please sign in again, and if the issue persists, contact support."
```
AntonioVentilii-DFINITY, line 131: Theoretically, in the user view, he is not yet signed in: maybe we can remove the "You've been signed out."

Or not?

src/frontend/src/lib/services/loader.services.ts
```diff
@@ -0,0 +1,37 @@
...
+import { i18n } from '$lib/stores/i18n.store';
+import type { ResultSuccess } from '$lib/types/utils';
+import { get } from 'svelte/store';
+
+/**
+ * Initializes the signer allowance by calling `allow_signing`.
+ *
+ * This function should be called once during boot time before retrieving ETH or BTC addresses.
+ * It allocates a cycles budget sufficient for a reasonable number of signer calls.
+ * A "reasonable" number is currently defined as 30 calls, allowing the user to retrieve their ETH and BTC addresses
```
AntonioVentilii-DFINITY, line 13: I know this is not the scope of this PR, but is this number definitive? What was the reason behind 30? You most probably already discussed it ad-nauseam

src/frontend/src/lib/services/loader.services.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @throws Will trigger a sign-out if `allow_signing` fails.
+ */
+export const initSignerAllowance = async (): Promise<ResultSuccess> => {
+	try {
+		const { identity } = get(authStore);
+
+		await allowSigning({ identity });
+	} catch (err: unknown) {
+		// In the event of any error, we sign the user out, as we assume that the Oisy Wallet cannot function without ETH or Bitcoin addresses.
+		await errorSignOut(get(i18n).init.error.allow_signing);
```
AntonioVentilii-DFINITY, line 31: would it make sense to call the `signOut` function here? to eventually clean the storage, etc (in the same way we do in case the loading of address is not successful)

src/frontend/src/lib/i18n/en.json
peterpeterparker: Sure!

src/frontend/src/lib/services/loader.services.ts
peterpeterparker: Comment is literally what @bitdivine told me. So is probably better suited to answer your question.

src/frontend/src/lib/services/loader.services.ts
peterpeterparker: not sure what you mean? it's literally what `errorSignOut` does.

src/frontend/src/lib/components/core/Loader.svelte
peterpeterparker: On it. Might take a bit, there are lots of things missing in the current test setup but, that way it will be set.

AntonioVentilii-DFINITY: nono, i was telling for another PR and another time eheh

src/frontend/src/lib/services/loader.services.ts
AntonioVentilii-DFINITY: I will ask separately, thank you!

AntonioVentilii-DFINITY (approved):
LGTM!

P.S. another instance where E2E test among CI would be very much useful

src/frontend/src/lib/components/core/Loader.svelte
peterpeterparker: no but you have a good point, we should stop f-ing around with quality particurly when it comes to the absolute core process of the dApp.

Anyways I'm not sure if we can merge it in main without arming next deploy on staging. So as long as I don't know that, that's something I can work on. If we can merge, I'll add a todo and work on the test on the side.

I'll let know.

peterpeterparker: The tests will be provided in another PR #2673

PR #2652, opened 2024-10-04
feat(frontend): Derived store to find pending transactions

lmuntaner:
# Motivation

We want to display a warning messange and disable the send form if the modal is for bitcoin and the user has pending transactions.

The bitcoin addresses are unique. Therefore, I wanted to take advantage of that instead of checking selected networks or tokens.

The idea is to create a derived store creator that will return `true`, `false` or `"loading"` depending on the address and whether the bitcoin addresses have been loaded:

* Return `true` when the address is a bitcoin address and the pending transactions array is not empty.
* Return `"loading"` if the addresses or the pending transactions are not loaded. This is trickier because we also need to check the `testnets`. If testnets are not enabled, then the addresses for testnet and regtest won't be loaded.
* Return `false` if the pending transactions are present and empty.

This derived store will be used in `IcSendForm`:

```
let hasPendingTransactionsStore: Readable<boolean | "loading">;
$: hasPendingTransactionsStore = initHasPendingTransactionsStore(source);

//...

{#if $hasPendingTransactionsStore}
  <WarningMessage>{...}</>
```

# Changes

* New derived store creator `initHasPendingTransactions`.

# Tests

* Add tests for the new derived store.

### DIFF
--- a/src/frontend/src/btc/derived/has-pending-transactions.derived.ts
+++ b/src/frontend/src/btc/derived/has-pending-transactions.derived.ts
@@ -1,0 +1,65 @@
+import { pendingTransactionsStore } from '$btc/stores/btc-pending-transactions.store';
+import { LOCAL } from '$lib/constants/app.constants';
+import {
+	btcAddressMainnet,
+	btcAddressRegtest,
+	btcAddressTestnet
+} from '$lib/derived/address.derived';
+import { testnets } from '$lib/derived/testnets.derived';
+import { isNullish } from '@dfinity/utils';
+import { derived, type Readable } from 'svelte/store';
+
+export const hasPendingTransactions = (address: string): Readable<boolean | 'loading'> =>
+	derived(
+		[btcAddressMainnet, btcAddressTestnet, btcAddressRegtest, testnets, pendingTransactionsStore],
+		([
+			$btcAddressMainnet,
+			$btcAddressTestnet,
+			$btcAddressRegtest,
+			$testnets,
+			$pendingTransactionsStore
+		]) => {
+			if (!$testnets) {
+				if (isNullish($btcAddressMainnet)) {
+					// Return loading while we don't have btc addresses and we can't tell
+					// whether there are pending transactions for a specific address.
+					return 'loading';
+				}
+				if ($btcAddressMainnet !== address) {
+					// If the address is not a bitcoin address, there are no pending transactions.
+					return false;
+				}
+				const pendingTransactions = $pendingTransactionsStore[address];
+				if (isNullish(pendingTransactions)) {
+					// Address is bitcoin but we pending transactions are not loaded yet.
+					return 'loading';
+				}
+				return pendingTransactions.data.length > 0;
+			}
+			if (
+				isNullish($btcAddressMainnet) ||
+				isNullish($btcAddressTestnet) ||
+				(LOCAL && isNullish($btcAddressRegtest))
+			) {
+				// Return loading while we don't have btc addresses and we can't tell
+				// whether there are pending transactions for a specific address.
+				return 'loading';
+			}
+
+			if (
+				$btcAddressMainnet !== address &&
+				$btcAddressTestnet !== address &&
+				$btcAddressRegtest !== address
+			) {
+				// If the address is not a bitcoin address, there are no pending transactions.
+				return false;
+			}
+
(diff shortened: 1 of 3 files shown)


lmuntaner:

This is still in Draft, but I want to confirm the approach before I write the tests and spend time improving the code.

peterpeterparker:
The pending transactions won't be fetched in a web worker? If there are, you can compute what you need there without arming the UI.

lmuntaner:
> The pending transactions won't be fetched in a web worker? If there are, you can compute what you need there without arming the UI.

Pending transactions don't need to be pulled regularly. That's why I don't think it makes sense to put them in a web worker.

Even if they are loaded in a web worker, they can still go into the store I merged in #2635 , right?

Then the component will read from this derived store and render the warning message or not.

peterpeterparker:
> Pending transactions don't need to be pulled regularly.

Are you sure? If I go to my Bitcoin wallet XYZ and send BTC to my Oisy Wallet, does this result in a "Pending Transaction"?

lmuntaner:
> > Pending transactions don't need to be pulled regularly.
> 
> Are you sure? If I go to my Bitcoin wallet XYZ and send BTC to my Oisy Wallet, does this result in a "Pending Transaction"?

Good question, I see the confusion. These are "sent" pending transactions. This is the only moment where we would disable sending.

Shall I rename the store "pendingTransactionsStore" to "btcPendingSentTransactionsStore"?

lmuntaner:
> Ah, sorry for the confusion. Maybe yes, let's rename the store even if verbose.

Shall I do it after this PR? In this PR? Or in another PR and then fix the merge conflicts in this PR?

src/frontend/src/btc/derived/has-pending-transactions.derived.ts
```diff
@@ -0,0 +1,66 @@
...
+				return pendingTransactionsData.data.length > 0;
+			}
+
+			if (!$testnets) {
+				if (isNullish($btcAddressMainnet)) {
+					// Return loading while we don't have btc addresses and we can't tell
+					// whether there are pending transactions for a specific address.
+					return 'loading';
+				}
+				if ($btcAddressMainnet !== address) {
```
peterpeterparker, line 34: Not sure if made the choice on purpose so just in case, you can inline/remove the above first check here:

```
if (nonNullish($btcAddressMainnet) && $btcAddressMainnet !== address) {
   return false;
}

return "loading";
```

src/frontend/src/btc/derived/has-pending-transactions.derived.ts
```diff
@@ -0,0 +1,66 @@
...
+					// If the address is not a bitcoin address, there are no pending transactions.
+					return false;
+				}
+
+				// We already checked that the pendingTransaction are present at the top.
+				// If we reach here is because they are not present.
+				return 'loading';
+			}
+			if (
+				isNullish($btcAddressMainnet) ||
```
peterpeterparker, line 44: maybe a dumb idea but, if you test if `address` is actually `nullish` earlier in this function, would you maybe spare some checks?

Just thinking at if there is a way to reduce the numbers of `if`.

peterpeterparker (commented):
Looks good. Just wanted two share two ideas that crossed my mind.

peterpeterparker:
> Shall I do it after this PR? In this PR? Or in another PR and then fix the merge conflicts in this PR?

No strong opinion. Your call.

src/frontend/src/btc/derived/has-pending-transactions.derived.ts
lmuntaner: Done! Thanks!

src/frontend/src/btc/derived/has-pending-transactions.derived.ts
lmuntaner: I have refactored the code using helper variables to explain the ocnditionals. I hope it's more readable now.

I was able to combine different conditions within one variable and still make it readable. Therefore, I removed one if statement.

lmuntaner:
> > Shall I do it after this PR? In this PR? Or in another PR and then fix the merge conflicts in this PR?
> 
> No strong opinion. Your call.

I'll refactor both stores in an upcoming PR after this one is merged.

lmuntaner:

peterpeterparker (approved):
LGTm, thx

PR #2654, opened 2024-10-04
refactor(frontend): move BTC components to folders

DenysKarmazynDFINITY:
# Motivation

We had some BTC component  directly in the general folder. Similar to all other places, I moved them to the proper folders.


peterpeterparker (approved):
LGTM, thx

PR #2655, opened 2024-10-04
chore(frontend): simplify button class removing center sub-class

AntonioVentilii-DFINITY:
# Motivation

The buttons classes `primary`, `secondary` and `tertiary` were using the sub-class `center` in all the instances. Furthermore, it makes sense that, as default, the button have centered content (according to new design too).

This PR removes the sub-class `center` and set the centered content directly in the classes.


peterpeterparker (approved):
LGTm, thx

PR #2662, opened 2024-10-04
feat(frontend): Load Bitcoin pending transactions

lmuntaner:
# Motivation

We want to disable sending bitcoin if the user has a pending transaction.

In this PR, after loading the btc address, a call is made to the backend of Oisy to get the pending transactions for that address.

# Changes

* Add code in `loadBtcAddress` to also load the pending transactions.

# Tests

Tested locally.

### DIFF
--- a/src/frontend/src/lib/services/address.services.ts
+++ b/src/frontend/src/lib/services/address.services.ts
@@ -1,3 +1,4 @@
+import { pendingTransactionsStore } from '$btc/stores/btc-pending-transactions.store';
 import { NETWORK_BITCOIN_ENABLED } from '$env/networks.btc.env';
 import {
 	BTC_MAINNET_TOKEN_ID,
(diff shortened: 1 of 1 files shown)


lmuntaner:

src/frontend/src/lib/services/address.services.ts
```diff
@@ -115,12 +118,33 @@ const loadBtcAddress = async ({
...
+			const address = await getBtcAddress({
 				identity,
-				network: mapToSignerBitcoinNetwork({ network }),
+				network: apiNetwork,
 				nullishIdentityErrorMessage: get(i18n).auth.error.no_internet_identity
-			}),
+			});
+			try {
+				// Endpoint raises an error locally if network is not regtest.
+				if (!LOCAL || network === 'regtest') {
```
peterpeterparker, line 130: I'm confuse why you are "loading pending transactions" within the callback that is, and should in my opinion be solely, use to load an address. Is there any particular reason?

lmuntaner: For different reasons, but I'm open to a refactor.

I want to load the pending transactions after I know the address.

peterpeterparker: There is two different sign-in flows and yes have a pattern for things that need to wait depending of the feature. Let's discuss Monday offline?

lmuntaner: Sounds good.

lmuntaner:
We will do a different approach to load the bitcoin pending sent transactions

PR #2664, opened 2024-10-04
feat(frontend): disable buy button if no OnRamper API key is provided [GIX-2854]

AntonioVentilii-DFINITY:
# Motivation

Just to be on the safe side, we disable the Buy button in case there is no OnRamper API key provided. This could happen only in case there is some inconsistency on the `.env` file.

### DIFF
--- a/src/frontend/src/lib/components/buy/BuyButton.svelte
+++ b/src/frontend/src/lib/components/buy/BuyButton.svelte
@@ -1,11 +1,17 @@
 <script lang="ts">
+	import { ONRAMPER_API_KEY } from '$env/onramper.env';
 	import IconBuy from '$lib/components/icons/IconBuy.svelte';
 	import ButtonHero from '$lib/components/ui/ButtonHero.svelte';
 	import { isBusy } from '$lib/derived/busy.derived';
 	import { i18n } from '$lib/stores/i18n.store';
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
Approving but, I'm hesitant in the sense if that would not be too much work I ask myself if I would not rather like the button to be always enabled and extend the `Buy` component to toast an error if not defined instead of openeing the buy modal. But that's maybe not useful and overengineered, therefore just sharing.

LGTM, thx

AntonioVentilii-DFINITY:
I think that maybe a pop over whenever that button is disabled just saying "temporary disabled, contact blah blah", since the user shouldn't be affected by faults in the build process, I think, no? this would happen only if we have not the correct env file when built

P.S. the best would be a pre-check of compulsory ENV variables (for example onramper prod api key, or ehterescan api key, both), before building

peterpeterparker:
> I think that maybe a pop over whenever that button is disabled just saying "temporary disabled, contact blah blah", 

I personally like to have button "enabled" for those kind of thing, to provide a positive user experience. It's probably an edge case so disabling is good, therefore approving the PR already. Just thought about sharing what crossed my mind.

> P.S. the best would be a pre-check of compulsory ENV variables (for example onramper prod api key, or ehterescan api key, both), before building

Yes!

PR #2665, opened 2024-10-04
refactor(frontend): change Onramper IC address to legacy one [GIX-2854]

AntonioVentilii-DFINITY:
# Motivation

Since it is not specified in the documentation, I asked to Onramper team about the choice of which address to provide for the IC network. After an internal investigation on their side, they said that it is safer to provide the legacy one instead of the principal.


src/frontend/src/lib/components/onramper/OnramperWidget.svelte
```diff
@@ -40,8 +40,8 @@
 			bitcoin: $btcAddressMainnet,
 			ethereum: $ethAddress,
 			erc20: $ethAddress,
-			icrc: $icrcAccountIdentifierText,
-			icp: $icrcAccountIdentifierText
+			icrc: $icpAccountIdentifierText,
```
peterpeterparker, line 43: That cannot work. Old ICP account identity is only for ICP.

AntonioVentilii-DFINITY: yes, i mean, for now on onramper there is only ICP, but i see your point: i moved back the Principal for ICRC

peterpeterparker (approved):
LGTM, thx

PR #2668, opened 2024-10-04
fix(frontend): adjust spacing in Warning and Info

AntonioVentilii-DFINITY:
# Motivation

The instances of `Warning` and `Info` were using `<p>` as children, giving an odd spacing. So, we substitute with `<span>`.

Furthermore, there was an hardcoded string in component InProgressWizard that is solved.

### Before

Screenshot 2024-10-04 at 16 50 34
Screenshot 2024-10-04 at 16 50 46

### After

Screenshot 2024-10-04 at 16 40 22
Screenshot 2024-10-04 at 16 40 59


peterpeterparker (approved):
LGTM, thx

PR #2672, opened 2024-10-04
feat(frontend): BtcSendForm component

DenysKarmazynDFINITY:
# Motivation

Implement `BtcSendForm` component.


src/frontend/src/btc/components/send/BtcSendForm.svelte
```diff
@@ -0,0 +1,45 @@
...
+
+		<!--	TODO: calculate and display transaction fee	-->
+
+		<ButtonGroup slot="toolbar">
+			<button type="button" class="secondary block flex-1" on:click={close}
+				>{$i18n.core.text.cancel}</button
+			>
+			<button class="primary block flex-1" type="submit" disabled={false}>
+				{$i18n.core.text.next}
+			</button>
```
AntonioVentilii-DFINITY, line 42: For these two buttons, I think we already have some components

```svelte
<ButtonCancel on:click={close} />

<Button disabled={false}>
  {$i18n.core.text.next}
</Button>
```

But, for the cancel button, my suggestion is to use a slot

`<slot name="cancel" />`

like the other forms, because sometimes you will give a cancel button and sometimes a close button (see the other forms `IcSendFOrm` and `SendForm`)

AntonioVentilii-DFINITY (commented):
The best would be to create a generic SendForm component with a few slots (I see there are a lot of common elements among all the Send forms), but I will put it in my Backlog, no worries

DenysKarmazynDFINITY:
@AntonioVentilii-DFINITY I actually already implemented a common `SendForm`, not sure why I decided not to proceed with it (probably, wasn't convinced that there is enough shared UI items). In any case, now I agree that it makes sense to have it, so I update the PR, please take a look.

src/frontend/src/btc/components/send/BtcSendForm.svelte
DenysKarmazynDFINITY: Good point, done.

src/frontend/src/lib/components/send/SendForm.svelte
```diff
@@ -0,0 +1,33 @@
...
+		<SendSource token={$token} balance={$balance} {source} />
+
+		<slot name="fee" />
+
+		<ButtonGroup slot="toolbar">
+			<slot name="cancel" />
+
+			<button class="primary block flex-1" type="submit" disabled={false}>
+				{$i18n.core.text.next}
+			</button>
```
AntonioVentilii-DFINITY, line 30: I would use this here (the same for the other forms)

```svelte
<Button disabled={nextDisabled}>
  {$i18n.core.text.next}
</Button>
```

and give `nextDisabled` as prop

src/frontend/src/lib/components/send/SendForm.svelte
```diff
@@ -0,0 +1,33 @@
...
+	const dispatch = createEventDispatcher();
+</script>
+
+<form on:submit={() => dispatch('icNext')} method="POST">
+	<ContentWithToolbar>
+		<slot name="destination" />
+
+		<slot name="amount" />
+
+		<SendSource token={$token} balance={$balance} {source} />
```
AntonioVentilii-DFINITY, line 21: my suggestion is to pass these things as external props, so that we can re-use it.

I agree that at some point these should be "internal". I am working on that with the send context, but still early

src/frontend/src/btc/components/send/BtcSendForm.svelte
```diff
@@ -0,0 +1,37 @@
...
+		bind:invalidDestination
+		{networkId}
+		on:icQRCodeScan
+	/>
+
+	<BtcSendAmount slot="amount" bind:amount bind:amountError />
+
+	<!--	TODO: calculate and display transaction fee	-->
+
+	<button slot="cancel" type="button" class="secondary block flex-1" on:click={close}
```
AntonioVentilii-DFINITY, line 34: I think, better to leave it as slot. This component will be used either as first step or as second step. In those cases, the "cancel" button changes (`close` or `back`)

src/frontend/src/btc/components/send/BtcSendForm.svelte
```diff
@@ -0,0 +1,37 @@
...
+<SendForm {source}>
+	<BtcSendDestination
+		slot="destination"
+		bind:destination
+		bind:invalidDestination
+		{networkId}
+		on:icQRCodeScan
+	/>
+
+	<BtcSendAmount slot="amount" bind:amount bind:amountError />
```
AntonioVentilii-DFINITY, line 30: I don't know the details, but will you add an `insufficientFunds` prop here at some point? or the check will be done differently?

src/frontend/src/lib/components/send/SendForm.svelte
DenysKarmazynDFINITY: Done 👍

src/frontend/src/lib/components/send/SendForm.svelte
DenysKarmazynDFINITY: Agreed, updated.

src/frontend/src/btc/components/send/BtcSendForm.svelte
DenysKarmazynDFINITY: Oh, I thought we were initially talking only about SendForm. Now, I updated it for BtcSendForm too.

AntonioVentilii-DFINITY: shouldn't it be?

`<slot name="cancel" slot="cancel" />`

src/frontend/src/btc/components/send/BtcSendForm.svelte
DenysKarmazynDFINITY: Yeah, the plan is to add `insufficientFunds` when I start working on the related part.

src/frontend/src/btc/components/send/BtcSendForm.svelte
DenysKarmazynDFINITY: Corrected.

AntonioVentilii-DFINITY (approved):
Sorry, forgot to approve! LGTM!

PR #2674, opened 2024-10-04
build(test): import fake-indexeddb

peterpeterparker:
# Motivation

When we start testing features around the addresses that are saved on the client side with Idb, we will need to fake Idb.
So, similarly as in NNS dapp, we can use `fake-indexeddb`

# Changes

- `npm i fake-indexeddb -D`
- Add related import in vitest setup

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,6 +44,7 @@
 				"@types/node": "^20.14.9",
 				"autoprefixer": "^10.4.20",
 				"dotenv": "^16.4.5",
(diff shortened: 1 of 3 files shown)


AntonioVentilii-DFINITY (approved):

AntonioVentilii-DFINITY (approved):
LGTM!

I imagine you will use it in another PR

peterpeterparker:
Yes, exactly.

PR #2673, opened 2024-10-04
test(frontend): init signer allowance before getting addresses

peterpeterparker:
# Motivation

Test for #2632 plus various other test related setup which I'll spit in various PR.

PR #2677, opened 2024-10-04
feat(frontend): expose an imperative identity setter for test

peterpeterparker:
# Motivation

I'm not the fan numero uno of this pattern used in NNS dapp but, I have to admint that it is super handy for test purposes to be able to mock the `identity` in the `auth.store`. Given that I need to mock the identity in #2673, I guess to replicate the pattern.

# Changes

- Add a flag `TEST` in `app.constant`
- Extend `auth.store` with `setForTesting` which set an identity only if `TEST` and `LOCAL` are `true`

### DIFF
--- a/src/frontend/src/lib/constants/app.constants.ts
+++ b/src/frontend/src/lib/constants/app.constants.ts
@@ -9,6 +9,8 @@ export const STAGING = MODE === 'staging';
 export const BETA = MODE === 'beta';
 export const PROD = MODE === 'ic';
 
+export const TEST = import.meta.env.TEST === true;
+
 const MAINNET_DOMAIN = 'icp0.io';
 
 export const REPLICA_HOST = LOCAL ? 'http://localhost:4943/' : 'https://icp-api.io';
--- a/src/frontend/src/lib/constants/app.constants.ts
+++ b/src/frontend/src/lib/constants/app.constants.ts
@@ -9,7 +9,7 @@ export const STAGING = MODE === 'staging';
 export const BETA = MODE === 'beta';
 export const PROD = MODE === 'ic';
 
-export const TEST = import.meta.env.TEST === true;
+export const TEST = JSON.parse(import.meta.env.TEST ?? 'false') === true;
 
 const MAINNET_DOMAIN = 'icp0.io';
 
--- a/src/frontend/src/lib/stores/auth.store.ts
+++ b/src/frontend/src/lib/stores/auth.store.ts
@@ -3,7 +3,6 @@ import {
 	AUTH_POPUP_HEIGHT,
 	AUTH_POPUP_WIDTH,
 	INTERNET_IDENTITY_CANISTER_ID,
-	LOCAL,
 	TEST
 } from '$lib/constants/app.constants';
 import type { OptionIdentity } from '$lib/types/identity';
@@ -92,8 +91,8 @@ const initAuthStore = (): AuthStore => {
 		},
 
 		setForTesting: (identity) => {
-			if (!LOCAL && !TEST) {
-				throw new Error('This function should only be used in local environment');
+			if (!TEST) {
+				throw new Error('This function should only be used in npm run test environment');
 			}
 
 			set({ identity });
--- a/src/frontend/src/lib/stores/auth.store.ts
+++ b/src/frontend/src/lib/stores/auth.store.ts
@@ -2,12 +2,15 @@ import {
 	AUTH_MAX_TIME_TO_LIVE,
 	AUTH_POPUP_HEIGHT,
 	AUTH_POPUP_WIDTH,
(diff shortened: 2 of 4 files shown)


src/frontend/src/lib/constants/app.constants.ts
```diff
@@ -9,6 +9,8 @@ export const STAGING = MODE === 'staging';
 export const BETA = MODE === 'beta';
 export const PROD = MODE === 'ic';
 
+export const TEST = import.meta.env.TEST === true;
```
AntonioVentilii-DFINITY, line 12: Should be added to `.env.test` file?

AntonioVentilii-DFINITY: And, out of curiosity, why not set it up like the other ENV variables? for example

```typescript
export const MODE = VITE_DFX_NETWORK;
export const LOCAL = MODE === 'local';
export const STAGING = MODE === 'staging';
export const BETA = MODE === 'beta';
export const PROD = MODE === 'ic';
```

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -85,6 +89,14 @@ const initAuthStore = (): AuthStore => {
 				...state,
 				identity: null
 			}));
+		},
+
+		setForTesting: (identity) => {
```
AntonioVentilii-DFINITY, line 94: WOWOWOW! Eheh! Now I understand why you are not a fan... hardcode a test-specific function

I have to say, that it is a smart pattern, but not ideal (not a blocker, just an opinion). Maybe add a `TODO` for posterity to think of something  else?

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -85,6 +89,14 @@ const initAuthStore = (): AuthStore => {
...
+			if (!LOCAL && !TEST) {
```
AntonioVentilii-DFINITY, line 95: Honest curiosity, since I am not so expert of `vitest`: why `LOCAL` is allowed? Isn't it just for testing? Doesn't work on local replica when you run `npm run test` ?

src/frontend/src/lib/constants/app.constants.ts
peterpeterparker: MODE = DFX_NETWORK comes from vite, from our own configuration plugin
TEST comes from Nodejs or vitest

src/frontend/src/lib/stores/auth.store.ts
peterpeterparker: In the NNS dapp, we used various solutions, including mocking the entire sign-in process, the `AuthClient` class, the store, etc. Ultimately, this was the final pattern we adopted.

Have a look to the test suite there and let me know if one or the other pattern better suits you. For sure happy to user another one.

src/frontend/src/lib/stores/auth.store.ts
peterpeterparker: Good point. Let's just use TEST.

peterpeterparker: Done.

src/frontend/src/lib/stores/auth.store.ts
AntonioVentilii-DFINITY: Not really any better idea! I would just put something as "documentation", maybe instead of `TODO` tag, use `HACK` tag?

AntonioVentilii-DFINITY (approved):
LGTM!

Just check my comment about a comment tag, if applicable, but nothing blocking

src/frontend/src/lib/stores/auth.store.ts
peterpeterparker: @AntonioVentilii-DFINITY can you please review this commit https://github.com/dfinity/oisy-wallet/pull/2677/commits/ee03926840bdedc1be92e917c665e5ff6b6f874b or the PR again?

- I added the documenation with warning
- I created a test (needed a second commit because I forgot to update a title https://github.com/dfinity/oisy-wallet/pull/2677/commits/cd7333e3b113669985ec8530a56286dc6f4d66a3)
- And actually discovered that the TEST constant was not correctly set

Let me know what you think.

vitest.config.ts
```diff
@@ -60,7 +60,7 @@ export default defineConfig(
 			environment: 'jsdom',
 			globals: true,
 			watch: false,
-			silent: true,
+			silent: false,
```
AntonioVentilii-DFINITY, line 63: I suppose this is necessary for the error string check

AntonioVentilii-DFINITY (approved):
New changes look good

vitest.config.ts
peterpeterparker: Thx! Leftover, reverted but, provided in another PR #2684

PR #2678, opened 2024-10-05
refactor(build): hardcode repetitive env variables

AntonioVentilii-DFINITY:
# Motivation

With the aim of simplifying the .env.* files, we move two variables inside the code directly:

- `VITE_ERC20_ICP_SYMBOL` --> it is a different value only for BETA and PROD (that is `ICP`), while the other environments have always another one (that is `ckICP`)

- `VITE_ERC20_ICP_REPO_URL` --> this is the same for all environments


peterpeterparker (approved):
LGTm, thx

PR #2680, opened 2024-10-05
fix(e2e): avoid false positive for log in tests

AntonioVentilii-DFINITY:
# Motivation

Since the 11th September 2024 10:14, the E2E environment in GitHub is not able to sign in the local replica. This PR does not fix it, but avoids receive a false positive for the "sign-in" tests.

We add verification for the presence of an element that is defined as the Signed In indicator (in our case it will be the menu button).

Then, we include a test that verifies that the system is still signed in after a certain interval: it may happen that some guards in the code will sign-out the user for issues.

# Note

The E2E tests will still fail for this PR, but this time even the sign-in test will fail (as it should).

# Useful Wokflows Links

- [Last working E2E snapshot workflow](https://github.com/dfinity/oisy-wallet/actions/runs/10807703489)
- [First failing E2E snapshot workflow](https://github.com/dfinity/oisy-wallet/actions/runs/10814147938)


peterpeterparker:
I'll let others review this PR, but for what it's worth, in the Oisy Signer Lib, the tests had been flaky for a long time until I figured out that I had to bump the "beforeAll" timeout to 2 minutes. Since then, the E2E tests have been running reliably. Thought I’d share in case it's relevant in this context.

lmuntaner (approved):
LGTM!

As @peterpeterparker mentions, maybe we want to increase the timeout to help witht he flakyness. It can be done in another PR.

AntonioVentilii-DFINITY:
@lmuntaner @peterpeterparker  I actually found the base error: the issue I found is that the e2e test is not logging in... independently of the wait. I have another PR ready for that, but this one I think it is useful either way, because it helped me identifying that it was not logging in

PR #2682, opened 2024-10-06
refactor(e2e): remove Docker from E2E snapshot workflow

AntonioVentilii-DFINITY:
# TL;DR

We remove Docker from the E2E workflow and put back the normal deployment flow, losing ~ one minute and half in processing time.

# Motivation

Given that I am not expert on Docker, what follows is just the simplest/easiest way that I found to solve the issue.

The E2E snapshots and tests workflows were not working properly. It started with a change on how we build the backend wasm, but then a lot of other PRs passed through the code.
After quite a lot of tries with `Dockerfile.e2e` and its related workflows, I gave up, since I wasn't able to find a solution.
I took a step back and removed Docker from the workflows and now it seems to be back to working correctly, but we spend one minute and half more for each workflow (that is the Cargo building process).

# Future

I am not giving up on Docker, but I think we need to re-assess how we use it. Using Docker directly in workflows is useful if we pass something from one job to another in the same workflow (each workflow is a separate machine). In that way, we don't have to re-build for each job.

However, in this specific workflow, we build just once...

The best scenario is to upload the Docker image somewhere and have both the workflow of snapshot and the one for tests to use it, to spare time. I am looking into it (I put a TODO in the Github action too, for posterity). 

# Changes

- Create `deploy-backend` action that sets dfx up, starts it and deploy the backend.
- Remove the Docker buil step from the `update-snapshots` workflow, in favor of the newly-created action above.
- Remove the custom timeout from the test, that now it is not needed anymore.

# Tests

I ran the [E2E snapshots workflow](https://github.com/dfinity/oisy-wallet/actions/runs/11210903200) on this branch and it worked: it updated the snapshots.


AntonioVentilii-DFINITY:
@bitdivine I would like you to have a look too, if possible, since maybe there is some Docker magic that easily solves this in another way

DenysKarmazynDFINITY (approved):
The changes look good to me, thanks!

It would be great to start using `update-snapshots` again, considering the amount of UI work being done at the moment. Later, we can re-evaluate if and how we want to use Docker here.

PR #2683, opened 2024-10-06
refactor(frontend): invert env variables for tokens mainnet

AntonioVentilii-DFINITY:
# Motivation

I would like to propose the inversion of the environment variables referring to enabling tokens mainnets, and use them accordingly:

- `VITE_BITCOIN_MAINNET` --> `VITE_BITCOIN_MAINNET_DISABLED`
- `VITE_ETHEREUM_MAINNET` --> `VITE_ETHEREUM_MAINNET_DISABLED`

My reasons:

1. They are by default set to `true` everywhere, apart from the test env file.
2. It makes sense that the default should be that the mainnets are enabled, BUT we leave the optionality to disable them for dev/testing purposes.
3. When we are using Docker (or in general a Virtual environment, like GitHub Workflows), we reduce the number of env variables needed.
4. For BITCOIN, we have the feature flag anyway.


AntonioVentilii-DFINITY:
@lmuntaner and @DenysKarmazynDFINITY please have a look at this too

src/frontend/src/env/networks.eth.env.ts
```diff
@@ -2,7 +2,7 @@ import type { Networkish } from '@ethersproject/networks';
 import { Network } from 'alchemy-sdk';
 
 export const ETH_MAINNET_ENABLED =
-	JSON.parse(import.meta.env.VITE_ETHEREUM_MAINNET ?? false) === true;
+	JSON.parse(import.meta.env.VITE_ETHEREUM_MAINNET_DISABLED ?? true) === false;
```
peterpeterparker, line 5: With this change of approach, shouldn't we interpret the abscence of the environment flag as the feature being enabled?

I mean default to `false`:

```
JSON.parse(VITE_ETHEREUM_MAINNET_DISABLED ?? false) === false;
```

If flag DISABLE not set => feature is enabled
If flag DISABLE is set to false => feature is enabled
If flag DISABLE set to true => feature is disabled

WDYT?

AntonioVentilii-DFINITY: you are totally right! i confused the logic

peterpeterparker (approved):
LGTM, thanks

You take care of updating various environments for deployment?

lmuntaner:
Isn't it dangerous that the default is true? Won't that mean that if we deploy to prod and there is a bug in the build new functoinality might be deployed?

lmuntaner:
It might also disrupt local environments. Any new variable will need to be added to everyone else as soon as it's merged.

Is the only advantage that we need less env vars for Docker?

DenysKarmazynDFINITY (approved):
If it helps to avoid some docker/GH actions overhead, then it's fine with me. The code also LGTM.

AntonioVentilii-DFINITY:
@lmuntaner @DenysKarmazynDFINITY that env is not the feature-flag, but the mainnet enabler. I checked, it is already `true`, in all environment files. That is why it seems a bit pointless to have it like that, and I think it even makes sense that it is always `true` by default, independently if the Bitcoin features are enabled or not (that is the real bug-stop).

lmuntaner:
> @lmuntaner @DenysKarmazynDFINITY that env is not the feature-flag, but the mainnet enabler. I checked, it is already `true`, in all environment files. That is why it seems a bit pointless to have it like that, and I think it even makes sense that it is always `true` by default, independently if the Bitcoin features are enabled or not (that is the real bug-stop).

Ah, ok. True, this is not the feature flag, but something else. LGTM!

lmuntaner (approved):

PR #2687, opened 2024-10-07
chore(frontend): Remove Unused Svelte Files

sa-github-api:
The following files were removed because they were not used in the project anymore:
```
src/frontend/src/btc/components/send/BtcSendAmount.svelte
src/frontend/src/btc/components/send/BtcSendDestination.svelte
```


peterpeterparker:
Those two components are under development.

PR #2700, opened 2024-10-07
ci(frontend): workflow to auto-deploy frontend in staging

AntonioVentilii-DFINITY:
# Motivation

# Changes

# Tests

### DIFF
--- a/.github/workflows/frontend-deploy-staging.yml
+++ b/.github/workflows/frontend-deploy-staging.yml
@@ -1,0 +1,36 @@
+name: Deploy Frontend to Staging
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  deploy-staging:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Prepare
+        uses: ./.github/actions/prepare
+
+      - name: Set up DFX
+        uses: dfinity/setup-dfx@main
+
+      - name: Install key
+        env:
+          DFX_DEPLOY_STAGING_KEY: ${{ secrets.DFX_DEPLOY_STAGING_KEY }}
+        run: |
+          key_pem=$(mktemp)
+          printenv "DFX_DEPLOY_STAGING_KEY" > "$key_pem"
+          dfx identity import --disable-encryption --force default "$key_pem"
+          rm "$key_pem"
+
+      - name: Pre-build
+        run: npm run build
+
+      - name: Deploy to Staging
+        run: |
+          ENV=staging dfx deploy frontend --network staging


AntonioVentilii-DFINITY:
@inc-man will take over this PR

Have a look at @lmuntaner 's suggestion: they are doing something similar here:

- https://github.com/dfinity/internet-identity/blob/1a185880684d646c1ae301ffd7e4152ae8a80d37/.github/workflows/deploy-rc.yml#L63

.github/workflows/frontend-deploy-staging.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Set up DFX
+        uses: dfinity/setup-dfx@main
+
+      - name: Install key
+        env:
+          DFX_DEPLOY_STAGING_KEY: ${{ secrets.DFX_DEPLOY_STAGING_KEY }}
+        run: |
+          key_pem=$(mktemp)
+          printenv "DFX_DEPLOY_STAGING_KEY" > "$key_pem"
+          dfx identity import --disable-encryption --force default "$key_pem"
```
bitdivine, line 53:
```suggestion
          dfx identity import --storage-mode=plaintext --force default <(printenv DFX_DEPLOY_STAGING_KEY)
```
* `--disable-encryption` has been deprecated and replaced with `--storage-mode=plaintext`
* If you like (not obligatory) you can use a herefile as shown.  Normally when dealing with key material one avoids letting it touch disk, but in this case the key material is landing in plaintext on disk anyway so it's a moot point.

.github/workflows/frontend-deploy-staging.yml
```diff
@@ -0,0 +1,62 @@
...
+          rm "$key_pem"
+
+      - name: Pre-build
+        run: npm run build
+
+      - name: Deploy to Staging
+        env:
+          ENV: staging
+        run: dfx deploy frontend --network staging
```
bitdivine, line 62: I think we probably want to deploy the backend as well, however we can do that in another PR.  That could be in this or another workflow.  [Here is an example of a workflow that deploys](https://github.com/dfinity/nns-dapp/blob/main/.github/workflows/deploy-to-app.yaml) another project with 2 canisters to a test environment.  It has a checklist for which canisters to deploy:
```
  workflow_dispatch:
    inputs:
      mode:
        type: choice
        description: "The dfx canister install mode.  See `dfx canister install --help` for details."
        options:
          - upgrade
          - reinstall
          - install
          - auto
      canisters:
        type: choice
        description: Which canisters to install
        options:
          - all
          - nns-dapp
          - sns_aggregator
```

.github/workflows/frontend-deploy-staging.yml
```diff
@@ -0,0 +1,62 @@
+name: Deploy Frontend to Staging
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
```
bitdivine, line 7: For workflow dispatch it would be nice to be able to choose the target network, so that this can also be used to deploy to the mainnet test canisters test_fe_1..4  But this can be done in another PR.

.github/workflows/frontend-deploy-staging.yml
bitdivine: I see that Internet Identity still uses `--disable-encryption`.  We should do a scan of all our repositories and update them all.  I'll add a Jira ticket to do that.

.github/workflows/frontend-deploy-staging.yml
AntonioVentilii-DFINITY: Indeed, I would first test if this one is all ok (deploying frontend takes less time, so less testing time)

.github/workflows/frontend-deploy-staging.yml
AntonioVentilii-DFINITY: Good one!

bitdivine:
I think it would be fine to go ahead and merge this.  What do you think?

PR #2707, opened 2024-10-07
feat(frontend): Add mapNetworkIdToBitcoinNetwork util

lmuntaner:
# Motivation

In the components I'll have a network id, but other levels need the `BitcoinNetwork`.

In this PR, I add an util to convert a network id into a `BitcoinNetwork` if it's one of the known bitcoin network ids.

# Changes

* New util `mapNetworkIdToBitcoinNetwork`.

# Tests

* Test new util `mapNetworkIdToBitcoinNetwork`.

### DIFF
--- a/src/frontend/src/lib/utils/network.utils.ts
+++ b/src/frontend/src/lib/utils/network.utils.ts
@@ -33,6 +33,15 @@ export const isNetworkIdBTCTestnet = (networkId: NetworkId | undefined): boolean
 export const isNetworkIdBTCRegtest = (networkId: NetworkId | undefined): boolean =>
 	BTC_REGTEST_NETWORK_ID === networkId;
 
+const mapper: Record<symbol, BitcoinNetwork> = {
+	[BTC_MAINNET_NETWORK_ID]: 'mainnet',
+	[BTC_TESTNET_NETWORK_ID]: 'testnet',
+	[BTC_REGTEST_NETWORK_ID]: 'regtest'
+};
+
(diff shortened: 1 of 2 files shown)


lmuntaner:

@DenysKarmazynDFINITY just FYI, this might be an interesting util for you too.

src/frontend/src/lib/utils/network.utils.ts
```diff
@@ -33,6 +33,15 @@ export const isNetworkIdBTCTestnet = (networkId: NetworkId | undefined): boolean
 export const isNetworkIdBTCRegtest = (networkId: NetworkId | undefined): boolean =>
 	BTC_REGTEST_NETWORK_ID === networkId;
 
+const mapper: Record<symbol, BitcoinNetwork> = {
+	[BTC_MAINNET_NETWORK_ID]: 'mainnet',
+	[BTC_TESTNET_NETWORK_ID]: 'testnet',
+	[BTC_REGTEST_NETWORK_ID]: 'regtest'
+};
+
+export const mapNetworkIdToBitcoinNetwork = (networkId: NetworkId): BitcoinNetwork | undefined =>
```
AntonioVentilii-DFINITY, line 42: can it happen that it is `undefined`? I mean, would it give some error if it's not the existing Keys? I sincerly don't know, but if that's the case, maybe create a test for this specific case?

AntonioVentilii-DFINITY (approved):
LGTM!

Small comment, not blocker

src/frontend/src/lib/utils/network.utils.ts
lmuntaner: Done!

PR #2708, opened 2024-10-07
refactor(frontend): reduce size of preview image

AntonioVentilii-DFINITY:
# Motivation

The Design team was able to reduce the size of the image in the preview of the landing page. Furthermore, we will have only one image for both desktop and mobile.

NOTE: I ran the Update Snapshot workflow on this PR.


src/frontend/src/lib/components/auth/LandingPage.svelte
```diff
@@ -21,9 +20,8 @@
 	<div
 		class="ml-auto min-w-[1127px] pt-12 md:m-0 md:flex md:h-full md:content-center md:items-center"
 	>
-		<picture aria-label={ariaLabel} class="w-full md:h-md:mt-auto" role="presentation">
-			<source srcset={previewSmall} media="(max-width: 767px)" />
-			<Img src={previewLarge} alt={ariaLabel} />
-		</picture>
+		<div class="w-full md:h-md:mt-auto" role="presentation">
```
peterpeterparker, line 23: The `role=presentation` can be remove on the div. It should be provided on the `img`

AntonioVentilii-DFINITY: ah right! thank you! it is the default for that component, so i just removed it

peterpeterparker (approved):
LGTM, thx

PR #2709, opened 2024-10-07
feat(frontend): Service to load btc pending sent txs

lmuntaner:
# Motivation

We want to load the btc pending sent transactions when the user visits the form to send tokens.

In this PR, I introduce the service that will be used by the component to load it.

# Changes

* New service `loadBtcPendingSentTransactions`.

# Tests

Tested locally by loading them in the IcSendForm

### DIFF
--- a/src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
+++ b/src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
@@ -18,7 +18,10 @@ export const loadBtcPendingSentTransactions = async ({
 	try {
 		const network = mapNetworkIdToBitcoinNetwork(networkId);
 		if (isNullish(network)) {
-			throw new Error(`Invalid networkId: ${networkId.toString}`);
+			return {
+				success: false,
+				err: new Error(`Invalid networkId: ${networkId.toString}`)
+			};
 		}
 		const pendingTransactions = await getPendingBtcTransactions({
 			identity,
--- a/src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
+++ b/src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
@@ -1,8 +1,10 @@
 import { btcPendingSentTransactionsStore } from '$btc/stores/btc-pending-sent-transactions.store';
 import { getPendingBtcTransactions } from '$lib/api/backend.api';
 import type { NetworkId } from '$lib/types/network';
+import type { ResultSuccess } from '$lib/types/utils';
 import { mapNetworkIdToBitcoinNetwork, mapToSignerBitcoinNetwork } from '$lib/utils/network.utils';
 import type { Identity } from '@dfinity/agent';
+import { isNullish } from '@dfinity/utils';
 
 export const loadBtcPendingSentTransactions = async ({
 	address,
@@ -12,10 +14,10 @@ export const loadBtcPendingSentTransactions = async ({
 	address: string;
 	identity: Identity;
 	networkId: NetworkId;
-}) => {
+}): Promise<ResultSuccess> => {
 	try {
 		const network = mapNetworkIdToBitcoinNetwork(networkId);
-		if (!network) {
+		if (isNullish(network)) {
 			throw new Error(`Invalid networkId: ${networkId.toString}`);
 		}
 		const pendingTransactions = await getPendingBtcTransactions({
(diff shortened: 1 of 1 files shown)


lmuntaner:

src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
```diff
@@ -0,0 +1,34 @@
+import { btcPendingSentTransactionsStore } from '$btc/stores/btc-pending-sent-transactions.store';
+import { getPendingBtcTransactions } from '$lib/api/backend.api';
+import type { NetworkId } from '$lib/types/network';
+import { mapNetworkIdToBitcoinNetwork, mapToSignerBitcoinNetwork } from '$lib/utils/network.utils';
+import type { Identity } from '@dfinity/agent';
+
+export const loadBtcPendingSentTransactions = async ({
```
peterpeterparker, line 7: For consistency can you return a `ResultSuccess`?

src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
```diff
@@ -0,0 +1,34 @@
...
+	identity,
+	networkId
+}: {
+	address: string;
+	identity: Identity;
+	networkId: NetworkId;
+}) => {
+	try {
+		const network = mapNetworkIdToBitcoinNetwork(networkId);
+		if (!network) {
```
peterpeterparker, line 18: `isNullish`?

src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
```diff
@@ -0,0 +1,34 @@
...
+			throw new Error(`Invalid networkId: ${networkId.toString}`);
```
peterpeterparker, line 19: - should this service toast an error?

- if you use the `ResultSuccess` pattern I guess this can become a return?

lmuntaner: Not toast an error (at least for now). The component will show a different error because the store is populated with the error.

src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
lmuntaner: Done!

src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
lmuntaner: Done!

lmuntaner:

src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
peterpeterparker: Regarding returning the error vs throwing it?

lmuntaner: I was assuming it was ok that this was thrown, got to the catch and return the error.

But I guess I can return here as well.

lmuntaner:

src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
peterpeterparker: It's ok but, if you also `return` instead of `throw`, then you spare the need of a `try catch` in your consumer component - i.e. you will be able to interpret solely the result.

peterpeterparker (approved):
LGTM, thx

PS.: @AntonioVentilii-DFINITY @DenysKarmazynDFINITY maybe we should discuss if we don't want to start writing tests for `services` since we started writing tests for `api`.

In that case, a test like "if backend call success, store gets updated with x" or `if backend fails, store is resetted" is maybe interesting? Particularly the latter which is a bit tricky to test with e2e? 

Unrelated to the PR itself again, just thinking out at loud.

lmuntaner:
> LGTM, thx
> 
> PS.: @AntonioVentilii-DFINITY @DenysKarmazynDFINITY maybe we should discuss if we don't want to start writing tests for `services` since we started writing tests for `api`.
> 
> In that case, a test like "if backend call success, store gets updated with x" or `if backend fails, store is resetted" is maybe interesting? Particularly the latter which is a bit tricky to test with e2e?
> 
> Unrelated to the PR itself again, just thinking out at loud.

I agree with this. E2E for happy paths is great, but it's harder to get the fine-grained scenarios.

I can add a test in a following PR.

PR #2714, opened 2024-10-07
feat(frontend): SendReview component

DenysKarmazynDFINITY:
# Motivation

Implement a shared `SendReview` component.

### DIFF
--- a/src/frontend/src/lib/components/send/SendReview.svelte
+++ b/src/frontend/src/lib/components/send/SendReview.svelte
@@ -1,28 +1,27 @@
 <script lang="ts">
 	import { nonNullish } from '@dfinity/utils';
-	import { createEventDispatcher } from 'svelte';
+	import { createEventDispatcher, getContext } from 'svelte';
+	import { SEND_CONTEXT_KEY, type SendContext } from '$icp-eth/stores/send.store';
 	import SendData from '$lib/components/send/SendData.svelte';
 	import Button from '$lib/components/ui/Button.svelte';
 	import ButtonBack from '$lib/components/ui/ButtonBack.svelte';
 	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
 	import ContentWithToolbar from '$lib/components/ui/ContentWithToolbar.svelte';
 	import { i18n } from '$lib/stores/i18n.store';
-	import type { OptionBalance } from '$lib/types/balance';
-	import type { OptionToken } from '$lib/types/token';
 
 	export let destination = '';
 	export let amount: number | undefined = undefined;
 	export let source: string;
-	export let token: OptionToken;
-	export let balance: OptionBalance;
 	export let disabled: boolean | undefined = false;
 
 	const dispatch = createEventDispatcher();
+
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/send/SendReview.svelte
```diff
@@ -0,0 +1,38 @@
...
+	import SendData from '$lib/components/send/SendData.svelte';
+	import Button from '$lib/components/ui/Button.svelte';
+	import ButtonBack from '$lib/components/ui/ButtonBack.svelte';
+	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
+	import ContentWithToolbar from '$lib/components/ui/ContentWithToolbar.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import type { OptionBalance } from '$lib/types/balance';
+	import type { OptionToken } from '$lib/types/token';
+
+	export let destination = '';
```
peterpeterparker, line 13: Why do you need all those parameters? We use a `SendContext` in modals, that pattern won't work here?

DenysKarmazynDFINITY: Only one of these props is coming from the context (token). This is how they look in `IcSendReview` in `main`:

Screenshot 2024-10-07 at 16 44 53

Also, since the component I implemented in this PR is placed in `$lib`, I assume we don't want to use a context from `$icp-eth` or other network folder and rather make it generic.

DenysKarmazynDFINITY: @peterpeterparker @AntonioVentilii-DFINITY would it be fine with you if we introduce the send store for BTC after we have all related components in place? I am implementing the flow from bottom to top (first child components, then parent ones), so without having the whole components tree in place, I find it a bit problematic to introduce and use a new store. Plus, I want to unblock @lmuntaner as he needs btc send modals to update them with pending transactions stuff.

peterpeterparker: When it comes to me, I would say: no. I think it's best if we get it right from the start because we already suffered having to rework the modals multiple times and I know it's quite an effort which can also lead to issues.

peterpeterparker: I can give a hand later today if you wish btw.

DenysKarmazynDFINITY: @peterpeterparker that would be great! My main questions are whether it's possible to setup such a store at this stage (where BtcSendTokenWizard doesn't exist yet) and do we want to move the related store files from `icp-eth` to `lib`?

DenysKarmazynDFINITY: I guess I can start using context, even if doesn't exist yet (it will be added later in `BtcSendModal`) since the components I'm adding are not in-use yet. 

And for now, I can re-use store from `icp-eth` and we can decide later where do we want to move.

DenysKarmazynDFINITY: I updated the component, please take a look.

peterpeterparker (approved):
LGTM, thx

PR #2717, opened 2024-10-07
fix(frontend): util formatToken was removing trailing zeros

AntonioVentilii-DFINITY:
# Motivation

The util formatToken was not formatting correctly the trailing zeros when the user wants to show them. We adjust it a bit and create tests to verify the effectiveness.

# Changes

# Tests


peterpeterparker (approved):
Rubber stamp.

I bet it was that way on purpose but, I generally dislike the topic "decimals" at the foundation.

Code wise look ok, thanks

PR #2718, opened 2024-10-07
fix(frontend): hardcode background color to avoid glitch [GIX-3067]

AntonioVentilii-DFINITY:
# Motivation

We need to hard-code the background color, since it causes a glitch at loading time (that is until the variable is loaded).


peterpeterparker (approved):
LGTM, thx

PR #2720, opened 2024-10-07
style(frontend): change network switcher menu style [GIX-3086]

AntonioVentilii-DFINITY:
# Motivation

We change the style of the dropdown for the Network selection. To do that, I create a new `Dropdown` component to be re-used.

It is basically a refactoring of the previous `NetworksSwitcher`, but breaking it down into smaller parts, to make it easier to change the style eventually. We use Tailwind for the entire PR (note that there is a repeated `blue-ribbon` color because the aim is to remove the RGB one soon enough).

# Changes

- Create `DropdownButton` element with its specific style for each state.
- Create `Dropdown` with `DropdownButton` and `items` slot.
- Wrap list of networks with `DropDown`.
- Remove Popover backdrop.
- Remove deprecated component.

# Tests

https://github.com/user-attachments/assets/33b50824-8168-4efd-ad00-3188086e0c0c

https://github.com/user-attachments/assets/cb027cd9-01ef-440c-8c86-544ce7ac6851


src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
AntonioVentilii-DFINITY, line 1: I don't know why Github shows the changes like that, but I basically only wrapped the entire list around the new `Dropdown` component...

src/frontend/src/lib/components/ui/DropdownButton.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
+	import { IconExpandMore } from '@dfinity/gix-components';
+
+	export let disabled = false;
+	export let button: HTMLButtonElement | undefined = undefined;
+	export let ariaLabel: string;
+	export let opened = false;
```
peterpeterparker, line 7: Instead of `opened` what do you think of the name `focused`?

src/frontend/src/lib/components/ui/DropdownButton.svelte
```diff
@@ -0,0 +1,28 @@
...
+</script>
+
+<button
+	class="text-secondary min-w-72 justify-between gap-2 rounded-xl border border-light-grey bg-white px-4 py-3 text-left font-medium leading-5 text-inherit hover:border-blue-ribbon"
+	bind:this={button}
+	on:click
+	aria-label={ariaLabel}
+	{disabled}
+	class:opacity-50={disabled}
+	class:border={opened}
```
peterpeterparker, line 17: You need this, the button has already a border in the class?

src/frontend/src/lib/components/ui/DropdownButton.svelte
```diff
@@ -0,0 +1,28 @@
...
+	class:border-blue-ribbon={opened}
+>
+	<slot />
+	<div
+		class="transform transition-transform duration-300 ease-in-out"
+		class:scale-y-[-1]={opened}
```
peterpeterparker, line 23: Out of curiosity, what does `scale-y-` do? I don't think we have it yet and I don't know it 😅.

tailwind.config.js
```diff
@@ -17,9 +17,11 @@ export default {
 			'off-white': '#fcfaf6',
 			dust: '#dbd9d6',
 			grey: '#c0bbc4',
+			'light-grey': '#ced4da',
```
peterpeterparker, line 20: This change is expected and if yes, isn't the corresponding variables missing in `colors.scss`?

src/frontend/src/lib/components/ui/DropdownButton.svelte
AntonioVentilii-DFINITY: this is the one that make the arrow icon "flip": instead of rotate, I think it was better to make it flip when the dropdown is open, WDYT?

tailwind.config.js
AntonioVentilii-DFINITY: Yes, it is one of the color of the pallet, but I didn't want to add it as variable color, since my plan is to slowly deprecate the CSS styles. In this way, the color is available only for Tailwind classes.

Of course, if at some point it is absolutely necessary to have it for CSS, I will add it

peterpeterparker: I would add it to the class because we gonna start having some variables set and some not.
Not against deprecating the style class but either we do it or not if you know what I mean?

src/frontend/src/lib/components/ui/DropdownButton.svelte
peterpeterparker: Ah now I get it. Looks cool 👍

src/frontend/src/lib/components/ui/DropdownButton.svelte
AntonioVentilii-DFINITY: right! removed

src/frontend/src/lib/components/ui/DropdownButton.svelte
AntonioVentilii-DFINITY: but you are right, it is not needed when not opened, and it can be used `-scale-y-100` that is the built-in

src/frontend/src/lib/components/ui/DropdownButton.svelte
AntonioVentilii-DFINITY: the idea was to really be the indicator that the dropdown is open/visible, in fact I am using it to flip the icon too

in the sense, it could be focused but not opened. Maybe a better name is `dropdownOpen`? or `clicked`? boh

peterpeterparker: No no it's fine, `opened` works, thanks.

tailwind.config.js
AntonioVentilii-DFINITY: Consistency, make sense!

peterpeterparker (approved):
LGTM, thx

PR #2721, opened 2024-10-08
feat(frontend): Load btc pending sent txs in BtcSendForm

lmuntaner:
# Motivation

We want to disable the review step if the user has pending sent transactions. Therefore, we load them in the form step so that they will be present in the review step.

# Changes

* Use `loadBtcPendingSentTransactions` in `BtcSendForm`.

# Tests

Tested manually that it does indeed load the data in the store.

### DIFF
--- a/src/frontend/src/btc/components/send/BtcSendForm.svelte
+++ b/src/frontend/src/btc/components/send/BtcSendForm.svelte
@@ -6,6 +6,9 @@
 	import { balance } from '$lib/derived/balances.derived';
 	import { token } from '$lib/stores/token.store';
 	import type { NetworkId } from '$lib/types/network';
+	import { authIdentity } from '$lib/derived/auth.derived';
+	import { nonNullish } from '@dfinity/utils';
+	import { loadBtcPendingSentTransactions } from '$btc/services/btc-pending-sent-transactions.services';
 
 	export let networkId: NetworkId | undefined = undefined;
 	export let amount: number | undefined = undefined;
@@ -14,6 +17,16 @@
 
 	let amountError: BtcAmountAssertionError | undefined;
 	let invalidDestination: boolean;
+
+	$: {
+		if (nonNullish($authIdentity) && nonNullish(networkId)) {
+			loadBtcPendingSentTransactions({
+				identity: $authIdentity,
+				networkId,
+				address: source
+			});
+		}
+	}
 </script>
 
 <SendForm {source} token={$token} balance={$balance}>
--- a/src/frontend/src/btc/components/send/BtcSendForm.svelte
+++ b/src/frontend/src/btc/components/send/BtcSendForm.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { nonNullish } from '@dfinity/utils';
+	import { onMount } from 'svelte';
 	import BtcSendAmount from './BtcSendAmount.svelte';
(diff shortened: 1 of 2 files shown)


lmuntaner:
@peterpeterparker as discussed offline, here is the last PR to load the btc pending sent transactions in the form step.

@DenysKarmazynDFINITY FYI, in case you are working in this compoent.

src/frontend/src/btc/components/send/BtcSendForm.svelte
```diff
@@ -14,6 +17,16 @@
 
 	let amountError: BtcAmountAssertionError | undefined;
 	let invalidDestination: boolean;
+
+	$: {
+		if (nonNullish($authIdentity) && nonNullish(networkId)) {
```
peterpeterparker, line 22: - Can we initiate the load in the `onMount` hook? 
- Should we do anything if `$authIdentity` is not defined?
- Can you add a comment that we on purpose does not await the response?

lmuntaner: > * Can we initiate the load in the `onMount` hook?  Or `networkId` can be manually selected by the user?

Not network id, but also the source. I guess that not at the moment because we don't support subaccounts, but if we support subaccounts in the future is better if this is triggered when `source` changes.

Does that make sense? I can put it in the `onMount` as well, not a strong opinion. Just sharing why I did so. I also put it in `onMount` first but then I thought of changing the `source`.

> * Should we do anything if `$authIdentity` is not defined?

This is just an edge case. An anonymous user can't reach this component. If they do, we also don't want to trigger the call. But I don't see a reason to do anything special either.

> * Can you add a comment that we on purpose does not await the response?

Added!

peterpeterparker: > Not network id...

Fair point. On the other side, if we support subaccounts, then we might need some debouncing as we might not want to trigger lots of call update if the users repeatdly switch between accounts, so we would have to comeback to this too.

For those reason, kind of feel like `onMount` would be nice for now as it would express the fact that we indeed only load it once.

> This is just an edge case

Given that we ignore the error anyway, maybe the service then can accept an `OptionIdentity` and return `{success: false}` if not set that way you don't have to handle the edge case in the component? If agree, we can add a TODO.

peterpeterparker (commented):
LGTM, thanks

Checkout my last answers before merge.

src/frontend/src/btc/components/send/BtcSendForm.svelte
lmuntaner: Done!

I also allowed `networkId` to be `undefined` in the service and I handled the scenario there.

lmuntaner:

src/frontend/src/btc/services/btc-pending-sent-transactions.services.ts
```diff
@@ -1,26 +1,34 @@
...
-	networkId: NetworkId;
+	identity: OptionIdentity;
+	networkId?: NetworkId;
 }): Promise<ResultSuccess> => {
 	try {
-		const network = mapNetworkIdToBitcoinNetwork(networkId);
+		if (isNullish(identity)) {
+			return {
+				success: false,
+				err: new Error('Identity not found')
```
peterpeterparker, line 22: You can use `get(i18n).auth.error.no_internet_identity` here for example.

peterpeterparker (approved):
LGTM, thx

PR #2726, opened 2024-10-08
chore(frontend): use Oisy placeholders for Oisy name

AntonioVentilii-DFINITY:
# Motivation

Some string in i18n file were using hard-coded Oisy name. We substitute with the placeholder.


peterpeterparker (approved):
LGTM, thanks

PR #2729, opened 2024-10-08
feat(ci): Change update-snapshots to create PR requests too

AntonioVentilii-DFINITY:
# Motivation

We would like to auto-update E2E snapshots every new commit of main.


DenysKarmazynDFINITY (approved):
Nice one, let's test it!

PR #2731, opened 2024-10-08
style(frontend): new ButtonAsLink for the footer [GIX-3049]

AntonioVentilii-DFINITY:
# Motivation

We want to change the style of the icons in the footer. That is actually a version of the `tertiary` button. So we adapt the code to use it.

# Changes

- Create a type for the button styles and use it in the button components.
- Change the style of `tertiary` button class to have a generic style and only specific after link.
- Create `ButtonAsLink` component that is basically the same as `Button` but with <a>.
- Use `ButtonAsLink` in the `Footer`.
- Unrelated: we adjust the footer spacing.

# Tests

Only changed the footer. The rest is the same.

Screenshot 2024-10-08 at 13 31 19


src/frontend/src/lib/components/core/Footer.svelte
```diff
@@ -13,31 +14,34 @@
...
 	<div class="flex flex-row items-center gap-4">
 		<a
 			href={OISY_REPO_URL}
 			rel="external noopener noreferrer"
 			target="_blank"
-			class="rounded-full bg-white p-2"
 			aria-label={$i18n.navigation.text.source_code_on_github}
 		>
-			<IconGitHub />
+			<ButtonIcon ariaLabel={$i18n.navigation.text.source_code_on_github}>
```
peterpeterparker, line 29: Mmmh a `button` within a `a`. Is that semantically speaking correct?

src/frontend/src/lib/components/core/Footer.svelte
```diff
@@ -13,31 +14,34 @@
...
+				{$i18n.navigation.text.source_code_on_github}
+			</ButtonIcon>
 		</a>
 
 		<a
 			href={OISY_TWITTER_URL}
 			rel="external noopener noreferrer"
 			target="_blank"
-			class="rounded-full bg-white p-2"
 			aria-label={$i18n.navigation.text.source_code_on_github}
```
peterpeterparker, line 39: GitHub? Shouldn't be Twitter?

src/frontend/src/lib/components/core/Footer.svelte
```diff
@@ -13,31 +14,34 @@
...
 		>
-			<IconTwitter />
+			<ButtonIcon>
+				<IconTwitter slot="icon" />
+				{$i18n.navigation.text.source_code_on_github}
```
peterpeterparker, line 43: Same same

src/frontend/src/lib/components/core/Footer.svelte
```diff
@@ -13,31 +14,34 @@
...
 
 		<a
 			href={OISY_TWITTER_URL}
 			rel="external noopener noreferrer"
 			target="_blank"
-			class="rounded-full bg-white p-2"
 			aria-label={$i18n.navigation.text.source_code_on_github}
 		>
-			<IconTwitter />
+			<ButtonIcon>
```
peterpeterparker, line 41: Same same

AntonioVentilii-DFINITY: As above

src/frontend/src/lib/components/core/Footer.svelte
AntonioVentilii-DFINITY: I was in doubt a bit about this, but we will have some button styles that will be applied all over, even to link <a> elements.

So, I am not sure how to tackle this. The other solutions is to include such classes for both <button> and <a>, but I don't really like the idea somehow

peterpeterparker: I should be a link. I think that from an accessibility and semantic point of view, using a button here, moreover stricly for styling purpose, is incorrect.

AntonioVentilii-DFINITY: Right, check last commit please: I created a new component for that, WDYT?

src/frontend/src/lib/components/ui/ButtonAsLink.svelte
```diff
@@ -0,0 +1,18 @@
+<script lang="ts">
```
peterpeterparker, line 1: I suggest to rename this component to `ExternalLinkIcon`.

src/frontend/src/lib/components/ui/ButtonIcon.svelte
```diff
@@ -1,11 +1,14 @@
 <script lang="ts">
+	import type { ButtonColorStyle } from '$lib/types/style';
+
 	export let button: HTMLButtonElement | undefined = undefined;
+	export let colorStyle: ButtonColorStyle = 'tertiary';
 	export let testId: string | undefined = undefined;
-	export let ariaLabel: string;
+	export let ariaLabel: string | undefined = undefined;
```
peterpeterparker, line 7: Do we need this change?

src/frontend/src/lib/i18n/en.json
```diff
@@ -41,7 +41,8 @@
 			"more_settings": "More settings",
 			"menu": "Your wallet address, settings, sign-out and external links",
 			"changelog": "Open the changelog of $oisy_name on GitHub to review the latest updates",
-			"submit_ticket": "Open the ticket submission form on GitHub to report an issue or request a feature"
+			"submit_ticket": "Open the ticket submission form on GitHub to report an issue or request a feature",
```
peterpeterparker, line 44: Can be simplified: "Report an issue or request a feature on GitHub"

src/frontend/src/lib/i18n/en.json
```diff
@@ -41,7 +41,8 @@
...
+			"open_twitter": "Open the X (Twitter) account of $oisy_name"
```
peterpeterparker, line 45: nitpick: Oisy does not have a Twitter account per sé, the foundation has. "Open the DFINITY X/Twitter feed"

src/frontend/src/lib/components/ui/ButtonIcon.svelte
AntonioVentilii-DFINITY: forgot to remove, removed

src/frontend/src/lib/i18n/en.json
AntonioVentilii-DFINITY: changed

src/frontend/src/lib/i18n/en.json
AntonioVentilii-DFINITY: right!

src/frontend/src/lib/components/ui/ButtonAsLink.svelte
AntonioVentilii-DFINITY: Yes! Make sense, changed

peterpeterparker (approved):
LGTM, thx

PR #2735, opened 2024-10-08
chore(devops): Add two more test envs

bitdivine:
# Motivation
Test environments are in demand right now and we don't have enough to go around.

# Changes
- Add two more test environments.

# Tests
The frontend has been deployed to these new canisters:
* test_fe_3: https://dxvmu-wiaaa-aaaah-aqzaa-cai.icp0.io/
* test_fe_4: https://gwqec-uqaaa-aaaak-qlnza-cai.icp0.io/
(deployed without env files, so the configuration is probably all wrong)


AntonioVentilii-DFINITY (approved):
Thank you!!

PR #2737, opened 2024-10-08
feat(frontend): consider the test environment the same as Staging

AntonioVentilii-DFINITY:
# Motivation

The test environment will be treated as Staging.


lmuntaner (approved):
Ah! Thanks!

PR #2738, opened 2024-10-08
fix(frontend): issue with loading BTC txs

DenysKarmazynDFINITY:
# Motivation

Another example of why we need to have zod for typing worker-client communication - there was an issue with the way of how BTC address was passed to the wallet worker.


lmuntaner (approved):
Thanks!

peterpeterparker (approved):
You don't have to convince me ;)

PR #2739, opened 2024-10-08
refactor(frontend): move the Zero Balance trigger in the token list [GIX-3046]

AntonioVentilii-DFINITY:
# Motivation

On Desktop and Tablet, we want to show the "Hide zero balances" button in stead of the context menu, on top of the token list. However, for Mobile, we leave the current menu (with a new Icon) and we hide the "Manage tokens" button (since it's inside the menu.

# Changes

- Adjust ButtonIcon component to be reused for the set of class colors required: we make it so `link` prop is by default `True` (the most used case FOR NOW), but can be adjusted.
- Wrap the Context menu into a `<div>` to manage the visibility and change the icon.
- Include the [new checkbox component](https://github.com/dfinity/oisy-wallet/pull/2808) and adapt visibility for larger screens.
- Hide `ManageTokensButton` for mobile.

# Tests

https://github.com/user-attachments/assets/ef2329c2-c45a-4cf5-9bf0-1768673c6d23


peterpeterparker (commented):
If you don't mind too much, can we have the refactoring - creating the new component - in a separate PR?

AntonioVentilii-DFINITY:
@peterpeterparker sure! First the component one:

- PR #2808

AntonioVentilii-DFINITY:
@peterpeterparker you were right! This is a way cleaner PR after having it split! Please re-review?

src/frontend/src/lib/components/tokens/TokensMenu.svelte
```diff
@@ -1,34 +1,42 @@
...
+	</ButtonIcon>
 
-		<div class="my">
-			<Hr />
-		</div>
+	<Popover bind:visible anchor={button} invisibleBackdrop direction="rtl">
+		<div class="flex flex-col gap-3">
+			<TokensZeroBalance />
+
+			<div class="my">
```
peterpeterparker, line 35: `my` exists? it's a shorthand for `my-0`?

peterpeterparker (approved):
LGTM, thx

I added a minor question.

Beside that, when reviewing the PR, I was really hesitant, for once, about asking myself if we should really use CSS instead of JavaScript to hide or display the components, as there are quite a few differences between the two states, particular if the content of the popovers grows. Let's see...

src/frontend/src/lib/components/tokens/TokensMenu.svelte
AntonioVentilii-DFINITY: that was actually there before, but you see modified because I wrapped the entire element and it shifted (Github strangly shows it as modified).

However, it is not a class, so I will remove it, since we are here

PR #2740, opened 2024-10-08
fix(frontend): Add more contrast to the icons in the transaction list

AntonioVentilii-DFINITY:
# Motivation

As noted by @DenysKarmazynDFINITY , I forgot to adjust the contrast between the background and the icons in the transaction list.

### Before

Screenshot 2024-10-08 at 18 49 45

### After

Screenshot 2024-10-08 at 18 49 33


peterpeterparker (approved):
Thx

PR #2741, opened 2024-10-08
fix(frontend): Footer was not stuck to bottom when logged in [GIX-3049]

AntonioVentilii-DFINITY:
# Motivation

When logged in, the footer is not completely stuck to the bottom, because it is centered in its container.

### Before

Screenshot 2024-10-08 at 19 21 14
Screenshot 2024-10-08 at 19 21 26

### After

https://github.com/user-attachments/assets/0ebe7ed4-3658-41b6-a552-23498499d9af

https://github.com/user-attachments/assets/db282e27-fcd3-4466-b4b4-6079f5a14776


peterpeterparker (approved):
LGTM, thx

PR #2742, opened 2024-10-08
style(frontend): shift the main content up when on larger screens [GIX-3046]

AntonioVentilii-DFINITY:
# Motivation

For the new layout, the first step is to move the main content in the middle of the header when the screen is larger.

Furthermore, we remove the `Alpha` element and we move `Hero` inside `<main>` since we will treat what's inside it as a single component.

NOTE: the changes are just for larger screens when logged in. The rest is unchanged.

https://github.com/user-attachments/assets/e42b9c34-8b05-4821-8303-61ce8235d689


AntonioVentilii-DFINITY:
Done with:

- PR #2747

src/frontend/src/lib/components/hero/Hero.svelte
```diff
@@ -7,6 +7,6 @@
 	export let back = false;
 </script>
 
-<article class="main relative flex flex-col items-center rounded-lg pb-6 pt-10">
+<article class="relative flex flex-col items-center rounded-lg">
```
peterpeterparker, line 10: The global style selector `.main` can be removed?

AntonioVentilii-DFINITY:
This PR requires to remove the Alpha warning, but we will remove it only once in production.

Since it is not critical to move the entire thing up, for now we close this

PR #2743, opened 2024-10-08
fix(frontend): Remove the back button when logged out

AntonioVentilii-DFINITY:
# Motivation

There was a bug so that when an User logs out from the setting page, the `Back` button will stay there in the `Header`. We check for logged in to display it now.


peterpeterparker (approved):
I'm not sure this was incorrect, feature wise. I was also hesitant to approve, asking myself if it's not the consumer that should be extended to set the property `back` to false or true because with this implementation the "Back" button has two layers of logic. In the consumer if is true or false depending of the route and within the component it is in addition displayed or not according the authentication.

That said, LGTM, for sure it works.

AntonioVentilii-DFINITY:
we are going to use a menu for navigation, so maybe the Back button will move elsewhere. Closing this

PR #2745, opened 2024-10-09
feat(devops): Support topping up mainnet

bitdivine:
# Motivation
As paid APIs are used in staging and production, we will need to top up the cycles ledger balance for those backends.  The topping up command currently supports only local deployments.

# Changes
- Extend the top up script to support topping up accounts on mainnet.

# Tests
- Local topping up is already exercised in CI
- Mainnet topping up was exercised as follows:
  - Topping up the production backend account
```
max@sinkpad:~/dfn/oisy-wallet/branches/papi-slice-2 (1:11)$ ./scripts/top-up-cycles-ledger-account --network ic 
Balance before: (9_997_000_000_000 : nat)
Transfer sent at block index 21_420
Balance after: (10_997_000_000_000 : nat)
```
- Topping up the staging backend account.
```
max@sinkpad:~/dfn/oisy-wallet/branches/papi-slice-2 (1:22)$ ./scripts/top-up-cycles-ledger-account --network staging
Balance before: (3_142_360_000_000 : nat)
Transfer sent at block index 21_421
Balance after: (4_142_360_000_000 : nat)
max@sinkpad:~/dfn/oisy-wallet/branches/papi-slice-2 (1:43)$ 
```


scripts/clap.bash
```diff
@@ -0,0 +1,168 @@
+# clap - a BASH argument parser
```
bitdivine, line 1: Note: This is a stock bash argument parser, used in nns-dapp and other repos.  It is copied here verbatim from the nns-dapp repo.  See the other file for usage.

scripts/top-up-cycles-ledger-account
```diff
@@ -1,38 +1,93 @@
...
 
-[[ "${1:-}" != "--help" ]] || {
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=n long=network desc="dfx network to use" variable=NETWORK default="local"
+clap.define short=c long=cycles desc="Number of cycles" variable=CYCLES default="1T"
+clap.define short=t long=to desc="The destinantion principal or canister name" variable=ACCOUNT default="backend"
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
```
bitdivine, line 32: This is where the argument parser sets NETWORK, CYCLES and ACCOUNT from the cli flags.

scripts/top-up-cycles-ledger-account
AntonioVentilii-DFINITY, line 1: since we are here, may you please add the extension to the file? just for consistency with the other scripts

scripts/top-up-cycles-ledger-account
```diff
@@ -1,38 +1,93 @@
...
-echo "Balance after: $(dfx canister call cycles_ledger icrc1_balance_of '(record { owner = principal "'"$account"'"; subaccount = null })')"
+: Converts e.g. 1T cycles to a number:
+CYCLES="${CYCLES//k/K}"
+CYCLES="${CYCLES//m/M}"
+CYCLES="${CYCLES//g/G}"
+CYCLES="${CYCLES//t/T}"
+CYCLES="${CYCLES//K/_000}"
+CYCLES="${CYCLES//M/_000_000}"
+CYCLES="${CYCLES//G/_000_000_000}"
+CYCLES="${CYCLES//T/_000_000_000_000}"
```
AntonioVentilii-DFINITY, line 48: shall there be a check for unexpected values here? or will it auto-fail?

scripts/top-up-cycles-ledger-account
```diff
@@ -1,38 +1,93 @@
...
+	'
+}
+top_up_mainnet() {
+  : "Tops up a cycles account on mainnet or similar, where cycles cannot be magicked out of nowhere."
+  dfx cycles transfer "$ACCOUNT" "$cycles" --network "$NETWORK"
+}
+is_test_env() {
+  : "Checks whether this is a test environment."
+  : TODO: Add support for devenv
+  [[ "$NETWORK" == "local" ]]
```
AntonioVentilii-DFINITY, line 76: shall we add `test_fe_` here too? if for example we are playing with those, do we want to top_up staging backend? I imagine yes, it is needed, but just to double-confirm

scripts/top-up-cycles-ledger-account
bitdivine: Can do.  I'm not a huge fan of suffixes on executable because that ties the executable into a particular implementation. If you look at say Mac builtin commands about 100 are python scripts, but you would never know when executing them.  A command is just a something that performs an action.  How it performs that action is an implementation detail.  On Ubuntu there are 122 builtin commands that don't end in .py, for example core tools like `keyring`:

```
$ grep -r -m1 -E '#!.*python' /bin /usr/bin/ | grep -v py: | wc -l
122
```

But you are right, consistency has value.

scripts/top-up-cycles-ledger-account
bitdivine: It will auto-fail.

scripts/top-up-cycles-ledger-account
bitdivine: For the purposes of this function, `test_fe_1` is on mainnet and mainnet is not a test environment.  It's a test canister on a production environment.  But thank you for asking, I'll update the comment to clarify this.

scripts/top-up-cycles-ledger-account
bitdivine: Done.

scripts/top-up-cycles-ledger-account
bitdivine: Done.

scripts/top-up-cycles-ledger-account
bitdivine: Another example occurred to me:  `npm` is implemented as a script but has no suffix.  For imports and libraries, the source code language matters.  Not usually for commands.

PR #2749, opened 2024-10-09
test(e2e): wait for balance to appear

AntonioVentilii-DFINITY:
# Motivation

The Update snapshots script would always save a different image because it would not wait for the balance to appear. So sometimes it would snapshot with the skeleton and sometimes without.


DenysKarmazynDFINITY (approved):
LGTM, thanks!

PR #2751, opened 2024-10-09
refactor(frontend): move decodeQrCode util to lib

DenysKarmazynDFINITY:
# Motivation

Move decodeQrCode util to lib so it can be reused by BTC.


AntonioVentilii-DFINITY (approved):
lgtm! tks

PR #2753, opened 2024-10-09
fix(build): Adjust style according to Sass warnings

AntonioVentilii-DFINITY:
# Motivation

As found by @peterpeterparker , a recent change in the button styles created Sass compiler to raise some deprecation warnings, for example:

```bash
Deprecation Warning: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.

More info: https://sass-lang.com/d/mixed-decls

    ╷
11  │ ┌     &:not([disabled]) {
12  │ │         cursor: pointer;
13  │ │     }
    │ └─── nested rule
... │
16  │       border-radius: var(--button-border-radius);
    │       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ declaration
    ╵
    src/frontend/src/lib/styles/global/button.scss 16:2  @import
    src/frontend/src/lib/styles/global.scss 12:9         root stylesheet
```


peterpeterparker (approved):
LGTm, thx

PR #2757, opened 2024-10-09
feat(frontend): formCancelAction prop for BtcSendTokenWizard

DenysKarmazynDFINITY:
# Motivation

Extending BtcSendTokenWizard with an additional prop that tells component which button should it render - close or back.


peterpeterparker (approved):
LGTM, thx

PR #2759, opened 2024-10-09
feat(devops): Add and include cfs_canister_id init

lmuntaner:
# Motivation

We want to remove the `address` parameter from the `SelectedUtxosFeeRequest` and derive it from the caller.

In this PR, I included the CFS canister id which will be used to fetch the ECDSA public key before deriving the address.

# Changes

## Manual changes

* New parameter `cfs_canister_id` in `Config` and `InitArg`.
* Add new parameter in deploy scripts `deploy.args.sh` and `deploy.backend.sh`.
* Add it to the move from init to the config.

## Automatic changes

Using `npm run generate`:
* Changes in candid files.
* Changes in declarations files.

# Tests

* Add new `cfx_canister_id` as `None` in the integration tests. It's not yet used.


lmuntaner:
@bitdivine please review

bitdivine (approved):

PR #2760, opened 2024-10-09
feat(frontend): display regtest BTC address

DenysKarmazynDFINITY:
# Motivation

We now need to display a separate BTC address for regtest (previously, it was always the same with testnet).


peterpeterparker (approved):
LGTM, thx

PR #2762, opened 2024-10-09
refactor(frontend): Use eth_address endpoint

bitdivine:
# Motivation
The signer canister `eth_address_of_caller` is identical to the `eth_address` endpoints with no arguments, falling back to the default.  So the `.._of_caller` variant is deprecated.

# Changes
- Use a different endpoint when getting the user's eth address.

# Tests
- Existing CI should suffice.
- Also deployed to a testnet:
```
dfx deploy frontend --network test_fe_3
```
I logged in and the eth balance was shown correctly.


src/frontend/src/lib/canisters/signer.canister.ts
```diff
@@ -66,11 +66,11 @@ export class SignerCanister extends Canister<SignerService> {
 	};
 
 	getEthAddress = async (): Promise<EthAddress> => {
-		const { eth_address_of_caller } = this.caller({
+		const { eth_address } = this.caller({
 			certified: true
 		});
 
-		const response = await eth_address_of_caller([
+		const response = await eth_address({ principal: [] }, [
```
peterpeterparker, line 73: What's `principal: []`? It's a bit a generic naming, maybe we can add a comment?

bitdivine: The arguments to `eth_address` are `(EthAddressRequest{principal: optional principal, defaulting to caller}, PaymentType{})`.  I can add a comment but I suspect that IDEs probably show function signatures on hover, if set up properly for typescript which I'm only minimally.  I'll add the comment.

peterpeterparker: >  I can add a comment but I suspect that IDEs probably show function signatures on hover, 

Haha, good one. It's not just that the documentation isn't copied to Candid types when automatically generated; the bindings are also unfriendly for IDEs. When I click on the function's usage, I end up in a generic interface of agent-js. Two DX issues in a single piece of code.

Having to specify the `principal None` is new, so I would just add a comment. Something like `// In OISY, we derive the ETH address from the caller. Therefore, we are not providing a principal as an argument` or something like that. WDYT?

PR #2763, opened 2024-10-09
test(frontend): hide console.error logs for query.ic methods

DenysKarmazynDFINITY:
# Motivation

As agreed with David offline, we want to keep unit tests log clean and therefore need to mock console.error for query.ic methods.


peterpeterparker (approved):
Super, thanks a lot!

PR #2764, opened 2024-10-09
feat(frontend): calculate max BTC amount on send

DenysKarmazynDFINITY:
# Motivation

Calculate max BTC amount on send, for now without fees.

### DIFF
--- a/src/frontend/src/btc/components/send/BtcSendAmount.svelte
+++ b/src/frontend/src/btc/components/send/BtcSendAmount.svelte
@@ -4,7 +4,6 @@
 	import type { BtcAmountAssertionError } from '$btc/types/btc-send';
 	import { SEND_CONTEXT_KEY, type SendContext } from '$icp-eth/stores/send.store';
 	import SendInputAmount from '$lib/components/send/SendInputAmount.svelte';
-	import { ZERO } from '$lib/constants/app.constants';
 	import { tokenDecimals } from '$lib/derived/token.derived';
 	import { getMaxTransactionAmount } from '$lib/utils/token.utils';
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/btc/components/send/BtcSendAmount.svelte
```diff
@@ -1,10 +1,32 @@
...
 	export let amountError: BtcAmountAssertionError | undefined;
+
+	const { sendToken, sendBalance } = getContext<SendContext>(SEND_CONTEXT_KEY);
+
+	$: calculateMax = (): number | undefined =>
+		isNullish($sendToken)
+			? undefined
+			: // TODO: Add fee to the calculation
+				getMaxTransactionAmount({
+					balance: $sendBalance ?? ZERO,
```
AntonioVentilii-DFINITY, line 21: i think you don't need a fallback here (it is included in the function)

I see that it is used with fallback in `SendForm`, but it is not needed there either

AntonioVentilii-DFINITY, line 21: i think you don't need a fallback here (it is included in the function)

I see that it is used with fallback in `SendForm`, but it is not needed there either

AntonioVentilii-DFINITY (approved):
LGTM!

Just check my comment before, but not a blocker

src/frontend/src/btc/components/send/BtcSendAmount.svelte
DenysKarmazynDFINITY: True, I changed it to `sendBalance ?? undefined` to comply with types (`balance | null | undefined` -> `balance | undefined`).

PR #2765, opened 2024-10-09
refactor(frontend): Extract Button that represents Done action

AntonioVentilii-DFINITY:
# Motivation

Another step to reduce duplication and concentrate components in one place: we extract the button the is used as "Done".

NOTE: The button is always at full width.


src/frontend/src/lib/components/ui/ButtonDone.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+	import Button from '$lib/components/ui/Button.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+
+	export let type: 'submit' | 'reset' | 'button' = 'submit';
```
peterpeterparker, line 5: Is it really ever used to submit a form? If not, should we also set it as `button` for all use case?

AntonioVentilii-DFINITY: I didn't want to change the logic to be honest, just in case it is needed as `submit` somewhere.

But it make sense that "Done" does not submit a form. 

I double checked and it is really never used in a form, you are right! Changed

peterpeterparker: Makes sense. THx

peterpeterparker (approved):
LGTM, thx

PR #2766, opened 2024-10-09
refactor(frontend): Extract Button that represents Next action

AntonioVentilii-DFINITY:
# Motivation

Another step to reduce duplication and concentrate components in one place: we extract the button the is used as "Next".


peterpeterparker (approved):
LGTM, thx

PR #2768, opened 2024-10-09
style(frontend): Adjust the Hero for four buttons

AntonioVentilii-DFINITY:
# Motivation

The spacing in the Hero for four buttons was off.

### Before

Screenshot 2024-10-09 at 19 46 55

### After

Screenshot 2024-10-09 at 19 53 50

### With 3 Buttons

Screenshot 2024-10-09 at 19 45 10


peterpeterparker (approved):
LGTM, thx

PS: We are at the limit, there isn't any text longer than "ckSepoliaETH" right?

PR #2769, opened 2024-10-09
docs(frontend): Comment on eth address derivation

bitdivine:
# Motivation
The reason why the caller address is needed isn't clear.

# Changes
- Explain why the caller principal is important when getting the user eth address.

# Tests
N/A


peterpeterparker (approved):
LGTM, thanks!

PR #2771, opened 2024-10-10
feat(frontend): Provide payment with signer methods

bitdivine:
# Motivation
Calls to the signer need to be paid.

# Changes
- Add the payment argument to all signer API calls.

# Tests
- The BTC calls already used the paid endpoints and had tests, so these tests just needed to be updated.
- Others needed to be switched to the paid API and so the API signatures used in the tests needed to be updated as well.
- This branch has been deployed to mainnet here: https://dxvmu-wiaaa-aaaah-aqzaa-cai.icp0.io/
  - [x] Login works (gets Eth address)
  - Redeployed with BTC enabled in .env:
    ```
    VITE_NETWORK_BITCOIN_ENABLED=true
    VITE_BITCOIN_MAINNET=true
    ```
  - The BTC balance is NOT appearing. There are messages about infura rate limit exceeded in the console.


src/frontend/src/lib/constants/app.constants.ts
```diff
@@ -57,6 +58,13 @@ export const SIGNER_CANISTER_ID = LOCAL
 		? import.meta.env.VITE_STAGING_SIGNER_CANISTER_ID
 		: import.meta.env.VITE_IC_SIGNER_CANISTER_ID;
 
+export const SIGNER_PAYMENT_TYPE = {
```
peterpeterparker, line 61: Will there be more constant? If yes, I would create a new module at the same level and name it for example `signer.constants.ts`

Actually even if there aren't more constant. In `app.constant` it's mostly IDs, env variable mapping or things like seconds. It feels those are quite distinctive in regards to this particular new constant.

src/frontend/src/lib/canisters/signer.canister.ts
```diff
@@ -92,28 +93,65 @@ export class SignerCanister extends Canister<SignerService> {
...
+			const {
+				Ok: { signature }
+			} = response;
+			return signature;
+		}
+
+		throw mapSignerCanisterGetEthAddressError(response.Err);
 	};
 
 	sendBtc = async ({
```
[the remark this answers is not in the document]
peterpeterparker: What does that mean?

src/frontend/src/lib/canisters/signer.canister.ts
```diff
@@ -42,7 +45,9 @@ export class SignerCanister extends Canister<SignerService> {
 			certified: true
 		});
 
-		const response = await btc_caller_address({ network, address_type: { P2WPKH: null } }, []);
+		const response = await btc_caller_address({ network, address_type: { P2WPKH: null } }, [
```
peterpeterparker, line 48: Should we also extract a constant for `{ P2WPKH: null }`?

src/frontend/src/lib/canisters/signer.canister.ts
```diff
@@ -92,28 +93,65 @@ export class SignerCanister extends Canister<SignerService> {
 		return address;
 	};
 
-	signTransaction = ({ transaction }: { transaction: SignRequest }): Promise<string> => {
-		const { sign_transaction } = this.caller({
+	signTransaction = async ({
```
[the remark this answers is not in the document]
peterpeterparker: Can you to please provide this in another PR?

i.e. one for the constants, one for the new API? This feel like two different topics

lmuntaner:
How about if the new endpionts are all added in one PR in signer.canister.ts.

Then, in another PR, we change the signer.api.ts to use the new endpoints?

Then, in another PR, we remove the old endpoints from signer.canister.ts?

src/frontend/src/lib/canisters/signer.canister.ts
bitdivine: Possibly.  @lmuntaner , what do you think?

src/frontend/src/lib/canisters/signer.canister.ts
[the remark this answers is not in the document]
bitdivine: @peterpeterparker  I have extracted the payment type here:  https://github.com/dfinity/oisy-wallet/pull/2779  Can you review, please?

For the bitcoin constant I'll wait for Llorenc's recommendation as that is his baby and not changed by this PR.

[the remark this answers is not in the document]
lmuntaner: I can do the constant in another PR when this migration is finished.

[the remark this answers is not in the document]
bitdivine: Done: Now this PR does nothing but provide payment for signer API calls.

src/frontend/src/lib/canisters/signer.canister.ts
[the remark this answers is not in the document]
bitdivine: This API call did not specify how it was going to pay, so the call would have failed once payment is enforced.

src/frontend/src/lib/constants/app.constants.ts
bitdivine: Done.

src/frontend/src/lib/canisters/signer.canister.ts
```diff
@@ -91,28 +97,65 @@ export class SignerCanister extends Canister<SignerService> {
...
 		});
 
-		return sign_transaction(transaction);
+		const response = await eth_sign_transaction(transaction, [SIGNER_PAYMENT_TYPE]);
+
+		// If the response does not match the type signature, so has neither `Ok` nor `Err`,
+		// will typescript have thrown an error before this point?  Ditto for the other APIs.
+		// It seems safer to check for `Ok` in response, and always throw an error if it's not there.
+
+		if ('Ok' in response) {
```
peterpeterparker, line 115: That's probably more a nitpick but, given that in existing functions we check for the `if Err` and then return the response, I would apply the same order here for consistency reasons.

```
if ("Err" in response)
   throw mapSigner...
   
const {Ok: {signature}} = response;
return signature
```

On the contrary, if it makes more sense to check for "Ok" first, then let's rework the existing functions in another PR.

WDYT?

src/frontend/src/lib/canisters/signer.canister.ts
```diff
@@ -91,28 +97,65 @@ export class SignerCanister extends Canister<SignerService> {
...
+	personalSign = async ({ message }: { message: string }): Promise<string> => {
+		const { eth_personal_sign } = this.caller({
 			certified: true
 		});
 
-		return personal_sign(message);
+		const request: EthPersonalSignRequest = { message };
+		const response = await eth_personal_sign(request, [SIGNER_PAYMENT_TYPE]);
+
+		if ('Ok' in response) {
```
peterpeterparker, line 133: same same

src/frontend/src/lib/canisters/signer.canister.ts
```diff
@@ -91,28 +97,65 @@ export class SignerCanister extends Canister<SignerService> {
...
+	signPrehash = async ({ hash }: { hash: string }): Promise<string> => {
+		const { eth_sign_prehash } = this.caller({
 			certified: true
 		});
 
-		return sign_prehash(hash);
+		const request: EthSignPrehashRequest = { hash };
+		const response = await eth_sign_prehash(request, [SIGNER_PAYMENT_TYPE]);
+
+		if ('Ok' in response) {
```
peterpeterparker, line 151: same same

src/frontend/src/lib/canisters/signer.canister.ts
bitdivine: I don't know whether typescript throws an exception if, at runtime, the wrong data type is provided by an API.  If it does throw an exception, it doesn't matter which patetrn is used.  If it doesn't, checking for Ok handles wacky data.  So I would rather convert everything to the defensive version, but if you are 100% sure that typescript enforces types at runtime I can switch to the pattern used elsewhere.

bitdivine: It looks as if there is no runtime check, so I'll convert the existing methods.  But I'll do that in a separate PR to avoid mixing topics.

peterpeterparker: At build time, TypeScript prevents it indeed.
At runtime, Candid should throws (I think), if anything other than Ok or Err is provided.

That said, I think you make a good point anyway, so let's rather convert everything to defensive and move on with this PR.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/canisters/signer.canister.ts
bitdivine: Thanks.  Yes, it would good to have a close look at the candid parser so that we know what integrity guarantees it provides.

PR #2775, opened 2024-10-10
style(frontend): adjust main content to have a menu column [GIX-3050]

AntonioVentilii-DFINITY:
# Motivation

We want to have a lateral column with the menu, but only for desktop and tablets. So, this PR will first adjust the main content layout. We will add the menu in another PR (there is a `TODO`).

# Changes

- Include Hero in the central part of the main content, and consequently adjust its layout (spacing, transition, etc).
- Create a component for the MainLayout:
  - On Desktop, it will show 3 columns (the menu on the left, the central main content, and an empty column on the right side).
  - On Tablet, only the menu and the central content.
  - On Mobile, only the central content.

# Tests

I just placed an example for the menu to visually see.

https://github.com/user-attachments/assets/b80025bf-f3d9-4383-93ae-8cc9d92a3fcc

Slowed transition:

https://github.com/user-attachments/assets/3aeb6670-6556-4c5c-b2c5-62679c143d37


peterpeterparker:
I haven't reviewed the code yet, so this is just initial feedback based on the video: In my opinion, the appearance and disappearance of the menu should have a smoother transition, rather than a 'on/off' effect. For inspiration, you might check out the transitions used on https://nns.ic0.app/ or https://recyclingcredit.org/

AntonioVentilii-DFINITY:
@peterpeterparker you are right! I didn't put it there yet, because the Alpha warning was practically causing issues (it shifts down, so the transitions was a bit glitchy). And since we are going to remove it soon enough, I would have done it after.

I put it as last commit in this PR anyway, just so that if you want you can have a look and decide if we keep it or remove it. The results without Alpha are:


https://github.com/user-attachments/assets/b0f34c36-9141-40b3-aba2-850a40eb7a6c

src/frontend/src/lib/components/hero/Hero.svelte
```diff
@@ -1,12 +1,16 @@
...
 	export let summary = false;
-	export let actions = true;
 	export let back = false;
 </script>
 
-<article class="main relative flex flex-col items-center rounded-lg pb-6 pt-10">
-	<HeroContent {usdTotal} {summary} {actions} {back} />
+<article
+	class="relative flex flex-col items-center rounded-lg pb-6"
+	transition:slide={{ delay: 0, duration: 250, easing: quintOut, axis: 'y' }}
```
peterpeterparker, line 13: Not for this PR, just thinking out at loud, maybe we can create a constant for `{ delay: 0, duration: 250, easing: quintOut, axis: 'y' }`

src/frontend/src/lib/components/hero/HeroContent.svelte
```diff
@@ -66,14 +65,12 @@
 		</div>
 	{/if}
 
-	{#if actions}
```
peterpeterparker, line 69: This is not related to this PR per sé, right?

src/frontend/src/lib/components/ui/MainLayout.svelte
```diff
@@ -0,0 +1,15 @@
+<main class="relative mx-auto max-w-screen-2.5xl pt-8 sm:flex sm:flex-row">
```
peterpeterparker, line 1: Do you mind if we we name this component `SplitPane`?

I generally do so, so it would help me if I need to retrieve it, and also that way we use the same naming convention as in gix-cmp for example.

src/frontend/src/lib/styles/global/layout.scss
```diff
@@ -1,13 +1,9 @@
...
+header {
 	margin: 0 auto;
-	padding-inline: var(--padding-2_5x);
-
-	max-width: media.$breakpoint-small;
 }
 
-header {
-	margin: 0 auto;
+.central-content {
```
peterpeterparker, line 7: What's central? It's the main paine?

src/frontend/src/routes/(app)/+layout.svelte
```diff
@@ -31,18 +32,23 @@
...
+					usdTotal={route === 'tokens'}
+					summary={route === 'transactions'}
+					back={route === 'transactions'}
+				/>
+			{/if}
+
 			<LoadersGuard>
 				<slot />
 			</LoadersGuard>
-		</main>
```
peterpeterparker, line 45: There is no more main do I get it right?

src/frontend/src/lib/components/hero/Hero.svelte
AntonioVentilii-DFINITY: YES! THANK YOU (Micheal Scott style)

src/frontend/src/lib/components/hero/HeroContent.svelte
AntonioVentilii-DFINITY: right! i put it here just to simplofy, but I can remove it once we merge this one, what do you prefer?

src/frontend/src/lib/components/ui/MainLayout.svelte
AntonioVentilii-DFINITY: Yes, for sure, will do!

P.S. I tried to use that GIX component but the requirements say that the central part is always "centered" for desktop, and I wasn't able to adapt it to the case

src/frontend/src/lib/styles/global/layout.scss
AntonioVentilii-DFINITY: correct, i will change the name: `main-content`?

src/frontend/src/routes/(app)/+layout.svelte
AntonioVentilii-DFINITY: Well, `main` is inside `MainLayout`, do you want to separate them?

src/frontend/src/lib/components/hero/Hero.svelte
AntonioVentilii-DFINITY: Done with PR:

- PR #2807

src/frontend/src/lib/components/hero/HeroContent.svelte
peterpeterparker: All good

src/frontend/src/lib/components/ui/MainLayout.svelte
AntonioVentilii-DFINITY: done!

src/frontend/src/lib/styles/global/layout.scss
AntonioVentilii-DFINITY: Changed to `main-content`

AntonioVentilii-DFINITY:
@peterpeterparker ready for another review, please

src/frontend/src/routes/(app)/+layout.svelte
peterpeterparker: Ah, my bad. It's good.

src/frontend/src/lib/components/hero/Hero.svelte
```diff
@@ -1,12 +1,16 @@
...
 	export let summary = false;
-	export let actions = true;
 	export let back = false;
 </script>
 
-<article class="main relative flex flex-col items-center rounded-lg pb-6 pt-10">
-	<HeroContent {usdTotal} {summary} {actions} {back} />
+<article
+	class="relative flex flex-col items-center rounded-lg pb-6"
+	transition:slide={SLIDE_PARAMS}
```
peterpeterparker, line 13: There is no `#if`, this transition has no effect no?

src/frontend/src/lib/components/hero/HeroContent.svelte
```diff
@@ -66,11 +65,9 @@
...
-		<div transition:slide={SLIDE_PARAMS} class="flex w-full justify-center text-left">
-			<Actions />
-		</div>
-	{/if}
+	<div transition:slide={SLIDE_PARAMS} class="flex w-full justify-center text-left">
```
peterpeterparker, line 68: ditto about the transition

peterpeterparker (commented):
Can you verify that the animation of the hero remains smooth when you navigate accross token? See comments.

src/frontend/src/lib/components/hero/Hero.svelte
AntonioVentilii-DFINITY: The IF condition is outside the component (see `+layout` page), so this one works

src/frontend/src/lib/components/hero/HeroContent.svelte
AntonioVentilii-DFINITY: in this case, the if is a few levels above, where a parent has already a `slide`, so we need to add `|local` to avoid concurrency

AntonioVentilii-DFINITY:
@peterpeterparker I included in the description a slowed down transition

src/frontend/src/lib/components/hero/Hero.svelte
peterpeterparker: Really? If was under the impression that Svelte requires an animation to be preceded by a IF. Good to know.

peterpeterparker (approved):
LGTm, thx

src/frontend/src/lib/components/hero/Hero.svelte
AntonioVentilii-DFINITY: Looking at [documentation](https://svelte.dev/docs/element-directives#transition-fn), it actually mention it, I didn't know before testing

> A transition is triggered by an element entering or leaving the DOM as a result of a state change.
> 
> When a block is transitioning out, all elements inside the block, including those that do not have their own transitions, are kept in the DOM until every transition in the block has been completed.

PR #2779, opened 2024-10-10
refactor(frontend): Extract payment type as a constant

bitdivine:
# Motivation
The payment type is the same for every paid signer API call.

# Changes
- Move the payment type into a constants file.

# Tests
Existing CI should suffice.


lmuntaner (approved):
LGTM

PR #2780, opened 2024-10-10
fix(devops): Deploy signer before backend

lmuntaner:
# Motivation

The backend needs the CFS canister id.

In this PR, first signer is deployed and then the backend in the `deploy.sh`.

# Changes

* Change order of scripts in `deploy.sh`

# Tests

I tested with a clean dfx replica running `npm run deploy`.


peterpeterparker (approved):
Thanks a ton!

PR #2781, opened 2024-10-10
fix(frontend): networkId always undefined in SendModal

DenysKarmazynDFINITY:
# Motivation

The `networkId` prop  that is passed down from `SendModal` is actually only defined when `targetNetwork` is there. Because of that, `BtcSendForm` was never receiving `networkId` even though it is relying on it to identify whether it's a regtest, testnet or mainnet btc.


peterpeterparker (approved):
LGTM, thx

PR #2782, opened 2024-10-10
feat(frontend): disable send BTC button if conditions are not met

DenysKarmazynDFINITY:
# Motivation

We need to disable send BTC button if conditions are not met.

### DIFF
--- a/src/frontend/src/btc/components/send/BtcSendForm.svelte
+++ b/src/frontend/src/btc/components/send/BtcSendForm.svelte
@@ -20,6 +20,7 @@
 	let amountError: BtcAmountAssertionError | undefined;
 	let invalidDestination: boolean;
 
+	// TODO: check if we can align this validation flag with other SendForm components (e.g IcSendForm)
 	let invalid = true;
 	$: invalid =
 		invalidDestination ||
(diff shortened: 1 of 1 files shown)


src/frontend/src/btc/components/send/BtcSendForm.svelte
```diff
@@ -18,6 +20,13 @@
 	let amountError: BtcAmountAssertionError | undefined;
 	let invalidDestination: boolean;
 
+	let invalid = true;
```
peterpeterparker, line 23: This will grows with more condition?

It's exactly the same rule as in IcSendForm, so just wondering if we should add a TODO to potentially refactor it some day.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/btc/components/send/BtcSendForm.svelte
DenysKarmazynDFINITY: Yes, in case of BTC this flag will have more conditions, but still it would be nice to align some basic validation checks with other network's SendForms. Added TODO comment.

PR #2783, opened 2024-10-10
refactor(frontend): Provide type for payment type

bitdivine:
# Motivation
The `SIGNER_PAYMENT_TYPE` inner value was fixed, but cleaner is simply to declare the type of the constant.

# Changes
- Declare the type of `SIGNER_PAYMENT_TYPE` instead of coercing the inner.

# Tests
Existing CI should suffice.


peterpeterparker (approved):
Nice, thanks!

PR #2786, opened 2024-10-10
feat(backend): Derive address from caller select utxos

lmuntaner:
# Motivation

Avoid misusage of `btc_select_user_utxos_fee` by using the address of the caller.

# Changes

## Manual changes

* Install `bitcoin` crate.
* Remove `source_address` from `SelectedUtxosFeeRequest`.
* Create `btc_principal_to_p2wpkh_address` in signer module. This is what the Signer does, but using the CFS canister id instead of its own.
* Get the source_address from signer module instead of the param in the endpoint logic.

## Automatic changes

Candid files and declarations by running `npm run generate`.

# Tests

I haven't been able to test because the UI doesn't use the endpoint yet. But the code added is the same as in CFS which is working.

I'll test it as soon as the UX uses the endpoint and fix any issues encountered.


lmuntaner:
@bitdivine please review

lmuntaner:
@DenysKarmazynDFINITY FYI, this is the PR with the breaking changes

src/backend/src/signer.rs
```diff
@@ -82,3 +90,48 @@ pub fn principal2account(principal: &Principal) -> ByteBuf {
...
         .into()
 }
+
+/// Computes the public key of the specified principal.
+// TODO: Cache CFS pubkey and derive it offline as in [ckBTC minter](https://github.com/dfinity/ic/blob/35153c7cb7b9d1da60472ca7e94c693e418f87bd/rs/bitcoin/ckbtc/minter/src/address.rs#L101-L101)
+async fn cfs_ecdsa_pubkey_of(principal: &Principal) -> Result<Vec<u8>, String> {
+    let (ecdsa_key_name, maybe_cfs_canister_id) =
+        read_config(|s| (s.ecdsa_key_name.clone(), s.cfs_canister_id));
+    // As done in [CFS](https://github.com/dfinity/chain-fusion-signer/blob/26b683c6de9971fdbf7bd4cebc04d427d1753289/src/signer/canister/src/derivation_path.rs#L28)
+    let derivation_path = vec![vec![0_u8], principal.as_slice().to_vec()];
```
bitdivine, line 100: Note that that `0u8` in the first place of the derivation path applies to BTC only.  `1u8` is Eth, `0xffu8` is generic.  It might be worth renaming tyhis function, then in the signer repo we can expose this function and then you will be able to get BTC, Eth and other derivation paths.

bitdivine (approved):
Thanks!

src/backend/src/signer.rs
lmuntaner: Done in a comment and variable

PR #2787, opened 2024-10-10
feat(frontend): converNumberToSatoshis util

DenysKarmazynDFINITY:
# Motivation

Since it will be used in a few places (btc-send.services), I makde this small function available as a util.


peterpeterparker (approved):
LGTM, thanks

@lmuntaner would that be an interesting utils to move to `@dfinity/ckbtc`?

lmuntaner:
> LGTM, thanks
> 
> @lmuntaner would that be an interesting utils to move to `@dfinity/ckbtc`?

Mhm, someone could needed without using ckbtc. If we had a `dfinity/btc` I believe it should go there. But it's also pretty small and the only hidden data is the number of decimals of Bitcoin, which devs know.

PR #2791, opened 2024-10-10
feat(frontend): min reservations 0

bitdivine:
# Motivation

We want to be able to request balance for different confirmations amount.

In this PR, we expose the parameter min_confirmations to the api layer.

# Changes

- Expose `minConfirmations` in getBtcBalance canister and api layer.

# Tests

- The relevant unit test has been updated and a new one added.


src/frontend/src/tests/lib/canisters/signer.canister.spec.ts
```diff
@@ -175,6 +175,29 @@ describe('signer.canister', () => {
 		);
 	});
 
+	it('returns correct BTC balance with min_confirmations', async () => {
+		const balance = 2n;
+		const minConfirmations = 3;
```
peterpeterparker, line 180: It crashes an error if zero is passed?

lmuntaner: No, it doesn't. But 0 is the same as 1.

peterpeterparker: Thanks for the feedback!

PR #2794, opened 2024-10-10
feat(frontend): btc-send.service - selectUtxosFee

DenysKarmazynDFINITY:
# Motivation

A service to get utxos fee for the provided BTC amount.

### DIFF
--- a/src/frontend/src/btc/services/btc-send.services.ts
+++ b/src/frontend/src/btc/services/btc-send.services.ts
@@ -1,0 +1,37 @@
+import type { UtxosFee } from '$btc/types/btc-send';
+import { convertNumberToSatoshis } from '$btc/utils/btc-send.utils';
+import { selectUserUtxosFee } from '$lib/api/backend.api';
+import type { BtcAddress } from '$lib/types/address';
+import { mapToSignerBitcoinNetwork } from '$lib/utils/network.utils';
+import type { Identity } from '@dfinity/agent';
+import type { BitcoinNetwork } from '@dfinity/ckbtc';
+
+const DEFAULT_MIN_CONFIRMATIONS = 6;
+
+export const selectUtxosFee = async ({
+	identity,
+	sourceAddress,
+	network,
+	amount
+}: {
+	identity: Identity;
+	network: BitcoinNetwork;
+	sourceAddress: BtcAddress;
+	amount: number;
+}): Promise<UtxosFee> => {
+	const satoshisAmount = convertNumberToSatoshis({ amount });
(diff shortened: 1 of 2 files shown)


AntonioVentilii-DFINITY (approved):
LGTM!

DenysKarmazynDFINITY:
All the functions used by this service are tested on the utils/API/canister level, so I think we don't need further tests here.

src/frontend/src/btc/services/btc-send.services.ts
```diff
@@ -0,0 +1,37 @@
...
+	identity,
+	sourceAddress,
+	network,
+	amount
+}: {
+	identity: Identity;
+	network: BitcoinNetwork;
+	sourceAddress: BtcAddress;
+	amount: number;
+}): Promise<UtxosFee> => {
```
peterpeterparker, line 21: Shouldn't this return a `ResultSuccess`

src/frontend/src/btc/services/btc-send.services.ts
```diff
@@ -0,0 +1,37 @@
...
+	const satoshisAmount = convertNumberToSatoshis({ amount });
+	const signerBitcoinNetwork = mapToSignerBitcoinNetwork({ network });
+
+	const { fee_satoshis, utxos } = await selectUserUtxosFee({
```
peterpeterparker, line 25: You are going to handle try catch and toast on the UI side? Won't implement this here?

src/frontend/src/btc/services/btc-send.services.ts
DenysKarmazynDFINITY: This is the call that will be triggered on the SendReview step, the response is needed to display the fee amount in the modal and to start the transaction in case user proceeds with the sending.

src/frontend/src/btc/services/btc-send.services.ts
DenysKarmazynDFINITY: Good point, better to handle it here, I'll create a separate PR.

DenysKarmazynDFINITY: I just saw that e.g. in case of IC we handle it on the UI side. So, to be consistent, will do the same for BTC.

peterpeterparker: Makes sense. Thanks for double checking.

src/frontend/src/btc/services/btc-send.services.ts
peterpeterparker: Given that you are going to handle the try / catch in the UI it's fine indeed

PR #2795, opened 2024-10-10
feat(frontend): do not pass address_type as a param

DenysKarmazynDFINITY:
# Motivation

For the consistency sake, let's keep hardcoded `adrdress_type` inside of `signer.canister` (same as it's done in `getBtcBalance`) instead of passing this value all the way down from the related service.

### DIFF
--- a/src/frontend/src/lib/canisters/signer.canister.ts
+++ b/src/frontend/src/lib/canisters/signer.canister.ts
@@ -165,7 +165,6 @@ export class SignerCanister extends Canister<SignerService> {
 	};
 
 	sendBtc = async ({
-		addressType,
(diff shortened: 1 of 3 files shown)


AntonioVentilii-DFINITY (approved):
I think it was discussed somewhere, but I cannot find it: this PR is a step to have it as `{ P2WPKH: null }` constant somewhere, right?

Since I imagine you already discussed it with the others, I approve, in case that is the case eheh

DenysKarmazynDFINITY:
Yes, for now, we should always pass `{ P2WPKH: null }` as `address_type`, same as it's done in `getBtcBalance`.

PR #2802, opened 2024-10-10
feat(frontend): defensive response assertions

peterpeterparker:
# Motivation

As discussed with @bitdivine, we decided to follow a "defensive" implementation of the signer and backend API by explicitly asserting that `Ok` is set in the provided response of the actors.

# Notes

This is an edge-case approach, and given that TypeScript validates the types at build time, I did not add extra complexity in the code to interpret potential unexpected errors. This is also asserted by Candid at runtime. Therefore, I only implemented a test that asserts it fails.

# Changes

- Update order of assertions by always checking for if `Ok` is present instead of `Err`
- Provide related tests.
- Group few tests for consistency.


lmuntaner (approved):

bitdivine:
You got there before me.  Thank you!

PR #2803, opened 2024-10-10
feat(frontend): remove Networks switcher [GIX-3087]

AntonioVentilii-DFINITY:
# Motivation

Time has arrived to remove the Networks switcher.

Since OISY is a multi-chain wallet by it's nature, we can remove the network selector from the home screen. It may be added a possible filter in the future, but not so bluntly, because the user should be driven to have all the tokens from multiple chains in one place.

### Note

The toggle for Testnet networks will be available only in Settings after this PR

# Results

Screenshot 2024-10-10 at 18 00 10
Screenshot 2024-10-10 at 18 00 24

### DIFF
--- a/src/frontend/src/lib/components/tokens/Tokens.svelte
+++ b/src/frontend/src/lib/components/tokens/Tokens.svelte
@@ -6,11 +6,12 @@
 	import TokensSignedOut from '$lib/components/tokens/TokensSignedOut.svelte';
 	import Header from '$lib/components/ui/Header.svelte';
 	import { authNotSignedIn, authSignedIn } from '$lib/derived/auth.derived';
+	import { i18n } from '$lib/stores/i18n.store';
 </script>
 
 <div class:pointer-events-none={$authNotSignedIn} class:blur-[1.5px]={$authNotSignedIn}>
 	<Header>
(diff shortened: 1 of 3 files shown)


peterpeterparker:
> The toggle for Testnet networks will be available only in Settings after this PR

That's nice and stuffs but, how do I find the test tokens?

SepoliaETH tokens and ICP ckTest tokens are not displayed on the pseudo network chain fusion.

AntonioVentilii-DFINITY:
@peterpeterparker right! I would have put it there in another PR, but it makes sense to first do that. Moving to draft

peterpeterparker:
> I would have put it there all together (not trivial)

We have always been against this change from a technical point of view because, as you said, it's not trivial.

Also, as we mentioned each time this was discussed, particularly in a dApp that is not well-tested enough, it introduces a higher risk that if a bug is introduced, some tokens that are supposed to target a test environment could actually end up being exchanged in production.

Is that something that can be discussed? Or are the stakeholders aware of that potential issue?

AntonioVentilii-DFINITY:
@peterpeterparker requirement deprecated for now, closing this one

peterpeterparker:
Great! Thanks

PR #2808, opened 2024-10-10
feat(frontend): Create checkbox to hide tokens with zero balance

AntonioVentilii-DFINITY:
# Motivation

On Desktop and Tablet, we want to show the "Hide zero balances" button in stead of the context menu, on top of the token list. However, for Mobile, we leave the current menu (with a new Icon) and we hide the "Manage tokens" button (since it's inside the menu.

So, this PR creates first the Checkbox component.

# Changes

- Create checkbox component for the required button
- We use an [existing component from Gix Components](https://gix.design/components/checkbox), but we need to change some specific styles regarding mostly colors.

# Tests

I included it as test in the main page:

https://github.com/user-attachments/assets/ef2329c2-c45a-4cf5-9bf0-1768673c6d23


src/frontend/src/lib/components/tokens/TokensZeroBalanceCheckbox.svelte
```diff
@@ -0,0 +1,18 @@
+<script lang="ts">
+	import { Checkbox } from '@dfinity/gix-components';
+	import { hideZeroBalances } from '$lib/derived/settings.derived';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { hideZeroBalancesStore } from '$lib/stores/settings.store';
+
+	let checked: boolean;
+	$: checked = $hideZeroBalances;
```
peterpeterparker, line 8: Should the value be assigned `onMount` only? Is `checked` triggered each time `toggleHide` is triggered so per extension click on the checkbox set checked twice to the same value?

AntonioVentilii-DFINITY: In my understanding, the flow is:

1. Click on checkbox
2. `toggleHide` triggered --> `hideZeroBalancesStore` is set to the opposite value
3. `hideZeroBalances` derived store is updated
4. `checked` is updated

No? I am missing something maybe...

I run it with `$: checked, console.log('checked', checked);` and it seems to update only once


https://github.com/user-attachments/assets/afd7ac3a-51a3-4977-b07a-09ba7bdaa3d4

peterpeterparker: Yeah sorry, badly explained. It's what I meant.

1. Click on checkbox
2. toggleHide triggered --> hideZeroBalancesStore is set to the opposite value
3. hideZeroBalances derived store is updated
4. checked is updated // <---------------------------------- This rerender is not needed

But thinking about it, it's fine. If we use onMount, then in the future, if the state is used, there might be a discrepancy.

Nvm my comment, thanks for the feedback.

peterpeterparker (approved):
LGTM, thx

PR #2810, opened 2024-10-10
test(e2e): await loader modal to close at login

AntonioVentilii-DFINITY:
# Motivation

A few E2E tests are timing out because now the access requires a bit more time. So we include a check to wait for the loader modal to appear and then disappear, trying to "buy some time" for the rest of the test.

# Changes

- Add test ID for the Loader modal.
- In `HomepageLoggedIn` add a sequential check to first verify that the modal is present, and then when it is hidden.

# Tests

I ran the E2E Tests workflow on this branch and it worked: https://github.com/dfinity/oisy-wallet/actions/runs/11387701538


DenysKarmazynDFINITY (approved):
LGTM, thanks!

PR #2812, opened 2024-10-11
feat(frontend): Add allowance method to Send services

AntonioVentilii-DFINITY:
# Context

# Example

    1. `approve` transaction for 10 USDT --> _successful_
    2. `transfer_from` transaction for 10 USDT --> _failed_ (for whatever reason)
    1. `approve` transaction for 20 USDT --> _failed_ (the previous approved amount is still there)
    2. `transfer_from` transaction for 20 USDT --> _failed_ (not enough approved amount)
    1. `approve` transaction for 7 USDT --> _failed_ (the previous approved amount is still there)
    2. `transfer_from` transaction for 7 USDT --> _successful_ (remaining approved amount is 3 USDT)

# Practical example

# Motivation

As first step to help the User with this: for the Send flow, it is useful to know the allowance of the ERC20 smart contract for the Helper smart contract. In this PR, we add the method in the provider and in the services.

Consequently, in another PR, we will use it.

### DIFF
--- a/src/frontend/src/eth/services/send.services.ts
+++ b/src/frontend/src/eth/services/send.services.ts
@@ -170,6 +170,28 @@ const ckErc20HelperContractPrepareTransaction = async ({
 };
 
 /**
+ * Get the current allowance of an Erc20 contract.
+ */
+const erc20ContractAllowance = async ({
+	token,
+	owner,
+	spender,
+	networkId
+}: {
+	networkId: NetworkId;
+	owner: EthAddress;
+	spender: EthAddress;
+} & Pick<SendParams, 'token'>): Promise<BigNumber> => {
+	const { allowance } = infuraErc20Providers(networkId);
+
+	return await allowance({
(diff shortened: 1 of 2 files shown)


src/frontend/src/eth/services/send.services.ts
```diff
@@ -169,6 +169,29 @@ const ckErc20HelperContractPrepareTransaction = async ({
 	});
 };
 
+/**
+ * Get the current allowance of an Erc20 contract.
+ */
+// TODO: this function is still not used in the codebase, so for now we export globally
```
peterpeterparker, line 175: Maybe an eslint exception instead of exposing it if not used?

peterpeterparker (approved):
LGTm, thx

src/frontend/src/eth/services/send.services.ts
AntonioVentilii-DFINITY: better!

PR #2813, opened 2024-10-11
refactor(frontend): Extract function that prepares and signs approval

AntonioVentilii-DFINITY:
# Motivation

Following up PR #2812, the next step is to extract the part of the flow that prepare and signs an `approve` transition for ERC20.

We will re-use this function because, if necessary, we need to reset to ZERO the allowance, before requesting a new allowance.


peterpeterparker (approved):
LGTM, thx

PR #2814, opened 2024-10-11
chore(frontend): slightly simplify params of approve ERC20

AntonioVentilii-DFINITY:
# Motivation

The param `nounce` was extracted from the params list, but it is used in a more direct way. So there is no need to extract it, to simplify it slightly.


peterpeterparker (approved):
That's my jam!

PR #2819, opened 2024-10-11
feat(frontend): add component TokenSymbol with network icon

AntonioVentilii-DFINITY:
# Motivation

We change a bit the format of the token list, moving the network icon from the token logo to the side of the symbol (in Black and White).

NOTE: as discussed offline, we still don't touch the Manage Tokens Modal, leaving it as it is.

# Changes

- Create Black and White (BW) icons for networks Bitcoin, Ethereum and ICP.
- Add `iconBW` prop to each network.
- Create new component `TokenSymbol` to manage the layout.
- Include new component in the token list.

# Tests

Screenshot 2024-10-11 at 11 26 11
Screenshot 2024-10-11 at 11 26 20


src/frontend/src/lib/components/tokens/TokenSymbol.svelte
```diff
@@ -0,0 +1,15 @@
+<!-- source: DFINITY foundation -->
```
peterpeterparker, line 1: This can be removed

src/frontend/src/lib/components/tokens/TokenSymbol.svelte
```diff
@@ -0,0 +1,15 @@
...
+	import type { Token } from '$lib/types/token';
+
+	export let token: Token;
+</script>
+
+<div class="flex flex-row items-center justify-between gap-2">
+	{token.symbol}
+
+	{#if nonNullish(token.network.iconBW)}
+		<svelte:component this={token.network.iconBW} />
```
peterpeterparker, line 13: - Shouldn't we use the `Logo` component?
- The accessibility hints are missing

src/frontend/src/lib/types/network.ts
```diff
@@ -10,6 +11,7 @@ export interface Network {
 	env: NetworkEnvironment;
 	name: string;
 	icon?: string;
+	iconBW?: ComponentType;
```
peterpeterparker, line 14: - I'm pretty that I will forgot what `BW` means I have to say.
- I was also wondering if we should, in another PR, group the icons?

```
env: ...
name: string;
icons: NetworkIcons;

type NetworkIcons {
   primary: ComponentType;
   negative: ComponentType;
}
```

(naming as example)

src/frontend/src/lib/components/tokens/TokenSymbol.svelte
AntonioVentilii-DFINITY: Yes... not really sure how it ended up there... thank you!

src/frontend/src/lib/types/network.ts
AntonioVentilii-DFINITY: Yes, but i agree, another PR, so i change the name directly there, WDYT?

peterpeterparker: Another PR 👍

peterpeterparker (approved):
LGTM, thx

PR #2820, opened 2024-10-11
refactor(frontend): rename slot in Card for amount sub-text

AntonioVentilii-DFINITY:
# Motivation

It was already mentioned that `sub-amount` was not a comprehensive definition for the text slot below the amount. So, this PR aims to rename it as `amount-description`.


peterpeterparker (approved):
LGTM, thx

Just thinking out at loud: If there is a new iteration in the future, maybe we can go to a generic slot naming like `top-left`. But let's see...

PR #2821, opened 2024-10-11
feat(frontend): fetch and display UTXOs fee on BTC send review step

DenysKarmazynDFINITY:
# Motivation

The goal of this PR is to use the btc-send service to get UTXOs fee and to display it on the review step. The PR could have been split into few pieces, but I think in this particular case it's useful to see the whole picture.

Loading state:
Screenshot 2024-10-11 at 11 42 31

UTXOs fee has been fetched:
Screenshot 2024-10-11 at 11 42 36

### DIFF
--- a/src/frontend/src/btc/components/send/BtcSendReview.svelte
+++ b/src/frontend/src/btc/components/send/BtcSendReview.svelte
@@ -1,10 +1,13 @@
 <script lang="ts">
+	import { isNullish } from '@dfinity/utils';
 	import type { Readable } from 'svelte/store';
 	import BtcSendHasPendingTransactions from './BtcSendHasPendingTransactions.svelte';
+	import BtcUtxosFee from '$btc/components/send/BtcUtxosFee.svelte';
 	import {
 		BtcPendingSentTransactionsStatus,
 		initPendingSentTransactionsStatus
 	} from '$btc/derived/btc-pending-sent-transactions-status.derived';
+	import type { UtxosFee } from '$btc/types/btc-send';
 	import SendReview from '$lib/components/send/SendReview.svelte';
 	import type { NetworkId } from '$lib/types/network';
 	import { invalidAmount } from '$lib/utils/input.utils';
@@ -14,14 +17,17 @@
 	export let amount: number | undefined = undefined;
 	export let networkId: NetworkId | undefined = undefined;
 	export let source: string;
+	export let utxosFee: UtxosFee | undefined = undefined;
 
 	let hasPendingTransactionsStore: Readable<BtcPendingSentTransactionsStatus>;
 	$: hasPendingTransactionsStore = initPendingSentTransactionsStatus(source);
 
 	let disableSend: boolean;
-	// We want to disable send if pending transactions are loading, there was an error or there are pending transactions.
+	// We want to disable send if pending transactions or UTXOs fee are loading, there was an error or there are pending transactions.
 	$: disableSend =
-		$hasPendingTransactionsStore !== BtcPendingSentTransactionsStatus.NONE || invalid;
+		$hasPendingTransactionsStore !== BtcPendingSentTransactionsStatus.NONE ||
+		isNullish(utxosFee) ||
+		invalid;
 
 	// Should never happen given that the same checks are performed on previous wizard step
 	let invalid = true;
@@ -33,6 +39,8 @@
 </script>
 
 <SendReview on:icBack on:icSend {source} {amount} {destination} disabled={disableSend}>
+	<BtcUtxosFee slot="fee" on:icSelectUtxosFee {utxosFee} />
+
 	<BtcSendHasPendingTransactions
 		slot="info"
 		pendingTransactionsStatus={$hasPendingTransactionsStore}
--- a/src/frontend/src/btc/components/send/BtcSendReview.svelte
+++ b/src/frontend/src/btc/components/send/BtcSendReview.svelte
@@ -9,6 +9,7 @@
 	} from '$btc/derived/btc-pending-sent-transactions-status.derived';
 	import type { UtxosFee } from '$btc/types/btc-send';
 	import SendReview from '$lib/components/send/SendReview.svelte';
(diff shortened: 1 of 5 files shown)


src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
```diff
@@ -44,10 +59,44 @@
 
 	// TODO: implement send function when related services are ready
 	const send = () => {};
+
+	const selectUtxosFee = async () => {
+		try {
+			// all required params should be already defined at this stage
+			if (nonNullish(amount) && nonNullish(networkId) && nonNullish($authIdentity)) {
```
peterpeterparker, line 66: can you invert the if for consistency?

```
if isNullish
   return
   
logic
```

src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
```diff
@@ -44,10 +59,44 @@
 
 	// TODO: implement send function when related services are ready
 	const send = () => {};
+
+	const selectUtxosFee = async () => {
```
peterpeterparker, line 63: is this use anywhere else than in the review step?
if no, shouldn't we move this within that particular component?

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
```diff
@@ -0,0 +1,36 @@
...
+	import { i18n } from '$lib/stores/i18n.store';
+	import { formatToken } from '$lib/utils/format.utils';
+
+	export let utxosFee: UtxosFee | undefined = undefined;
+
+	const dispatch = createEventDispatcher();
+
+	onMount(() => {
+		if (isNullish(utxosFee)) {
+			dispatch('icSelectUtxosFee');
```
peterpeterparker, line 18: if we ultimately do not move the related function in this component, what do you think of using the `steps` component in the parent instead of a dispatcher and listener?

src/frontend/src/btc/components/send/BtcSendReview.svelte
```diff
@@ -14,14 +17,17 @@
...
 	let hasPendingTransactionsStore: Readable<BtcPendingSentTransactionsStatus>;
 	$: hasPendingTransactionsStore = initPendingSentTransactionsStatus(source);
 
 	let disableSend: boolean;
-	// We want to disable send if pending transactions are loading, there was an error or there are pending transactions.
+	// We want to disable send if pending transactions or UTXOs fee are loading, there was an error or there are pending transactions.
 	$: disableSend =
-		$hasPendingTransactionsStore !== BtcPendingSentTransactionsStatus.NONE || invalid;
+		$hasPendingTransactionsStore !== BtcPendingSentTransactionsStatus.NONE ||
+		isNullish(utxosFee) ||
```
peterpeterparker, line 29: The comments says "when UTXOs fee are loading" but, the if says "when utxos are not defined or null". Is that correct?

src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
peterpeterparker: actually, why not having this logic directly in `BtcUtxosFee`?

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
```diff
@@ -0,0 +1,36 @@
...
+
+<Value ref="utxos-fee" element="div">
+	<svelte:fragment slot="label">{$i18n.fee.text.fee}</svelte:fragment>
+
+	{#if isNullish(utxosFee)}
+		<span class="mt-2 block w-full max-w-[140px]"><SkeletonText /></span>
+	{:else}
+		{formatToken({
+			value: BigNumber.from(utxosFee.feeSatoshis),
+			unitName: BTC_DECIMALS,
```
peterpeterparker, line 31: shouldn't we use the `sendTokenDecimals` here?

src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
DenysKarmazynDFINITY: Sure, done.

src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
DenysKarmazynDFINITY: Good point. We can only keep the value in the wizard (it will be needed for the send step), but the fetching logic can be moved to the child component. It's now done like this.

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
DenysKarmazynDFINITY: Moved the logic to `BtcUtxosFee`.

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
DenysKarmazynDFINITY: Indeed, updated.

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
```diff
@@ -0,0 +1,74 @@
...
+	});
+</script>
+
+<Value ref="utxos-fee" element="div">
+	<svelte:fragment slot="label">{$i18n.fee.text.fee}</svelte:fragment>
+
+	{#if isNullish(utxosFee)}
+		<span class="mt-2 block w-full max-w-[140px]"><SkeletonText /></span>
+	{:else}
+		{formatToken({
```
peterpeterparker, line 67: maybe add an animation here, like an `in:fade` to smoothen the display?

peterpeterparker (approved):
LGTm, thx

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
DenysKarmazynDFINITY: Good idea, added.

PR #2823, opened 2024-10-11
feat(frontend): Disable if amount higher than balance

lmuntaner:
# Motivation

We want to give fast feedback to the user if a transaction can't be made.

In this PR, compare the amount that the user wants to send to the current balance and show an error is the amount is larger.

This doesn't consider the fee because the fee is calculated based on the amount, the utxos and the current fee of the network in the backend.

# Changes

* Introduce a `customValidate` in `BtcSendAmount`.

# Tests

Tested locally.

### DIFF
--- a/src/frontend/src/btc/components/send/BtcSendAmount.svelte
+++ b/src/frontend/src/btc/components/send/BtcSendAmount.svelte
@@ -1,11 +1,14 @@
 <script lang="ts">
-	import { isNullish } from '@dfinity/utils';
+	import { isNullish, nonNullish } from '@dfinity/utils';
 	import { getContext } from 'svelte';
 	import type { BtcAmountAssertionError } from '$btc/types/btc-send';
 	import { SEND_CONTEXT_KEY, type SendContext } from '$icp-eth/stores/send.store';
 	import SendInputAmount from '$lib/components/send/SendInputAmount.svelte';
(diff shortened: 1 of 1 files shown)


lmuntaner:

src/frontend/src/btc/components/send/BtcSendAmount.svelte
```diff
@@ -1,10 +1,13 @@
 <script lang="ts">
-	import { isNullish } from '@dfinity/utils';
+	import { isNullish, nonNullish } from '@dfinity/utils';
+	import type { BigNumber } from 'alchemy-sdk';
 	import { getContext } from 'svelte';
 	import type { BtcAmountAssertionError } from '$btc/types/btc-send';
+	import { IcAmountAssertionError } from '$icp/types/ic-send';
```
DenysKarmazynDFINITY, line 6: Please use `BtcAmountAssertionError` here.

AntonioVentilii-DFINITY (approved):
Pending @DenysKarmazynDFINITY suggestion, the rest LGTM!

src/frontend/src/btc/components/send/BtcSendAmount.svelte
lmuntaner: Done!

PR #2825, opened 2024-10-11
feat(devops): Add VITE_BLOCKCHAIN_API_URL to staging deploy

lmuntaner:
# Motivation

We want to enable bitcoin in Staging and other environments but bitcoin requires an extra env var apart from the flags that was missing in the current env vars: `VITE_BLOCKCHAIN_API_URL`.

# Changes

* Add `VITE_BLOCKCHAIN_API_URL` in the GH workflow that deploys the frontend.

# Tests

We will test when this PR is merged.

### DIFF
--- a/.github/workflows/frontend-deploy-staging.yml
+++ b/.github/workflows/frontend-deploy-staging.yml
@@ -9,7 +9,7 @@ on:
       network:
         required: true
         type: choice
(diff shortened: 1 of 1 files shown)


lmuntaner:
@AntonioVentilii-DFINITY please review

PR #2829, opened 2024-10-11
fix(frontend): set Sandbox for Onramper iframe [GIX-3110]

AntonioVentilii-DFINITY:
# Motivation

The security review pointed out that we should use the [sandbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#sandbox) attribute for the iFrame of Onramper.

> Observation: Oisy Onramper Integration is loaded in an unsandboxed iframe.
> 
> Risk: If the Onramper app turns malicious, it could redirect the user from Oisy to a phishing page or initiate downloads.
> 
> Recommendation: Apply least privilege principle: Use the [sandbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#sandbox) attribute on the iframe with only the flags that are needed for functioning.


src/frontend/src/lib/components/onramper/OnramperWidget.svelte
```diff
@@ -71,4 +71,5 @@
 	height="660px"
 	width="100%"
 	allow="accelerometer; autoplay; camera; gyroscope; payment; microphone"
+	sandbox="allow-forms allow-popups allow-popups-to-escape-sandbox allow-same-origin allow-scripts"
```
peterpeterparker, line 74: In addition to sandbox itself, is that the list of privileges prodsec recommended?

AntonioVentilii-DFINITY: They said the ones that are needed for functioning, and those seems like the ones they need to me. I tagged @tmu0 in this PR too, so that he can review

peterpeterparker (approved):
LGTM; thx

Please merge when you get their feedback.

PR #2832, opened 2024-10-11
feat(frontend): send btc txId to BE after transaction send has succeeded

DenysKarmazynDFINITY:
# Motivation

We need to send btc txId to BE after transaction send has succeeded.

### DIFF
--- a/src/frontend/src/btc/services/btc-send.services.ts
+++ b/src/frontend/src/btc/services/btc-send.services.ts
@@ -59,6 +59,8 @@ export const sendBtc = async ({
 	// TODO: use txid returned by this method to register it as a pending transaction in BE
 	const { txid } = await send({ progress, utxosFee, network, identity, ...rest });
 
+	progress(ProgressStepsSendBtc.RELOAD);
+
 	await addPendingBtcTransaction({
 		identity,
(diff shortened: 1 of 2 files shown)


src/frontend/src/btc/services/btc-send.services.ts
```diff
@@ -46,9 +48,24 @@ export const selectUtxosFee = async ({
...
+	network,
+	source,
+	identity,
+	...rest
+}: SendBtcParams): Promise<void> => {
 	// TODO: use txid returned by this method to register it as a pending transaction in BE
-	await send({ progress, ...rest });
+	const { txid } = await send({ progress, utxosFee, network, identity, ...rest });
+
+	await addPendingBtcTransaction({
```
AntonioVentilii-DFINITY, line 62: shall this be before or after the `RELOAD` step? The previous step is `SEND`, that theoretically is finished at this point, right?

DenysKarmazynDFINITY: I think we can do it after RELOAD to avoid waiting for the call to be completed. Updated.

AntonioVentilii-DFINITY (approved):
LGTM!

PR #2833, opened 2024-10-11
feat(frontend): Use Network wallets for Onramper Widget

AntonioVentilii-DFINITY:
# Motivation

As Security review suggested, it is better to use the param `networkWallets` of Onramper instead of `wallets`, to reduce the link length and make it more scalable.

> Currently the same wallet addresses for each individual token are specified in the onramper wallets URL parameter, e.g.: https://buy.onramper.dev/?wallets=icp_icp:9dd073a8b088ce77f7e0362d8a3b51752f20c304eb019fc83feb4511d1fa914e,eth:0x0a12ef5b74541C7De3c36B9f0265902869741b6A,usdc_ethereum:0x0a12ef5b74541C7De3c36B9f0265902869741b6A,link_ethereum:0x0a12ef5b74541C7De3c36B9f0265902869741b6A,...  IIUC: if wallet addresses are the same for a network these could be specified in  networkWallets like this https://buy.onramper.dev/?wallets=icp_icp:9dd073a8b088ce77f7e0362d8a3b51752f20c304eb019fc83feb4511d1fa914e&networkWallets=ethereum:0x0a12ef5b74541C7De3c36B9f0265902869741b6A...  Doing this is highly recommended because otherwise you could run into issues with the URL length if more tokens get added over time

# Changes

- Create mapping util to link each `network` from Oisy to a `networkWallet` for Onramper (plus tests).
- Include `networkWallets` among the parameters to build the source link.
- Add the list of network (and mapping to address) to the Onramper Widget, removing the list of crypto wallets. 

# Note

I would like to leave the old utils and parameters there, just in case. But it can be discussed.

# Tests

Manual test in local replica worked as usual.

[GIX-3110]

[GIX-3110]: https://dfinity.atlassian.net/browse/GIX-3110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/frontend/src/lib/components/onramper/OnramperWidget.svelte
```diff
@@ -33,30 +34,29 @@
...
-		tokens: $tokens,
-		walletMap: {
-			bitcoin: $btcAddressMainnet,
-			ethereum: $ethAddress,
-			erc20: $ethAddress,
-			icrc: $icrcAccountIdentifierText,
-			icp: $icpAccountIdentifierText
-		}
+	let networkWallets: OnramperNetworkWallet[];
+	$: networkWallets = mapOnramperNetworkWallets({
```
peterpeterparker, line 38: The `icrcAccountIdentifierText` disappeared?

src/frontend/src/tests/lib/utils/onramper.utils.spec.ts
```diff
@@ -22,6 +35,10 @@ describe('buildOnramperLink', () => {
 				{ cryptoId: 'btc', wallet: 'bitcoin_wallet_address' },
 				{ cryptoId: 'icp', wallet: 'icp_wallet_address' }
 			],
+			networkWallets: [
```
peterpeterparker, line 38: Can we use real addresses for the mocks? That would be a bit more readable regarding what's expected for me at least.

peterpeterparker (approved):
LGTM, thx

PR #2835, opened 2024-10-11
refactor(frontend): Extract button that closes the modal

AntonioVentilii-DFINITY:
# Motivation

To avoid repetition of code and concentrate the button styling, we extract the Button that is always used to close any modal.

# Note

This button is used sometimes as `primary` and sometimes as `secondary`. In This PR I kept the same way, but I think we should agree on one definition of closing the modal. My personal opinion is that a "Close" button should be considered `secondary`.

### Right now

Screenshot 2024-10-11 at 17 14 58
Screenshot 2024-10-11 at 17 15 11

### Possible alternatives for both cases

Screenshot 2024-10-11 at 17 14 51
Screenshot 2024-10-11 at 17 15 22


peterpeterparker (approved):
Nice!

peterpeterparker:
Checks do not pass though...

AntonioVentilii-DFINITY:
Sorry, forgot to remove a prop...

PR #2836, opened 2024-10-11
refactor(frontend): Re-use Button in the code

AntonioVentilii-DFINITY:
# Motivation

There are a few instances in the code where we can re-use the `Button` component, if we add the optionality to include more classes (for example, margin).

There is no change in the style.


peterpeterparker (commented):
LGTM, thx

peterpeterparker (approved):

PR #2843, opened 2024-10-11
feat(frontend): Add SendContext to Ethereum fee context

AntonioVentilii-DFINITY:
# Motivation

The component `EthereumFeeContext` is used only in `IcSendTokenWizard`, that is always wrapped with the Send context in both its uses.

So, we make use of the Send Context inside the component, removing all the stores that fetch the token from the token store, and substitute them with the ones from Send context.

# Note

We don't change the logic, we just replace element with the same value from the store to the context. 

# Changes

- Add sub-stores `sendTokenAsIcToken` and `sendTokenWithFallbackAsIcToken` to the Send context, that are identical in logic to the respective global stores.
- Replace all the global stores in `EthereumFeeContext` with the ones from the context: one-to-one.

# Tests

No changes in local replica: the Ethereum fee was fetched regularly both on Send and on Convert.

Screenshot 2024-10-11 at 20 28 26
Screenshot 2024-10-11 at 20 28 54
Screenshot 2024-10-11 at 20 29 04
Screenshot 2024-10-11 at 20 29 11


peterpeterparker (commented):
The `send.store.ts` is used by ETH and BTC therefore I don't think it should be extended by anything related to IC.

On the contrary, it should be move to `lib`. Provided this in PR #2848.

AntonioVentilii-DFINITY:
@peterpeterparker I checked and in component `IcSendTokenWizard` we already call the `SEND_CONTEXT`

that is what you meant?

AntonioVentilii:
This PR is from my old account and it is not actual anymore.

PR #2844, opened 2024-10-12
build(backend): Set minimal Bash version

AntonioVentilii-DFINITY:
# Motivation

In script `build.signer.sh` we use the upper-case syntax `^^` that is available only starting from [Bash version 4.0](https://tldp.org/LDP/abs/html/bashver4.html)

```bash
asset_url="${asset^^}_URL"
asset_file="${asset^^}_FILE"
```

# Changes

- Define a minimum Bash version in `engines`.

# Tests


AntonioVentilii-DFINITY:
@bitdivine I actually need you help here: is this enough? Shall it be changed elsewhere or in another way?

package.json
```diff
@@ -109,6 +109,7 @@
 		"elliptic": "^6.5.7"
 	},
 	"engines": {
-		"node": "^20"
+		"node": "^20",
+		"bash": ">=4.0"
```
bitdivine, line 113: I'm not super familiar with this so I'll add notes on what I find out here:

* bash 4 [was released more than a decade ago](https://en.wikipedia.org/wiki/Bash_(Unix_shell)#Release_history):	2009-02-20
* The npm documentation mensions engines as a field for setting the node version, but I see no hint that other programming languages can be configured here as well: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines

In what context does this problem appear?

AntonioVentilii-DFINITY: Basically, I had an outdated version of bash and I ran `npm run build:` nested inside there was this snippet that gave an issue, since my version does not support the syntax

```bash
asset_url="${asset^^}_URL"
asset_file="${asset^^}_FILE"
```

The idea is to guarantee that bash is at least upgraded to an useful version

AntonioVentilii-DFINITY:
As discussed offline with @bitdivine , this solution will not help (not effective), so closing

PR #2848, opened 2024-10-14
refactor(frontend): move send.store to lib

peterpeterparker:
# Motivation

The `send.store` module is now used by BTC as well. Therefore we move it from package `icp-eth` to `lib`.


AntonioVentilii-DFINITY (approved):
LGTM! Tks

PR #2849, opened 2024-10-14
build(frontend): bump eslint-config-oisy-wallet and require-await

peterpeterparker:
# Motivation

We want to use the last eslint rules which introduces `require-await`.

# Changes

- Bump eslint-config-oisy-wallet
- Adapt all error related to the eslint rule `require-await` (few ignored but most either resolved or turn into not promises)


src/frontend/src/routes/sitemap.xml/+server.ts
```diff
@@ -4,6 +4,7 @@ const staticPages: string[] = [];
 
 export const prerender = true;
 
+// eslint-disable-next-line require-await
```
AntonioVentilii-DFINITY, line 7: I did't quite understand this exception, why is it needed? because entire function is not awaitable?

src/frontend/src/lib/components/send/SendModal.svelte
```diff
@@ -60,6 +60,7 @@
 			}
 		}
 
+		// eslint-disable-next-line require-await
```
AntonioVentilii-DFINITY, line 63: why shall this be async?

src/frontend/src/lib/services/auth.services.ts
```diff
@@ -104,10 +104,11 @@ const emptyIdbAddress = async (deleteIdbAddress: (principal: Principal) => Promi
 	}
 };
 
-const emptyIdbBtcAddressMainnet = async () => emptyIdbAddress(deleteIdbBtcAddressMainnet);
+const emptyIdbBtcAddressMainnet = (): Promise<void> => emptyIdbAddress(deleteIdbBtcAddressMainnet);
 
-const emptyIdbEthAddress = async () => emptyIdbAddress(deleteIdbEthAddress);
+const emptyIdbEthAddress = (): Promise<void> => emptyIdbAddress(deleteIdbEthAddress);
 
+// eslint-disable-next-line require-await
```
AntonioVentilii-DFINITY, line 111: same question here

AntonioVentilii-DFINITY (approved):
Just a few doubts about exceptions, that you probalby already, checked, so just for my info

The rest LGTM! Thank you for the big fix

src/frontend/src/routes/sitemap.xml/+server.ts
peterpeterparker: Because there is nothing to await within the function I think but given that's not really important here, I mean that's the root that generate the sitemap, I spared myself to dig deeper 😅. Works for you?

src/frontend/src/lib/components/send/SendModal.svelte
peterpeterparker: Because `loadTokenAndRun` requires the callback to be async. Few functions needs it to load data for example opening a modal that calls `autoLoadUserToken`.

src/frontend/src/lib/services/auth.services.ts
peterpeterparker: For simplicity reason. This function, `clearTestnetsOption`, execution is grouped with `emptyIdbBtcAddressMainnet` and `emptyIdbEthAddress`  in an existing `await Promise.all`. The two "idb" functions have to be promised.

peterpeterparker:
@AntonioVentilii-DFINITY I have merged the PR because this is not the kind of PR that should remains open for a long time given the number of changes. If anything does not fit you regarding above answers, let me know and we resolve this in separate PRs.

src/frontend/src/routes/sitemap.xml/+server.ts
AntonioVentilii-DFINITY: yes yes, it was really a doubt, I was not understanding, thank you for the explanation!

src/frontend/src/lib/services/auth.services.ts
AntonioVentilii-DFINITY: right! thank you!

src/frontend/src/lib/components/send/SendModal.svelte
AntonioVentilii-DFINITY: ah that's true! thank you!

PR #2862, opened 2024-10-14
feat(frontend): disable send BTC button for userAmount <= 0

DenysKarmazynDFINITY:
# Motivation

The goal is to add additional checks to validate the provided user amount in the BTC send flow.


AntonioVentilii-DFINITY (approved):
LGTM!

P.S. this is checked during the `send` function too, I see; we leave it there too as redundancy?

DenysKarmazynDFINITY:
@AntonioVentilii-DFINITY yes, all values are double-checked on the `send` step, even though we make sure that they are all correct even before that.

AntonioVentilii-DFINITY (approved):

PR #2866, opened 2024-10-14
feat(frontend): take fees into account while calculating BTC tx value

DenysKarmazynDFINITY:
# Motivation

The previous implementation didn't take utxos fee into account while calculating a transaction value for "send" txs. In this PR, I update the implementation in order to get complete numbers. The mapper is implemented the way that it loops through `inputs` and `out` just once to get all required vars.

Previous values:
Screenshot 2024-10-14 at 12 20 13

Updated values:
Screenshot 2024-10-14 at 12 20 18

Note: please ignore the UI, it will be tackled separately.

### DIFF
--- a/src/frontend/src/btc/utils/btc-transactions.utils.ts
+++ b/src/frontend/src/btc/utils/btc-transactions.utils.ts
@@ -4,7 +4,7 @@ import {
 } from '$btc/constants/btc.constants';
 import type { BtcTransactionUi } from '$btc/types/btc';
 import type { BtcAddress } from '$lib/types/address';
-import type { BitcoinOutput, BitcoinTransaction } from '$lib/types/blockchain';
+import type { BitcoinTransaction } from '$lib/types/blockchain';
 import { isNullish, nonNullish } from '@dfinity/utils';
 
 export const mapBtcTransaction = ({
@@ -16,10 +16,50 @@ export const mapBtcTransaction = ({
 	btcAddress: BtcAddress;
 	latestBitcoinBlockHeight: number;
 }): BtcTransactionUi => {
-	const isTypeSend = inputs.some(({ prev_out }) => prev_out.addr === btcAddress);
-	const output: BitcoinOutput | undefined = out.find(({ addr }) =>
-		isTypeSend ? addr !== btcAddress : addr === btcAddress
+	const { totalInputValue, isTypeSend } = inputs.reduce<{
+		totalInputValue: number;
+		isTypeSend: boolean;
+	}>(
+		(acc, { prev_out: { value, addr } }) => ({
+			totalInputValue: acc.totalInputValue + value,
+			isTypeSend: !acc.isTypeSend ? addr === btcAddress : acc.isTypeSend
+		}),
+		{
+			totalInputValue: 0,
+			isTypeSend: false
+		}
 	);
+
+	const { totalOutputValue, value, to } = out.reduce<{
+		totalOutputValue: number;
+		value: number | undefined;
+		to: string | undefined;
+	}>(
+		(acc, { addr, value }) => {
+			const isValidOutput =
+				(isTypeSend && addr !== btcAddress) || (!isTypeSend && addr === btcAddress);
+
+			if (isNullish(acc.value) && isValidOutput) {
+				acc.value = (acc.value ?? 0) + value;
+			}
+
+			if (isNullish(acc.to) && isValidOutput) {
+				acc.to = addr;
+			}
+
+			return {
+				...acc,
+				totalOutputValue: acc.totalOutputValue + value
+			};
+		},
+		{
+			totalOutputValue: 0,
+			value: undefined,
+			to: undefined
+		}
+	);
+
+	const utxosFee = totalInputValue - totalOutputValue;
 
 	const confirmations = nonNullish(block_index)
 		? latestBitcoinBlockHeight - block_index
@@ -35,11 +75,11 @@ export const mapBtcTransaction = ({
 	return {
 		id: hash,
 		timestamp: BigInt(time),
-		value: nonNullish(output?.value) ? BigInt(output.value) : undefined,
+		value: nonNullish(value) ? BigInt(isTypeSend ? value + utxosFee : value) : undefined,
 		status,
 		blockNumber: block_index ?? undefined,
 		type: isTypeSend ? 'send' : 'receive',
 		from: isTypeSend ? btcAddress : inputs[0].prev_out.addr,
-		to: nonNullish(output) ? (isTypeSend ? output.addr : btcAddress) : undefined
+		to
 	};
 };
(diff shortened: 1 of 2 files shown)


peterpeterparker:
Unrelated to the PR @DenysKarmazynDFINITY but to your screenshots, shouldn't the date on the UI not be displayed when "Pending..." is displayed?

DenysKarmazynDFINITY:
@peterpeterparker the plan is to display all BTC transactions in the list (unconfirmed, pending, confirmed).

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -16,11 +16,51 @@ export const mapBtcTransaction = ({
...
-	const isTypeSend = inputs.some(({ prev_out }) => prev_out.addr === btcAddress);
-	const output: BitcoinOutput | undefined = out.find(({ addr }) =>
-		isTypeSend ? addr !== btcAddress : addr === btcAddress
+	const { totalInputValue, isTypeSend } = inputs.reduce<{
+		totalInputValue: number;
+		isTypeSend: boolean;
+	}>(
+		(acc, { prev_out: { value, addr } }) => ({
+			totalInputValue: acc.totalInputValue + value,
+			isTypeSend: !acc.isTypeSend ? addr === btcAddress : acc.isTypeSend
```
AntonioVentilii-DFINITY, line 25: i know it is just a detail eheh, but maybe invert the condition for readability (for me)?

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -16,11 +16,51 @@ export const mapBtcTransaction = ({
 	btcAddress: BtcAddress;
 	latestBitcoinBlockHeight: number;
 }): BtcTransactionUi => {
-	const isTypeSend = inputs.some(({ prev_out }) => prev_out.addr === btcAddress);
-	const output: BitcoinOutput | undefined = out.find(({ addr }) =>
-		isTypeSend ? addr !== btcAddress : addr === btcAddress
+	const { totalInputValue, isTypeSend } = inputs.reduce<{
```
AntonioVentilii-DFINITY, line 19: My understanding is that a transaction is mapped here and if there is no address related to me (neither in `inputs` nor in `outputs`), we still run the loops. So, I just wanted to double confirm: I imagine that you are filtering the transactions before feed them here?

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -35,11 +75,11 @@ export const mapBtcTransaction = ({
...
 		id: hash,
 		timestamp: BigInt(time),
-		value: nonNullish(output?.value) ? BigInt(output.value) : undefined,
+		value: nonNullish(value) ? BigInt(isTypeSend ? value + utxosFee : value) : undefined,
 		status,
 		blockNumber: block_index ?? undefined,
 		type: isTypeSend ? 'send' : 'receive',
 		from: isTypeSend ? btcAddress : inputs[0].prev_out.addr,
-		to: nonNullish(output) ? (isTypeSend ? output.addr : btcAddress) : undefined
+		to
```
AntonioVentilii-DFINITY, line 83: same question here for `type`, `from` and `to`: i imagine that the transactions that arrive in this function are either send from me or receive to me, and not unrelated

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -16,11 +16,51 @@ export const mapBtcTransaction = ({
...
 	);
 
+	const { totalOutputValue, value, to } = out.reduce<{
+		totalOutputValue: number;
+		value: number | undefined;
+		to: string | undefined;
+	}>(
+		(acc, { addr, value }) => {
+			const isValidOutput =
+				(isTypeSend && addr !== btcAddress) || (!isTypeSend && addr === btcAddress);
```
AntonioVentilii-DFINITY, line 40: I was wondering in which case it would appear nullish `value`: what happens if, for example, I stupidly send BTC to myself? what will be `value`? My understanding from the code is that it stays `undefined` because there will be no valid output. If yes, is that on purpose?

P.S. same question for `to` address

AntonioVentilii-DFINITY (commented):
I few observation, nothing big, mostly doubts that I have.

But if I can give a deeper feedback: it would be very useful to have some sort of documentation in the code, some explanation for example on the reason why using `inputs` and `outputs`, why there may be bitcoin sent to my address by me, even if I sent to someone else (very low level, I mean, some very basic eplxanation).  

I am no expert at all of BTC transactions, but I was a bit aware of inputs/outputs/UTXOS, and it helped me a lot to understand the code better.  
However, I am imagining that not all devs that will encounter the code will be expert in the matter, and theoretically we want to facilitate their life, I imagine

WDYT?

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: We don't need additional filters here since we are fetching transactions for a specific address (endpoint: `{API_URL}/rawaddr/${btcAddress}`). In case there were no transactions related to an address, txs will be an empty array, hence this method won't be called at all.

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: correct, it will be a transaction related to your current BTC address

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: Good question, I think we'll need to test this kind of corner cases, because I'm not sure how the Blockchain API response will look like in the provided example. Not sure how it's done for other tokens, but maybe it makes sense to disallow sending to the current address?

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: Sure 🙂

DenysKarmazynDFINITY:
@AntonioVentilii-DFINITY totally agree re. adding documentation. Considering we're a bit in a rush, would it work for you if I create a JIRA ticket for that and tackle it separately? For now, I added a TODO comment.

src/frontend/src/btc/utils/btc-transactions.utils.ts
AntonioVentilii-DFINITY: I would not disallow completely, maybe just a warning... I could be used as test, to send to itself, or something similar.

But about the case, maybe try to send to your self something as a test? Just to check? And if you don't want to tackle it in this one, add a TODO?

DenysKarmazynDFINITY: @AntonioVentilii-DFINITY you're right, value is undefined when transaction's sender and receiver addresses are the same:

Screenshot 2024-10-15 at 11 31 14

Since we already have TODO in the code, I suggest we handle this corner case separately. WDYT?

AntonioVentilii-DFINITY: For me all good! Hopefully people don't have BTC to waste like that eheh

PR #2867, opened 2024-10-14
feat(frontend): dapps mvp 1.0

loki344:
# Motivation

Implement a basic data structure / routing / layout for dapps page. NOTE: This implementation will change when implementing the final figma design

# Changes

Add components for Dapps

# Tests

no tests added because this is not the final implementation


loki344:
Will split up in smaller PR's

PR #2868, opened 2024-10-14
build(backend): remove wallet option from dfx script

AntonioVentilii-DFINITY:
# Motivation

It seems like the --wallet option for deploy with dfx is not really needed: the canisters will have enough cycles to spend. We tested with staging and it worked. So, we remove it


bitdivine (approved):
Thanks!

PR #2872, opened 2024-10-14
refactor(frontend): move Alpha warning inside the Hero when signed in

AntonioVentilii-DFINITY:
# Motivation

Since we are going to change the general layout of the elements when signed in, we need to move the Alpha warning to a place where it will not be affected by that.

Screenshot 2024-10-14 at 16 31 03
Screenshot 2024-10-14 at 16 31 15
Screenshot 2024-10-14 at 16 31 22


peterpeterparker (approved):
LGTM, thx

PR #2873, opened 2024-10-14
feat(frontend): BTC transaction status based on current block height

DenysKarmazynDFINITY:
# Motivation

In this PR, I fetch the current BTC block height to determine whether a transaction is confirmed, unconfirmed or pending. This will be used in:

1. Transaction list - an item will have a status displayed, plus pending txs will have "disabled" background.
2. The number of confirmations will be displayed in the tx detail view.

The UI PRs will be done separately.

### DIFF
--- a/src/frontend/src/btc/utils/btc-transactions.utils.ts
+++ b/src/frontend/src/btc/utils/btc-transactions.utils.ts
@@ -26,7 +26,7 @@ export const mapBtcTransaction = ({
 		: undefined;
 
 	const status =
-		isNullish(confirmations) || confirmations === PENDING_BTC_TRANSACTION_MIN_CONFIRMATIONS
+		isNullish(confirmations) || confirmations <= PENDING_BTC_TRANSACTION_MIN_CONFIRMATIONS
 			? 'pending'
 			: confirmations >= CONFIRMED_BTC_TRANSACTION_MIN_CONFIRMATIONS
 				? 'confirmed'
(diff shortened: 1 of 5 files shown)


src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -1,26 +1,41 @@
...
 	);
 
+	const confirmations = nonNullish(block_index)
+		? latestBitcoinBlockHeight - block_index
+		: undefined;
+
+	// 0 confirmations (or undefined) - status "pending"
+	// 1 - 5 confirmations - status "unconfirmed"
+	// 6 and more confirmations - status "confirmed"
+	const status =
```
AntonioVentilii-DFINITY, line 27: maybe these can be hardcoded as constants somewhere? it will be easier to manage, and can be re-used in the tests

DenysKarmazynDFINITY: Sounds good, updated.

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -1,26 +1,42 @@
...
 	const output: BitcoinOutput | undefined = out.find(({ addr }) =>
 		isTypeSend ? addr !== btcAddress : addr === btcAddress
 	);
 
+	const confirmations = nonNullish(block_index)
+		? latestBitcoinBlockHeight - block_index
+		: undefined;
+
+	const status =
+		isNullish(confirmations) || confirmations === PENDING_BTC_TRANSACTION_MIN_CONFIRMATIONS
```
AntonioVentilii-DFINITY, line 29: I think here should be <= right? i know we have 0 for now, but just in case

AntonioVentilii-DFINITY (approved):
LGTM!

Quick question below

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: Good point.

PR #2876, opened 2024-10-14
style(frontend): adjust layout to have sticky header footer and menu [GIX-3046]

AntonioVentilii-DFINITY:
# Motivation

We aim to leave the Header, the Footer and the Navigational Menu sticky for medium and larger screens. However, the main content should be able to freely scroll in the space occupied by the Header and the Footer, when it not overlaps (basically larger screens). Meaning that the class `sticky` is not apt to be used in that case.

So, we use `fixed` for all three, but being careful that it should applied only when signed in (the landing page is ok as it is).

# Results

https://github.com/user-attachments/assets/88c2375c-6d27-4a91-8725-70aed20181ab

# Landing page unchanged

https://github.com/user-attachments/assets/a098e72d-c7c0-43fe-9572-298dd5734afa


AntonioVentilii-DFINITY:
@peterpeterparker just to avoid something that i could not see, may you please check it live on test env number 4? or in a local replica too, obviously

src/frontend/src/lib/components/auth/AuthGuard.svelte
```diff
@@ -7,7 +7,7 @@
 {#if $authNotSignedIn}
 	<LandingPage />
 {:else}
-	<div in:fade>
+	<div in:fade class="sm:block sm:overflow-y-scroll">
```
peterpeterparker, line 10: I would use `overflow-y-auto` instead of scroll. That way the scrollbar on Mac is only displayed if necessary.

peterpeterparker (approved):
LGTM, thanks

Please have a look at my comment before merge.

On another note, do you plan to review the "Developed with love..." in another PR? On few edge cases  it is not gracefully displayed according your screenshots video.

AntonioVentilii-DFINITY:
@peterpeterparker RE "Developed with love..." yes, exactly. I was handling here too, but it was a bit too much maybe and out-of-scope, so I reduced to the minium.

But I will tackle in another PR

src/frontend/src/lib/components/auth/AuthGuard.svelte
AntonioVentilii-DFINITY: changed!

PR #2878, opened 2024-10-14
feat(frontend): enable networks switcher only for when testnets are visible [GIX-3046]

AntonioVentilii-DFINITY:
# Motivation

We change the main layout so that the Networks Switcher is visible ONLY when the toggle for testnet is turned on.

Since OISY is a multi-chain wallet by it's nature, we can remove the network selector from the home screen in a normal situation. It may be added a possible filter in the future, but not so bluntly, because the user should be driven to have all the tokens from multiple chains in one place.

### Note

The toggle for Testnet networks will be available only in Settings after this PR

# Results

Screenshot 2024-10-10 at 18 00 10
Screenshot 2024-10-10 at 18 00 24


src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
```diff
@@ -47,9 +46,8 @@
 			{/each}
 		</ul>
 
-		<div class="mb-4 mt-8 flex items-center justify-between">
-			<span class="px-4.5 font-bold">{$i18n.networks.show_testnets}</span>
-			<NetworksTestnetsToggle />
+		<div class="mb-5 mt-8 flex items-center justify-between">
```
peterpeterparker, line 49: Not sure what is this change? The list of testenet should still be present in the network switcher when displayed.

src/frontend/src/lib/components/tokens/Tokens.svelte
```diff
@@ -7,11 +7,17 @@
...
+	import { i18n } from '$lib/stores/i18n.store';
 </script>
 
 <div class:pointer-events-none={$authNotSignedIn} class:blur-[1.5px]={$authNotSignedIn}>
 	<Header>
-		<NetworksSwitcher disabled={$authNotSignedIn} />
+		{#if $testnetsEnabled}
+			<NetworksSwitcher disabled={$authNotSignedIn} />
+		{:else}
+			<h2 class="text-base">{$i18n.tokens.text.title}</h2>
```
peterpeterparker, line 19: Spacingwise, those that h2 occupies as much height as the closed `NetworkSwitcher`?

src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
AntonioVentilii-DFINITY: yes yes, it is, this specific part just removes the toggle from the list, when the switcher is displayed. Is that the question?

src/frontend/src/lib/components/tokens/Tokens.svelte
AntonioVentilii-DFINITY: not really, but, they occupy the same as the title in the "Transactions" page, so that we have a smooth transition there.

While the switcher will appear ONLY when the user is in the Settings page, so there will be no transition needed, since the user will never see it

src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
peterpeterparker: Ahhhhh just the toggle. Gotcha.

- Do we still need the component `NetworksTestnetsToggle`?

- Do we sill need the above div `<div class="mb-5 mt-8 flex...` given that only a text will now be displayed?

src/frontend/src/lib/components/tokens/Tokens.svelte
peterpeterparker: Alright, thanks for the feedback.

src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
AntonioVentilii-DFINITY: - Yes, we use it in Settings page the same one

- You are right, no, merged

peterpeterparker: Great! Thanks

peterpeterparker (approved):
LGTM, thx

PR #2880, opened 2024-10-14
style(frontend): background color of Hero for each network [GIX-3047]

AntonioVentilii-DFINITY:
# Motivation

The background color of the Hero will change according to the network selected (or token selected).

https://github.com/user-attachments/assets/08b83a7e-0ca6-49f0-8d5b-2a97f384ddc0


src/frontend/src/lib/components/hero/Hero.svelte
```diff
@@ -16,5 +16,6 @@
 	<div class="py-4">
 		<Alpha />
 	</div>
+
```
peterpeterparker, line 19: Format?

peterpeterparker (approved):
Nice, thanks!

src/frontend/src/lib/components/hero/Hero.svelte
AntonioVentilii-DFINITY: yep! but strangely it was my IDE and not prettier... but for prettier (and the CIs) it was acceptable. I will have to take a look into it at some point

PR #2886, opened 2024-10-15
feat(devops): Add new backend canister for beta

bitdivine:
# Motivation
As discussed in the standup, we would like to have a dedicated backend for beta.

# Changes
- Add a canister for the beta backend, instead of using prod.

Note:
- I have added controllers to match staging.
- I have added the new canister to cycleops.
- We will need to deploy the backend & frontend the next time we deploy to beta

# Tests


AntonioVentilii-DFINITY (approved):
Thank you!!!

PR #2889, opened 2024-10-15
feat(frontend): blockstream API

DenysKarmazynDFINITY:
# Motivation

Unfortunately, we cannot get both transactions and latest block data from Blockchain API since they have a strict API rate limit (1 request per 10 seconds). Therefore, we need to use another provider to get the height of the latest mined BTC block. Llorenc already sent Blockstream API for security review, we will merge it only after it's approved. The code can be already reviewed, though.

Additional info from Slack:
```
At the moment, this (blockstream API) will be used to complement the information we get from [blockchain.com](http://blockchain.com/).
We want to show the number of confirmations for each transaction. To do that, we need to compare the block of each transaction with the current block.
The current block will come from [blockstram.info](http://blockstream.info/) and the transaction information from [blockchain.com](http://blockchain.com/).
In the future, we might want to do something as you say about a double checkmark or compare transactions. But this won’t be in the first iteration.
```


lmuntaner:
> LGTM, thx
> 
> That said my review is closed to a rubber stamp this time.
> 
> Unrelated but related: Can we potentially remove blockchain.info to just use blockstream.info ?

Probably we could, yes. But according to security, we should use two endpoints to fetch the transaction and compare the results before showing them to the user.

peterpeterparker:
@lmuntaner gotcha, thanks for the explanation. That wasn't clear for me reading the description of the PR I have to say.

PR #2890, opened 2024-10-15
fix(build): Put Onramper CSP policy in frames [GIX-3024]

AntonioVentilii-DFINITY:
# Motivation

In PR #2827, I forgot to put the Onramper Frame CSP policy in `frame-src`.


peterpeterparker (approved):
LGTM, thx.

I feel like I lose a piece of my soul every time we have to make an exception for an iframe in a CSP.

PR #2892, opened 2024-10-15
ci(deploy): Add concurrency for jobs

AntonioVentilii-DFINITY:
# Motivation

When we commit sequentially, we want to be sure that only no commit is before the previous one. So, we make it so only one job at the time run, to guarantee that each commit has its chance to be deployed, but not before the previous ones.


bitdivine (approved):
Good idea!

PR #2895, opened 2024-10-15
feat(frontend): "unconfirmed" BTC transaction UI

DenysKarmazynDFINITY:
# Motivation

In this PR, "unconfirmed" transaction label is added to the `Transaction` component. The intermediate design as shown on the screenshot was agreed with Artem. 

Note: after ETH and IC Transaction components are refactored to use `lib/Transaction`, `TransactionPending` can be deleted in favour of `TransactionStatus`.

Screenshot 2024-10-15 at 13 46 50

### DIFF
--- a/src/frontend/src/btc/components/transactions/BtcTransaction.svelte
+++ b/src/frontend/src/btc/components/transactions/BtcTransaction.svelte
@@ -17,6 +17,9 @@
 
 	let pending: boolean;
 	$: pending = status === 'pending';
+
+	let unconfirmed: boolean;
+	$: unconfirmed = status === 'unconfirmed';
 </script>
 
 <Transaction
@@ -25,4 +28,5 @@
 	{type}
 	{timestamp}
 	{pending}
+	{unconfirmed}
 />
--- a/src/frontend/src/btc/components/transactions/BtcTransaction.svelte
+++ b/src/frontend/src/btc/components/transactions/BtcTransaction.svelte
@@ -14,12 +14,6 @@
 	let type: TransactionType;
 
 	$: ({ type, status, value, timestamp } = transaction);
-
-	let pending: boolean;
-	$: pending = status === 'pending';
-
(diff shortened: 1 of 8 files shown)


src/frontend/src/btc/components/transactions/BtcTransaction.svelte
```diff
@@ -25,4 +28,5 @@
 	{type}
 	{timestamp}
 	{pending}
+	{unconfirmed}
```
peterpeterparker, line 31: Can it be pending and unconfirmed at the same time? If not, would it be safer to pass the status here?

peterpeterparker (commented):
Not really a technical question, maybe none of my business but, are we sure we want to display both a date and status next to each other?

DenysKarmazynDFINITY:
@peterpeterparker very valid question and, luckily, it's planned to update the design of the list items, here is an example of how it's gonna look:
image

src/frontend/src/btc/components/transactions/BtcTransaction.svelte
DenysKarmazynDFINITY: True, we can just pass `status`. I did it like this now, it lead to a small refactoring of the types, please take a look.

src/frontend/src/lib/components/transactions/TransactionStatus.svelte
```diff
@@ -0,0 +1,18 @@
...
+
+	let pending: boolean;
+	$: pending = status === 'pending';
+
+	let unconfirmed: boolean;
+	$: unconfirmed = status === 'unconfirmed';
+</script>
+
+{#if pending || unconfirmed}
+	<span class="ml-2 text-goldenrod">
```
peterpeterparker, line 15: Can this be displayed with a delay? Should we add an `in:fade`?

peterpeterparker (approved):
LGTm, thx

PR #2896, opened 2024-10-15
feat(frontend): handle BTC latest block height endpoint error

DenysKarmazynDFINITY:
# Motivation

Similar as with fetching BTC txs data, we don't want worker to fail if the latest block height cannot be fetched. If this value is not available, it also means that we cannot properly parse transactions response. For now, I added a simple check, but in the future we might want to handle this case more gracefully (e.g. we still display txs but mention to users that we cannot identify confirmations number and hence tx status).

### DIFF
--- a/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
+++ b/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
@@ -13,7 +13,7 @@ import type {
 	PostMessageDataResponseWallet
 } from '$lib/types/post-message';
 import type { CertifiedData } from '$lib/types/store';
-import { assertNonNullish, isNullish, jsonReplacer } from '@dfinity/utils';
+import { assertNonNullish, isNullish, jsonReplacer, nonNullish } from '@dfinity/utils';
 
 interface BtcWalletStore {
 	balance: CertifiedData<bigint> | undefined;
@@ -47,11 +47,10 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> 
 		});
 	}
 
-	private async loadBtcTransactions({
-		btcAddress
-	}: {
-		btcAddress: BtcAddress;
-	}): Promise<BitcoinTransaction[]> {
+	private async loadBtcTransactionsData({ btcAddress }: { btcAddress: BtcAddress }): Promise<{
+		newTransactions: BitcoinTransaction[];
+		latestBitcoinBlockHeight: number | undefined;
+	}> {
 		try {
 			const { txs: fetchedTransactions } = await btcAddressData({ btcAddress });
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -123,16 +124,18 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
 
 		const balance = await this.loadBtcBalance({ identity, bitcoinNetwork });
-		const newTransactions = data?.shouldFetchTransactions
-			? await this.loadBtcTransactions({ btcAddress })
+		const { newTransactions, latestBitcoinBlockHeight } = data?.shouldFetchTransactions
+			? await this.loadBtcTransactionsData({ btcAddress })
+			: { newTransactions: [], latestBitcoinBlockHeight: undefined };
+
+		// TODO: handle the case when tx data is available but latestBitcoinBlockHeight is undefined
+		const uncertifiedTransactions = nonNullish(latestBitcoinBlockHeight)
```
AntonioVentilii-DFINITY, line 132: should it still post a message is there are no new transactions? meaning, does the balance change if there are no new transactions?

same question above where you check for `shouldFetchTransactions`: if it shouldn't fetch transactions, do you still want to post a message?

DenysKarmazynDFINITY: 1. Since we are fetching balance from one source (IC) and txs from the other one (blockchain API), there is a chance that they not 100% synced (hence balance can get an update even with no new transactions). Therefore, it makes sense to post a message (at least, for now), and try to solve this race condition problem globally. 
2. `shouldFetchTransactions` is `true` only for mainnet. It means that we have to post a message for testnet and regtest even if we don't have any transactions at all.

AntonioVentilii-DFINITY: thank you for the explanation!

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -123,16 +124,18 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
 		assertNonNullish(btcAddress, 'No BTC address provided to get BTC transactions.');
 
 		const balance = await this.loadBtcBalance({ identity, bitcoinNetwork });
-		const newTransactions = data?.shouldFetchTransactions
-			? await this.loadBtcTransactions({ btcAddress })
+		const { newTransactions, latestBitcoinBlockHeight } = data?.shouldFetchTransactions
```
AntonioVentilii-DFINITY, line 127: small thing: would you mind using `nonNullish(data) && data.shouldFetchTransactions` here please? Since we use it like that

AntonioVentilii-DFINITY (approved):
LGTM tks!

Small request please, have a look

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: Of course!

PR #2898, opened 2024-10-15
feat(frontend): Add short OISY name

AntonioVentilii-DFINITY:
# Motivation

Including OISY short name among the placeholders for `i18n`. It is useful to have it, in fact we already use it.


peterpeterparker (approved):
LGTM, thx

PR #2899, opened 2024-10-15
feat(frontend): Add info menu at bottom of navigational menu

AntonioVentilii-DFINITY:
# Motivation

We want to add some informative elements at the bottom of the navigational Menu.

# Changes

- Adjust lateral menu column to extend until bottom of the viewport.
- Split the menu into two elements to put distance automatically between them: the top part is the navigation menu, the bottom part will be mostly informative and hyperlinks.
- Use the About Modal buttons as elements to put in the "info" part.
- Adjust the buttons to have a shorter string there and update the icon (using Lucide as per design).

# Tests

https://github.com/user-attachments/assets/73e2c0df-cffe-46e7-a2c7-4b4ba9137868

### DIFF
--- a/src/frontend/src/lib/components/hero/about/AboutHow.svelte
+++ b/src/frontend/src/lib/components/hero/about/AboutHow.svelte
@@ -7,6 +7,7 @@
 	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
 
 	export let asMenuItem = false;
+	export let shortLabel = false;
 
 	const dispatch = createEventDispatcher();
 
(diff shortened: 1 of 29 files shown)


src/frontend/src/lib/components/hero/about/AboutHow.svelte
```diff
@@ -17,5 +17,5 @@
 </script>
 
 <AboutItem {asMenuItem} on:click={openModal} testId={ABOUT_HOW_MODAL_OPEN_BUTTON}>
-	<span slot="label">{replaceOisyPlaceholders($i18n.about.how.text.label)}</span>
+	<span slot="label">{replaceOisyPlaceholders($i18n.about.how.text.label)} </span>
```
peterpeterparker, line 20: This space was added on purpose?

peterpeterparker (approved):
LGTm, thx

src/frontend/src/lib/components/hero/about/AboutHow.svelte
AntonioVentilii-DFINITY: nope, removed! thank you for noticing!

PR #2900, opened 2024-10-15
ci(e2e): cache build for snapshots

inc-man:
# Motivation

Run time of E2E Tests was too long. 

# Changes

added Caching of npm and cargo build jobs to the action.

# Tests

ran the improved E2E-Test on branch with a runtime of 6min40s instead of 9min40s


.github/actions/deploy-backend/action.yml
```diff
@@ -7,9 +7,31 @@ description: Build and deploy the backend after having installed dfx
...
   steps:
+  
     - name: Checkout
       uses: actions/checkout@v4
 
+    - name: Restore npm cache
+      uses: actions/cache@v4
+      with:
+        path: ~/.npm
+        key: ${{ runner.os }}-npm-${{ hashFiles('package-lock.json') }}
```
AntonioVentilii-DFINITY, line 18: may you please instead of `npm` put `e2e-npm`

.github/actions/deploy-backend/action.yml
```diff
@@ -7,9 +7,31 @@ description: Build and deploy the backend after having installed dfx
...
+      with:
+        path: |
+          /home/runner/.cargo/registry
+          /home/runner/.cargo/git
+          target/
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+        key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
```
AntonioVentilii-DFINITY, line 31: and here, instead of `cargo` put `e2e-cargo`

.github/actions/deploy-backend/action.yml
```diff
@@ -7,9 +7,28 @@ description: Build and deploy the backend after having installed dfx
...
   steps:
+
     - name: Checkout
       uses: actions/checkout@v4
 
+    - name: Restore npm cache
+      uses: actions/cache@v4
+      with:
+        path: ~/.npm
+        key: ${{ runner.os }}-e2e-npm-${{ hashFiles('package-lock.json') }}
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: btw, shouldn't the hash be including `package.json` too?

.github/actions/deploy-backend/action.yml
```diff
@@ -7,9 +7,28 @@ description: Build and deploy the backend after having installed dfx
...
+
+    - name: Restore cargo cache
+      uses: actions/cache@v4
+      with:
+        path: |
+          /home/runner/.cargo/registry
+          /home/runner/.cargo/git
+          target/
+          ~/.cargo/bin/
+        key: ${{ runner.os }}-e2e-cargo-${{ hashFiles('Cargo.lock') }}
```
AntonioVentilii-DFINITY, line 28: shouldn't the has be including all the rest too?

`hashFiles('Cargo.toml', 'Cargo.lock', 'rust-toolchain.toml', 'src/backend/**/*', 'src/shared/**/*')`

inc-man: as of my knowledge the lock files handle all the dependencies, so if either in npm or cargo builds the dependencies change, we only have to check the lock file.

we can add the other files too, but this could result in more cache files. This could lead to potential issues due to the cache size.

AntonioVentilii-DFINITY: for `cargo`, I think we should consider all the changes related to backend: example, we change the code in backend, but cargo stays the same, or if we change some workspace in cargo.toml, etc etc

.github/actions/deploy-backend/action.yml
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: better to include it, anyway, just to be sure, there are instances where the package.json changes but not the lock, so just to be on the safe side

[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Before merging, may you please add `package.json` here?

AntonioVentilii-DFINITY (approved):
LGTM! Just maybe check last request before merging

PR #2901, opened 2024-10-15
feat(frontend): use blockstream API to get latest block height

DenysKarmazynDFINITY:
# Motivation

Use blockstream API to get latest block height.


peterpeterparker (approved):
LGTM, thx

PS: I guess we delete  `btcLatestBlock` in another PR?

PR #2903, opened 2024-10-15
fix(frontend): adjust BTC tx confirmations calculation

DenysKarmazynDFINITY:
# Motivation

I realised I used incomplete formula to calculate BTC tx confirmations, it should be `confirmations = currentBlockHeight - transactionBlockHeight + 1`. Without adding +1, we didn't account for the block where the transaction was first included, which resulted in 0 confirmations immediately after the block is mined, which isn't correct.

### DIFF
--- a/src/frontend/src/btc/utils/btc-transactions.utils.ts
+++ b/src/frontend/src/btc/utils/btc-transactions.utils.ts
@@ -64,7 +64,7 @@ export const mapBtcTransaction = ({
 	const utxosFee = totalInputValue - totalOutputValue;
 
 	const confirmations = nonNullish(block_index)
-		? latestBitcoinBlockHeight - block_index
+		? latestBitcoinBlockHeight - block_index + 1
 		: undefined;
 
 	const status =
--- a/src/frontend/src/btc/utils/btc-transactions.utils.ts
+++ b/src/frontend/src/btc/utils/btc-transactions.utils.ts
@@ -1,7 +1,4 @@
-import {
-	CONFIRMED_BTC_TRANSACTION_MIN_CONFIRMATIONS,
-	PENDING_BTC_TRANSACTION_MIN_CONFIRMATIONS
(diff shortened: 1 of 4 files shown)


src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -63,16 +60,16 @@ export const mapBtcTransaction = ({
 
 	const utxosFee = totalInputValue - totalOutputValue;
 
+	// +1 is needed to account for the block where the transaction was first included
 	const confirmations = nonNullish(block_index)
-		? latestBitcoinBlockHeight - block_index
+		? latestBitcoinBlockHeight - block_index + 1
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: I updated the existing tests and mocks, `confirmations` value is now either a number that >= 1 or `undefined`. Please let me know in case you have some additional test cases in mind.

src/frontend/src/tests/btc/utils/btc-transactions.utils.spec.ts
```diff
@@ -31,6 +31,7 @@ describe('mapBtcTransaction', () => {
 		const expectedResult = {
 			...mockBtcTransactionUi,
 			blockNumber: undefined,
+			confirmations: undefined,
```
peterpeterparker, line 34: can it be `null`?  asking because the comment say

> // no confirmations - transaction status "pending"

peterpeterparker (approved):
LGTM, thx

Just asked another question if interesting before merge.

src/frontend/src/tests/btc/utils/btc-transactions.utils.spec.ts
DenysKarmazynDFINITY: It can only be either `undefined` or a number. I rephrased the comment a bit.

PR #2904, opened 2024-10-15
feat(frontend): extend BtcTransactionUi type with "confirmations"

DenysKarmazynDFINITY:
# Motivation

We need to extend BtcTransactionUi type with "confirmations" in order to display it in the UI.


peterpeterparker (approved):
LGTm, thx

PR #2907, opened 2024-10-15
fix(frontend): Wrong flex for buttons in modal

AntonioVentilii-DFINITY:
# Motivation

For mobiles, the buttons in the modal were stretching too much, because the were not wrapped into a `flex`.

### Before

Screenshot 2024-10-15 at 17 35 51

### After

Screenshot 2024-10-15 at 17 36 09


peterpeterparker (approved):
LGTM, thx

PR #2909, opened 2024-10-15
feat(frontend): update BTC transaction modal

DenysKarmazynDFINITY:
# Motivation

The goal of this PR is to update BTC transaction modal with the following data:
1. Properly parse and display BTC tx status (previously, "unconfirmed" was missing).
2. Display number of confirmations if it's available.

Screenshot 2024-10-15 at 18 01 35

### DIFF
--- a/src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
+++ b/src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
@@ -59,12 +59,14 @@
 	sendToLabel={$i18n.transaction.text.to}
 	typeLabel={type === 'send' ? $i18n.send.text.send : $i18n.receive.text.receive}
 >
-	{#if nonNullish(confirmations)}
-		<Value ref="confirmations" slot="transaction-confirmations">
-			<svelte:fragment slot="label">{$i18n.transaction.text.confirmations}</svelte:fragment>
-			{confirmations}
-		</Value>
-	{/if}
+	<svelte:fragment slot="transaction-confirmations">
+		{#if nonNullish(confirmations)}
+			<Value ref="confirmations">
+				<svelte:fragment slot="label">{$i18n.transaction.text.confirmations}</svelte:fragment>
+				{confirmations}
+			</Value>
+		{/if}
+	</svelte:fragment>
 
 	<Value ref="status" slot="transaction-status">
 		<svelte:fragment slot="label">{$i18n.transaction.text.status}</svelte:fragment>
--- a/src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
+++ b/src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
@@ -19,6 +19,7 @@
 	let timestamp: bigint | undefined;
 	let id: string;
 	let blockNumber: number | undefined;
+	let confirmations: number | undefined;
 	let status: BtcTransactionStatus;
 
 	let explorerUrl: string | undefined;
@@ -30,7 +31,7 @@
 				: BTC_MAINNET_EXPLORER_URL;
 	}
 
-	$: ({ from, value, timestamp, id, blockNumber, to, type, status } = transaction);
+	$: ({ from, value, timestamp, id, blockNumber, to, type, status, confirmations } = transaction);
 
 	let txExplorerUrl: string | undefined;
 	$: txExplorerUrl = nonNullish(explorerUrl) ? `${explorerUrl}/tx/${id}` : undefined;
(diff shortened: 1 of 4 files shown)


src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
```diff
@@ -58,9 +59,21 @@
 	sendToLabel={$i18n.transaction.text.to}
 	typeLabel={type === 'send' ? $i18n.send.text.send : $i18n.receive.text.receive}
 >
-	<!--	TODO: Implement BtcTransactionStatus component	-->
+	{#if nonNullish(confirmations)}
+		<Value ref="confirmations" slot="confirmations">
```
peterpeterparker, line 63: This compiles? Shouldn't the `slot=confirmations` be provided with a `svelte:fragement` and the `if nonNullish(confirmation` find place within the fragment?

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
```diff
@@ -58,9 +59,21 @@
...
+			<svelte:fragment slot="label">{$i18n.transaction.text.confirmations}</svelte:fragment>
+			{confirmations}
+		</Value>
+	{/if}
+
 	<Value ref="status" slot="transaction-status">
 		<svelte:fragment slot="label">{$i18n.transaction.text.status}</svelte:fragment>
-		{`${status === 'pending' ? $i18n.transaction.text.pending : 'Confirmed'}`}
+		{`${
+			status === 'pending'
```
peterpeterparker, line 72: It's me or we already had this piece of code? Worth extracting a component or funtion for it?

src/frontend/src/lib/components/transactions/TransactionModal.svelte
```diff
@@ -67,6 +67,8 @@
 
 		<slot name="transaction-status" />
 
+		<slot name="confirmations" />
```
peterpeterparker, line 70: We have `transaction-status` but `confirmation`, should it be `transaction-confirmations` or should we rename the other or that's not really related?

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
DenysKarmazynDFINITY: Yes, it works fine, it's the way how it's used in other components too. I think it would be required to do like you said if there wasn't a wrapper component (`Value` in this case).

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
DenysKarmazynDFINITY: We have a similar component that only displays either "Unconfirmed" or "Pending" (`TransactionStatus`), plus adds additional styling to the label (which should not be applied in `BtcTransactionModal`). In case of the modal, we also want to display the third option ("Confirmed"). 

Having said that, I think for now we can keep it as it is since I can't foresee another use-case where all 3 labels are needed. But I'm also fine with creating a component. WDYT?

src/frontend/src/lib/components/transactions/TransactionModal.svelte
DenysKarmazynDFINITY: True, renamed it to `transaction-confirmations`.

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
peterpeterparker: TIL I did not knew that. Really thought the linter would complain about that. Coolio, thx for letting me know.

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
peterpeterparker: Maybe not in the PR but, yeah it feels like it would deserve a bit of cleanup, agree?

peterpeterparker (approved):
LGTm, thx

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
DenysKarmazynDFINITY: Sorry, my bad - I checked it once again and it does have to be done like you said. Wondering how it worked for me the first time. Even though Svelte is a really nice instrument, I still find some of its concepts a bit confusing 🙃

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
DenysKarmazynDFINITY: Sounds good, agreed.

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
peterpeterparker: Wait for it, in Svelte v5 they removed slots 🤪

PR #2911, opened 2024-10-16
style(frontend): buttons success and error

peterpeterparker:
# Motivation

In the signer flow we need green and red buttons.

# Changes

- Add `.success` and `.error` button global classes

# Screenshot

Capture d’écran 2024-10-16 à 09 32 41

PR #2912, opened 2024-10-16
style(frontend): replace color mountain-meadow with green-crayola

peterpeterparker:
# Motivation

The success button provided in #2911 uses color `green-crayola` therefore for consistency this PR replaces `mountain-meadow` use to set the GIX component color.

# Notes

Note sure exactly where this style is used but nevertheless and given the visual difference between the two colors, it felt accurate to provide this PR.

# Changes

- Remove `mountain-meadow`
- Use `green-crayola` for gix


AntonioVentilii-DFINITY (approved):
Tks!

PR #2913, opened 2024-10-16
style(frontend): Update ExtrenalLink hover state color [OISY-62]

AntonioVentilii-DFINITY:
# Motivation

The component `ExternalLink` should be updated with the new color pallets when hovered/active.

### DIFF
--- a/src/frontend/src/lib/components/ui/ExternalLink.svelte
+++ b/src/frontend/src/lib/components/ui/ExternalLink.svelte
@@ -20,6 +20,8 @@
 	class:text-primary={color === 'blue'}
 	class:hover:text-inherit={color === 'blue'}
 	class:active:text-inherit={color === 'blue'}
+	class:hover:text-blue-ribbon={color === 'inherit'}
+	class:active:text-blue-ribbon={color === 'inherit'}
 	class:w-full={fullWidth}
 >
 	{#if iconVisible}


src/frontend/src/lib/components/ui/ExternalLink.svelte
```diff
@@ -20,6 +20,8 @@
 	class:text-primary={color === 'blue'}
 	class:hover:text-inherit={color === 'blue'}
 	class:active:text-inherit={color === 'blue'}
+	class:hover:text-blue-ribbon={color === 'inherit'}
```
peterpeterparker, line 23: it's a bit weird no? Blue is inherit but inherit is text-blue-ribbon.

Should we rename the value 'inherit'?

AntonioVentilii-DFINITY: well, those are the `hover`/`active` classes right? The idea is that `blue`/`inherit` refers to the main color of the text in the normal state

peterpeterparker: I would honestly expect `class:hover:text-inherit={color === 'inherit'}`

peterpeterparker: Ah now I understand. Yeah it's a bit confusing.

peterpeterparker: The color is the idle state and the class is the active state.

peterpeterparker (approved):
LGTM, thx

PR #2914, opened 2024-10-16
style(frontend): Set different hover color for tertiary link button

AntonioVentilii-DFINITY:
# Motivation

The button class `tertiary link` will have a new color when hovered and pressed.


peterpeterparker (approved):
LGTm, thx

PR #2917, opened 2024-10-16
feat(frontend): sort BTC transactions

DenysKarmazynDFINITY:
# Motivation

Sort BTC transactions with the following logic:
1. Pending status + timestamp.
2. Unconfirmed status + timestamp.
3. Confirmed status + timestamp.

### DIFF
--- a/src/frontend/src/btc/utils/btc-transactions.utils.ts
+++ b/src/frontend/src/btc/utils/btc-transactions.utils.ts
@@ -83,3 +83,38 @@ export const mapBtcTransaction = ({
 		confirmations
 	};
 };
+
+export const sortBtcTransactions = ({
+	transactionA: { status: statusA, timestamp: timestampA },
+	transactionB: { status: statusB, timestamp: timestampB }
+}: {
+	transactionA: BtcTransactionUi;
+	transactionB: BtcTransactionUi;
+}): number => {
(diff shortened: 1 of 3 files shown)


src/frontend/src/tests/btc/utils/btc-transactions.utils.spec.ts
AntonioVentilii-DFINITY, line 1: Ideally, the tests will check each case, for example, what if they all have the same timestamp, but different states, or same state and different timestamp

May just add a TODO, since we are close with time eheh

src/frontend/src/btc/utils/btc-transactions.utils.ts
AntonioVentilii-DFINITY, line 1: nitpick: not a call to action, but what about address and/or amount as last sorting criteria when everything else is the same?

AntonioVentilii-DFINITY (approved):
LGTM! Thank you!

Two no-call-to-action items, so already approving and you decide on it

src/frontend/src/tests/btc/utils/btc-transactions.utils.spec.ts
DenysKarmazynDFINITY: Added a TODO comment 🙂

PR #2918, opened 2024-10-16
style(frontend): body background

peterpeterparker:
# Motivation

There should be no glitch given that the background is embeded in the html file but, just had this idea that we can maybe actually set the background color on the `:root` element.

# Changes

- Move background color from background container to root.
- Remove `:root` specified as transparent background and black color. The first will be specifically set in the HTML page, the second is the default.

# Screenshots

Before if I delete the background image:

Capture d’écran 2024-10-16 à 11 35 14

After if I remove the bg image:

Capture d’écran 2024-10-16 à 11 35 01

### DIFF
--- a/src/frontend/src/app.html
+++ b/src/frontend/src/app.html
@@ -58,6 +58,10 @@
 		<!-- ROUTE_STYLE -->
 
 		<style>
+			body {
+				background: #e8f1ff;
+			}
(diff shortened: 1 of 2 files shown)


AntonioVentilii-DFINITY (approved):
Thank you!

PR #2920, opened 2024-10-16
refactor(frontend): hide partially the footer DFINITY string [OISY-62]

AntonioVentilii-DFINITY:
# Motivation

The idea is to hide the "Developed with..." string for middle screens, leaving just the DFINITY logo. However, for narrower middle screen, hide the logo too.

# Changes

- Make the entire string be an hyperlink.
- Modify the footer so that it is hidden on narrow middle screens, and only a logo on middle screens.
- The rest is unchanged.

# Tests

https://github.com/user-attachments/assets/72d9e1b8-65c0-41f8-a0fc-c75f8886d0b5

### DIFF
--- a/src/frontend/src/lib/components/core/Footer.svelte
+++ b/src/frontend/src/lib/components/core/Footer.svelte
@@ -5,7 +5,7 @@
 	import ExternalLink from '$lib/components/ui/ExternalLink.svelte';
 	import ExternalLinkIcon from '$lib/components/ui/ExternalLinkIcon.svelte';
 	import { OISY_REPO_URL } from '$lib/constants/oisy.constants';
-	import { authNotSignedIn } from '$lib/derived/auth.derived';
+	import { authNotSignedIn, authSignedIn } from '$lib/derived/auth.derived';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
 
@@ -19,6 +19,9 @@
 	class:md:h-md:grid={$authNotSignedIn}
 	class:md:h-md:grid-cols-4={$authNotSignedIn}
 	class:md:h-md:pr-0={$authNotSignedIn}
+	class:sm:fixed={$authSignedIn}
+	class:sm:inset-x-0={$authSignedIn}
+	class:sm:bottom-0={$authSignedIn}
 >
 	<div class="flex flex-row items-center gap-4">
 		<ExternalLinkIcon href={OISY_REPO_URL} ariaLabel={$i18n.navigation.text.source_code_on_github}>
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/core/Footer.svelte
```diff
@@ -34,20 +37,24 @@
...
-				href="https://dfinity.org"
-				ariaLabel={replaceOisyPlaceholders($i18n.footer.alt.dfinity)}
-				color="blue"
-				iconVisible={false}
-			>
-				{replaceOisyPlaceholders($i18n.footer.text.dfinity)}
-			</ExternalLink>
-		</span>
+		<ExternalLink
+			href="https://dfinity.org"
```
peterpeterparker, line 48: Wait what, I missed that in a previous review. It's a link to DFINITY, not the Internet Computer?!!??!? That's correct and expected?

src/frontend/src/lib/components/core/Footer.svelte
```diff
@@ -34,20 +37,24 @@
...
+			ariaLabel={replaceOisyPlaceholders($i18n.footer.alt.dfinity)}
+			iconVisible={false}
+		>
+			<div class="flex flex-row items-center gap-2">
+				<IconDfinity />
```
peterpeterparker, line 53: I missed that too, did I review the previous PR 😅. It's `IconDfinity` not `IconInternetComputer` or something?

peterpeterparker (approved):
LGTM, thanks

Two questions unrelated to this PR, therefore approving the code.

src/frontend/src/lib/components/core/Footer.svelte
AntonioVentilii-DFINITY: I am pretty sure... I will double confirm, but the string says `Developed with... by DFINITY`, so I would imagine that is where it should point. But good question, I will ask

src/frontend/src/lib/components/core/Footer.svelte
AntonioVentilii-DFINITY: eheh same as above... i will re-check

src/frontend/src/lib/components/core/Footer.svelte
peterpeterparker: Please do. Thank in advance.

AntonioVentilii-DFINITY: Just as follow up: I checked with both PM and Design team and that is indeed on purpose all DFINITY related.

src/frontend/src/lib/components/core/Footer.svelte
AntonioVentilii-DFINITY: Just as follow up: I checked with both PM and Design team and that is indeed on purpose all DFINITY related.

src/frontend/src/lib/components/core/Footer.svelte
peterpeterparker: Thanks for double checking.

PR #2922, opened 2024-10-16
fix(frontend): reduce the size of action button in QR modal

AntonioVentilii-DFINITY:
# Motivation

For mobile, the button ion the QR modal was expanding too much. To solve it we re-use component `ContentWithToolbar`.

### Before

Screenshot 2024-10-16 at 12 47 50

#### After

Screenshot 2024-10-16 at 12 47 42


peterpeterparker (approved):
LGTm, thx

A good use case for an E2E test btw.

PR #2924, opened 2024-10-16
style(frontend): Rearrange the Hero layout [OISY-62]

AntonioVentilii-DFINITY:
# Motivation

A few changes in the Hero layout regarding the Token Logo:

- Add black&white network icon to the `TokenLogo`
- Re-arrange the logo to be in the same row of the back button and the context menu
- General adjustments in the spacing.

# Results

A small video where i slowed down the animation

Screenshot 2024-10-16 at 13 04 20

https://github.com/user-attachments/assets/aaa93d23-1a96-4bc7-a80b-9dccf7f898b9

# Rest is Unchanged

The other instance where `TokenLogo` is used with the network icon is only one and it is unchanged:

Screenshot 2024-10-16 at 12 59 12


src/frontend/src/lib/components/hero/HeroContent.svelte
```diff
@@ -36,34 +34,28 @@
...
 >
 	{#if summary}
-		<div
-			transition:slide={SLIDE_PARAMS}
-			class="grid w-full grid-cols-[1fr_auto_1fr] flex-row items-start justify-between"
-		>
-			{#if back}
-				<Back color="current" />
-			{/if}
+		<div transition:slide={SLIDE_PARAMS} class="flex w-full flex-col gap-5">
```
peterpeterparker, line 37: Not for this PR, just thinking out at loud and opiniated but, maybe we should spend a bit of time trying other animations. From your video it feels like slide is maybe not the best animation anymore because of the paddings.

peterpeterparker (approved):
LGTm, thx

src/frontend/src/lib/components/hero/HeroContent.svelte
AntonioVentilii-DFINITY: right! good point! putting it among my backlog and will talk with Stefan too

peterpeterparker: I would just do and to not make a big story of it. Feels trivial. 

"Does the animation looks good? Yes/no. Better like this? Yes. Let's go."

AntonioVentilii-DFINITY: This was solved by:

- PR #3221

PR #2925, opened 2024-10-16
feat(frontend): integrate RC (⚠️) version of the signer library

peterpeterparker:
# Motivation

The OISY Wallet Signer library has undergone a security review. A few security points remain open, but we agreed (as confirmed by @samuelburri) that we can proceed with its integration into the wallet.

The library is about to be open-sourced, but it is not yet public, and therefore its code cannot be published to npm.

As it is in our best interest to move forward, this PR adds the dependency as a static package.

# Notes

- Although the code is not yet public, the goal is to open-source the library. Moreover, the library will be released under the Apache-2.0 license. While including the library means also exposing its bundled code on the web, I don't foresee anyone opposing it, particularly given that the foundation wants to advance this initiative.

- The library was packaged on my machine using the following steps: `npm ci && npm run build && cd dist && npm pack` in the project repository of the library.

- Commit reference from which the library was build: [58fe36cc838476337af315fa9027958f3d0f82a8](https://github.com/dfinity/oisy-wallet-signer/commit/58fe36cc838476337af315fa9027958f3d0f82a8)

# Changes

- Copy `dfinity-oisy-wallet-signer-0.0.1-rc.1.tgz` at root of the project (library `v0.0.1-rc.1`)
- `npm i dfinity-oisy-wallet-signer-0.0.1-rc.1.tgz`


AntonioVentilii-DFINITY (approved):
LGTM!

I imagine no `TODO` needed, since you will notice that we need to remove the bundle when you will install the open library in `package.json`

peterpeterparker:
I would not know where to pub a TODO anyway 🤷‍♂️. I force sure keep track of this! Thanks for the review.

PR #2927, opened 2024-10-16
feat(frontend): initiate new /sign route

peterpeterparker:
## Motivation

The signer feature of OISY will be available under a sub-route `/sign`, e.g., `https://oisy.com/sign`. This URL was decided by the product and design teams (Mary and Artem). This PR aims to initiate this route without any specific capabilities.

## Notes

These features do not require any of the core OISY Wallet functionalities, such as fetching USD, Ethereum, or Bitcoin transactions from third-party Web2 APIs. Similarly, they do not need ETH or BTC keys derived from the Crosschain Signer canister. The reason behind this is that the signer acts primarily as a proxy for messages. It receives requests, asserts and prompts the user for interaction, and propagates those requests if approved—nothing more, nothing less.

For this reason, this PR creates a new group in the routing, `(sign)`, which does not use the `LoadersGuard`.

It also does not use the `AuthGuard` because it is still custom and contains another call to action for sign-in.

## Security Notes

Since a few points are still being discussed with the security team and the library is not yet open source, the UI will include a warning banner to inform users about the ongoing work. I will keep track of this and provide a PR to remove the banner once everything is resolved.

# Changes

- Init new route `src/frontend/src/routes(sign)` with placeholders

# Screenshot

Capture d’écran 2024-10-16 à 13 41 32

### DIFF
--- a/src/frontend/src/routes/(sign)/sign/+layout.svelte
+++ b/src/frontend/src/routes/(sign)/sign/+layout.svelte
@@ -1,0 +1,18 @@
+<script lang="ts">
+	import Modals from '$lib/components/core/Modals.svelte';
+	import OisyWalletLogoLink from '$lib/components/core/OisyWalletLogoLink.svelte';
+	import WarningBanner from '$lib/components/ui/WarningBanner.svelte';
+</script>
+
+<main class="mx-auto flex flex-col gap-y-6 px-5 pt-10 sm:w-sm">
+	<div class="flex justify-center"><OisyWalletLogoLink /></div>
+
+	<WarningBanner>
+		This feature is unreviewed. DO NOT USE until it has undergone proper code and security
+		assessment!
+	</WarningBanner>
+
+	<slot />
+</main>
+
+<Modals />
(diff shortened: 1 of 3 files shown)


AntonioVentilii-DFINITY (approved):
LGTM! Tks!

Note: I am assuming that the `(sign)` page will not use any of the stores set by the Loaders component in `(app)`, such as `tokenStore`. Or there is not risk of conflict if an user is logged in in Oisy and uses the same browser?

peterpeterparker:
> I am assuming that the (sign) page will not use any of the stores set by the Loaders component in (app), such as tokenStore. 

That's correct but, it's an excellent point. There is no security in place - such as an eslint rule - that some developer in the future incorrectly start using a store that is actually not loaded but, I assume that if they would do so, they would notice the store is empty.

> Or there is not risk of conflict if an user is logged in in Oisy and uses the same browser?

There is no conflicts. When user nagivates from root to `/sign` or the contrary, the the related `+page.ts` and `+layout` are executed so either the information are loaded or resetted.

PR #2928, opened 2024-10-16
feat(frontend): dapps data

loki344:
# Motivation

In order to display dapps in the explorer, we manage the data in a json structure.

# Changes

- add parsing of json
- add json data


src/frontend/src/env/dapp-descriptions.json
```diff
@@ -0,0 +1,45 @@
...
+		"oneLiner": "The one stop token shop - Trade tokens across all chains with ease using KongSwap.",
+		"website": "https://www.kongswap.io/",
+		"twitter": "https://x.com/kongswap",
+		"github": "https://github.com/KongSwap",
+		"tags": ["DEX", "Signer Standard"],
+		"description": "KongSwap is a fully decentralized exchange (DEX) offering bridgeless cross-chain trading across Bitcoin, Ethereum, and Solana. <br/><br/>Using Chain Fusion technology and the Internet Computer Protocol, it reduces fees, simplifies user experience, and avoids liquidity fragmentation. Both frontend and backend operate on-chain via smart contracts, delivering a seamless DeFi experience.",
+		"usesInternetIdentity": false,
+		"display": "Large",
+		"stats": "",
+		"logo": "/images/dapps/kong-swap-logo.svg",
```
peterpeterparker, line 14: Should we have a small job that checks that the assets mentionned in the JSON files are actually present at their location to avoid broken links at runtime?

peterpeterparker (approved):

src/frontend/src/env/dapp-descriptions.json
loki344: I let @Stefan-Berger-DFINITY decide if that is needed.

Stefan-Berger-DFINITY: Since this list is always referred to as "highly curated", I think we can have the MVP version of the dapps explorer w/o this additional check. But once we have more and more dapps listed, it might be worth to consider it.
I will make a planing ticket for now.

PR #2293, opened 2024-09-09
feat(frontend): signer support

peterpeterparker:
# Motivation

A branch I'll use as WIP to develop the integration of the OISY Wallet Signer library and signer features.

# Integration

This PR has been split into few PRs, notably:

- https://github.com/dfinity/oisy-wallet/pull/2968
- https://github.com/dfinity/oisy-wallet/pull/2965
- https://github.com/dfinity/oisy-wallet/pull/2964
- https://github.com/dfinity/oisy-wallet/pull/2953
- https://github.com/dfinity/oisy-wallet/pull/2955
- https://github.com/dfinity/oisy-wallet/pull/2960
- https://github.com/dfinity/oisy-wallet/pull/2957
- https://github.com/dfinity/oisy-wallet/pull/2956
- https://github.com/dfinity/oisy-wallet/pull/2952
- https://github.com/dfinity/oisy-wallet/pull/2950
- https://github.com/dfinity/oisy-wallet/pull/2947
- https://github.com/dfinity/oisy-wallet/pull/2946
- https://github.com/dfinity/oisy-wallet/pull/2945
- https://github.com/dfinity/oisy-wallet/pull/2940
- https://github.com/dfinity/oisy-wallet/pull/2939
- https://github.com/dfinity/oisy-wallet/pull/2938
- https://github.com/dfinity/oisy-wallet/pull/2937
- https://github.com/dfinity/oisy-wallet/pull/2934
- https://github.com/dfinity/oisy-wallet/pull/2927
- https://github.com/dfinity/oisy-wallet/pull/2925
- https://github.com/dfinity/oisy-wallet/pull/2912
- https://github.com/dfinity/oisy-wallet/pull/2911

PR #2934, opened 2024-10-16
feat(frontend): signer context

peterpeterparker:
# Motivation

To implement the OISY Wallet Signer features, a consumer must create a `Signer` object and implement various prompts.

In the OISY Wallet integration, I ended up using a context to handle all subscriptions instead of, for example, registering prompts at the component level. This approach felt cleaner and more aligned with our general architecture.

I also did not declare a global store, but rather a context. This way, we can potentially reuse the signer features more easily.

# Notes

This PR itself might not make much sense, but it is part of the branch split from #2293, which might provide more context.

# Changes

- Create `signer.store` context initializer and interfaces.

# Tests

The tests are minimal. I tried to implement more features, such as ensuring that the various prompts are triggered upon receiving Signer Standards window post messages, but ultimately, it felt like an overcomplication and a bit of a reinventing the wheel, particularly since the library itself is already well-covered by tests. What we want to assert here is that the components (in upcoming PRs) behave as they should, and this can be validated through E2E tests.


AntonioVentilii-DFINITY (approved):
LGTM! tks

PR #2935, opened 2024-10-16
style(frontend): Footer will be expanded when logged out

AntonioVentilii-DFINITY:
# Motivation

We want to expand the footer when logged out, so that the "Developed with" string expands when it has space. However, we still leave only the logo for narrower middle screen, until it becomes a mobile.

To do that, I wrapped the entire footer in a `<div>` that when signed out fits one of of the columns of the grid. 

# Results

https://github.com/user-attachments/assets/b0fee411-8d79-4007-85de-7bd98b640bea

https://github.com/user-attachments/assets/42c6c7f5-d7a2-4b01-9bf2-43e933892cac


AntonioVentilii-DFINITY:
[OISY-62]

peterpeterparker (approved):
LGTM, thx

Unsolicited opinion: the more I look at it the more I would remove it from the login page 😅

PR #2936, opened 2024-10-16
feat(frontend): Update about what modal to new design

loki344:
# Motivation

Adjust the about what modal to match the new Design.

# Changes

- Update cover image
- Update text
- Add IC icon
- Make button primary to match design

Before:
Bildschirmfoto 2024-10-16 um 16 23 35

Now:
Bildschirmfoto 2024-10-16 um 16 22 41


peterpeterparker (commented):
The `cover-features.png` is lighter, which is great. I noticed that its width and height are now twice as large. If we keep the same dimensions, can we reduce its file size even further? Or we need it to be larger for rendering reasons?

peterpeterparker:
Also just noticed that the spacing has changed. Should the new image keep the same proportion as the previous image or should the CSS be updated?

Capture d’écran 2024-10-16 à 16 43 10

src/frontend/src/lib/assets/cover-features.png
AntonioVentilii-DFINITY, line 1: I like it! Less kb! Magic!

src/frontend/src/lib/i18n/en.json
```diff
@@ -636,11 +636,11 @@
...
-				"use_ic_dapps": "<strong>Use with Internet Computer dapps (Coming soon)</strong><br>Use your $oisy_name wallet in any <a rel=\"noreferrer noopener\" target=\"_blank\" href=\"https://internetcomputer.org/ecosystem\">dapp on the Internet Computer</a> that supports the signer standards, including <a rel=\"noreferrer noopener\" target=\"_blank\" href=\"https://oc.app/\">OpenChat</a>, <a rel=\"noreferrer noopener\" target=\"_blank\" href=\"https://www.icpswap.com/\">ICPSwap</a>, and the <a rel=\"noreferrer noopener\" target=\"_blank\" href=\"https://nns.ic0.app/\">NNS dapp</a>."
+				"label": "Wallet Features",
+				"title": "$oisy_short Wallet Features",
+				"hold_crypto": "<strong>Hold crypto across networks</strong><br>Manage and transact with native Ethereum, Bitcoin, and Internet Computer assets. Powered by <a rel=\"noreferrer noopener\" target=\"_blank\" href=\"https://internetcomputer.org/chainfusion\">chain fusion</a>.",
```
AntonioVentilii-DFINITY, line 641: Shouldn't Chain Fusion be upper case on the first letter? I remember this was discussed in the past, but I am not sure what is the official denomination

loki344:
> The `cover-features.png` is lighter, which is great. I noticed that its width and height are now twice as large. If we keep the same dimensions, can we reduce its file size even further? Or we need it to be larger for rendering reasons?

Yes, was able to further reduce it.

src/frontend/src/lib/i18n/en.json
```diff
@@ -636,11 +636,11 @@
...
-				"use_eth_dapps": "<strong>Use with Ethereum dapps</strong><br>$oisy_name is fully compatible with Ethereum Defi like <a rel=\"noreferrer noopener\" target=\"_blank\" href=\"https://uniswap.org/\">Uniswap</a> thanks to the <a rel=\"noreferrer noopener\" target=\"_blank\" href=\"https://walletconnect.com/\">WalletConnect</a> framework.",
-				"use_ic_dapps": "<strong>Use with Internet Computer dapps (Coming soon)</strong><br>Use your $oisy_name wallet in any <a rel=\"noreferrer noopener\" target=\"_blank\" href=\"https://internetcomputer.org/ecosystem\">dapp on the Internet Computer</a> that supports the signer standards, including <a rel=\"noreferrer noopener\" target=\"_blank\" href=\"https://oc.app/\">OpenChat</a>, <a rel=\"noreferrer noopener\" target=\"_blank\" href=\"https://www.icpswap.com/\">ICPSwap</a>, and the <a rel=\"noreferrer noopener\" target=\"_blank\" href=\"https://nns.ic0.app/\">NNS dapp</a>."
+				"label": "Wallet Features",
+				"title": "$oisy_short Wallet Features",
```
AntonioVentilii-DFINITY, line 640: here you can use `$oisy_name` instead of `$oisy_short Wallet`

AntonioVentilii-DFINITY (commented):
I see that the label for the buttons has been changed: may you please provide some screenshot of the general landing page and the logged in (+menu) just to double check?

loki344:
> Also just noticed that the spacing has changed. Should the new image keep the same proportion as the previous image or should the CSS be updated?
> 
> Assuming that this is not expected given that no CSS is modified in your PR. Or course let me know if it is expected.
> 
> Capture d’écran 2024-10-16 à 16 43 10

Thanks for spotting this. The ImgBanner seems not to grow its image. By using the size small and increasing margin we can match the design.
Bildschirmfoto 2024-10-16 um 16 56 49

new:
Bildschirmfoto 2024-10-16 um 16 55 06

src/frontend/src/lib/i18n/en.json
loki344: TBH I matched figma because that's in the ticket. I don't know about discussions.
image

loki344:
> I see that the label for the buttons has been changed: may you please provide some screenshot of the general landing page and the logged in (+menu) just to double check?

Note that the icon and order not yet match figma (comes in a separate PR)

Bildschirmfoto 2024-10-16 um 17 01 23
Bildschirmfoto 2024-10-16 um 17 01 47
Bildschirmfoto 2024-10-16 um 17 02 27

Bildschirmfoto 2024-10-16 um 17 02 45

peterpeterparker:
> The ImgBanner seems not to grow its image.

On purpose. Without min height the UI becomes glitchy on mainnet due to the delay to load the image. Thanks for the adjustement.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/i18n/en.json
AntonioVentilii-DFINITY: maybe double-check with the design team, but I think it should be Chain Fusion because on the website is like that

https://internetcomputer.org/docs/current/developer-docs/multi-chain/overview

loki344: Thanks, done

PR #2944, opened 2024-10-16
feat(frontend): Update menu items to new design

loki344:
# Motivation

The Design changes the order of the items and adds a new icon. 

# Changes

- add star icon
- add slot with fallback to info icon for AboutItem
- use star icon for wallet features
- change order on landing page

Before:
Bildschirmfoto 2024-10-16 um 17 41 12
Bildschirmfoto 2024-10-16 um 17 41 20
Bildschirmfoto 2024-10-16 um 17 41 29
Bildschirmfoto 2024-10-16 um 17 41 36

New:
Bildschirmfoto 2024-10-16 um 17 39 22
Bildschirmfoto 2024-10-16 um 17 39 30
3044">
Bildschirmfoto 2024-10-16 um 17 40 07


src/frontend/src/lib/components/hero/about/AboutItem.svelte
```diff
@@ -12,7 +12,9 @@
 >
 	<div class="flex items-center gap-2">
 		{#if asMenuItem}
-			<IconInfo />
+			<slot name="icon">
+				<IconInfo />
```
AntonioVentilii-DFINITY, line 16: i would say to remove the `IconInfo` from here, leave the slot and add the `IconInfo` in the slot directly inside `AboutHow` too (similar to `AboutWhat`)

WDYT?

loki344: I agree. May I suggest a renaming of the component? Because this altered version would be better described by: `IconItem`. WDYT?

loki344: I meant: AboutItem => IconItem

AntonioVentilii-DFINITY: Maybe more `ButtonAbout` or `AboutButton` if you want to rename it, since there is both text icon, no?

AntonioVentilii-DFINITY:
@loki344 I think the `AboutWhat` is missing: right now it will have no icon. May you please add it as slot as discussed?

AntonioVentilii-DFINITY (approved):
LGTM, tks!

AntonioVentilii-DFINITY:
As discussed with @loki344 , I am going to merge in his stead, since I need a component from here

PR #2945, opened 2024-10-16
feat(frontend): signer accounts

peterpeterparker:
# Motivation

In the integration of the signer feature in OISY, the user's account will be automatically shared with the relying party if permission is granted. This is because we currently manage a single ICRC account per user, and while the library provides a prompt, we will resolve it automatically.

That's why this PR introduces a component that contains a prompt that resolves automatically.

# Notes

I could have resolved this feature differently than by using a component, but given that other prompts are also handled with components, it felt consistent to proceed this way.

Integrating this account prompt in OISY, while the component related to permissions is not yet reviewed, is not an issue. The library ensures that prompts are not triggered if permissions have not been granted.

# Changes

- Created a component to handle the prompt for the signer standard account. This component acts as a container since it requires no interaction.
- Integrated the component into the `/sign` route.


AntonioVentilii-DFINITY (approved):
LGTM! tks

PR #2946, opened 2024-10-16
feat(frontend): signer origin

peterpeterparker:
# Motivation

This component is useful to inform the user about the source - relying party - that is requesting anything through the signer.

# Notes

Given that the lib is strongly typed, it's unlikely that an invalid URL ends being processed in this new component.

# Changes

- A new component `SignerOrigin`

# Screenshot

Usage:

Capture d’écran 2024-10-16 à 17 54 15


src/frontend/src/lib/components/signer/SignerOrigin.svelte
```diff
@@ -0,0 +1,45 @@
...
+	// Null being used if mapping the origin does not work - i.e. invalid origin. Probably an edge case.
+	// eslint-disable-next-line local-rules/use-option-type-wrapper
+	let host: string | null | undefined;
+	$: host = mapHost(origin);
+</script>
+
+{#if nonNullish(origin)}
+	<p class="mb-6 break-normal text-center">
+		{$i18n.signer.origin.text.request_from}
+		{#if nonNullish(host)}<span class="font-bold text-brandeis-blue"
```
AntonioVentilii-DFINITY, line 36: is this the blue that was decided officially? because otherwise the prop color='blue' of `ExternalLink` should provide the other "official" blue

src/frontend/src/lib/components/signer/SignerOrigin.svelte
```diff
@@ -0,0 +1,45 @@
...
+			const { host } = new URL(origin);
+			return host;
+		} catch {
+			return null;
+		}
+	};
+
+	// Null being used if mapping the origin does not work - i.e. invalid origin. Probably an edge case.
+	// eslint-disable-next-line local-rules/use-option-type-wrapper
+	let host: string | null | undefined;
```
AntonioVentilii-DFINITY, line 29: why not using `OptionString` here? it is the same of the mapping function no?

src/frontend/src/lib/components/signer/SignerOrigin.svelte
```diff
@@ -0,0 +1,45 @@
...
+	import ExternalLink from '$lib/components/ui/ExternalLink.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import type { Option } from '$lib/types/utils';
+
+	export let payload: Option<PayloadOrigin>;
+
+	let origin: Origin | undefined;
+	$: origin = payload?.origin;
+
+	const mapHost = (origin: Origin | undefined): Option<string> => {
```
AntonioVentilii-DFINITY, line 13: there is an `OptionString` type already built, IIRC

src/frontend/src/lib/components/signer/SignerOrigin.svelte
peterpeterparker: Just implemented what was provided in the Figma file I got 🤷‍♂️

src/frontend/src/lib/components/signer/SignerOrigin.svelte
peterpeterparker: Great idea!

peterpeterparker: Done

src/frontend/src/lib/components/signer/SignerOrigin.svelte
peterpeterparker: `Origin | undefined` comes from the OISY Wallet Signer library. This is correct.

```
export const UrlSchema = z
  .string()
  .url()
  .refine(
    (url): boolean => {
      try {
        const {protocol, hostname} = new URL(url);

        // We allow http for development locally
        if (['localhost', '127.0.0.1'].includes(hostname)) {
          return ['http:', 'https:'].includes(protocol);
        }

        return protocol === 'https:';
      } catch (_err: unknown) {
        return false;
      }
    },
    {
      message: 'Invalid URL.'
    }
  );
  
export const OriginSchema = UrlSchema;

const PayloadOriginSchema = z.object({
  origin: OriginSchema
});
```

AntonioVentilii-DFINITY: What i meant is that the return is `Option<string>`, but we have a type already called `OptionString` that is that

peterpeterparker: Ahhhhhhhhhhhhhhhhh

peterpeterparker: Thx. Done!

AntonioVentilii-DFINITY (approved):
LGTM tks!

PR #2947, opened 2024-10-16
feat(frontend): signer loading

peterpeterparker:
# Motivation

When the signer is doing some work, like loading the consent message or effectivelly processing the call to a canister, we want to display a spinner with a message.

# Changes

- Add a component to present a loading msg.

# Screenshot

Capture d’écran 2024-10-16 à 17 58 45


AntonioVentilii-DFINITY (approved):
LGTM! tks

PR #2950, opened 2024-10-16
feat(frontend): signer idle

peterpeterparker:
# Motivation

When the signer is idle - i.e. awaiting messages and not processing anything - we want to display an idle related animation.

# Notes

This can notably happens if relying parties do not close the OISY Wallet window after interactions.

I was hesitant to place one the new animated component in icon folder but, ultimately felt like it was something related to the signer only for now. So provided it as a `Signer...` component as well.

# Changes

- New idle component and animation

# Screenshot

demo


AntonioVentilii-DFINITY (approved):
Wow!!!

LGTM! Tks

peterpeterparker:
Fun fact: Artem developed the animation!

PR #2952, opened 2024-10-16
feat(frontend): init and reset signer context on route sign

peterpeterparker:
# Motivation

At the top of the `/sign` route we want to initialize and reset the signer context.

# Changes

- Subscribe to auth and destroy hook and init or reset the context.


AntonioVentilii-DFINITY (approved):
LGTM! tks

PR #2953, opened 2024-10-16
feat(frontend): signer permissions

peterpeterparker:
# Motivation

The OISY Wallet Signer library handles requests and permission needs. On the OISY Wallet side, we need to manage prompt requests, present the information to the user, and provide them with a way to reject or approve the permissions.

# Notes

While the integration with the demo begins by requesting all permissions, according to the specification, individual permissions can be requested separately. That’s why the component can also be triggered independently.

The wallet address is section is rendered only if the related "accounts" permissions are requested.

# Changes

- New component `SignerPermissions` which uses the prompt and provide the call to action for the user
- Integrate the component in the `/sign` route

# Screenshot

Capture d’écran 2024-10-16 à 18 35 28


src/frontend/src/lib/components/signer/SignerPermissions.svelte
```diff
@@ -0,0 +1,136 @@
...
+	import { toastsError } from '$lib/stores/toasts.store';
+	import { shortenWithMiddleEllipsis } from '$lib/utils/format.utils';
+	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+
+	const {
+		permissionsPrompt: { payload, reset: resetPrompt }
+	} = getContext<SignerContext>(SIGNER_CONTEXT_KEY);
+
+	let scopes: IcrcScope[] | undefined;
+	$: scopes = $payload?.requestedScopes;
```
AntonioVentilii-DFINITY, line 29: I saw that in the script you are "coalescing" `scopes` when you use it (you never analyze the case where scopes is `nullish`, I mean). Would it make sense to do it directly here?

```
let scopes: IcrcScope[];
$: scopes = $payload?.requestedScopes ?? []
```

src/frontend/src/lib/components/signer/SignerPermissions.svelte
```diff
@@ -0,0 +1,136 @@
...
+		confirm((scopes ?? []).map((scope) => ({ ...scope, state: ICRC25_PERMISSION_GRANTED })));
+
+		resetPrompt();
+	};
+
+	const onReject = () => ignorePermissions();
+
+	const onApprove = () => approvePermissions();
+
+	const listItems: Record<
```
AntonioVentilii-DFINITY, line 69: `let listItems...` here, I think, since you are using `$i18n`?

src/frontend/src/lib/components/signer/SignerPermissions.svelte
```diff
@@ -0,0 +1,136 @@
...
+	};
+
+	let requestAccountsPermissions = false;
+	$: requestAccountsPermissions = nonNullish(
+		scopes?.find(({ scope: { method } }) => method === ICRC27_ACCOUNTS)
+	);
+</script>
+
+{#if nonNullish(scopes) && nonNullish($payload)}
+	<form in:fade on:submit|preventDefault={onApprove} method="POST">
```
AntonioVentilii-DFINITY, line 93: I remember you said once that we should be careful using `|preventDefault`: just for my info, why are we using it in this case?

src/frontend/src/lib/components/signer/SignerPermissions.svelte
```diff
@@ -0,0 +1,136 @@
...
+			label: replaceOisyPlaceholders($i18n.signer.permissions.text.icrc27_accounts)
+		},
+		icrc49_call_canister: {
+			icon: IconShield,
+			label: $i18n.signer.permissions.text.icrc49_call_canister
+		}
+	};
+
+	let requestAccountsPermissions = false;
+	$: requestAccountsPermissions = nonNullish(
```
AntonioVentilii-DFINITY, line 87: no need to add `scopes`, after `$:` here right?

`$: scopes, requestAccountsPermissions = nonNullish(....`

src/frontend/src/lib/components/signer/SignerPermissions.svelte
```diff
@@ -0,0 +1,136 @@
...
+
+					<output id="ic-wallet-address" class="break-all"
+						>{shortenWithMiddleEllipsis({ text: $icrcAccountIdentifierText ?? '' })}</output
+					>
+				</div>
+			</div>
+		{/if}
+
+		<ButtonGroup>
+			<button type="button" class="error block flex-1" on:click={onReject}
```
AntonioVentilii-DFINITY, line 130: isn't `block` the default? with or without is the same?

TBD for me: include this CSS class into `Button` component

src/frontend/src/lib/components/signer/SignerPermissions.svelte
```diff
@@ -0,0 +1,136 @@
...
+				>{$i18n.core.text.reject}</button
+			>
+			<button type="submit" class="success block flex-1">{$i18n.core.text.approve}</button>
```
AntonioVentilii-DFINITY, line 133: same question

src/frontend/src/lib/components/signer/SignerPermissions.svelte
peterpeterparker: Sure, done!

src/frontend/src/lib/components/signer/SignerPermissions.svelte
peterpeterparker: Absolutely, thx

src/frontend/src/lib/components/signer/SignerPermissions.svelte
peterpeterparker: Copy/paste. We don't need it I think indeed, thanks!

src/frontend/src/lib/components/signer/SignerPermissions.svelte
peterpeterparker: No need indeed. Because `scopes` is used in the evaluation.

src/frontend/src/lib/components/signer/SignerPermissions.svelte
peterpeterparker: same answer ;)

AntonioVentilii-DFINITY (approved):
LGTM! Tks

PR #2955, opened 2024-10-16
feat(frontend): signer consent message

peterpeterparker:
# Motivation

Before submitting any call to the IC, the OISY Wallet Signer library first fetches a consent message from the targeted canister (using the specified methods and arguments as payload) and prompts the user to decline or approve the call.

# Notes

I anticipate several iterations in the future regarding this. However, from a specification standpoint, there isn’t much more we can do in terms of formatting. The standards are generic for any canister on the IC, and therefore the response currently contains only unstructured information.

Approving the consent message triggers the library to effectively execute the call canister.

# Changes

- Create a new consent message related component which handles the prompt and offer the user options to decline or approve the message.
- Integrate the component in the `/sign` route

# Screenshot

Capture d’écran 2024-10-16 à 18 43 13


src/frontend/src/lib/components/signer/SignerConsentMessage.svelte
```diff
@@ -0,0 +1,153 @@
...
+	} = getContext<SignerContext>(SIGNER_CONTEXT_KEY);
+
+	let approve: ConsentMessageApproval | undefined;
+	$: approve = $payload?.status === 'result' ? $payload?.approve : undefined;
+
+	let reject: Rejection | undefined;
+	$: reject = $payload?.status === 'result' ? $payload?.reject : undefined;
+
+	let consentInfo: icrc21_consent_info | undefined;
+	$: consentInfo = $payload?.status === 'result' ? $payload?.consentInfo : undefined;
```
AntonioVentilii-DFINITY, line 30: can the three declarations above be merged?

`{ approve, reject, consentInfo } = $payload?.status...`

src/frontend/src/lib/components/signer/SignerConsentMessage.svelte
```diff
@@ -0,0 +1,153 @@
...
+
+		displayMessage =
+			nonNullish(consentInfo) && 'GenericDisplayMessage' in consentInfo.consent_message
+				? consentInfo.consent_message.GenericDisplayMessage
+				: undefined;
+	};
+
+	$: $payload, onPayload();
+
+	type Text = { title: string; content: string } | undefined;
```
AntonioVentilii-DFINITY, line 65: shouldn't this be among the `types` folder?

src/frontend/src/lib/components/signer/SignerConsentMessage.svelte
```diff
@@ -0,0 +1,153 @@
...
+		if (isNullish(reject)) {
+			toastsError({
+				msg: { text: $i18n.signer.consent_message.error.no_reject_callback }
+			});
+
+			resetPrompt();
+			return;
+		}
+
+		reject?.();
```
AntonioVentilii-DFINITY, line 108: the coalescing `?` is not needed here, right?

AntonioVentilii-DFINITY (approved):
Just some small observations, if useful, but nothing necessary, so already approving, tks!

src/frontend/src/lib/components/signer/SignerConsentMessage.svelte
peterpeterparker: Sure. Done.

```
let approve: ConsentMessageApproval | undefined;
	let reject: Rejection | undefined;
	let consentInfo: icrc21_consent_info | undefined;

	$: ({ approve, reject, consentInfo } =
		nonNullish($payload) && $payload.status === 'result'
			? $payload
			: { approve: undefined, reject: undefined, consentInfo: undefined });
```

src/frontend/src/lib/components/signer/SignerConsentMessage.svelte
peterpeterparker: It's used nowhere else and really specific for this function and component therefore I would just keep it here.

src/frontend/src/lib/components/signer/SignerConsentMessage.svelte
peterpeterparker: Absolutely right. Thanks.

PR #2956, opened 2024-10-16
feat(frontend): signer centered content component

peterpeterparker:
# Motivation

I realized there is few screen that are all centered in the signer flow. Therefore I extracted a component from the "loading" component that places the content at the center of the signer article.

# Changes

- Extract `SignerCenteredContent` component.


AntonioVentilii-DFINITY (approved):
LGTM!

peterpeterparker:
Generally I would say so but, it felt really specific for the signer because we got nowhere this "article in the center of the page that has that particular style". But in other PR I think it's a good idea if we notice we can use it elsewhere.

PR #2959, opened 2024-10-16
fix(frontend): Clickable header and footer when signed in

AntonioVentilii-DFINITY:
# Motivation

When signed in and in large screens, a part of the token list is overlapped by the Header and the Footer. That means that those part are un-clickable (for example Alpha and Manage list).

To solve it, we make the entire Header and Footer with disabled pointer events, enabling it ONLY for the elements inside it.


AntonioVentilii-DFINITY:
[OISY-62]

DenysKarmazynDFINITY (approved):
LGTM, thanks for fixing it!

PR #2963, opened 2024-10-16
feat(e2e): No need to set viewport when the device is a mobile

AntonioVentilii-DFINITY:
# Motivation

We will be adding mobile devices to the E2E tests, so in those cases there is no need to re-set the viewport, since we are interested in the default viewport of each one and how the screen would appear.


DenysKarmazynDFINITY (approved):
Makes sense, thanks

PR #2965, opened 2024-10-17
style(frontend): remove block class

peterpeterparker:
# Motivation

Remove class `block` in Signer buttons as discussed in https://github.com/dfinity/oisy-wallet/pull/2953#discussion_r1804106006.


AntonioVentilii-DFINITY (approved):
Tks!

PR #2966, opened 2024-10-17
feat(frontend): dapp modal

loki344:
# Motivation

Implement a modal to show the details of a dapp.

# Changes

- implement DappModal
- add modal to store
- enhance external link with class so that it can be better reused (IOC)

mobile with screenshot
image
mobile without screenshot
image
desktop with screenshot
image
desktop without screenshot
image


src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -0,0 +1,84 @@
...
+	import ExternalLinkIcon from '$lib/components/ui/ExternalLinkIcon.svelte';
+	import Img from '$lib/components/ui/Img.svelte';
+	import ImgBanner from '$lib/components/ui/ImgBanner.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+	import type { DappDescription } from '$lib/types/dappDescription';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+
+	export let dAppDescription: DappDescription;
+	const formattedUrl = new URL(dAppDescription.website).hostname.replace('www.', '');
```
peterpeterparker, line 17: 1. This can crash an error

2. Why removing `www`? 

3. nitpick: for coding style consistency prefer destructing objects

src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -0,0 +1,84 @@
...
+</script>
+
+<Modal on:nnsClose={modalStore.close}>
+	<div slot="title">
```
peterpeterparker, line 21: The title should be a fragment

src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -0,0 +1,84 @@
...
+		<span class="text-center text-xl">{dAppDescription.name}</span>
+	</div>
+
+	<div class="stretch flex flex-col gap-4 pt-4">
+		{#if nonNullish(dAppDescription.screenshots) && dAppDescription.screenshots.length > 0}
```
peterpeterparker, line 26: Isn't this the same piece of code as the one we reviewed recently? Shouldn't you create a component for it?

src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -0,0 +1,84 @@
...
+			<div class="overflow-hidden rounded-3xl">
+				<ImgBanner
+					src={dAppDescription.screenshots[0]}
+					alt={replacePlaceholders($i18n.dapps.alt.website, { $dAppname: dAppDescription.name })}
+				/>
+			</div>
+		{/if}
+
+		<div class="rounded-3xl p-5 shadow">
```
peterpeterparker, line 35: Should this be an article?

src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -0,0 +1,84 @@
...
+			<div class="flex items-center gap-4 border-b border-light-grey pb-5">
+				<div class="h-10 w-10">
```
peterpeterparker, line 37: Why not providing a size to the image to spare a div?

src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -0,0 +1,84 @@
...
+					<Img
+						src={dAppDescription.logo}
+						alt={replacePlaceholders($i18n.dapps.alt.logo, { $dAppname: dAppDescription.name })}
+					/>
+				</div>
+				<div>
+					<div class="text-lg font-bold">{dAppDescription.name}</div>
+					<div class="text-sm text-misty-rose">{formattedUrl}</div>
```
peterpeterparker, line 45: This should be a link I guess?

src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -0,0 +1,84 @@
...
+				</div>
+				<div class="ml-auto flex space-x-4">
+					{#if dAppDescription.twitter}
+						<ExternalLinkIcon
+							href={dAppDescription.twitter}
+							ariaLabel={$i18n.dapps.alt.open_twitter}
```
peterpeterparker, line 51: There isn't a placeholder in the label that should be replaced with the dapp name?

src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -0,0 +1,84 @@
...
+						>
+							<IconTwitter />
+						</ExternalLinkIcon>
+					{/if}
+					{#if dAppDescription.github}
+						<ExternalLinkIcon
+							href={dAppDescription.github}
+							ariaLabel={$i18n.dapps.alt.source_code_on_github}
```
peterpeterparker, line 59: There isn't a placeholder in the label that should be replaced with the dapp name?

src/frontend/src/lib/components/dapps/DappTags.svelte
```diff
@@ -0,0 +1,18 @@
+<script lang="ts">
+	import { i18n } from '$lib/stores/i18n.store';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+
+	export let tags: string[] = [];
```
peterpeterparker, line 5: Are you going to reuse this component elsewhere? If not, should we pass the dapp description and destructure the tags and dappme within the component to simplify its interface?

src/frontend/src/lib/components/ui/ExternalLink.svelte
```diff
@@ -14,7 +14,7 @@
 	{href}
 	rel="external noopener noreferrer"
 	target="_blank"
-	class="inline-flex items-center gap-2 no-underline"
+	class="inline-flex items-center gap-2 no-underline {$$restProps.class || ''}"
```
peterpeterparker, line 17: Shouldn't we provide this in another PR? Seems unrelated.

src/frontend/src/lib/stores/modal.store.ts
```diff
@@ -97,6 +99,7 @@ const initModalStore = <T>(): ModalStore<T> => {
 		openReceiveBitcoin: () => set({ type: 'receive-bitcoin' }),
 		openAboutWhat: () => set({ type: 'about-what' }),
 		openAboutHow: () => set({ type: 'about-how' }),
+		openDappDetails: () => set({ type: 'dapp-details' }),
```
peterpeterparker, line 102: Shouldn't opening the modal also receive the data as parameter - i.e. the dapp description?

src/frontend/src/lib/components/dapps/DappTags.svelte
loki344: I'm going to reuse it

src/frontend/src/lib/components/ui/ExternalLink.svelte
loki344: In my opinion not - This change is necessary to implement this modal. But let me know if you want a separate PR and I will make one.

src/frontend/src/lib/components/dapps/DappModal.svelte
loki344: 1. I see your point, we validate the data with zod to be a valid URL. So we would expect someone to use this component with other data than provided by the json in order to provide an invalid URL. I can handle this case gracefully by not rendering any links to this app - Which makes this modal a pure info modal, but I think this is fine for this edge case.
2. matching the design
3. done

src/frontend/src/lib/components/dapps/DappModal.svelte
loki344: The promobanner expects a screenshot. This modal can display dapps without screenshots. The promobanner needs position absolute etc. I think the abstraction with the ImgBanner is good enough.

src/frontend/src/lib/components/ui/ExternalLink.svelte
AntonioVentilii-DFINITY: on a separate note, should there be `nonNullish` here or `??` coalescing? I thought the linter would spot the second one...

src/frontend/src/lib/stores/modal.store.ts
loki344: Oh I like this, did not catch this pattern. Trying to figure out how to get the data in the modal if it's passed like this.

src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -0,0 +1,84 @@
...
+	export let dAppDescription: DappDescription;
+	const formattedUrl = new URL(dAppDescription.website).hostname.replace('www.', '');
+</script>
+
+<Modal on:nnsClose={modalStore.close}>
+	<div slot="title">
+		<span class="text-center text-xl">{dAppDescription.name}</span>
+	</div>
+
+	<div class="stretch flex flex-col gap-4 pt-4">
```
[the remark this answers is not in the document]
loki344: I would have to add an extension point to pass in the classname. Also, I'm not really sure if my content here has a 'toolbar' in that sense

src/frontend/src/lib/components/dapps/DappModal.svelte
loki344: made it reactive

loki344:
rdy for second round

src/frontend/src/lib/components/dapps/DappTags.svelte
peterpeterparker: Oki doki, let's see then. Thanks for the feedback!

src/frontend/src/lib/components/dapps/DappModal.svelte
peterpeterparker: 1. We don't have tests in place for either the types using Zod or this component. Therefore no certainty that this won't break someday. Please handle any potential errors.

2. Can you please review this with the designer?

- I assume that if a dapp has chosen to use www, they most likely want to communicate using that URL.
- If we decide to ignore their preferred URL, but a non-www URL is available, we can configure this at the metadata level.
- Given the critical importance of the hostname on the IC, particularly with respect to Internet Identity, I believe we should avoid altering the URL of projects.

src/frontend/src/lib/components/dapps/DappModal.svelte
peterpeterparker: Oki doki

src/frontend/src/lib/components/ui/ExternalLink.svelte
peterpeterparker: I think this should have been a separate PR because we don't spread classes but use a property named `styleClass` when passing a class to a component. However, if you want to keep this in this PR, please adjust the code accordingly for consistency reasons.

src/frontend/src/lib/components/dapps/DappModal.svelte
loki344: It's only about the visual appearance of the url, the link in the href is not touched by this formatting. I already checked with the designer: "Cool, let's then make it different. I was interested in cleaner look, so no https:// for sure, we can also try and drop www"

peterpeterparker: "Again, given the importance of hostnames on the IC, I personally would not mess with those, but if that's the wish, then let's not solve it at the code level and instead use a URL without 'www' in the metadata.

loki344: Hm ok, as he stated 'try and drop www'. I go with your feedback. thanks

src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -0,0 +1,112 @@
...
+	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+	import type { DappDescription } from '$lib/types/dappDescription';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+
+	export let dAppDescription: DappDescription;
+	$: ({ website, screenshots, twitter, github, tags, name, description, logo } = dAppDescription);
+
+	let isURLValid: boolean;
+	let formattedUrl: string | undefined;
```
peterpeterparker: To spare two potential variables that are not in sync, you can use a type `formattedUrl: OptionString` and set the value to `null` of the parsing fails in error.

We generally use this pattern "undefined === not set, null == we set that explicitely to null" for those use cases.

src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -0,0 +1,109 @@
...
+				</div>
+			</div>
+
+			<p class="m-0 my-5 text-sm">
+				<Html text={description} />
+			</p>
+			<DappTags dAppName={name} {tags} />
+		</article>
+
+		{#if websiteURL}
```
peterpeterparker, line 96: `nonNullish`

Note that we have a custom eslint rule in progress to check those things.

peterpeterparker (approved):
LGTM, thx

loki344:
@AntonioVentilii-DFINITY also good for you?

AntonioVentilii-DFINITY (approved):
LGTM tks

PR #2975, opened 2024-10-17
style(frontend): Remove legacy blue color

AntonioVentilii-DFINITY:
# Motivation

Alas, the legacy color defined as `blue` (but more on the purple side) was being used still. This PR removes its references in favor of the new blue decided for the pallet.


peterpeterparker (approved):
LGTM, thx

PR #2976, opened 2024-10-17
feat(frontend): Add USD balance in Hero for token page

AntonioVentilii-DFINITY:
# Motivation

We include the USD balance in the `Hero`, right below the token balance, when the user is in the transaction page.

Note that when the token has no USD price, we treat the USD balance in the same way, showing `$-` instead of zero.

# Results

Screenshot 2024-10-17 at 16 47 46
Screenshot 2024-10-17 at 16 47 54
Screenshot 2024-10-17 at 16 48 03
Screenshot 2024-10-17 at 16 48 09
Screenshot 2024-10-17 at 16 48 15


src/frontend/src/lib/components/hero/Balance.svelte
```diff
@@ -1,23 +1,45 @@
...
 	import { tokenSymbol } from '$lib/derived/token.derived';
+	import { balancesStore } from '$lib/stores/balances.store';
+	import { token } from '$lib/stores/token.store';
+	import { formatUSD } from '$lib/utils/format.utils';
+	import { calculateTokenUsdBalance } from '$lib/utils/token.utils';
+
+	let usdBalance: number | undefined;
+	$: usdBalance = nonNullish($token)
+		? calculateTokenUsdBalance({
+				token: $token,
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: Done! Better

src/frontend/src/lib/components/hero/Balance.svelte
```diff
@@ -1,23 +1,45 @@
...
+			<span><Amount amount={$balance} /> {$tokenSymbol}</span>
 		{:else}
-			<span class="text-5xl font-bold" class:animate-pulse={isNullish($balance)}>0.00</span>
+			<span class:animate-pulse={isNullish($balance)}>0.00 {$tokenSymbol}</span>
 		{/if}
+	</output>
 
-		{#if $erc20UserTokensInitialized && nonNullish(tokenSymbol)}
-			<span class="opacity-100">{$tokenSymbol}</span>
+	<span class="text-xl font-bold" class:opacity-50={usdBalance ?? 0 === 0}>
```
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: I think I did it with the latest commits, please have a look

src/frontend/src/lib/components/exchange/ExchangeTokenValue.svelte
```diff
@@ -1,18 +1,12 @@
 <script lang="ts">
-	import { nonNullish } from '@dfinity/utils';
+	import TokenExchangeBalance from '$lib/components/tokens/TokenExchangeBalance.svelte';
 	import TokenExchangeValueSkeleton from '$lib/components/tokens/TokenExchangeValueSkeleton.svelte';
 	import type { TokenUi } from '$lib/types/token';
-	import { formatUSD } from '$lib/utils/format.utils';
 
 	export let token: TokenUi;
+	$: ({ balance, usdBalance } = token);
```
peterpeterparker, line 7: Can you declare `balance` and `usdBalance` for consistency? Did not double check but, I think we always do so.

src/frontend/src/lib/components/tokens/TokenExchangeBalance.svelte
```diff
@@ -0,0 +1,16 @@
...
+
+	export let balance: TokenFinancialData['balance'];
+	export let usdBalance: TokenFinancialData['usdBalance'];
+</script>
+
+<output class="break-all">
+	{#if nonNullish(balance) && nonNullish(usdBalance)}
+		{formatUSD({ value: usdBalance })}
+	{:else}
+		{formatUSD({ value: 0, options: { minFraction: 0, maxFraction: 0 } }).replace('0', '-')}
```
peterpeterparker, line 14: It's correct we want to display zero when not initialized, not a `&ZeroWidthSpace`? If correct, shouls it be animated `class:animate-pulse`?

src/frontend/src/lib/components/exchange/ExchangeTokenValue.svelte
AntonioVentilii-DFINITY: I don't know why I didn't... strange! changed

src/frontend/src/lib/components/tokens/TokenExchangeBalance.svelte
AntonioVentilii-DFINITY: my understanding was that indeed: to display 0 not pulsating while the balance pulsates and loads. 

But you are right.... it doesn't make sense: I included pulsating, the same as balance.

It will pulsate ONLY when `balance` is nullish, as it was done before. That is because this component is used alsewhere and if `balance` is not nullish, the only case where `usdBalance` is nullish is when there are no exchange rates and in that case we already display `$-` in the rest of the code

peterpeterparker (approved):
LGTm, thx

PR #2978, opened 2024-10-18
refactor(frontend): Remove token store from HeroContent

AntonioVentilii-DFINITY:
# Motivation

The `HeroContent` component is using the `token` store. However it is not needed for its scope since it could directly use the `pageToken` store.

In fact, in this context, the token store is set as  such: `token.set($pageToken)`. We are not removing this, we are just substituting the store used to provide the same value.

Some of its sub-components use the `token` store too, but this PR change will not affect them, since we are just reading values that are equals.
We will analyze them in other PRs.


peterpeterparker (approved):
Coolio, thx

AntonioVentilii-DFINITY:
[OISY-99]

PR #2979, opened 2024-10-18
feat(frontend): Use page token for balance in Hero

AntonioVentilii-DFINITY:
# Motivation

In `HeroContent`, the component `Balance` was using the `balance` store, that it is a derived of the `token` store by itself. However, in the context of the usage of `Balance` component, we don't need to rely on the `token` store, and we can use directly the `pageToken` store.

In fact, in this context, the token store is set as such: token.set($pageToken).

So, we adapt `pageToken` store to have the balance values too.

# Changes

- Create new variable `pageTokenUi` in HeroContent that maps `pageToken` to have financial data.
- Make `Balance` component accepts a `token` prop.
- Adapt Balance component to use the token prop instead of global stores/derived.

# Tests

No changes in the local replica, same values showed.


peterpeterparker (commented):
Did you double-check if this change introduces additional re-renders? i.e. does `$pageToken` lead to as many render as using `$balance`?

AntonioVentilii-DFINITY:
@peterpeterparker it should actually reduce them (if not the same), because `balance` was a derived of both `balancesStore` and `token` store.
And in the specific context of the Hero, `token` store is set to be the same as `pageToken` store.

So, what we are basically doing is:
1. substituting `token` store with `pageToken`
2. calculate the balance using the `balancesStore` (as it was done before)

So, the re-render will happen every time that:

### Before
`token` store and `balancesStore` are changed (and, as you know, `token` store is handled a few times)

### After
`pageToken` and the `balancesStore` are changed)


Does that answer the question?

peterpeterparker (approved):
Thanks for the additional details. LGTM, thx

AntonioVentilii-DFINITY:
[OISY-99]

PR #2984, opened 2024-10-18
feat(frontend):move filter to derived function

darioloetsc:
# Motivation

Preparation for group view integration.

# Changes

Moving the filter to derived function needs to be wrapped again later.

# Tests

None

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
@@ -1,16 +1,15 @@
 <script lang="ts">
 	import { debounce } from '@dfinity/utils';
-	import { combinedDerivedSortedNetworkTokensUi } from '$lib/derived/network-tokens.derived';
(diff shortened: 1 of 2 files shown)


AntonioVentilii-DFINITY (commented):
Before reviewing, maybe I missed a step: are you not going to use the tokens list that `TokensDisplayHandler` provides? Meaning, is it not the grouping being done directly inside the `TokensSignedIn` component's list?

AntonioVentilii-DFINITY:
@darioloetsc thank you for the explanation! Then why should the balance filter be moved to a separate derived store? For example, why not using directly `sortedTokens` list

darioloetsc:
Please provide a change request, sry I cannot follow. The idea of this commit is to make it easier for later mapping and not pile wrappers within the handler. If you have a better idea of how to clean up the handler, please go ahead.

darioloetsc:
As discussed, we dont do it. To many derived chained.

PR #2985, opened 2024-10-18
feat(frontend): implement info message for BTC mainnet transactions

DenysKarmazynDFINITY:
# Motivation

In this PR, 2 things are added:
1. "Transaction" headline for all BTC tx pages  (it was missing previously).
2. A collapsable text block with an info button as a trigger for mainnet BTC only.

Inactive:
Screenshot 2024-10-18 at 15 26 12

Active:
Screenshot 2024-10-18 at 15 26 20


AntonioVentilii-DFINITY (approved):
LGTM! Tks!

I imagine it was like this in the requirements, but maybe a small arrow would be useful on the side of the info button when `Collapsible` is expanded? Just as an idea, not a call to action

DenysKarmazynDFINITY:
Thanks for the feedback, I'll check if we want an arrow there 👍

PR #2986, opened 2024-10-18
refactor(frontend): Create specific Checkbox component to be re-used

AntonioVentilii-DFINITY:
# Motivation

With the aim of making it as much closer as the design as possible (but even improving the general estetic of it), we substitute the built-in Checkbox component with a new one with the same functionalities and smaller adaptations. For example, it will now have a central checkmark.

### Default

Screenshot 2024-10-18 at 16 36 55

### Hover

Screenshot 2024-10-18 at 16 37 03

### Checked

Screenshot 2024-10-18 at 16 36 50

### Checked and Hover

Screenshot 2024-10-18 at 16 37 15

### Video of usage

https://github.com/user-attachments/assets/b03ba38b-ea8b-4c26-a85b-96f07ab63d92

# Note

There is no easy way of modifying the an element `<input type="checkbox" ... />` that I know of. In fact, if one wants to customize part of it, it must first annul the previous CSS appearances (`appearance-none`) and then re-build the important things, like the checkmark.

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensZeroBalanceCheckbox.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensZeroBalanceCheckbox.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { Checkbox } from '@dfinity/gix-components';
+	import Checkbox from '$lib/components/ui/Checkbox.svelte';
 	import { hideZeroBalances } from '$lib/derived/settings.derived';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { hideZeroBalancesStore } from '$lib/stores/settings.store';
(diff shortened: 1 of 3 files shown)


peterpeterparker (commented):
I suggest to not create a new checkbox component but continue to use the one of `gix-components` given that it offers more features.

Is there anything else more that the UI offset we discussed that would need to be addressed?

AntonioVentilii-DFINITY:
@peterpeterparker not really, no, only the position of the checkmark indeed, but to modify it I should first modify it in the gix components repo, ok? if yes, i will close this PR

peterpeterparker:
> I should first modify it in the gix components repo, ok?

If it's possible, sounds good to me. There are few devs that use the gix-cmp some it might be useful for others too some day.

AntonioVentilii-DFINITY:
Moving to draft first, I will check if the gix-components modifications go well

AntonioVentilii:
This topic is being solved directly in the GIX Components repo:

- PR https://github.com/dfinity/gix-components/pull/539

PR #2988, opened 2024-10-19
refactor(e2e): extract list of tokens to be looped

AntonioVentilii-DFINITY:
# Motivation

We are going to create more tests based on what happens in each token page, or for each token that we would test. That is why we extract a re-usable token list for the tests.

# Note

We don't use directly the list of tokens constants, but their symbol. That is because in the tokens we are importing the SVG icons that are not correctly recognized by `playwright` (see [workflow](https://github.com/dfinity/oisy-wallet/actions/runs/11415076238/job/31764716557?pr=2988)). This is a [know issue](https://github.com/microsoft/playwright/issues/22657) with `playwright` apparently.


AntonioVentilii-DFINITY:
apparently the issue with importing static assets with playwright is more profound, and not hackable in this PR unfortunately. For now, we keep hard-coding all the tokens we want to test, and not loop them

PR #2990, opened 2024-10-20
feat(frontend): Check allowance when sending ERC20 tokens

AntonioVentilii-DFINITY:
# Context

# Example

    1. `approve` transaction for 10 USDT --> _successful_
    2. `transfer_from` transaction for 10 USDT --> _failed_ (for whatever reason)
    1. `approve` transaction for 20 USDT --> _failed_ (the previous approved amount is still there)
    2. `transfer_from` transaction for 20 USDT --> _failed_ (not enough approved amount)
    1. `approve` transaction for 7 USDT --> _failed_ (the previous approved amount is still there)
    2. `transfer_from` transaction for 7 USDT --> _successful_ (remaining approved amount is 3 USDT)

# Practical example

# Motivation

Following up PR https://github.com/dfinity/oisy-wallet/pull/2812 and https://github.com/dfinity/oisy-wallet/pull/2813, we now introduce a check before requesting am amount approval.
The code will first check if there is already a pre-approved amount for the contract (`allowance`), and if that is enough, it will not request a new approval.
On the contrary, if the `allowance` shows that the approved amount will not be enough for the amount required by the current transaction, we must first reset the `allowance` to `ZERO`, and then request a new approval for the correct amount.

# Changes

- Deconstruct some params from input of `approve` function: these will be used by other functions inside it.
- Add `allowance` query and check against the required amount: if enough, already return a positive outcome.
- Reset `allowance` if not enough.
- Note that we don't pass `progress` to the reset approval, since it is not part of `ProgressStepsSend`.

# Tests

I deployed this branch in env test_fe_2 and used my wallet address [0xc0828a992f7f2f0eb6165ac8f19b4aeb9bd07ed8](https://sepolia.etherscan.io/address/0xc0828a992f7f2f0eb6165ac8f19b4aeb9bd07ed8) to simulate some USDC --> ckUSDC conversions in Sepolia.

### Current Allowance = 0

1. I first [approved an amount 1 USDC](https://sepolia.etherscan.io/tx/0xe6eddd6e95e2b5ff45f2b5e03d991480e35b635087905938888df8bd6b571e79) in an independent manner.
2. Then I [asked for a conversion of 2 USDC to ckUSD](https://sepolia.etherscan.io/tx/0x01d3386a6e82baf01a49ac492a3df097c0ed0200ce0240bfe7c8e7413e784496): it failed as expected, with the old methods. And it would keep to fail if I would not reset it.

### Current Allowance = 1 USDC

3. I implemented the new method and asked for a conversion of 0.7 USD to ckUSDC. Since the pre-approved amount was enough, it just did a [`deposit` transaction](https://sepolia.etherscan.io/tx/0x00520333d8accd0c9dbc6d65db5487b9a18372531f17908b461a96d1a3e4c0bb).

### Current Allowance = 0.3 USDC

4. I then requested a conversion of 2 USDC to ckUSD. Since it was not enough:
   1. First it [reset the allowance to ZERO](https://sepolia.etherscan.io/tx/0xe9b8e98defbd9902421e1dc78644c27f88fdf051cca0cc7bd9db2d8d57ab2900).
   2. Then it [requested an approval for the required amount](https://sepolia.etherscan.io/tx/0xef3b5c76e8262abf1ee2e1710cac9745bd5524b9c9d0b35d6302e6b1891b6552).
   3. Finally, it send the [conversion `deposit` transaction](https://sepolia.etherscan.io/tx/0x23bd30b94df299cb3999285d0591db99bb5c1ec897a1b867282c09e740474166).

### Current allowance = 0 USDC

5. Last case scenario, it is the most common one: the user has no pre-approved amount: the conversion worked smoothly as expected:
   1. First it [requested an approval](https://sepolia.etherscan.io/tx/0x5da20ebd0c023a1d7476b88a777b23b93b882726f7314edf99eb4026d1982ea4) for the required amount.
   2. Then it made a [`deposit` transaction for the conversion](https://sepolia.etherscan.io/tx/0x546b2ff584ba4fb703f985ee1b89b5db486a3ec2094c6e2a66615638c0398be1).

### Etherscan Screenshot

Screenshot 2024-10-21 at 00 08 30


src/frontend/src/eth/services/send.services.ts
```diff
@@ -508,14 +511,45 @@ const approve = async ({
 			erc20HelperContractAddress
 		})
 	) {
-		return { transactionApproved: false };
+		return { transactionApproved: false, nonce };
 	}
 
+	const preApprovedAmount = await erc20ContractAllowance({
```
peterpeterparker, line 517: Can we extract this logic in a dedicated function? That would scope the guard logic and make the code a bit more easy to understand, thinking at the future.

src/frontend/src/eth/services/send.services.ts
```diff
@@ -508,14 +511,45 @@ const approve = async ({
...
+			...rest,
+			nonce,
+			amount: ZERO,
+			token,
+			sourceNetwork,
+			spender: erc20HelperContractAddress
+		});
+	}
+
+	const nonceApproval = preApprovedAmount.gt(ZERO) ? nonce + 1 : nonce;
```
peterpeterparker, line 542: Can you add a comment about why we increment the nonce here? I asked myself why we do so because the caller of the function also increment the nonce (`nonceTransaction`).

src/frontend/src/eth/services/send.services.ts
```diff
@@ -508,14 +511,45 @@ const approve = async ({
...
+		token,
+		owner: from,
+		spender: erc20HelperContractAddress,
+		networkId
+	});
+
+	// If there is already an approved allowance that is enough for the required amount, we don't need to approve again.
+	if (preApprovedAmount.gte(amount)) {
+		progress(ProgressStepsSend.APPROVE);
+		return { transactionApproved: false, nonce };
```
peterpeterparker, line 527: I'm confused about `transactionApproved: false` because I interpret it as if it were an error. I'm also confused about whether we should distinguish between this state and, although not yet used, a potential `success: false` in `prepareAndSignApproval`.

Do you know what I mean?

src/frontend/src/eth/services/send.services.ts
```diff
@@ -508,14 +511,45 @@ const approve = async ({
...
+	}
+
+	// If the existing pre-approved amount is not enough but non-null, we need to reset the allowance first and then approve the new amount.
+	if (preApprovedAmount.gt(ZERO)) {
+		await prepareAndSignApproval({
```
peterpeterparker, line 532: 1. For readability maybe a local function?

```
const resetExistingApprovalToZero = async () => await prepareAndSignApproval({...

await resetExistingApprovalToZero();
```

2. `prepareAndSignApproval` update the progression in the UI, is it correct to do so at this stage? Isn't the UI going to go back an forth between steps SIGN_APPROVE -> APPROVE -> SIGN_APPROVE -> APPROVE?

AntonioVentilii-DFINITY: 1. Yes, I was wondering that myself, it is cleaner

2. Not really, because we are not passing `progress` to the `reset` function. So, in the UI, the difference is that the initialization would take more time (I tested it), before going to `SIGN_APPROVE`.

I was thinking of making it with more steps, if we need to reset it. But it is not that easy: conditional steps that change while the progress is already "in progress".

What do you suggest? Ideally, I could just change the label of the INITIALIZATION step: it would dynamically change from "Initializing transaction" to "Initializing transaction (reset allowance)", or something similar, boh

src/frontend/src/eth/services/send.services.ts
AntonioVentilii-DFINITY: Yes, not that clear indeed: I wanted to keep the same logic in the sense that `transactionApproved` indicates that we actually created an "approval" transaction. For example, if the function is run for something that it is not ck-conversion, the code returns `transactionApproved: false`, meaning that there was no need to approve anything, nor create a transaction to approve anything. But that does not indicates an error, it is just used to increase the `nonce` or not.

I would suggest to rename it as `approvalTransactionNeeded`, or something similar. Or remove it completely and just use a nullish check on the `hash` return: if hash is undefined (no new approval transaction), then do not increment the nonce

src/frontend/src/eth/services/send.services.ts
AntonioVentilii-DFINITY: done

src/frontend/src/eth/services/send.services.ts
AntonioVentilii-DFINITY: I think I maybe understood: may you please have a look at the latests commits, WDYT?

src/frontend/src/eth/services/send.services.ts
AntonioVentilii-DFINITY: Check this one please, does it make sense?

https://github.com/dfinity/oisy-wallet/pull/3004

src/frontend/src/eth/services/send.services.ts
peterpeterparker: I'm not sure UX wise that we should go too deep in the details. Since we are not passing `progress` I guess my concern is already addressed.

src/frontend/src/eth/services/send.services.ts
```diff
@@ -484,20 +469,75 @@ const prepareAndSignApproval = async ({
...
+	token,
+	from,
+	spender,
+	amount,
+	sourceNetwork,
+	...rest
+}: Omit<ApproveParams, 'to' | 'minterInfo' | 'progress'> & {
+	nonce: number;
+	spender: EthAddress;
+}): Promise<{ existingApprovalIsEnough: boolean; approvalNeededReset: boolean }> => {
```
peterpeterparker, line 494: The case `true` & `true` is not handled, so instead of two boolean, I was thinking, can we return a `type`?

Something like:

```
result: "existingApprovalsEnough" | "approveNeededReset" | "none"
```

WDYT?

src/frontend/src/eth/services/send.services.ts
AntonioVentilii-DFINITY: cool!

src/frontend/src/eth/services/send.services.ts
AntonioVentilii-DFINITY: Yes, better! Changed!

peterpeterparker (approved):
LGTM, thx

If one of these days you feel like it, adding some tests with mocks here to ensure that certain steps are performed would probably be useful because it's probably tricky to setup with E2E.

PR #2991, opened 2024-10-20
ci(deploy): Set description for run-name

AntonioVentilii-DFINITY:
# Motivation

Useful to have exactly what are the chosen parameters of the workflow in case of `workflow_dispatch`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #2992, opened 2024-10-20
style(frontend): Break ck-tokens name in Hero conversion button

AntonioVentilii-DFINITY:
# Motivation

Since we cannot know beforehand the length of a possible future ckToken (or ERC20 token), we make it so in the conversion button in `Hero`, the name breaks if overflows.
We have already a few examples with Sepolia tokens, where their ck-counterparty is too long to fit.

NOTE: there will be a change in the button, probably the label will change to "Convert" directly. So, for now this is just a visual fix.

### Before

Screenshot 2024-10-21 at 00 37 39
Screenshot 2024-10-21 at 00 37 47
Screenshot 2024-10-21 at 00 37 55

### After

Screenshot 2024-10-21 at 00 33 11
Screenshot 2024-10-21 at 00 33 19
Screenshot 2024-10-21 at 00 33 27


peterpeterparker (approved):
LGTM, thx

Note: Asked myself if the button should not just be named "Convert" as a pragmatic solution

PR #2997, opened 2024-10-21
feat(frontend): dapp route

loki344:
# Motivation

Implement the route for the dapp explorer to make the curated list of dapps available to users. We introduce a Navigation item and do not display the hero section on the route dapps.

# Changes

- Add route dapps
- Add menu item dapps
- Implement DappsSignedIn
- Implement DappsSignedOut

image
image


peterpeterparker (commented):
Thanks for the PR. A few inputs before reviewing the code:

1. I discussed with Artem, the route should be named`/explore`

2. If there are no technical issues that could delay the implementation, the route should be public since it contains only static information. This way, we might be able to pre-render the information in the future, and it would also anticipate any UI/UX changes if stakeholders decide to make the information public, following a similar pattern to what is implemented in the NNS dapp. If you foresee any blockers, please let me know which ones?

3. Could you split the work into smaller PRs? For example, first provide the new route, then its content (in one or two PRs), and once everything is in place, submit the PR to add the feature to the menu. This will make features scoped and reviews easier.

4. I'm not sure if the designer intended it this way, but the "Featured" text above the KongSwap banner doesn’t seem to have enough contrast.

loki344:
> Thanks for the PR. A few inputs before reviewing the code:
> 
> 1. I discussed with Artem, the route should be named`/explore`
> 2. If there are no technical issues that could delay the implementation, the route should be public since it contains only static information. This way, we might be able to pre-render the information in the future, and it would also anticipate any UI/UX changes if stakeholders decide to make the information public, following a similar pattern to what is implemented in the NNS dapp. If you foresee any blockers, please let me know which ones?
> 3. Could you split the work into smaller PRs? For example, first provide the new route, then its content (in one or two PRs), and once everything is in place, submit the PR to add the feature to the menu. This will make features scoped and reviews easier.
> 4. I'm not sure if the designer intended it this way, but the "Featured" text above the KongSwap banner doesn’t seem to have enough contrast.

1. Will change it
2. Yes, I'll make it public
3. I'll try
4. Yeah the text will be blue, I have a todo in the banner anyways for activate the button that we commented out in a previous PR

loki344:
> > Thanks for the PR. A few inputs before reviewing the code:
> > 
> > 1. I discussed with Artem, the route should be named`/explore`
> > 2. If there are no technical issues that could delay the implementation, the route should be public since it contains only static information. This way, we might be able to pre-render the information in the future, and it would also anticipate any UI/UX changes if stakeholders decide to make the information public, following a similar pattern to what is implemented in the NNS dapp. If you foresee any blockers, please let me know which ones?
> > 3. Could you split the work into smaller PRs? For example, first provide the new route, then its content (in one or two PRs), and once everything is in place, submit the PR to add the feature to the menu. This will make features scoped and reviews easier.
> > 4. I'm not sure if the designer intended it this way, but the "Featured" text above the KongSwap banner doesn’t seem to have enough contrast.
> 
> 1. Will change it
> 2. Yes, I'll make it public
> 3. I'll try
> 4. Yeah the text will be blue, I have a todo in the banner anyways for activate the button that we commented out in a previous PR

Regarding point 2: On second look I think it's better if we begin with the protected route and switch to a public route in a second step. The OISY setup with the Authguard differs from the nns dapp setup and I don't want to introduce additional risks by refactoring that logic.

loki344:
This PR was split into several PR's and is no longer relevant.

PR #3004, opened 2024-10-21
refactor(frontend): Rename results param in send function

AntonioVentilii-DFINITY:
# Motivation

The param `transactionApproved` indicates actually the needed for an `approval` transaction to be executed, before the rest. For example, in a ck-conversion.

So, to make it less ambiguous, we rename it as `transactionNeededApproval`.

Note: any better naming suggestions are welcomed.


peterpeterparker (approved):
LGTm, thx

PR #3005, opened 2024-10-21
feat(frontend): remove analytics duration

peterpeterparker:
# Motivation

We temporarly added some metrics about call duration for some testing purpose. Given that this is not required anymore, we remove those to make the code a bit more readable.

# Changes

- Remove constants and functions related to tracking analytics durations


AntonioVentilii-DFINITY (approved):
LGTM! Tks

PR #3007, opened 2024-10-21
chore(frontend): Remove SASS deprecation warning for imports

AntonioVentilii-DFINITY:
# Motivation

The syntax `@import` was deprecated in favor of `@use` and `@forward` (see [announcement](https://sass-lang.com/blog/import-is-deprecated/)). As per the latest versions of SASS, we are receiving deprecation warnings about it, for example:

```bash
Deprecation Warning: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.

More info and automated migrator: https://sass-lang.com/d/import

  ╷
6 │ @import './global/fonts';
  │         ^^^^^^^^^^^^^^^^
  ╵
    src/frontend/src/lib/styles/global.scss 6:9  root stylesheet
```

So we adjust the only usages of `@import` left.

To do that I used directly the command suggested in the article:

`sass-migrator module --migrate-deps ./global.scss`

### DIFF
--- a/src/frontend/src/lib/styles/global.scss
+++ b/src/frontend/src/lib/styles/global.scss
@@ -1,7 +1,3 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
-
 @use './global/font-faces';
 @use './global/fonts';
 @use './global/theme';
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
Nice, thanks!

PR #3011, opened 2024-10-21
feat(frontend): copy i18n

peterpeterparker:
# Motivation

i18n key was missing copy.

# Changes

- Extract "copy" text.


AntonioVentilii-DFINITY (approved):
Good catch!

PR #3012, opened 2024-10-21
feat(frontend): add route for dapp explorer

loki344:
# Motivation

We want to have a route for the dapp explorer so users can see it.

# Changes
- add the route setup for the dapp-explorer
- the route is under /explore
- content is empty and will follow in a further PR


peterpeterparker (commented):
Ultimately, what's the call? Are we going to introduce this new `/explore` route as a public route, or will it only be available to signed-in users?

If it's the latter, all good—I'll review this PR.

If it's the former, the route shouldn't be placed under `(app)` I would say because it's layout uses the authentication guard. We could, for example, create a new route group `(public)`, but I'm not sure what the decision is?

loki344:
> Ultimately, what's the call? Are we going to introduce this new `/explore` route as a public route, or will it only be available to signed-in users?
> 
> If it's the latter, all good—I'll review this PR.
> 
> If it's the former, the route shouldn't be placed under `(app)` I would say because it's layout uses the authentication guard. We could, for example, create a new route group `(public)`, but I'm not sure what the decision is?
> 
> PS: sorry noticing now the PR is a draft. Don't mind my comment if too early to ask.

No Problem, I answered in the other MR. For now let's add it as a protected route as the landing page designs do not indicate any links to the explore page. And because the setup with the AuthGuard differs from the nns-dapp repo I don't feel confident setup the public route on short notice. Is that fine with you?

peterpeterparker:
>  For now let's add it as a protected route

Alright, let's do the following. We'll integrate it as a protected route with the PRs you've prepared. That way, the feature will be available, and once everything is merged, I'll take care of making it public.

peterpeterparker (approved):
LGTM, thx

PR #3013, opened 2024-10-21
feat(frontend): review address block

peterpeterparker:
# Motivation

We want to redesign a bit the address we present in modals. We keep the address and buttons but, we change the block and buttons style. We also want to display an icon for the network.

# Changes

- Pass network to consume `ReceiveAddress`
- Update `ReceiveActions` to use button icon and new `ReceiveCopy` component

# Screenshots

Capture d’écran 2024-10-22 à 06 51 57
Capture d’écran 2024-10-22 à 06 51 34
Capture d’écran 2024-10-22 à 06 51 16
Capture d’écran 2024-10-22 à 06 50 50


AntonioVentilii-DFINITY (approved):
LGTM!

Unsolicited opinion: shouldn't the address be ellipsed? I know it was already "expanded", but since we are changing it...

peterpeterparker:
> shouldn't the address be ellipsed? I know it was already "expanded", but since we are changing it...

I asked Artem the same question. After having a look at other wallets, which seem to also display full length, we decided to stick to a full length address.

PR #3014, opened 2024-10-21
refactor(frontend): extract copy to clipboard to a module

peterpeterparker:
# Motivation

In #3013 we will need a specific copy button icon. I'm not sure if it will be use elsewhere, that's why instead of enhancing the existing `Copy` button, I rather like to create a specific "Receive related button". Also why it's handy to extract the "copy to clipboard" function to a module for reusability purposes.

# Changes

- Move existing copy to clipboard function to a utils.

# Tests

No particular tests.


AntonioVentilii-DFINITY (approved):
LGTM! tks

PR #3019, opened 2024-10-21
feat(frontend): dapp route content

loki344:
# Motivation

Render the dapp explorer component under /explore

# Changes

- Implement DappExplorer with a Promobanner, a row to filter the dapps by the tags and a card grid.

# Screenshots
image

image


src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
```diff
@@ -0,0 +1,80 @@
...
+	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+	import {
+		type DappDescription,
+		dAppDescriptions,
+		type FeaturedDappDescription
+	} from '$lib/types/dappDescription';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+
+	let selectedTag: string | null = null;
```
peterpeterparker, line 17: nitpick: we actually generally used  "something or `undefined`". `null` works for sure just fine but is less consistent

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
```diff
@@ -0,0 +1,80 @@
...
+	const featuredDapp: FeaturedDappDescription | undefined = dAppDescriptions.find(
```
peterpeterparker, line 18: - Shouldn't it be a `filter` - i.e. aren't multiple `featuredDapps` possible?

- the cast is needed? `dAppDescriptions` is not strongly typed enough?

- nitpick: not against fancy deconstruction

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
```diff
@@ -0,0 +1,80 @@
...
+		(dApp) => dApp.featured && nonNullish(dApp.screenshots) && dApp.screenshots.length > 0
+	) as FeaturedDappDescription | undefined;
+
+	const uniqueTags = new Set(
+		dAppDescriptions.flatMap((dapp) => dapp.tags).sort((tagA, tagB) => tagA.localeCompare(tagB))
+	);
+
+	$: filteredDapps = nonNullish(selectedTag)
+		? dAppDescriptions.filter((dApp) => dApp.tags.includes(selectedTag!))
```
peterpeterparker, line 27: Please avoid the use of `!`.

cc @AntonioVentilii-DFINITY can you setup an eslint rule?

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
```diff
@@ -0,0 +1,80 @@
...
+		: dAppDescriptions;
+
+	$: selectedDapp = $modalDAppDetails
```
peterpeterparker, line 30: can you declare `selectedDapp` for consistency reason?

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
```diff
@@ -0,0 +1,80 @@
...
+{#if nonNullish(featuredDapp) && nonNullish(featuredDapp.screenshots)}
+	<div class="mb-10">
+		<DappPromoBanner
+			on:click={() => modalStore.openDappDetails(featuredDapp)}
+			dAppDescription={featuredDapp}
+		/>
+	</div>
+{/if}
+
+<div class="flex flex-wrap gap-4 {$$restProps.class || ''}">
```
peterpeterparker, line 46: - There are style classes that are passed to the components?

- Prefer `??` over `||`

cc @AntonioVentilii-DFINITY we have an eslint rule for that?

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
```diff
@@ -0,0 +1,80 @@
...
+			on:click={() => (selectedTag = tag)}
+			styleClass="text-nowrap max-w-fit text-sm"
+			colorStyle={selectedTag === tag ? 'primary' : 'tertiary'}>{tag}</Button
+		>
+	{/each}
+</div>
+
+<ul class="mt-8 grid list-none grid-cols-1 gap-x-4 gap-y-7 md:grid-cols-3">
+	{#each filteredDapps as dApp}
+		<li class="flex">
```
peterpeterparker, line 67: Would a short in fade animation smoothen the experience?

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
```diff
@@ -0,0 +1,80 @@
...
+				on:click={() => {
+					modalStore.openDappDetails(dApp);
+				}}
+				dAppDescription={dApp}
+			/>
+		</li>
+	{/each}
+</ul>
+
+{#if $modalDAppDetails && selectedDapp}
```
peterpeterparker, line 78: `nonNullish(selectedDapp)`

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
loki344: Ah I see, it lints, but my idea does not like it. Presumably because of the nonNullish wrapper.. 
image

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
loki344: - For the moment there will be only one featured dapp according to Artem.
- seems as it lints - my IDEA is not happy, but that's ok.
- I don't understand this, do you want deconstruction inside the find lambda?

peterpeterparker: > For the moment there will be only one featured dapp according to Artem.

Can you add a comment or create a function - like "selectFirstFeaturedDapp" - to make this requirement explicit?

> deconstruction

```
// For the moment only the first featured dapp is highlighted
const selectFirstFeaturedDapp = (): FeaturedDappDescription | undefined => 
  dAppDescriptions.find(({featured, screenshots}) => featured && nonNullish(screenshots) && screenshots.length);

const featuredDapp = selectFirstFeaturedDapp()
```

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
loki344: just wondering.. I had to apply a cast to the selectedTag because the npm run check did not succeed. Is this the normal case when using nonNullish? Using the normal comparison, the compiler would be happy.
image

peterpeterparker: It seems unrelated to `nonNullish`. If you `selectedTag !== undefined` the linter complains the same way. Tried last version of TypeScript, same same. I guess it cannot defers it.

```
Error: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'. (ts)
		selectedTag !== undefined
			? dAppDescriptions.filter((dApp) => dApp.tags.includes(selectedTag))
			: dAppDescriptions;


====================================
svelte-check found 1 error and 0 warnings in 1 file
```

Let's add an additional test instead of a cast:

```
$: filteredDapps = nonNullish(selectedTag)
		? dAppDescriptions.filter(({ tags }) => nonNullish(selectedTag) && tags.includes(selectedTag))
		: dAppDescriptions;
```

Or a bit less performant but code wise more concise:

```
$: filteredDapps = dAppDescriptions.filter(
		({ tags }) => isNullish(selectedTag) || tags.includes(selectedTag)
	);
```

loki344: gotcha, you're right, I forgot to remove the bang (!) 😄

peterpeterparker (approved):
LGTM, thx

PR #3027, opened 2024-10-22
build(ci): automated release process

inc-man:
# Motivation

The previously merged release process had its flaws, so we want an new version.

# Changes

fixed snytax errors
adjusted release-drafter configuration
better formatting


.github/release-drafter.yml
```diff
@@ -26,8 +26,18 @@ categories:
...
+template: |
+  ## What's Changed
+
+  ### Detailed Changes
+  {% for change in changes %}
+  - **Title**: {{ change.title }}
+  - **PR**: [#{{ change.number }}]({{ change.html_url }})
+  - **Author**: @{{ change.author.login }}
+  - **Motivation**:
+    {{ change.body | paragraph_start_with("# Motivation") | default("No specific motivation provided.") }}
```
bitdivine, line 40: Will this get the whole markdown section, or just the first paragraph in the section?

bitdivine (approved):
This looks like progress.  Thanks!

.github/release-drafter.yml
bitdivine: I couldn't find the documentation for `paragraph_start_with` to confirm. Do you know where that is?

inc-man: There is no real documentation on that. Basically the release-drafter.yml uses Handlebars as template-engine. But paragraph_start_with is not part of it. So I want to try and see what happens. Otherwise I feel like I can't really get just the Motivation or just the Changes.

PR #3032, opened 2024-10-22
feat(frontend): i18n for the Carousel component

DenysKarmazynDFINITY:
# Motivation

i18n keys/types for the Carousel components.

### DIFF
--- a/src/frontend/src/lib/i18n/en.json
+++ b/src/frontend/src/lib/i18n/en.json
@@ -726,5 +726,12 @@
 				"cannot_call": "The call cannot be processed"
 			}
 		}
+	},
+	"carousel": {
+		"text": {
+			"next_slide": "Next slide",
+			"prev_slide": "Previous slide",
+			"indicator": "Jump to slide $index"
+		}
 	}
 }
--- a/src/frontend/src/lib/types/i18n.d.ts
+++ b/src/frontend/src/lib/types/i18n.d.ts
@@ -629,6 +629,10 @@ interface I18nSigner {
 	};
 }
 
+interface I18nCarousel {
+	text: { next_slide: string; prev_slide: string; indicator: string };
+}
+
 interface I18n {
 	lang: Languages;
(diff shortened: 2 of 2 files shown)


src/frontend/src/lib/i18n/en.json
```diff
@@ -726,5 +726,12 @@
 				"cannot_call": "The call cannot be processed"
 			}
 		}
+	},
+	"carousel": {
+		"text": {
+			"next_slide": "Next slide",
```
DenysKarmazynDFINITY, line 732: Those keys will be used as `aria-label` for the respective control/indicator buttons.

peterpeterparker (commented):
Are you going to build a generic carousel or one specifically designed for the showcased applications?

If the former, texts seem accurate, if the later, they probably have to be more specific no? Like "Jump to a specific showcased application" or "Next application highlighted on OISY explorer" those kind of text.

WDYT?

peterpeterparker:
Ah and if generic caroussel, have you discussed building it in Gix-cmp - i.e. is NNS dapp going to use such a component one day?

DenysKarmazynDFINITY:
@peterpeterparker the plan is to build as generic as possible carousel component with indicators and controls as shown on the screenshot. There were no discussions re. if we should implement it on the components library side, so I'm implementing it here in Oisy, but eventually it can be moved to GIX-cmp, especially when there is another use-case for the carousel that can help us understand if it's 100% generic.

Screenshot 2024-10-22 at 12 19 31

peterpeterparker:
I see, let me start a thread internaly.

peterpeterparker (approved):
Artem confirmed that we have no other choice but to implement it now in OISY, given that the reusability of the component was not included in the task estimation. It's a pity, as the foundation may have to spend twice the effort, but at least I asked.

LGTM, thanks

DenysKarmazynDFINITY:
@peterpeterparker thanks for double-checking!

PR #3034, opened 2024-10-22
chore(ci): fixed deployment task

inc-man:
# Motivation

run-name needs to be executed after the on: block

# Changes

put run-name after the on: block


bitdivine (approved):
Thanks!

PR #3036, opened 2024-10-22
feat(frontend): receive modal title with symbol

peterpeterparker:
# Motivation

The designers want to display the token symbol within the "Receive" modal title.

# Changes

- Since `addressToken` is per definition potentially `undefined` either just display "Receive" or "Receive $token"
- Create a component to extract the logic for the title (could have been a function but felt cleaner that way. no strong opinion here)
- Consume the title component in receive modals

# Screenshots

Capture d’écran 2024-10-22 à 13 45 03
Capture d’écran 2024-10-22 à 13 44 55
Capture d’écran 2024-10-22 à 13 44 50
Capture d’écran 2024-10-22 à 13 44 45
Capture d’écran 2024-10-22 à 13 44 38
Capture d’écran 2024-10-22 à 13 44 30
Capture d’écran 2024-10-22 à 13 44 26


src/frontend/src/lib/components/receive/ReceiveTitle.svelte
```diff
@@ -0,0 +1,17 @@
...
+	export let addressToken: Token | undefined;
+
+	let symbol: string | undefined;
+	$: symbol = addressToken?.symbol;
+</script>
+
+{isNullish(symbol)
+	? $i18n.receive.text.receive
+	: replacePlaceholders($i18n.receive.text.receive_token, {
+			$token: addressToken?.symbol ?? ''
```
AntonioVentilii-DFINITY, line 16: `$token: symbol` (without coalescing either)

or is it on purpose for a reason?

AntonioVentilii-DFINITY (approved):
LGTM! Tks! One small observation please

P.S. could the new component `ReceiveTitle` be used for `IcReceiveCkEthereumModal` in another PR? Or was it already discussed with designers

src/frontend/src/lib/components/receive/ReceiveTitle.svelte
peterpeterparker: Good catch! Thanks!

peterpeterparker:
> P.S. could the new component ReceiveTitle be used for IcReceiveCkEthereumModal in another PR? Or was it already discussed with designers

Maybe indeed. But let's implement the big mess I started in  #3061 which requires lots of changes in the receive modal before starting any other development on the modals I would say.

PR #3038, opened 2024-10-22
feat(frontend): back and logo transition

peterpeterparker:
# Motivation

I find the transition between "OISY logo" and "Back" in the header a bit rough. Therefore I proposed to add an small slide animation between the two states.

# Changes

- Use an `in:slide` on `x` axis when components are displayed

# Screenshots

Google Chrome

### DIFF
--- a/src/frontend/src/lib/components/hero/Header.svelte
+++ b/src/frontend/src/lib/components/hero/Header.svelte
@@ -1,4 +1,6 @@
 <script lang="ts">
+	import { quintOut } from 'svelte/easing';
+	import { slide, type SlideParams } from 'svelte/transition';
 	import WalletConnect from '$eth/components/wallet-connect/WalletConnect.svelte';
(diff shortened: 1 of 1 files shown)


AntonioVentilii-DFINITY (approved):
LGTM! Tks!

For future TODO, maybe we can create a `SLIDE_PARAMS_X` and `SLIDE_PARAMS_Y`, instead of a single `SLIDE_PARAMS`, if we use it more

peterpeterparker:
> For future TODO, maybe we can create a SLIDE_PARAMS_X and SLIDE_PARAMS_Y, instead of a single SLIDE_PARAMS, if we use it more

I thought about it but, in addition to `x` it also has a particularly short duration therefore was thinking that at it's point it's rather something specific that reusable. But for sure, in the future if we notice it's used again, let's do that, agree!

PR #3040, opened 2024-10-22
feat(frontend): Create listener for multiple tokens

AntonioVentilii-DFINITY:
# Motivation

The `TokenGroupCard` component created in PR #3054 is still lacking the listeners for the tokens.

Basically, some tokens requires the `Listener` component to update the balance. We cannot put it in the single-token cards that are sub-elements of the group card, because they will not be mounted if the group is not expanded.

So, to have the correct `Listener` objects, we will not wrap the cards with the `Listener`, but we wrap the entire Token Group card with a new component `MultipleListeners`.

This component is simply a stack of single-token `Listeners`, but only the ones that are required.

# Changes

- Create util to map a Listener to a specific token: `BTC` --> `BtcListener`, for example.
- Create component MultipleListener to stack the required `Listeners` given a list of tokens.
- Wrap the token group card with the new `Listener` component.

# Test

Created some tests for the util function.


peterpeterparker (commented):
I lack a bit of context to review this notably on how and when it will be effectivelly used. 
If it's not a blocker, I'll review it in few PRs once it's a bit more clear.

AntonioVentilii-DFINITY:
[OISY-69]

AntonioVentilii-DFINITY:
@peterpeterparker maybe we are ready for another review: I am showing the place we are using the new component and I changed the description

peterpeterparker (commented):
Given the time constrain, agree no other choice. LGTM, thx

AntonioVentilii-DFINITY:
@peterpeterparker may you please re-review? After our offline discussion, I changed it a bit to make it slightly better (I think). WDYT?

peterpeterparker (approved):
Better thanks!

PR #3028, opened 2024-10-22
feat(frontend): Token grouping visuals

darioloetsc:
# Motivation

Display grouped tokens in the list with new visuals.

# Changes

- Add token group view
- Trigger token grouping
- Fix styling based on main changes / merge conflicts

# Tests

No additional tests

PR #3041, opened 2024-10-22
feat(frontend): Provide token count in Token logo

AntonioVentilii-DFINITY:
# Motivation

For PR #3028, to group tokens, we want to show the number of tokens grouped in a certain token card.

So, we refactor to have a prop that defines the sub-logo of `TokenLogo`, defaulting to `undefined`.

Note: we are checking for a number greater than 0 (and not greater than 1), because supposedly this feature will be used only for groups of 2 or more elements. However, we want to leave the choice of creating a group with 1 element if needed/required.

Screenshot 2024-10-22 at 14 41 55


src/frontend/src/lib/components/tokens/TokenLogo.svelte
```diff
@@ -1,11 +1,13 @@
 <script lang="ts">
+	import { nonNullish } from '@dfinity/utils';
 	import Logo from '$lib/components/ui/Logo.svelte';
 	import { i18n } from '$lib/stores/i18n.store';
 	import type { Token } from '$lib/types/token';
 	import { replacePlaceholders } from '$lib/utils/i18n.utils';
 
 	export let token: Token;
 	export let color: 'dust' | 'off-white' | 'white' = 'dust';
+	export let tokenCount: number | undefined = undefined;
```
peterpeterparker, line 10: Should we have a single property for both count and network because the first exclude the second. In addition, `showNetworkIcon` is by default to `true` so it's a bit misleading to have it default but hidden if token count is set. WDYT?

src/frontend/src/lib/components/tokens/TokenLogo.svelte
```diff
@@ -25,7 +27,13 @@
 		{color}
 		{ring}
 	/>
-	{#if showNetworkIcon}
+	{#if nonNullish(tokenCount) && tokenCount > 0}
+		<span
```
peterpeterparker, line 31: a11y?

src/frontend/src/lib/components/tokens/TokenLogo.svelte
```diff
@@ -25,7 +27,13 @@
...
+			class="token-count absolute -right-2.5 bottom-0 flex h-6 w-6 items-center justify-center rounded-full border-[0.5px] border-light-grey bg-white text-sm font-semibold"
```
peterpeterparker, line 32: `text--[var(--color-secondary)]` instead of `token-count` ?

AntonioVentilii-DFINITY: for sure, changed!

src/frontend/src/lib/components/tokens/TokenLogo.svelte
AntonioVentilii-DFINITY: sorry, forgot! Have a look, i put 

`"token_group_number": "Total number of tokens in the group for native token $token"`

src/frontend/src/lib/components/tokens/TokenLogo.svelte
AntonioVentilii-DFINITY: yes, make sense: I refactored a bit, maybe check last commit, WDYT?

src/frontend/src/lib/components/tokens/TokenLogo.svelte
AntonioVentilii-DFINITY: or shall i add the number there too?

`"token_group_number": "There are $count tokens in the group for native token $token"`

peterpeterparker: I guess the current implementation - `"token_group_number": "Total number of tokens in the group for native token $token"` - is fine because the number itself contained in the span is also readable. Thx.

peterpeterparker (approved):
LGTM, thx

AntonioVentilii-DFINITY:
[OISY-69]

PR #3043, opened 2024-10-22
refactor(frontend): Include native token in token group property

AntonioVentilii-DFINITY:
# Motivation

It is useful to have the native token of the token group as defined prop, even if theoretically it is always the first element of the list, just to have a place to get it in an "obvious" way.

Note: in the same PR, we adjust the breakdown of the tests for function `groupTokensByTwin`, related to this change.


src/frontend/src/lib/types/token.ts
```diff
@@ -71,6 +71,7 @@ export type OptionTokenUi = Option<TokenUi>;
 
 export interface TokenGroupUi {
 	header: TokenMetadata;
+	nativeToken: TokenUi;
```
peterpeterparker, line 74: We don't need `nativeNetwork` anymore then? We add a TODO to remove it?

peterpeterparker: Likewise, do we still need `header` afterwards? TODO to be removed?

AntonioVentilii-DFINITY: i would remove it afterwards, but YES we don't need them anymore. Adding a TODO

peterpeterparker (approved):
LGTM, thx

AntonioVentilii-DFINITY:
[OISY-69]

PR #3046, opened 2024-10-22
feat(frontend): Carousel Controls components

DenysKarmazynDFINITY:
# Motivation

This PR includes Carousel Controls component.

https://github.com/user-attachments/assets/f5705783-55df-4dfa-b6cb-f784c5f17da5

### DIFF
--- a/src/frontend/src/lib/components/carousel/Controls.svelte
+++ b/src/frontend/src/lib/components/carousel/Controls.svelte
@@ -1,0 +1,27 @@
+<script lang="ts">
+	import { IconWest, IconEast } from '@dfinity/gix-components';
+	import { i18n } from '$lib/stores/i18n.store';
+
+	export let onNextClick: () => void;
+	export let onPreviousClick: () => void;
+
+	const classes =
+		'w-7 h-7 bg-white rounded-md shadow p-1 hover:text-blue-ribbon transition-colors duration-700';
+</script>
+
+<div class="flex w-[25%] items-center justify-center">
+	<button
+		aria-label={$i18n.carousel.text.prev_slide}
+		class={`${classes} mr-1`}
+		on:click={() => onPreviousClick()}
+	>
+		<IconWest />
+	</button>
+	<button
+		aria-label={$i18n.carousel.text.next_slide}
+		class={classes}
+		on:click={() => onNextClick()}
+	>
+		<IconEast />
+	</button>
+</div>
--- a/src/frontend/src/lib/components/carousel/Controls.svelte
+++ b/src/frontend/src/lib/components/carousel/Controls.svelte
@@ -1,27 +1,17 @@
 <script lang="ts">
 	import { IconWest, IconEast } from '@dfinity/gix-components';
+	import ButtonControl from '$lib/components/ui/ButtonControl.svelte';
 	import { i18n } from '$lib/stores/i18n.store';
 
-	export let onNextClick: () => void;
-	export let onPreviousClick: () => void;
-
-	const classes =
-		'w-7 h-7 bg-white rounded-md shadow p-1 hover:text-blue-ribbon transition-colors duration-700';
+	export let onNext: () => void;
+	export let onPrevious: () => void;
 </script>
 
 <div class="flex w-[25%] items-center justify-center">
-	<button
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/carousel/Controls.svelte
```diff
@@ -0,0 +1,27 @@
+<script lang="ts">
+	import { IconWest, IconEast } from '@dfinity/gix-components';
+	import { i18n } from '$lib/stores/i18n.store';
+
+	export let onNextClick: () => void;
```
peterpeterparker, line 5: What do you think of removing the suffix `Click` for a bit less verbosity?

src/frontend/src/lib/components/carousel/Controls.svelte
```diff
@@ -0,0 +1,27 @@
...
+	export let onPreviousClick: () => void;
+
+	const classes =
```
peterpeterparker, line 8: What about creating a component `ButtonControl`? We generally do not have a pattern in which we assign classes as const.

src/frontend/src/lib/components/carousel/Controls.svelte
```diff
@@ -0,0 +1,27 @@
...
+		'w-7 h-7 bg-white rounded-md shadow p-1 hover:text-blue-ribbon transition-colors duration-700';
+</script>
+
+<div class="flex w-[25%] items-center justify-center">
+	<button
+		aria-label={$i18n.carousel.text.prev_slide}
+		class={`${classes} mr-1`}
+		on:click={() => onPreviousClick()}
```
peterpeterparker, line 16: `on:click={onPreviousClick}` would do the job if you want to spare some characters

src/frontend/src/lib/components/carousel/Controls.svelte
DenysKarmazynDFINITY: Makes sense, create a new button component as suggested.

src/frontend/src/lib/components/carousel/Controls.svelte
DenysKarmazynDFINITY: I definitely do 🙂 Adjusted.

src/frontend/src/lib/components/ui/ButtonControl.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+	export let ariaLabel: string;
+	export let styleClass: string | undefined = undefined;
+</script>
+
+<button
+	aria-label={ariaLabel}
+	class={`${styleClass} h-7 w-7 rounded-md bg-white p-1 shadow transition-colors duration-700 hover:text-blue-ribbon`}
```
peterpeterparker, line 8: `active:text-blue-ribbon` too?

peterpeterparker, line 8: `styleClass ?? ""` otherwise are we maybe ending with a class name "undefined" in the DOM?

DenysKarmazynDFINITY: Added.

DenysKarmazynDFINITY: True, I changed the default value for `styleClass` to `''` instead of `undefined`

src/frontend/src/lib/components/ui/ButtonControl.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+	export let ariaLabel: string;
+	export let styleClass: string | undefined = '';
```
peterpeterparker, line 3: It's a bit weird to read I have to say: `styleClass` can be undefined but is never undefined 🤪. One consumer can set `styleClass={undefined}` and it's still renders "undefined" in the DOM.

I suggest to fallback to `""` within the class assignation.

peterpeterparker (approved):
LGTM, thx

Please have a look at my comment before merge.

src/frontend/src/lib/components/ui/ButtonControl.svelte
DenysKarmazynDFINITY: Ok 👍

PR #3048, opened 2024-10-22
feat(frontend): remove header Back

peterpeterparker:
# Motivation

The new route behavior `/explore` introduced an inconsistency in the navigation in regards of `/settings`. The former kept the OISY logo in the header while the latter triggered the display of a "Back" button instead of the logo in the header.

After discussion with Artem, we decided to remove the "Back" button in the "Header" to resolve the consistency issues.

# Changes

- Remove usage of `Back` in `Header`
- Add a TODO about the pages and navigation for a later review as the topic feels a bit messy


AntonioVentilii-DFINITY (approved):
LGTM! TKs!

P.S. Good TODO! necessary indeed

PR #3050, opened 2024-10-22
chore(frontend): format i18n type file

AntonioVentilii-DFINITY:
# Motivation

Maybe in some PR, an unformatted `i18n` type file slipped in. This is not checked by the format CI, since it is an auto-generated file.

# Future Todo

- [ ] Add CI to check auto-generated files like this one (I believe it already exists something similar for the backend)
- [ ] Add a CI to auto-commit this?


AntonioVentilii-DFINITY:
@inc-man tagging you here just to have some context of the TODO we discussed offline

peterpeterparker (approved):
Thx!

PR #3051, opened 2024-10-22
refactor(build): Change tailwind config to Typescript

AntonioVentilii-DFINITY:
# Motivation

Just to use more Typescript, we change the Tailwind configs to a Typescript file as suggested by [documentation](https://tailwindcss.com/docs/configuration#using-esm-or-type-script).


peterpeterparker (approved):
Thanks!

PR #3052, opened 2024-10-22
test(frontend): Add test ID for token group in TokenCard

AntonioVentilii-DFINITY:
# Motivation

The component `TokenCard` will be re-used for token groups too. So we create a test ID based on its usage: either as single-token card or as header token card for a group.


peterpeterparker (approved):
LGTM, thx

PR #3053, opened 2024-10-22
feat(frontend): Introduce OisySymbol prop for appearence interface

AntonioVentilii-DFINITY:
# Motivation

For grouping tokens, we are going to re-use the `TokenCard` component that takes a few props of a given token, for example `name` and `symbol`. However, for the group, we will provide some different data to be shown, just for appearance sake.

For prop `name`, we already have `oisyName` substitute to show something different in place. But, for `symbol`, there is nothing.

So, similar to `oisyName`, we create an optional prop `oisySymbol` to be provided if we want to show some substitute to the usual `token.symbol`, just for different appearance.

### DIFF
--- a/src/frontend/src/lib/types/token.ts
+++ b/src/frontend/src/lib/types/token.ts
@@ -26,11 +26,16 @@ export interface TokenMetadata {
 }
 
 export interface TokenAppearance {
+	oisySymbol?: TokenOisySymbol;
 	oisyName?: TokenOisyName;
 }
 
+export interface TokenOisySymbol {
+	oisySymbol: string;
+}
+
 export interface TokenOisyName {
-	prefix: string | undefined;
+	prefix?: string | undefined;
 	oisyName: string;
 }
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/types/token.ts
```diff
@@ -26,11 +26,16 @@ export interface TokenMetadata {
...
 	oisyName?: TokenOisyName;
 }
 
+export interface TokenOisySymbol {
+	oisySymbol: string;
+}
+
 export interface TokenOisyName {
-	prefix: string | undefined;
+	prefix?: string | undefined;
```
peterpeterparker, line 38: this change is expected?

AntonioVentilii-DFINITY: ah yes, sorry, I forgot to mention it in the description, but it is sort of unrelated, just a "since we are here..."

Do you prefer it in another PR?

peterpeterparker: This PR is ok.

Since we are here, you didn't want to shorten it to `prefix?: string` or keeping the `| undefined` is on purpose?

AntonioVentilii-DFINITY: nono, you are right, it is better

peterpeterparker (approved):
LGTM, thx

AntonioVentilii-DFINITY:
[OISY-69]

PR #3055, opened 2024-10-22
feat(frontend): Remove redundant prop to show back button in Hero

AntonioVentilii-DFINITY:
# Motivation

The `Back` button showed in the Hero was dependent on prop `back` provided to the Hero. However, the same button is dependent on `summary` prop by its parent.

Both `summary` and `back` are `true` if the user is in the `transactions` page (and both `false` otherwise): that make sense, since the `Back` button is shown only when an user is visiting that page for a specific token.

However, having both variables is redundant, since `summary` is enough to trigger everything that should be shown in the `transactions` page.


peterpeterparker (approved):
LGTM, thx

PR #3057, opened 2024-10-23
feat(frontend): qr code min height to prevent modal glitch

peterpeterparker:
# Motivation

The QR code has no minimum height. As a result, when the user navigates while the QR dependencies are not yet loaded, the modal briefly resizes, causing a visual glitch.

# Changes

- Set a min height for the QR code

# Screenshots

Before:

glitch

Fix:

fi


AntonioVentilii-DFINITY (approved):
Tks

PR #3058, opened 2024-10-23
refactor(frontend): extract qr code address content component

peterpeterparker:
# Motivation

Two components that display a QR-code where using a similar content - a qr code, and address and a copy button. As we have to update their UI, it felt like the good time to extract a common component.

Plus, I spotted a missing i18n key usage.

# Changes

- Extract new component `ReceiveAddressQRCodeContent` to present the qr code and address


AntonioVentilii-DFINITY (approved):
LGTM tks

PR #3059, opened 2024-10-23
fix(ci): fix run-name and remove ic

inc-man:
# Motivation

We don't need the workflow to deploy to ic, since we do that with orbit. Additionally we set defaults so the run-name will be fixed.

# Changes

- remove ic related code
- add default choices


AntonioVentilii-DFINITY (approved):
Tks

PR #3060, opened 2024-10-23
feat(frontend): add token group store

darioloetsc:
# Motivation

Introduce a token group store to preserve expanded settings in the grouping view.

# Changes

- Add token group store

# Tests

### DIFF
--- a/src/frontend/src/lib/stores/token-group.store.ts
+++ b/src/frontend/src/lib/stores/token-group.store.ts
@@ -1,0 +1,11 @@
+import { initStorageStore} from '$lib/stores/storage.store';
+
+export interface TokenGroupData {
+	expanded: boolean;
+}
+
+export interface TokenGroupsData {
+	[key: string]: TokenGroupData;
+}
+
+export const tokenGroupStore = initStorageStore<TokenGroupsData>({ key: 'token-groups' });
--- a/src/frontend/src/lib/stores/token-group.store.ts
+++ b/src/frontend/src/lib/stores/token-group.store.ts
@@ -1,11 +1,13 @@
-import { initStorageStore } from '$lib/stores/storage.store';
+import { initCertifiedSetterStore } from '$lib/stores/certified-setter.store';
 
 export interface TokenGroupData {
 	expanded: boolean;
 }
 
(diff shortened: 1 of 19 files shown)


src/frontend/src/lib/stores/token-group.store.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export interface TokenGroupData {
+	expanded: boolean;
+}
+
+export interface TokenGroupsData {
+	[key: string]: TokenGroupData;
+}
+
+export const tokenGroupStore = initStorageStore<TokenGroupsData>({ key: 'token-groups' });
```
AntonioVentilii-DFINITY, line 11: if you plan for the keys to be TokenId (I imagine it will be the native token ID of the group), you can use directly

`export const tokenGroupStore = initCertifiedSetterStore<TokenGroupData>({ key: 'token-groups' });`

defaulting the certified part to either `true` or `false`, you decide

Basically, this store will have the structure you need `[token] : data` and some useful `set` and `reset` functions

darioloetsc: sounds good, need to refactore the store to deal with generic key / symbol. Currently only tokenId is supported.

peterpeterparker: `symbol` can be persisted in the local storage?

peterpeterparker:
Why is there suddently 19 files modified in this PR when it was first provided with a single file change? Is this a draft?

darioloetsc:
@peterpeterparker—The proposed store needs to be made generic. The 19 additional changed files are type adjustments.

peterpeterparker:
> The proposed 

so it originally needed to be a draft, I see. thanks for the feedback.

src/frontend/src/lib/stores/token-group.store.ts
darioloetsc: No, symbol cannot be persisted in local storage. The CertifiedSetterStore seems not to store to local storage and be implemented on TokenId (symbol).

darioloetsc:
will split in two, and refactoring is bigger than expected

PR #3062, opened 2024-10-23
feat(frontend): Include balances in token group

AntonioVentilii-DFINITY:
# Motivation

Another useful PR for grouping tokens: we need to have the balances (both token and USD) for each group. Note that for nullish values for both tokens in group, we keep them as they are but prioritizing the native token. So for example:

### Case 1

- native token --> `null`
- twin token --> `undefined`
- group --> `null`

### Case 2

- native token --> `undefined`
- twin token --> `null`
- group --> `undefined`

# Changes

- Change `TokenGroupUi` to be a type instead of interface and add the financial data props.
- Add balances calculation to `createTokenGroup`.

# Tests

No changes in local replica, since we are not using those values yet.


darioloetsc (approved):

AntonioVentilii-DFINITY:
[OISY-69]

src/frontend/src/lib/types/token.ts
```diff
@@ -70,11 +70,11 @@ export type TokenUi = Token & TokenFinancialData;
 export type OptionTokenUi = Option<TokenUi>;
 
 // TODO: remove header and nativeNetwork, since we added nativeToken that has all their data, and they became redundant
-export interface TokenGroupUi {
+export type TokenGroupUi = {
```
peterpeterparker, line 73: unrelated: why is this type suffixed with `Ui` btw?

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -190,7 +190,19 @@ const createTokenGroup = ({
 	},
 	nativeToken,
 	nativeNetwork: nativeToken.network,
-	tokens: [nativeToken, twinToken]
+	tokens: [nativeToken, twinToken],
+	balance:
+		nonNullish(nativeToken.balance) && nonNullish(twinToken.balance)
```
peterpeterparker, line 195: - should we extract a small inner function for readability?

- can you add a comment that we assume that twin token and native token are always 1 to 1? e.g. 1 ckBTC === 1 BTC

- should we extend the test?

src/frontend/src/lib/types/token.ts
AntonioVentilii-DFINITY: it is to indicate that the group contains TokenUi and not just Token: in fact it should be used only in UI context (appearance, show the list, etc)

peterpeterparker: so, shouldn't it be called `TokenUiGroup`?

src/frontend/src/lib/utils/token.utils.ts
AntonioVentilii-DFINITY: - done! and created test

- good catch! I completely missed that... 

- see first point

please review it again, because i separated the smaller functions and created tests about it

src/frontend/src/lib/types/token.ts
AntonioVentilii-DFINITY: right! make sense! can i do it in another PR? because I am pretty sure it will change some other components and should be indeed a different topic

peterpeterparker: definitely another PR, thanks

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -141,6 +141,42 @@ export const mapTokenUi = ({
 	})
 });
 
+/** Function to sum the balances of two tokens.
+ *
+ * If the decimals of the tokens are the same, the balances are added together.
+ * If the decimals are different, the function returns null.
```
peterpeterparker, line 147: Maybe I'm misleading it but,`null` is maybe not specified as return type of the function?

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -141,6 +141,42 @@ export const mapTokenUi = ({
...
+ * If one of the balances is nullish, the function returns the other balance.
+ * If both balances are nullish, the function prioritize the first token
+ * NOTE: the function assumes that the two tokens are always 1:1 twins, for example BTC and ckBTC, or ETH and SepoliaETH
+ *
+ * @param token1
+ * @param token2
+ * @returns The sum of the balances or nullish value.
+ */
+export const sumTokenBalances = ([token1, token2]: TokenUi[]): TokenUi['balance'] =>
```
peterpeterparker, line 156: - Should we add a guard if there are more than two tokens?

- If exactly two tokens, should the array be actually typed as `[TokenUi, TokenUi]`?

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -141,6 +141,42 @@ export const mapTokenUi = ({
...
+	nonNullish(token1.balance) && nonNullish(token2.balance)
+		? token1.decimals === token2.decimals
+			? token1.balance.add(token2.balance)
+			: null
+		: isNullish(token2.balance)
```
peterpeterparker, line 161: what if token2 balance is zero but token1 balance is >0 ?

peterpeterparker: same question for usd

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -141,6 +141,42 @@ export const mapTokenUi = ({
...
+			? token1.balance
+			: token2.balance;
+
+/** Function to sum the USD balances of two tokens.
+ *
+ * If one of the balances is nullish, the function returns the other balance.
```
peterpeterparker, line 167: Does it really matter here? It's all dollars, shouldn't the array just be reduced to sum all usd balance?

AntonioVentilii-DFINITY: ah! make sense what about this?

```typescript
export const sumTokenUsdBalances = ([token1, token2]: TokenUi[]): TokenUi['usdBalance'] =>
	nonNullish(token1.usdBalance) || nonNullish(token2.usdBalance)
		? (token1.usdBalance ?? 0) + (token2.usdBalance ?? 0)
		: undefined;
```
		
because `usdBalance` can be either `number` or `undefined`

AntonioVentilii-DFINITY: i would prefer to leave `undefined` if both are `undefined`

src/frontend/src/lib/utils/token.utils.ts
AntonioVentilii-DFINITY: it will be added in both cases, so it will stay the same as the other. I didn't put the check for a zero balance, do you think it is necessary? Meaning, either way, summing 0 is ok no?

src/frontend/src/lib/utils/token.utils.ts
AntonioVentilii-DFINITY: you are right! I was already thinking about multiple tokens in the group! Changed to `[TokenUi, TokenUi]` that should act as guard too, no?

src/frontend/src/lib/utils/token.utils.ts
AntonioVentilii-DFINITY: I defined the return type as `TokenUi['balance']` which definition is 

Screenshot 2024-10-23 at 13 24 26

AntonioVentilii-DFINITY: I have a few tests to check the different returns for `null` and `undefined`

peterpeterparker: Cool

src/frontend/src/lib/utils/token.utils.ts
peterpeterparker: Yes, works for me.

src/frontend/src/lib/utils/token.utils.ts
peterpeterparker: I oversaw it, good point.

src/frontend/src/lib/utils/token.utils.ts
peterpeterparker: For consistency with `sumTokenBalances`, shouldn't `sumTokenUsdBalances` returns `null` then if both tokens are `undefined`?

Or should the function always returns a zero or value, hence suggesting a reduce and sum?

AntonioVentilii-DFINITY: Good question...

The only issue I see is to keep it consistent with definition of `TokenUi.usdBalance` that can be either `number` or `undefined`, but not `null`.

When it is `undefined`, we show it differently in the page `"$-"`, so I thought to keep that same behavior: if both `usdBalance` are `undefined`, the sum is `undefined` and it will showed in the same way `"$-"`

The alternative could be to show `zero`, as you suggested, but I am not sure it could be `null`... WDYT?

peterpeterparker: Makes sense, thanks for the explanation.

src/frontend/src/lib/types/token.ts
AntonioVentilii-DFINITY: Done:

- PR #3075

PR #3064, opened 2024-10-23
feat(frontend): network icon black white for receive

peterpeterparker:
# Motivation

The network icon displayed in the receive modals should be black and white according Figma.


loki344 (approved):
Thanks

PR #3065, opened 2024-10-23
feat(frontend): Add padding prop to Button component

AntonioVentilii-DFINITY:
# Motivation

To align more with the Design structure and to be more re-usable, we add an optional parameter to component Button, that is padding size, mapped as such it reflects the Design team plan:

```typescript
const paddings:Record<PaddingSize, string> = {
  xs: 'p-1',
  sm: 'px-3 py-2',
  md: 'px-4 py-3',
  lg: 'px-5 py-4',
  xl: 'p-5',
  none: 'p-0',
};
```

It will default to `lg` that is more or less the current default value defined in CSS.

- CSS --> `padding-x = 21.6px ; padding-y = 14.4px`
- new one --> `padding-x = 20px ; padding-y = 16px`

Note that the previous definition of `paddingSmall` is now defined as padding `sm`.

# Changes

- Create new type for padding size.
- Define padding size map in Button component.
- Replace `paddingSmall` usage with `padding='sm'`
- Remove `paddingSmall` prop.

# Tests

Small visual change in the buttons

### Before

Screenshot 2024-10-23 at 11 03 33

### After

Screenshot 2024-10-23 at 11 03 57

### DIFF
--- a/src/frontend/src/lib/components/dapps/DappPromoBanner.svelte
+++ b/src/frontend/src/lib/components/dapps/DappPromoBanner.svelte
@@ -34,7 +34,7 @@
 				<h4 class="text-white">{dAppDescription.name}</h4>
 			</div>
 
-			<Button padding="md" styleClass="flex-grow-0 text-sm" colorStyle="secondary" on:click>
+			<Button padding="sm" styleClass="flex-grow-0 text-sm" colorStyle="secondary" on:click>
 				{$i18n.core.text.view}
 			</Button>
 		</div>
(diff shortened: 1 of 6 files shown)


loki344:
This might be opinionated so feel free to ignore: Going down the road of implementing a custom button, I would try to mimic established component frameworks for these things. TBH I've never encountered a padding prop there. Normally you go with a size property, which then controls padding and also font-size. This approach would also have the benefit of matching the figma better, as there are also different button sizes (which then implement different paddings and font-sizes). 


image

AntonioVentilii-DFINITY:
@loki344 yes yes, completely agree, this was just a test branch for me

PR #3066, opened 2024-10-23
fix(ci): fix release process automation

inc-man:
# Motivation

During previous runs the new release branch was set as base and the workflow didn't complete as intended. We want that fixed.

# Changes

- added a new checkout so after creating the release branch we are back at main for the PR.


.github/workflows/bump-version.yml
AntonioVentilii-DFINITY, line 1: is the step `Commit version bump on release` branch needed? I think the `create-pr` action already takes care of everything, no? At least, it does it in the other workflows: have a look at `Remove Unused Svelte Files`

inc-man: you are absolutely correct. I did not read the create-pr action correctly. Removed everything that is not necessary.

AntonioVentilii-DFINITY (approved):
LGTM! Tks

PR #3067, opened 2024-10-23
feat(frontend): Carousel utils

DenysKarmazynDFINITY:
# Motivation

This PR includes utils needed for the Carousel component. The idea is to take provided to `Carousel` as children slides (can be any HTML elements, components, etc.) and to create a slider frame - a div that has `slidesNumber * slideWidth` width (only active slide is visible) and includes all slides with equal (percentage) size. The other 3 functions are small helpers to enable/disable (we need to disable animation when we do last-to-first and first-to-last switches) animation and to change a current slide by applying `transform` to the frame.   

Please let me know if you need more context to review this PR (e.g. you also want to include the UI component itself here).

https://github.com/user-attachments/assets/0cf5618b-1e26-45d2-8b9f-b2d1c3ce771c

### DIFF
--- a/src/frontend/src/lib/utils/carousel.utils.ts
+++ b/src/frontend/src/lib/utils/carousel.utils.ts
@@ -1,0 +1,96 @@
+import { isNullish } from '@dfinity/utils';
+
+interface CommonParams {
+	sliderFrame: HTMLElement | undefined;
+}
+
+/**
+ * A util for building carousel slider frame item.
+ */
+export const buildCarouselSliderFrameItem = ({
+	slide,
+	totalSlides
+}: {
+	slide: Node;
+	totalSlides: number;
+}): HTMLDivElement => {
+	// Create slider frame item and apply styling
+	const frameItem = document.createElement('div');
+	frameItem.style.width = `${100 / totalSlides}%`;
+
+	frameItem.appendChild(slide);
+
+	return frameItem;
+};
+
+/**
+ * A util for building carousel slider frame.
+ */
+export const buildCarouselSliderFrame = ({
+	slides,
+	slideWidth
+}: {
+	slides: Node[];
+	slideWidth: number;
+}): HTMLDivElement => {
+	// We add the last slide as the first element to properly animate last-to-first slide transition.
+	// The same applies to the last element - we use the first slide as the last item of the array.
+	const extendedSlides = [slides[slides.length - 1], ...slides, slides[0]];
+	const totalSlides = extendedSlides.length;
+
+	// Create slider frame and apply styling
+	const sliderFrame = document.createElement('div');
+	sliderFrame.style.width = `${slideWidth * totalSlides}px`;
+	sliderFrame.style.display = 'flex';
+
+	extendedSlides.forEach((slide) => {
+		sliderFrame.appendChild(
+			buildCarouselSliderFrameItem({
+				slide: slide.cloneNode(true),
+				totalSlides
+			})
+		);
+	});
+
+	return sliderFrame;
+};
+
+/**
+ * Disable transition for the provided slider frame.
+ */
+export const disableTransition = ({ sliderFrame }: CommonParams) => {
+	if (isNullish(sliderFrame)) {
+		return;
+	}
+
+	sliderFrame.style.transition = 'none';
+};
+
+/**
+ * Enable transition for the provided slider frame.
+ */
+export const enableTransition = ({
+	sliderFrame,
+	duration,
+	easing
+}: { duration: number; easing: string } & CommonParams) => {
+	if (isNullish(sliderFrame)) {
+		return;
+	}
+
+	sliderFrame.style.transition = `all ${duration}ms ${easing}`;
+};
+
+/**
+ * Update visible part of the provided slider frame.
+ */
+export const updateSliderFrameVisiblePart = ({
+	animateTo,
+	sliderFrame
+}: { animateTo: number } & CommonParams): void => {
+	if (isNullish(sliderFrame)) {
+		return;
+	}
+
+	sliderFrame.style.transform = `translate3d(${animateTo}px, 0, 0)`;
+};
--- a/src/frontend/src/lib/utils/carousel.utils.ts
+++ b/src/frontend/src/lib/utils/carousel.utils.ts
@@ -16,7 +16,10 @@ export const buildCarouselSliderFrameItem = ({
 }): HTMLDivElement => {
 	// Create slider frame item and apply styling
 	const frameItem = document.createElement('div');
-	frameItem.style.width = `${100 / totalSlides}%`;
+
+	if (totalSlides > 0) {
+		frameItem.style.width = `${100 / totalSlides}%`;
+	}
 
 	frameItem.appendChild(slide);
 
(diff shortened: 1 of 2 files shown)


peterpeterparker (commented):
Before reviewing the code: have you considered other approaches that does not require manipulating the DOM and calculating width, such as using CSS scroll snap?

DenysKarmazynDFINITY:
@peterpeterparker yes, the CSS approach was my first thought, but, unfortunately, it doesn't align with the requirements for the wanted carousel component. To name a few reasons why we do need to manipulate the DOM: the last-to-first and first-to-last switches won't work like it's shown on video (and requested by the product) without manually adding 2 additional slides; synced animation between slides and indicators won't be possible since snap-scroll easing/duration differs from browser to browser and it's not possible to adjust it. Also, the current approach gives us more control, hence allows us to extend Carousel in the future with further functionality. 

P.S.: it's the first time I miss `.tsx` - in React, the code could have been much simpler.

peterpeterparker:
Thanks for the explanation. I then have a follow-up question: why manipulating the DOM manually and not resolving the implementation using Svelte?

Fortunatelly Svelte has not virtual DOM but, I'm still always a bit relunctant to manipulate the DOM directly, feels like bypassing the framework we are using in first place.

DenysKarmazynDFINITY:
@peterpeterparker also a valid question, and it was my initial intention to do so, but it appeared it's not possible to work with `slots` in Svelte similar to `React.children`: looping through them, getting first-last elements, etc., Here is a big thread on why they don't want to add a native support for it https://github.com/sveltejs/svelte/issues/5381. It could have been solved by passing slides to carousel in some different than `slot` way, but it would decrease the flexibility of the component - my goal was to make `Carousel` completely unaware of what type of slides are passed down.

peterpeterparker:
Feels like we should bump Svelte v5 so that you resolve this with the new `Snippet` pattern.

Do you have a WIP branch to share in which those utils are used? Just to grasp a better sense of how they will be used.

DenysKarmazynDFINITY:
@peterpeterparker sure, this is the PR with the Carousel component - https://github.com/dfinity/oisy-wallet/pull/3069. There will be one more with a small adjustments to the already merged Indicators components, after which the whole feature will be available in `main` as shown on the video.

loki344:
maybe take a look at: https://svelte.dev/playground/c379be902322415bb1e10f855442f40a?version=4.2.1
Seems to use transform magic

loki344:
Or.. use a existing library instead..? https://www.embla-carousel.com/examples/predefined/#auto-scroll

src/frontend/src/lib/utils/carousel.utils.ts
```diff
@@ -0,0 +1,96 @@
...
+export const buildCarouselSliderFrameItem = ({
+	slide,
+	totalSlides
+}: {
+	slide: Node;
+	totalSlides: number;
+}): HTMLDivElement => {
+	// Create slider frame item and apply styling
+	const frameItem = document.createElement('div');
+	frameItem.style.width = `${100 / totalSlides}%`;
```
peterpeterparker, line 19: should we prevent division by zero?

src/frontend/src/lib/utils/carousel.utils.ts
```diff
@@ -0,0 +1,96 @@
...
+	// The same applies to the last element - we use the first slide as the last item of the array.
+	const extendedSlides = [slides[slides.length - 1], ...slides, slides[0]];
+	const totalSlides = extendedSlides.length;
+
+	// Create slider frame and apply styling
+	const sliderFrame = document.createElement('div');
+	sliderFrame.style.width = `${slideWidth * totalSlides}px`;
+	sliderFrame.style.display = 'flex';
+
+	extendedSlides.forEach((slide) => {
```
peterpeterparker, line 46: `append()` or something like that accept appending multiple elements at one - no need of `forEach`

src/frontend/src/lib/utils/carousel.utils.ts
```diff
@@ -0,0 +1,96 @@
...
+	slides: Node[];
+	slideWidth: number;
+}): HTMLDivElement => {
+	// We add the last slide as the first element to properly animate last-to-first slide transition.
+	// The same applies to the last element - we use the first slide as the last item of the array.
+	const extendedSlides = [slides[slides.length - 1], ...slides, slides[0]];
+	const totalSlides = extendedSlides.length;
+
+	// Create slider frame and apply styling
+	const sliderFrame = document.createElement('div');
```
peterpeterparker, line 42: Do we need to create this element? It can be an existing reference?

src/frontend/src/lib/utils/carousel.utils.ts
```diff
@@ -0,0 +1,96 @@
...
+	sliderFrame.style.width = `${slideWidth * totalSlides}px`;
```
peterpeterparker, line 43: It needs a width even if each slides is already defined with their width?

src/frontend/src/lib/utils/carousel.utils.ts
```diff
@@ -0,0 +1,96 @@
...
+	sliderFrame.style.display = 'flex';
+
+	extendedSlides.forEach((slide) => {
+		sliderFrame.appendChild(
+			buildCarouselSliderFrameItem({
+				slide: slide.cloneNode(true),
```
peterpeterparker, line 49: We clone the slide? Shouldn't we just move those?

peterpeterparker (commented):
A first round of comments 😉

src/frontend/src/lib/utils/carousel.utils.ts
DenysKarmazynDFINITY: Only the new 2 elements (first and last slides) must be cloned. I adjusted the implementation.

src/frontend/src/lib/utils/carousel.utils.ts
DenysKarmazynDFINITY: With some adjustments - yes. That was a good idea 👍

src/frontend/src/lib/utils/carousel.utils.ts
DenysKarmazynDFINITY: Done.

src/frontend/src/lib/utils/carousel.utils.ts
DenysKarmazynDFINITY: Done.

src/frontend/src/lib/utils/carousel.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+ */
+export const buildCarouselSliderFrameItem = ({
+	slide,
+	totalSlides
+}: {
+	slide: Node;
+	totalSlides: number;
+}): HTMLDivElement => {
+	// Create slider frame item and apply styling
+	const frameItem = document.createElement('div');
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Done.

src/frontend/src/lib/utils/carousel.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+	slideWidth
+}: {
+	slides: Node[];
+	slideWidth: number;
+} & CommonParams) => {
+	if (isNullish(sliderFrame)) {
+		return;
+	}
+
+	// We add the last slide as the first element to properly animate last-to-first slide transition.
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Added additional TODO which suggests to use either Svelte 5 (when available) or potential CSS solution to avoid the hackiness 🙂

src/frontend/src/lib/utils/carousel.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+	// The same applies to the last element - we use the first slide as the last item of the array.
+	const extendedSlides = [
+		slides[slides.length - 1].cloneNode(true),
+		...slides,
+		slides[0].cloneNode(true)
+	];
+	const totalSlides = extendedSlides.length;
+
+	// Calculate and set width
+	sliderFrame.style.width = `${slideWidth * totalSlides}px`;
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: This could work 👍  I'll check it and create a new PR if everything works as expected.

src/frontend/src/lib/utils/carousel.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+		return;
+	}
+
+	sliderFrame.style.transition = `all ${duration}ms ${easing}`;
+};
+
+/**
+ * Update visible part of the provided slider frame.
+ */
+export const updateSliderFrameVisiblePart = ({
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Done.

PR #3069, opened 2024-10-23
feat(frontend): Carousel component

DenysKarmazynDFINITY:
# Motivation

This PR includes the main Carousel component. It is based on the Carouse-utils branch, it will be switched to `main` after that PR is merged.

https://github.com/user-attachments/assets/1fd6da41-1da1-4ad4-b6ba-0672942006d6

### DIFF
--- a/src/frontend/src/lib/components/carousel/Carousel.svelte
+++ b/src/frontend/src/lib/components/carousel/Carousel.svelte
@@ -5,7 +5,7 @@
 	import Indicators from '$lib/components/carousel/Indicators.svelte';
 	import {
 		updateSliderFrameVisiblePart,
-		buildCarouselSliderFrame,
+		extendCarouselSliderFrame,
 		disableTransition,
 		enableTransition
 	} from '$lib/utils/carousel.utils';
@@ -73,26 +73,26 @@
 	 * Build slider frame and set required variables
 	 */
 	const initializeCarousel = () => {
-		if (isNullish(container)) {
+		if (isNullish(container) || isNullish(sliderFrame)) {
 			return;
 		}
 
 		// Save this data only on mount
 		if (isNullish(slides)) {
-			slides = [...container.children];
+			slides = [...sliderFrame.children];
 			totalSlides = slides.length;
 		}
 
 		containerWidth = container.offsetWidth;
 
-		sliderFrame = buildCarouselSliderFrame({
+		// Clean previous HTML if the frame is being re-built (e.g. in case of window resize)
+		sliderFrame.innerHTML = '';
+
+		extendCarouselSliderFrame({
+			sliderFrame,
 			slides,
 			slideWidth: containerWidth
 		});
-
-		// Clean previous HTML if the frame is being re-built (e.g. in case of window resize)
-		container.innerHTML = '';
-		container.appendChild(sliderFrame);
 
 		goToSlide(currentSlide);
 	};
@@ -225,8 +225,10 @@
 </script>
 
 <div class={`${styleClass ?? ''} relative overflow-hidden rounded-3xl bg-white p-4 pb-16 shadow`}>
-	<div class="overflow-hidden" bind:this={container}>
-		<slot></slot>
+	<div class="w-full overflow-hidden" bind:this={container}>
+		<div class="flex" bind:this={sliderFrame}>
+			<slot></slot>
+		</div>
 	</div>
 	<div class="absolute bottom-4 left-0 flex w-full justify-center">
 		<Indicators {onIndicatorClick} {totalSlides} {currentSlide} />
--- a/src/frontend/src/lib/components/carousel/Carousel.svelte
+++ b/src/frontend/src/lib/components/carousel/Carousel.svelte
@@ -3,12 +3,7 @@
 	import { onMount } from 'svelte';
 	import Controls from '$lib/components/carousel/Controls.svelte';
 	import Indicators from '$lib/components/carousel/Indicators.svelte';
-	import {
-		updateSliderFrameVisiblePart,
-		extendCarouselSliderFrame,
-		disableTransition,
-		enableTransition
-	} from '$lib/utils/carousel.utils';
+	import { moveSlider, extendCarouselSliderFrame } from '$lib/utils/carousel.utils';
 
 	export let autoplay = 5000;
 	export let duration = 300;
@@ -44,30 +39,30 @@
 	let slideTransformTimer: NodeJS.Timeout | undefined = undefined;
 
 	/**
-	 * A const to be used for enabling transition animation
-	 */
-	const animationParams = {
-		easing,
-		duration
-	};
-
-	/**
 	 * Initialise all required data and attach/detach events
 	 */
 	onMount(() => {
+		initializeSlides();
 		initializeCarousel();
 		initialiseAutoplayTimer();
-
-		// Resize listener to re-calculate slide frame width
-		window.addEventListener('resize', onResize);
 
 		return () => {
 			clearAutoplayTimer();
 			clearSlideTransformTimer();
-
-			window.removeEventListener('resize', onResize);
 		};
 	});
+
+	/**
+	 * Initialise slides-related vars on component mount
+	 */
+	const initializeSlides = () => {
+		if (isNullish(sliderFrame)) {
+			return;
+		}
+
+		slides = [...sliderFrame.children];
+		totalSlides = slides.length;
+	};
 
 	/**
 	 * Build slider frame and set required variables
@@ -75,12 +70,6 @@
 	const initializeCarousel = () => {
 		if (isNullish(container) || isNullish(sliderFrame)) {
 			return;
-		}
-
-		// Save this data only on mount
-		if (isNullish(slides)) {
-			slides = [...sliderFrame.children];
-			totalSlides = slides.length;
 		}
 
 		containerWidth = container.offsetWidth;
@@ -94,7 +83,11 @@
 			slideWidth: containerWidth
 		});
 
-		goToSlide(currentSlide);
+		if (nonNullish(sliderFrame.children)) {
+			goToSlide({
+				slide: currentSlide
+			});
+		}
 	};
 
 	/**
@@ -110,10 +103,8 @@
 	 * Clear autoplay timer
 	 */
 	const clearAutoplayTimer = () => {
-		if (nonNullish(autoplayTimer)) {
-			clearInterval(autoplayTimer);
-			autoplayTimer = undefined;
-		}
+		clearInterval(autoplayTimer);
+		autoplayTimer = undefined;
 	};
 
 	/**
@@ -128,10 +119,8 @@
 	 * Clear slide transform timer
 	 */
 	const clearSlideTransformTimer = () => {
-		if (nonNullish(slideTransformTimer)) {
-			clearInterval(slideTransformTimer);
-			slideTransformTimer = undefined;
-		}
+		clearInterval(slideTransformTimer);
+		slideTransformTimer = undefined;
 	};
 
 	/**
@@ -147,14 +136,14 @@
 	const goToNextSlide = () => {
 		const nextCurrentSlide = currentSlide + 1;
 
-		enableTransition({ sliderFrame, ...animationParams });
-		goToSlide(nextCurrentSlide);
+		goToSlide({
+			slide: nextCurrentSlide
+		});
 
 		if (nextCurrentSlide >= totalSlides) {
 			// Switch to the first slide without animation
 			slideTransformTimer = setTimeout(() => {
-				disableTransition({ sliderFrame });
-				goToSlide(0);
+				goToSlide({ slide: 0, withTransition: false });
 			}, duration);
 		}
 	};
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+	const animationParams = {
+		easing,
+		duration
+	};
+
+	/**
+	 * Initialise all required data and attach/detach events
+	 */
+	onMount(() => {
+		initializeCarousel();
```
peterpeterparker, line 58: The carousel is not dynamic and can only be initialized once, is this expected?

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+		initialiseAutoplayTimer();
+
+		// Resize listener to re-calculate slide frame width
+		window.addEventListener('resize', onResize);
+
+		return () => {
```
peterpeterparker, line 64: In Svelte4, returning a function here works similarly as an `onDestroy` hook?

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+
+	/**
+	 * Initialise all required data and attach/detach events
+	 */
+	onMount(() => {
+		initializeCarousel();
+		initialiseAutoplayTimer();
+
+		// Resize listener to re-calculate slide frame width
+		window.addEventListener('resize', onResize);
```
peterpeterparker, line 62: We generally use `<svelte:window>` to register global events

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+	/**
+	 * Build slider frame and set required variables
+	 */
+	const initializeCarousel = () => {
+		if (isNullish(container)) {
+			return;
+		}
+
+		// Save this data only on mount
+		if (isNullish(slides)) {
```
peterpeterparker, line 81: Should we rather extract this in a dedicated initialization function?

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+			slides = [...container.children];
+			totalSlides = slides.length;
```
peterpeterparker, line 83: Why not derive `totalSlides` from `slides`?

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+			slides,
+			slideWidth: containerWidth
+		});
+
+		// Clean previous HTML if the frame is being re-built (e.g. in case of window resize)
+		container.innerHTML = '';
+		container.appendChild(sliderFrame);
+
+		goToSlide(currentSlide);
+	};
```
peterpeterparker, line 98: Many things need to be initialized, I'm a bit afraid we end up in a partial state in which some information are set and some not. Should we prevent those kind of desynchronization?

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+		containerWidth = container.offsetWidth;
+
+		sliderFrame = buildCarouselSliderFrame({
+			slides,
+			slideWidth: containerWidth
+		});
+
+		// Clean previous HTML if the frame is being re-built (e.g. in case of window resize)
+		container.innerHTML = '';
+		container.appendChild(sliderFrame);
```
peterpeterparker, line 95: Would something like `container.replaceChild` would work instead of resetting and appending?

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+			return;
+		}
+
+		// Save this data only on mount
+		if (isNullish(slides)) {
+			slides = [...container.children];
+			totalSlides = slides.length;
+		}
+
+		containerWidth = container.offsetWidth;
```
peterpeterparker, line 86: Shouldn't we derive `containerWidth` from the container rather than imperatively setting it's value?

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+		autoplayTimer = setInterval(() => {
+			goToNextSlide();
+		}, autoplay);
+	};
+
+	/**
+	 * Clear autoplay timer
+	 */
+	const clearAutoplayTimer = () => {
+		if (nonNullish(autoplayTimer)) {
```
peterpeterparker, line 113: nitpick: I think `clearInterval` actually acccepts `undefined` parameter

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+	 */
+	const onResize = () => {
+		initializeCarousel();
+	};
+
+	/**
+	 * Switch to the next slide
+	 */
+	const goToNextSlide = () => {
+		const nextCurrentSlide = currentSlide + 1;
```
peterpeterparker, line 148: Should we prevent overflow?

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+
+		enableTransition({ sliderFrame, ...animationParams });
+		goToSlide(nextCurrentSlide);
+
+		if (nextCurrentSlide >= totalSlides) {
+			// Switch to the first slide without animation
+			slideTransformTimer = setTimeout(() => {
+				disableTransition({ sliderFrame });
+				goToSlide(0);
```
peterpeterparker, line 157: Have you thhought to passing the animation as an option of `goToSlide` to avoid having to enable or disable before the going to the slide?

```
goToSlide({ slide: 0, transition: none })
```

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -0,0 +1,235 @@
...
+
+		// We add +1 here to scroll frame to the end of the current slide
+		const offset = (currentSlide + 1) * containerWidth;
+		updateSliderFrameVisiblePart({ sliderFrame, animateTo: -offset });
+	};
+</script>
+
+<div class={`${styleClass ?? ''} relative overflow-hidden rounded-3xl bg-white p-4 pb-16 shadow`}>
+	<div class="overflow-hidden" bind:this={container}>
+		<slot></slot>
```
peterpeterparker, line 229: nitpick: `<slot />`

peterpeterparker (commented):
Not sure if I was supposed to review this one but, as it is tight with #3067, I also had a first look.

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: Yes, it only needs to be initialised on mount and on resize.

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: Yep, the return function here is called on unmount. The only thing - it cannot be async, which is perfectly fine in my case.

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: Makes sense.

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: It's used in 5 different places, so `totalSlides` is just a shortcut for `slides.length`. No strong opinion here, I can remove the var if you prefer so.

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: Since `container` is initially `undefined`, I added this var to avoid `nonNullish` check in every place where we need this value. Plus, it can only change on mount and on resize (both events will trigger `initializeCarousel`), so the var is always up-to-date. And, finally, it's again a bit shorter than using `container.offsetWidth`.

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: True, adjusted.

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: There is a check in `onNext`/`onPrevious` functions. `goToNextSlide` is indeed also used directly by the autoplay timer, but the overflow won't happen because of the next lines of this method - it will be silently (w/o animation) switched to slide number 0 after the transform animation is completed.

src/frontend/src/lib/components/carousel/Carousel.svelte
peterpeterparker: Can you add a comment somewhere because this is kind of a limitation.

src/frontend/src/lib/components/carousel/Carousel.svelte
peterpeterparker: Did not knew, thought in Svelte v4 onDestroy was required. Cool.

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: I added a small check to see if sliderFrame includes slides, couldn't come up with any other solutions here

src/frontend/src/lib/components/carousel/Carousel.svelte
peterpeterparker: What I mean was:

```
let containerWidth = 0;
$: containerWidth = container?.offsetWidth ?? 0
```

Would a less imperactive approach be interesting here?

src/frontend/src/lib/components/carousel/Carousel.svelte
peterpeterparker: Same as my other comment, I meant:

```
let totalSlides: number;
$: totalSlides = slides?.length ?? 0
```

Is this interesting here?

src/frontend/src/lib/components/carousel/Carousel.svelte
peterpeterparker: Oh that's another thing that needs to be documented - if a slide gets removed dynamically after everything is mounted then I guess the all slider is broken?

DenysKarmazynDFINITY: I don't think it can be the case, I added that check mostly to attempt to prevent the above mentioned desynchronization. If it looks confusing, I can remove it.

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: Done in this case.

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: In this case, I would keep defining containerWidth onMount (right before we generate the frame).

peterpeterparker: oki doki

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: Adding now, missed this comment.

src/frontend/src/lib/components/carousel/Carousel.svelte
peterpeterparker: What I mean is, if I understand correctly, the carousel will work fine as long as there are no dynamic changes to its content, like adding or removing slides. I'm also not sure how it would behave if one of the slides is a Svelte component that gets rendered while the carousel is mostly a static item.

Long story short, I foresee quite a potential for errors if this component is reused in any dynamic environment.

It's not a blocker for this PR, but definitely something to tackle once the due date has passed.

peterpeterparker (approved):
LGTM, thanks

I'm approving the PR, but as mentioned in my comments, I believe that the fact this component does not seem to support dynamic re-rendering could potentially lead to issues and costs in the future.

src/frontend/src/lib/components/carousel/Carousel.svelte
DenysKarmazynDFINITY: I will keep it in mind while implementing the next step where I actually use the component. I think it won't be a big deal to adjust some values to be dynamic, but in general your point totally makes sense.

PR #3072, opened 2024-10-23
style(frontend): add bg for loading phase of dapp promo banner

loki344:
# Motivation

The buttons and text of the DappPromoBanner are not visible during loading of the image. Design-team wants a simple background to provide contrast during loading time.

# Changes

- Add background
- improve backdrop

before: 
image

after:
image


loki344:
> Note however that at first sight, the text feels being not accessible - i.e. not contrasting enough.

agree - This now matches the design which might not be ideal here. I'll forward this

PR #3075, opened 2024-10-23
refactor(frontend): Rename type for token groups

AntonioVentilii-DFINITY:
# Motivation

As discussed, the more obvious name for the variable should be `TokenUiGroup` since it represents a group of `TokenUi` types.


loki344 (approved):
Thanks

PR #3077, opened 2024-10-23
style(frontend): Increase breakpoint for mobile view

AntonioVentilii-DFINITY:
# Motivation

As a recent feedback, we are increasing the breakpoint for the "mobile" view to be on `md` instead of `sm` (just for logged in).

https://github.com/user-attachments/assets/562c5f21-9f17-4bab-885d-6f122fcb3625

### DIFF
--- a/src/frontend/src/lib/components/auth/AuthGuard.svelte
+++ b/src/frontend/src/lib/components/auth/AuthGuard.svelte
@@ -7,7 +7,7 @@
 {#if $authNotSignedIn}
 	<LandingPage />
 {:else}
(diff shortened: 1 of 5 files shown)


AntonioVentilii-DFINITY:
[OISY-167]

peterpeterparker (approved):
LGTM, thx

Side note: Should we introduce a variable shorthand for this special case? If a stakeholder changes their mind in two weeks and wants to switch `md` back to `sm`, it would be easier to update.

AntonioVentilii-DFINITY:
@peterpeterparker that is a good to-do: I will put it in my backlog to check what's doable

PR #3079, opened 2024-10-23
refactor(frontend): Limit data provided to token cards

AntonioVentilii-DFINITY:
# Motivation

We want to re-use TokenCard and its related components for the token groups. However, to abstract the data provided, instead of Token, we create a set of data that are common/required in both cases: single token or token group.

This is just a refactoring PR.

# Changes

- Create type with common props required for the token card, called `CardData`
- Substitute the input types of all token card components from `Token`/`TokenUi` to `CardData`.
- Substitute the input name of all token card components from `token` to `data`.


peterpeterparker (approved):
LGTM, thx

PR #3083, opened 2024-10-24
build(frontend): use @dfinity/utils next

AntonioVentilii-DFINITY:
# Motivation

We want to use the the latest release (next) of @dfinity/utils because we want to use the recently-introduced class `AgentManager`.


AntonioVentilii-DFINITY:
Unresolved dependencies in ic-js, closing this

PR #3084, opened 2024-10-24
refactor(frontend): Rename subLogo to badge

AntonioVentilii-DFINITY:
# Motivation

As per @peterpeterparker 's suggestion, we rename the prop `subLogo` of `TokenLogo` component to `badge`, since now its scope is wider, not only restricted to show another logo.


peterpeterparker (approved):
LGTM, thx

PR #3085, opened 2024-10-24
fix(ci): adjusted checkout and pull strategy

inc-man:
# Motivation

The last run of the auto update failed because of no branch, this should be fixed.

# Changes

- adjusted checkout so we are always in the correct repo and correct branch from the pr
- adjusted pull strategy in add and commit action


.github/workflows/auto-update-i18n.yml
```diff
@@ -15,6 +15,8 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
```
AntonioVentilii-DFINITY, line 18: i don't think you need any input here, i don't see any in the other CI

.github/workflows/auto-update-i18n.yml
```diff
@@ -45,6 +47,7 @@ jobs:
         uses: EndBug/add-and-commit@v9
         with:
           add: '.'
+          default_author: github_actions
           message: '🤖 Updated i18n files'
           push: true
```
AntonioVentilii-DFINITY, line 52: and i think neither `push` nor `pull` are needed, i don't see in the other usages

.github/workflows/auto-update-i18n.yml
inc-man: it is needed for the add-and-comit action. They state that if working with PRs you need to add that. Thats why the last run I did failed. 

It is also in ic-js/.github/workflows/docs.yml

.github/workflows/auto-update-i18n.yml
inc-man: ic-js/.github/workflows/docs.yml uses a faulty "pull" because they don't want to pull. "Default" value is not the default so I think we can leave it like that. Push could be left out, but I feel like for someone who reads the code for a first time it makes sense to see what this action does. Idk no strong opinion on that one.

.github/workflows/auto-update-i18n.yml
AntonioVentilii-DFINITY: Ah understood! thanks for the explanation

.github/workflows/auto-update-i18n.yml
AntonioVentilii-DFINITY: ah understood! then, maybe, could you please adjust the snapshot workflow too in another PR?

inc-man: just add the push / pull to the snapshot workflow?

AntonioVentilii-DFINITY (approved):

PR #3086, opened 2024-10-24
feat(frontend): Util to map a token group to card data

AntonioVentilii-DFINITY:
# Motivation

The header card for a token group will be a summary of what the token group consists of. To do that we first need a mapping function that defines what are the "summary" data to be shown in the Token Card of the header.


peterpeterparker (approved):
Nice, thanks!

AntonioVentilii-DFINITY:
[OISY-69]

PR #3091, opened 2024-10-24
feat(frontend): Extract Setter store from Certified store

AntonioVentilii-DFINITY:
# Motivation

It is useful to have a setter store that is not strictly related to certified data (`certified` flag).


AntonioVentilii:
This PR is not necessary anymore

PR #3094, opened 2024-10-24
feat(frontend): create Storage Setter store

AntonioVentilii-DFINITY:
# Motivation

For the token group store, we need a mix between the `Storage` store and the map provided by the `Certified` store. Meaning that we want saved in local storage some thing that is

```typescript
{
  id1: data1,
  id2: data2,
  ...
}
```

In this PR we create a new type of `Storage` store that resembles the mapping provided by the `Certified` store.

### DIFF
--- a/src/frontend/src/lib/stores/storage-setter.store.ts
+++ b/src/frontend/src/lib/stores/storage-setter.store.ts
@@ -6,12 +6,13 @@ import { type Readable } from 'svelte/store';
 export type StorageSetterStoreData<T> = Option<Record<TokenId, T | null>>;
 
 export interface StorageSetterStore<T> extends Readable<StorageSetterStoreData<T>> {
-	reset: (tokenId: TokenId) => void;
 	set: (params: { tokenId: TokenId; data: T }) => void;
+	reset: (tokenId: TokenId) => void;
+	resetAll: (params: { key: string }) => void;
 }
 
 export const initStorageSetterStore = <T>({ key }: { key: string }): StorageSetterStore<T> => {
-	const { set, subscribe, update } = initStorageStore<StorageSetterStoreData<T>>({ key });
+	const { set, subscribe, update, reset } = initStorageStore<StorageSetterStoreData<T>>({ key });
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/stores/storage.store.ts
```diff
@@ -1,24 +1,27 @@
...
 
-	const { set, subscribe } = writable<StorageStoreData<T>>(initialValue);
+	const { set, subscribe, update } = writable<StorageStoreData<T>>(initialValue);
 
 	return {
 		set: ({ key, value }) => {
 			setStorage({ key, value });
 			set(value);
 		},
+		update,
```
peterpeterparker, line 24: I think we should not leak "update" because a consumer can use it and the value does not get saved in the local storage.

Is there another way?

src/frontend/src/lib/stores/storage-setter.store.ts
```diff
@@ -0,0 +1,43 @@
...
+				const newState = {
+					...state,
+					[tokenId]: null
+				};
+
+				set({ key, value: newState });
+
+				return newState;
+			}),
+		resetAll: reset,
```
peterpeterparker, line 40: is this correct?

peterpeterparker (commented):
Worth writing a small test too?

src/frontend/src/lib/stores/storage-setter.store.ts
AntonioVentilii-DFINITY: my idea is that `resetAll` would reset the entire storage

I can rename it `resetStorage`?

src/frontend/src/lib/stores/storage.store.ts
AntonioVentilii-DFINITY: i tried without it and it is really not what i would do... basically it uses `subscribe` and not in a functional way... let me think a bit and come back on this

AntonioVentilii:
This PR is not necessary anymore

PR #3098, opened 2024-10-24
chore(frontend): Remove unused prop of TokenUiGroup

AntonioVentilii-DFINITY:
# Motivation

We remove the props that are not needed from `TokenUiGroup` and we keep only the needed ones.

### DIFF
--- a/src/frontend/src/tests/lib/utils/token-card.utils.spec.ts
+++ b/src/frontend/src/tests/lib/utils/token-card.utils.spec.ts
@@ -1,5 +1,4 @@
-import { ETHEREUM_NETWORK, ICP_NETWORK } from '$env/networks.env';
-import { BTC_MAINNET_TOKEN } from '$env/tokens.btc.env';
+import { ICP_NETWORK } from '$env/networks.env';
 import { ETHEREUM_TOKEN, ICP_TOKEN } from '$env/tokens.env';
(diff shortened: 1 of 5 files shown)


AntonioVentilii-DFINITY:
[OISY-69]

src/frontend/src/tests/lib/utils/token-card.utils.spec.ts
```diff
@@ -30,7 +22,7 @@ describe('mapHeaderData', () => {
 			decimals: ICP_TOKEN.decimals,
 			icon: ICP_TOKEN.icon,
 			network: ICP_NETWORK,
-			oisyName: { oisyName: BTC_MAINNET_TOKEN.name },
+			oisyName: { oisyName: ICP_NETWORK.name },
```
peterpeterparker, line 25: It's fine but maybe for a safer test we should pick another token. The ICP network and token have the same name.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/tests/lib/utils/token-card.utils.spec.ts
AntonioVentilii-DFINITY: right! I will use SEPOLIA_ETH in another PR:

- PR #

PR #3100, opened 2024-10-24
fix(frontend): Show balance in Token Group Card

AntonioVentilii-DFINITY:
# Motivation

To show the balance (and USD balance) of the summary of the token group, we need to use `TokenCardContent` instead of `TokenCard`.


AntonioVentilii-DFINITY:
[OISY-69]

peterpeterparker (approved):
LGTM, thx

PR #3101, opened 2024-10-24
test(frontend): Use Sepolia token for card tests

AntonioVentilii-DFINITY:
# Motivation

Since ICP token name is the same as ICP network name, to be completely sure of the correct implementation of mapHeaderData, we use the Sepolia token, since the name of the token is slightly different than the name of the network.


peterpeterparker (approved):
Thx

PR #3107, opened 2024-10-24
feat(frontend): make DappModal accessible across all routes

DenysKarmazynDFINITY:
# Motivation

This PR is about moving `DappModal` from `DappsExplorerSignedIn` to global `Modals` so the modal can be opened from different places (e.g. list item on the explorer page, carousel slides on the token page). Also, to avoid having `selectedDapp` retrieval logic in the global component (`Modals`), I created `DappModalDetails` where all the checks will be placed.

### DIFF
--- a/src/frontend/src/lib/components/dapps/DappModalDetails.svelte
+++ b/src/frontend/src/lib/components/dapps/DappModalDetails.svelte
@@ -1,0 +1,16 @@
+<script lang="ts">
+	import { nonNullish } from '@dfinity/utils';
+	import DappModal from '$lib/components/dapps/DappModal.svelte';
+	import { modalDAppDetails } from '$lib/derived/modal.derived';
+	import { modalStore } from '$lib/stores/modal.store';
+	import type { OisyDappDescription } from '$lib/types/oisyDappDescription';
+
+	let selectedDapp: OisyDappDescription | undefined = undefined;
+	$: selectedDapp = $modalDAppDetails
+		? ($modalStore?.data as OisyDappDescription | undefined)
+		: undefined;
+</script>
+
+{#if nonNullish(selectedDapp)}
+	<DappModal dAppDescription={selectedDapp} />
+{/if}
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/dapps/DappModalDetails.svelte
```diff
@@ -0,0 +1,16 @@
...
+	import { modalStore } from '$lib/stores/modal.store';
+	import type { OisyDappDescription } from '$lib/types/oisyDappDescription';
+
+	let selectedDapp: OisyDappDescription | undefined = undefined;
+	$: selectedDapp = $modalDAppDetails
+		? ($modalStore?.data as OisyDappDescription | undefined)
+		: undefined;
+</script>
+
+{#if nonNullish(selectedDapp)}
```
peterpeterparker, line 14: You might want to also use a symbol here to identity this particular modal. Just in case it would be reused (some day).

DenysKarmazynDFINITY: There is already a condition for `modalDAppDetails` to be true a few lines above which under the hood checks `$modalStore?.type === 'dapp-details'`. Or did I get your comment wrong?

peterpeterparker: Indeed ;)

For example `SendButtonWithModal` - see the `const modalId = Symbol();`

```
<script lang="ts">
	import SendButton from '$lib/components/send/SendButton.svelte';

	export let open: (modalId: symbol) => void;
	export let isOpen: boolean;

	const modalId = Symbol();
</script>

<SendButton on:click={() => open(modalId)} />

{#if isOpen && $modalStore?.data === modalId}
	<slot name="modal" />
{/if}
```

I just want to avoid to ends with two modals in the DOM in case somethine goes wrong.  But let's do that in another PR if it makes sense to you too.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/components/dapps/DappModalDetails.svelte
DenysKarmazynDFINITY: Ah, got it. Yeah, let's do it in a new PR.

PR #3111, opened 2024-10-24
feat(frontend): adjust Carousel initialisation function

DenysKarmazynDFINITY:
# Motivation

While implementing Carousel usages (2 instances rendered on the token page to comply with the provided design, one for XL screens, another one for smaller ones), I noticed that the frame and timers are being initialised even if one of carousel is not actually visible (display: none). In this PR, I add a check to avoid that. The check works and stops all processes both on mount and on resize.

### DIFF
--- a/src/frontend/src/lib/components/carousel/Carousel.svelte
+++ b/src/frontend/src/lib/components/carousel/Carousel.svelte
@@ -45,11 +45,9 @@
 	onMount(() => {
 		initializeSlides();
 		initializeCarousel();
-		initialiseAutoplayTimer();
 
 		return () => {
-			clearAutoplayTimer();
-			clearSlideTransformTimer();
+			clearTimers();
 		};
 	});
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -75,6 +81,12 @@
 
 		containerWidth = container.offsetWidth;
 
+		// Clear timers and stop further initialisation in case container is rendered but not visible (e.g. display: none)
+		if (containerWidth === 0) {
```
peterpeterparker, line 85: Does that work in every browser and devices (including mobile phone)?

Should we plug quickly an intersection oberserve to be a on the safest side?

DenysKarmazynDFINITY: Safari, Firefox, Chrome - ✅ 

I think IntersectionObserver could lead to some unwanted clearing/initialising processes - if Carousel is rendered and does not have display:none but still out of the screen (e.g. long tokens list on tablet and the carousel stays on top), `isIntersecting` will be giving false/true on scroll, so the events, timers, and etc. will re-initialise which is not really necessary. WDYT?

peterpeterparker: Too some extention that can be considered as a feature. 😉
But I hear you.

peterpeterparker (approved):
LGTM, as discussed offline. Thanks.

PR #3112, opened 2024-10-24
feat(frontend): integrate group store

darioloetsc:
# Motivation

Integrate group store with token group card to enable storing the isExpanded state of the card.

# Changes

- Rename expanded to isExpanded for consistency
- use Symbol.for on group creation to prevent mismatch of logically same group
- Glue store and component

# Tests


AntonioVentilii-DFINITY (approved):
LGTM tks!

PR #3113, opened 2024-10-24
refactor(frontend): Extract sub-functions to sum balances

AntonioVentilii-DFINITY:
# Motivation

We need to re-use the sub-functions of `sumTokenBalances` and `sumTokenUsdBalances`, specifically to provide only the balances instead of the entire token object.


peterpeterparker (approved):
LGTM, thx

PR #3114, opened 2024-10-24
feat(frontend): Grouping util for token list

AntonioVentilii-DFINITY:
# Motivation

We want to change the way we group tokens, to have more than two elements in the group, and to try and avoid repetitions/glitches when the order of the token list is not the usual one.
Plus, the 

# Definitions

### Main Token

The "original" token of a certain group or of a certain other token; for example `BTC` is the "main token" of `ckBTC`, or `ETH` is the "main token" of `SepoliaETH`. Typically, a token and its "main token" should be 1:1.

### Secondary token

A token that has a "main token", even if not present in the list. It is just to indicate that it should always be grouped if the latter is present.

# Implementation

Following PR #3296 ,tThe function loops through the tokens list and groups them according to a prop key that links a token to its "main token". For example, it could be `twinToken` for ck-tokens. But the logic can be extended to other props, if necessary.

The tokens with no "main token" will still be included in a group, but it will be a single-element group, where the the token itself takes the place of the "main token". That, in general, makes sense for:
- tokens that are not "secondary tokens" of a "main token";
- tokens that have a "main token" but it is not present in the list.

The returned list respects the sorting order of the initial tokens list, meaning that the group is created at each position of the first encountered token of the group. So, independently of being a "main token" or a "secondary token", the group will replace the first token of the group in the list. That is useful if a "secondary token" is before the "main token" in the list; for example, if the list is sorted by balance.

# Tests

I created a few tests mocking some twin tokens and specific scenarios: different sorting, multiple "secondary tokens", missing "main token".


AntonioVentilii-DFINITY:
[OISY-164]

AntonioVentilii-DFINITY:
@peterpeterparker I think ready for another review

AntonioVentilii-DFINITY:

src/frontend/src/lib/utils/token-group.utils.ts
```diff
@@ -164,3 +165,52 @@ export const groupMainToken = ({ token, tokenGroup }: GroupTokenParams): TokenUi
...
+ */
+export const groupTokens = (tokens: TokenUi[]): TokenUiGroup[] => {
+	const tokenGroupsMap = tokens.reduce<{
+		[id: TokenId]: TokenUiGroup | undefined;
+	}>(
+		(acc, token) => ({
+			...acc,
+			...('twinToken' in token &&
+			isToken(token.twinToken) &&
+			token.decimals === token.twinToken.decimals
```
peterpeterparker, line 195: Not necesseraly for this PR but the comparison of decimals feels like something important so maybe worth a standalone function for it?

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -183,3 +185,6 @@ export const sumUsdBalances = ([usdBalance1, usdBalance2]: [
  */
 export const isRequiredTokenWithLinkedData = (token: Token): token is RequiredTokenWithLinkedData =>
 	'twinTokenSymbol' in token && typeof token.twinTokenSymbol === 'string';
+
+export const isToken = (token: Option<unknown>): token is Token =>
```
peterpeterparker, line 189: Should we add a test for this?

peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
Superseded by:

-

PR #3116, opened 2024-10-25
feat(frontend): Page for Terms and Conditions

AntonioVentilii-DFINITY:
# Motivation

We need a Terms and Condition page for OISY. The text was approved by Legal team. So, in this PR, we define the component to be used.

https://github.com/user-attachments/assets/f457db63-d6e8-48d8-9982-f93f71376baf


AntonioVentilii-DFINITY:
[OISY-122]

src/frontend/src/lib/components/terms-and-conditions/TermsAndConditions.svelte
```diff
@@ -0,0 +1,34 @@
...
+		$i18n.terms_and_conditions.text.indemnity,
+		$i18n.terms_and_conditions.text.governing_law,
+		$i18n.terms_and_conditions.text.entire_agreement,
+		$i18n.terms_and_conditions.text.assignment,
+		$i18n.terms_and_conditions.text.no_waiver,
+		$i18n.terms_and_conditions.text.english_version
+	];
+
+	let terms: string;
+	$: terms = replaceOisyPlaceholders(termsList.join('<br><br>'));
```
peterpeterparker, line 27: I would render them in paragraph, feel a bit more clean semantically speaking WDYT?

src/frontend/src/routes/(terms-and-conditions)/terms-and-conditions/+page.ts
```diff
@@ -0,0 +1,5 @@
+import { loadRouteParams, type RouteParams } from '$lib/utils/nav.utils';
```
peterpeterparker, line 1: I would use the same `page.ts` as for the signer route because those parameters are not used here

peterpeterparker (commented):
I suggest to rename the route group `(terms-and-conditions)` to `(public)`.

The `+layout` can maybe remain at the root of the group.

Ideally I would provide two PRs, one to initialize the public route, another for its content but fine if we keep it here.

src/frontend/src/routes/(terms-and-conditions)/terms-and-conditions/+page.ts
AntonioVentilii-DFINITY: yes, I thought about it, but I was thinking that maybe at some point there will be a link inside the logged in page. So, if the user go "back", it would need the parameters

But that we can see when we cross that bridge! Changed

(and opened a new PR for the route)

AntonioVentilii-DFINITY:
@peterpeterparker you are right, new PR for the route:

- PR #3120

src/frontend/src/lib/components/terms-and-conditions/TermsAndConditions.svelte
AntonioVentilii-DFINITY: Done! Changed, have a look please

src/frontend/src/lib/components/terms-and-conditions/TermsAndConditions.svelte
```diff
@@ -0,0 +1,34 @@
...
+		$i18n.terms_and_conditions.text.indemnity,
+		$i18n.terms_and_conditions.text.governing_law,
+		$i18n.terms_and_conditions.text.entire_agreement,
+		$i18n.terms_and_conditions.text.assignment,
+		$i18n.terms_and_conditions.text.no_waiver,
+		$i18n.terms_and_conditions.text.english_version
+	];
+
+	let terms: string;
+	$: terms = replaceOisyPlaceholders(termsList.map((term) => `<p>${term}</p>`).join(''));
```
peterpeterparker, line 27:
```
let termsList: string[];
$: termsList = [...].map(replaceOisyPlaceholders);

<section class="mt-12">
  {#each termsList as terms}
     <p><Html text={terms} /></p>
  {/each}
</seciton>
```

WDYT? A bit less performant agree.

AntonioVentilii-DFINITY: Ah! that was almost exactly one of my iterations, but I thought you didn't want too many components ahahahaahah

Done! Have a look

peterpeterparker (approved):
LGTm, thx

src/frontend/src/lib/components/terms-and-conditions/TermsAndConditions.svelte
peterpeterparker: Always Html > JS 😉

Thanks

PR #3120, opened 2024-10-25
feat(frontend): New public routes

AntonioVentilii-DFINITY:
# Motivation

We would like to have "public" routes in OISY, for example for Terms and Conditions. So, we set an independent page, agnostic to signed in or not.

Screenshot 2024-10-25 at 12 48 55


peterpeterparker (approved):
Looks good, already approving.

Just a question: it was requested to name the route `/terms-and-conditions`? not just `/terms`?

AntonioVentilii-DFINITY:
@peterpeterparker no compulsory specification actually, but this was like that in the requirements. I just asked the Design team + Legal and they say to go with `/terms-and-conditions`, even if they checked that some other app uses just `/terms`

AntonioVentilii-DFINITY:
[OISY-122]

peterpeterparker:
> even if they checked that some other app uses just 

Indeed `/terms` and `/privacy` are quite common I got the feeling. Thanks for the feedback.

PR #3121, opened 2024-10-25
feat(frontend): Map Tailwind colors as Design Team

AntonioVentilii-DFINITY:
# Motivation

- Base Colors: native Hex colors with assigned name to be easily manageable.
- Theme Colors: these will be colors fixed for each theme and they will have separate CSS variables; for now we only have `default` theme.
- Variable Mapping: these will be the map that will change according to the theme. That means that the definitions will not change, but its mapped color will.

# Changes

# Tests


src/frontend/src/lib/components/hero/HeroContent.svelte
```diff
@@ -33,7 +33,7 @@
 </script>
 
 <div
-	class="class:from-blue-ribbon-light flex h-full w-full flex-col content-center items-center justify-center rounded-[40px] bg-blue-ribbon bg-gradient-to-b from-blue-ribbon via-absolute-blue bg-size-200 bg-pos-0 p-6 text-center text-white transition-all duration-500 ease-in-out"
```
AntonioVentilii-DFINITY, line 36: the removed one is unrelated: it has no effect to the class and was there by mistake

AntonioVentilii-DFINITY:
@loki344 have a look please, this is a mix between your proposal (PR #2897) and what I had in mind. Do you think it could work? You proposed that having theme CSS variables is better to switch themes?

tailwind.config.ts
```diff
@@ -43,18 +44,18 @@ export default {
...
-			cobalt: '#004abe',
 			'resolution-blue': '#012f80',
 			zumthor: '#e8f1ff',
 			onahau: '#d1e3ff',
 			anakiwa: '#b0cdff',
 			beer: '#f7931a',
 			fulvous: '#de7900',
-			water: '#d1e3ff'
+			water: '#d1e3ff',
+			...colorsBase,
```
loki344, line 57: @AntonioVentilii-DFINITY  We should not add the colorsBase here: You don't want people to use colors directly, they should always be referenced via a theme variable (eg. foreground-brand-primary) because otherwise theming will be impossible. 
Example: If someone uses a certain blue directly, instead of assigning the theme-mapped-variable, this component then always will be blue, no matter which theme is applied.

src/frontend/src/lib/styles/tailwind/colors/theme-default.ts
```diff
@@ -0,0 +1,11 @@
+import { colorsBase } from './base';
+
+export const themeDefault = {
```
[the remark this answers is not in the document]
loki344: see 
image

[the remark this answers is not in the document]
loki344: note that this will not be the actual theme, this is only declaring the available primitives

src/frontend/src/lib/styles/tailwind/colors/base.ts
loki344, line 1: I still see no benefit of having to invent a new name for each color we're adding just to have a named hex here.. But I'll leave it

AntonioVentilii-DFINITY: My idea is to have an "easy" reference point (easier to talk about name than hexes eheh), but even because I am pretty sure we will re-use a few of them for other themes.

That means that we don't need to duplicate the HEX, but just its name. I don't know, it feels easier, at least for me eheh! But feel free to remove that one and just define HEX values directly

src/frontend/src/lib/styles/tailwind/colors/theme-default.ts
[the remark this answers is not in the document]
AntonioVentilii-DFINITY: so, this will be the theme default defined as Primitive, right? and a new theme will be for example `modernPrimitives` or `themePrimitivesModern`? or something similar?

tailwind.config.ts
loki344: Agree, done

AntonioVentilii-DFINITY:
superseded by :

- PR #3246

PR #3124, opened 2024-10-25
fix(ci): add release to pr check

inc-man:
# Motivation

We can adjust the naming of the release branch to as it was before. Like this we don't need to adjust the PR-check workflow.

# Changes

- remove release from PR-check
- change release to chore(release) in both PR branch and PR title


.github/workflows/bump-version.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       with:
         token: ${{ secrets.GIX_CREATE_PR_PAT }}
         branch: "release/v${{ steps.bump_version.outputs.new_version }}"
-        title: "Release: v${{ steps.bump_version.outputs.new_version }}"
+        title: "release: v${{ steps.bump_version.outputs.new_version }}"
```
AntonioVentilii-DFINITY, line 53: you don't need `release(something):` right?

AntonioVentilii-DFINITY (approved):
Already approving, check below

.github/workflows/bump-version.yml
inc-man: I'm pretty sure its not necessary. If I think about it I wouldn't really know what to put in the brackets.

release(oisy)? because its a release for everything right?

AntonioVentilii-DFINITY: i would say `release(ci)` or `ci(release)`

AntonioVentilii-DFINITY: correction: i saw that in the previous releases we would do `chore(release)` or `chore(devops)`

AntonioVentilii-DFINITY (approved):

PR #3125, opened 2024-10-25
fix(frontend): join oisy name on group from tokens

darioloetsc:
# Motivation

create oisy name for group by joining token symbols.

# Changes

replace name creation

# Tests


AntonioVentilii-DFINITY (approved):
LGTM! Tks

PR #3126, opened 2024-10-25
feat(frontend): include License Agreement link in landing page

AntonioVentilii-DFINITY:
# Motivation

We add a link to the License Agreement in the main page, when logged out, below the Authenticate button.

Screenshot 2024-10-25 at 14 34 50
Screenshot 2024-10-25 at 14 35 00
Screenshot 2024-10-25 at 14 35 11


loki344 (approved):
LGTM, thx

peterpeterparker (commented):
Are you planning to link the agreement everywhere the user can sign in—for example, in the signer standard flow—as well as in another PR, or is this not planned?

And if planned, would it be worthwhile to provide the agreement in a dedicated component, or are you planning to refactor it afterward?

peterpeterparker:
Also, unsolicited design feedback, should the font size be a bit smaller?

AntonioVentilii-DFINITY:
@peterpeterparker yes, you are correct: the plan is to show it everywhere (even after logged in actually); however, I didn't refactor it in a single component already because I am not sure it will be done in the same way in all instances.

I imagine that for all the uses of `ButtonAuthenticate` will be the same (so, yes, signer standard too), but I would prefer to await the Design team and refactor in another PR. WDYT? Do you prefer to hold this one and confirm with Design Team first?

peterpeterparker (approved):
LGTM, thx

Discussed add-ons provided in other PRs work for me.

PR #3128, opened 2024-10-25
style(frontend): only apply button animation related classes for loading

loki344:
# Motivation

Animation styles of the button were introduced for the loading state. Some of these classes are applied permanently to the button. This leads to unwanted side effects when changing the colorStyle of the button.

# Changes

Apply classes for loading animation only if loading.

# Tests

before: 

https://github.com/user-attachments/assets/db14dc8e-d1e9-4fb3-a7e0-50a3b6045964

after:

https://github.com/user-attachments/assets/3c807455-25be-4311-aa86-adef43eb1244

loading state verification:

https://github.com/user-attachments/assets/f6d3f8bf-4ce6-47b5-a1b1-63f1ece0cd8a


AntonioVentilii-DFINITY:
will the transition from `loading` to not `loading` be smooth in this way? may you please provide a small video test?

PR #3130, opened 2024-10-25
feat(frontend): no hamburger for why oisy on small screens

loki344:
# Motivation

On small screens, showing a hamburger icon with just one menu item is no good UX. Until we have a second item besides why oisy, we replace the menu with the actual item.

# Changes

- replace AboutWhatMenu with de facto only item aboutWhyOisy
- 
# Tests

before: 
image
after:
image

https://github.com/user-attachments/assets/521098ee-fe7d-4981-83eb-3db7574ebe7e


peterpeterparker (approved):
LGTM, thx

PR #3141, opened 2024-10-28
feat(frontend): extract common input text component

peterpeterparker:
# Motivation

I noticed that all input type `text` are similar. Given that I want to add a property to all of them, I thought that we can get started by extracting a common component.

# Changes

- create `InputText` for all text inputs and consume it


AntonioVentilii-DFINITY (approved):
LGTM! tks!

PR #3142, opened 2024-10-28
feat(frontend): wallet connect input placeholder i18n

peterpeterparker:
# Motivation

A noticed an i18n key was missing in wallet connect input.


AntonioVentilii-DFINITY (approved):
Tks

PR #3143, opened 2024-10-28
feat(frontend): input text autocomplete off

peterpeterparker:
# Motivation

Someone on Twitter shared a video in which they are using OISY: https://x.com/AaronBremser_/status/1849408317325545787

I noticed that the input fields were always suggesting previous values which I find annoying in terms of UX as it does not make much sense that OISY prompts for values like "1" or "2".

# Changes

- Disable autocomplete for all input text fields.


AntonioVentilii-DFINITY (approved):
Thank you

PR #3146, opened 2024-10-28
feat(frontend): e2e test ids

darioloetsc:
# Motivation

Current e2e tests are read only, this pr requests to add test-ids mostly on buttons and forms. Preparing for the integration of a e2e send test case.

# Changes

- Add multiple test ids


src/frontend/src/lib/components/receive/ReceiveActions.svelte
```diff
@@ -7,12 +7,13 @@
 	export let address: string;
 	export let qrCodeAriaLabel: string;
 	export let copyAriaLabel: string;
+	export let testCopyId: string | undefined = undefined;
```
DenysKarmazynDFINITY, line 10: I noticed that this var is called in some files `testCopyId` and in others - `testIdCopy`. First of all, let's align the naming. Second of all, how about `copyButtonTestId`? It sounds a bit more cleaner.

darioloetsc: Fine for me 09097d39

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3157, opened 2024-10-28
fix(ci): fix dependabot issue

inc-man:
# Motivation

The dependabot workflows fail because of insufficient access rights. So we want to exclude the dependabot workflows.

# Changes

- added condition to exclude dependabot workflows


AntonioVentilii-DFINITY (approved):
Tks for the quick fix

PR #3174, opened 2024-10-28
chore(frontend): remove inexistent class

AntonioVentilii-DFINITY:
# Motivation

Just noticed that there is an inexistent class, probably a residual of some copy/paste process. It has no effect so we can remove it safely.


peterpeterparker (approved):
LGTm, thx

PR #3177, opened 2024-10-28
style(frontend): Reduce padding of Hero

AntonioVentilii-DFINITY:
# Motivation

To make it less dispersive, we reduce the padding on top of Hero, for the different brakpoints.

https://github.com/user-attachments/assets/5da80529-c68c-49f5-9c57-c8b57b53cc1e


peterpeterparker (approved):
LGTm, thx

PR #3155, opened 2024-10-28
style(frontend): spacing on receive modal two addresses

peterpeterparker:
# Motivation

On the receive modal, when there are two addresses, there is a lack of space around the `<hr />`.

# Screenshots

Before:

Capture d’écran 2024-10-28 à 09 26 40

After:

Capture d’écran 2024-10-28 à 09 26 07

PR #3181, opened 2024-10-28
feat(frontend): hr spacing

peterpeterparker:
# Motivation

Follow-up of https://github.com/dfinity/oisy-wallet/pull/3155#pullrequestreview-2398266006. We want to have the spacing for HR embeded within the component.

# Changes

- Add property and style for spacing around `hr`

### DIFF
--- a/src/frontend/src/lib/components/ui/Hr.svelte
+++ b/src/frontend/src/lib/components/ui/Hr.svelte
@@ -1,1 +1,9 @@
-<hr class="bg-dark-blue opacity-10" style="width: 100%; border: 0.05rem solid" />
+<script lang="ts">
+	export let spacing: 'none' | 'lg' = 'none';
+</script>
+
+<hr
+	class="bg-dark-blue opacity-10"
+	class:my-6={spacing === 'lg'}
+	style="width: 100%; border: 0.05rem solid"
+/>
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/ui/Hr.svelte
```diff
@@ -1 +1,9 @@
-<hr class="bg-dark-blue opacity-10" style="width: 100%; border: 0.05rem solid" />
+<script lang="ts">
+	export let spacing: 'none' | 'lg' = 'none';
+</script>
+
+<hr
+	class="bg-dark-blue opacity-10"
+	class:my-6={spacing === 'lg'}
```
AntonioVentilii-DFINITY, line 7: I dont't know if it makes sense to you, but we have `Size` type now

`type Size = 'xs' | 'sm' | 'md' | 'lg' | 'xl';`

so you could use `Pick<Size, 'lg'>` if you want

(though it is not globally exported yet...)

peterpeterparker: `Pick<Size, 'lg'>` is not possible though because `Size` is a `type` or I am missing something?

AntonioVentilii-DFINITY: right! sorry! maybe `Extract<Size, 'lg'>`? But for just one, maybe not worth it? boh

peterpeterparker: Nice, I did not knew about `Extract`! Done. https://github.com/dfinity/oisy-wallet/pull/3181/commits/88382d8c51d25b0d6ceb8e32b40f3f3b6aea66e5

PR #3183, opened 2024-10-28
chore(frontend): update Twitter URL

AntonioVentilii-DFINITY:
# Motivation

We update the OISY twitter URL and make it part of the metadata file.

### DIFF
--- a/src/frontend/src/env/types/env-oisy-metadata.ts
+++ b/src/frontend/src/env/types/env-oisy-metadata.ts
@@ -6,5 +6,6 @@ export const oisyMetadata = z.object({
 	OISY_ONELINER: z.string(),
 	OISY_DESCRIPTION: z.string(),
 	OISY_REPO_URL: z.string().url(),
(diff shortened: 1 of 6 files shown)


DenysKarmazynDFINITY (approved):
Nice, thanks!

src/frontend/src/env/types/env-oisy-metadata.ts
```diff
@@ -6,5 +6,6 @@ export const oisyMetadata = z.object({
 	OISY_ONELINER: z.string(),
 	OISY_DESCRIPTION: z.string(),
 	OISY_REPO_URL: z.string().url(),
-	OISY_ALPHA_WARNING_URL: z.string().url()
+	OISY_ALPHA_WARNING_URL: z.string().url(),
+	OISY_TWITTER_URL: z.string().url()
```
peterpeterparker, line 10: If I would be picky I would say that the zod parser should validate the URL is indeed a Twitter URL 😉

PR #3171, opened 2024-10-28
feat(frontend): new modal design

peterpeterparker:
# Motivation

New design for the modal as in Figma.

# TODO

We need E2E tests that take screenshots of the modals. It was incredibly painful to test each modal repeatedly. I ended up taking screenshots for this PR about 20 times.

# Leftovers

It was already too much work. I'll have a look to the WalletConnect modal afterwards.

# Changes

- Adapt style for the gix-cmp modals
- Create a mixin for the card-content of the modal because we need a special tricks on the "Manage tokens" modal
- Adapt various paddings
- Modify "Manage Tokens" to fit this new design
- Remove custom style in "Dapp modal"

# Screenshots

Capture d’écran 2024-10-29 à 09 42 13
Capture d’écran 2024-10-29 à 09 42 23
Capture d’écran 2024-10-29 à 09 42 25
Capture d’écran 2024-10-29 à 09 42 32
Capture d’écran 2024-10-29 à 09 43 03
Capture d’écran 2024-10-29 à 09 43 25
Capture d’écran 2024-10-29 à 09 43 27
Capture d’écran 2024-10-29 à 09 43 33
Capture d’écran 2024-10-29 à 09 43 42
Capture d’écran 2024-10-29 à 09 43 46
Capture d’écran 2024-10-29 à 09 43 49
Capture d’écran 2024-10-29 à 09 43 55
Capture d’écran 2024-10-29 à 09 43 57
Capture d’écran 2024-10-29 à 09 44 03
Capture d’écran 2024-10-29 à 09 44 07
Capture d’écran 2024-10-29 à 09 44 16
Capture d’écran 2024-10-29 à 09 44 23
Capture d’écran 2024-10-29 à 09 44 33
Capture d’écran 2024-10-29 à 09 44 45
Capture d’écran 2024-10-29 à 09 44 48
Capture d’écran 2024-10-29 à 09 44 51
Capture d’écran 2024-10-29 à 09 44 54
Capture d’écran 2024-10-29 à 09 44 55
Capture d’écran 2024-10-29 à 09 44 59
Capture d’écran 2024-10-29 à 09 45 01
Capture d’écran 2024-10-29 à 09 45 03
Capture d’écran 2024-10-29 à 09 45 10
Capture d’écran 2024-10-29 à 09 45 12
Capture d’écran 2024-10-29 à 09 45 17
Capture d’écran 2024-10-29 à 09 45 19
Capture d’écran 2024-10-29 à 09 45 23
[...]

PR #3186, opened 2024-10-29
style(frontend): add color ghost-white

peterpeterparker:
# Motivation

We need a new background colors for the modals  #3171.

# Changes

- Init color #f8f9fa as discussed with Artem


AntonioVentilii-DFINITY (approved):
I imagine you are going to use it in another PR, so approved

PR #3188, opened 2024-10-29
chore(frontend): Set missing Typescript in components

AntonioVentilii-DFINITY:
# Motivation

Just a small re-ordering: two components were missing `lang="ts"`. Not that it was necessary, but just for consistency.


peterpeterparker (approved):
Thx

PR #3192, opened 2024-10-29
fix(ci): adjust run-name in auto deploy

inc-man:
# Motivation

Currently the run-name only shows on workflow_dispatch but not on auto deploy.

# Changes

- adjusted run-name to work in every possible case
- removed obsolete entries


.github/workflows/deploy-to-environment.yml
```diff
@@ -13,22 +13,23 @@ on:
         options:
           - staging
           - beta
-        default: staging
```
AntonioVentilii-DFINITY, line 16: why removing the defaults?

.github/workflows/deploy-to-environment.yml
```diff
@@ -13,22 +13,23 @@ on:
...
       canister:
         required: true
         type: choice
         description: 'Select the canister to deploy.'
         options:
           - frontend
           - backend
-        default: frontend
```
AntonioVentilii-DFINITY, line 24: ditto

.github/workflows/deploy-to-environment.yml
```diff
@@ -13,22 +13,23 @@ on:
...
         required: false
         type: boolean
         description: 'Force backend deployment even if no changes are detected.'
         default: false
 
-run-name: Deploying ${{ inputs.canister }} to ${{ inputs.network }}
+run-name: >-
+  ${{ github.event_name == 'push'
+    && 'Deploying Frontend / Backend to Staging'
+    || format('Deploying {0} to {1} and force backend = {2}', inputs.canister, inputs.network, inputs.force-backend) }}
```
AntonioVentilii-DFINITY, line 32: can you make it so the "force backend" part appears only when the deploy is to Backend AND the force is true? is it possible?

.github/workflows/deploy-to-environment.yml
inc-man: I added the defaults in an earlier attempt to fix to run-name. Thought that by adding them, the run name would get those defaults. Didn't work so no real need.

.github/workflows/deploy-to-environment.yml
inc-man: see above

.github/workflows/deploy-to-environment.yml
inc-man: what if we add a 3rd option to the canister choice so we have:
- frontend
- backend
- both

with that the force-backend becomes obsolete. Or am I missing a case? Since chosing backend and force-backend is the exact same thing.

.github/workflows/deploy-to-environment.yml
AntonioVentilii-DFINITY: thank you for the explanation

.github/workflows/deploy-to-environment.yml
inc-man: yeah you are right. I think its possible but it might just result in an ugly string. Let me see what I can do

AntonioVentilii-DFINITY (approved):
LGTM! tks

PR #3193, opened 2024-10-29
feat(frontend): handle BTC in the Import modal

DenysKarmazynDFINITY:
# Motivation

In this PR, BTC networks are filtered out from the selector in the Import modal. Also, an additional message is added in case user opens the modal while having an active BTC network.

Screenshot 2024-10-29 at 10 30 39

Screenshot 2024-10-29 at 10 28 46


AntonioVentilii-DFINITY (approved):
LGTM! Tks

PR #3202, opened 2024-10-29
chore(frontend): cleanup dapp description types

peterpeterparker:
# Motivation

Consistency.

# Changes

- Rename module
- Upper case for schema
- Add toto to move constants which should not find place in types


AntonioVentilii-DFINITY (approved):
Good one! THank you

PR #3203, opened 2024-10-29
feat(frontend): input text with action

peterpeterparker:
# Motivation

I've been bothered for quite some time by input fields that lack a call to action yet don't use the full width for the input. The root cause of the issue is that the slot must be declared at the top, even if it's not used, which causes gix-cmp to assume that a slot is always in use.

To resolve this, I created a new component, `InputTextWithAction`. This allows us to choose either a simple input or an input with an action, as needed.

# Screenshots

Before:

Capture d’écran 2024-10-29 à 13 37 29
Capture d’écran 2024-10-29 à 13 37 35
Capture d’écran 2024-10-29 à 13 37 38

After:

Capture d’écran 2024-10-29 à 13 36 58
Capture d’écran 2024-10-29 à 13 36 53
Capture d’écran 2024-10-29 à 13 36 49

### DIFF
--- a/src/frontend/src/lib/components/manage/ManageTokens.svelte
+++ b/src/frontend/src/lib/components/manage/ManageTokens.svelte
@@ -26,7 +26,7 @@
 	import ButtonCancel from '$lib/components/ui/ButtonCancel.svelte';
 	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
 	import Card from '$lib/components/ui/Card.svelte';
-	import InputText from '$lib/components/ui/InputText.svelte';
+	import InputTextWithAction from '$lib/components/ui/InputTextWithAction.svelte';
 	import { exchanges } from '$lib/derived/exchange.derived';
 	import {
 		networkEthereum,
@@ -195,7 +195,7 @@
 </script>
 
 <div class="mb-4">
-	<InputText
+	<InputTextWithAction
 		name="filter"
 		required={false}
 		bind:value={filter}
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/ui/InputText.svelte
```diff
@@ -15,7 +15,4 @@
 	{placeholder}
 	spellcheck={false}
 	autocomplete="off"
-	on:nnsInput
->
-	<slot name="inner-end" slot="inner-end" />
-</Input>
+/>
```
peterpeterparker, line 18: In case of `InputText` the `on:nnsInput` is currently unused therefore I removed it.

src/frontend/src/lib/components/ui/InputText.svelte
AntonioVentilii-DFINITY, line 1: Can `InputText` be just at this point, for example just 
`<InputTextWithAction .... />` ? Or is that the issue you mentioned with gix-comp that do not recognize absent slots? You already checked `if $$slots...` solves something, right?

peterpeterparker: Indeed that's the issue. There isn't really a solution, maybe with Svelte v5.

AntonioVentilii-DFINITY: cool, tks!

AntonioVentilii-DFINITY (approved):
LGTM! Tks

PR #3207, opened 2024-10-29
feat(frontend): chain-fusion receive with copy new design block

peterpeterparker:
# Motivation

In the recent / new design the new "copy qrcode block" is used on the main "Receive" modal.

# Notes

Reviewed by Artem. We know that it could look better on mobile.

# Changes

- Align `ReceiveAddresses` to new design
- Remove unused component given that the addresses are now displayed as any other addresses
- Make `address` optional in `ReceiveAddress` and replicate skeleton logic that existed in the removed component `ReceiveAddressWithLogo`
- Subtext removed under "ICP for exchanges" as requested by design

# Screenshots

Before:

Capture d’écran 2024-10-29 à 15 03 14
Capture d’écran 2024-10-29 à 15 03 18

After:

Capture d’écran 2024-10-29 à 14 59 41
Capture d’écran 2024-10-29 à 14 59 39
Capture d’écran 2024-10-29 à 14 59 30
Capture d’écran 2024-10-29 à 14 59 28
Capture d’écran 2024-10-29 à 14 59 21
Capture d’écran 2024-10-29 à 14 59 19


src/frontend/src/lib/components/receive/ReceiveAddress.svelte
```diff
@@ -44,9 +47,21 @@
 				/>
 			</div>
 
-			<output id="ic-wallet-address" class="break-all text-sm" data-tid={testId}>{address}</output>
+			{#if isNullish(address)}
```
AntonioVentilii-DFINITY, line 50: maybe invert the condition to be the same as the condition below?

AntonioVentilii-DFINITY (approved):
One observation below but not blocker, so approved already, your call!

Unsolicited additional feedback: what if the addresses are ellipsed on mobile only?

peterpeterparker:
> what if the addresses are ellipsed on mobile only?

I thought about it too! Maybe something we can come with agree.

src/frontend/src/lib/components/receive/ReceiveAddress.svelte
peterpeterparker: Good idea! Done. Thx

PR #3215, opened 2024-10-29
feat(frontend): convert to ckBTC button for BTC mainnet

DenysKarmazynDFINITY:
# Motivation

Add the convert-to-ckBTC button for BTC mainnet transactions page. 

Screenshot 2024-10-29 at 15 40 31

### DIFF
--- a/src/frontend/src/btc/convert/ConvertToCkBTC.svelte
+++ b/src/frontend/src/btc/convert/ConvertToCkBTC.svelte
@@ -1,0 +1,12 @@
+<script lang="ts">
+	import { BTC_MAINNET_TOKEN } from '$env/tokens.btc.env';
+	import IconCkConvert from '$lib/components/icons/IconCkConvert.svelte';
+	import ButtonHero from '$lib/components/ui/ButtonHero.svelte';
+	import { isBusy } from '$lib/derived/busy.derived';
+	import { i18n } from '$lib/stores/i18n.store';
+</script>
+
+<ButtonHero disabled={$isBusy} ariaLabel={$i18n.convert.text.convert_to_ckbtc}>
+	<IconCkConvert size="28" slot="icon" />
+	<span>{BTC_MAINNET_TOKEN.twinTokenSymbol}</span>
+</ButtonHero>
--- a/src/frontend/src/lib/components/hero/Actions.svelte
+++ b/src/frontend/src/lib/components/hero/Actions.svelte
@@ -1,6 +1,8 @@
 <script lang="ts">
 	import { page } from '$app/stores';
 	import BtcReceive from '$btc/components/receive/BtcReceive.svelte';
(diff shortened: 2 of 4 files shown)


src/frontend/src/btc/convert/ConvertToCkBTC.svelte
```diff
@@ -0,0 +1,12 @@
...
+	import { BTC_MAINNET_TOKEN } from '$env/tokens.btc.env';
+	import IconCkConvert from '$lib/components/icons/IconCkConvert.svelte';
+	import ButtonHero from '$lib/components/ui/ButtonHero.svelte';
+	import { isBusy } from '$lib/derived/busy.derived';
+	import { i18n } from '$lib/stores/i18n.store';
+</script>
+
+<ButtonHero disabled={$isBusy} ariaLabel={$i18n.convert.text.convert_to_ckbtc}>
+	<IconCkConvert size="28" slot="icon" />
+	<span>{BTC_MAINNET_TOKEN.twinTokenSymbol}</span>
```
AntonioVentilii-DFINITY, line 11: we have `ckTestBTC` as token, no? wouldn't it better to use the `$pageToken` store in this case? (not inside the modal I mean, just for the button)

or just for mainnet indeed?

src/frontend/src/lib/components/hero/Actions.svelte
```diff
@@ -18,19 +20,24 @@
...
 
 	let convertErc20 = false;
 	$: convertErc20 = $erc20ToCkErc20Enabled && $erc20UserTokensInitialized;
 
+	let convertCkBtc = false;
+	$: convertCkBtc = $tokenCkBtcLedger && $erc20UserTokensInitialized;
+
 	let convertBtc = false;
-	$: convertBtc = $tokenCkBtcLedger && $erc20UserTokensInitialized;
+	$: convertBtc = BTC_TO_CKBTC_EXCHANGE_ENABLED && isNetworkIdBTCMainnet($networkId);
```
AntonioVentilii-DFINITY, line 40: same question

src/frontend/src/btc/convert/ConvertToCkBTC.svelte
DenysKarmazynDFINITY: For the time being, my assumption is that I implement it only for mainnet. After all the UI and logic for mainnet is finished, this will be an easy thing to make it available for testnet too. WDYT?

AntonioVentilii-DFINITY: understood, thank you for the explanation!

src/frontend/src/lib/components/hero/Actions.svelte
AntonioVentilii-DFINITY: same answer as above , I guess

AntonioVentilii-DFINITY (approved):
LGTM! tks

PR #3221, opened 2024-10-29
Fix/glitchy navigation

peterpeterparker:
# Motivation

The navigation has become noticeably glitchy. The root causes are:

- The slide effect in the hero section has slowed down due to the increased data and components.
- The addition of the dApp explorer carousel, which uses transitions on certain pages, causes content to linger too long. This results in double content being displayed briefly, leading to layout shifts.

# Solution

- Remove the slide effect in the hero section.
- Implement the web API `startViewTransition` with its default values to smooth the transition between pages.

# Changes

- Remove slide effects in hero components.
- Add a `startViewTransition` snippet (copied from its usage on https://proposals.network).
- Add type support with `@types/dom-view-transitions`.
- Configure `tsconfig` to include these types.

# Screenshots

Before:

issue

After:

fix

### DIFF
--- a/src/frontend/src/lib/components/hero/HeroContent.svelte
+++ b/src/frontend/src/lib/components/hero/HeroContent.svelte
@@ -43,7 +43,7 @@
 	class:to-bright-lilac={$networkEthereum}
 >
 	{#if summary}
-		<div transition:slide={SLIDE_PARAMS} class="flex w-full flex-col gap-6">
+		<div in:slide={SLIDE_PARAMS} class="flex w-full flex-col gap-6">
 			<div class="grid w-full grid-cols-[1fr_auto_1fr] flex-row items-center justify-between">
 				<Back color="current" onlyArrow />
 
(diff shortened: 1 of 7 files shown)


src/frontend/src/lib/components/hero/HeroContent.svelte
```diff
@@ -71,12 +71,12 @@
...
 
 	{#if usdTotal}
-		<div transition:slide={SLIDE_PARAMS}>
+		<div in:slide={SLIDE_PARAMS}>
 			<ExchangeBalance />
 		</div>
 	{/if}
 
-	<div transition:slide|local={SLIDE_PARAMS} class="flex w-full justify-center text-left">
+	<div in:slide|local={SLIDE_PARAMS} class="flex w-full justify-center text-left">
```
AntonioVentilii-DFINITY, line 79: not for this PR, but I introduced the `|local` when I introduced the transition of the parent `Hero` too. Not really sure, but maybe it can be removed

AntonioVentilii-DFINITY (approved):

src/frontend/src/lib/components/hero/HeroContent.svelte
peterpeterparker: Good point! We can have a look afterwards indeed

PR #3226, opened 2024-10-29
refactor(frontend): Include Zod in Token interfaces

AntonioVentilii-DFINITY:
# Motivation

We create Zod schemas for a few interfaces of `Token` type.

Note: the types `TokenOisySymbol` and `TokenOisyName` were not used anymore, not even globally. So, we removed them.


peterpeterparker (approved):
Thanks!

PR #3227, opened 2024-10-29
refactor(frontend): Use Zod for NetworkId

AntonioVentilii-DFINITY:
# Motivation

We want to use Zod to safe parse the types of `Token`. Its interface uses the type `Network` too. So, we change type `NetworkId`, branding it specifically for NetworkId. This is similar to PR #3189 

# Changes

- Create schema for `NetworkId`.
- Use it to define `NetworkId` type.
- Create util to parse `NetworkId` type.
- Adapt the code in each usage of `NetworkId`.

# Tests

I didn't create a test for the util because I am not sure how to make it throw errors, but open to suggestions.


peterpeterparker (approved):
LGTM, thx

PR #3229, opened 2024-10-29
feat(e2e): add clickByTestId function

AntonioVentilii-DFINITY:
# Motivation

We add function `clickByTestId` to class `Homepage`, because it can come in handy when we need to click in extension of the class, but we don't want to "publish" the prop `#page`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3230, opened 2024-10-29
test(e2e): Add tests for transactions pages

AntonioVentilii-DFINITY:
# Motivation

We want to create snapshots test for each native token transaction page: BTC, ETH, and ICP.

# Changes

- Add additional customizable function to extend `HomepageLoggedIn.waitForReady`
- Create class `TransactionPage` extending `HomepageLoggedIn`.
- Create tests for BTC, ETH and ICP.

# Tests

In this PR we already updated the new screenshots. Please check committed files.


e2e/utils/pages/homepage.page.ts
```diff
@@ -22,7 +22,7 @@ interface HomepageParams {
 	viewportSize?: ViewportSize;
 }
 
-type HomepageLoggedInParams = {
+export type HomepageLoggedInParams = {
```
DenysKarmazynDFINITY, line 25: Obviously not in this PR, but I guess it's time to extract shared between all pages things from `Homepage` into `Page` class and use it as a parent for all further pages.

DenysKarmazynDFINITY (approved):
The code LGTM, thanks! I assume not much we can do with the txs call on the ETH page?

AntonioVentilii:
>  I assume not much we can do with the txs call on the ETH page?

 
 I tried to check, but it is really a question of API keys, and for now no, unfort...

e2e/utils/pages/homepage.page.ts
AntonioVentilii: yes, makes sense! Included in my backlog

PR #3232, opened 2024-10-30
style(frontend): sets Header z-index to 20 to ensure that it is displayed over overlays

BonomoAlessandro:
sets root z-index also to 20 to ensure that modals are still displayed over the menu

# Motivation

The wallet connect modal and some other parts of the header do have the wrong z-index configuration, which leads to be displayed behind some background elements.

# Changes

- z-index

# Tests

before:
image
to fix this, the z-index of the Header is set to 20 but this leads to following problem:
image
thats why the root z-index is also changed from 10 to 20

after:
image

image


AntonioVentilii-DFINITY (commented):
I feel a bit paranoid about the z-index: may you please provide a quick video with a few more cases, like the modals for receive/send/convert, the loader modal on login, the info modal, etc

BonomoAlessandro:
Here is a short video that shows some modals with different viewports

https://github.com/user-attachments/assets/98c7c084-b3aa-4b4c-8472-a9b253180258

AntonioVentilii-DFINITY (commented):
Sorry, I clicked the wrong button, it wasn't meant to be approved already: why not changing the z-index of the component that is causing issues?

BonomoAlessandro:
You are right, this is an easy solution.
now i set the z-index of the header to 12, which means it is over the overlay but behind the modals, which is exactly what we want.

AntonioVentilii-DFINITY:
I think that removing the `z-10` from `<DappPromoBanner>` would work, may you please check?

AntonioVentilii-DFINITY (approved):
LGTM tks

PR #3233, opened 2024-10-30
ci(test): Merge test and checks

AntonioVentilii-DFINITY:
# Motivation

There is some concurrency error with the status check `may-merge` of our CIs. Apparently, if one of the status check is positive, and the rest is not, the PR is cleared to merge.

We had a few instances of it in the past, but here is a practical example with PR #3207 : in the last commit before merge, [the workflow for the checks failed](https://github.com/dfinity/oisy-wallet/actions/runs/11576857312), but the one for the tests passed. So the latest may-merge superseded the first one, and the PR was allowed to merge.

Screenshot 2024-10-30 at 09 28 58

Screenshot 2024-10-30 at 09 29 27

To try and solve it, we first  merge the two workflows, so that they share the same status check `may-merge`


.github/workflows/frontend-checks.yml
```diff
@@ -28,8 +28,19 @@ jobs:
       - name: Lint
         run: npm run check
 
+  test:
```
peterpeterparker, line 31: moving the test here is related to the issue?

peterpeterparker (approved):
LGTM, thx

.github/workflows/frontend-checks.yml
AntonioVentilii-DFINITY: correct: this should avoid that the `may-merge` coming from tests workflow conflicts with the `may-merge `coming from checks workflow, since now they are the same workflow

PR #3234, opened 2024-10-30
style(frontend): fixes display of token balances

BonomoAlessandro:
# Motivation

There should be a consistent way to display the token balance and the usd balance.

Token card:
- If token balance and usd balance are defined, both should be displayed.
- If token balance is undefined and usd balance is defined, `n/a` should be displayed for the balance and `-. for the usd balance`.
- If token balance is defined and usd balance is undefined, the usd balance should be displayed as `n/a`.
- If token balance and usd balance are undefined, the balance should be displayed as `n/a` and the usd balance as `-`.

Token view:
When the usd balance is not defined, the text `$ value is not available` should be displayed.

# Changes

- displayed values

# Tests

### **before:**

**token card**

balance and usdBalance defined:
image

balance null and usdBalance defined:
image

balance defined and usdBalance undefined:
image

balance and usdBalance undefined:
image

**token view**

usd balance defined:
image

usd balance undefined:
image

### **after:**

**token card**

balance and usdBalance defined:
image

balance null and usdBalance defined:
image

balance defined and usdBalance undefined:
image

balance and usdBalance undefined:
image

**token view**

usd balance defined:
image

usd balance undefined:
image


AntonioVentilii-DFINITY:
So, this PR I think is not correct: it was decided that when the USD balance is undefined (not loaded yet) it should be shown as missing, since it indicates that we don't have a price feed. In fact, if you try to do that with a non-null token amount, you will see that the token amount is for example 4 and the USD amount is 0, and that is not correct.

BonomoAlessandro:
The token card looks now like this:

balance and usdBalance defined:
image

balance null and usdBalance defined:
image

balance defined and usdBalance undefined:
image

balance null and usdBalance undefined:
image

BonomoAlessandro:
@AntonioVentilii-DFINITY pls review

src/frontend/src/lib/components/hero/Balance.svelte
```diff
@@ -19,6 +19,10 @@
 	</output>
 
 	<span class="text-xl font-bold opacity-50">
-		<TokenExchangeBalance balance={token?.balance} usdBalance={token?.usdBalance} />
+		<TokenExchangeBalance
+			balance={token?.balance}
+			usdBalance={token?.usdBalance}
+			nullishBalanceMessage="$ value is not available"
```
AntonioVentilii, line 25: `i18n` ?

src/frontend/src/lib/components/tokens/TokenBalance.svelte
```diff
@@ -17,10 +16,7 @@
 				unitName: data.decimals
 			})}
 		{:else}
-			{formatToken({
-				value: ZERO,
-				unitName: data.decimals
-			}).replace('0', '-')}
+			<span>n/a</span>
```
AntonioVentilii, line 19: ditto

src/frontend/src/lib/components/tokens/TokenExchangeBalance.svelte
```diff
@@ -5,17 +5,15 @@
 
 	export let balance: TokenFinancialData['balance'];
 	export let usdBalance: TokenFinancialData['usdBalance'];
+	export let nullishBalanceMessage = '-';
```
AntonioVentilii, line 8: why a prop? I think you can use directly inside the component since that is its usage

BonomoAlessandro: This needs to be a prop because by default, this property will display `-` when the value is null or undefined. However, for the `Hero` element, it should display `'$ value is not available'` instead.

BonomoAlessandro:
@AntonioVentilii-DFINITY pls review

AntonioVentilii:
@BonomoAlessandro may you please update the description and the screenshots?

src/frontend/src/lib/components/tokens/TokenExchangeBalance.svelte
```diff
@@ -1,21 +1,21 @@
...
 </script>
 
 <output class="break-all">
 	{#if nonNullish(balance) && nonNullish(usdBalance)}
 		{formatUSD({ value: usdBalance })}
+	{:else if isNullish(balance)}
+		<span class:animate-pulse={isNullish(balance)}>{nullishBalanceMessage}</span>
```
AntonioVentilii, line 16: you don't need the condition here anymore, right? it will always be true

src/frontend/src/lib/components/tokens/TokenExchangeBalance.svelte
AntonioVentilii: I am still not convinced about this, i would prefer to default is as `undefined` and the use coalescing to overwrite it with `'-'`

src/frontend/src/lib/components/tokens/TokenExchangeBalance.svelte
```diff
@@ -1,21 +1,21 @@
...
 	{:else}
-		<span class:animate-pulse={isNullish(balance)}
-			>{formatUSD({
-				value: 0,
-				options: { minFraction: 0, maxFraction: 0 }
-			}).replace('0', '-')}
-		</span>
+		<span class:animate-pulse={isNullish(balance)}>{$i18n.tokens.balance.error.not_applicable}</span
```
AntonioVentilii, line 18: personally, I would not like to see a bunch of `'n/a'` on the screen... especially in the ICP network, I don't know if it gives a good impression to the users, @StefanBerger-DFINITY

src/frontend/src/lib/components/tokens/TokenBalance.svelte
```diff
@@ -17,10 +17,7 @@
 				unitName: data.decimals
 			})}
 		{:else}
-			{formatToken({
-				value: ZERO,
-				unitName: data.decimals
-			}).replace('0', '-')}
+			<span>{$i18n.tokens.balance.error.not_applicable}</span>
```
AntonioVentilii, line 20: here will be the same, but there are less tokens where this is true; maybe here just greyed out?

src/frontend/src/lib/components/tokens/TokenExchangeBalance.svelte
BonomoAlessandro: No this is still needed. The `if` above checks if the `balance` AND `usdBalance` is non nullish. In case that the `balance` is defined but the `usdbalance` is nullish this check also failes and the third option will take place.

src/frontend/src/lib/components/tokens/TokenExchangeBalance.svelte
BonomoAlessandro: @AntonioVentilii at the moment it is defined like this. I guess in case of a change it would be solved with another bug ticket.

src/frontend/src/lib/components/tokens/TokenBalance.svelte
BonomoAlessandro: ditto

src/frontend/src/lib/components/tokens/TokenExchangeBalance.svelte
AntonioVentilii: but `animate-pulse={isNullish(balance)}` will always be true, since it's inside `if isNullish(balance)}`

BonomoAlessandro:
@AntonioVentilii Description is updated and the not needed `isNullish` check is removed.

AntonioVentilii (approved):
LGTM, tks

PR #3237, opened 2024-10-30
feat(frontend): Show loading skeleton for group if one token is loading

AntonioVentilii-DFINITY:
# Motivation

The request was to show a skeleton for the `balance` (`balance = undefined`) if at least one of the tokens in the group has balance `undefined` (meaning that it is still loading). To do that we add a case in the `sumBalance` function for `undefined` case.


peterpeterparker (approved):
LGTM, thx

PR #3239, opened 2024-10-30
refactor(frontend): move SendTokenContext to lib

DenysKarmazynDFINITY:
# Motivation

SendTokenContext is not only used by `eth`, so it makes sense to have it in `lib`.


AntonioVentilii-DFINITY (approved):
LGTM tks!

PR #3240, opened 2024-10-30
style(frontend): fixes distortion images in ImgBanner

BonomoAlessandro:
# Motivation

Images in AboutWhyOisyModal, Loader, DappPromoBanner and DappModal are distorted and not behaving responsively.

# Changes

- simplify ImgBanner => Achieve space reserving via tailwind ratio, responsiveness via h-auto w-full (https://www.w3schools.com/howto/howto_css_image_responsive.asp)
- Adjust _all_ usages of ImgBanner

# Tests

before: 

https://github.com/user-attachments/assets/d856645c-ed0d-4350-ba70-7677eef855a6

after:

https://github.com/user-attachments/assets/6e1d6bfd-6c35-411a-b269-ab4e79a02a6a

Verification of space reserving on slow loading times:

DappPromoBanner

https://github.com/user-attachments/assets/43cd4c31-56cd-4395-92de-1513acc238ba

DappModal

https://github.com/user-attachments/assets/3d31e390-2e3d-482a-bd8c-8c8946914354

AboutWhyOisyModal

https://github.com/user-attachments/assets/d6f1b336-4ac5-4dfa-9f04-7b88ef8a34a4

Loader

https://github.com/user-attachments/assets/0070069e-94fd-490f-8855-3fa549eff70d


src/frontend/src/lib/components/ui/ImgBanner.svelte
```diff
@@ -11,6 +11,6 @@
 
 <div class="relative block" class:h-56={size === 'big'} class:h-40={size === 'small'}>
 	<div class="absolute h-full w-full">
-		<Img {src} {alt} {width} {height} {fitHeight} />
+		<Img {src} {alt} {width} {height} {fitHeight} styleClass="object-cover" />
```
AntonioVentilii-DFINITY, line 14: just to double confirm: this is the expected behaviors for all of the use cases of `ImgBanner`, correct? Info modal and Loader included

BonomoAlessandro: yes also the images on the loader or the info modal should not be distorted.

StefanBerger-DFINITY: Yes, this is cleaner than a distorted image. We might later want to reduce the height of the image box when it gets squeezed, so its ratio stays in a reasonable range. but also then, cover is better than stretch.

loki344:
Convert to draft  after discussion with @StefanBerger-DFINITY and @BonomoAlessandro

loki344:
@BonomoAlessandro After discussion with @StefanBerger-DFINITY: Instead of cropping the images (object-cover) let's resize the images responsively.
@AntonioVentilii-DFINITY please review

src/frontend/src/lib/components/dapps/DappPromoBanner.svelte
```diff
@@ -1,27 +1,23 @@
...
-				fitHeight={true}
-				height="100%"
-				width="100%"
-				styleClass="object-cover"
-				src={dAppDescription.screenshots[0]}
-				alt={replacePlaceholders($i18n.dapps.alt.website, { $dAppName: dAppDescription.name })}
-			/>
-		</div>
+		<ImgBanner
+			styleClass="bg-brand-subtle-alt h-64 max-h-64"
```
AntonioVentilii, line 15: I think better to keep `bg-brand-subtle-alt` class in the parent `<div>`

src/frontend/src/lib/components/dapps/DappModal.svelte
```diff
@@ -49,9 +49,7 @@
 		{#if nonNullish(screenshots) && screenshots.length > 0}
 			<div class="overflow-hidden rounded-3xl">
 				<ImgBanner
-					fitHeight={true}
-					height="100%"
-					width="100%"
+					styleClass="h-64 max-h-64"
```
AntonioVentilii, line 52: if height is defined, do you need `max-h-`?

src/frontend/src/lib/components/core/Loader.svelte
```diff
@@ -128,7 +128,7 @@
 		<div in:fade={{ delay: 0, duration: 250 }}>
 			<Modal testId={LOADER_MODAL}>
 				<div class="stretch">
-					<ImgBanner width="100%" src={banner} size="small" />
+					<ImgBanner styleClass="" src={banner} />
```
AntonioVentilii, line 131: since `styleClass` defaults to `undefined`, no need to pass it empty here, right?

src/frontend/src/lib/components/dapps/DappPromoBanner.svelte
BonomoAlessandro: I mean like this we do not need a parent `<div>` anymore, which i think is cleaner.

src/frontend/src/lib/components/dapps/DappModal.svelte
BonomoAlessandro: the `max-h-` is needed but the `h-` is unnecessary because the image can shrink and does not have a fixes height

src/frontend/src/lib/components/core/Loader.svelte
BonomoAlessandro: you are right

src/frontend/src/lib/components/dapps/DappPromoBanner.svelte
AntonioVentilii: but then you need to provide `styleClass`, while right now you could provide only `maxSize` prop for example (`large`, `small`, `none`)

BonomoAlessandro: I don't think it makes sense to replace the `styleClass` with a prop `maxSize` that takes values like: `large`, `small`, `none`. At some points we need the `<ImgBanner` with a max height of 64px and at other points with a max height of 56px. This means it would be necessary to provide the max height in px (for example: `maxSize: 64px`) instead of `large`, etc. This would be more or less the same as with the styleClass.

BonomoAlessandro:
@AntonioVentilii-DFINITY pls review

src/frontend/src/lib/components/dapps/DappPromoBanner.svelte
```diff
@@ -1,27 +1,23 @@
...
-				fitHeight={true}
-				height="100%"
-				width="100%"
-				styleClass="object-cover"
-				src={dAppDescription.screenshots[0]}
-				alt={replacePlaceholders($i18n.dapps.alt.website, { $dAppName: dAppDescription.name })}
-			/>
-		</div>
+		<ImgBanner
+			styleClass="bg-brand-subtle-alt max-h-64"
```
AntonioVentilii, line 15: ideally, we don't pass colors to components as general class. May you please put it outside `ImgBanner`?

src/frontend/src/lib/components/dapps/DappPromoBanner.svelte
AntonioVentilii: I think i didn't understand: you are using `styleClass` either for `max-h-64` and `max-h-56` or `none` (where the numbers are not px, but a size value). Why not just using it as defined multiple choice prop?

BonomoAlessandro: @AntonioVentilii Now i moved the `bg-brand-subtle-alt` to the parent `<div>` and this requires that `max-h-64` is also moved to the parent `<div>`. So we do not need the `styleClass` anymore on the '<ImgBanner'.

AntonioVentilii: ah perfect! sorry for the back and forth, but it seems cleaner in a way

AntonioVentilii (approved):

PR #3241, opened 2024-10-30
fix(e2e): Adjust test ID for QR code

AntonioVentilii-DFINITY:
# Motivation

In one of the latest commit, the QR button of the address modal was moved out of the element defined by the test ID of the `token-modal-section`. That means that the E2E test was trying to locate the QR button nested in the element but could not find it.

So, with this PR, we move the test ID for the main `token-modal-section` and we correct the one that should point to the address label.

# Tests

The [Update E2E Snapshots workflow](https://github.com/dfinity/oisy-wallet/actions/runs/11592711081) worked correctly on this branch (there is a commit of the new images too).


DenysKarmazynDFINITY (approved):
LGTM, thanks

peterpeterparker (approved):
Thx for the fix!

PR #3250, opened 2024-10-31
feat(frontend): Hide Send button for zero balance

AntonioVentilii-DFINITY:
# Motivation

We want to hide the Send button in the Hero when the user has no tokens, and display a different message. That is ONLY for the main page, and not the transactions page.

Note that we are talking about token balance and not USD balance, since for some tokens we don't have an exchange rate.

# Changes

- Create util (and tests) to find any possible non-zero balance in the `balancesStore`.
- Create derived to find any non-zero balance, using the util created.
- Adjust `Actions` component to show `Send` button in the main page ONLY when there are at least some non-zero-balance tokens.
- Adjust `ExchangeBalance` to show a difference sentence if there are no tokens.

# Tests

For the derived, I decided to create a specific util and test that one. I think I covered all possible scenarios, but any input welcome.

For the results, here it is (no changes for users with non-zero balances):

Screenshot 2024-10-31 at 09 12 02
Screenshot 2024-10-31 at 09 12 11
Screenshot 2024-10-31 at 09 12 28
Screenshot 2024-10-31 at 09 13 09
Screenshot 2024-10-31 at 09 13 18
Screenshot 2024-10-31 at 09 17 20

# Transition

I tried to add a fade transition for the Send button when suddenly some token arrives, however, it messes up the layout, and since the Button layout is quite nested, I am not sure how to tackle this.

```svelte
{#if sendAction}
	<div in:fade>
		<Send {isTransactionsPage} />
	</div>
{/if}
```

Screenshot 2024-10-31 at 09 17 08


AntonioVentilii-DFINITY:
[OISY-64]

DenysKarmazynDFINITY (approved):
The code LGTM, thanks!

Re. the transition - IMO, not event element should have it. In your particular case, having an animation for a middle button would probably look more like a layout shift. Therefore, I would proceed without it.

src/frontend/src/lib/derived/balances.derived.ts
```diff
@@ -17,3 +18,7 @@ export const balanceZero: Readable<boolean> = derived(
 		nonNullish($balanceStore?.[$token.id]) &&
 		$balanceStore[$token.id]?.data.isZero() === true
 );
+
+export const anyBalanceNonZero: Readable<boolean> = derived([balancesStore], ([$balanceStore]) =>
```
peterpeterparker, line 22: You need a global derived store? Isn't this use in a single place?

src/frontend/src/lib/components/hero/Actions.svelte
```diff
@@ -41,6 +42,9 @@
 
 	let isTransactionsPage = false;
 	$: isTransactionsPage = isRouteTransactions($page);
+
+	let sendAction = false;
+	$: sendAction = $anyBalanceNonZero || isTransactionsPage;
```
peterpeterparker, line 47: Is this correct, your PR description does not mention an exception or am I missing something?

peterpeterparker (approved):
Here maybe a max width to avoid displaying solely "it" on a second line - i.e. displaying more words on the second line?

image

src/frontend/src/lib/components/hero/Actions.svelte
AntonioVentilii-DFINITY: the requirement should be to be hidden ONLY in the main page (not the transaction page), but ONLY when there is at least a token with non-zero balance

you are right, it was not clear in the description: I updated it

src/frontend/src/lib/derived/balances.derived.ts
AntonioVentilii-DFINITY: I am using both in `Actions` (show/hide `Send`) and `ExchangeBalance` (change the string). If you want I can extract it and put it directly in the `Hero` and pass the prop (removing the derived)

peterpeterparker: Already two, I guess it will be quickly three or four usages so makes sense, thanks for the explanation.

peterpeterparker:
I like this one because I firstly read "start using it!", that's a strong call to action.

image

peterpeterparker (approved):

PR #3251, opened 2024-10-31
feat(frontend): Use Zod for Network type

AntonioVentilii-DFINITY:
# Motivation

We change `Network` type to use Zod directly.

### DIFF
--- a/src/frontend/src/lib/types/network.ts
+++ b/src/frontend/src/lib/types/network.ts
@@ -16,18 +16,20 @@ const NetworkAppMetadataSchema = z.object({
 	explorerUrl: UrlSchema
 });
 
+const NetworkSchema = z.object({
+	id: NetworkIdSchema,
+	env: NetworkEnvironmentSchema,
+	name: z.string(),
+	icon: z.string().optional(),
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/types/network.ts
```diff
@@ -16,18 +16,20 @@ const NetworkAppMetadataSchema = z.object({
 	explorerUrl: UrlSchema
 });
 
+const NetworkSchema = z.object({
+	id: NetworkIdSchema,
+	env: NetworkEnvironmentSchema,
+	name: z.string(),
+	icon: z.string().optional(),
```
peterpeterparker, line 23: Not for this PR but, should we be a bit more expressive on the type of the icons?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/types/network.ts
AntonioVentilii-DFINITY: yes, define the type of string for icons, I will have a look

AntonioVentilii: Done with:

- PR #3875

PR #3252, opened 2024-10-31
feat(frontend): Use Zod for Token type

AntonioVentilii-DFINITY:
# Motivation

We ultimately use `Zod` for `Token` type. That requires a few more sub-interfaces to use `Zod` too (for example the ones related to OnRamper).


src/frontend/src/lib/types/token.ts
```diff
@@ -31,32 +31,41 @@ const TokenAppearanceSchema = z.object({
...
+});
+
+const TokenSchema = z
+	.object({
+		id: TokenIdSchema,
+		network: NetworkSchema,
+		standard: TokenStandardSchema,
+		category: TokenCategorySchema
+	})
+	.merge(TokenMetadataSchema)
```
peterpeterparker, line 50: Is `extend` closer to original definition?

peterpeterparker (approved):
Thx!

src/frontend/src/lib/types/token.ts
AntonioVentilii-DFINITY: Unless I am mistaken (I haven't use `.merge` or `.extend` in Zod in a long time), to use `.extend`, I shall do

`.extend(TokenMetadataSchema.shape)`

Or not?

But if I look at `.merge` [documentation](https://zod.dev/?id=merge), it says that it is equivalent to it:

Screenshot 2024-10-31 at 14 22 51


What is the most common way of doing it? WDYT?

peterpeterparker: I don't think you need `.shape`, at least in the `oisy-wallet-signer` I just do extend.

https://github.com/dfinity/oisy-wallet-signer/blob/832a9d237109c99bb1fcec07e1b57c8457728eb3/src/types/rpc.ts#L36

But yeah if it's equivalent then all good.

peterpeterparker (approved):
LGTm, thx

src/frontend/src/lib/types/token.ts
peterpeterparker: Ah maybe it's because in `oisy-wallet-signer` I extend with an object that contains the schema, gotcha.

PR #3255, opened 2024-10-31
feat(frontend): disable send and swap on transactions page if no balance

loki344:
# Motivation

If a user is on the transaction page and has no token balance of the current token, we want to disable actions that need a blance (send, convert). This provides a better UX for the user.

# Changes

- Add property disabled to actions
- Set property disabled for send and convert 

# Tests

Before:

0 Balance:

https://github.com/user-attachments/assets/2b8373b6-22b1-4d2b-8fa4-d347613fef95

Greater 0 Balance:

https://github.com/user-attachments/assets/a3985c52-e177-44d0-a4d3-a21245de62da

After:
0 Balance:

https://github.com/user-attachments/assets/ccc6861d-de71-40d1-a1ef-bf1eb475188e

Greater 0 Balance

https://github.com/user-attachments/assets/95473809-8935-4626-abf6-1a9c48813d03


AntonioVentilii-DFINITY:
With PR #3276 , we introduced the Hero Context, to avoid bubbling the `loading` property (and similar). Is it the case to use it here too, avoiding bubbling the `disabled` function? So for example, you would create some property in the context the identify if `Send` or `Convert` are `disabled`, and call the context in the buttons

loki344:
> With PR #3276 , we introduced the Hero Context, to avoid bubbling the `loading` property (and similar). Is it the case to use it here too, avoiding bubbling the `disabled` function? So for example, you would create some property in the context the identify if `Send` or `Convert` are `disabled`, and call the context in the buttons

Yeah ok, I can do this.

loki344:
@AntonioVentilii-DFINITY I extended the hero store and used the property directly from the context without passing it through the component hierarchy.

src/frontend/src/lib/stores/hero.store.ts
```diff
@@ -3,16 +3,19 @@ import { derived, writable, type Readable, type Writable } from 'svelte/store';
...
-
 	const loaded = derived(loading, ($loading) => !$loading);
 
+	const isExpenseActionsDisabled = writable<boolean>(true);
+
 	return {
 		loading,
-		loaded
+		loaded,
+		isExpenseActionsDisabled
```
AntonioVentilii-DFINITY, line 18: maybe just `outflowActionsDisabled`?

AntonioVentilii-DFINITY (approved):
LGTM, tks!

Small suggestion on naming

src/frontend/src/lib/stores/hero.store.ts
loki344: Yeah ok

PR #3256, opened 2024-10-31
refactor(frontend): Use colors mapping instead of blue-ribbon

AntonioVentilii-DFINITY:
# Motivation

Only for Tailwind, we want to replace the use of `blue-ribbon` (`primary`) with the new colors mapping.


peterpeterparker (approved):
LGTM, thx

PR #3257, opened 2024-10-31
refactor(frontend): Separate module for token-group

AntonioVentilii-DFINITY:
# Motivation

We re-organize the `token-group` related functions in a separate file.


peterpeterparker (approved):
LGTm, thx

PR #3262, opened 2024-10-31
fix(frontend): Quickfix sorting for Token Groups

AntonioVentilii-DFINITY:
# Motivation

It may happen that after the grouping, a particular group results with a balance that is greater than another group/token above it. So the sorting could be wrong:

image

We are going to refactor the grouping util, but not in this PR.

For now, it was requested a "quickfix" to at least have the correct sorting. To achieve that, we sort the final list of `[tokens + groups]` by balances (token and USD).

It is guaranteed that the only sorting criteria that is being mixed is the one concerning the balances, because a group could only have a greater (or same) balance than the single tokens it was born from.

The other sorting criteria (name, network.name) can be ignored because the resulting position given by them will be maintained during grouping.


AntonioVentilii-DFINITY:
[OISY-244]

src/frontend/src/lib/utils/token-group.utils.ts
```diff
@@ -81,7 +82,11 @@ export const groupTokensByTwin = (tokens: TokenUi[]): TokenUiOrGroupUi[] => {
 		return token;
 	});
 
-	return mappedTokensWithGroups.filter(
-		(t) => isTokenUiGroup(t) || !groupedTokenTwins.has(t.symbol)
-	);
+	return mappedTokensWithGroups
+		.filter((t) => isTokenUiGroup(t) || !groupedTokenTwins.has(t.symbol))
+		.sort(
```
peterpeterparker, line 87: Can we have a test?

AntonioVentilii-DFINITY: Sorry, forgot! Just included, have a look please

peterpeterparker (approved):
LGTm, thx

PR #3263, opened 2024-10-31
refactor(frontend): Move third parties ENV variables to folder

AntonioVentilii-DFINITY:
# Motivation

We will have more third-parties ENV variables to remove from the ENV files. So, to be more organized, we create a specific folder in path `/env` , and we start moving Onramper variables there.


peterpeterparker (approved):
You know there is way too many Web2 third party providers used when you need to start structuring their environment variables 🥲

LGTM, thx

PR #3264, opened 2024-10-31
refactor(frontend): hard-code Coingecko ENV variables

AntonioVentilii-DFINITY:
# Motivation

The URL for Coingecko API can be removed from the .env files and hard-coded, since it is constant for all environment.

That said, we move all Coingecko-related "env" variables to a specific module in `/env` folder

# Changes

- Hard-code Coingecko API URL.
- Move API URL and KEY variables to an `env` module.
- Adapt the code.
- Remove the API URL from the template `.env` files.


src/frontend/src/env/rest/coingecko.env.ts
```diff
@@ -0,0 +1,5 @@
+import { UrlSchema } from '$lib/validation/url.validation';
+
+export const COINGECKO_API_URL = UrlSchema.parse('https://pro-api.coingecko.com/api/v3/');
```
peterpeterparker, line 3: `parse` can throw an error at runtime which can per extension lead to the all app to not mount since this is executed at the root of the module.

We should probably rather use `safeParse`?

AntonioVentilii-DFINITY: my idea is that since they are hard-coded, the error would appear safely with CIs failing, in any PR that changes/updates/modifies it

so we would be aware of it, before merging a possible not-valid-URL

with `safeParse`, if the URL is wrongly parsed, the app would work either way, but only the `COINGECKO_API` functions will possibly fail

peterpeterparker: How the CI would fail? `safe` is a runtime feature.

AntonioVentilii-DFINITY: What I meant was that with `parse` will fail, and that would be good

### `parse`

- CI runs and find an error in parsing --> fails --> we know that the URL we wrote in the code is somehow wrong

### `safeParse`

```typescript
export const COINGECKO_API_URL = UrlSchema.safeParse(...).success ? ... : fallbackValue;
```

- CI runs --> there is an error in parsing, but it falls back on the fallback value --> we think it is all good, but the Coingecko features are not working


Or you mean to not have a fallback value but raise the error? Maybe I misunderstood what you meant as implementation

NOTE: with E2E tests obviously this is not a discussion eheh

peterpeterparker: I guess the CI don't run E2E tests against the production build, hence the CI won't run the production application and it can potentially ends with an application that does not start / mount.

It would be an edge case but, it's still a possible case therefore I think it's not safe to actually use `parse` at the root of a module / of the application without catching potential errors.

AntonioVentilii-DFINITY: yes, makes sense! What do you think would be better then? Throw an error in case of parsing unsuccessful or just fallback to an empty string?

peterpeterparker: What's the current behavior if someone forgot to setup an env variable, empty string?
Should we just keep that approach (for the best or worst) and iterate if necessary?

AntonioVentilii-DFINITY: it would default to an `undefined`, so I keep the same; have a look at last commit

peterpeterparker: > it would default to an undefined, so I keep the same

I'm confuse, current code default to `''`

AntonioVentilii-DFINITY: very sorry, I forgot to push:

- Before: if no env variable is set, it is `undefined`;
- After: if `safeParse` fails, it falls back on `undefined`

src/frontend/src/env/rest/coingecko.env.ts
```diff
@@ -0,0 +1,9 @@
+import { UrlSchema } from '$lib/validation/url.validation';
+
+const parsedCoingeckoApiUrl = UrlSchema.safeParse('https://pro-api.coingecko.com/api/v3/');
```
peterpeterparker, line 3: If you want fancyness `const {success, data} = ...` 😉

peterpeterparker (approved):
LGTM, thx

src/frontend/src/env/rest/coingecko.env.ts
peterpeterparker: Thx

src/frontend/src/env/rest/coingecko.env.ts
AntonioVentilii-DFINITY: good one!

PR #3266, opened 2024-10-31
refactor(frontend): Blockstream API URL is hard-coded

AntonioVentilii-DFINITY:
# Motivation

The Blockstream API URL can be hard-coded since it is constant throughout the envs.


DenysKarmazynDFINITY (approved):
LGTM, thanks!

PR #3271, opened 2024-11-01
feat(frontend): Bubble loading prop of ButtonHero

AntonioVentilii-DFINITY:
# Motivation

We want to slowly bubble the loading prop of `ButtonHero`, so that we can manage it directly in the `Hero`. However, to avoid creating a huge PR, we go step by step.

### DIFF
--- a/src/frontend/src/btc/components/convert/ConvertToCkBTC.svelte
+++ b/src/frontend/src/btc/components/convert/ConvertToCkBTC.svelte
@@ -4,9 +4,11 @@
 	import ButtonHero from '$lib/components/ui/ButtonHero.svelte';
 	import { isBusy } from '$lib/derived/busy.derived';
 	import { i18n } from '$lib/stores/i18n.store';
+
+	export let loading = false;
 </script>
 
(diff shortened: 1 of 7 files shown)


AntonioVentilii-DFINITY:
[OISY-64]

peterpeterparker (commented):
As you mention that you are planing multiple PRs to bubble props, have you though about a context and would it avoid too many level of properties drilling? Or does not make sense here?

AntonioVentilii-DFINITY:
@peterpeterparker I actually first thought of a global store/derived eheh! Never thought about the context, but it most probably be quite the same; shall I have a try at it and come back to you?

peterpeterparker:
> it most probably be quite the same; shall I have a try at it and come back to you?

Worth having a look if there is 3 or 4 level down of passing properties but, sure if it's just one or two, that's maybe an overkill.

AntonioVentilii-DFINITY:
@peterpeterparker I think it is better! Good suggestion! Have a look:

- PR #3276

AntonioVentilii-DFINITY:
This was superseded by PR #3276 , as discussed above

PR #3272, opened 2024-11-01
refactor(frontend): Simplify props of Hero

AntonioVentilii-DFINITY:
# Motivation

The props given to `Hero` are somehow a bit confused. They were ok before, since the `Hero` was used in all routes. With the latest changes, the scope of the `Hero` was reduced to only the main page (tokens) and the transaction page.

To clean-up a bit we refactor the Hero to be a bit more obvious.

# Changes

- Create types for routes and for the routes for the `Hero`.
- Pass `route` prop to `Hero` related component.
- Use `route` to define the content of the Hero.
- Remove deprecated props `summary` and `usdTotal`.

# Tests

No changes in local replica.

### DIFF
--- a/src/frontend/src/lib/components/hero/HeroContent.svelte
+++ b/src/frontend/src/lib/components/hero/HeroContent.svelte
@@ -16,11 +16,11 @@
 	import { networkBitcoin, networkEthereum, networkICP } from '$lib/derived/network.derived';
 	import { pageToken } from '$lib/derived/page-token.derived';
 	import { balancesStore } from '$lib/stores/balances.store';
+	import type { HeroRoute } from '$lib/types/route';
 	import type { OptionTokenUi } from '$lib/types/token';
 	import { mapTokenUi } from '$lib/utils/token.utils';
 
-	export let usdTotal = false;
-	export let summary = false;
+	export let route: HeroRoute;
 
 	let pageTokenUi: OptionTokenUi;
 	$: pageTokenUi = nonNullish($pageToken)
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/hero/HeroContent.svelte
```diff
@@ -68,9 +68,7 @@
 
 			<Balance token={pageTokenUi} />
 		</div>
-	{/if}
-
-	{#if usdTotal}
+	{:else}
```
AntonioVentilii-DFINITY, line 71: this is guaranteed to be `'tokens'` route because the type of `HeroRoute` is either `'transactions'` or `'tokens'`

src/frontend/src/lib/types/route.ts
```diff
@@ -0,0 +1,3 @@
+export type Route = 'transactions' | 'tokens' | 'settings' | 'explore';
```
peterpeterparker, line 1: I'm not convinced by have a global type route that actually does not match the routing of `$page`.

`let route: ...` was just a handy local variable.

In NNS dapp if I remember correctly, we can have a look, I think we use an enum that contains the route names.

peterpeterparker: For example in NNS dapp it's

```
export enum AppPath {
  Accounts = "/accounts",
  Wallet = "/wallet",
```

With `/accounts` being the href.

```
export const ROUTE_ID_GROUP_APP = "/(app)";
export const ROUTE_ID_GROUP_UNIVERSE = "/(u)";
```


```
$: routes = [
    {
      context: "accounts",
      href: AppPath.Tokens,
      selected: isSelectedPath({
        currentPath: $pageStore.path,
        paths: [AppPath.Accounts, AppPath.Wallet, AppPath.Tokens],
      }),
      title: $i18n.navigation.tokens,
      icon: IconWallet,
    },
    ...
```

AntonioVentilii:
Superseded by:

-

PR #3277, opened 2024-11-01
feat(frontend): util to safe parse schemas

AntonioVentilii-DFINITY:
# Motivation

Small util to have an out-of-the-box `safeParse` function.

UNRELATED: the `validation` folder for tests was outside `lib`, so we moved it.


src/frontend/src/lib/validation/utils.validation.ts
```diff
@@ -0,0 +1,19 @@
...
+type SafeParseReturn<T, Fallback> = Fallback extends undefined ? T | undefined : T;
+
+export const safeParse = <T, Fallback extends T | undefined = undefined>({
+	schema,
+	value,
+	fallback = undefined
+}: SafeParseParams<T, Fallback>): SafeParseReturn<T, Fallback> => {
+	const { success, data } = schema.safeParse(value);
+
+	return success ? data : (fallback as SafeParseReturn<T, Fallback>);
```
peterpeterparker, line 18: I guess no other way to avoid casting?

peterpeterparker (approved):
Super duper cool! Thx

src/frontend/src/lib/validation/utils.validation.ts
AntonioVentilii-DFINITY: unfortunately not... I tried/searched a lot and it seems a limitation of the conditional typing; however it works: simulating the exact return type for the cases where we specify a `fallback`

Screenshot 2024-11-01 at 22 05 08
Screenshot 2024-11-01 at 22 05 18

and if we remove the fallback

PR #3276, opened 2024-11-01
style(frontend): Make ButtonHero loading synchronized with Balance using HeroContext

AntonioVentilii-DFINITY:
# Motivation

We want to make the Buttons in the Hero in loading state, synchronized with the loading of the balance.

We have 2 situations:

- Main page --> `loading` is linked to exchange store initialized
- Transactions page --> `loading` is linked to the token balance being nullish

To transmit the loading information we create a Hero Context (suggestion by @peterpeterparker in [comment](https://github.com/dfinity/oisy-wallet/pull/3271#pullrequestreview-2409866758) of PR #3271)

# Changes

- Create `HeroContext` with `loading` store.
- Initiate `HeroContext` in the `Hero` setting it by default to `false` and updating according to page logic.
- Adapt balance components to fetch the `HeroContext`.
- Set the `ButtonHero` component (used ONLY for the `Hero` actions) to use the same context.

# Tests

https://github.com/user-attachments/assets/09f9ddaf-0766-4483-90d1-def1f7b97b8d

PR #3279, opened 2024-11-01
refactor(frontend): Move ButtonHero to Hero folder

AntonioVentilii-DFINITY:
# Motivation

As discussed in PR #3276 , `ButtonHero` is not an UI component anymore, but more related to `Hero` (we introduced hero context). So we move it to the appropriate folder.


peterpeterparker (approved):
Thanks!

peterpeterparker:
All those imports makes me think it wasn't that a bad decision to use a context. 😃

AntonioVentilii-DFINITY:
And that was just the first level! Eheh

PR #3280, opened 2024-11-01
style(frontend): Add error colors to mapping

AntonioVentilii-DFINITY:
# Motivation

We include the `error` colors from the Design library, to be re-used as standard in the code (for now only Tailwind).


peterpeterparker (approved):
LGTm, thx

PR #3293, opened 2024-11-04
feat(frontend): toggle addresses

peterpeterparker:
# Motivation

# Changes

# Tests


peterpeterparker:
Discussed offline we are not going to implement this now as the feature, filtering the addresses, lead to some other questions.

PR #3295, opened 2024-11-04
refactor(frontend): remove prefix from new colors

AntonioVentilii-DFINITY:
# Motivation

It is too verbose to have bg-background- or border-border- for the new color mapping. So, we simplify it extending the colors of specific scopes with the correct mapping:

1. Background: `bg-` and `gradient-`
2. Border: `border-` and `ring-`
3. Foreground: `text-`

Thanks @peterpeterparker for the suggestion!


peterpeterparker (approved):
LGTM, thx

PR #3296, opened 2024-11-04
feat(frontend): Utils to group main and secondary tokens

AntonioVentilii-DFINITY:
# Motivation

We want to change the way we group tokens, to have more than two elements in the group, and to try and avoid repetitions/glitches when the order of the token list is not the usual one.

# Definitions

### Main Token

The "original" token of a certain group or of a certain other token; for example `BTC` is the "main token" of `ckBTC`, or `ETH` is the "main token" of `SepoliaETH`. Typically, a token and its "main token" should be 1:1.

### Secondary token

A token that has a "main token", even if not present in the list. It is just to indicate that it should always be grouped if the latter is present.

# Implementation

To improve the grouping util, we first break it down into smaller utils: one for grouping a `main` token, and one to create a `secondary` token.

If the token has no "main token", it is either a "main token" for an existing group,
or a "main token" for a group that still does not exist, or a single-element group (but still its "main token").

If a group already exists for the "main token" of the current token, add it to the existing group.
Otherwise, create a new group with the current token as a placeholder "main token".
This is to avoid that the group is created with an empty "main token", if the current token's "main token" is not in the list.
Instead, it should be considered a single-element group, until the "main token" may or may not be found.

# Tests

Created tests for new utils.

### DIFF
--- a/src/frontend/src/lib/utils/token-group.utils.ts
+++ b/src/frontend/src/lib/utils/token-group.utils.ts
@@ -5,9 +5,11 @@ import type { TokenUi } from '$lib/types/token';
 import type { TokenUiGroup, TokenUiOrGroupUi } from '$lib/types/token-group';
 import {
 	isRequiredTokenWithLinkedData,
+	sumBalances,
 	sumTokenBalances,
 	sumUsdBalances
 } from '$lib/utils/token.utils';
+import { nonNullish } from '@dfinity/utils';
 
 /**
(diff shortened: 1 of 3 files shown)


AntonioVentilii-DFINITY:
[OISY-164]
[OISY-141]
[OISY-142]

peterpeterparker (approved):
LGTM, thx

PR #3300, opened 2024-11-04
test(frontend): transactions store

peterpeterparker:
# Motivation

Before breaking everything, here's a bare minimal test for the `transactions.store`.


src/frontend/src/tests/lib/stores/transactions.store.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+
+				store.subscribe((state) => {
+					expect(state?.[tokenId]).toHaveLength(2);
+
+					done();
+				})();
+			}));
+	});
+
+	describe('append', () => {
```
AntonioVentilii-DFINITY, line 64: maybe here the same test to check if no duplicates? as the one for prepend?

src/frontend/src/tests/lib/stores/transactions.store.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+				const store = initTransactionsStore<IcTransactionUi>();
+
+				const transactions = [
+					createCertifiedTransaction('tx1'),
+					createCertifiedTransaction('tx2'),
+					createCertifiedTransaction('tx3')
+				];
+
+				store.append({ tokenId, transactions });
+				store.cleanUp({ tokenId, transactionIds: ['tx1', 'tx3'] });
```
AntonioVentilii-DFINITY, line 97: if you add `tx4` just to check the behavior for non existing transactions, does the test fails? or is it no issue? I don't remember the code

AntonioVentilii-DFINITY (approved):
Already good, tks!

A few suggestions for further tests, but only if you think necessary

src/frontend/src/tests/lib/stores/transactions.store.spec.ts
peterpeterparker: that's actually a good point, turns out `append` do not deduplicate. Should we improve the store in another PR and provide the test at the same time?

src/frontend/src/tests/lib/stores/transactions.store.spec.ts
peterpeterparker: It just do nothing. Added a test 632a63dfa38f32a61ee2290f47a43ba19bdbfe19.

src/frontend/src/tests/lib/stores/transactions.store.spec.ts
AntonioVentilii-DFINITY: Yes! I didn't even know that it wasn't doing it, good thing you tested!

PR #3301, opened 2024-11-04
feat(frontend): nullify transactions store

peterpeterparker:
# Motivation

We want to add support for tokens without Index canister which means tokens won't have transactions. As a result we decided to extend transactions store with a `nullify` option.

# Changes

- Set content of the transactions store to potentially be `null`.
- Add function `nullify`.
- Mark derived store as returning `NonNullable` answer.
- Add a type in BTC store for consistency.


AntonioVentilii-DFINITY (approved):
Good naming! LGTM! Tks

PR #3302, opened 2024-11-04
style(frontend): fix modal covering

BonomoAlessandro:
# Motivation

The vh measurement, shorthand for Viewport Height, refers to the height of the viewing area of your device or browser. When you define something as 100vh, it means it should take up 100% of the viewport height. However, this measure doesn't consider changes in the size of the browser's interface elements like the URL bar. Consequently, when these elements alter their size dynamically, your modals that use vh could end up being partially hidden under these elements, as 100vh can sometimes span beyond the visible area.
To solve this, we use 100dvh, where dvh stands for dynamic viewport height. This measure dynamically recalculates the height of the viewport (vh) considering any changes to the display of the browser's interface elements, like the URL bar. With 100dvh, you're ensuring that your modals are displayed within the actually visible part of the page, rather than being potentially hidden by dynamically changing browser interfaces.

# Changes

- uses dynamic height for modals

# Tests

**before:**

Brave browser:
image

Chrome browser:
image

**after:**

Brave browser:
image

Chrome browser:
image


src/frontend/src/lib/styles/global/gix.scss
```diff
@@ -60,7 +60,7 @@
 
 	--alert-width: calc(100% - var(--padding-8x));
 	--alert-max-width: calc(100vw - var(--padding-4x));
-	--alert-max-height: calc(100vh - var(--padding-4x));
+	--alert-max-height: calc(100dvh - var(--padding-4x));
```
peterpeterparker, line 63: `dvh` is not supported yet on Firefox (https://caniuse.com/?search=dvh) we should use the same snippet as @AntonioVentilii-DFINITY introduced last Friday.

BonomoAlessandro: @peterpeterparker isn't it just the static method `dvh()` that is not supported by Firefox? The Viewport units that i used are supported.
image

peterpeterparker: Oh great then, my bad!

cc @AntonioVentilii-DFINITY

PR #3310, opened 2024-11-04
test(frontend): should duplicate transactions with same id

peterpeterparker:
# Motivation

As discussed on- and offline, `transactionsStore.append` should not deduplicate transactions with same ID.

Follow-up of PR: https://github.com/dfinity/oisy-wallet/pull/3300#discussion_r1827906055

# Changes

- Add a test which duplicate ID


AntonioVentilii-DFINITY (approved):
thank you!

PR #3311, opened 2024-11-04
chore(frontend): vitest is global - no import needed

peterpeterparker:
# Motivation

Just a bit of cleanup since most `vitest` imports can be spared as those are available globally.


AntonioVentilii-DFINITY (approved):
TKS!

PR #3314, opened 2024-11-04
fix(e2e): Make ButtonHero loaded when exchange disabled

AntonioVentilii-DFINITY:
# Motivation

The buttons in the Hero were permanently loading when the Exchange is are not enabled (never initialized). That mostly happens in test environments, such as the one used for E2E test. We can see the artifacts of [workflow E2E test ran on main](https://github.com/dfinity/oisy-wallet/actions/runs/11662656812):

Screenshot 2024-11-04 at 18 43 08

To solve it, we include the flag for enabled exchanges among the checks for loading `ButtonHero`.

# Changes

- Extract ENV variable `EXCHANGE_DISABLED`.
- Add it to set function for `HeroContext`.
- Disable exchanges for playwright.

# Note

One alternative could have been to check directly the ENV variable that has the API_KEY for the exchanges, but this approach felt more re-usable.

# Tests

I run the [screenshots workflow on this branch](https://github.com/dfinity/oisy-wallet/actions/runs/11670799340) and it worked.


playwright.config.ts
```diff
@@ -11,6 +11,9 @@ dotenv.populate(
 	})
 );
 
+// We disable the exchange in the E2E tests to load the buttons in the Hero
+process.env.VITE_EXCHANGE_DISABLED = 'true';
```
peterpeterparker, line 15: You cannot solve that with a particular `.env` file for the e2e test?

src/frontend/src/lib/components/hero/HeroContent.svelte
```diff
@@ -45,7 +46,9 @@
 	});
 
 	$: loading.set(
-		isRouteTransactions($page) ? isNullish(pageTokenUi?.balance) : !$exchangeInitialized
+		isRouteTransactions($page)
+			? isNullish(pageTokenUi?.balance)
+			: !EXCHANGE_DISABLED && !$exchangeInitialized
```
peterpeterparker, line 51: If the `EXCHANGE_DISABLED`, should `$exchangeInitialized` always be true?

AntonioVentilii-DFINITY: I thought about it and i didn't want to touch that derived, because in the end, it is not initialized, right? it is useful to have a derived that checks the initialization even if the exchange is enabled

playwright.config.ts
peterpeterparker: > Github Secrets that is for the E2E CI

Not sure, why?

I was thinking something like

```
# .env.e2e
VITE_EXCHANGE_DISABLED = 'true'
```

and

```
# playwright.config.ts
import { config } from 'dotenv';

config({ path: `.env.e2e` });
```

Not sure if it works.I would just avoid having business flag defined in `playwright.config.ts`.

src/frontend/src/lib/components/hero/HeroContent.svelte
peterpeterparker: On the other hand, we use `$exchangeInitialized` in various places, so we are introducing a discrepancy here that is not covered by any particular tests.

playwright.config.ts
AntonioVentilii: ah! you mean to add the `.env.e2e` file to the repo, like `.env.template` or `.env.example` ? yes, that would work

peterpeterparker: Yes

AntonioVentilii: done!

src/frontend/src/lib/components/hero/HeroContent.svelte
AntonioVentilii: mmmm probably you are right... in the end, that derived store shall say that the exchange are "initialized" if they are not enabled

PR #3315, opened 2024-11-04
ci(e2e): simplify script to install playwright

AntonioVentilii-DFINITY:
# Motivation

There is no need to install `playwright` twice, we can do it once directly with deps.


peterpeterparker (approved):
LGTM, thx

PR #3317, opened 2024-11-04
test(frontend): Tests for certified store init

AntonioVentilii-DFINITY:
# Motivation

We create some base tests for `initCertifiedStore` and `initCertifiedSetterStore`

Note that we are not creating tests for the `update` function since it is replicated as it is from `writable` constructor of `svelte`.


AntonioVentilii:
Since it was my old account, it is superseded by:

- PR #

PR #3319, opened 2024-11-04
refactor(frontend): remove deprecated logged out status in Tokens

AntonioVentilii-DFINITY:
# Motivation

The main page that loads the `Tokens` component has currently an Authorization guard, so we will never have access to it with the user signed out.

# Changes

- Remove all auth-related conditions in `Tokens` component.
- Delete deprecated components: `TokensSignedOut` and `TokenCardSignedOut`


peterpeterparker (approved):
LGTM, thx

In a next PR we rename `TokensSignedIn`?

PR #3320, opened 2024-11-04
refactor(frontend): remove deprecated logged out status in Settings

AntonioVentilii-DFINITY:
# Motivation

The route that loads the `Settings` component has currently an Authorization guard, so we will never have access to it with the user signed out.

# Changes

- Remove all auth-related conditions in `Settings` component.
- Delete deprecated component `SettingsSignedOut`


peterpeterparker (approved):
LGTM, thx

We rename `SettingsSignedIn` afterwards? Or we just merge it into `Settings`?

PR #3321, opened 2024-11-04
refactor(frontend): remove deprecated logged out status in Transactions

AntonioVentilii-DFINITY:
# Motivation

The route that loads the `Transactions` component has currently an Authorization guard, so we will never have access to it with the user signed out.

# Changes

- Remove all auth-related conditions in `Transactions` component.
- Delete deprecated component `TransactionsSignedOut`


peterpeterparker (approved):
LGTM, thx

LGTM, thx

We rename `TransactionsSignedIn` afterwards? Or we just merge it into `Transactions`?

PR #3323, opened 2024-11-04
refactor(frontend): simplify utils filterTokensForSelectedNetwork

AntonioVentilii-DFINITY:
# Motivation

Util function `filterTokensForSelectedNetwork` was using `pseudoNetworkChainFusion` derived, that however is by itself a simple derived of `selectedNetwork` (another dependency that `filterTokensForSelectedNetwork` is already using).

So, we simplify it, removing `pseudoNetworkChainFusion` dependency and substituting with its logic.

# Changes

- Replace `pseudoNetworkChainFusion` dependency from filterTokensForSelectedNetwork with `isNullish(selectedNetwork)`
- Remove deprecated `pseudoNetworkChainFusion` dependency.
- Adapt the code.
- Create tests.

# Tests

Furthermore, we create some tests for `filterTokensForSelectedNetwork`, that were missing.

### DIFF
--- a/src/frontend/src/lib/components/manage/ManageTokens.svelte
+++ b/src/frontend/src/lib/components/manage/ManageTokens.svelte
@@ -101,8 +101,7 @@
 			...(manageEthereumTokens ? allErc20Tokens : []),
 			...(manageIcTokens ? allIcrcTokens : [])
 		],
-		$selectedNetwork,
-		$pseudoNetworkChainFusion
+		$selectedNetwork
 	]);
 
 	let allTokensSorted: Token[] = [];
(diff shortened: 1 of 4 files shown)


peterpeterparker (commented):
I stopped reviewing the code because the PR is labeled as "test" but provides various features and fixes. Could you review the description of the PR? If it addresses multiple concerns, could you consider splitting it as well?

AntonioVentilii-DFINITY:
@peterpeterparker you are right, changed

AntonioVentilii:
Old account, replacing with

PR #3324, opened 2024-11-04
refactor(frontend): use enum for app paths

AntonioVentilii-DFINITY:
# Motivation

Inspired by [`nns-dapp` codebase](https://github.com/dfinity/nns-dapp/blob/main/frontend/src/lib/constants/routes.constants.ts), we use an enum to define the routes of the App.

Unrelated: we remove unused function `isSubRoute`.


src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -30,14 +36,14 @@ const tokenUrl = ({
 	path
 }: {
 	token: Token;
-	path: '/transactions/' | '/';
+	path: AppPath.Transactions | undefined;
```
peterpeterparker, line 39: It couldn't be `undefined` before, the previous description was incorrect?

peterpeterparker (approved):
LGMT, thx

Maybe we can add some tests before (ideally) or after for `nav.utils` as it seems not covered yet.

src/frontend/src/lib/utils/nav.utils.ts
AntonioVentilii-DFINITY: I removed the leading '/' from the input, and moved it inside the function here

``${path ?? ''}/?${TOKEN_PARAM}=...`

so it seemed more correct to have `undefined` input instead of empty string

peterpeterparker: When the user hits `https://oisy.com` the path will be undefined?

AntonioVentilii: correct!

for example

`https://oisy.com/?token=ETH`

will have undefined path

peterpeterparker (approved):
LGTM, thx

Issue #2883, opened 2024-10-15
Make index canister as optional 


tcpim:
**Is your feature request related to a problem? Please describe.**

I love the progress on Oisy! I am confident that Oisy has the potential to be the dominant self custodial wallet in IC ecosystem. For Defi users, it is critical to import any tokens with just the ledger canister because not all tokens have index canister. Say there is a new popular meme token and people want to use Oisy in ICPSwap to trade it, we need to import the token no matter if it has the index canister.

**Describe the solution you'd like**
Just like the new feature in NNS https://www.youtube.com/watch?v=zcXmarBk2L4, Oisy can convey a clear message that without index canister, you won't be able to see the transaction history. 

**Describe alternatives you've considered**

**Additional context**
Congrats on the good work so far! Oisy has a really huge potential, especially with the ability to connect via WalletConnect to connect to the outside world. But first, the most wanted feature in IC right now is a easy to use and reliable self custody wallet for Defi. Plug and Stoic is no longer good. And we need Oisy to take the lead. Thank you!

tcpim:
Also added a forum post here https://forum.dfinity.org/t/can-oisy-support-only-ledger-canister/36185. Thanks!

AntonioVentilii-DFINITY:
@tcpim   Thank you for the suggestion and the praises! Eheh! The entire team (and beyond) is working very hard on the project and it is always good to hear feedbacks!

Regarding the feature, we are including it in our backlog, and we will try to scope it as soon as it we are able to; we will keep you informed

tcpim:
@AntonioVentilii-DFINITY Does Oisy have plan to let IC Dapp to connect to Oisy? Wondering what would be the experience. Do users need to approve/sign when initiate a transfer after connect to Oisy?

peterpeterparker:
> Does Oisy have plan to let IC Dapp to connect to Oisy?

@tcpim Yes, we are currently reviewing the necessary work for this.

> Do users need to approve/sign when initiate a transfer after connect to Oisy?

Yes, the process is similar to the flow you might already be familiar with in other wallets.

On our side, the work is still in progress, and the related library will soon be open-sourced. In the meantime, you can check out [Identity Kit](https://www.identitykit.xyz/), which is a client for the new ICRC standards. These capabilities are being built on top of those standards.

[...]

PR #3326, opened 2024-11-05
feat(frontend): ic token zod

peterpeterparker:
# Motivation

Convert ic token related definition to zod.

# Changes

- Extract schema validation and tests
- Convert types
- Reorder types to avoid having declaration referencing types declared afterwards


src/frontend/src/lib/types/token.ts
```diff
@@ -40,7 +40,7 @@ const TokenBuyableSchema = z.object({
 	buy: z.custom<AtLeastOne<TokenBuy>>().optional()
 });
 
-const TokenSchema = z
+export const TokenSchema = z
```
peterpeterparker, line 43: I did not provided a test for this. I guess we should move it to a `token.validation` in another PR and add a test. I can do it before this PR if you wish?

peterpeterparker: https://github.com/dfinity/oisy-wallet/pull/3340

AntonioVentilii (commented):
LGTM tks!

One note: when I did it for the `Token` type, there were a few sub-types that required only the schema, and not the `infer type of...`. We currently use all of them? Apart for the tests I mean

AntonioVentilii (approved):
LGTM tks!

One note: when I did it for the Token type, there were a few sub-types that required only the schema, and not the infer type of.... We currently use all of them? Apart for the tests I mean

peterpeterparker:
> We currently use all of them?

Only one was not use, `IcAppMetadata`. I prefered to not removed it here since it was not really the goal of the PR.

peterpeterparker:

PR #3327, opened 2024-11-05
feat(frontend): principal zod validation

peterpeterparker:
# Motivation

A zod schema for principal as text. We need it to assert for example `CanisterIdText`.

Copied from https://github.com/dfinity/oisy-wallet-signer


AntonioVentilii-DFINITY (approved):
LGTM! Tks!

P.S. this seems something useful to have in `ic-js` too

src/frontend/src/lib/validation/principal.validation.ts
```diff
@@ -0,0 +1,16 @@
...
+	(principal) => {
+		try {
+			Principal.fromText(principal);
+			return true;
+		} catch (err: unknown) {
+			return false;
+		}
+	},
+	{
+		message: 'Invalid textual representation of a Principal.'
```
AntonioVentilii-DFINITY, line 14: here I imagine no `i18n` because is dev error

peterpeterparker: Because I copied the code 😉.

But yeah I would say it's a low level dev error, not sure we should add i18n there.

peterpeterparker:

PR #3330, opened 2024-11-05
refactor(frontend): Extract About components folder

AntonioVentilii-DFINITY:
# Motivation

Components folder `About` was nested in `Hero`. We extract it, since it is not an Hero-only topic anymore.


peterpeterparker (approved):
LGTM, thx

PR #3339, opened 2024-11-05
style(frontend): improves dapps explorer cards

BonomoAlessandro:
# Motivation

Before, when the title was longer or there were too many tags, they would stick out of the card. Now, the tags are fixed to the bottom of the card and the displayed text can take as much space as needed. This makes the dAppCard look cleaner and simpler.

# Changes

- lines that can be used for tags
- lines that can be used for title
- lines that can be used for text

# Tests

**before:**
desktop:
image

mobile:
image

**after:**
desktop:
image

mobile:
image


src/frontend/src/lib/components/dapps/DappCard.svelte
```diff
@@ -2,11 +2,29 @@
...
+	import type {
+		OisyDappDescription,
+		OisyDappDescriptionDimension
+	} from '$lib/types/dapp-description';
 	import { replacePlaceholders } from '$lib/utils/i18n.utils';
 
 	export let dAppDescription: OisyDappDescription;
 	$: ({ name: dAppName, logo, oneLiner, tags } = dAppDescription);
+
+	let calculateClampValue = (dAppDescription: OisyDappDescriptionDimension) =>
```
peterpeterparker, line 14: Please use stricly CSS for styling purpose.

BonomoAlessandro: It seems unlikely that this can be achieved with plain CSS. While the `text-overflow: ellipsis` feature allows line clamping, it doesn't support multi-line clamping. For this, we need to use the `--webkit-line-clamp: x` property, where x defines the number of lines displayed. However, in our case, the number depends on the number of tags and the title. Hence, it appears that we'll have to calculate it using TypeScript.

peterpeterparker: Thanks for the explanation.

Then we should consider a different approach to this problem. Styling with JavaScript ddoesn’t align with the project’s overall codebase and approach (separation of concerns, maintainability, adherence to best practices, etc.).

BonomoAlessandro:
After discussion with @StefanBerger-DFINITY: Now, we position the tags fixed at the bottom, and the displayed text can flow beneath them. This is acceptable because we have control over the texts and can modify them as needed.

BonomoAlessandro:
@peterpeterparker pls review

peterpeterparker:
The lack of spacing in this screenshot is expected?

383505270-1f3bf121-e8bd-41ac-bd30-653edafbb1c9

BonomoAlessandro:
Yess, this is expected. The displayed text can move under the Tags which looks bad but we do have control over the text so we can change the text if needed. This is discussed with @StefanBerger-DFINITY

peterpeterparker:
Thanks for confirming.

peterpeterparker (approved):
LGTM, thx

PR #3344, opened 2024-11-05
test(frontend): test utils filterTokensForSelectedNetwork

AntonioVentilii:
# Motivation

Util function `filterTokensForSelectedNetwork` was using `pseudoNetworkChainFusion` derived, that however is by itself a simple derived of `selectedNetwork` (another dependency that `filterTokensForSelectedNetwork` is already using).

So, we simplify it, removing `pseudoNetworkChainFusion` dependency and substituting with its logic.

# Changes

- Replace `pseudoNetworkChainFusion` dependency from filterTokensForSelectedNetwork with `isNullish(selectedNetwork)`
- Remove deprecated `pseudoNetworkChainFusion` dependency.
- Adapt the code.
- Create tests.

# Tests

Furthermore, we create some tests for `filterTokensForSelectedNetwork`, that were missing.

### DIFF
--- a/src/frontend/src/lib/derived/network-tokens.derived.ts
+++ b/src/frontend/src/lib/derived/network-tokens.derived.ts
@@ -10,7 +10,7 @@ import { derived, type Readable } from 'svelte/store';
 /**
  * All user-enabled tokens matching the selected network or chain fusion.
  */
-export const enabledNetworkTokens: Readable<Token[]> = derived(
+const enabledNetworkTokens: Readable<Token[]> = derived(
 	[enabledTokens, selectedNetwork],
 	filterTokensForSelectedNetwork
 );
--- a/src/frontend/src/lib/derived/network-tokens.derived.ts
+++ b/src/frontend/src/lib/derived/network-tokens.derived.ts
@@ -1,5 +1,5 @@
 import { exchanges } from '$lib/derived/exchange.derived';
-import { pseudoNetworkChainFusion, selectedNetwork } from '$lib/derived/network.derived';
+import { selectedNetwork } from '$lib/derived/network.derived';
 import { enabledTokens, tokensToPin } from '$lib/derived/tokens.derived';
 import { balancesStore } from '$lib/stores/balances.store';
 import type { Token, TokenUi } from '$lib/types/token';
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -10,8 +10,8 @@ import { derived, type Readable } from 'svelte/store';
 /**
  * All user-enabled tokens matching the selected network or chain fusion.
  */
-const enabledNetworkTokens: Readable<Token[]> = derived(
-	[enabledTokens, selectedNetwork, pseudoNetworkChainFusion],
+export const enabledNetworkTokens: Readable<Token[]> = derived(
```
peterpeterparker, line 13: Not sure why we export this dervied store?

src/frontend/src/lib/utils/network.utils.ts
```diff
@@ -45,18 +45,17 @@ export const mapNetworkIdToBitcoinNetwork = (networkId: NetworkId): BitcoinNetwo
...
+	Network | undefined
+]): T[] =>
 	$tokens.filter((token) => {
 		const {
 			network: { id: networkId, env }
 		} = token;
 
 		return (
-			($pseudoNetworkChainFusion && !isTokenIcrcTestnet(token) && env !== 'testnet') ||
+			(isNullish($selectedNetwork) && !isTokenIcrcTestnet(token) && env !== 'testnet') ||
```
peterpeterparker, line 58: Not sure about the change. `pseudoNetworkChainFusion` is a bit more readable and adding `isNullish($selectedNetwork)` here kind of duplicates the logic.

Do you plan to remove `$pseudoNetworkChainFusion` in a future PR?

src/frontend/src/lib/derived/network-tokens.derived.ts
AntonioVentilii: it slipped in, sorry, removed

src/frontend/src/lib/utils/network.utils.ts
peterpeterparker: The `pseudoNetworkChainFusion` was indeed a pseudo-network selector but also included the idea that it did not contained any test network. Withouht strong types and lots of tests I think it's a risk to mix test and production tokens. Therefore, I would not rename it to `noSelectedNetwork` if this makes sense? I would even not touch it until tests are properly setup.

AntonioVentilii:
Superseded by:

- https://github.com/dfinity/oisy-wallet/pull/5035

PR #3345, opened 2024-11-05
test(e2e): Add tests for transactions pages

AntonioVentilii:
# Motivation

Duplicate of #3230 , since I changed my account

PR #3347, opened 2024-11-05
refactor(frontend): Simplify props of Hero

AntonioVentilii:
# Motivation

The props given to `Hero` are somehow a bit confused since derived from routing. They were ok before, since the `Hero` was used in all routes. With the latest changes, the scope of the `Hero` was reduced to only the main page (tokens) and the transaction page.

To clean-up a bit we refactor the Hero to be a bit more obvious. In another PR, we will probably merge `Hero` and `HeroContent`.

# Changes

- Use navigations utils directly in `HeroContent` and `+layout`
- Remove deprecated props from Hero.


src/frontend/src/tests/lib/utils/nav.utils.spec.ts
```diff
@@ -0,0 +1,88 @@
+import { AppPath, ROUTE_ID_GROUP_APP } from '$lib/constants/routes.constants';
```
peterpeterparker, line 1: This shouldn't be part of another PR?

src/frontend/src/routes/(app)/+layout.svelte
```diff
@@ -15,18 +15,16 @@
...
-				? 'transactions'
-				: 'tokens';
+	let tokensRoute: boolean;
+	$: tokensRoute = isRouteTokens($page);
+
+	let transactionsRoute: boolean;
+	$: transactionsRoute = isRouteTransactions($page);
+
+	let showHero: boolean;
+	$: showHero = tokensRoute || transactionsRoute;
```
peterpeterparker, line 27: Out of curiosity, not a call to action: we already have E2E screenshots of every pages? That way we can assert that indeed the hero is displayed only when needed.

AntonioVentilii: I agree! But we don't have for all yet, I was working on it. I have the first PR for the transactions page already in draft:

- #3345

src/frontend/src/tests/lib/utils/nav.utils.spec.ts
AntonioVentilii: Ah yes sure:

- PR #3358

peterpeterparker: Not just the test, you mean the utils as well?

AntonioVentilii: correct, this PR is just for the `Hero` logic

peterpeterparker (approved):
LGTM, thx

PR #3352, opened 2024-11-05
feat(frontend): New Activity page

AntonioVentilii:
# Note

Duplicate of #3316 , because I changed account.

# Motivation

We want to create a new page with the list of the user activities, mainly the unified transactions. As first step we create the route similar to the the main page route.

In the layout we set a `MultipleListener` for the entire list of tokens enabled for the current network, so that the transactions are updated. This reflects what is being done with the transaction page specific to a token, where there is a single `Listener`.

Screenshot 2024-11-04 at 20 01 53


src/frontend/src/lib/derived/network-tokens.derived.ts
```diff
@@ -10,7 +10,7 @@ import { derived, type Readable } from 'svelte/store';
 /**
  * All user-enabled tokens matching the selected network or chain fusion.
  */
-const enabledNetworkTokens: Readable<Token[]> = derived(
+export const enabledNetworkTokens: Readable<Token[]> = derived(
```
peterpeterparker, line 13: Since we expose it, do you want to create a test for it? In another PR.

peterpeterparker (commented):
Isn't a `+page.ts` that reset the token and network to null missing? Sorry lost track of what we discussed.

```
export const load: PageLoad = (_$event: LoadEvent): RouteParams => ({
	token: null,
	network: null,
	uri: null
});
```

src/frontend/src/lib/derived/network-tokens.derived.ts
AntonioVentilii: so, I wanted to do it in a separate PR, because it requires to mock `page` store and I found out it is not that easy

or maybe you already encountered something similar? any suggestion?

AntonioVentilii:
> Isn't a +page.ts that reset the token and network to null missing?

yes, sorry, i missed to push a commit from the old one....

src/frontend/src/lib/derived/network-tokens.derived.ts
peterpeterparker: I think we are on the same page, I also wrote "In another PR" 😉.
Coolio.

src/frontend/src/routes/(app)/activity/+page.ts
```diff
@@ -0,0 +1,10 @@
+import { type RouteParams } from '$lib/utils/nav.utils';
+import type { LoadEvent } from '@sveltejs/kit';
+import type { PageLoad } from './$types';
+
+// We reset the data because the "activity" route operates without a network or token selected.
+export const load: PageLoad = (_$event: LoadEvent): RouteParams => ({
```
peterpeterparker, line 6: Not fore this PR but, in another one, if you can refactor this to extract a function it would be nice. We have the same loader in three different routes now.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/derived/network-tokens.derived.ts
AntonioVentilii: ah! i didn't read it, sorry eheh

src/frontend/src/routes/(app)/activity/+page.ts
AntonioVentilii: > P.S. I will extract the function in another PR

I wrote the same [above](https://github.com/dfinity/oisy-wallet/pull/3352#issuecomment-2457394871) ahahaha!

AntonioVentilii:
[OISY-130]

src/frontend/src/routes/(app)/activity/+page.ts
AntonioVentilii: Done:

- PR #3362

PR #3353, opened 2024-11-05
feat(frontend): add balance to icrc-ledger.api

peterpeterparker:
# Motivation

We aim to support tokens without an index canister. This means that we want to support tokens without transactions, but we still want to fetch balances. These balances can be queried directly from the ledger.

# Changes

- Add `balance` to `icrc-ledger.api`
- Comment and add todo for missing test to existing functions


AntonioVentilii (approved):
LGTM! Tks

peterpeterparker:

PR #3355, opened 2024-11-05
test(frontend): mockIdentity transformRequest

peterpeterparker:
# Motivation

Spotted this check in NNS dapp which prevents the `mockIdentity` to be used without knowing. Thought it can be interesting for future proof to have it in OISY as well.


peterpeterparker:
> Another good one to put in ic-js eheh

Yeah, generally speaking, I also feel there's no documentation on how to mock agent-js for Vitest or Jest. Plus, there isn't a library available that would allow developers to avoid reinventing the wheel in every project.

peterpeterparker:

PR #3334, opened 2024-11-05
feat(frontend): optional index canister

peterpeterparker:
# Motivation

Make the index canister ID optional in the core type definition without changing the way we load information - i.e. in a backwards compatible manner.

# Notes

This PR is relatively large, so Antonio and I reviewed the code together offline.

# Changes

- `custom-token.services.ts`: extend `toCustomToken` to support optional Index canister ID
- `IcTransactions.svelte`: silently ignore listing transactions if no index
- `ic-token.schema.ts`: set Index Canister as optional in the base type `IcCanisters`
- `ic-add-custom-tokens.service.ts`: extend support to, in the future, to allow user to register custom token without index canister id
- `icrc-wallet.worker.ts`: assert index is set. in the future this should not be called without index
- add few TODOs
- adapt tests

PR #3357, opened 2024-11-05
feat(frontend): isIcToken and isIcTokenCanistersStrict utils

peterpeterparker:
# Motivation

We will need to know if IC tokens are provided with or without Index Canister ID. Unrelated but, somehow related, we also want to know if a token is just a token or an IC token.

# Notes

Some test uncommented and will be activated in PR #3334 because they cannot be active unless the Index Canister ID is set to optional.

# Changes

- New utils `isIcToken`, `isNotIcToken`, `isIcTokenCanistersStrict` and `isNotIcTokenCanistersStrict`


AntonioVentilii (approved):
LGTM tks!

I have a generic question: shall this checking function be in `validation` or in `utils`? Asking because I am doing a similar one for `TokenSchema`

peterpeterparker:
As discussed offline: we ultimally agreed those were "validation" and not "utils". So we proceeded with some refactoring and I finally renamed the code of this PR from "utils" to "validation".

peterpeterparker:

PR #3359, opened 2024-11-05
refactor(frontend): ic-token.validation -> schema

peterpeterparker:
# Motivation

As discussed offline in the context of #3357. We actually adopt a new convention in which Zod schema related are provided in sub-folder `schema`.

This prevents the introduction of circular dependencies when the validation uses both types and Zod schema.

# Changes

- Rename `ic-token.validation` to `ic-token.schema`


AntonioVentilii (approved):
tks!

peterpeterparker:

PR #3360, opened 2024-11-05
refactor(frontend): token.validation -> schema

peterpeterparker:
# Motivation

Similar to #3359. We are introducing a new convention for schema.

# Changes

- Split `token.validation` into `token.schema` and validation. The schema, well, go to schema and in validation we keep the "utils".


AntonioVentilii (approved):
Thank you for the quick fix!

P.S. have a look, I think there is an additional row in the some of the imports, don't know if it is on purpose

peterpeterparker:

PR #3363, opened 2024-11-05
feat(frontend): check if exchange enabled for initialization

AntonioVentilii:
# Motivation

The derived store `exchangeInitialized` should be considered `true` when the exchange is disabled. This mainly happens for test environments (like E2E) where we don't provide API keys and we forcibly turn off the exchanges via env variable `VITE_EXCHANGE_DISABLED`: it is mostly used for loading states, so just for visual changes.

# Changes

- Extract parsed env variable `EXCHANGE_DISABLED` and export it globally.
- Include it in the logic of `exchangeInitialized`: it is true if either `exchangeInitialized` data is non-nullish or if exchange is disabled.

# Tests

Created some tests for the derived.


src/frontend/src/tests/lib/derived/exchange.derived.spec.ts
```diff
@@ -0,0 +1,31 @@
+import * as exchangeEnv from '$env/exchange.env';
+import { exchangeInitialized } from '$lib/derived/exchange.derived';
+import { exchangeStore } from '$lib/stores/exchange.store';
+import { get } from 'svelte/store';
+
+describe('exchange.derived', () => {
```
peterpeterparker, line 6: Maybe add a test with `EXCHANGE_DISABLED` explicitely set to `undefined` and `false`, just in case?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/tests/lib/derived/exchange.derived.spec.ts
AntonioVentilii: done!

peterpeterparker (approved):
LGTm, thx

PR #3366, opened 2024-11-06
test(frontend): ic-add-custom-tokens.service

peterpeterparker:
# Motivation

On one hand, it is worthwhile to have tests for the `loadAndAssertAddCustomToken` service. On the other hand, we're about to make the Index canister ID optional, which will require modifications to the service in #3334. Therefore, it makes sense to track and verify these changes with tests.


AntonioVentilii (approved):
LGTM tks!

peterpeterparker:

PR #3368, opened 2024-11-06
docs(frontend): add curve finance, uppercase aave

loki344:
# Motivation

Add Curve Finance to dapp explorer

# Changes

- Add description
- Add assets
- fix capitalization for aave

# Tests

image
image


src/frontend/src/env/dapp-descriptions.json
```diff
@@ -138,7 +138,7 @@
 	},
 	{
 		"id": "aave",
-		"name": "aave",
+		"name": "Aave",
```
AntonioVentilii, line 141: even if it is a small change, it is always better to separate in another PR, since it is another topic

loki344: No we cannot capitalize. See zkPoker which is displayed with small letter first.

peterpeterparker: Gotcha! Thanks for the explanation

PR #3373, opened 2024-11-06
test(frontend): remove mockedAgent

peterpeterparker:
# Motivation

A bit opiniated but, I feel like it's cleaner to not have the `mock<HttpAgent>()` provided as a global mock module. Feel like it's particular to every test suite. Hence, removing the existing mock use in two tests.


AntonioVentilii (approved):
Makes sense, if we risk involuntary conflicts in tests

peterpeterparker:

PR #3374, opened 2024-11-06
style(frontend): fix-hero-balance-opacity

BonomoAlessandro:
# Motivation

Zero balances in the `Hero` element are displayed with `opacity-50` but zero balances should also be displayed like non zero balances

# Changes

- changes opacity

# Tests

**before:**

zero balance:
image

non zero balance:
image

**after:**

zero balance:
image

non zero balance:
image


AntonioVentilii (approved):
Tks!

PR #3375, opened 2024-11-06
test(frontend): Test derived enabledEthereumTokens

AntonioVentilii:
Some tests for derived `enabledEthereumTokens`.

### DIFF
--- a/src/frontend/src/tests/eth/derived/tokens.derived.spec.ts
+++ b/src/frontend/src/tests/eth/derived/tokens.derived.spec.ts
@@ -1,0 +1,37 @@
+import * as ethEnv from '$env/networks.eth.env';
+import { ETHEREUM_TOKEN, SEPOLIA_TOKEN } from '$env/tokens.env';
+import { enabledEthereumTokens } from '$eth/derived/tokens.derived';
+import * as appContants from '$lib/constants/app.constants';
+import { testnetsStore } from '$lib/stores/settings.store';
+import { get } from 'svelte/store';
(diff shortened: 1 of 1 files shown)


src/frontend/src/tests/eth/derived/tokens.derived.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+describe('tokens.derived', () => {
+	describe('enabledEthereumTokens', () => {
+		beforeEach(() => {
+			vi.resetAllMocks();
+
+			testnetsStore.reset({ key: 'testnets' });
+
+			vi.spyOn(ethEnv, 'ETH_MAINNET_ENABLED', 'get').mockImplementation(() => true);
+
+			vi.spyOn(appContants, 'LOCAL', 'get').mockImplementation(() => false);
```
peterpeterparker, line 18: In two other tests we mock local as following:

```
vi.mock(import('$lib/constants/app.constants'), async (importOriginal) => {
	const actual = await importOriginal();
	return {
		...actual,
		LOCAL: false
	};
});
```

Not for this PR but, maybe we should find a way to have a utils or something?

peterpeterparker (approved):
Thanks!

src/frontend/src/tests/eth/derived/tokens.derived.spec.ts
peterpeterparker: Ah it's not applicable here, nevermind my comment. All good!

PR #3377, opened 2024-11-06
test(frontend): token.services

peterpeterparker:
# Motivation

To make the index canister optional in #3334 we need to test the `custom-token.services` which itself uses the `token.services`. That's why we need test for that services. Of course, we also need test generally speaking.


src/frontend/src/tests/lib/services/token.services.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+		});
+
+		it('should handle asynchronous callback', () =>
+			new Promise<void>((done) => {
+				// eslint-disable-next-line require-await
+				const callback = async () => {
+					done();
+				};
+
+				loadTokenAndRun({ token: mockValidToken, callback });
```
AntonioVentilii, line 43: i am not really sure here, how is this test "tested"?

src/frontend/src/tests/lib/services/token.services.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+		const mockErrorMessage = 'Error loading token';
+
+		let params: AutoLoadTokenParams<SaveUserToken, Erc20Token>;
+
+		let spyToastsError: MockInstance;
+		let spyBusyStart: MockInstance;
+		let spyBusyStop: MockInstance;
+
+		// we mock console.error just to avoid unnecessary logs while running the tests
+		vi.spyOn(console, 'error').mockImplementation(() => undefined);
```
AntonioVentilii, line 74: Just as note: I had to do the same in another test, maybe it is worth to create some sort of `disableConsole` util for tests

src/frontend/src/tests/lib/services/token.services.spec.ts
peterpeterparker: With `done`. If it is not called, the test fails. It assert the promise of the callback is correctly handled.

src/frontend/src/tests/lib/services/token.services.spec.ts
peterpeterparker: Yes, good idea!

src/frontend/src/tests/lib/services/token.services.spec.ts
AntonioVentilii: ah! right!!! thank you for the explanation

AntonioVentilii (approved):
LGTM tks!

peterpeterparker:

src/frontend/src/tests/lib/services/token.services.spec.ts
AntonioVentilii: Done with:

- PR https://github.com/dfinity/oisy-wallet/pull/5237

PR #3380, opened 2024-11-06
test(frontend): derived store enabledNetworkTokens

AntonioVentilii:
# Motivation

A few tests for derived store `enabledNetworkTokens`.

To help us mock the `page` store of svelte, we create a smaller util to mock the route params.


src/frontend/src/tests/mocks/page.store.mock.ts
```diff
@@ -1,13 +1,30 @@
+import { NETWORK_PARAM, TOKEN_PARAM, URI_PARAM } from '$lib/constants/routes.constants';
+import type { Page } from '@sveltejs/kit';
 import { writable } from 'svelte/store';
 
+export const mockPageData = ({
```
peterpeterparker, line 5: In NNS dapp we replace the page with a store for test purpose that exposes a `mock` function. That way in the test you can call `page.mock`.

Would that be something interesting?

https://github.com/dfinity/nns-dapp/blob/95d58213187190f606cffe7b43e6fc857d50528d/frontend/__mocks__/%24app/stores.ts#L40

AntonioVentilii: Yes! Good suggestion! Done:

- PR #3415

AntonioVentilii:
@peterpeterparker please re-review, I included the new mock page store

src/frontend/src/tests/lib/derived/network-tokens.derived.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+					network: ETHEREUM_NETWORK,
+					tokens: [ETHEREUM_TOKEN, mockEr20UserToken]
+				},
+				{
+					network: SEPOLIA_NETWORK,
+					tokens: [SEPOLIA_TOKEN]
+				}
+			];
+
+			networkMap.forEach(({ network, tokens }) => {
```
peterpeterparker, line 85: instead of looping with the test, what do you think of using `it.each` or `describe.each` here?

AntonioVentilii: Makes sense! Corrected

peterpeterparker (approved):
Nice, thanks!

PR #3385, opened 2024-11-06
style(frontend): fix cktokens hide option

BonomoAlessandro:
# Motivation

Until now, when the `Manage list` was opened, there were tokens (ckTokens) that were enabled but couldn't be switched off. However, it was still possible to disable them via the `Hero` element, which should not be possible.

# Changes

- uses same logic for hero element as for toggle items

# Tests

**before:**
Manage list: (like it is desired)
image

Hero: (should not be possible to hide this token)
image

**after:**
Manage list:
image

Hero: (it's not possible anymore to hide this token, which is good)
image


src/frontend/src/icp/types/icrc-custom-token.ts
```diff
@@ -13,3 +15,11 @@ export type IcrcCustomTokenWithoutId = UserTokenState & IcTokenWithoutIdExtended
 export type IcrcCustomToken = TokenToggleable<IcToken> & IcrcCustomTokenExtra;
 
 export type OptionIcrcCustomToken = Option<IcrcCustomToken>;
+
+export function isIcrcCustomTokenDisabled(token: IcrcCustomToken): boolean {
```
AntonioVentilii, line 19: same for this function, that possibly is a mix between zod and an util (plus tests)

src/frontend/src/lib/derived/token.derived.ts
```diff
@@ -29,7 +34,20 @@ export const tokenDecimals: Readable<number | undefined> = derived(
 	([$token]) => $token?.decimals
 );
 
-export const tokenToggleable: Readable<boolean> = derived(
-	[token],
-	([$token]) => nonNullish($token) && 'enabled' in $token
-);
+export const tokenToggleable: Readable<boolean> = derived([token], ([$token]) => {
```
AntonioVentilii, line 37: since we are here, some tests for this derived?

src/frontend/src/lib/derived/token.derived.ts
```diff
@@ -29,7 +34,20 @@ export const tokenDecimals: Readable<number | undefined> = derived(
...
+	if (nonNullish($token)) {
+		let toggleable = false;
```
AntonioVentilii, line 39: OISY codebase has preference to a functional approach

src/frontend/src/lib/derived/token.derived.ts
```diff
@@ -29,7 +34,20 @@ export const tokenDecimals: Readable<number | undefined> = derived(
...
+
+		if (icTokenIcrcCustomToken($token)) {
+			toggleable = !isIcrcCustomTokenDisabled($token);
+		} else if (icTokenEthereumUserToken($token)) {
+			toggleable = !isEthereumUserTokenDisabled($token);
```
AntonioVentilii, line 44: just noticing, isn't it better to create the functions that check if it is enabled? since you are negating it in all usage (or not?)

src/frontend/src/lib/derived/token.derived.ts
BonomoAlessandro: I will write some tests for this function in a separate PR

BonomoAlessandro:
@AntonioVentilii-DFINITY as discussed i moved the functions from above to the utils and wrote some tests. 
The tests for the token.derived.ts will come in a separate PR.

src/frontend/src/lib/derived/token.derived.ts
```diff
@@ -29,7 +35,15 @@ export const tokenDecimals: Readable<number | undefined> = derived(
...
+		if (icTokenIcrcCustomToken($token)) {
+			return isIcrcCustomTokenEnabled($token);
+		}
+		if (icTokenEthereumUserToken($token)) {
+			return isEthereumUserTokenEnabled($token);
+		}
+	}
+
+	return false;
+});
```
AntonioVentilii, line 49: is it possible to refactor this to avoid if statements?

`nonNullissh ? icTokenIcrcCustomToken ? ... isIcrcCustomTokenEnabled `

src/frontend/src/tests/lib/utils/token-toggle.utils.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+import type { IcrcCustomToken } from '$icp/types/icrc-custom-token';
+import {
+	isEthereumUserTokenDisabled,
+	isIcrcCustomTokenDisabled
+} from '$lib/utils/token-toggle.utils';
+import { describe, expect } from 'vitest';
+
+describe('isEthereumUserTokenDisabled', () => {
+	it('should check if default ethereum user token is disabled for token toggle', () => {
+		const token: EthereumUserToken = {
```
AntonioVentilii, line 15: here you can use directly `ETHEREUM_TOKEN` i think, and expand it if necessary

src/frontend/src/tests/lib/utils/token-toggle.utils.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+			standard: 'ethereum',
+			enabled: false,
+			category: 'default'
+		};
+
+		expect(isEthereumUserTokenDisabled(token)).toBe(true);
+	});
+
+	it('should check if custom ethereum user token is disabled for token toggle', () => {
+		const token: EthereumUserToken = {
```
AntonioVentilii, line 30: ditto

src/frontend/src/tests/lib/utils/token-toggle.utils.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+		};
+
+		expect(isEthereumUserTokenDisabled(token)).toBe(false);
+	});
+});
+
+describe('isIcrcCustomTokenDisabled', () => {
+	it('should check if icrc custom token is disabled for token toggle', () => {
+		const token: IcrcCustomToken = {
+			id: ICP_TOKEN_ID,
```
AntonioVentilii, line 48: here same for `ICP_TOKEN`

src/frontend/src/tests/lib/utils/token-toggle.utils.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+			enabled: false,
+			fee: ICP_TRANSACTION_FEE_E8S
+		};
+
+		expect(isIcrcCustomTokenDisabled(token)).toBe(true);
+	});
+
+	it('should check if custom icrc custom token is disabled for token toggle', () => {
+		const token: IcrcCustomToken = {
+			id: ICP_TOKEN_ID,
```
AntonioVentilii, line 68: ditto

src/frontend/src/tests/lib/utils/token-toggle.utils.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+			enabled: false,
+			fee: ICP_TRANSACTION_FEE_E8S
+		};
+
+		expect(isIcrcCustomTokenDisabled(token)).toBe(false);
+	});
+
+	it('should check if outdated custom icrc custom token is disabled for token toggle', () => {
+		const token: IcrcCustomToken = {
+			id: ICP_TOKEN_ID,
```
AntonioVentilii, line 88: ditto

src/frontend/src/tests/lib/utils/token-toggle.utils.spec.ts
AntonioVentilii, line 1: can you pleas provide some tests where you pass other tokens apart from ETH and ICP? Some ERC20 and some ICRC  maybe

src/frontend/src/lib/utils/token-toggle.utils.ts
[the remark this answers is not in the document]
BonomoAlessandro: hmm you are right. I think your suggestion with `ethereumTokenToggleDisabled` is pretty good.

src/frontend/src/tests/lib/utils/token-toggle.utils.spec.ts
BonomoAlessandro: @AntonioVentilii I don't understand why this would make sense on this level. I mean `isEthereumTokenToggleDisabled` for example is a function that takes a param of type `EthereumUserToken`. 

I guess it makes more sense to check the function one layer above (`tokenToggleable` for which i will provide test cases in another PR).

AntonioVentilii: no, my point is: if you pass an ERC20 token or an ICRC token to the function, it should return `false`, no? shouldn't we test it, since we test purely the function?

BonomoAlessandro: @AntonioVentilii ahh i see. I have added now some tests for erc20 and icrc tokens.

PR #3392, opened 2024-11-07
test(frontend): more custom token

peterpeterparker:
# Motivation

We need few more tests before being able to complete #3334.


peterpeterparker:

PR #3396, opened 2024-11-07
ci(e2e): add devices to config

inc-man:
# Motivation

We want our e2e tests to run against multiple devices and browsers.

# Changes

- add multiple different devices and browsers


AntonioVentilii:
May you please run the workflow on this branch to check if it works?

PR #3397, opened 2024-11-07
refactor(frontend): use common component for TransactionStatusComponent

AntonioVentilii:
# Motivation

We want to try and create more common components for the Transaction components. One that is already ready to use is the TransactionStatusComponent. So we replace `TransactionPending` with `TransactionStatusComponent`.

The only difference between them is that the new one consider even the `unconfirmed` status (proper of BTC transactions) beyond the `pending` status.


peterpeterparker (approved):
Have you thought about using `TransactionStatusComponent` within `TransactionPending` ? That way you don't have to duplicate the boolean logic.

Just sharing in case interesting, approving already.

AntonioVentilii:
Ah! Didn't think about it sincerely. Let me check

AntonioVentilii:
[OISY-130]

PR #3398, opened 2024-11-07
refactor(frontend): change small props of Transaction component

AntonioVentilii:
# Motivation

To keep it consistent with the other components, a small change in naming and typing of the prop of `Transaction`:

- `value` becomes `amount`
- `timestamp` becomes a `number` instead of a `bigint`


peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3399, opened 2024-11-07
refactor(frontend): rename all ETH transactions components

AntonioVentilii:
# Motivation

All the transactions components and utils related to Ethereum have no prefix (for example `BtcTransaction`, or `IcTransaction`). It becomes a bit confusing, especially now that we are trying to homologate all the transactions in one page.

So, as first step we rename all the transaction components to have the `Eth` prefix.


peterpeterparker (approved):

AntonioVentilii:
[OISY-130]

PR #3401, opened 2024-11-07
feat(frontend): use single component for BTC and ETH transactions

AntonioVentilii:
# Motivation

The component `Transaction` could attend both needs of `BtcTransaction` and `EthTransaction` with a few tweaks. Mainly, expanding the type of accepted transactions types and adjust the icon and the label.

# Changes

- Increase the logic of `icon` selection in `Transaction` to include other types (for example conversions).
- Change `label` to be a `slot` instead of an internal variable.
- Adapt component `BtcTransaction`: this means only providing the `label` as `slot`.
- Replace the `Card` in `EthTransaction` with `Transaction`.

# Tests

No changes in local replica.


peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3403, opened 2024-11-07
refactor(frontend): rename prop of IcTransaction

AntonioVentilii:
# Motivation

We rename the prop `transactionType` to `type`, for consistency.

Furthermore, elsewhere in the code, in the util `mapIcrcTransaction`, the type is defined as

```typescript
const type: IcTransactionType = nonNullish(fromNullable(approve))
  ? 'approve'
  : nonNullish(fromNullable(burn))
    ? 'burn'
    : isMint
      ? 'mint'
      : source.incoming === false
        ? 'send'
        : 'receive';
```

Meaning that we can simply use `type === 'send'` in the same way that we use `incoming === false`

### DIFF
--- a/src/frontend/src/icp/components/transactions/IcTransaction.svelte
+++ b/src/frontend/src/icp/components/transactions/IcTransaction.svelte
@@ -25,13 +25,7 @@
 	let timestamp: bigint | undefined;
 	let incoming: boolean | undefined;
 
-	$: ({
-		type,
-		typeLabel: transactionTypeLabel,
-		value,
-		timestamp,
-		incoming
-	} = transaction);
+	$: ({ type, typeLabel: transactionTypeLabel, value, timestamp, incoming } = transaction);
 
 	let pending = false;
 	$: pending = transaction?.status === 'pending';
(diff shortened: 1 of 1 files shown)


src/frontend/src/icp/components/transactions/IcTransaction.svelte
```diff
@@ -19,19 +19,13 @@
 
 	export let transaction: IcTransactionUi;
 
-	let transactionType: IcTransactionType;
+	let type: IcTransactionType;
```
peterpeterparker, line 22: isn't `type` some sort of reserved TypeScript keyword? I would bet that's why `transactionType` was used.

AntonioVentilii: i thought so, but we use type in ETH and BTC too

```typescript
let type: TransactionType;

let type: EthTransactionType;
```

so, maybe I shall rename those and leave this one as it is?

peterpeterparker: Sounds good then!

peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3412, opened 2024-11-07
chore(frontend): remove unnecessary guard from Loaders

AntonioVentilii:
# Motivation

The `Loaders` component is now inside the `AuthGuard`, so it does not need a guard itself.

We make the layout use directly `Loaders` instead of `LoadersGuard`, and remove deprecated components.


peterpeterparker (approved):
LGTM, thx

PR #3413, opened 2024-11-07
test(frontend): exclude folders from coverage analysis

AntonioVentilii:
# Motivation

We exclude some folders from the coverage analysis, since there is nothing we can do to improve those (those are not project folders).

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -62,7 +62,10 @@ export default defineConfig(
 			watch: false,
 			silent: false,
 			setupFiles: ['./vitest.setup.ts'],
-			include: ['./src/**/*.{test,spec}.?(c|m)[jt]s?(x)']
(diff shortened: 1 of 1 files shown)


vitest.config.ts
```diff
@@ -62,7 +62,10 @@ export default defineConfig(
 			watch: false,
 			silent: false,
 			setupFiles: ['./vitest.setup.ts'],
-			include: ['./src/**/*.{test,spec}.?(c|m)[jt]s?(x)']
+			include: ['./src/**/*.{test,spec}.?(c|m)[jt]s?(x)'],
+			coverage: {
+				exclude: ['build', '.dfx', '**/.svelte-kit']
```
peterpeterparker, line 67: It works now? I wonder what I mistyped when we tried together the other day.

peterpeterparker (approved):
LGTM, thx

PR #3418, opened 2024-11-08
test(e2e): create test id for navigation items

AntonioVentilii:
# Motivation

We want to start testing each page available in OISY, so as first step, we set some test IDs for the navigation items in the main menu.


DenysKarmazynDFINITY (approved):
Cool, thanks

PR #3422, opened 2024-11-11

sa-github-api:


peterpeterparker:
The ICGhost's Index canister seems to have been out of cycles when the batch ran. The issue seems to have been fixed in the meantime.

PR #3432, opened 2024-11-11
fix(frontend): global GIX styles

DenysKarmazynDFINITY:
# Motivation

Since most of the GIX UI components use the same CSS classes namings (.content, .header, .container, .etc), the styles, that were supposed to be applied to the Modal component, affect all other GIX components inside a modal (e.g. in case token conversion flow I use `Collapsible`). This PR fixes it by applying the global styles from gix.scss only to direct children of the specified selectors.


AntonioVentilii (approved):
Tks

PR #3439, opened 2024-11-11
style(frontend): add 1px height to DappTag section in DappCard

loki344:
# Motivation

On MacOS + Chrome, there is a slight issue with not visible borders in the DappCard (Looked fine on Windows..)

# Changes

- add 1px height

# Tests

before: 

image

after
image

### DIFF
--- a/src/frontend/src/lib/components/dapps/DappCard.svelte
+++ b/src/frontend/src/lib/components/dapps/DappCard.svelte
@@ -30,7 +30,7 @@
 				{oneLiner}
 			</p>
 		</section>
-		<section class="absolute bottom-4 left-4 right-4 max-h-6 min-h-6 overflow-hidden md:max-h-14">
+		<section class="absolute bottom-4 left-4 right-4 max-h-6 min-h-6 overflow-hidden md:max-h-[calc(3.5rem+1px)]">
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/dapps/DappCard.svelte
```diff
@@ -30,7 +30,9 @@
 				{oneLiner}
 			</p>
 		</section>
-		<section class="absolute bottom-4 left-4 right-4 max-h-6 min-h-6 overflow-hidden md:max-h-14">
+		<section
+			class="absolute bottom-4 left-4 right-4 max-h-6 min-h-6 overflow-hidden md:max-h-[calc(3.5rem+1px)]"
```
peterpeterparker, line 34: > there is a slight issue with not visible borders

Instead of adding 1px to the height, should we add a border-1 transparent given the description of the issue?

loki344: Hmm, I don't get this to work with the overflow hidden. Do you want to give it a try?

peterpeterparker: There you go: https://github.com/dfinity/oisy-wallet/pull/3446

PR #3440, opened 2024-11-11
feat(frontend): post-message icrc strict type

peterpeterparker:
# Motivation

In #3437 we will implement web workers that behaves differently if the index canister id is provided or not. That's is why we two types for the post message following the same pattern as with the `IcCanisters` type.

# Changes

- Introduces `PostMessageDataRequestIcrcStrictSchema`


peterpeterparker:

PR #3441, opened 2024-11-11
style(frontend): removes opacity definition of balance component

BonomoAlessandro:
# Motivation

Even if the token balance is 0 it should be displayed with the same opacity as a balance of 100 or something else.

# Changes

- removes opacity calculation

# Tests

before:
image

after:
image


src/frontend/src/lib/components/hero/Balance.svelte
```diff
@@ -13,7 +13,7 @@
 
 <span class="flex flex-col gap-2">
 	<output
-		class={`break-words ${(token?.balance?.toBigInt() ?? 0n) === 0n ? 'opacity-50' : 'opacity-100'} inline-flex w-full flex-row justify-center gap-3 text-4xl font-bold lg:text-5xl`}
+		class={`inline-flex w-full flex-row justify-center gap-3 break-words text-4xl font-bold lg:text-5xl`}
```
AntonioVentilii, line 16: since it is not required that is a variable anymore, can you please remove the brackets?

BonomoAlessandro:
@AntonioVentilii unused brackets are removed

AntonioVentilii (approved):

PR #3442, opened 2024-11-11
feat(frontend): Ethereum transaction loader for multiple tokens

AntonioVentilii:
# Motivation

We need to create a transaction loader of all Ethereum tokens for the unified transactions list. The implementation is very similar to the `LoaderBalances` component that loads the balance for all the Ethereum tokens.

# Changes

- Create new component using all enabled Ethereum tokens.
- Create tests.
- Use in `+page`.

# Tests

Tests provided.


src/frontend/src/eth/components/loaders/LoaderMultipleEthTransactions.svelte
```diff
@@ -0,0 +1,22 @@
+<script lang="ts">
+	import { debounce } from '@dfinity/utils';
+	import { enabledEthereumTokens } from '$eth/derived/tokens.derived';
+	import { loadTransactions } from '$eth/services/transactions.services';
+	import { enabledErc20Tokens } from '$lib/derived/tokens.derived';
+
+	const load = async () => {
+		await Promise.allSettled(
```
peterpeterparker, line 8: Just a debounce to prevent multiple load triggers feels a bit light. Did you test to ensure that it effectively prevents unnecessary loads?

AntonioVentilii: you are right! it was a very lax measure: so, I make it run only when the ERC20 tokens store is initialized. I included more tests to verify that it is not called more than once if the stores are not loaded.

peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3443, opened 2024-11-11
test(frontend): Missing test for component LoaderEthTransactions

AntonioVentilii:
# Motivation

We add an important test for component `LoaderEthTransactions`: when the load transactions function fails, the component should understand that it should re-load if necessary, instead of considering a token that was already processed.


peterpeterparker (approved):
Thx!

AntonioVentilii:
[OISY-130]

PR #3444, opened 2024-11-11
style(frontend): add neutral colors to theme

loki344:
# Motivation

Adds neutral colors.

# Changes

Colors according to figma.

# Note

Just declares colors for future features to use.


AntonioVentilii (commented):
Careful with this one, because the class `text-primary` and `text-secondary` were defined in another way:

```css
--color-primary-rgb: var(--color-blue-ribbon-rgb);
--color-primary: rgb(var(--color-primary-rgb));
--color-secondary-rgb: var(--color-black-rgb);
--color-secondary: rgb(var(--color-secondary-rgb));

.text-primary {
	color: var(--color-primary);
}

.text-secondary {
	color: var(--color-secondary);
}
```

i think, before this, we should replace `text-primary`

loki344:
It might be better to have a separate PR to address the secondary / primary text color mapping

AntonioVentilii (approved):
LGTM tks

PR #3446, opened 2024-11-11
style(frontend): fix dapp badge height

peterpeterparker:
# Motivation

Parts of the badges in the dapp card are hidden by the overflow. This happens because the badge, actualy `span`, is not exactly equals to the container `li`.

# Notes

Resolve similar issue as #3439

# Changes

- Set `li` to be `flex` as well.

# Screeshots

Before:

Capture d’écran 2024-11-11 à 16 26 19
Capture d’écran 2024-11-11 à 16 26 46

After:

Capture d’écran 2024-11-11 à 16 26 05


AntonioVentilii (approved):
:LGTM ts!

PR #3448, opened 2024-11-11
feat(frontend): include ID in Ethereum transactions UI

AntonioVentilii:
# Motivation

To normalize all transactions we first include an optional ID in the Ethereum transactions for UI.


AntonioVentilii:
[OISY-130]

PR #3451, opened 2024-11-11
feat(frontend): ConvertInputAmount component

DenysKarmazynDFINITY:
# Motivation

In this PR, `ConvertInputAmount` component is implemented. Note: unfortunately, I needed to "hack" some internal GIX Input styles in order to match the design from Figma. 

Empty state:
Screenshot 2024-11-11 at 18 34 41

Normal state:
Screenshot 2024-11-11 at 18 40 21

Disabled state:
Screenshot 2024-11-11 at 18 37 21

Error state:
Screenshot 2024-11-11 at 18 34 50

### DIFF
--- a/src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
+++ b/src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
@@ -5,6 +5,7 @@
 	import { fade } from 'svelte/transition';
 	import InputCurrency from '$lib/components/ui/InputCurrency.svelte';
 	import { i18n } from '$lib/stores/i18n.store';
+	import type { ConvertAmountErrorType } from '$lib/types/convert';
 	import type { Token } from '$lib/types/token';
 	import { invalidAmount } from '$lib/utils/input.utils';
 	import { parseToken } from '$lib/utils/parse.utils';
@@ -13,31 +14,37 @@
 	export let amount: number | undefined = undefined;
 	export let name = 'convert-amount';
 	export let disabled: boolean | undefined = undefined;
-	export let customValidate: (userAmount: BigNumber | undefined) => void = () => {};
-	export let error: Error | undefined = undefined;
-	export let placeholder = '0.00';
+	export let customValidate: (userAmount: BigNumber) => ConvertAmountErrorType = () => undefined;
+	export let errorType: ConvertAmountErrorType = undefined;
+	export let placeholder = '0';
 
 	const onReset = () => {
 		amount = undefined;
 	};
 
 	const validate = () => {
-		const parsedValue = !invalidAmount(amount)
-			? parseToken({
-					value: `${amount}`,
-					unitName: token.decimals
-				})
-			: undefined;
+		if (invalidAmount(amount)) {
+			errorType = undefined;
+			return;
+		}
 
-		customValidate(parsedValue);
+		const parsedValue = parseToken({
+			value: `${amount}`,
+			unitName: token.decimals
+		});
+
+		errorType = customValidate(parsedValue);
 	};
 
 	const debounceValidate = debounce(validate, 300);
 
+	let errorState: boolean;
+	$: errorState = nonNullish(errorType) && errorType === 'insufficient-balance';
+
 	$: amount, token.decimals, debounceValidate();
 </script>
 
-<div class="convert-input-amount text-3xl font-bold" class:error={nonNullish(error)}>
+<div class="convert-input-amount text-3xl font-bold" class:error={errorState}>
 	<InputCurrency
 		{name}
 		testId="convert-amount"
@@ -53,7 +60,7 @@
 					data-tid="convert-amount-reset"
 					aria-label={$i18n.convert.text.input_reset_button}
 					on:click|preventDefault={onReset}
-					class={nonNullish(error) ? 'text-error' : 'text-aurometalsaurus'}
+					class={errorState ? 'text-error' : 'text-aurometalsaurus'}
 				>
 					<IconClose />
 				</button>
@@ -62,29 +69,22 @@
 	</InputCurrency>
 </div>
 
-<style lang="postcss">
-	:global(.convert-input-amount) {
-		&.error {
-			& > .input-block > div.input-field {
-				& input[id],
-				& input[id]:focus {
-					border-color: var(--color-error-default);
-					color: var(--color-error-default);
-				}
-			}
-		}
+<style lang="scss">
+	:global(.convert-input-amount div.input-block) {
+		--padding: 0;
+	}
 
-		& > .input-block {
-			--padding: 0;
+	:global(.convert-input-amount.error div.input-field input[id]) {
+		border-color: var(--color-error-default);
+		color: var(--color-error-default);
+	}
 
-			& > div.input-field input[id] {
-				@apply h-14;
+	:global(.convert-input-amount div.input-field input[id]) {
+		@apply h-14;
 
-				&:disabled {
-					--disable: var(--color-bright-gray);
-					--input-background: var(--color-ghost-white);
-				}
-			}
+		&:disabled {
+			--disable: var(--color-bright-gray);
+			--input-background: var(--color-ghost-white);
 		}
 	}
 </style>
--- a/src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
+++ b/src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
@@ -39,7 +39,7 @@
 	const debounceValidate = debounce(validate, 300);
 
 	let errorState: boolean;
-	$: errorState = nonNullish(errorType) && errorType === 'insufficient-balance';
+	$: errorState = nonNullish(errorType) && errorType === 'insufficient-funds';
 
 	$: amount, token.decimals, debounceValidate();
 </script>
--- a/src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
+++ b/src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
@@ -69,7 +69,7 @@
 	</InputCurrency>
 </div>
 
-<style lang="scss">
+<style lang="postcss">
 	:global(.convert-input-amount div.input-block) {
 		--padding: 0;
 	}
--- a/src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
+++ b/src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
@@ -1,14 +1,17 @@
 <script lang="ts">
 	import { IconClose } from '@dfinity/gix-components';
-	import { debounce, isNullish, nonNullish } from '@dfinity/utils';
+	import { debounce, nonNullish } from '@dfinity/utils';
 	import { BigNumber } from '@ethersproject/bignumber';
+	import { fade } from 'svelte/transition';
 	import InputCurrency from '$lib/components/ui/InputCurrency.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
 	import type { Token } from '$lib/types/token';
 	import { invalidAmount } from '$lib/utils/input.utils';
 	import { parseToken } from '$lib/utils/parse.utils';
 
 	export let token: Token;
 	export let amount: number | undefined = undefined;
+	export let name = 'convert-amount';
 	export let disabled: boolean | undefined = undefined;
 	export let customValidate: (userAmount: BigNumber | undefined) => void = () => {};
 	export let error: Error | undefined = undefined;
@@ -36,7 +39,8 @@
 
 <div class="convert-input-amount text-3xl font-bold" class:error={nonNullish(error)}>
 	<InputCurrency
-		name="convert-amount"
+		{name}
+		testId="convert-amount"
 		bind:value={amount}
 		decimals={token.decimals}
 		{placeholder}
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
```diff
@@ -0,0 +1,84 @@
...
+					class:hidden={isNullish(amount) || disabled}
+				>
+					<IconClose />
+				</button>
+			{/if}
+		</svelte:fragment>
+	</InputCurrency>
+</div>
+
+<style lang="postcss">
```
peterpeterparker, line 59: `postcss`?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
```diff
@@ -0,0 +1,84 @@
...
+	:global(.convert-input-amount) {
```
peterpeterparker, line 60: is there a way without using global?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
```diff
@@ -0,0 +1,84 @@
...
+	};
+
+	const debounceValidate = debounce(validate, 300);
+
+	$: amount, token.decimals, debounceValidate();
+</script>
+
+<div class="convert-input-amount text-3xl font-bold" class:error={nonNullish(error)}>
+	<InputCurrency
+		name="convert-amount"
```
peterpeterparker, line 39: should this static value be a property?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
```diff
@@ -0,0 +1,84 @@
...
+		bind:value={amount}
+		decimals={token.decimals}
+		{placeholder}
+		{disabled}
+	>
+		<svelte:fragment slot="inner-end">
+			{#if nonNullish(amount) && !disabled}
+				<button
```
peterpeterparker, line 47: a `fade` effect?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
```diff
@@ -0,0 +1,84 @@
...
+					on:click|preventDefault={onReset}
+					class={nonNullish(error) ? 'text-error' : 'text-aurometalsaurus'}
+					class:hidden={isNullish(amount) || disabled}
```
peterpeterparker, line 50: You need `class:hidden`? it's not rendered if nullish or disabled

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
peterpeterparker, line 47: isn't the accessibility missing?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
```diff
@@ -0,0 +1,84 @@
...
+		customValidate(parsedValue);
+	};
+
+	const debounceValidate = debounce(validate, 300);
+
+	$: amount, token.decimals, debounceValidate();
+</script>
+
+<div class="convert-input-amount text-3xl font-bold" class:error={nonNullish(error)}>
+	<InputCurrency
```
peterpeterparker, line 38: no placeholder?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
```diff
@@ -0,0 +1,84 @@
...
+	import { BigNumber } from '@ethersproject/bignumber';
+	import InputCurrency from '$lib/components/ui/InputCurrency.svelte';
+	import type { Token } from '$lib/types/token';
+	import { invalidAmount } from '$lib/utils/input.utils';
+	import { parseToken } from '$lib/utils/parse.utils';
+
+	export let token: Token;
+	export let amount: number | undefined = undefined;
+	export let disabled: boolean | undefined = undefined;
+	export let customValidate: (userAmount: BigNumber | undefined) => void = () => {};
```
peterpeterparker, line 13: in SendInputAmount we declare it as following, therefore for consistency I would keep the same:

```
	export let customValidate: (userAmount: BigNumber) => Error | undefined = () => undefined;
```

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
```diff
@@ -0,0 +1,84 @@
...
+	export let error: Error | undefined = undefined;
```
peterpeterparker, line 14: This is maybe a new pattern? In which case will this be use?

peterpeterparker (commented):
Can you also provide some tests for the component?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: On the line 80, I'm applying a tailwind class (`@apply h-14`) to the input. Why tailwind and not CSS - just wanted to use the exact same height value as the "token-info" container. I think I saw another place in the code where "postcss" is used, but please let me know if you think there are some constraints here I'm not aware of.

Screenshot 2024-11-12 at 10 18 18

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: I did try everything I could come up with - unfortunately, without `global` the rules are considered as "local" and therefore are not applied to the children components.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: Good idea, done

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: I did try both `fade` and `slide`, seemed a little bit "too much" for this button. But I don't have a strong opinion here, so let's have `fade` here.

DenysKarmazynDFINITY: Just tried with some smaller `duration` value - looking better than with the default one.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: Sorry, leftover from the previous version. Not needed, therefore removed.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: True, added it.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: Placeholder is passed as a prop on the line 46.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: This is related to the above approach I described. Parent component passes it as `error={insufficientFundsError}`. This will set the input to error state only when provided amount is bigger than token balance. The input will stay in non-error state in case of `insufficientFundsForFee` is defined (the error message will be displayed above buttons).

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
peterpeterparker: It has to be `postcss`? It does not work as well with our standard `scss` annotation?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
peterpeterparker: Thanks for the explanation, makes sense. However, not fully convinced by the API.

- Should it be use a more meaningful return type than `void`? Like `{success: boolean}`?

- Shouldn't the default value be `undefined` that way the property becomes optional, similar to the other custom validation that I shared?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
peterpeterparker: We are mixing some imperative and declarative approach, plus we are sharing up and down states, feels like we might end up with some desynchronization at some point. Is there another pattern that would allow to scope those errors within the component or transmit the information one way only with a single prop?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
peterpeterparker: Ah cool.

Speaking of, `0.00` is a correct default place holder?! Two decimals per default, always?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
peterpeterparker: Not for this PR at all but, maybe afterwards, we can extract a global constant similar to `SLIDE_DURATION`. Like `FADE_SHORT_DURATION` or something.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
peterpeterparker: And there is no CSS4 variables already available for those? Maybe in another PR, we should then extend the underneath components with variables and then get rid of this global. That's prone to maintainance issue. We gonna need an E2E test with screenshot for it.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: You're right, it can be done with `scss`, I was confused by WebStorm - it doesn't recognise `apply` mixin when set to `lang="scss"`. Updated.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: For some rules, I used the variables to update the styling. But, for example, for `error` state I had to rewrite the whole property. And yeah, totally agree that we need to improve this part as well as have e2e screenshot for such a case.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: noted👍

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: So I reworked it a bit, and now it's pretty similar to the approach we have in "send". To me, looks much cleaner now. WDYT?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: I think it's now resolved with the changes to the approach I just pushed.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: This is how it looks from the parent side:
```
if (isNullish(totalFee)) {
    return;
}
...
if (userAmount.gt(parsedSendBalance)) {
	return 'insufficient-balance';
}

if (nonNullish(parsedTotalFee) && parsedTotalFee.add(userAmount).gt(parsedSendBalance)) {
	return 'insufficient-balance-for-fee';
}
```

and 

```
<ConvertInputAmount
		slot="amount"
		token={$sourceToken}
		bind:amount={sendAmount}
		{customValidate}
		disabled={isZeroBalance}
		bind:errorType
	/>
```

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
peterpeterparker: Ok. I was just asking myself if the number of decimals should equals the token decimals.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
peterpeterparker: - how is `errorType` used in the parent?

- what do you think of assigning the value to `errorType` in the parent and not in the `ConvertInputAmount`, would that be cleaner?

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
peterpeterparker: Let's continue in the other thread.

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: This is how the error vars look like in the parent component:
Screenshot 2024-11-12 at 14 58 01

I just tried moving `errorType` up to the parent, and to be honest the current way seems to be a bit more cleaner (mostly, because we would need to call `customValidate` when value is `undefined` to reset errors, as well as we would still need to pass somehow to `ConvertInputAmount` information about the error to display it with proper CSS classes).

DenysKarmazynDFINITY: P.S.: I updated the type string to use `funds` instead of `balance` for consistency.

peterpeterparker: Alright, thanks for trying and explanation. Let's do the following. We go on with the PR but, if you don't mind, after the first usage of this component and pattern - before we use it everywhere - you allow me to have just a quick look, just in case I suddently remember a pattern we would use in other dapps once I got the code in front of my eyes 😉.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/components/convert/ConvertInputAmount.svelte
DenysKarmazynDFINITY: Agreed 🙂

PR #3450, opened 2024-11-11
test(frontend): icp-wallet.worker

peterpeterparker:
# Motivation

Few tests to cover the `icp-wallet.worker` and its scheduler.

# Changes

- Provide tests

PR #3457, opened 2024-11-12
feat(frontend): ic wallet worker initialized

peterpeterparker:
# Motivation

While developing the tests of #3450 I noticed that we can spare one `postMessage` in the IC wallet worker.
i.e. we can spare sending an additional time the balance without transactions to the frontend if we were able to fetch transactions

# Tests

Covered by the tests in #3450.

# Changes

- Set `initalized` when there were some transactions detected as well.


AntonioVentilii (approved):
Good one! Tks

peterpeterparker:

PR #3449, opened 2024-11-11
feat(frontend): create initial component for unified transactions

AntonioVentilii:
# Motivation

As initial step to show the unified transactions, we create a basic component, that show a list of transactions with their specific card component (each network has a specific one), or a placeholder if no transaction is present.

We will implement each list separately in other PRs.

# Changes

- Create type for transaction with associated component.
- Create component AllTransactions.
- Use it in `Activity`.

# Tests

Screenshot 2024-11-11 at 21 26 25
Screenshot 2024-11-11 at 21 26 36
Screenshot 2024-11-11 at 21 26 42

PR #3458, opened 2024-11-12
feat(frontend): util to map unified transactions list

AntonioVentilii:
# Motivation

Following up PR #3449 , we implement the util to create the list of unified transactions, starting with BTC transactions.

# Changes

- Move the util from inside the `AllTransactionsList` component to a specific module.
- Use the existing transaction store for BTC and map their component.

# Tests

Created some additional tests for both the util and the component. Here are the result loading a BTC address that has a transaction. Please note that the UI will be modified later.

Screenshot 2024-11-12 at 15 24 56
Screenshot 2024-11-12 at 15 25 03


src/frontend/src/lib/utils/transactions.utils.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export const mapAllTransactionsUi = ({
+	tokens,
+	$btcTransactions
+}: {
+	tokens: Token[];
+	$btcTransactions: CertifiedStoreData<TransactionsData<BtcTransactionUi>>;
+}): AllTransactionsUi =>
+	tokens.reduce<AllTransactionsUi>((acc, { id: tokenId, network: { id: networkId } }) => {
+		if (isNetworkIdBTCMainnet(networkId) && nonNullish($btcTransactions)) {
```
peterpeterparker, line 28: Maybe remove the `nonNullish($btcTransactions)` and handle the case within the if? That way the loop won't proceed with the other ifs.

src/frontend/src/tests/lib/components/transactions/AllTransactionsList.spec.ts
```diff
@@ -1,28 +1,72 @@
...
-import { ethTransactionsStore } from '$eth/stores/eth-transactions.store';
+import BtcTransaction from '$btc/components/transactions/BtcTransaction.svelte';
 import AllTransactionsList from '$lib/components/transactions/AllTransactionsList.svelte';
+import { mapAllTransactionsUi } from '$lib/utils/transactions.utils';
+import { createMockBtcTransactionsUi } from '$tests/mocks/btc.mock';
 import en from '$tests/mocks/i18n.mock';
 import { render } from '@testing-library/svelte';
+import type { MockedFunction } from 'vitest';
+
+vi.mock('$lib/utils/transactions.utils', () => ({
```
peterpeterparker, line 9: why mocking the utils?

AntonioVentilii: I want to assume that the util function works correctly, and I just verify how the component behaves with different outputs of the util. I don't think it is necessary to test the function inside the component, since the component behavior changes only in two cases: either empty return or non-empty return, right?

src/frontend/src/lib/utils/transactions.utils.ts
peterpeterparker: yes

src/frontend/src/tests/lib/components/transactions/AllTransactionsList.spec.ts
peterpeterparker: Yeah but you also want to test the rendering so per extension you need the utils, no?

AntonioVentilii: yes yes, but i mock the util return, for example, I mock an empty return or a return with n transactions, to test the rendering in both cases; is that what you meant? or do you mean that it should render something else?

src/frontend/src/lib/utils/transactions.utils.ts
AntonioVentilii: done!

src/frontend/src/tests/lib/components/transactions/AllTransactionsList.spec.ts
peterpeterparker: I just meant: don't mock the utils 😁

If possible. If we introduce various level of mocks within the all code base, then the tests become less accurate and might actually miss a case between mocked layers. That's why when it's not too difficult, kind of feel like it's not too bad to not mock, if that makes sense.

AntonioVentilii: Ah understood now! Done! have a look please

AntonioVentilii:
[OISY-130]

PR #3461, opened 2024-11-12
test(frontend): configure vitest to use data-tid

AntonioVentilii:
# Motivation

Since we use `data-tid` as test ID, it is useful that vitest uses it for functions like `getByTestId`.


vitest.config.ts
```diff
@@ -10,6 +11,10 @@ process.env = {
 	...readCanisterIds({ prefix: 'VITE_' })
 };
 
+configure({
```
peterpeterparker: What about adding this to `vitest.setup` instead?

AntonioVentilii: indeed, it was my mistake, i changed it to setup

PR #3462, opened 2024-11-12
test(frontend): tests for skeleton cards

AntonioVentilii:
# Motivation

Small useful tests for Skeleton cards components.


src/frontend/src/lib/components/ui/Card.svelte
```diff
@@ -4,7 +4,7 @@
 	import CardAmountDescription from '$lib/components/ui/CardAmountDescription.svelte';
 
 	export let noMargin = false;
-	export let testId: string | undefined = undefined;
+	export let testId = 'card';
```
peterpeterparker, line 7: Why did those properties changed? Couldn't we keep the default undefined?

AntonioVentilii: No particular reason, i thought to use the same way that gix-components do for UI components

But changed to undefined

peterpeterparker (approved):
Thanks!

PR #3463, opened 2024-11-12
feat(frontend): findTwinToken util

DenysKarmazynDFINITY:
# Motivation

The goal is to create a util for finding a twin token that can be reused in token-grouping.util as well as in the upcoming `convert.store`.

### DIFF
--- a/src/frontend/src/lib/utils/token.utils.ts
+++ b/src/frontend/src/lib/utils/token.utils.ts
@@ -4,6 +4,8 @@ import {
 } from '$env/networks.icrc.env';
 import { ERC20_SUGGESTED_TOKENS } from '$env/tokens.erc20.env';
 import type { ContractAddressText } from '$eth/types/address';
+import type { IcCkToken } from '$icp/types/ic-token';
+import { isIcCkToken } from '$icp/utils/icrc.utils';
 import { ZERO } from '$lib/constants/app.constants';
 import type { BalancesData } from '$lib/stores/balances.store';
 import type { CertifiedStoreData } from '$lib/stores/certified.store';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -183,3 +185,22 @@ export const sumUsdBalances = ([usdBalance1, usdBalance2]: [
...
+	tokenToPair,
+	tokens
+}: {
+	tokenToPair: Token;
+	tokens: Token[];
+}): IcCkToken | undefined =>
+	isRequiredTokenWithLinkedData(tokenToPair)
+		? (tokens.find(
+				(token) => token.symbol === tokenToPair.twinTokenSymbol && isIcCkToken(token)
+			) as IcCkToken | undefined)
```
AntonioVentilii, line 205: shouldn't `isIcCkToken` already cast it as `IcCkToken`? is the casting necessary?

`export const isIcCkToken = (token: Token): token is IcCkToken =>...`

src/frontend/src/tests/lib/utils/token.utils.spec.ts
```diff
@@ -275,3 +280,39 @@ describe('sumUsdBalances', () => {
...
+			tokens: [...mockTokens, ckBtcToken],
+			tokenToPair: BTC_MAINNET_TOKEN
+		});
+
+		expect(result).toBe(ckBtcToken);
+	});
+
+	it('should return undefined if no twin token found', () => {
+		const result = findTwinToken({
+			tokens: mockTokens,
```
AntonioVentilii, line 301: `tokens: [...mockTokens, ckBtcToken]` maybe just to show that there is at least a twin token?

DenysKarmazynDFINITY: done!

src/frontend/src/lib/utils/token.utils.ts
DenysKarmazynDFINITY: Unfortunately, still necessary.

DenysKarmazynDFINITY:
@AntonioVentilii ready for another look

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -207,3 +209,22 @@ export const sumUsdBalances = ([usdBalance1, usdBalance2]: [
...
```
AntonioVentilii, line 229: I just noticed one thing, sorry for the late note:

can you use `IcCkTokenSchema.safeParse` here?

DenysKarmazynDFINITY: I am now using `import { isIcCkToken } from '$icp/validation/ic-token.validation';` to validate if it's a IcCkToken. Is this something you had in mind?

AntonioVentilii: even better! thank you!

PR #3464, opened 2024-11-12
feat(frontend): split ic-wallet.scheduler

peterpeterparker:
# Motivation

We will need another scheduler to support tokens without index canister. Therefore, we split the existing `ic-wallet.scheduler` in two. An abstract class for the start, stop etc. features and the existing implementation to `ic-wallet-transactions.scheduler? .

# Notes

No features or code was implemented here. It's a feature but can also be considered as refactoring.

# Changes

- Split code as described
- Use new "child" class in workers


AntonioVentilii (approved):
LGTM tks!

peterpeterparker:

PR #3465, opened 2024-11-12
style(frontend): use scss instead of postcss

peterpeterparker:
# Motivation

We are using `scss` as CSS processor. In PR https://github.com/dfinity/oisy-wallet/pull/3451#discussion_r1837930148 we discussed `postcss` and noticed similar Tailwind imports can be resolved with our standard approach, therefore this PR removes an inconsistency.


DenysKarmazynDFINITY (approved):
Nice, thanks!

peterpeterparker:

PR #3466, opened 2024-11-12
feat(backend): Verify Derivation Origin's Id Alias from POUH Credential

lmuntaner:
# Motivation

Fix the issue in the POUH credential which didn't check the derivation origin of the POUH credential.

# Changes

* Upgrade `ic-verifiable-credentials` which has the logic to verify the new parameter.
* Expect the new parameter in backend endpoint.
* Update candid and generate related files.
* Pass the derivation origin or the window origin when adding the credential.

# Tests

Tested locally and it works.
Fix tests.


lmuntaner:
@peterpeterparker @AntonioVentilii please review

src/backend/tests/it/utils/mock.rs
```diff
@@ -13,10 +13,11 @@ pub const WEENUS_SYMBOL: &str = "Weenus";
...
+pub const VC_HOLDER: &str = "7eboi-tyuys-aqm4c-w2l7i-vgucm-xvawx-lemzx-6kq2g-f53u7-yvfh2-nae";
 pub const ISSUER_CANISTER_ID: &str = "qdiif-2iaaa-aaaap-ahjaq-cai";
 pub const ISSUER_ORIGIN: &str = "https://dummy-issuer.vc/";
-pub const II_CANISTER_ID: &str = "rdmx6-jaaaa-aaaaa-aaadq-cai";
+pub const VC_DERIVATION_ORIGIN: &str = "https://l7rua-raaaa-aaaap-ahh6a-cai.ic0.app";
+pub const II_CANISTER_ID: &str = "fgte5-ciaaa-aaaad-aaatq-cai";
```
AntonioVentilii, line 21: out of curiosity, why the II canister changed too for the mocks?

lmuntaner: Because I created the JWT with a different II.

AntonioVentilii: Ah understood! tks for the explanation!

lmuntaner: Are you going to approve it? Or should I ask for @peterpeterparker 's review specifically?

AntonioVentilii: honestly, to me it looks good, it seems trivial, but for this I would prefer to rely on @peterpeterparker or @bitdivine to review

peterpeterparker (commented):
Can you provide a frontend test for `requestPouchCredential` service?

I believe it hasn't been covered yet. Ideally, providing the test before this PR, so that the PR explicitly extends it with the changes. However, if we're under time pressure, providing it afterward for sure is also fine.

peterpeterparker:
I went running and give it some more though, so I would have two more questions (😅):

- I know what's the `derivation_origin` of II, but, what's the `derivation_origin` here? Is there any specification about it?

- Is it actually secure to pass this information from the frontend to backend?

src/backend/backend.did
```diff
@@ -5,6 +5,7 @@ type AddUserCredentialError = variant {
   UserNotFound;
 };
 type AddUserCredentialRequest = record {
+  derivation_origin : text;
```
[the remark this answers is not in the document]
lmuntaner: Yes, it's necessary for the new VC SDK and that is precisely what security asked for.

src/frontend/src/lib/services/request-pouh-credential.services.ts
```diff
@@ -112,6 +118,8 @@ export const requestPouhCredential = ({
 }): Promise<ResultSuccess> => {
 	const credentialSubject = identity.getPrincipal();
 	const { auth: authI18n } = get(i18n);
+	const derivationOriginParams = getOptionalDerivationOrigin();
```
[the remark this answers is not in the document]
lmuntaner: Sure, I can do it later.

lmuntaner:
> * I know what's the `derivation_origin` of II, but, what's the `derivation_origin` here? Is there any specification about it?

That is percisely the same concept as the `derivation_origin` in II. That's why I used the same function for logging in. The difference is that when logging in, if not provded, it uses the window origin. In here, we need to be specific of which one.

> * Is it actually secure to pass this information from the frontend to backend?

I don't see any problem. But I will ask security to review this PR.

lmuntaner:
Closing this PR because it's a security issue to pass the derivation origin from the frontend.

PR #3467, opened 2024-11-12
chore(devops): Add backend test canister

bitdivine:
# Motivation
We have lots of test frontend canisters (test-fe-{1..4}) but no similar backend canisters.

# Changes
- Added a test canister for pure-backend testing.
  - Added the canister to cycleops
  - Added the same controllers as for the staging backend

# Tests
- I have used this test canister for testing automated cycles ledger top-up.


AntonioVentilii (approved):
LGTM tks

PR #3468, opened 2024-11-12
test(frontend): icrc-wallet.worker

peterpeterparker:
# Motivation

We extend the test of `icp-wallet.worker` to `icrc-wallet.worker` given that both basically works the same way and with the same scheduler.

# Changes

- Rename `icp-wallet.worker.spect` to `ic-wallet`
- Refactor code to reuse sam `describe` and `it` for both ICP and Icrc
- Integrate ICRC Wallet tests


src/frontend/src/icp/workers/icrc-wallet.worker.ts
```diff
@@ -61,16 +61,20 @@ const mapTransaction = ({
 	return mapIcrcTransaction({ transaction, identity });
 };
 
-const scheduler: IcWalletScheduler<
+// Exposed for test purposes
```
AntonioVentilii, line 64: i don't think it is necessary this comment, right? i mean, i exposed a lot of unnecessary ones just for testing, eheh

Or shall we add this comment in the others too?

AntonioVentilii (approved):
LGTM tks

src/frontend/src/icp/workers/icrc-wallet.worker.ts
peterpeterparker: No strong opinion, I dislike exposing anything just for testing purpose because the production bundles ends with module features exposed for no reason, that's why I compromise with adding a comment which basically mean for me: "I had no other choice".

peterpeterparker:

PR #3474, opened 2024-11-12
feat(frontend): add transaction modals to unified transactions list

AntonioVentilii:
# Motivation

We need to include the transactions modals to provide the details of the selected transactions.

Since there is no unique way of identifying a transaction for each network, we define different selected transactions for each one.

This is what is done in all the other network-specific transaction components.

# Note

Another approach to this, would have been to do like we did for the Transaction card, meaning mapping the modal to each transaction. No particular preference on either method, just pointing out:

```svelte
{#if nonNullish(selectedTransaction) && ($modalBtcTransaction || $modalEthTransaction || $modalIcTransaction)}
	<svelte:component this={selectedTransaction.modal} transaction={selectedTransaction} />
{/if}
```

# Tests

Screenshot 2024-11-14 at 19 07 15


peterpeterparker (commented):
We generally use a `const modalId = Symbol();` pattern for the modals, this is not required here?

AntonioVentilii:
@peterpeterparker for the transaction modals we are not using it apparently: shall we add it (in another PR, I mean)

src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
```diff
@@ -24,6 +36,21 @@
 	$: sortedTransactions = transactions.sort((a, b) =>
 		sortTransactions({ transactionA: a, transactionB: b })
 	);
+
+	let selectedBtcTransaction: BtcTransactionUi | undefined;
```
peterpeterparker, line 40: Since we are already there, would it be worth it creating the Zod types for those transaction and using safe parse?

peterpeterparker (approved):
LGTM, thx

peterpeterparker:
> shall we add it (in another PR, I mean)

Maybe? WDYT?

src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
AntonioVentilii: I have a plan to do it for all the transaction types, but the ETH one is based on an external type. So I will create them soon enough with a zod refine

AntonioVentilii:
[OISY-130]

AntonioVentilii:
> > shall we add it (in another PR, I mean)
> Maybe? WDYT?

Done with:

- https://github.com/dfinity/oisy-wallet/pull/6088

PR #3475, opened 2024-11-12
feat(frontend): remove unnecessary param from the utxos service

DenysKarmazynDFINITY:
# Motivation

"progress" was required but never used in that service. This popped up when I started reusing this method in the conversion flow.


peterpeterparker (approved):
Nice, thx

PR #3478, opened 2024-11-12
feat(frontend): ConvertAmountExchange component

DenysKarmazynDFINITY:
# Motivation

`ConvertAmountExchange` component in action (when some of the values is undefined, just a regular skeleton will be shown):

Screenshot 2024-11-12 at 18 40 11

### DIFF
--- a/src/frontend/src/tests/lib/components/convert/ConvertAmountExchange.spec.ts
+++ b/src/frontend/src/tests/lib/components/convert/ConvertAmountExchange.spec.ts
@@ -1,6 +1,4 @@
 import ConvertAmountExchange from '$lib/components/convert/ConvertAmountExchange.svelte';
-import { formatUSD } from '$lib/utils/format.utils';
-import { assertNonNullish } from '@dfinity/utils';
 import { render } from '@testing-library/svelte';
 
 describe('ConvertAmountExchange', () => {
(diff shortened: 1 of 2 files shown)


peterpeterparker (commented):
Can you provide a test given that there is a bit of logic in the component?

DenysKarmazynDFINITY:
@peterpeterparker sure, added tests, please take a look

src/frontend/src/tests/lib/components/convert/ConvertAmountExchange.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+		exchangeRate
+	};
+
+	const divSelector = 'div[data-tid="convert-amount-exchange"]';
+	const skeletonSelector = 'div[data-tid="convert-amount-exchange-skeleton"]';
+
+	it('should display value with tilde sign if amount is provided', () => {
+		const { container, getByText } = render(ConvertAmountExchange, { props });
+
+		const div: HTMLDivElement | null = container.querySelector(divSelector);
```
peterpeterparker, line 21: You can maybe simplify and use `getByTestId`? Or you explicitely want to use the container to retrieve the elements?

src/frontend/src/tests/lib/components/convert/ConvertAmountExchange.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+		expect(
+			getByText(
+				`~${formatUSD({
+					value: amount * exchangeRate
+				})}`
+			)
+		).toBeInTheDocument();
+	});
+
+	it('should display value without tilde sign if amount is zero', () => {
```
peterpeterparker, line 33: where is the "tilde sign" in this test? should we hardcode the expected value to make it explicit?

src/frontend/src/tests/lib/components/convert/ConvertAmountExchange.spec.ts
DenysKarmazynDFINITY: Let's make them simpler indeed. Updated.

src/frontend/src/tests/lib/components/convert/ConvertAmountExchange.spec.ts
DenysKarmazynDFINITY: Updated to the hardcoded values.

peterpeterparker (approved):
LGTM, thx

PR #3483, opened 2024-11-13
feat(backend): Top up the cycles ledger account

bitdivine:
# Motivation
We wish the backend canister to periodically top up the cycles ledger account.

#3469 defined an API and structures for doing this.  The backend canister sends a proportion of its own cycles to the cycles ledger if the cycles ledger balance is below some threshold.

# Changes
- Implement the body of the top-up function.
- Call the topping up function periodically.
- Fix typos

# Tests
- This is _not_ tested with pocket-ic because the pocket-ic test environment does not include the necessary set-up and we do not wish to wait for that.  pocket-ic tests must follow in a future PR.  See: https://dfinity.atlassian.net/browse/OISY-357
- This code is deployed to `test-be-1`.  The canister balance and cycles balance are tracked manually with:
   ```
   while true ; do
     dfx canister status --network test_be_1 backend | grep Balance | tr "\n" "\t"
     dfx cycles balance --owner "$(dfx canister id backend --network test_be_1)" --ic
     sleep 1m
   done
   ```


src/backend/src/lib.rs
```diff
@@ -147,12 +147,32 @@ fn set_config(arg: InitArg) {
     });
 }
 
+/// Runs housekeeping tasks immediately, then periodically:
+/// - `hourly_housekeeping_tasks`
+fn start_periodic_housekeeping_timers() {
+    // Run housekeeping tasks once, immediately but asynchronously.
+    let immediate = Duration::from_secs(0);
```
peterpeterparker, line 154: There is a `Duration::ZERO` if interesting.

.config/spellcheck.dic
bitdivine, line 1: Updating the spelling dictionary

.config/spellcheck.dic
```diff
@@ -3,3 +3,22 @@ Ethereum
...
+backend
+backend's
+readonly
+pricelist
+ckBTC
++
+1x
+3x
+10x
+80T
```
bitdivine, line 24: Yes, the file really is this primitive.

.config/spellcheck.dic
peterpeterparker: Shouldn't this find place in another PR?

src/backend/src/lib.rs
```diff
@@ -147,12 +147,32 @@ fn set_config(arg: InitArg) {
...
+
+    // Then periodically:
+    let hour = Duration::from_secs(60 * 60);
+    let _ = set_timer_interval(hour, || ic_cdk::spawn(hourly_housekeeping_tasks()));
+}
+
+/// Runs hourly housekeeping tasks:
+/// - Top up the cycles ledger.
+async fn hourly_housekeeping_tasks() {
+    let _ = top_up_cycles_ledger(None).await;
```
peterpeterparker, line 165: Should we `ic_cdk::print` the error? Those can be read with canister logs. It can offers us a quick way to debug if necessary. I would maybe not log the result, that would be verbose.

src/backend/src/signer.rs
```diff
@@ -151,7 +159,52 @@ pub async fn btc_principal_to_p2wpkh_address(
...
+    let account = Account {
+        owner: ic_cdk::id(),
+        subaccount: None,
+    };
+    let (ledger_balance,): (Nat,) = cycles_ledger
+        .icrc_1_balance_of(&account)
+        .await
+        .map_err(|_| TopUpCyclesLedgerError::CouldNotGetBalanceFromCyclesLedger)?;
+
+    if ledger_balance < request.threshold() {
```
peterpeterparker, line 173: Not for this PR but, realizing now: should we have two thresholds?

- One threshold for the ledger balance, as in this line

but also

- Another threshold for the backend canister cycles itself under which we do not want to topup the account?

src/backend/src/signer.rs
```diff
@@ -151,7 +159,52 @@ pub async fn btc_principal_to_p2wpkh_address(
...
+        // Decide how many cycles to keep and how many to send to the cycles ledger.
+        let own_canister_cycle_balance = Nat::from(ic_cdk::api::canister_balance128());
+        let to_send = own_canister_cycle_balance.clone() / Nat::from(100u32)
+            * Nat::from(request.percentage());
```
peterpeterparker, line 177: Should we add an assertion for percentage? > 0 & <= 100

src/backend/src/signer.rs
```diff
@@ -151,7 +159,52 @@ pub async fn btc_principal_to_p2wpkh_address(
...
+        let to_retain = own_canister_cycle_balance.clone() - to_send.clone();
+
+        // Top up the cycles ledger.
+        let arg = DepositArgs {
+            to: account,
+            memo: None,
```
peterpeterparker, line 183: Maybe set `created_at_time` as well?

src/backend/src/signer.rs
```diff
@@ -151,7 +159,52 @@ pub async fn btc_principal_to_p2wpkh_address(
...
+        };
+        let to_send_128: u128 =
+            to_send.clone().0.try_into().unwrap_or_else(|err| {
+                unreachable!("Failed to convert cycle amount to u128: {}", err)
```
peterpeterparker, line 187: Noob question: trap and unreachable do not stop the timer right?

src/backend/src/signer.rs
```diff
@@ -151,7 +159,52 @@ pub async fn btc_principal_to_p2wpkh_address(
...
+                .await
+                .map_err(|_| TopUpCyclesLedgerError::CouldNotTopUpCyclesLedger {
+                    available: own_canister_cycle_balance,
+                    tried_to_send: to_send.clone(),
+                })?;
+        let ledger_balance = result.balance;
+
+        Ok(TopUpCyclesLedgerResponse {
+            ledger_balance,
+            backend_cycles: to_retain,
```
peterpeterparker, line 200: Do not know how precise this value should be but, `to_retain` is not exactly the remaining cycles because the call to the ledger consumed few cycles.

.config/spellcheck.dic
bitdivine: Sure, can do.  Done here: https://github.com/dfinity/oisy-wallet/pull/3503

src/backend/src/lib.rs
bitdivine: Used.  Thanks!

src/backend/src/lib.rs
bitdivine: Yes, that is worthwhile.  I would like proper stats but something is better than nothing.

src/backend/src/signer.rs
bitdivine: I think there is no need, as the command will fail with values of 100 and above anyway.  And if the caller is stingy enough to set 0 they can't be helped! :-D But it doesn't hurt, I suppose.

bitdivine: Done.  Added an error type & pocket-ic test.

src/backend/src/signer.rs
bitdivine: I don't believe that will be needed.  Or what am I missing?  This is used immediately.

src/backend/src/signer.rs
bitdivine: Correct.

src/backend/src/signer.rs
bitdivine: I'm not looking for great precision.  Maybe when writing tests I'll tweak the numbers to make it precise but when monitoring cycle balances I'm not concerned with tiny fractions of a Frank, unlike someone I know. :-D

src/backend/src/signer.rs
peterpeterparker: The `created_at_time` is used for the [deduplication](https://github.com/dfinity/ICRC-1/tree/main/standards/ICRC-1#transaction-deduplication-) mechanism of an icrc1 ledger. That's why I end generally always setting it. But indeed maybe not necessary here.

src/backend/src/signer.rs
peterpeterparker: Oki doki

src/backend/src/signer.rs
bitdivine: If neither memo nor time is set, a standards compliant ledger should not deduplicate, IIRC.  Indeed:

> If the client did not set the created_at_time field, the ledger SHOULD NOT deduplicate the transaction.

So I think we are fine.

PR #3484, opened 2024-11-13
refactor(frontend): simplify component RoundedIcon

AntonioVentilii:
# Motivation

Maybe this PR is not required, but I wanted to check if there are any objections when removing some props from `RoundedIcon`. 

Being a UI component, maybe it is good to have customizable props, but they are really not used for now.

Additionally, we replace `style` with Tailwind classes.

### DIFF
--- a/src/frontend/src/lib/components/ui/RoundedIcon.svelte
+++ b/src/frontend/src/lib/components/ui/RoundedIcon.svelte
@@ -6,6 +6,6 @@
 </script>
 
 <div class="relative">
-	<div class="rounded-full bg-brand-subtle-alt" style={`width: 2.9rem; aspect-ratio: 1/1;`} />
+	<div class="aspect-square w-12 rounded-full bg-brand-subtle-alt" />
 	<svelte:component this={icon} styleClass={`inset-center ${iconStyleClass}`} />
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/ui/RoundedIcon.svelte
```diff
@@ -2,16 +2,10 @@
...
 	export let iconStyleClass = '';
-	export let additionalStyleClass = '';
 </script>
 
 <div class="relative">
-	<div
-		class={`rounded-full ${backgroundStyleClass}`}
-		style={`width: ${iconSize}; aspect-ratio: 1/1; ${additionalStyleClass}`}
-	/>
+	<div class="rounded-full bg-brand-subtle-alt" style={`width: 2.9rem; aspect-ratio: 1/1;`} />
```
peterpeterparker, line 9: Since we are already there, Tailwind classes instead of style?

AntonioVentilii: yes!!! I was hoping for this eheh! done

peterpeterparker (approved):
LGTM, thx

PR #3485, opened 2024-11-13
ci(e2e): create prepare macos action

inc-man:
# Motivation

To run E2E Tests on Apple devices we want to have a prepare action that installs everything required on the github runner.

# Changes

- completely new action

# Tests

- tested on branch


.github/actions/prepare-macos/action.yml
```diff
@@ -0,0 +1,35 @@
+name: 'Setup LLVM and Bash on macOS'
+description: 'Installs LLVM and Bash 5 on macOS and sets up environment variables for wasm32 target'
+
+inputs:
+  os:
+    description: 'Operating system (e.g., ubuntu-latest, macos-latest)'
```
AntonioVentilii, line 6: do we need a choice here? isn't always macos? and if so, i would just use latest

inc-man: Its not really a choice, when we use the action later in the workflow, we use 

with: matrixOS

then the prepare-macos will only be run, if it is macOS and not on Ubuntu. Would leave it like that

AntonioVentilii: wait, maybe I misunderstood: are you not calling this action ONLY when it is the macOS? Why are you calling it for ubuntu too? I mean, you can even put a condition in the action by itself, or when you call it, no?

inc-man: the action itself is in type "composite" and thus cannot use if statements in the action. But yes in the workflow we can use either:

      - name: Prepare macOS
        uses: ./.github/actions/prepare-macos
        with:
          os: ${{ matrix.os }}
          
or:

      - name: Prepare macOS
        if: github_runnerOS == macOS
        uses: ./.github/actions/prepare-macos
        
I personally think the upper version that I am currently using is better and is tested, but really up to you.

AntonioVentilii: my point is: we shouldn't give the choice of any other OS for this action if it is intended to just the macos

so, I would say the correct way to go for this is the second one you proposed: it should be called ONLY when the OS is macOS, without any inputs

inc-man: but we already do that inside the action, every step checks matrixOS and if it is macOS it will run the tasks. But yeah can change it, but need to test again if it actually works

AntonioVentilii (approved):
LGTM tks

PR #3493, opened 2024-11-13
feat(frontend): add skeletons to unified transactions

AntonioVentilii:
# Motivation

We want to have a skeletons on loading the page Activity. It will be shown based on 2 factors:

- at least one transactions store is not initialized;
- all transactions stores are missing at least a token from the required list.

While the above is still `true`, the skeletons will not be shown if at least one token has transactions. It doesn't matter the rest.

# Note

We exclude the Bitcoin transactions store from the checks when locally because it could take longer time to be initialized and in case of no transactions (for example, a new user), it would be stuck to show the skeletons.

As discussed with @DenysKarmazynDFINITY offline, it should be absolutely fine to exclude it from equation locally, we do it already in other places when it comes to BTC.

# Changes

- Create util to check if a transactions store is initialized: it means that all the tokens in the required list must have data there, either a list, an empty list, or `null`. If it is `undefined` it means that it is still not loaded.
- Create util to check if a transactions store is entirely empty: it checks that every token in the required list has `nullish` data or an empty list of transactions.
- Create util to merge the checks: the transactions stores are still `loading` if one of them is nullish, or if one of them have at least a missing token. However, it is not considered in `loading` state anymore the instant that one token has at least a transaction.
- Create skeletons component for `AllTransactionsList`.

# Tests

Apart for the tests of the new utils and component, we tested in the local replica, once with transactions and once without.

https://github.com/user-attachments/assets/d4780fee-7805-442d-9514-eaf29031d0c0

https://github.com/user-attachments/assets/8a46e15a-ef52-4e06-942b-c68c8149d466


DenysKarmazynDFINITY (approved):
Nice one, LGTM

PR #3495, opened 2024-11-13
ci(e2e): added firefox for tests

inc-man:
# Motivation

We would like to not only test against Chrome but also Firefox Browsers.

# Changes

added Firefox Browser in Playwright config.

# Tests

tested on branch


AntonioVentilii (commented):
Was this tested? I don't see any new screenshot on the branch

AntonioVentilii (dismissed):
I see the screenshots were updated, tks!

playwright.config.ts
```diff
@@ -38,9 +38,14 @@ export default defineConfig({
 		...(DEV && { headless: false })
 	},
 	projects: [
+		/* Test against desktop browsers. */
 		{
 			name: 'Google Chrome',
-			use: { ...devices['Desktop Chrome'], channel: 'chrome' }
+			use: { ...devices['Desktop Chrome'] }
```
AntonioVentilii, line 44: honest doubt: is it ok to remove the channel? what is the default?

AntonioVentilii (commented):
Why there are removed screenshots?

playwright.config.ts
inc-man: The channel is only relevant in Chrome. Basically when there is no channel playwright will use the latest Chromium browser. If there is a channel it uses a specific version. I don't think its necessary to test a specific version of chrome, so we can also go with latest Chromium.

e2e/activity-page.spec.ts
```diff
@@ -7,5 +7,5 @@ testWithII('should display activity page', async ({ page, iiPage }) => {
 
 	await activityPage.waitForReady();
 
-	await expect(page).toHaveScreenshot({ fullPage: true });
```
AntonioVentilii, line 10: why removing this? don't we do it for all of them the same?

playwright.config.ts
AntonioVentilii: Asking because it seems this change changed all the screenshots that we already had for Chrome: you can notice that they are quite different (scrollbar, less viewport but same size, strange)

e2e/activity-page.spec.ts
inc-man: I was running the tests on this branch locally and the activity page was throwing random errors. Sometimes the toHaveScreenshot was expecting something that should not be. I then checked what fullPage actually does, and turns out its not necessary unless the screenshot should include "scrollable" content.

AntonioVentilii (approved):
LGTM tks

PR #3496, opened 2024-11-13
ci(e2e): added macos config

inc-man:
# Motivation

We also want to test against macOS devices.

# Changes

- add config to allow testing against macOS

# Tests

tested on branch


AntonioVentilii (approved):
Was this tested? I don't see any new screenshot in the branch

inc-man:
for this to be available for testing I need all the other changes in first so we have macOS runners that work correctly

AntonioVentilii:
@inc-man then we will put it in draft before merging, then await the rest and test it ok?

PR #3499, opened 2024-11-13
ci(e2e): set status check compulsory

AntonioVentilii:
# Motivation

I would like to enforce the status check for E2E test modifications, but to do so, we need to make the status check run always. In fact, if I enable `may-merge-e2e` as status check, no PR will be merged if no E2E test workflow is ran.

So, we turn on the E2E test workflow for all PR, BUT we limit it to changed to the relevant E2E files. Basically, the `may-merge-e2e` status check will be true either:
- no changes for E2E folder or playwright config file;
- there are changes to the relevant files, but the tests passes.

# Tests

The last two commits of this PR are respectively one with the changes in E2E files and one without. In both cases, it worked as expected.


DenysKarmazynDFINITY (approved):
Really good idea, I like it! Approving already.

Just as an idea: how about also check for `canister_e2e_ids.json` and `.env.e2e`?

PR #3502, opened 2024-11-13
feat(frontend): load custom token without index canister

peterpeterparker:
# Motivation

Given that we have now a worker scheduler that periodically fetches for balances without index canister, we can allow the UI to load those kind of tokens.

# Changes

- Remove guard to skip custom tokens without index canister

# Tests

There might be some more use cases that need to be covered by the tests provided here but, given there were none before, feels like what's provided here covers the common use cases at least.


src/frontend/src/icp/services/icrc.services.ts
```diff
@@ -171,7 +165,7 @@ const loadCustomIcrcTokensData = async ({
 				? meta
 				: await metadata({ ledgerCanisterId: ledgerCanisterIdText, identity, certified }),
 			ledgerCanisterId: ledgerCanisterIdText,
-			indexCanisterId: indexCanisterId.toText(),
+			...(nonNullish(indexCanisterId) && { indexCanisterId: indexCanisterId.toText() }),
```
peterpeterparker, line 168: If you rather like to have `indexCanisterId: undefined` within the memory not against. Instead of

```
			...(nonNullish(indexCanisterId) && { indexCanisterId: indexCanisterId.toText() }),
```

we can also do

```
indexCanisterId: indexCanisterId?.toText()
```

As you rather like Antonio.

AntonioVentilii: nono, i like it like this, so that the index canister prop is really missing and not `undefined`

AntonioVentilii (approved):
LGTM tks

PR #3505, opened 2024-11-13
feat(ci): add check for PR description

AntonioVentilii:
# Motivation

Since it is an open repo, we would like to avoid to provide DFINITY Jira/Atlassian links in the description.

# Tests

I tested on this same branch, adding a Jira link in the description, and then removing it.


peterpeterparker (approved):
Yes! LGTM, thx

You know, when it comes to me, we can fully get rid of Jira anyway 😉

PR #3506, opened 2024-11-13
feat(frontend): add json import for additional icrc env tokens

loki344:
# Motivation

We want to load a selection of additional icrc tokens. This prepares the json data structure with zod parsing and typing.

# Changes

- Add types with zod
- Add parsing
- add empty json structure

### DIFF
--- a/src/frontend/src/env/types/env-icrc-token.ts
+++ b/src/frontend/src/env/types/env-icrc-token.ts
@@ -1,5 +1,5 @@
-import { z } from 'zod';
 import { envTokenSymbol } from '$env/types/env-token-common';
+import { z } from 'zod';
 
 export const envIcrcTokenMetadata = z.object({
 	decimals: z.number(),
--- a/src/frontend/src/env/types/env-icrc-token.ts
+++ b/src/frontend/src/env/types/env-icrc-token.ts
@@ -31,6 +31,7 @@ export type EnvIcrcToken = z.infer<typeof envIcrcToken>;
 
 export type EnvIcrcTokens = z.infer<typeof envIcrcTokens>;
 
+// TODO, extract the union into it's own schema
 const envAdditionalIcrcTokens = z.record(envTokenSymbol, z.union([z.undefined(), envIcToken]));
 
 export type EnvAdditionalIcrcTokens = z.infer<typeof envAdditionalIcrcTokens>;
--- a/src/frontend/src/env/types/env-icrc-token.ts
+++ b/src/frontend/src/env/types/env-icrc-token.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import { envTokenSymbol } from '$env/types/env-token-common';
 
 export const envIcrcTokenMetadata = z.object({
 	decimals: z.number(),
@@ -29,3 +30,12 @@ export type EnvIcrcTokenMetadata = z.infer<typeof envIcrcTokenMetadata>;
 export type EnvIcrcToken = z.infer<typeof envIcrcToken>;
 
 export type EnvIcrcTokens = z.infer<typeof envIcrcTokens>;
+
+const envAdditionalIcrcTokens = z.record(envTokenSymbol, z.union([z.undefined(), envIcToken]));
+
+export type EnvAdditionalIcrcTokens = z.infer<typeof envAdditionalIcrcTokens>;
+
+export const envTokensAdditionalIcrc = z.object({
+	production: envAdditionalIcrcTokens,
+	staging: envAdditionalIcrcTokens
+});
--- a/src/frontend/src/env/schema/env-additional-icrc-token.schema.ts
+++ b/src/frontend/src/env/schema/env-additional-icrc-token.schema.ts
@@ -1,10 +1,10 @@
 import { z } from 'zod';
-import { envTokenSymbol } from '$env/types/env-token-common';
 import { EnvIcTokenSchema } from '$env/schema/env-icrc-token.schema';
+import { EnvTokenSymbolSchema } from '$env/schema/env-token-common.schema';
 
 // TODO, extract the union into it's own schema
 export const EnvAdditionalIcrcTokensSchema = z.record(
-	envTokenSymbol,
+	EnvTokenSymbolSchema,
 	z.union([z.undefined(), EnvIcTokenSchema])
 );
 
--- a/src/frontend/src/env/schema/env-additional-icrc-token.schema.ts
+++ b/src/frontend/src/env/schema/env-additional-icrc-token.schema.ts
@@ -1,6 +1,6 @@
-import { z } from 'zod';
 import { EnvIcTokenSchema } from '$env/schema/env-icrc-token.schema';
 import { EnvTokenSymbolSchema } from '$env/schema/env-token-common.schema';
+import { z } from 'zod';
 
 // TODO, extract the union into it's own schema
 export const EnvAdditionalIcrcTokensSchema = z.record(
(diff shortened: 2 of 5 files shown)


AntonioVentilii (approved):
LGTM tks

src/frontend/src/env/types/env-icrc-token.ts
```diff
@@ -29,3 +30,12 @@ export type EnvIcrcTokenMetadata = z.infer<typeof envIcrcTokenMetadata>;
 export type EnvIcrcToken = z.infer<typeof envIcrcToken>;
 
 export type EnvIcrcTokens = z.infer<typeof envIcrcTokens>;
+
+const envAdditionalIcrcTokens = z.record(envTokenSymbol, z.union([z.undefined(), envIcToken]));
+
+export type EnvAdditionalIcrcTokens = z.infer<typeof envAdditionalIcrcTokens>;
+
+export const envTokensAdditionalIcrc = z.object({
```
peterpeterparker, line 38: envAdditionalIcrcTokens
EnvAdditionalIcrcTokens
envTokensAdditionalIcrc

Isn't it a bit inconsistent?
And the first is a schema.

peterpeterparker:
Any other idea than "additional icrc" for the naming? It's bit a generic and to some extention, any ledger is additional.

src/frontend/src/env/types/env-icrc-token.ts
```diff
@@ -29,3 +30,12 @@ export type EnvIcrcTokenMetadata = z.infer<typeof envIcrcTokenMetadata>;
 export type EnvIcrcToken = z.infer<typeof envIcrcToken>;
 
 export type EnvIcrcTokens = z.infer<typeof envIcrcTokens>;
+
+const envAdditionalIcrcTokens = z.record(envTokenSymbol, z.union([z.undefined(), envIcToken]));
```
peterpeterparker, line 34: I'm a bit out of context, probably why I don't get it but, why undefined here?

src/frontend/src/env/types/env-icrc-token.ts
loki344: I tried to be aligned with https://github.com/dfinity/oisy-wallet/blob/main/src/frontend/src/env/types/env-token-ckerc20.ts

envCkErc20Tokens => envAdditionalIcrcTokens
EnvCkErc20Tokens => EnvAdditionalIcrcTokens
envTokensCkErc20 => envTokensAdditionalIcrc

src/frontend/src/env/types/env-icrc-token.ts
loki344: Aiming for this structure: https://github.com/dfinity/oisy-wallet/blob/de85087df27d0ce3267cacfad8740046da7065d2/src/frontend/src/env/types/env-token-ckerc20.ts#L11C1-L12C22

loki344:
> Any other idea than "additional icrc" for the naming? It's bit a generic and to some extention, any ledger is additional.

Yeah I'm open to suggestions. I'm unsure if 'envIcrcTokens' isn't enough as we already have a set of hardcoded ic tokens in the codebase.

src/frontend/src/env/types/env-icrc-token.ts
peterpeterparker: Ah thanks for letting know! I'll have a look (afterwards)

src/frontend/src/env/types/env-icrc-token.ts
peterpeterparker: Gotcha! In another PR, we should extract a generator/schema for the union `z.union([z.undefined(), T])`.
This way we avoid duplicate code and we can express the goal of the `undefined`.

loki344: Ok, I add a todo for this and will address it. Although TBH I'm unsure what the goal of the undefined is. :D

loki344:
@peterpeterparker as you also reviewed, I'm waiting for your OK

peterpeterparker:
> I'm unsure if 'envIcrcTokens' isn't enough as we already have a set of hardcoded ic tokens in the codebase.

Actually:

- The index canister in `envIcToken` should become optional
- `envIcToken` should be renamed to something "Schema"
- All other schema and definition in the related module - `env-icrc-token` - should be moved away and renamed to reflect the fact that they are related to Snses 

> Yeah I'm open to suggestions.

Once everything is done, we can actually rename "additionalIcrcTokens" to "icrcTokens". Not a must but then, the naming will be available. 

I was also thinking at `envIcCoinTokens`.

loki344:
> Actually:
> 
> * The index canister in `envIcToken` should become optional

Would it make sense to do that in the PR that enables the application to handle tokens without an index canister?

> * `envIcToken` should be renamed to something "Schema"

I think this would go against all other places where we define a schema without such a suffix to indicate it's a schema, no?

> * All other schema and definition in the related module - `env-icrc-token` - should be moved away and renamed to reflect the fact that they are related to Snses
> Or "icrcTokens" contains both SNSes tokens and "additional" tokens, then yeah maybe additional make sense.

Ok, I like the idea of having a differentiation between the sns and this 'additiona' icrc tokens. As far as I understand it, this would mean we aim for more consistent naming of the sns tokens. So the 'icrc-custom-tokens.services.ts' which essentially deals with loading the snsTokens from the json has currently a misleading name, correct?

peterpeterparker:
> Would it make sense to do that in the PR that enables the application to handle tokens without an index canister?

Yes, all the steps I listed in separate PR first. 

> such a suffix to indicate it's a schema, no?

I think @AntonioVentilii and I are also using the term schema and have started adapting iteratively - when we notice something - those usages where the convention is not used.

> So the 'icrc-custom-tokens.services.ts' which essentially deals with loading the snsTokens

Nope, this is not related to SNS.

peterpeterparker:
Is this topic urgent? If not, I can take care of it next week.

loki344:
> > such a suffix to indicate it's a schema, no?
> 
> I think @AntonioVentilii and I are also using the term schema and have started adapting iteratively - when we notice something - those usages where the convention is not used.
> 

Ok, I see. Will rename it.

> > So the 'icrc-custom-tokens.services.ts' which essentially deals with loading the snsTokens
> 
> Nope, this is not related to SNS.

Sorry maybe we do not talk about the same code, or maybe I misunderstand something, but this should read the json with the snsTokens in the "buildIcrcCustomTokens" method.. https://github.com/dfinity/oisy-wallet/blob/92169570667fe665082f59e8c1595481950672cf/src/frontend/src/icp/services/icrc-custom-tokens.services.ts#L33

loki344:
> Is this topic urgent? If not, I can take care of it next week.

Depends on what you mean by this topic. Getting in an additional ICRC Token is the goal of this week.

peterpeterparker:
Regarding `buildIcrcCustomTokens`: not for this week but, once we have extended the new environment and if needed, we can rename it. For example, is still solely related to Sns, we can rename it `mapSnsToIcrcCustomTokens` or something. Just thinking out at loud again, no call to action.

Regarding SNS, I provided a refactoring PR #3514.

peterpeterparker:
Regarding naming: let's go with "additional" and if we have an idea along the way, let's refactor later

Regarding "index canister" optional, let's do it when necessary in a separate PR with tests

After #3516, I'll provide one more PR and then we are ready to continue with this PR.

peterpeterparker:
@loki344 alright #3517 is in. Can you rebase with main and resolve the conflicts. Then we should be good.

peterpeterparker (approved):
LGTM, thx

PR #3507, opened 2024-11-13
refactor(frontend): use correct Ck token validation util

AntonioVentilii:
# Motivation

There was a duplicated `isIcCkToken` function that not really fits the correct validation. So we replace with the correct one that uses zod.


peterpeterparker (approved):
LGTM, thx

PR #3509, opened 2024-11-13
feat(frontend): convert store

DenysKarmazynDFINITY:
# Motivation

In this PR, a new store is created. It will be used across the conversion flow files. According to the new designs, we need to display data about both source and destination tokens, therefore the store contains the same set of info for both provided tokens.

### DIFF
--- a/src/frontend/src/lib/stores/convert.store.ts
+++ b/src/frontend/src/lib/stores/convert.store.ts
@@ -1,0 +1,75 @@
+import { exchanges } from '$lib/derived/exchange.derived';
+import { balancesStore } from '$lib/stores/balances.store';
+import type { Token } from '$lib/types/token';
+import { nonNullish } from '@dfinity/utils';
+import { BigNumber } from '@ethersproject/bignumber';
+import { derived, writable, type Readable } from 'svelte/store';
+
+export interface ConvertData {
+	sourceToken: Token;
+	destinationToken: Token;
+}
+
+export interface ConvertStore extends Readable<ConvertData> {
+	set: (data: ConvertData) => void;
+}
+
+export const initConvertStore = (data: ConvertData): ConvertStore => {
+	const { subscribe, set: setStore } = writable<ConvertData>(data);
+
+	return {
+		subscribe,
+
+		set(data: ConvertData) {
+			setStore(data);
+		}
+	};
+};
+
+export const initConvertContext = (convertData: ConvertData): ConvertContext => {
+	const convertStore = initConvertStore(convertData);
+
+	const sourceToken = derived([convertStore], ([{ sourceToken }]) => sourceToken);
+	const destinationToken = derived([convertStore], ([{ destinationToken }]) => destinationToken);
+
+	const sourceTokenBalance = derived(
+		[balancesStore, sourceToken],
+		([$balancesStore, $sourceToken]) =>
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/stores/convert.store.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export const initConvertContext = (convertData: ConvertData): ConvertContext => {
+	const convertStore = initConvertStore(convertData);
+
+	const sourceToken = derived([convertStore], ([{ sourceToken }]) => sourceToken);
+	const destinationToken = derived([convertStore], ([{ destinationToken }]) => destinationToken);
+
+	const sourceTokenBalance = derived(
+		[balancesStore, sourceToken],
+		([$balancesStore, $sourceToken]) =>
```
peterpeterparker, line 37: That's a new pattern. I like the idea, but I'm not sure if I like placing it in the `/src/**/stores` package. It gives me a bit of a recursive import vibe, like `derived` using `stores`, which then uses `derived`. What do you think?

src/frontend/src/lib/stores/convert.store.ts
```diff
@@ -0,0 +1,75 @@
...
+export interface ConvertData {
+	sourceToken: Token;
+	destinationToken: Token;
+}
+
+export interface ConvertStore extends Readable<ConvertData> {
+	set: (data: ConvertData) => void;
+}
+
+export const initConvertStore = (data: ConvertData): ConvertStore => {
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Good point - it's not really needed. Simplified it.

src/frontend/src/lib/stores/convert.store.ts
DenysKarmazynDFINITY: I think `send.store` uses a similar pattern for `sendBalance`. Maybe in another PR we can introduce a new folder `contexts` (or something like `context-stores`) and move there all similar context files where no custom stores needed? It does not solve the "recursion" feeling completely, though.

peterpeterparker: Ah yeah right, don't know why I forgot about it. I think at some point we already discussed the idea of a `contexts` package with @AntonioVentilii so sounds good to me.

src/frontend/src/tests/lib/stores/convert.store.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+	it('should ensure derived stores update at most once when the store changes', async () => {
+		await testDerivedUpdates(() =>
+			initConvertContext({
+				destinationToken: ETHEREUM_TOKEN,
+				sourceToken: ICP_TOKEN
+			})
+		);
+	});
+
+	it('should have all expected properties', () => {
```
peterpeterparker, line 21: Can you extend the test one step further be feeding some data in the `balanceStore` and `exchanges` and assert that the derived values are expected?

DenysKarmazynDFINITY: Yes, sure, should have done it right away 🙂

DenysKarmazynDFINITY:
@peterpeterparker tests are ready, please take a look

peterpeterparker (approved):
LGTM, thx

PR #3512, opened 2024-11-13
feat(frontend): support for no index on transactions page

peterpeterparker:
# Motivation

On the transactions page, if the ic transactions store is set to `null`, we assume this is because the token has no index canister and per extension, we want to show a related explanation and placeholder.

# Changes

- New "no index" placeholder
- Use placeholder
- Update skeleton to load while `undefined` instead of `nullish`

# Tests

I just provide a test to check if this or that placeholder is set. We need more tests but, felt like it's quite some work not related to this feature.

# Screenshot

Capture d’écran 2024-11-13 à 17 32 51


src/frontend/src/lib/components/icons/IconSearchClose.svelte
```diff
@@ -0,0 +1,27 @@
+<!-- source: DFINITY foundation -->
```
AntonioVentilii, line 1: this is not from Lucide? I had understood that we were using only from Lucide

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/components/icons/IconSearchClose.svelte
peterpeterparker: I explicitely asked Artem if I should export from Figma and he said yes. So assumed it was de facto a DFINITY foundation icon. Is that incorrect?

peterpeterparker:

PR #3513, opened 2024-11-13
fix(frontend): modal behaviour on back navigation

BonomoAlessandro:
# Motivation

When using the browser's back button, for instance, the modals were not closing as expected, leading to unforeseen behavior during navigation. In order to resolve this issue, I added an event listener to the open modals. This ensures correct handling of the browser's back navigation.

# Changes

- fixes modal behaviour on nagivating

# Tests

before:

https://github.com/user-attachments/assets/120baaa5-bc00-4524-9177-e4b0d410e50b

after:

https://github.com/user-attachments/assets/f5d7ec41-6222-412c-a85a-7284c17de674


AntonioVentilii (commented):
I am not really sure I understood the changes: aren't you calling `modalStore` that is initiated by `initModalStore` inside the same function? isn't it a loop?

Maybe it would be useful to create some tests for this store?

peterpeterparker (changes requested):
In my opinion, a few things need to be addressed:

1. With the exception of the auth store, which has a specific historical design, stores are meant to hold state only, without embedded logic. Adding logic here to close the modal under certain conditions diverges from our intended architecture.

2. The implementation introduces a public `set` function, exposing a generic setter across the application. This approach conflicts with our use of `open...` patterns for handling modals.

3. There's no guarantee that the event listeners will be removed. Without a destructor or cleanup mechanism, event listeners may persist if a modal is destroyed without explicitly calling close.

4. If the intention is to treat the modal as part of navigation history, why not consider adding an entry to the session history stack on modal open, rather than hijacking the close event?

5. Tests are missing and should be provided.

I’m marking the PR as "Requesting Changes" for the reasons outlined above and because I believe it should not be merged into main if implemented in its current form.

BonomoAlessandro:
@AntonioVentilii @peterpeterparker 
Would it be a possibility to define the onpopstate event handler in the root layout? 
this means: 
- there will be no embedded logic in the store, which means it does not break the intended architecture. 
- no additional event listeners are added.

BonomoAlessandro:
@peterpeterparker 
I don't understand why we shouldn't solely rely on the popstate event and why we should add an entry to the history? Is you argument that this would be a cleaner conceptual solution?

If you have a look at the code i provided in this PR, you can see that i updated the window.onpopstate event handler so that open modals get closed. Like this we shouldn't have a problem that the listener is not removed automatically.

src/frontend/src/routes/+layout.svelte
```diff
@@ -82,6 +83,14 @@
 		const spinner = document.querySelector('body > #app-spinner');
 		spinner?.remove();
 	})();
+
+	/**
+	 * Event handler
+	 */
+
+	window.onpopstate = function () {
```
peterpeterparker, line 91: - I'm confused about the goal of the implementation: Is the goal of the feature to call the `close` function of the modal on navigation back, or is it to treat the modal as an element of the navigation history? The implementation seems to address the former.

- Why not using `svelte:window`, which handles destroying the listener automatically?

- Nitpick: We use arrow functions.

- Not sure this belongs in the root layout. Maybe we should create a component like `Toasts` or `Busy` to scope this implementation? It would be easier to test that way.

- Maybe not for this PR, but if the application is `$isBusy`, should we prevent the modal from being closed?

peterpeterparker: > yes, the goal is to close open modals when clicking on the browser back button to prevent undefined behaviour.

So just closing, right? For example, if a user browses to "Activity," then "Assets," and then opens the "Receive modal", what should happen when the user clicks back?

1. Close the modal and redirect the user to "Activity".
2. Close the modal and remain on the "Assets" page.

> I don't understand why

Because the main layout initalize the authentication, the analytics, the transations, synchronize the authentication, track some analytics, mount some workers for the authentication etc. etc. so sure it can be tested be you'll have to deal with all those things when you'll write a test.

BonomoAlessandro: @peterpeterparker 

> So just closing, right? For example, if a user browses to "Activity," then "Assets," and then opens the "Receive modal", what should happen when the user clicks back?

It should do the first option. This means, closing the modal and redirect the user to the "Activity" page.

> Because the main layout initalize the authentication, the analytics...

Ok, i see. I will scope this implementation into its own component.

BonomoAlessandro: @peterpeterparker I just introduced the `ModalExitHandler`.

src/frontend/src/lib/components/ui/ModalExitHandler.svelte
```diff
@@ -0,0 +1,5 @@
+<script>
```
peterpeterparker, line 1: for consistency reason: `<script lang="ts>`

src/frontend/src/lib/components/ui/ModalExitHandler.svelte
```diff
@@ -0,0 +1,5 @@
...
+	import { modalStore } from '$lib/stores/modal.store';
+</script>
+
+<svelte:window on:popstate={() => modalStore.close()} />
```
peterpeterparker, line 5: if you like it, can be shortened:

```
<svelte:window on:popstate={modalStore.close} />
```

peterpeterparker (commented):
LGTM, thanks for the changes.

Can you provide a test as well?

BonomoAlessandro:
@peterpeterparker i just had a chat with stefan and he also meant that this would be a good idea. I will resolve this in another PR.

PR #3514, opened 2024-11-13
refactor(frontend): rename icrc env tokens specific for sns

peterpeterparker:
# Motivation

On one hand, we will extend the environment with some "additional icrc tokens" #3506, on the other, existing "sns env tokens" are not solely "icrc tokens" because they contains addition information such as the root canister.

To clean up a bit the naming and data inconsistency, this PR move and rename the sns related env types to a particular module and also add the "Schema" suffix we have started to use for our zod schema.

# Tests

Skipped.

### DIFF
--- a/src/frontend/src/icp/services/icrc-custom-tokens.services.ts
+++ b/src/frontend/src/icp/services/icrc-custom-tokens.services.ts
@@ -1,7 +1,7 @@
 import { SNS_EXPLORER_URL } from '$env/explorers.env';
 import { ICP_NETWORK } from '$env/networks.env';
 import snsTokens from '$env/tokens.sns.json';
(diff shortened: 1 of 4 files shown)


src/frontend/src/icp/services/icrc-custom-tokens.services.ts
```diff
@@ -28,26 +28,24 @@ export const buildIndexedIcrcCustomTokens = (): Record<
  */
 export const buildIcrcCustomTokens = (): IcTokenWithoutIdExtended[] => {
```
loki344, line 29: Would it make sense to rename this? I find it rather confusing that we're talking about customTokens here (and in the filename)

loki344: Ah, saw your comment in the other pr, all good

loki344 (approved):
Thank you

PR #3516, opened 2024-11-13
feat(frontend): env icrc token schema suffix and module

peterpeterparker:
# Motivation

Continuing with some cleanup of our env tokens by splitting schema and types (as we newly do) and using suffix Schema.

Follow-up of #3514.

# Tests

Skipped.


loki344 (approved):
Thank you

PR #3520, opened 2024-11-13
refactor(frontend): env common symbol schema

peterpeterparker:
# Motivation

Use our naming and structure convention for common env symbol.

# Tests

Skipped.

### DIFF
--- a/src/frontend/src/env/schema/env-token-ckerc20.schema.ts
+++ b/src/frontend/src/env/schema/env-token-ckerc20.schema.ts
@@ -1,5 +1,5 @@
 import { EnvIcTokenSchema } from '$env/schema/env-icrc-token.schema';
-import { envTokenSymbol } from '$env/types/env-token-common';
+import { EnvTokenSymbolSchema } from '$env/schema/env-token-common.schema';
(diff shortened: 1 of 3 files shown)


src/frontend/src/env/schema/env-token-ckerc20.schema.ts
```diff
@@ -13,7 +13,7 @@ const EnvCkErc20TokenDataSchema = EnvIcTokenSchema.extend({
 });
 
 export const EnvCkErc20TokensSchema = z.record(
-	envTokenSymbol,
+	EnvTokenSymbolSchema,
 	z.union([z.undefined(), EnvCkErc20TokenDataSchema])
```
AntonioVentilii, line 17: you don't want to move this one in this same PR, right?

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/env/schema/env-token-ckerc20.schema.ts
peterpeterparker: That's another task, which I wasn't planing to solve today. We need a generic `z.union([z.undefined(), T])` because this union is used in #3506 as well.

PR #3521, opened 2024-11-13
feat(frontend): btc post message strict

peterpeterparker:
# Motivation

BTC was using the same interfaces for post message as the ICP/ICRC tokens. Given that the index becomes optional for the later, it would introduce a potential nullability for BTC "newTransactions" which actually is incorrecty, given that we always fetch transactions for BTC.

Long story short, we need a separate post message strict types for BTC.

# Changes

- Create schema and types for strict btc post message
- Use those
- Add todos for some existing types which should be reviewed


src/frontend/src/icp/types/btc.post-message.ts
```diff
@@ -1,3 +1,6 @@
+// TODO: zod
```
AntonioVentilii, line 1: i missed it: why is this TODO removed?

src/frontend/src/tests/btc/schema/btc-post-message.schema.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+		it('should fail if newTransactions is missing', () => {
+			const invalidData = {
+				wallet: {
+					balance: mockValidBalance
+				}
+			};
+			expect(() => BtcPostMessageDataResponseWalletSchema.parse(invalidData)).toThrow();
+		});
+
+		it('should validate if balance is not a bigint because of zod custom', () => {
```
AntonioVentilii, line 31: that is why I am afraid of zod custom eheh! but nothing we can do i imagine

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/icp/types/btc.post-message.ts
peterpeterparker: the TODO is added 😉

src/frontend/src/tests/btc/schema/btc-post-message.schema.spec.ts
peterpeterparker: We have some TODOs to remove the custom but yeah, as long we use custom, nothing we can do except writing a refine but, not sure it's worth it if we want to get rid of the custom in first place

src/frontend/src/icp/types/btc.post-message.ts
AntonioVentilii: 🤦‍♂️

PR #3522, opened 2024-11-13
refactor(frontend): commot props for transaction UI

AntonioVentilii:
# Motivation

A few props are in common with all the transaction UI. So we use `TransactionUiCommon`.

# Tests

Unfortunately, the Transaction type is derived from an external type: meaning that ZOD will not correctly test correct parsing. But I accept any kind of suggestion


AntonioVentilii:
test failed

PR #3523, opened 2024-11-13
feat(frontend): add burn icrc token from json

loki344:
# Motivation

We add the additional icrc token "BURN".

# Changes

- adjust type reference
- add token to json
- merge data into icrc tokens

# Tests

In discussion with @AntonioVentilii-DFINITY I aimed for manual testing because testing this functionality in a unit test would include multiple stores and a deeper knowledge of how they work together.

Acceptance criteria:
User w/o the token: user does not see the token in the list, and sees the token in the list settings and can enable it

2650331

https://github.com/user-attachments/assets/12139ae4-a4ee-4d60-94d7-6c152a63856d

User w/ the token, a balance, active: User still sees the token and the balance in the list. and only once in “manage list” and can enable it

2650332

https://github.com/user-attachments/assets/7baad085-ecca-4f3d-a841-726c512cc81e

image

User w/ the token, a balance, inactive: User does not see the token in the list. and only once in “manage list” and can enable it

2650334

https://github.com/user-attachments/assets/e4abf017-b0f1-456e-b31f-792071a58883


src/frontend/src/env/networks.icrc.env.ts
```diff
@@ -385,6 +387,35 @@ const CKXAUT_IC_DATA: IcCkInterface | undefined = nonNullish(CKERC20_PRODUCTION_
 		}
 	: undefined;
 
+/**
+ * Additional ICRC tokens from JSON file
+ */
+const mapIcrcData = (
```
AntonioVentilii, line 393: may you please create some tests for this one?

and shouldn't this go somewhere in utils? or is it done in the same way elsewere?

src/frontend/src/env/networks.icrc.env.ts
```diff
@@ -385,6 +387,35 @@ const CKXAUT_IC_DATA: IcCkInterface | undefined = nonNullish(CKERC20_PRODUCTION_
...
+	icrcTokens: EnvAdditionalIcrcTokens
+): Record<EnvTokenSymbol, Omit<IcInterface, 'position'>> =>
+	Object.entries(icrcTokens).reduce(
+		(acc, [key, value]) => ({
+			...acc,
+			...((STAGING || BETA || PROD) &&
```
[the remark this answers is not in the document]
loki344: **Why only these env vars?**
Because it somehow seems to matter here: https://github.com/dfinity/oisy-wallet/blob/22c0653650a84615e049c0d6ec61f21e45637797/src/frontend/src/env/networks.icrc.env.ts#L225
So i thought it might also matter for this mapping

**Why I only use it for the production tokens?**
I could add a line like this: 
`const ADDITIONAL_ICRC_STAGING_DATA = mapIcrcData(additionalIcrcTokensStaging);`

but because there is no staging token that we want to read right now, this variable will be unused. So i think is better to introduce it when it's needed.

src/frontend/src/env/networks.icrc.env.ts
loki344: done

loki344:
@AntonioVentilii can you take another look so that we get this in today?

src/frontend/src/env/networks.icrc.env.ts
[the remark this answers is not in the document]
AntonioVentilii: ah understood! thank you for the explanation! then, shouldn't the check be only for BETA and PROD?

src/frontend/src/env/map-icrc-data.ts
AntonioVentilii, line 1: not really confortable with an util mixed in the env folder; may you please move it in the util folder maybe inside the icp/utils ?

src/frontend/src/tests/env/map-icrc-data.spec.ts
AntonioVentilii, line 1: good one! however, the tests should be for each env, so that if at some point we change something, they fail

src/frontend/src/env/networks.icrc.env.ts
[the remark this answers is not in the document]
loki344: I don't feel strongly about this as I just replicated the existing pattern, i'll remove the staging for now

src/frontend/src/env/map-icrc-data.ts
loki344: yeah I can move it

AntonioVentilii (approved):
LGTM tks

src/frontend/src/env/tokens.icrc.json
```diff
@@ -1,4 +1,9 @@
 {
-	"production": {},
+	"production": {
+		"BURN": {
+			"indexCanisterId": "nrant-tyaaa-aaaag-atsjq-cai",
+			"ledgerCanisterId": "egjwt-lqaaa-aaaak-qi2aa-cai"
```
peterpeterparker, line 5: Similar to SNS tokens, is there any plan to pre-render the metadata of the ledgers we statically add to OISY to avoid a query and an update call for loading the metadata at runtime?

peterpeterparker: I assumed we were already pre-fetching metadata for ICRC but, we actually only do so for SNS. So this might be a nice improvement afterwards but, not strictly related to this task.

peterpeterparker (approved):
LGTM, thx

Note that I did not really tested. Happy to do so once it's on staging or beta.

AntonioVentilii:
Merging since @loki344 is on vacation/holidays

PR #3525, opened 2024-11-13
feat(frontend): nullify ic transactions store on post message

peterpeterparker:
# Motivation

When the IC web worker scheduler runs without index canister to solely fetch the balance from the ledger, the related ic transactions store entry for the token should be nullified.

# Changes

- `nullify` ic transactions store if undefined transactions is received (no matter what - i.e. regardless if there were transactions in memory)


AntonioVentilii (approved):
LGTM tks!

PR #3526, opened 2024-11-13
feat(frontend): util to format a timestamp to a normalized date

AntonioVentilii:
# Motivation

For grouping the transactions by date, we create an util that formats a timestamp to a normalized date, for example, using today as reference:

- `Today` if it's today, i.e. `13 November 2024`
- `Yesterday` if it's yesterday, i.e.`12 November 2024`
- `17 June` if it's in the same year
- `31 December 2023` if is in the past years

However , we provide an optional param that is the date of reference, since we want to maintain the same reference if used in a loop. Maybe over-engineered, but it feels more correct.

# Note

The final scope is to present the transactions in smaller groups, by date, for example (this is just a mockup):

Screenshot 2024-11-13 at 17 09 24
Screenshot 2024-11-13 at 17 10 09

# Tests

Included some tests.


peterpeterparker (commented):
Why not using ic-js? The function is different?

https://github.com/dfinity/ic-js/tree/main/packages/utils#gear-secondstoduration

peterpeterparker:
I think so, might be worth giving it a try. In anycase, if the foundation is representing the dates a certain way in NNS dapp, might be worth using the display in OISY, or change both, just saying. If we can spare reinventing the wheel also it's not too bad.

AntonioVentilii:
@peterpeterparker , [as you suggested](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/RelativeTimeFormat#using_the_auto_option), I used `numeric: auto` and it worked !

src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -68,6 +68,30 @@ export const formatNanosecondsToDate = (nanoseconds: bigint): string => {
...
 
+export const formatSecondsToNormalizedDate = ({
+	seconds,
+	currentDate
+}: {
+	seconds: number;
+	currentDate?: Date;
+}): string => {
+	const date = new Date(seconds * 1000);
+	const today = currentDate ?? new Date();
```
peterpeterparker, line 79: Naming question: when is current date not the current date or when is current date provided not as the current date?

src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -68,6 +68,30 @@ export const formatNanosecondsToDate = (nanoseconds: bigint): string => {
...
+	const daysDifference = Math.ceil((date.getTime() - today.getTime()) / (1000 * 60 * 60 * 24));
```
peterpeterparker, line 80: Not sure if it would becomes too verbose but we have some constants like `SECONDS_IN_MINUTE` etc.

src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -68,6 +68,30 @@ export const formatNanosecondsToDate = (nanoseconds: bigint): string => {
...
+
+	if (Math.abs(daysDifference) < 2) {
+		return new Intl.RelativeTimeFormat('en', { numeric: 'auto' }).format(daysDifference, 'day');
+	}
+
+	// Same year, return day and month name
+	if (date.getFullYear() === today.getFullYear()) {
+		return date.toLocaleDateString('en', { day: 'numeric', month: 'long' });
```
peterpeterparker, line 88: I just read the following on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString)

> When the method is called many times with the same arguments, it is better to create a [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) object and use its [format()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/format) method, because a DateTimeFormat object remembers the arguments passed to it and may decide to cache a slice of the database, so future format calls can search for localization strings within a more constrained context.

Maybe for this implementation and our existing implementation, we should avoid to recreate always `new Intl` or `date.toLocal...` to improve the performance?

**NOT** for this PR, just sharing and asking.

src/frontend/src/lib/utils/format.utils.ts
AntonioVentilii: the idea is to provide the date of reference, so if used in a loop, all transactions have the same base to calculate. If you want, i can remove the `const today = currentDate ?? new Date();` and just make the user provide the today date as input always

src/frontend/src/lib/utils/format.utils.ts
AntonioVentilii: makes sense, done

src/frontend/src/lib/utils/format.utils.ts
AntonioVentilii: yes, it makes sense! i add a TODO and put it in my backlog

src/frontend/src/lib/utils/format.utils.ts
peterpeterparker: No no, now I understand. Maybe a comment to describe this usage?

peterpeterparker (approved):
LGTm, thx

peterpeterparker (approved):

src/frontend/src/lib/utils/format.utils.ts
AntonioVentilii: Sorry, I missed in this PR; done with:

- #3543

src/frontend/src/lib/utils/format.utils.ts
AntonioVentilii: Done with :

- PR #3603

AntonioVentilii:
[OISY-130]

PR #3529, opened 2024-11-13
feat(frontend): util to group transactions by date

AntonioVentilii:
# Motivation

After PR #3526 , we create a function to group transactions by date.

# Note

The final scope is to present the transactions in smaller groups, by date, for example (this is just a mockup):

Screenshot 2024-11-13 at 17 09 24
Screenshot 2024-11-13 at 17 10 09

# Tests

Included some tests.


src/frontend/src/lib/utils/transaction.utils.ts
```diff
@@ -29,3 +34,34 @@ export const mapTransactionIcon = ({
...
+ * Group a list of transactions by date.
+ *
+ * It assumes that the transactions are already sorted by timestamp in descending order.
+ * It does not sort the transactions, nor the groups.
+ * If a transaction has no timestamp, it will be grouped under the key 'undefined'.
+ *
+ * @param transactions - List of transactions to group.
+ * @returns Object where the keys are the date and the values are the transactions for that date.
+ */
+export const groupTransactionsByDate = <
```
peterpeterparker, line 48: Can you type the return type?

src/frontend/src/lib/utils/transaction.utils.ts
```diff
@@ -29,3 +34,34 @@ export const mapTransactionIcon = ({
...
+) => {
+	const currentDate = new Date();
+
+	return transactions.reduce<Record<string, T[]>>((acc, transaction) => {
+		if (isNullish(transaction.timestamp)) {
+			return { ...acc, undefined: [...(acc['undefined'] ?? []), transaction] };
+		}
+
+		const date = formatSecondsToNormalizedDate({
+			seconds: Number(transaction.timestamp),
```
peterpeterparker, line 61: We are sure the timestamp is always seconds here?

src/frontend/src/lib/utils/transaction.utils.ts
```diff
@@ -29,3 +34,34 @@ export const mapTransactionIcon = ({
...
+			currentDate
+		});
+
+		return { ...acc, [date]: [...(acc[date] ?? []), transaction] };
```
peterpeterparker, line 65: the order of the transaction within the group does not matter?

AntonioVentilii: so, this function does not resort, it uses the same order that the transactions are provided in. So, if they are already sorted by date, for example, the order will be respected

src/frontend/src/lib/utils/transaction.utils.ts
AntonioVentilii: done! any better naming?

`export type AnyTransactionUi = BtcTransactionUi | EthTransactionUi | IcTransactionUi;`

src/frontend/src/lib/utils/transaction.utils.ts
AntonioVentilii: So, I would have managed this when using this function: the list of transaction provided would be all with timestamp in seconds; but I can put a sort of guard here: `if timestamp is same order of now*1000 or now*1000000 ....`

AntonioVentilii: I would not modify the transactions to their core to make them all in seconds (BTW, only IC transactions are in nanoseconds)

peterpeterparker: If we are sure those are seconds it's fine, I was just thinking that it would be nice to not start displaying date like Dec 20 3098 or Jan 1 1970. So if there is a risk, yeah maybe a guard would be nice.

src/frontend/src/lib/utils/transaction.utils.ts
peterpeterparker: Maybe add a comment for our later selfs

peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3534, opened 2024-11-13
feat(frontend): onSend callbacks in btc-send.services

DenysKarmazynDFINITY:
# Motivation

The idea is to decouple btc-send.services from the send flow so the service can be re-used by the conversion wizard too.

### DIFF
--- a/src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
+++ b/src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
@@ -113,8 +113,13 @@
 				network,
 				source,
 				identity: $authIdentity,
-				onSendStarted: () => progress(ProgressStepsSendBtc.SEND),
-				onSendCompleted: () => progress(ProgressStepsSendBtc.DONE)
+				onProgress: () => {
+					if (sendProgressStep === ProgressStepsSendBtc.INITIALIZATION) {
+						progress(ProgressStepsSendBtc.SEND);
+					} else if (sendProgressStep === ProgressStepsSendBtc.SEND) {
+						progress(ProgressStepsSendBtc.DONE);
+					}
+				}
 			});
 
 			sendProgressStep = ProgressStepsSendBtc.DONE;
--- a/src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
+++ b/src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
@@ -110,10 +110,11 @@
 				destination,
 				amount,
 				utxosFee,
-				progress,
 				network,
 				source,
-				identity: $authIdentity
+				identity: $authIdentity,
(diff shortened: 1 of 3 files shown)


src/frontend/src/btc/services/btc-send.services.ts
```diff
@@ -77,12 +78,12 @@ const send = async ({
 	network,
 	amount,
 	utxosFee,
-	progress
+	onSendStarted
 }: Omit<SendBtcParams, 'source'>): Promise<SendBtcResponse> => {
```
AntonioVentilii, line 82: should you omit `onSendCompleted`? or is there a future use-case?

DenysKarmazynDFINITY:
@peterpeterparker @AntonioVentilii please take a look at https://github.com/dfinity/oisy-wallet/pull/3553 first. I'll come back to this PR after the tests are merged.

peterpeterparker (commented):
I have to say I rather like a single callback. By having two, there is no guarantee that somehow someday, start and complete for some bug reason are triggered at the same time, you know what I mean?

DenysKarmazynDFINITY:
@peterpeterparker yep, agree. 

@peterpeterparker @AntonioVentilii I have updated the code to use a single callback. WDYT?

peterpeterparker (approved):
I like `onProgress`, thanks.

Since @AntonioVentilii also participated maybe wait for his feedback too before merge.

src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
```diff
@@ -110,10 +110,16 @@
...
 				utxosFee,
-				progress,
 				network,
 				source,
-				identity: $authIdentity
+				identity: $authIdentity,
+				onProgress: () => {
+					if (sendProgressStep === ProgressStepsSendBtc.INITIALIZATION) {
+						progress(ProgressStepsSendBtc.SEND);
+					} else if (sendProgressStep === ProgressStepsSendBtc.SEND) {
```
AntonioVentilii, line 119: just a smaller question, unrelated: did the linter flagged something here or not?

AntonioVentilii (approved):

AntonioVentilii (approved):
LGTM! Tks

src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
DenysKarmazynDFINITY: Nope, linter didn't complain here. I think it would do it if I had something like:
```
if (sendProgressStep === ProgressStepsSendBtc.INITIALIZATION) {
  progress(ProgressStepsSendBtc.SEND);
  return;
} else if (sendProgressStep === ProgressStepsSendBtc.SEND) {
...
```

PR #3535, opened 2024-11-13
test(frontend): activate undefined index canister checks

peterpeterparker:
# Motivation

We can now activate the two tests for the `undefined` Index canister ID in the `custom-token.services`.


AntonioVentilii (approved):
Good one!

PR #3536, opened 2024-11-13
feat(frontend): align ICP wallet worker pattern and tests

peterpeterparker:
# Motivation

As discussed in a previous PR, we aim to test the "initWalletScheduler" function rather than directly hooking into the initializer for transactions or balance. However, this requires a change because the ICP wallet worker always fetches the balance from the Index, and this pattern hasn’t been updated in the same way as for ICRC.

This PR aligns the pattern and updates the tests accordingly.


AntonioVentilii (approved):
LGTM tks!

PR #3538, opened 2024-11-13
test(frontend): TokenLogo component

AntonioVentilii:
# Motivation

Some tests for TokenLogo Component.


src/frontend/src/lib/components/tokens/TokenLogo.svelte
```diff
@@ -42,6 +46,7 @@
 				src={badge.blackAndWhite ? networkIconBW : networkIcon}
 				alt={replacePlaceholders($i18n.core.alt.logo, { $name: networkName })}
 				{color}
+				testId={badgeTestId}
```
peterpeterparker, line 49: Is it expected that you use twice the same `testId` within the component?

AntonioVentilii: the component will render one or the other badge, it is exclusive; shall i maybe add a suffix/prefix, line `token-count-{badgeTestId}` and `network-{badgeTestId}` ?

peterpeterparker: Maybe? Somehow always thought `testId` are kind of unique.

AntonioVentilii: done! not really sure if should be prefix or suffix; have a look, WDYT? any preference?

peterpeterparker (approved):
LGTM, thx

PR #3540, opened 2024-11-14
refactor(cycles-ledger-client): Move cycles ledger client to a subdirectory

bitdivine:
# Motivation
Previously didc was able to generate only the cycles ledger client. Now we would like it to generate:
- The API types only (data transfer objects), as a crate
- The client API only (using the types above)
- The pocket-ic bindings only (using the types above)

These crates are to be at:
- `src/$canister/types` - Data Transfer Objects
- `src/$canister/client` - Client for use inside canister code
- `src/$canister/pic` - Bindings for use in pocket-ic tests

Notes:
* The client and pic crates cannot be compiled into the same crate without causing dependency issues.  The canister code will use the client, pocket-ic tests will use the pic bindings.
* Having the same types available from two locations (the the pic crate and the client crate) yields confusing compiler errors.  Hence the choice to have a common types crate.

# Changes
Move the client bindings from `src/cycles_ledger_client` to `src/cycles_ledger/client`

# Tests
Existing CI should suffice.


AntonioVentilii (approved):
LGTM tks

PR #3541, opened 2024-11-14

sa-github-api:


peterpeterparker:
ICGhost Index canister was reported as outdated which happens when the canister is out of cycles but, it's respondid now. 
Likewise the API is up-to-date. Therefore closing.

PR #3542, opened 2024-11-14
feat(frontend): hide custom token without Index canister

peterpeterparker:
# Motivation

Now that we support custom tokens without an Index canister, it should also be possible to hide them. However, this is not yet the case, as the hide function currently requires an Index canister to be provided.

# Changes

- Handle nullable Index canister ID in "Hide" component.
- Remove unused label key.

# Tests

Manually tested.


AntonioVentilii (approved):
LGTM tks!

PR #3546, opened 2024-11-14
feat(devops): Manually deploy breaking changes via CI

bitdivine:
# Motivation
If an API changes, dfx will typically ask the user whether they really want to deploy.  In CI, we need a way of saying "yes", but probably don't want this to be the default.

There is an existing checkbox for forcing the backend deployment to take place, no matter what.  It seems reasonable to use `--yes` when that is selected, so we do not need a new input.

# Changes
- If `force-backend` is selected in the CI deployment task, deploy with `--yes` and `--upgrade-unchanged`.

# Tests
Executed the action by:
- Disabling the check that allows deploying only from main.
- Running the action with: `gh workflow run deploy-to-environment.yml --ref "yes" --field network=staging --field canister=backend --field force-backend=true`
- Re-enabling the main check.

### DIFF
--- a/.github/workflows/deploy-to-environment.yml
+++ b/.github/workflows/deploy-to-environment.yml
@@ -25,11 +25,6 @@ on:
         type: boolean
         description: 'Force backend deployment even if no changes are detected.'
         default: false
-      backend-yes:
-        required: false
-        type: boolean
-        description: 'Allow breaking changes and other blockers to backend deployment'
-        default: false
 
 run-name: >-
   ${{ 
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-to-environment.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         if: ${{ github.ref != 'refs/heads/main' }}
         run: |
           echo "This workflow can only be manually triggered with workflow_dispatch on the main branch"
-          exit 1
+          # exit 1
```
bitdivine, line 48: Disabled for testing.  I actually believe that it should be possible to push any ref, not just head of main.  E.g. it should be possible to deploy a tagged release candidate to beta even if the tag is no longer at head of main.  But that is the subject of another PR.

scripts/deploy.backend.sh
```diff
@@ -60,7 +60,7 @@ if [ -n "${ENV+1}" ]; then
          };
          ic_root_key_der = $ic_root_key_der;
      }
-  })" --network "$ENV"
+  })" --network "$ENV" "${@}"
```
bitdivine, line 63: This passes cli flags through to `dfx`.

.github/workflows/deploy-to-environment.yml
```diff
@@ -154,14 +154,17 @@ jobs:
 
       - name: Deploy Backend to Environment
         run: |
+          DFX_DEPLOY_FLAGS=() # Arguments to be added to the dfx deploy command.
+          [[ "${{ github.event.inputs.force-backend }}" != "true" ]] || DFX_DEPLOY_FLAGS+=('--yes' '--upgrade-unchanged') # Corresponds to: dfx deploy --yes --upgrade-unchanged
```
peterpeterparker, line 158: `!= "true"` is correct here? it's not the contrary?

bitdivine: It is correct.  If true, more flags are added.  If undefined, false or any other value, no flags are added.

peterpeterparker: Gotcha.

peterpeterparker (approved):
LGTM, thx

PR #3547, opened 2024-11-14
feat(frontend): add Ethereum transactions to unified list

AntonioVentilii:
# Motivation

Following PR #3458 , we include Ethereum transactions in the unified list. It simply uses the same logic of the BTC transactions.

# Note

A few notes:
- For now, we include Sepolia, because it is easy for testing; I included a TODO to filter after the feature is complete.
- We do not test `mapEthTransactionUi` (it is tested elsewhere), so we provide undefined props for it. This function is used mostly for defining the ID and the transaction type, that is not really relevant for the tests of this PR.
- We still don't treat sorting and grouping in this PR, we are just focusing on unifing.

# Changes

- Create util to verify Sepolia networks.
- Add ETH transactions store in mapping.

# Tests

I included more tests, both for the util and the component.


peterpeterparker (commented):
Can you split the PR in two? Like the UI and the utils separated

AntonioVentilii:
@peterpeterparker so, first the PR:

- #3557 

Then this one will be just for the util. I will provide the UI in another one

AntonioVentilii:
@peterpeterparker ready for another review (the linked PR was merged)

src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
```diff
@@ -9,9 +9,13 @@
...
 
 	let transactions: AllTransactionsUi;
+	// TODO: add ethTransactions to mapAllTransactionsUi
 	$: transactions = mapAllTransactionsUi({
 		tokens: $enabledTokens,
-		$btcTransactions: $btcTransactionsStore
+		$btcTransactions: $btcTransactionsStore,
+		$ethTransactions: {},
+		$ckEthMinterInfo: {},
+		$ethAddress: undefined
```
peterpeterparker, line 18: undefined? not needed?

src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
```diff
@@ -9,9 +9,13 @@
 	import { mapAllTransactionsUi } from '$lib/utils/transactions.utils';
 
 	let transactions: AllTransactionsUi;
+	// TODO: add ethTransactions to mapAllTransactionsUi
```
peterpeterparker, line 12: and `ckEthMinterInfo` I guess?

Speaking of: are we sure? If I'm not wrong those info are currently not loaded on the tokens page, this will add some more stress.

src/frontend/src/lib/utils/transactions.utils.ts
```diff
@@ -40,8 +58,24 @@ export const mapAllTransactionsUi = ({
...
-			return acc;
+			// TODO: remove Sepolia transactions when the feature is complete; for now we use it for testing
+			const isSepoliaNetwork = isNetworkIdSepolia(networkId);
+
+			return [
+				...acc,
+				...($ethTransactions[tokenId] ?? []).map((transaction) => ({
+					...mapEthTransactionUi({
+						transaction,
+						ckMinterInfoAddresses: toCkMinterInfoAddresses({
```
peterpeterparker, line 69: I'm not sure how "heavy" is but, maybe interesting to compute it only once and not within the reduce?

src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
AntonioVentilii: it is just a placeholder, since the UI part will come in another PR, as discussed

src/frontend/src/lib/utils/transactions.utils.ts
AntonioVentilii: it varies from token to token, but you mean to get the one for mainnet and the one for testnet, and then assign each one, depending on the token? so that we call the function `toCkMinterInfoAddresses` only once?

peterpeterparker: Even if `toCkMinterInfoAddresses` appears to actually be trivial, I would maybe still do so, WDYT?

src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
AntonioVentilii: corrected!

> are we sure?

Good catch! I thought it was among the `Loaders`, but indeed it is specific to Ethereum transaction page (via the `ConvertToCkETH` button) or in IC transactions page (via `ConvertToEth` button and component `IcTransactionsCkEthereumListeners`). I will add a listener in another PR, then, good for you?

src/frontend/src/lib/utils/transactions.utils.ts
AntonioVentilii: yes agreed, have a look at last commit, I had to use mutable variables.... not really a fan of it

Obviously i can just calculate it upfront, but it seemed correct to calculate it only if and when we have the need

AntonioVentilii: or do you prefer immutable?

src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
peterpeterparker: ok

src/frontend/src/lib/utils/transactions.utils.ts
```diff
@@ -40,8 +61,36 @@ export const mapAllTransactionsUi = ({
 		}
 
 		if (isNetworkIdEthereum(networkId)) {
-			// TODO: Implement Ethereum transactions
-			return acc;
+			// TODO: remove Sepolia transactions when the feature is complete; for now we use it for testing
```
peterpeterparker, line 64: Can you move those initialization outside of the reduce? It's what I meant but, maybe that's not possible?

AntonioVentilii: ah yes, i can, but only of two ways:

1) we initialize always, independently of having Eth/Sepolia tokens

OR

2) I initialize only if we have Eth/Sepolia tokens outside the reduce, but that means that i have to "scan" the token list twice, with `find` first and then in reduce

Which one is preferable? I am more on the first one eheh

peterpeterparker: I tend for the 1., not the 2. 
But if you rather like to keep it in the reduce after all, I can also understand now.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/utils/transactions.utils.ts
AntonioVentilii: changed to do the 1! I prefer it too

AntonioVentilii:
[OISY-130]

PR #3558, opened 2024-11-14
feat(frontend): add additional measure to not load multiple times the same token transactions

AntonioVentilii:
# Motivation

As @peterpeterparker , we could maybe cross the limit of requests we can do to third-parties APIs. So, to try and reduce this number, we try to avoid reloading transactions for the tokens that were already loaded, in `LoaderMultipleEthTransaction`.

Just to clarify, the use-case is when the list changes, not when the list stays the same: we want to load the transactions only for new tokens.

# Note

I tried to make it more functional  as possible, but I couldn't reach a good solution... any suggestion welcomed.

# Tests

Added some new test.


peterpeterparker (approved):
I would not be against something more functional but, works out. Thanks

AntonioVentilii:
[OISY-130]

PR #3559, opened 2024-11-14
feat(frontend): ConvertFee component

DenysKarmazynDFINITY:
# Motivation

A small wrapper around `ConvertAmountDisplay` to display provided `feeAmount` (or `zeroAmountLabel` if fee is zero and the label is there).

Small heads up: total fee will have a separate component, since it should be displayed differently.

P.S.: I haven't added unit tests for this one as the child component as well as all used here utils are tested separately.

Screenshot 2024-11-14 at 14 46 20

### DIFF
--- a/src/frontend/src/lib/components/convert/ConvertFee.svelte
+++ b/src/frontend/src/lib/components/convert/ConvertFee.svelte
@@ -1,0 +1,27 @@
+<script lang="ts">
+	import { nonNullish } from '@dfinity/utils';
+	import { BigNumber } from '@ethersproject/bignumber';
+	import ConvertAmountDisplay from '$lib/components/convert/ConvertAmountDisplay.svelte';
+	import { formatToken } from '$lib/utils/format.utils';
+
+	export let feeAmount: bigint | undefined = undefined;
+	export let symbol: string;
+	export let decimals: number;
+	export let exchangeRate: number | undefined = undefined;
+	export let zeroAmountLabel: string | undefined = undefined;
+
+	let formattedFeeAmount: number | undefined;
+	$: formattedFeeAmount = nonNullish(feeAmount)
+		? Number(
+				formatToken({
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/convert/ConvertFee.svelte
```diff
@@ -0,0 +1,27 @@
...
+	export let feeAmount: bigint | undefined = undefined;
+	export let symbol: string;
+	export let decimals: number;
+	export let exchangeRate: number | undefined = undefined;
+	export let zeroAmountLabel: string | undefined = undefined;
+
+	let formattedFeeAmount: number | undefined;
+	$: formattedFeeAmount = nonNullish(feeAmount)
+		? Number(
+				formatToken({
```
AntonioVentilii, line 16: can formatToken be used inside Number? I really don't know, I imagine you tested and it went all good?

(maybe a short unittest for this component, just for this part of logic?)

AntonioVentilii: If you are sure that i works (tested manually) and are tight with time, maybe add a TODO?

Or even better, since it seems you are using `Number(formatToken(...))` more times, maybe a dedicated function with its own test (in another future PR)?

DenysKarmazynDFINITY: Yep, all good, it can be used like this. This pattern is already in the codebase, e.g. `exhange.utils -> usdValue`. Regarding a util - yes, a good idea, I am indeed using it in a plenty of files across the conversion flow. Add todo comment for now.

AntonioVentilii: Tks for checking!

AntonioVentilii (approved):
LGTM tks

PR #3562, opened 2024-11-14
feat(frontend): util to normalize a timestamp to seconds

AntonioVentilii:
# Motivation

Since we are going to re-use the logic, we create a small util that normalize any milliseconds or nanoseconds timestamp to seconds.


peterpeterparker (approved):
LGTM, thx

PR #3563, opened 2024-11-14
test(frontend): Test service requestPouhCredential

lmuntaner:
# Motivation

We want to change the code in the `requestPouhCredential` service.

Before we do that, we add a test with the current functionality.

# Changes

* Add two test scenarios for service `requestPouhCredential`.

# Tests

Only test changes.


lmuntaner:

vitest.setup.ts
```diff
@@ -8,6 +8,8 @@ vi.mock('$app/stores', () => ({
 	page: mockPage
 }));
 
+vi.stubEnv('VITE_LOCAL_POUH_ISSUER_CANISTER_ID', 'qbw6f-caaaa-aaaah-qdcwa-cai');
```
lmuntaner, line 11: There is a race condition between importing the file that uses the env vars and stubbing the env vars.

To make sure that the env vars are stubbed we have two options:
* Stub in the test specifically and import everything dynamically, also breaking some mocks that depend on general import.
* Stub in the setup file (as is now). The disadvantage is that I can't change it for different tests scenarios. But I don't need to change this env var for different scenarios.

src/frontend/src/tests/lib/services/request-pouh-credential.services.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+				profile: userProfileWithCredential
+			});
+		});
+
+		it('should not store the credential if validating credential fails', async () => {
+			addUserCredentialMock.mockResolvedValueOnce({ Err: { InvalidCredential: null } });
+
+			const result = await requestPouhCredential({ identity });
+
+			expect(result.success).toBe(true);
```
peterpeterparker, line 98: success = true is correct here? The backend throws `InvalidCredential`.

vitest.setup.ts
peterpeterparker: Do we actually need to stub the env var? Can we specify it in the `.env.test`, would that work too?

src/frontend/src/tests/lib/services/request-pouh-credential.services.spec.ts
lmuntaner: It's a matter of meaning. In this case, the request succeeded, the validation didn't though. Already a toast is shown with the validation error. But I can change that if you think that makes more sense.

vitest.setup.ts
lmuntaner: let me try

lmuntaner: Nice, it worked!

lmuntaner:

src/frontend/src/tests/lib/services/request-pouh-credential.services.spec.ts
peterpeterparker: Thanks for the explanation!

vitest.setup.ts
peterpeterparker: Cool!

peterpeterparker (approved):
LGTM, thx

PR #3565, opened 2024-11-14
feat(frontend): util to sort unified transaction list

AntonioVentilii:
# Motivation

I thought a lot about what would be the best way to reduce loops for the unified transaction list, but the only flow that I was able to reach and respect the requirements is:

- First loop --> unify the transactions looping the enabled tokens
- Second loop --> sort the transactions
- Third loop --> group by date (that gives automatically the correct sorting)

I verified if performance was better when the sorting would happen after, for dates and for each date group, but it was not better.

Furthermore there would be another method, but it would create the date groups directly when rendering inside the component. We yield the "latest date" and every-time it changes, we add a title with the new date, so we don't need to group anymore, just an if condition inside the component rendering. However, I think that is too much logic in the component.

So, for now we still create another util to sort the transactions, only by timestamp. The ones without timestamp will be put automatically on top (pending).

We will use it in the UI in another PR.

# Tests

Added tests.


peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3567, opened 2024-11-14
fix(frontend): avoid concurrent loading with fast reloading of LoaderMultipleEthTransactions

AntonioVentilii:
# Motivation

# Changes

# Tests


AntonioVentilii:
This branch was a test to check if this solution would work to avoid concurrency in the calls. But it didn't

Issue #3480, opened 2024-11-12
[BUG] Bitcoin sending doesn't work


DeLemonCH:
Several bugs:

1. When still in status confirmation, then there is a strange error message when calculating the fee (see first picture)
2. Max auto-input button/click of amount is not working
3. Sending is not working - ending with an error message (see second picture)
<img width="538" alt="btc_still_in_confirmation" src="https://github.com/user-attachments/assets/06112717-f8f2-47b2-b9f4-58b608f2271a">
<img width="621" alt="btc_sending" src="https://github.com/user-attachments/assets/97959149-1e56-4c7d-8853-217dbc809461">

DeLemonCH:
@AntonioVentilii any feedback regarding this - I think a small fix in the frontend is need for serialization

DeLemonCH:
Thank you!

StefanBerger-DFINITY:
Hi @DeLemonCH . We filed the remaining 2 issues, and will add them into one of our next sprints.
Thanks again for filing these items!

DeLemonCH:
Hi guys, I can't still not get mt Bitcoins out - now because another issue with the funds ...
Thanks for checking!
<img width="617" alt="image" src="https://github.com/user-attachments/assets/1d660c64-924b-405e-be22-d579bd596951">

DeLemonCH:
[StefanBerger-DFINITY](https://github.com/StefanBerger-DFINITY) Merci!

StefanBerger-DFINITY:
@DeLemonCH will address that asap. @DenysKarmazynDFINITY  that should be in your todo list in the next few minutes

StefanBerger-DFINITY:
The diff is 0.00002066, which is about the transaction fees of ~2 USD.
So this is still the MAX problem I assume.
(You could try to send 0.001097544914 instead of the 0.00111783, that should then end up to 0 with the fees..)

StefanBerger-DFINITY:
@DeLemonCH sorry to bother again.
The root problem here is of course that the fees must be paid in btc, and they are deducted from the wallet account and not the transfer amount (paid by sender). We currently don't have the fee information at this step, so we can't offer a MAX button there.

But we are working on an improved flow that will show the expected fees, and then we will also provide a MAX button.

DeLemonCH:
Hi Stefan, yes I could send it with considering the fee in the full amount - though would have expected it to be taken from the transfer amount.
Thanks anyway!

PR #3569, opened 2024-11-14
fix(frontend): use jsonReplacer to stringify bigints

DenysKarmazynDFINITY:
# Motivation

The PR fixes one of the issues reported in https://github.com/dfinity/oisy-wallet/issues/3480. Also, I noticed a few more places where the same kinda of error could have appeared.


AntonioVentilii (approved):
Thank you for the quick fix and the catch! 

P.S. maybe at some point we should extract it as separate function

DenysKarmazynDFINITY:
@AntonioVentilii definitely, a util something like `safeJsonStringify` would be useful for this kind of cases.

peterpeterparker (commented):
Thx!

PR #3575, opened 2024-11-15
feat(backend): Add derivation_origin to backend config

lmuntaner:
# Motivation

We want to validate the new field "derivationOrigin" in the id alias credential.

Before we can do that, the backend needs to know the derivation origin.

This PR, adds the derivation origin to the configuration parameter of the backend.

# Changes

## Manual Changes

* Add a new field `derivation_origin` in InitArg and Config struts.
* Add new field in `deploy.backend.sh` script.

## Automatic Changes

* `.did` and related file changes after running `npm run generate`.

# Tests

* Add a `derivation_origin` in the pocket ic init args of the backend canister.

### DIFF
--- a/scripts/deploy.backend.sh
+++ b/scripts/deploy.backend.sh
@@ -53,7 +53,7 @@ if [ -n "${ENV+1}" ]; then
          ecdsa_key_name = \"$ECDSA_KEY_NAME\";
          allowed_callers = vec {};
          cfs_canister_id = opt principal \"$SIGNER_CANISTER_ID\";
-         derivation_origin = \"$DERIVATION_ORIGIN\";
+         derivation_origin = opt \"$DERIVATION_ORIGIN\";
          supported_credentials = opt vec {
             record {
               credential_type = variant { ProofOfUniqueness };
@@ -73,7 +73,7 @@ else
          ecdsa_key_name = \"$ECDSA_KEY_NAME\";
          allowed_callers = vec {};
          cfs_canister_id = opt principal \"$SIGNER_CANISTER_ID\";
-         derivation_origin = \"$DERIVATION_ORIGIN\";
+         derivation_origin = opt \"$DERIVATION_ORIGIN\";
          supported_credentials = opt vec {
             record {
               credential_type = variant { ProofOfUniqueness };
--- a/scripts/deploy.backend.sh
+++ b/scripts/deploy.backend.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 II_CANISTER_ID="$(dfx canister id internet_identity --network "${ENV:-local}")"
+BACKEND_CANISTER_ID="$(dfx canister id backend --network "${ENV:-local}")"
 POUH_ISSUER_CANISTER_ID="$(dfx canister id pouh_issuer --network "${ENV:-local}")"
 SIGNER_CANISTER_ID="$(dfx canister id signer --network "${ENV:-local}")"
 
@@ -12,6 +13,7 @@ case $ENV in
   # URL used by issuer in the issued verifiable credentials (typically hard-coded)
   # Represents more an ID than a URL
   POUH_ISSUER_VC_URL="https://${POUH_ISSUER_CANISTER_ID}.icp0.io/"
+  DERIVATION_ORIGIN="https://tewsx-xaaaa-aaaad-aadia-cai.icp0.io"
   ;;
 "ic")
   ECDSA_KEY_NAME="key_1"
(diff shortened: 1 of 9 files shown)


bitdivine (approved):
LGTM.  Thanks!

scripts/deploy.backend.sh
```diff
@@ -20,6 +22,7 @@ case $ENV in
   # URL used by issuer in the issued verifiable credentials (tipically hard-coded)
   # Represents more an ID than a URL
   POUH_ISSUER_VC_URL="https://id.decideai.xyz/"
+  DERIVATION_ORIGIN="https://oisy.com"
```
bitdivine, line 25: Note: Is there a reason to use a new variable, not e.g. `VITE_OISY_URL=https://oisy.com`  or `VITE_AUTH_DERIVATION_ORIGIN=https://tewsx-xaaaa-aaaad-aadia-cai.icp0.io` (used for II logins)?

If a different variable is used, will it need to be added to 1password & .github/workflows/deploy-to-environment.yml ?

If the new variable is usually the same as say the II derivation origin, would it make sense to default to using that if not specfied?

peterpeterparker (commented):
- Generally speaking, should we use a more specific name—for example, by adding a suffix or prefix to`derivation_origin` and `DERIVATION_ORIGIN`? If I understand correctly, there are two types of derivation origin: one for authentication sign-in and one for POUH. This could lead to confusion about which one is being referred to.

- If we reinstall the backend canister, won’t we lose the custom tokens and other user metadata? If so, how do we plan to set this new value?

scripts/deploy.backend.sh
lmuntaner: Yes, it should be the same derivation origins as in the frontend. However, in the frontend, the frontend origin might not be set, which reverts to using the window location origin. But in the backend I need this env var always set.

Does the backend have access to the same env vars as the frontend?

lmuntaner:
> * Generally speaking, should we use a more specific name—for example, by adding a suffix or prefix to`derivation_origin` and `DERIVATION_ORIGIN`? If I understand correctly, there are two types of derivation origin: one for authentication sign-in and one for POUH. This could lead to confusion about which one is being referred to.

They are the same for signing in and for the POUH.

> * If we reinstall the backend canister, won’t we lose the custom tokens and other user metadata? If so, how do we plan to set this new value?

Why would we lose the custom tokens? This is an optional new parameter in the config. I believe I even saw some tests checking that the custom tokens are persisted across updates.

peterpeterparker:
> They are the same for signing in and for the POUH.

Gotcha 👍

> This is an optional new parameter in the config

What I meant is how are you going to initialize the config?

scripts/deploy.backend.sh
bitdivine: Normally the backend build doesn't care about the frontend env vars but for sure we can make it care, either with a built in variable or by providing this as as install arg (nicer, as then we still have the same Wasm everywhere).  Do you want help with this or maybe sync with @inc-man who takes care of deployments to staging, beta and soon also production?

lmuntaner:
> What I meant is how are you going to initialize the config on staging or production?

By passing a different derivation origin when installing the canister. Can't Orbit do that? Aren't the parameters part of the proposal?

scripts/deploy.backend.sh
lmuntaner: This PR provides it as install arg, isn't that enough?

Let me sync with @inc-man !

peterpeterparker:
> By passing a different derivation origin when installing the canister. Can't Orbit do that? Aren't the parameters part of the proposal?

Isn't the `post_init` lifecycle function triggered only when you install the canister, not when you upgrade it? If that's correct, my question is: are you planning to re-install the canister to set those new parameters? And if so, won't we lose the state?

Note that this isn't a blocker for the implementation. I think it's correct to have those as parameters—I just want to understand how we are going to handle upgrades in production with these changes and how we are going to configure the existing canister.

lmuntaner:
> Isn't the post_init lifecycle function triggered only when you install the canister

Do you mean the `post_upgrade`? The `post_upgrade` is called when upgrading a canister. I agree that we don't want to reinstall the wasm and lose all state.

Where do you see `post_init`?

peterpeterparker:
> Where do you see post_init?

Typo, I meant `init`, sorry for that.

> I agree that we don't want to reinstall the wasm and lose all state.

Great. So, we gonna add a public API endpoint to set the value to the existing production canister?

Note again, code wise it's all good, really just asking about how to move forward. Feel free to merge.

lmuntaner:
> Great. So, we gonna add a public API endpoint to set the value to the existing production canister?

No, this is done through upgrade which is needed to also store the new config. It uses the [post_upgrade hook already present](https://github.com/dfinity/oisy-wallet/blob/d9dae50a1a5e8bb023574543bf6b7f3f7d49452f/src/backend/src/lib.rs#L189).

Does this make sense? I can explain in a call if necessary.

peterpeterparker:
> Does this make sense? I can explain in a call if necessary.

I did not knew the init arguments were passed to the post upgrade. All clear. Thanks.

inc-man (approved):
LGTM

PR #3576, opened 2024-11-15
feat(frontend): component to show group of transactions by date

AntonioVentilii:
# Motivation

In this PR we introduce a component that will show a date with its transactions, as preparation for the unified transaction list.

Screenshot 2024-11-15 at 10 06 37
Screenshot 2024-11-15 at 10 06 46
Screenshot 2024-11-15 at 10 06 56
Screenshot 2024-11-15 at 10 07 04

PR #3577, opened 2024-11-15
feat(frontend): use component to group of dates in AllTransactionsList

AntonioVentilii:
# Motivation

After PR #3576 , we include the group component into the main transaction list.

Screenshot 2024-11-15 at 10 06 37
Screenshot 2024-11-15 at 10 06 46
Screenshot 2024-11-15 at 10 06 56
Screenshot 2024-11-15 at 10 07 04


src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
```diff
@@ -58,10 +65,8 @@
 
 <!--TODO: include skeleton for loading transactions and remove nullish checks-->
 {#if nonNullish(sortedTransactions) && sortedTransactions.length > 0}
-	{#each transactions as transaction, index (`${transaction.id}-${index}`)}
-		<div in:slide={SLIDE_DURATION}>
-			<svelte:component this={transaction.component} {transaction} token={transaction.token} />
-		</div>
+	{#each Object.entries(groupedTransactions) as [date, transactions], index (`${date}-${index}`)}
```
DenysKarmazynDFINITY, line 68: I think with grouping by date, we can be sure that `date` is unique, otherwise it may indicate that "something went wrong".

src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
```diff
@@ -58,10 +65,8 @@
 
 <!--TODO: include skeleton for loading transactions and remove nullish checks-->
 {#if nonNullish(sortedTransactions) && sortedTransactions.length > 0}
```
DenysKarmazynDFINITY, line 67: Maybe let's update `groupedTransaction` to be `undefined` if no `sortedTransactions` available, and then update this line to check `groupedTransaction` instead?

src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
AntonioVentilii: so, the function per se guarantees it, since it returns a `Record`

but if you prefer I can put a pre-check, in the component directly, some sort or guard, not really sure which

src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
AntonioVentilii: good catch! done!

src/frontend/src/lib/components/transactions/AllTransactionsList.svelte
DenysKarmazynDFINITY: Yeah, that's exactly what I meant - since `date` is unique, we don't need `index` here. Not a must though, if you still want to keep it

AntonioVentilii: ah! understood now! good one! yes, removed!

AntonioVentilii:
[OISY-130]

PR #3578, opened 2024-11-15
feat(frontend): formatTokenBigintToNumber util

DenysKarmazynDFINITY:
# Motivation

Before I start introducing new places with the similar formatting, creating the util that we discussed with @AntonioVentilii .


AntonioVentilii (approved):
Thank you! LGTM!

PR #3570, opened 2024-11-14
feat(frontend): add token prop to transaction card components

AntonioVentilii:
# Motivation

After PR #3537 , we can provide the token to the Transaction card components, to show the Token logo in the unified transaction list.

# Changes

- Make prop `token` for `EthTransaction` and adapt the usage, removing the derived store and providing the input from its parents, like `EthTransactions`.
- Map `token` for all the UI transactions.
- Adapt `AllTransactionsList`.
 
# Tests

Test adapted and here results.

Screenshot 2024-11-14 at 18 09 30

PR #3579, opened 2024-11-15
fix(frontend): do not show token logo in token transaction page

AntonioVentilii:
# Motivation (deprecated, see Edit)

I don't have a better idea: the token logo was being showed in token-specific transaction page, because it is given as input and it is required in components like `BtcTransaction` or `EthTransaction`.

However, that same component is needed for the unified transactions (example PR https://github.com/dfinity/oisy-wallet/pull/3570) with the token logo.

So, the only idea I had was to make it so by default the component show the token logo, but not under request (for example in the token-specific transaction page).

# Edit

After @peterpeterparker [suggestion](https://github.com/dfinity/oisy-wallet/pull/3579#discussion_r1844114080) , I changed the logic to provide a prop that defines the icon type (either `token` or `transaction`). In my opinion is a better solution.


src/frontend/src/lib/components/transactions/Transaction.svelte
```diff
@@ -31,7 +32,7 @@
 		<span class="inline-block first-letter:capitalize"><slot /></span>
 
 		<div slot="icon">
-			{#if nonNullish(token)}
+			{#if nonNullish(token) && !hideTokenLogo}
```
peterpeterparker, line 35: The token is mandatory now, I guess we can remove `nonNullish(token)`?

src/frontend/src/lib/components/transactions/Transaction.svelte
```diff
@@ -18,6 +18,7 @@
 	export let timestamp: number | undefined;
 	export let styleClass: string | undefined = undefined;
 	export let token: Token;
+	export let hideTokenLogo = false;
```
peterpeterparker, line 21: Another option is something like as following

```
export let badge: "token" | "logo" = "logo";
```

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/components/transactions/Transaction.svelte
AntonioVentilii: make sense, done!

src/frontend/src/lib/components/transactions/Transaction.svelte
AntonioVentilii: way better!!! thank you! changed to iconType! have a look!

AntonioVentilii:
@peterpeterparker even if you already approved, please re-review, since I followed [your suggestion](https://github.com/dfinity/oisy-wallet/pull/3579#discussion_r1844114080) and changed the logic

peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3580, opened 2024-11-15
refactor(frontend): remove token derived stores from Amount component

AntonioVentilii:
# Motivation

The component `Amount` was using a derived of `token` store. However, it is now used in a context where the token may change and it is not defined, for example, it is mounted on each transaction on the unified list.

So, I thought of giving the adequate properties instead of using the store. But that means that we need to propagate such properties.

I thought of creating a context too, but it will change for each Amount component mounted in the unified transaction list, so not a possible solution for now.

# Changes

- Add prop `decimals` to component `Amount`.
- Remove derived store dependency from `Amount`.
- Adapt component `Balance` that is using `Amount`.
- Adapt component `Transaction` that is using `Amount`.
- Adapt all transaction components for each network, providing at the highest level the token as token with fallback (that was what it was used before inside the `Amount` component, so no real change).
- We force the token to be non-optional for convenience, since in the future, we will be removing the `tokenWithFallback` store and we will need these components to have a defined token.

# Tests

No changes in local replica.

### DIFF
--- a/src/frontend/src/btc/components/transactions/BtcTransactions.svelte
+++ b/src/frontend/src/btc/components/transactions/BtcTransactions.svelte
@@ -13,6 +13,7 @@
 	import TransactionsSkeletons from '$lib/components/transactions/TransactionsSkeletons.svelte';
 	import { SLIDE_DURATION } from '$lib/constants/transition.constants';
 	import { modalBtcTransaction } from '$lib/derived/modal.derived';
+	import { tokenWithFallback } from '$lib/derived/token.derived';
 	import { modalStore } from '$lib/stores/modal.store';
 
 	let selectedTransaction: BtcTransactionUi | undefined;
@@ -26,7 +27,7 @@
 <TransactionsSkeletons loading={$btcTransactionsNotInitialized}>
 	{#each $sortedBtcTransactions as transaction (transaction.data.id)}
 		<div transition:slide={SLIDE_DURATION}>
-			<BtcTransaction transaction={transaction.data} />
+			<BtcTransaction transaction={transaction.data} token={$tokenWithFallback} />
 		</div>
 	{/each}
 
--- a/src/frontend/src/icp/components/transactions/IcTransactions.svelte
+++ b/src/frontend/src/icp/components/transactions/IcTransactions.svelte
@@ -30,6 +30,7 @@
 	import { WALLET_PAGINATION } from '$lib/constants/app.constants';
 	import { authIdentity } from '$lib/derived/auth.derived';
 	import { modalIcToken, modalIcTransaction } from '$lib/derived/modal.derived';
+	import { tokenWithFallback } from '$lib/derived/token.derived';
 	import { nullishSignOut } from '$lib/services/auth.services';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { modalStore } from '$lib/stores/modal.store';
@@ -117,7 +118,7 @@
 			<InfiniteScroll on:nnsIntersect={onIntersect} disabled={disableInfiniteScroll}>
 				{#each $icTransactions as transaction, index (`${transaction.data.id}-${index}`)}
 					<li in:slide={{ duration: transaction.data.status === 'pending' ? 250 : 0 }}>
-						<IcTransaction transaction={transaction.data} />
+						<IcTransaction transaction={transaction.data} token={$tokenWithFallback} />
 					</li>
 				{/each}
 			</InfiniteScroll>
--- a/src/frontend/src/lib/components/ui/Amount.svelte
+++ b/src/frontend/src/lib/components/ui/Amount.svelte
@@ -1,22 +1,22 @@
 <script lang="ts">
 	import type { BigNumber } from '@ethersproject/bignumber';
 	import { EIGHT_DECIMALS } from '$lib/constants/app.constants';
-	import { tokenWithFallback } from '$lib/derived/token.derived';
 	import { formatToken } from '$lib/utils/format.utils';
 
 	export let amount: BigNumber;
+	export let decimals: number;
 
 	let detailedValue: string;
 	$: detailedValue = formatToken({
 		value: amount,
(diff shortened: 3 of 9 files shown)


src/frontend/src/btc/components/transactions/BtcTransactions.svelte
```diff
@@ -26,7 +27,7 @@
 <TransactionsSkeletons loading={$btcTransactionsNotInitialized}>
 	{#each $sortedBtcTransactions as transaction (transaction.data.id)}
 		<div transition:slide={SLIDE_DURATION}>
-			<BtcTransaction transaction={transaction.data} />
+			<BtcTransaction transaction={transaction.data} token={$tokenWithFallback} />
```
peterpeterparker, line 30: Mmm the fallback is Ethereum, is this a good idea or should we assert that the token is given and BTC?

src/frontend/src/icp/components/transactions/IcTransactions.svelte
```diff
@@ -117,7 +118,7 @@
 			<InfiniteScroll on:nnsIntersect={onIntersect} disabled={disableInfiniteScroll}>
 				{#each $icTransactions as transaction, index (`${transaction.data.id}-${index}`)}
 					<li in:slide={{ duration: transaction.data.status === 'pending' ? 250 : 0 }}>
-						<IcTransaction transaction={transaction.data} />
+						<IcTransaction transaction={transaction.data} token={$tokenWithFallback} />
```
peterpeterparker, line 121: Same as for Btc

src/frontend/src/lib/components/ui/Amount.svelte
```diff
@@ -1,22 +1,22 @@
 <script lang="ts">
 	import type { BigNumber } from '@ethersproject/bignumber';
 	import { EIGHT_DECIMALS } from '$lib/constants/app.constants';
-	import { tokenWithFallback } from '$lib/derived/token.derived';
 	import { formatToken } from '$lib/utils/format.utils';
 
 	export let amount: BigNumber;
+	export let decimals: number;
```
peterpeterparker, line 7: I realize that we can have an amount for a token x with a number of decimals of a token y. Not. for this PR and not sure neither what we can do, just thinking out at loud.

AntonioVentilii: indeed! but it is difficult to link directly amount with a token, meaning, either an enforced type, but it could always fail, no?

src/frontend/src/btc/components/transactions/BtcTransactions.svelte
AntonioVentilii: you mean `$token ?? BTC_MAINNET_TOKEN` , instead of `$tokenWithFallback`? if I understood correctly, then yes, totally agreed!

I was just trying to respect the old logic, that is flawed too...

peterpeterparker: Great. Yes, either your suggestion or not showing the transaction unless it's bitcoin. One or the other yes.

src/frontend/src/lib/components/ui/Amount.svelte
peterpeterparker: Agree, not sure how, I was just thinking out at loud.

src/frontend/src/btc/components/transactions/BtcTransactions.svelte
AntonioVentilii: done!

src/frontend/src/icp/components/transactions/IcTransactions.svelte
AntonioVentilii: same as above, but with fallback ICP token

src/frontend/src/env/tokens.btc.env.ts
```diff
@@ -51,3 +52,9 @@ export const BTC_REGTEST_TOKEN: Token = {
 	decimals: BTC_DECIMALS,
 	icon: bitcoinTestnet
 };
+
+export const SUPPORTED_BITCOIN_TOKENS: [...Token[], Token] = [
```
peterpeterparker, line 56: Maybe add a comment (or a test) that the order matters here.

src/frontend/src/icp/components/transactions/IcTransactions.svelte
```diff
@@ -117,7 +118,7 @@
 			<InfiniteScroll on:nnsIntersect={onIntersect} disabled={disableInfiniteScroll}>
 				{#each $icTransactions as transaction, index (`${transaction.data.id}-${index}`)}
 					<li in:slide={{ duration: transaction.data.status === 'pending' ? 250 : 0 }}>
-						<IcTransaction transaction={transaction.data} />
+						<IcTransaction transaction={transaction.data} token={$token ?? ICP_TOKEN} />
```
peterpeterparker, line 121: ckErc20 is 6 or 18 decimals 🫠

AntonioVentilii: ah! argh! mmm thinking.... not really sure what to choose... render only when `token` is defined? either way, before it was Ethereum for any ICRC eheh!

What do you suggest?

src/frontend/src/env/tokens.btc.env.ts
AntonioVentilii: done!

src/frontend/src/icp/components/transactions/IcTransactions.svelte
peterpeterparker: Let's keep ICP in this PR and think about it. Rendering when token is defined does not solve it neither when I think about it too long because then, it's also what we discussed earlier, the decimals of the token might not match the amount.

peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3585, opened 2024-11-15
feat(frontend): validateConvertAmount util

DenysKarmazynDFINITY:
# Motivation

The util that will be used in `ConvertAmountSource` as `customValidate` function for the input. I extracted it from the component for the unit testing purposes.

### DIFF
--- a/src/frontend/src/lib/utils/convert.utils.ts
+++ b/src/frontend/src/lib/utils/convert.utils.ts
@@ -1,10 +1,11 @@
 import { ZERO } from '$lib/constants/app.constants';
 import type { ConvertAmountErrorType } from '$lib/types/convert';
 import { formatToken } from '$lib/utils/format.utils';
-import { isNullish, nonNullish } from '@dfinity/utils';
+import { nonNullish } from '@dfinity/utils';
 import { BigNumber } from '@ethersproject/bignumber';
 import { Utils } from 'alchemy-sdk';
 
+// TODO: standardise and re-use this util between Send and Convert flows
 export const validateConvertAmount = ({
 	userAmount,
 	decimals,
@@ -16,9 +17,6 @@ export const validateConvertAmount = ({
 	balance?: BigNumber;
 	totalFee?: bigint;
 }): ConvertAmountErrorType => {
-	if (isNullish(totalFee)) {
-		return;
-	}
 	// We should align balance and userAmount to avoid issues caused by comparing formatted and unformatted BN
 	const parsedSendBalance = nonNullish(balance)
 		? Utils.parseUnits(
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/utils/convert.utils.ts
```diff
@@ -0,0 +1,41 @@
...
+	balance,
+	totalFee
+}: {
+	userAmount: BigNumber;
+	decimals: number;
+	balance?: BigNumber;
+	totalFee?: bigint;
+}): ConvertAmountErrorType => {
+	if (isNullish(totalFee)) {
+		return;
```
AntonioVentilii, line 20: I saw that in the code we do the same elsewhere, but I don't remember why? is it possible that the validation comes in another component when the fee is nullish?

src/frontend/src/lib/utils/convert.utils.ts
```diff
@@ -0,0 +1,41 @@
...
+				}),
+				decimals
+			)
+		: ZERO;
+
+	if (userAmount.gt(parsedSendBalance)) {
+		return 'insufficient-funds';
+	}
+
+	if (nonNullish(totalFee) && userAmount.add(totalFee).gt(parsedSendBalance)) {
```
AntonioVentilii, line 38: theoretically here the `nonNullish` function is not needed, right? you already checked above

AntonioVentilii (commented):
In component `SendAmount` we have a similar logic: is there any way that the two functions can be merged? Meaning, can we use your function there too?

src/frontend/src/lib/utils/convert.utils.ts
DenysKarmazynDFINITY: ah, good catch, removing it!

DenysKarmazynDFINITY:
@AntonioVentilii they are a few differences between them (mostly, in return types - one uses `strings`, and another one `Error`), but my idea was that they indeed can be standardised and reused in the future. I added a TODO comment for now.

src/frontend/src/lib/utils/convert.utils.ts
DenysKarmazynDFINITY: That's a good point. After thinking about it, I think it makes sense to validate `insufficient-funds` case even if `totalFee` is nullish. Therefore, I removed this check and adjusted the tests.

src/frontend/src/lib/utils/convert.utils.ts
DenysKarmazynDFINITY: Doch - the check stays because of the comments above :)

DenysKarmazynDFINITY:
@AntonioVentilii ready for another look

AntonioVentilii (approved):
LGTM tks!

PR #3586, opened 2024-11-15
feat(frontend): ConvertAmountSource component

DenysKarmazynDFINITY:
# Motivation

Component for handling source token amount input and all related information appearance.

Desktop:
Screenshot 2024-11-15 at 13 49 33

Mobile:
Screenshot 2024-11-15 at 13 49 41

### DIFF
--- a/src/frontend/src/lib/components/convert/ConvertAmountSource.svelte
+++ b/src/frontend/src/lib/components/convert/ConvertAmountSource.svelte
@@ -1,0 +1,95 @@
+<script lang="ts">
+	import { isNullish, nonNullish } from '@dfinity/utils';
+	import { BigNumber } from '@ethersproject/bignumber';
+	import { getContext } from 'svelte';
+	import { slide } from 'svelte/transition';
+	import ConvertInputAmount from '$lib/components/convert/ConvertInputAmount.svelte';
+	import ConvertInputsContainer from '$lib/components/convert/ConvertInputsContainer.svelte';
+	import ConvertToken from '$lib/components/convert/ConvertToken.svelte';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/convert/ConvertAmountSource.svelte
```diff
@@ -0,0 +1,95 @@
...
+		{:else}
+			<div transition:slide={SLIDE_DURATION}>
+				{formatUSD({ value: convertAmountUSD })}
+			</div>
+		{/if}
+	</div>
+
+	<button
+		slot="balance"
+		class={`rounded px-2 py-0.5 ${isZeroBalance ? 'bg-error-subtle-alt text-error' : 'bg-zumthor text-blue-ribbon'}`}
```
AntonioVentilii, line 84: i think that both `zumthor` and `blue-ribbon` can be replaced with the new colors

blue ribbon is text-brand-primary, and the other bg-brand-subtle (please double check)

src/frontend/src/lib/components/convert/ConvertAmountSource.svelte
```diff
@@ -0,0 +1,95 @@
...
+		on:click|preventDefault={setMax}
```
AntonioVentilii, line 85: do we need `preventDefault`? we do the same elsewhere?

src/frontend/src/lib/components/convert/ConvertAmountSource.svelte
DenysKarmazynDFINITY: good catch, updated!

AntonioVentilii (approved):
LGTM, tks!

PR #3590, opened 2024-11-15
fix(frontend): remove circular reference in tokens env file

AntonioVentilii:
# Motivation

We move `SUPPORTED_BITCOIN_TOKENS` since it was giving circular reference on initiation.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3592, opened 2024-11-15
fix(frontend): incorrect "value" type in InputCurrency

DenysKarmazynDFINITY:
# Motivation

While working on the conversion flow, I noticed one suspicious thing: the `value` prop that is controlled by `InputCurrency` is actually a `string` and not a `number` as it is expected to be according to the types in our codebase. It means that we incorrectly assume across the whole `send` flow that `amount` is `number | undefined` when it's in reality `string | undefined`. The same applies to the convert flow: `sourceAmount` was never a number. 

After reading [the GIX Input docs](https://gix.design/components/input), I found this paragraph: 
```
If the inputType is set to icp, the value bind by the component is a number. On the contrary, if bind to currency, the value is a string. This to avoid issue with scientific notation enforced by JavaScript. It is then up to you to parse the currency according your need, for example to bigint or BigNumber.
```

Having said that, we have quite an error-prompt situation: I noticed this issue while debugging one of the conversion services - `sendAmount === 0` was always giving me `false` after typing 0, even though TS-wise the statement looks correct. Since in the code all `send` components also expect it to be a `number`, this may lead to the same issue. 

Solution 1 (implemented in this PR): as suggested by GIX, I added an internal `inputValue` to the `InputCurrency` component which has `string | undefined` type, and on `nnsInput` I convert it to `number` before notifying parent. Also, provided some tests to test this behaviour. 

Solution 2: update `amount` type everywhere where it's used. Would require a lot of changes.

### DIFF
--- a/src/frontend/src/lib/components/ui/InputCurrency.svelte
+++ b/src/frontend/src/lib/components/ui/InputCurrency.svelte
@@ -1,5 +1,7 @@
 <script lang="ts">
 	import { Input } from '@dfinity/gix-components';
+	import { nonNullish } from '@dfinity/utils';
+	import { createEventDispatcher } from 'svelte';
 
 	export let value: number | undefined = undefined;
 	export let disabled: boolean | undefined = undefined;
@@ -8,20 +10,34 @@
 	export let required = true;
 	export let decimals: number | undefined = undefined;
 	export let testId: string | undefined = undefined;
+
+	const dispatch = createEventDispatcher();
+
+	let inputValue: string | undefined;
+	// Convert value to string if it comes from parent (e.g. initial state or set-to-max button)
+	$: inputValue = nonNullish(value) ? `${value}` : undefined;
+
+	const onInput = () => {
+		// Convert inputValue to number before updating parent's value
+		value = nonNullish(inputValue) ? Number(inputValue) : undefined;
+
+		// Bubble nnsInput as consumers might require the event as well
+		dispatch('nnsInput');
+	};
 </script>
 
 <Input
 	{name}
 	inputType="currency"
 	{required}
-	bind:value
+	bind:value={inputValue}
 	{decimals}
 	{placeholder}
 	spellcheck={false}
 	autocomplete="off"
 	{testId}
 	{disabled}
-	on:nnsInput
+	on:nnsInput={onInput}
 >
 	<slot name="inner-end" slot="inner-end" />
 </Input>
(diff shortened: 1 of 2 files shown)


DenysKarmazynDFINITY:
@peterpeterparker @AntonioVentilii please take a look at the PR's description and code.

src/frontend/src/tests/lib/components/ui/InputCurrency.spec.ts
```diff
@@ -0,0 +1,54 @@
+import InputCurrency from '$lib/components/ui/InputCurrency.svelte';
+import { assertNonNullish } from '@dfinity/utils';
+import { fireEvent, render } from '@testing-library/svelte';
+
+describe('ConvertAmountExchange', () => {
```
AntonioVentilii, line 5: I think here is InputCurrency, right?

src/frontend/src/lib/components/ui/InputCurrency.svelte
```diff
@@ -8,20 +10,34 @@
...
+
+	let inputValue: string | undefined;
+	// Convert value to string if it comes from parent (e.g. initial state or set-to-max button)
+	$: inputValue = nonNullish(value) ? `${value}` : undefined;
+
+	const onInput = () => {
+		// Convert inputValue to number before updating parent's value
+		value = nonNullish(inputValue) ? Number(inputValue) : undefined;
+
+		// Bubble nnsInput as consumers might require the event as well
```
AntonioVentilii, line 24: not really sure we need this comment: I think it is a common pattern in the code of Oisy

AntonioVentilii (commented):
It looks good, but thinking out loud: have you thought of using a context instead of binding the value? No preference, I am just asking if there is any worth in it or not

peterpeterparker:
Wow! Good catch, it's been there forever!

Actually, I rather like solution 2. Because properties can flow up and down, I feel like it's more maintanable if we can avoid having to manipulate it therefore, I gave it a quick try and, at least from a linter perspective, it doesn’t seem too bad to implement, as I managed to put it together quickly in #3609.

WDYT?

DenysKarmazynDFINITY:
@peterpeterparker yeah, solution 2 also work for me perfectly fine, thanks for preparing the PR! I

src/frontend/src/tests/lib/components/ui/InputCurrency.spec.ts
DenysKarmazynDFINITY: Totally right, but since we will most likely proceed with the Solution 2, this PR won't be merged.

peterpeterparker:
> solution 2 also work for me perfectly fine

Cool, then let's proceed with solution 2.

DenysKarmazynDFINITY:
Closing this PR.

PR #3593, opened 2024-11-15
fix(frontend): avoid concurrency of calls in Loader Transactions

AntonioVentilii:
# Motivation

Etherscan API has a limit of calls per seconds, depending on which plan we use. With the unified transaction we query all the tokens

# Changes

# Tests


src/frontend/src/eth/components/loaders/LoaderMultipleEthTransactions.svelte
```diff
@@ -1,9 +1,10 @@
 <script lang="ts">
+	import pLimit from 'p-limit';
```
peterpeterparker, line 2: What's this?

AntonioVentilii:
This PR was a draft to test a possible solution to the issue. It is not needed anymore

PR #3594, opened 2024-11-15
feat(frontend): bottleneck control of Etherescan API call limit

AntonioVentilii:
# WIP

# Motivation

# Changes

# Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
 		"@metamask/detect-provider": "^2.0.0",
 		"@walletconnect/web3wallet": "^1.14.0",
 		"alchemy-sdk": "^3.4.1",
+		"bottleneck": "^2.19.5",
 		"buffer": "^6.0.3",
 		"ethers": "^5.7.0",
 		"idb-keyval": "^6.2.1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -67,6 +67,7 @@
 		"@metamask/detect-provider": "^2.0.0",
 		"@walletconnect/web3wallet": "^1.14.0",
 		"alchemy-sdk": "^3.4.1",
+		"bottleneck": "^2.19.5",
```
peterpeterparker, line 70: What's this?

AntonioVentilii: still a draft, but this [package](https://www.npmjs.com/package/bottleneck) "controls" how many requests per second or concurrent are done to a specific function; it will help to "bottleneck" requests to Etherescan and avoid touching the limit of calls/second

### Context

I am trying several solutions to solve the issue of the Etherescan limit calls/second...

peterpeterparker: Let’s discuss Monday then. Adding any dependency introduces technical debt and potential security risks, especially when the library is not actively maintained, as is the case here. At the foundation—or at least within the Gix team—we generally avoid taking this route unless absolutely necessary. Additionally, I’d like to better understand the issue and explore the various solutions available.

AntonioVentilii: agreed!

AntonioVentilii:
This PR was a draft to test a possible solution to the issue. It is not needed anymore

PR #3600, opened 2024-11-15
refactor(frontend): simplify logic in LoaderMultipleEthTransactions

AntonioVentilii:
# Motivation

A very small simplification in `LoaderMultipleEthTransactions`.


peterpeterparker (approved):
Personally, as commented in previous PR, I would rather like a more functional approach for this but, if we have to keep the imperative implementation, I actually rather like the existing I have to say.

That said, it's an opiniated answer so not a reason to not approve the PR 😉

AntonioVentilii:
@peterpeterparker nono, i completely agree! still working on it...

AntonioVentilii:
[OISY-130]

PR #3603, opened 2024-11-15
refactor(frontend): optimize formatSecondsToNormalizedDate

AntonioVentilii:
# Motivation

As mentioned by @peterpeterparker in https://github.com/dfinity/oisy-wallet/pull/3526#discussion_r1841610934 , we could optimize function `formatSecondsToNormalizedDate` , avoiding creating a new formatter every time.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString)

> When the method is called many times with the same arguments, it is better to create a [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) object and use its [format()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/format) method, because a DateTimeFormat object remembers the arguments passed to it and may decide to cache a slice of the database, so future format calls can search for localization strings within a more constrained context.


src/frontend/src/tests/lib/utils/format.utils.spec.ts
AntonioVentilii, line 1: this one is just a wrap inside `describe('format.utils', () => {...`

nothing changed

src/frontend/src/icp/utils/date.utils.ts
```diff
@@ -11,3 +11,5 @@ export const normalizeTimestampToSeconds = (timestamp: number | bigint): number
 			? ts / Number(MILLISECONDS_IN_SECOND)
 			: ts;
 };
+
+export const relativeTimeFormatter = new Intl.RelativeTimeFormat('en', { numeric: 'auto' });
```
peterpeterparker, line 15: If not required, I would keep this formatter in `format.utils`. This way it remains scoped and we avoid one reference between two modules.

peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3604, opened 2024-11-15
feat(frontend): component NetworkLogo

AntonioVentilii:
# Motivation

We render a few times the logo for a network in the code. So this PR just extracts such component.


peterpeterparker (approved):
LGTM, thx

This PR give me an idea: we could have some sort of static page for E2E screenshots testing for all those "display purposes" component that accept tokens or networks as property. It would be easy to setup and can also be useful to compare with Figma when discussing with designers.

AntonioVentilii:
[OISY-130]

PR #3537, opened 2024-11-13
feat(frontend): add token logo to Transaction component

AntonioVentilii:
# Motivation

For the unified transaction list, we will require different icons for the Transaction card. We need to have the token logo and the transaction icon, to be more identifiable.

# Changes

- Provide an optional token prop to `Transaction`, that will define the type of icon.
- Change `TokenLogo` to accept another type of badge: transaction icon.
- Adapt the code.

# Tests

Added more test, but this is the manual result.

Screenshot 2024-11-13 at 21 09 13

PR #3605, opened 2024-11-15
refactor(frontend): remove custom class prop from RoundedIcon

AntonioVentilii:
# Motivation

The prop iconStyleClass seemed unnecessary for just opacity. So we substitute with a direct boolean prop (reference https://github.com/dfinity/oisy-wallet/pull/3537#discussion_r1842195829).

### DIFF
--- a/src/frontend/src/lib/components/ui/RoundedIcon.svelte
+++ b/src/frontend/src/lib/components/ui/RoundedIcon.svelte
@@ -8,7 +8,5 @@
 <div
 	class="relative flex items-center justify-center rounded-full bg-primary p-3 ring-2 ring-brand-subtle"
 >
-	<div class:opacity-10={withOpacity}>
-		<svelte:component this={icon}  />
-	</div>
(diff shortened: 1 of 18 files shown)


src/frontend/src/lib/components/ui/RoundedIcon.svelte
```diff
@@ -2,11 +2,11 @@
 	import type { ComponentType } from 'svelte';
 
 	export let icon: ComponentType;
-	export let iconStyleClass = '';
+	export let withOpacity = false;
```
peterpeterparker, line 5: We can also call the property just "opacity"

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/components/ui/RoundedIcon.svelte
AntonioVentilii: changed

AntonioVentilii:
[OISY-130]

PR #3606, opened 2024-11-16
feat(frontend): add resetAll to initTransactionsStore

AntonioVentilii:
# Motivation

Mostly for testing purposes, it is useful to have a way of "resetting" the entire transaction store.

### DIFF
--- a/src/frontend/src/lib/stores/transactions.store.ts
+++ b/src/frontend/src/lib/stores/transactions.store.ts
@@ -16,6 +16,7 @@ export interface TransactionsStore<T> extends CertifiedStore<TransactionsData<T>
 	append: (params: { tokenId: TokenId; transactions: CertifiedTransaction<T>[] }) => void;
 	cleanUp: (params: { tokenId: TokenId; transactionIds: string[] }) => void;
 	nullify: (tokenId: TokenId) => void;
+	resetAll: () => void;
 }
 
 export const initTransactionsStore = <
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/stores/transactions.store.ts
```diff
@@ -64,6 +65,7 @@ export const initTransactionsStore = <
 				[tokenId]: null
 			})),
 		reset,
+		resetAll: () => update(() => ({})),
```
peterpeterparker, line 68: We generally use null for reset because it explicitly indicates a decision to reset, such as when an error occurs. However, in this case, you are setting to the initial value, which creates a slight discrepancy.

AntonioVentilii: I was trying to replicate what is happening for the ETH transactions store too. But for sure I can set `null` on this, and then, in another PR, I adjust that store too

peterpeterparker: I would say so, makes sense?

AntonioVentilii: just to clarify, that means setting `null` in all keys, like `{tokenId1: null, tokenId2: null, ...}`, and NOT that the entire store is `null` instead of `{}`

correct?

peterpeterparker: what does not reset?

AntonioVentilii:
this PR is not needed anymore

PR #3607, opened 2024-11-16
refactor(frontend): move hide info utils to common folder

AntonioVentilii:
# Motivation

Since we are going to re-use them, we move the hide info utils to the common `lib` folder. And we add tests.

# Note

I am not using zod for the keys list because I am replicating the way we do for modals, defining explicit string types. I am not 100% sure that zod enforces the strings in the same way.


peterpeterparker (commented):
The PR description says "move" but, things are also renamed. Should it be two PRs?

AntonioVentilii:
@peterpeterparker you are right, I will do the rename in another PR

peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3608, opened 2024-11-16
refactor(frontend): make modalId a prop

AntonioVentilii:
# Motivation

We would like to pass both `data` and `modalId` to the modal store. But it seems fair first that passing modalId is differentiated from passing any data.

So, we add the prop modalId in the modal store and adapt the code where needed.

# Changes

- Add optional prop `modalId` to modal store type.
- Create sub-functions to define the type of modals:
  - no input
  - `modalId` input
  - `data` input
- Adapt the list of modals functions to use the sub-functions.
- Adapt usage of modal store when inquired on `modalId`.

# Tests

Provided some tests.


src/frontend/src/lib/stores/modal.store.ts
```diff
@@ -32,20 +32,21 @@ export interface Modal<T> {
 		| 'btc-transaction'
 		| 'dapp-details';
 	data?: T;
+	modalId?: symbol;
```
peterpeterparker, line 35: I was thinking, it can just be called `id`.

src/frontend/src/lib/stores/modal.store.ts
```diff
@@ -32,20 +32,21 @@ export interface Modal<T> {
...
-	openSend: <D extends T>(data: D) => void;
-	openBuy: <D extends T>(data: D) => void;
+	openEthReceive: (modalId: symbol) => void;
+	openIcpReceive: (modalId: symbol) => void;
+	openIcrcReceive: (modalId: symbol) => void;
+	openCkBTCReceive: (modalId: symbol) => void;
+	openCkETHReceive: (modalId: symbol) => void;
+	openBtcReceive: (modalId: symbol) => void;
+	openReceive: (modalId: symbol) => void;
+	openSend: (modalId: symbol) => void;
```
peterpeterparker, line 48: This one `modalId` is not set (see `Send` cmp). We improve that in another PR?

src/frontend/src/lib/stores/modal.store.ts
```diff
@@ -68,38 +69,47 @@ export interface ModalStore<T> extends Readable<ModalData<T>> {
 	close: () => void;
 }
 
-const initModalStore = <T>(): ModalStore<T> => {
+export const initModalStore = <T>(): ModalStore<T> => {
```
peterpeterparker, line 72: Why's that?

src/frontend/src/lib/stores/modal.store.ts
```diff
@@ -68,38 +69,47 @@ export interface ModalStore<T> extends Readable<ModalData<T>> {
...
 	const { subscribe, set } = writable<ModalData<T>>(undefined);
 
+	const setType = (type: Modal<T>['type']) => () => set({ type });
```
peterpeterparker, line 75: Does this work? If I get it right, you are declaring function that returns function but, often we consume the store directly in the component so I'm wondering if the function are actually triggered?

For example

```
<button on:click={modalStore.openSend} />
```

src/frontend/src/lib/stores/modal.store.ts
AntonioVentilii: testing eheh!

src/frontend/src/lib/stores/modal.store.ts
AntonioVentilii: yes, better!

src/frontend/src/lib/stores/modal.store.ts
AntonioVentilii: Apparently it works: since the return is a function of the same type, so it runs it as function.

BTW, this gave me the idea, that we can create a simple E2E test for all the Hero buttons, without necessary following the flow

Screenshot 2024-11-18 at 08 44 08

https://github.com/user-attachments/assets/34e28e9d-2d90-4a35-82b7-dad6ceb90f42

src/frontend/src/lib/stores/modal.store.ts
AntonioVentilii: It is actually used in the sub-component `SendButtonWithModal`. I think the idea is that this specific button is used only once, while the `Receive` ones is used both for receive and convert (with different props), so they need a `modalId` one level higher

src/frontend/src/lib/stores/modal.store.ts
peterpeterparker: Can you add a comment about it?

src/frontend/src/lib/stores/modal.store.ts
peterpeterparker: > Apparently it works: since the return is a function of the same type, so it runs it as function.

Ah cool! Thanks for the explanation.

> BTW, this gave me the idea, that we can create a simple E2E test for all the Hero buttons, without necessary following the flow

That would be cool!

peterpeterparker (approved):
LGTM; thx

AntonioVentilii:
[OISY-130]

PR #3609, opened 2024-11-16
fix(frontend): extend "value" type in InputCurrency with string

peterpeterparker:
# Motivation

"Solution 2" of #3592 - i.e. @DenysKarmazynDFINITY spotted a bug that was there since ages. We interpreted the value of the `Input currency` as `number | undefined` while actually the type is a `string | number | undefined`.

# TODO

Create a type `type OptionCurrencyValue = string | number | undefined` or something.

# Changes

- Property `value` in `InputCurrency` must be of type `string | number | undefined` - i.e. `string` was missing
- Populate changes in consumers
- Add `Number()` cast to few utils when property is defined


DenysKarmazynDFINITY (approved):
Cool, thanks for taking care of it :)

PR #3611, opened 2024-11-16
refactor(frontend): extract type for list of ProgressStep

AntonioVentilii:
# Motivation

We extract the type `[ProgressStep, ...ProgressStep[]]` to a generic one for reusability.

# Note

I didn't call the new type `ProgressSteps` because it could go in conflict with the component of the same name from GIX components. If any stronger opinion to have it as `ProgressSteps`, we can easily change the import of the GIX component when needed:

`import { ProgressSteps as ProgressStepsCmp } from '@dfinity/gix-components';`


peterpeterparker (approved):
LGTm, thx

Side notes: naming `ProgressSteps` could have worked too.

PR #3614, opened 2024-11-17
ci(e2e): run E2E snapshots with label

AntonioVentilii:
# Motivation

We still don't run the E2E snapshots automatically on each PR (only when merged).

However it is useful for a dev to decide when he wants that a PR needs to have the CI run on it.

So, to leave less manual work, we make the CI triggered with the addition of a new label `run-e2e-snapshots` label. The label will make it so each new PR.

# Tests

We added the label to this own PR to test it and it worked (see CIs).


DenysKarmazynDFINITY (approved):
Yesss, I like it! Thanks!

PR #3617, opened 2024-11-17
refactor(frontend): move generic type of hide info key

AntonioVentilii:
# Motivation

Since we are going to add more keys to the "Hide Info" flow, it seems correct to move the generic type `HideInfoKey` to the `lib` folder.

# Note

I am not using zod for the keys list because I am replicating the way we do for modals, defining explicit string types. I am not 100% sure that zod enforces the strings in the same way.


peterpeterparker (commented):
Do we really need that extra complexity?

peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3620, opened 2024-11-18
feat(frontend): remove indexCanisterVersion

peterpeterparker:
# Motivation

All SNSes have upgraded their Index Canister version to `-ng` several months ago. Therefore, the `indexCanisterVersion` information is no longer necessary.

# Notes

The weekly job was interesting because it reported Index Canisters as "outdated" when they were out of cycles. However, detecting such issues is not entirely our responsibility. Additionally, the job incorrectly flagged them as "outdated" due to a failing assertion that checked the error message to determine if they were out of cycles. This assertion stopped working because the error message text had changed. Long story short, we decided to remove the check.

# Changes

- Remove assertion from bi-weekly script.
- Remove type and usage of `indexCanisterVersion` in code.


AntonioVentilii (approved):
LGTM tks!

PR #3621, opened 2024-11-18
chore(engine): use Node LTS (v22)

peterpeterparker:
# Motivation

Node active LTS is now version v22. Therefore I suggest to use this version.

# Changes

- Set Node v22 as engine in package.json
- Update GitHub Actions with v22
- Setup Node v22 in Dockerfile similar to what's done in II
- Add missing .node-version file to set Node version for Docker ("missing" because was already referenced by Dockerfile.frontend)
- Keyword assert was replaced with with for JS import (in i18n script)


AntonioVentilii (approved):
LGTM tks

PR #3633, opened 2024-11-18
fix(frontend): ConvertInputsContainer border color

DenysKarmazynDFINITY:
# Motivation

After switching to the colors from theme file, the input started to have a blueish color when it should be light grey. Fixed it.


peterpeterparker (approved):
LGTM, thx

PR #3634, opened 2024-11-18
feat(frontend): BtcConvertFees component

DenysKarmazynDFINITY:
# Motivation

A component that uses already existing UI and contexts to display BTC-related fees.


src/frontend/src/btc/components/convert/BtcConvertFees.svelte
```diff
@@ -0,0 +1,55 @@
...
+
+	const { store: storeUtxosFeeData } = getContext<UtxosFeeContext>(UTXOS_FEE_CONTEXT_KEY);
+
+	let kytFee: bigint | undefined;
+	$: kytFee = $ckBtcMinterInfoStore?.[$destinationToken.id]?.data.kyt_fee;
+
+	let satoshisFee: bigint | undefined;
+	$: satoshisFee =
+		isNullish(sendAmount) || Number(sendAmount) === 0
+			? 0n
```
peterpeterparker, line 25: It's correct to display 0 when amount is undefined?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/btc/components/convert/BtcConvertFees.svelte
DenysKarmazynDFINITY: I think for the moment this should fine, but ideally we need to somehow show a small heads up to users that utxos fee can only be calculated for sendAmount > 0.

PR #3639, opened 2024-11-18
feat(frontend): analytics on prod only

peterpeterparker:
# Motivation

I've got some hiccup with the analytics that are calculated on the fly when a filter is selected. Given that we are not really interested by staging and beta metrics, we disable those for now that way we can release on reading the analytics without filtering.

# Changes

- if not prod no analytics


AntonioVentilii (approved):
LGTM tks

PR #3642, opened 2024-11-18
feat(frontend): ConvertForm component

DenysKarmazynDFINITY:
# Motivation

A container component for the first step of the conversion flow.

### DIFF
--- a/src/frontend/src/lib/components/convert/ConvertForm.svelte
+++ b/src/frontend/src/lib/components/convert/ConvertForm.svelte
@@ -1,0 +1,44 @@
+<script lang="ts">
+	import { createEventDispatcher } from 'svelte';
+	import ConvertAmount from '$lib/components/convert/ConvertAmount.svelte';
+	import Button from '$lib/components/ui/Button.svelte';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/convert/ConvertForm.svelte
```diff
@@ -0,0 +1,44 @@
...
+	export let receiveAmount: number | undefined;
+	export let totalFee: bigint | undefined;
+	export let disabled: boolean;
+	export let insufficientFunds: boolean;
+	export let insufficientFundsForFee: boolean;
+
+	const dispatch = createEventDispatcher();
+</script>
+
+<form on:submit={() => dispatch('icNext')} method="POST">
```
peterpeterparker, line 20: We have other components that wraps a form around everything? This is needed here?

DenysKarmazynDFINITY: I think I saw it somewhere in the send flow, but you're right, there is no need in it. Removed it.

peterpeterparker (approved):
LGTm, thx

PR #3644, opened 2024-11-18
feat(frontend): divide LoaderMultipleEthTransactions into buckets of promises

AntonioVentilii:
# Motivation

The Etherscan API has a maximum [calls/second limit](https://docs.etherscan.io/support/rate-limits) and with the unified transaction we are always breaking it.

There are a few ways to handle/solve this issue, but one of the possible help would be to break the component `LoaderMultipleEthTransactions` into smaller buckets of promises (limited by the API key calls/second limit).

# Edit

As suggested by @peterpeterparker in https://github.com/dfinity/oisy-wallet/pull/3644#discussion_r1846867061 and as discussed offline, I implemented a service that uses `AsyncGenerator` to yield the batches of promises.

# Changes

- Add limit of Etherscan calls/second as variable.
- Remove dependency from store `erc20UserTokensNotInitialized` that to be substituted by nullish check of the other two store dependencies.
- Filter tokens by already loaded.
- Break the promises in buckets using `AsyncGenerator`.
- Execute the promises buckets with a delay of 1 second between them.
- Include initial `debounce` as a safe measure.

# Tests

Adjusted tests.


src/frontend/src/eth/components/loaders/LoaderMultipleEthTransactions.svelte
```diff
@@ -9,23 +10,38 @@
...
 					}
-				}
-			)
+				];
+				return acc;
+			},
+			[]
 		);
+
+		for (const promises of promisesBuckets) {
```
peterpeterparker, line 36: https://github.com/junobuild/juno-js/blob/a0cb36fd538dac9aa588f43502614b41aef9217c/packages/storage/src/api/storage.api.ts#L86

AntonioVentilii:
@peterpeterparker , I updated the PR with your suggestion, have a look pls

peterpeterparker (commented):
Thanks.

- I was under the impression that we would develop a generic service that can chunks calls? But maybe there isn't much generic capatibilities to extract?

- I also did not review the code yet as I think it would be better to provide the service in a separate, dedicated PR. This way reviews are more scoped as the provided tests. Would you mind doing so?

AntonioVentilii:
@peterpeterparker 

- Have a look at:

  - PR #3651 
  
  Is that what you meant? To be re-used here


- I added the following PR, to be reviewed after the above:
  
  - PR #3652

AntonioVentilii:
This PR was superseded by the following one because it is cleaner

-

PR #3649, opened 2024-11-19
feat(frontend): make MessageBox closable

AntonioVentilii:
# Motivation

We add an icon to make the `MessageBox` component closable.

Screenshot 2024-11-19 at 09 31 13
Screenshot 2024-11-19 at 09 31 18

https://github.com/user-attachments/assets/6f55bf86-d122-4f9a-b094-0060022c4d46


src/frontend/src/lib/components/ui/MessageBox.svelte
```diff
@@ -1,25 +1,44 @@
...
+	import { slide } from 'svelte/transition';
+	import IconClose from '$lib/components/icons/lucide/IconClose.svelte';
 	import IconInfo from '$lib/components/icons/lucide/IconInfo.svelte';
+	import { SLIDE_EASING } from '$lib/constants/transition.constants';
+	import { i18n } from '$lib/stores/i18n.store';
 
 	export let level: 'plain' | 'info' | 'light-warning' | 'error' = 'info';
+	export let closable = false;
+
+	let closed = false;
```
peterpeterparker, line 11: It would be more readable to use an "open" or "visible" (for consistency) state?

(I know existing `InfoBox` is not consistent 😉)

src/frontend/src/lib/components/ui/MessageBox.svelte
```diff
@@ -1,25 +1,44 @@
 <script lang="ts">
+	import { slide } from 'svelte/transition';
+	import IconClose from '$lib/components/icons/lucide/IconClose.svelte';
 	import IconInfo from '$lib/components/icons/lucide/IconInfo.svelte';
+	import { SLIDE_EASING } from '$lib/constants/transition.constants';
+	import { i18n } from '$lib/stores/i18n.store';
 
 	export let level: 'plain' | 'info' | 'light-warning' | 'error' = 'info';
+	export let closable = false;
```
peterpeterparker, line 9: In `SendInfo` the `MessageBox` is already displayed or not, do you plan to refactor this afterwards?

AntonioVentilii:
> This screenshot you shared should be here?!?!


sorry, i uploaded the wrong file, it was a video

src/frontend/src/lib/components/ui/MessageBox.svelte
AntonioVentilii: ahahahaha sure, going to use `visible`, it seems the most consistent

src/frontend/src/lib/components/ui/MessageBox.svelte
AntonioVentilii: by default the `MessageBox` is not closable, so, in `SendInfo` case, the logic is based more on the context; and it is not allowed to be closable

I think I can leave it as it is, no? Meaning that if at some point we want to make it closable in that case too, we can change that logic

WDYT?

peterpeterparker: Works for me for sure.

peterpeterparker (approved):
LGTm, thx

AntonioVentilii:
[OISY-130]

PR #3651, opened 2024-11-19
feat(frontend): service to batch promises

AntonioVentilii:
# Motivation

Splitting PR #3644 , we create first a services useful to break a list of promises in smaller batches.

# Tests

Test provided.


src/frontend/src/lib/services/batch.services.ts
```diff
@@ -0,0 +1,30 @@
+export const batchPromisesPerSecond = async function* <T, R>({
```
peterpeterparker, line 1: Maybe just `batchPromises` or `batch`? I like `batch` actually wdyt?

src/frontend/src/lib/services/batch.services.ts
```diff
@@ -0,0 +1,30 @@
...
+	items,
+	maxCallsPerSecond,
+	promise,
+	validate,
+	afterPromise
+}: {
+	items: T[];
+	maxCallsPerSecond: number;
```
peterpeterparker, line 9: I did not understood `maxCallsPerSecond` from the naming. Now that I read the code I get it but, it was not natural. Any other naming idea?

peterpeterparker (commented):
What about simplifying and just passing all the promises?

Also not sure what's the use case for `afterPromise`?

src/frontend/src/lib/services/batch.services.ts
AntonioVentilii: batch!

src/frontend/src/lib/services/batch.services.ts
AntonioVentilii: I went with `batchSize` ? WDYT

AntonioVentilii:
@peterpeterparker as discussed offline, I simplified the logic and now it takes just a list of promises as input

src/frontend/src/lib/services/batch.services.ts
```diff
@@ -0,0 +1,14 @@
+export const batch = async function* <T>({
+	promises,
+	batchSize
```
peterpeterparker, line 3: Just `size`? Don't know, today I'm in the mood of shortening any variable names 😂

peterpeterparker (approved):
LGTm, thx

src/frontend/src/lib/services/batch.services.ts
AntonioVentilii: ahahahahah! I noticed! I am just afraid that it is not clear enough, it should say "how many promises per batch"

AntonioVentilii:
[OISY-130]

PR #3653, opened 2024-11-19
test(frontend): test modal store directly

AntonioVentilii:
# Motivation

As discussed offline and in https://github.com/dfinity/oisy-wallet/pull/3608#discussion_r1845974057 , it does not make really sense to export `initModalStore` to test it, when we can test directly the initiated store modalStore.

In this PR we replace it.


peterpeterparker (approved):
Nice!

PR #3656, opened 2024-11-19
feat(frontend): BtcConvertProgress component

DenysKarmazynDFINITY:
# Motivation

This PR contains a basic version of the BtcConvertProgress component. There is already a new, much fancier designs in Figma for this modal, therefore I kept it simple in the first iteration.

Screenshot 2024-11-19 at 12 24 50

### DIFF
--- a/src/frontend/src/btc/components/convert/BtcConvertProgress.svelte
+++ b/src/frontend/src/btc/components/convert/BtcConvertProgress.svelte
@@ -1,0 +1,29 @@
+<script lang="ts">
+	import type { ProgressStep } from '@dfinity/gix-components';
+	import InProgressWizard from '$lib/components/ui/InProgressWizard.svelte';
+	import { ProgressStepsConvert } from '$lib/enums/progress-steps';
+	import { i18n } from '$lib/stores/i18n.store';
+
+	export let convertProgressStep: string = ProgressStepsConvert.INITIALIZATION;
+
+	let steps: [ProgressStep, ...ProgressStep[]];
+	$: steps = [
+		{
+			step: ProgressStepsConvert.INITIALIZATION,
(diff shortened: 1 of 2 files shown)


src/frontend/src/btc/components/convert/BtcConvertProgress.svelte
```diff
@@ -0,0 +1,29 @@
+<script lang="ts">
+	import type { ProgressStep } from '@dfinity/gix-components';
+	import InProgressWizard from '$lib/components/ui/InProgressWizard.svelte';
+	import { ProgressStepsConvert } from '$lib/enums/progress-steps';
+	import { i18n } from '$lib/stores/i18n.store';
+
+	export let convertProgressStep: string = ProgressStepsConvert.INITIALIZATION;
+
+	let steps: [ProgressStep, ...ProgressStep[]];
```
peterpeterparker, line 9: @AntonioVentilii somewhere there should be a type for this now right?

peterpeterparker (approved):
LGTM, thx

Just tagged Antonio because there is a shorthand type floating around somewhere which later on can replace the type.

src/frontend/src/btc/components/convert/BtcConvertProgress.svelte
DenysKarmazynDFINITY: Just searched through the codebase and I found this - https://github.com/dfinity/oisy-wallet/blob/main/src/frontend/src/lib/types/transaction.ts#L44, which is not really general. So, maybe it's still only planned to create some kind of `NonEmptyArray<T>`?

peterpeterparker: Sure. Already approved hours ago ;)

PR #3657, opened 2024-11-19
feat(frontend): save closable setting of MessageBox

AntonioVentilii:
# Motivation

Similar to other info box components, we want to persist the "closed" state of `MessageBox`. However we give it as optional prop.

# Note

The naming is not the best, since "hide info" does not really match `MessageBox` that is a component for info, warns, errors, etc.

# Changes

- Include "hide info" functions in component `MessageBox`.
- Add prop to define the key.
- Apply the new feature to the Bitcoin transactions disclaimer.

# Tests

https://github.com/user-attachments/assets/7c9de6db-361b-4f0f-bcb4-96cf30cdee0b


src/frontend/src/lib/components/ui/MessageBox.svelte
```diff
@@ -1,17 +1,25 @@
...
 	import { SLIDE_EASING } from '$lib/constants/transition.constants';
 	import { i18n } from '$lib/stores/i18n.store';
+	import { type HideInfoKey, saveHideInfo, shouldHideInfo } from '$lib/utils/info.utils';
 
 	export let level: 'plain' | 'info' | 'light-warning' | 'error' = 'info';
 	export let closable = false;
+	export let hideInfoKey: HideInfoKey | undefined = undefined;
 
 	let visible = true;
+	$: visible = nonNullish(hideInfoKey) ? !shouldHideInfo(hideInfoKey) : true;
```
peterpeterparker, line 15: The linter does not complain that it rather like an assignation? Just in case useful, something like that would work maybe?

```
$: visible = isNullish(hideInfoKey) || !shouldHideInfo(hideInfoKey);
```

src/frontend/src/lib/utils/info.utils.ts
```diff
@@ -3,7 +3,8 @@ import { browser } from '$app/environment';
 export type HideInfoKey =
 	| 'oisy_ic_hide_bitcoin_info'
 	| 'oisy_ic_hide_ethereum_info'
-	| 'oisy_ic_hide_erc20_info';
+	| 'oisy_ic_hide_erc20_info'
+	| 'oisy_ic_hide_bitcoin_transactions_disclaimer';
```
peterpeterparker, line 7: `activity` instead of `transactions_disclaimer` since it's the activity page? No strong opinion though

src/frontend/src/lib/components/ui/MessageBox.svelte
```diff
@@ -1,17 +1,25 @@
...
 
 	const close = () => {
 		visible = false;
+
+		if (nonNullish(hideInfoKey)) {
```
peterpeterparker, line 20: nitpick:

```
if (isNullish(hideInfoKey)) {
   return;
}

saveHideInfo(hideInfoKey);
```

peterpeterparker (commented):
Is there a use case where `closable` is not set but `hideInfoKey` is used?
If not, shouldn't we have a single prop?

src/frontend/src/lib/components/ui/MessageBox.svelte
AntonioVentilii: done!

src/frontend/src/lib/components/ui/MessageBox.svelte
AntonioVentilii: indeed it doesn't! I will add a rule in the linter

peterpeterparker:
> WDYT? better naming?

Better naming and I like that there is a single prop but, is there a use case `closable` is not set but `hideInfoKey` is used or the contrary? You know what I mean, do we really need two props?

AntonioVentilii:
@peterpeterparker the only case i see for now is that `closable` is given but not `hideInfoKey`, meaning that we want that the message box can be closed, but not persisted

however no practical use-cases no, indeed there are none, for now

 so I can merge the two pros in just one given key `closableKey` (WDYT?), if you agree; and then, when the time arrives, we adjust

AntonioVentilii:

peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3659, opened 2024-11-19
style(frontend): combine dapps tags

BonomoAlessandro:
# Motivation

We want to reduce the number of tags. To achieve this we want to replace the `DEX` and `Liquidity` Tags with a single `DeFi` Tag.

# Changes

- replace tags

# Tests

**before:**
image

**after:**
image


peterpeterparker (approved):
LGTM, thx

PR #3660, opened 2024-11-19
feat(frontend): remove token store from IcTransactionModal

AntonioVentilii:
# Motivation

The component `IcTransactionModal` was using the token store to define the token to which the transaction refers too.

# Note

An alternative could be to create a context for the transaction modal.

# Changes

- Add prop token to `IcTransactionModal`.
- Make the `modalStore` call pass not only the transaction but the token too.
- Create function to "extract" the transaction data from the modal store. Plus tests.
- Adapt both `IcTransactions` and `AllTransactionsList`.

# Tests

Screenshot 2024-11-19 at 15 11 27

### DIFF
--- a/src/frontend/src/icp/components/transactions/IcTransaction.svelte
+++ b/src/frontend/src/icp/components/transactions/IcTransaction.svelte
@@ -43,7 +43,7 @@
 </script>
 
 <Transaction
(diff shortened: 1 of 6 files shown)


peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3662, opened 2024-11-19
feat(frontend): divide LoaderMultipleEthTransactions into buckets of promises

AntonioVentilii:
# Motivation

The Etherscan API has a maximum [calls/second limit](https://docs.etherscan.io/support/rate-limits) and with the unified transaction we are always breaking it.

There are a few ways to handle/solve this issue, but one of the possible help would be to break the component `LoaderMultipleEthTransactions` into smaller buckets of promises (limited by the API key calls/second limit).

Using the service created in PR #3651 , we use `batch` for the component.

# Changes

- Change `batch` so that it may return the `Promise` values.
- Create list of promises inside `LoaderMultipleEthTransactions`.
- Create the `AsyncGenerator` with the limitations of Etherscan calls/second.
- Update the tokens already loaded using the generator.

# Tests

Adjusted tests.


src/frontend/src/env/rest/etherscan.env.ts
```diff
@@ -1 +1,5 @@
+import { BETA, PROD } from '$lib/constants/app.constants';
+
 export const ETHERSCAN_API_KEY = import.meta.env.VITE_ETHERSCAN_API_KEY;
+
+export const ETHERSCAN_MAX_CALLS_PER_SECOND = BETA || PROD ? 10 : 5;
```
peterpeterparker, line 5: Staging should not be considered as prod?

src/frontend/src/eth/components/loaders/LoaderMultipleEthTransactions.svelte
```diff
@@ -1,31 +1,56 @@
...
 			return;
 		}
 
-		await Promise.allSettled(
-			[...$enabledEthereumTokens, ...$enabledErc20Tokens].map(
-				async ({ network: { id: networkId }, id: tokenId }) => {
-					if (!tokensLoaded.includes(tokenId)) {
-						await loadEthereumTransactions({ tokenId, networkId });
-						tokensLoaded.push(tokenId);
+		const promises = [...$enabledEthereumTokens, ...$enabledErc20Tokens].reduce<
```
peterpeterparker, line 19: with the component, can you extract a function for this? it would make reading the `load` function and de facto the review a bit easier

src/frontend/src/eth/components/loaders/LoaderMultipleEthTransactions.svelte
```diff
@@ -1,31 +1,56 @@
...
+			return acc;
+		}, []);
+
+		const loader = batch<ResultSuccess & { tokenId: TokenId }>({
+			promises,
+			batchSize: ETHERSCAN_MAX_CALLS_PER_SECOND
+		});
+
+		for await (const results of loader) {
+			tokensLoaded = results.reduce<TokenId[]>(
```
peterpeterparker, line 39: a function here too would be nice

src/frontend/src/eth/components/loaders/LoaderMultipleEthTransactions.svelte
```diff
@@ -1,31 +1,56 @@
...
-		await Promise.allSettled(
-			[...$enabledEthereumTokens, ...$enabledErc20Tokens].map(
-				async ({ network: { id: networkId }, id: tokenId }) => {
-					if (!tokensLoaded.includes(tokenId)) {
-						await loadEthereumTransactions({ tokenId, networkId });
-						tokensLoaded.push(tokenId);
+		const promises = [...$enabledEthereumTokens, ...$enabledErc20Tokens].reduce<
+			Promise<ResultSuccess & { tokenId: TokenId }>[]
+		>((acc, { network: { id: networkId }, id: tokenId }) => {
+			if (!tokensLoaded.includes(tokenId)) {
```
peterpeterparker, line 22: For readability reason I would do the contrary

```
if (tokensLoaded.includes(tokenId)) {
  return acc;
}

// Stuffs here
```

src/frontend/src/eth/components/loaders/LoaderMultipleEthTransactions.svelte
```diff
@@ -1,31 +1,56 @@
...
 		}
 
-		await Promise.allSettled(
-			[...$enabledEthereumTokens, ...$enabledErc20Tokens].map(
-				async ({ network: { id: networkId }, id: tokenId }) => {
-					if (!tokensLoaded.includes(tokenId)) {
-						await loadEthereumTransactions({ tokenId, networkId });
-						tokensLoaded.push(tokenId);
+		const promises = [...$enabledEthereumTokens, ...$enabledErc20Tokens].reduce<
+			Promise<ResultSuccess & { tokenId: TokenId }>[]
```
peterpeterparker, line 20: If within the component you create a type for `Promise<ResultSuccess & { tokenId: TokenId }>` that can help readability too

src/frontend/src/env/rest/etherscan.env.ts
AntonioVentilii: Well, for staging we have an API key that is the Free plan, hence the lower limit. Unless we decide to change it, but I think this should be another topic, offline eheh

src/frontend/src/eth/components/loaders/LoaderMultipleEthTransactions.svelte
AntonioVentilii: done!

src/frontend/src/eth/components/loaders/LoaderMultipleEthTransactions.svelte
AntonioVentilii: makes sense, changed

src/frontend/src/eth/components/loaders/LoaderMultipleEthTransactions.svelte
AntonioVentilii: same!

AntonioVentilii:

src/frontend/src/eth/services/eth-transactions-batch.services.ts
```diff
@@ -0,0 +1,51 @@
...
+	tokens: Token[];
+	tokensAlreadyLoaded: TokenId[];
+}): AsyncGenerator<PromiseSettledResult<ResultByToken>[]> => {
+	const promises = tokens.reduce<PromiseResult[]>(
+		(acc, { network: { id: networkId }, id: tokenId }) => {
+			if (tokensAlreadyLoaded.includes(tokenId)) {
+				return acc;
+			}
+
+			const promise = (async (): PromiseResult => {
```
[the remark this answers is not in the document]
AntonioVentilii: yes, it can be done both ways actually, but if i don't execute it here (i don't create the promise but only the function), then I think I need to change the batch util to be
`const results = await Promise.allSettled(batch.map((fn) => fn())`

src/frontend/src/eth/services/eth-transactions-batch.services.ts
```diff
@@ -0,0 +1,51 @@
...
+	});
+};
+
+export const batchResultsToTokenId = (results: PromiseSettledResult<ResultByToken>[]): TokenId[] =>
+	results.reduce<TokenId[]>((acc, result) => {
+		if (result.status === 'rejected') {
+			return acc;
+		}
+
+		if (!result.value.success) {
```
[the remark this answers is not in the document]
AntonioVentilii: it is part of the results input, since they have the type

PromiseSettledResult<ResultByToken>

that in case of fulfilled promise is something like

```typescript
{
	status: 'fulfilled',
	value: { success, tokenId }
}
```

src/frontend/src/eth/services/eth-transactions-batch.services.ts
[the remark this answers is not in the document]
peterpeterparker: I wonder if the later would not be nicer? But not a show stopper for this PR.

peterpeterparker (approved):
LGTm, thx

AntonioVentilii:
[OISY-130]

PR #3663, opened 2024-11-19
feat(frontend): nullify transactions instead of resetting

AntonioVentilii:
# Motivation

In case of error, the function `loadEthereumTransactions` will generally reset the entire `ethTransactionsStore`. That was acceptable when the store was used only in the token-specific transaction page.

But even in that case, we would erase the entire transaction store, even if the error may have happened only for a token.

Now, with the unified transaction list, we would like to manage each token differently. So we add the `nullify` function to the store, that will "nullify" the transactions for a specific token, and not for all. This was inspired by what was done in the `transactionStore` too.

# Changes

- Add `nullify` function to `ethTransactionsStore` that sets `null` for a specific token.
- Use such function in `loadEthereumTransactions` and its sub-functions.

# Tests

Create a few tests for `ethTransactionsStore`, that were missing.


peterpeterparker (approved):
LGTM; thx

AntonioVentilii:
[OISY-130]

PR #3637, opened 2024-11-18
feat(frontend): service util for rest requests

AntonioVentilii:
# Motivation

The Etherscan API has a maximum [calls/second limit](https://docs.etherscan.io/support/rate-limits) and with the unified transaction we are always breaking it.

There are a few ways to handle/solve this issue, but, as first important guard, we would like to create a way to handle retries in our requests to the Rest API. Right now a request would fail at the first try (if it fails).

So, we create a service handler (similar to `queryAndUpdate`) that accepts a request and handles success, errors and retries.

# Tests

Provided some tests.

PR #3664, opened 2024-11-19
refactor(frontend): simplify restRequest

AntonioVentilii:
# Motivation

In PR #3637 , we created an util to send REST requests with a certain amount of retries, before failing.

However it was over-engineered, and some logic didn't quite make sense. For example, it doesn't make sense to have `onSuccess` and `onError` return values. This should be handled on the caller side.

So, we greatly simplify the function and we make it even nested (more functional).

# Changes

- Remove `onSuccess`: the function will return the request results, if successful.
- Remove `onError`: the function will throw directly the error if it exceeded the max attempts.
- Create a sub-fuction that is called nested inside the main function.

# Tests

Adapted the tests, just removing the parts concerning `onSuccess` and `onError`.


src/frontend/src/lib/services/rest.services.ts
```diff
@@ -10,35 +8,29 @@ export interface RestRequestParams<Response, Error = unknown, Success = void> {
  * A utility function to make a REST request with possible retries.
  *
  * @param request - The request function to call.
- * @param onSuccess - The function to call when the request is successful.
- * @param onError - The optional function to call when the request fails and the max retries are reached.
  * @param onRetry - The optional function to call when the request fails and a retry is needed.
  * @param maxRetries - The maximum number of retries to attempt.
- * @returns The result of the onSuccess/onError function or undefined if the max retries are reached.
+ * @returns The result of the request or it throws an error if the max retries are reached.
```
peterpeterparker, line 13: I think you may be able to document the error with `@throws`

src/frontend/src/lib/services/rest.services.ts
```diff
@@ -10,35 +8,29 @@ export interface RestRequestParams<Response, Error = unknown, Success = void> {
...
  */
-export const restRequest = async <Response, Error, Success>({
+export const restRequest = async <Response, Error>({
```
peterpeterparker, line 15: unrelated: just thought at the naming `retryRequest` or `retry` since the moto of the day is shortening names 😉

src/frontend/src/lib/services/rest.services.ts
```diff
@@ -10,35 +8,29 @@ export interface RestRequestParams<Response, Error = unknown, Success = void> {
...
+			if (retryCount >= maxRetries) {
 				console.error(`Max retries reached. Error:`, error);
-				return onError?.(error as Error);
+				throw error;
 			}
 
 			await onRetry?.({ error: error as Error, retryCount });
-			console.warn(`Request attempt ${retryCount} failed. Retrying...`);
+			console.warn(`Request attempt ${retryCount + 1} failed. Retrying...`);
+			return executeRequest(retryCount + 1);
```
peterpeterparker, line 31: isn't the keyword `await` missing here?

AntonioVentilii: very true! strange that it was not flagged...

src/frontend/src/lib/services/rest.services.ts
AntonioVentilii: ahahah! i will go with `retry`

src/frontend/src/lib/services/rest.services.ts
AntonioVentilii: done!

peterpeterparker (approved):
LGTM, thx

Added a comment for future development.

src/frontend/src/lib/services/rest.services.ts
```diff
@@ -10,35 +8,30 @@ export interface RestRequestParams<Response, Error = unknown, Success = void> {
...
-			return onSuccess(response);
+			return await request();
 		} catch (error: unknown) {
-			retryCount++;
-
-			if (retryCount > maxRetries) {
+			if (retryCount >= maxRetries) {
 				console.error(`Max retries reached. Error:`, error);
-				return onError?.(error as Error);
+				throw error;
```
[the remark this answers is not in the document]
AntonioVentilii: I was thinking that this could be handled with onRetry: if you want specific errors to be raised and not retried, you can do it in that function

`onRetry: (error) => isSpecificError ? throw : ....`

or viceversa, like in our use-case

`onRetry: (error) => isCallsLimitError ? ... : throw`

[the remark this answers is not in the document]
peterpeterparker: Handling it with `onRetry` sounds good + I would end the process with a custom `throw new RetryError()` or something but not throwing the last error

AntonioVentilii:
[OISY-130]

PR #3667, opened 2024-11-19
feat(frontend)/remove-token-store-from-EthTransactionModal

AntonioVentilii:
# Motivation

Similar to PR #3660 , the component `EthTransactionModal` was using the token store to define the token to which the transaction refers too.

### DIFF
--- a/src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
+++ b/src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
@@ -11,7 +11,6 @@
 	import ExternalLink from '$lib/components/ui/ExternalLink.svelte';
 	import Value from '$lib/components/ui/Value.svelte';
 	import { ethAddress } from '$lib/derived/address.derived';
-	import { tokenWithFallback } from '$lib/derived/token.derived';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { modalStore } from '$lib/stores/modal.store';
 	import type { Transaction } from '$lib/types/transaction';
@@ -21,8 +20,10 @@
 		shortenWithMiddleEllipsis
 	} from '$lib/utils/format.utils';
 	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+	import type { OptionToken } from '$lib/types/token';
 
 	export let transaction: Transaction;
+	export let token: OptionToken;
 
 	let from: string;
 	let to: string | undefined;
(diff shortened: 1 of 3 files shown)


src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
```diff
@@ -124,14 +125,16 @@
...
 			<svelte:fragment slot="label">{$i18n.core.text.amount}</svelte:fragment>
-			<output>
-				{formatToken({
-					value,
-					unitName: $tokenWithFallback.decimals,
-					displayDecimals: $tokenWithFallback.decimals
-				})}
-				{$tokenWithFallback.symbol}
-			</output>
+			{#if nonNullish(token)}
```
peterpeterparker, line 128: Then you have the label "Amount:" but no amount? Shoudn't the all block be hidden if no token is provided?

src/frontend/src/lib/utils/transaction.utils.ts
```diff
@@ -69,7 +70,7 @@ export const groupTransactionsByDate = <T extends AnyTransactionUi>(
 	}, {});
 };
 
-export const mapTransactionModalData = <T extends AnyTransactionUi>({
+export const mapTransactionModalData = <T extends AnyTransactionUi | Transaction>({
```
peterpeterparker, line 73: Doesn't `AnyTransactionUi` extends `Transaction`?

`<T extends Transaction>`?

AntonioVentilii: No indeed, only `EthTransactionUi`, but the Ethereum token-specific transaction list uses type `Transaction`, I will analyze the reason why

src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
AntonioVentilii: ideally the token MUST be provided, it is a safeguard only because for now we still use the `token` store

i replicated what we were doing in `IcTransactionModal`, but I think the idea is that if for some reason there is a lag in loading the prop token, at least we already have the `Amount:` label and then we load the amount?

Not really sure UX-wise what is better, WDYT?

peterpeterparker: Personally I would hide the all block if not provided. If you want to keep the amount, then you should probably add a `&ZeroWidthSpace;` for the empty value otherwise there might be a lack of space in the UI?

AntonioVentilii: let's hide the entire block, makes more sense

peterpeterparker (approved):
LGTm, thx

AntonioVentilii:
[OISY-130]

PR #3668, opened 2024-11-19
feat(frontend): remove token store from BtcTransactionModal

AntonioVentilii:
# Motivation

Similar to PR #3660 , the component `BtcTransactionModal` was using the token store to define the token to which the transaction refers too.


peterpeterparker (approved):
LGTM, thx

AntonioVentilii:
[OISY-130]

PR #3671, opened 2024-11-20
chore(e2e): fix IDE import cannot find module $lib

peterpeterparker:
# Motivation

Both Webstorm and Code were not able to resolve the module of `$lib` from the e2e modules.

# Changes

- Add specific tsconfig

# Screenshot

Capture d’écran 2024-11-20 à 06 49 22

PR #3672, opened 2024-11-20
chore(e2e): use $lib import

peterpeterparker:
# Motivation

Spotted to import in e2e that can be shortened with `$lib` particularly after #3671 is set.


DenysKarmazynDFINITY (approved):
Thanks!

PR #3674, opened 2024-11-20
fix(frontend): adjust input JSON for ICRC default tokens

AntonioVentilii:
# Motivation

In PR #3646 , I forgot to include the modifications of the JSON file that gives the default ICRC tokens.


peterpeterparker (approved):
We need a test...

LGTM, thx

PR #3677, opened 2024-11-20
feat(frontend): ConvertWizard component

DenysKarmazynDFINITY:
# Motivation

`ConvertWizard` - semi-last component in the Conversion flow tree.

### DIFF
--- a/src/frontend/src/lib/components/convert/ConvertWizard.svelte
+++ b/src/frontend/src/lib/components/convert/ConvertWizard.svelte
@@ -1,0 +1,33 @@
+<script lang="ts">
+	import type { WizardStep } from '@dfinity/gix-components';
+	import { getContext } from 'svelte';
+	import BtcConvertTokenWizard from '$btc/components/convert/BtcConvertTokenWizard.svelte';
+	import MessageBox from '$lib/components/ui/MessageBox.svelte';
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/convert/ConvertWizard.svelte
```diff
@@ -0,0 +1,33 @@
...
+		{formCancelAction}
+		bind:sendAmount
+		bind:receiveAmount
+		bind:convertProgressStep
+		on:icBack
+		on:icNext
+		on:icClose
+	/>
+{:else}
+	<div class="mt-6"><MessageBox>{$i18n.convert.text.unsupported_token_conversion}</MessageBox></div>
```
peterpeterparker, line 32: Can you provide a screenshot? How does it looks like

DenysKarmazynDFINITY: Sure, here it is:
Screenshot 2024-11-20 at 11 16 34

This message can be visible only if something went wrong with the way how we use `ConvertWizard`. In other places, we render an empty slot in such cases:
Screenshot 2024-11-20 at 11 16 49

peterpeterparker (approved):
LGTM, thx

PR #3679, opened 2024-11-20
feat(frontend): do not show Your Addresses in Activity

AntonioVentilii:
# Motivation

We don't want to show "Your Addresses" in the Activity page.

Screenshot 2024-11-20 at 10 50 11

### DIFF
--- a/src/frontend/src/lib/components/core/Menu.svelte
+++ b/src/frontend/src/lib/components/core/Menu.svelte
@@ -57,7 +57,7 @@
 	$: activityRoute = isRouteActivity($page);
 
 	let addressesOption = true;
-	$: addressesOption = !settingsRoute && !dAppExplorerRoute;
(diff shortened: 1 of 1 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -57,7 +57,7 @@
 	$: activityRoute = isRouteActivity($page);
 
 	let addressesOption = true;
-	$: addressesOption = !settingsRoute && !dAppExplorerRoute;
+	$: addressesOption = !settingsRoute && !dAppExplorerRoute && !activityRoute;
```
peterpeterparker, line 60: For a long term solution, shouldn't it be "if assets route then visible"?

AntonioVentilii:
[OISY-130]

PR #3681, opened 2024-11-20
fix(frontend): modal styling issue

DenysKarmazynDFINITY:
# Motivation

Adding `flex` to `ContentWithToolbar` introduced in some particular cases some unwanted spacings between elements. Solution - make the component accept `styleClass` to allow consumers adjust the styling as needed. Having both `styleClass` and `minHeight` doesn't make too much sense, therefore I removed the latter.

Capture d’écran 2024-11-20 à 10 17 26

Capture d’écran 2024-11-20 à 10 21 28

### DIFF
--- a/src/frontend/src/lib/components/convert/ConvertReview.svelte
+++ b/src/frontend/src/lib/components/convert/ConvertReview.svelte
@@ -15,7 +15,7 @@
 	const dispatch = createEventDispatcher();
 </script>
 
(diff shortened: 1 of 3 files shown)


peterpeterparker (approved):
LGTm, thx

peterpeterparker:
For history purpose, maybe you can add two screenshots to the description?

PR #3684, opened 2024-11-20
feat(frontend): remove network id from settings nav

peterpeterparker:
# Motivation

We are fine, at least for now, not supporting network ID in the URL for the navigation - i.e. we are fine when user open assets if it restart with no network selected given that's an advanced option anyway.

# Changes

- Remove network id from settings url


AntonioVentilii (approved):
LGTM tks!

PR #3687, opened 2024-11-20
feat(frontend): ConvertModal component

DenysKarmazynDFINITY:
# Motivation

The top-level conversion modal component.


src/frontend/src/tests/lib/components/convert/ConvertModal.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+			[
+				CONVERT_CONTEXT_KEY,
+				{
+					sourceToken: readable(BTC_MAINNET_TOKEN),
+					destinationToken: readable(ICP_TOKEN)
+				}
+			]
+		]);
+
+	it('should display correct modal title after navigating between steps', async () => {
```
DenysKarmazynDFINITY, line 21: We have tests for all children components (where it was applicable), so in case of `ConvertModal`, I couldn't come up with some additional test suites that have not been covered yet. Also, Svelte doesn't really provide us with the possibility of checking component internal vars. IMO, all further tests belong to the e2e level.

peterpeterparker (approved):
LGTm, thx

PR #3689, opened 2024-11-20

cgundy:
# Motivation

This is part of a larger effort to replace PATs generated my service accounts with GitHub Apps.

# Changes

No functional changes.

# Tests

No tests run, but has been successfully installed in other repos.


AntonioVentilii (commented):
@cgundy thank you for this! Do you think it makes sense that it becomes an action instead of a step in each workflow? Just to avoid repetition of code

cgundy:
@AntonioVentilii do you mean like a composite action? I'm not sure I understand.

AntonioVentilii:
@cgundy yes, correct! a composite. I see you changed it in a few CIs. I am just asking if it is possible at all eheh!

AntonioVentilii (approved):
LGTM tks!

PR #3691, opened 2024-11-20
feat(frontend): adjust day difference calculation

DenysKarmazynDFINITY:
# Motivation

To properly decide whether "today" or "yesterday" label should be displayed, `daysDifference` needs to be calculated with vars in dates-only format.


AntonioVentilii (approved):
LGTM tks!

P.S. it would be cool to test this edge case

DenysKarmazynDFINITY:
@AntonioVentilii I added a todo comment so we don't forget it.

PR #3692, opened 2024-11-20
chore(e2e): add name to workflow

AntonioVentilii:
# Motivation

For more clarity, we add a name to the step that saves the report.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3698, opened 2024-11-20
feat(frontend): show timestamp instead of created_at for ICP

AntonioVentilii:
# Motivation

For ICP token (not ICRC, but ICP indeed), we were showing the `created_at_time` as reference date. However, since that date can be `null`, or somehow it is mostly an input set when creating the transaction, it does not seem like a good indicator for the transaction history.

So, similar to ICRC, we will use the `timestamp` field instead.

### Before

Screenshot 2024-11-20 at 14 50 23

### After

Screenshot 2024-11-20 at 14 55 48


src/frontend/src/icp/utils/icp-transactions.utils.ts
```diff
@@ -56,11 +56,11 @@ export const mapIcpTransaction = ({
 	transaction: IcpTransaction;
 	identity: OptionIdentity;
 }): IcTransactionUi => {
-	const { operation, created_at_time, transferToSelf } = transaction;
+	const { operation, timestamp, transferToSelf } = transaction;
```
peterpeterparker, line 59: Did you double check with FI team or are we sure the timestamp is a better source?

That said, already approving.

peterpeterparker (approved):
Approving, just check my question.

src/frontend/src/icp/utils/icp-transactions.utils.ts
AntonioVentilii: As per ongoing discussion with FI, they confirmed the scope of timestamp (as you said); they still need to confirm what is the use-case of "missing" timestamp; but since it seems something not really on our control. So, to order the transactions, it seems ok

PR #3699, opened 2024-11-20
feat(frontend): update ConvertToCkBTC component

DenysKarmazynDFINITY:
# Motivation

The last piece of the puzzle - updating ConvertToCkBTC component.

### DIFF
--- a/src/frontend/src/btc/components/convert/ConvertToCkBTC.svelte
+++ b/src/frontend/src/btc/components/convert/ConvertToCkBTC.svelte
@@ -1,19 +1,80 @@
 <script lang="ts">
-	import { getContext } from 'svelte';
+	import { isNullish, nonNullish } from '@dfinity/utils';
+	import { setContext, getContext } from 'svelte';
 	import { BTC_MAINNET_TOKEN } from '$env/tokens.btc.env';
+	import IcCkListener from '$icp/components/core/IcCkListener.svelte';
+	import { loadAllCkBtcInfo } from '$icp/services/ckbtc.services';
+	import { initCkBTCMinterInfoWorker } from '$icp/services/worker.ck-minter-info.services';
+	import { ckBtcMinterInfoStore } from '$icp/stores/ckbtc.store';
+	import type { IcCkToken } from '$icp/types/ic-token';
+	import ConvertModal from '$lib/components/convert/ConvertModal.svelte';
 	import ButtonHero from '$lib/components/hero/ButtonHero.svelte';
 	import IconCkConvert from '$lib/components/icons/IconCkConvert.svelte';
+	import { authIdentity } from '$lib/derived/auth.derived';
 	import { isBusy } from '$lib/derived/busy.derived';
+	import { modalConvertBTCToCkBTC } from '$lib/derived/modal.derived';
+	import { tokens } from '$lib/derived/tokens.derived';
+	import {
+		initConvertContext,
+		CONVERT_CONTEXT_KEY,
+		type ConvertContext
+	} from '$lib/stores/convert.store';
 	import { HERO_CONTEXT_KEY, type HeroContext } from '$lib/stores/hero.store';
 	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+	import { findTwinToken } from '$lib/utils/token.utils';
+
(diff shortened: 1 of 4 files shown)


src/frontend/src/btc/components/convert/ConvertToCkBTC.svelte
```diff
@@ -1,19 +1,80 @@
...
+	let ckBtcToken: IcCkToken | undefined;
+	$: ckBtcToken = findTwinToken({
+		tokenToPair: BTC_MAINNET_TOKEN,
+		tokens: $tokens
+	});
+
+	$: ckBtcToken,
+		(() =>
+			nonNullish(ckBtcToken) &&
+			setContext<ConvertContext>(
```
peterpeterparker, line 36: It does not lead to some unexpected issue to not declare a context straight went the component is mounted?

Or shouldn't the context be declared within the modal? Not sure what's our pattern anymore.

src/frontend/src/btc/components/convert/ConvertToCkBTC.svelte
```diff
@@ -1,19 +1,80 @@
...
+					sourceToken: BTC_MAINNET_TOKEN,
+					destinationToken: ckBtcToken
+				})
+			))();
+
+	let minterInfoLoaded = false;
+
+	const openConvert = async () => {
+		if (isNullish(ckBtcToken)) {
+			return;
```
peterpeterparker, line 48: This check is only there for TS and race condition? Maybe adding a comment?

src/frontend/src/btc/components/convert/ConvertToCkBTC.svelte
```diff
@@ -1,19 +1,80 @@
...
+		}
+
+		minterInfoLoaded = nonNullish($ckBtcMinterInfoStore?.[ckBtcToken.id]);
+
+		await loadAllCkBtcInfo({
```
peterpeterparker, line 53: You on purpose does not display an animation while loading things?

We have a pattern for this, something like "Hold tight while we are loading ...." displayed over a backdrop until the modal gets opened

src/frontend/src/btc/components/convert/ConvertToCkBTC.svelte
DenysKarmazynDFINITY: Correct, for TS only. Added a comment.

src/frontend/src/btc/components/convert/ConvertToCkBTC.svelte
peterpeterparker: Ah coolio

peterpeterparker (approved):
LGTM, thx

src/frontend/src/btc/components/convert/ConvertToCkBTC.svelte
DenysKarmazynDFINITY: I haven't noticed any issues with it, mostly because we make sure that the modal, that already expects context to be defined, cannot be opened if `ckBtcToken` is not loaded yet.

peterpeterparker: Oops, approved too quickly. It would be nice to double-check if this is not a discrepancy with our existing patterns.

DenysKarmazynDFINITY: In the other flows (e.g. ConvertToCkETH or ConverToCkErc20), we also define context on the same level is here (not conditionally, though). Then, the below modals (ConvertModal) already uses vars from the context.

DenysKarmazynDFINITY: IMO, having it on this level is actually a nice thing, because here we know exactly what should be converted to what.

peterpeterparker: > ConvertToCkETH or ConverToCkErc20

Not realy, on those we defined the context when the components are mounted no?

In this implementation we rely on a reactive assignment of `ckBtcToken`.

peterpeterparker (approved):

src/frontend/src/btc/components/convert/ConvertToCkBTC.svelte
peterpeterparker: Thanks!

PR #3705, opened 2024-11-20
fix(frontend): glitchy navigation between pages

peterpeterparker:
# Motivation

Not sure exactly why but the navigation between pages from assets to other pages become glitchy again.

# Changes

- Use an animation to display the "Manage tokens" button only on enter

# Screenshots

Before:

demo 18 06 12

After:

after


AntonioVentilii (approved):
LGTM tks!

PR #3708, opened 2024-11-20
ci(release): adjust tag and name of releases

AntonioVentilii:
# Motivation

A few adjustments in the CI workflow that draft the release notes:

- Remove `token` (it is not used).
- Add `tag` and `name`.
- Set the release to be latest and published.


.github/workflows/tag-release.yml
```diff
@@ -36,5 +36,8 @@ jobs:
       id: draft_release
```
bitdivine, line 36: Is it still a draft if publish is set to true?

AntonioVentilii: good point! changed

PR #3710, opened 2024-11-21
refactor(frontend): move ETH tokens to specific file

AntonioVentilii:
# Motivation

It seems a good moment to do this, since we just released a version and theoretically there is a small stasis of PRs.

Basically, I would like to organize the `env` folder, creating two sub-folders: `tokens` and `network`.

But we cannot do that directly, we will do it in small phases (look how many files were changed with just one).

# Changes

- Move ETH native tokens to a file inside env/tokens folder (it seems fair, since BTC has the same).


peterpeterparker (approved):
LGTM, thx

PR #3713, opened 2024-11-21
chore(backend): Upgrade ic-verifiable-credentials crate

lmuntaner:
# Motivation

We want Oisy to also check the "derivation_origin" claim of the ID Alias credential. This check is performed in the latest `ic-verifiable-credentials` version.

In this PR, we upgrade the library to the one published in crates.io.

# Changes

* Upgrade library `ic-verifiable-credentials` to `"1.0.1"`.
* Use the `derivation_origin` in the config to verify the credentials with the library.

# Tests

* Upgrade tests to use a credential that has the new claim `derivation_origin`.


inc-man (approved):
LGTM

PR #3715, opened 2024-11-21
feat(frontend): fetch pending BTC txs in the conversion flow

DenysKarmazynDFINITY:
# Motivation

I noticed that information about pending sent transactions is not fetched in the conversion flow. Adding the missing call.


src/frontend/src/btc/components/convert/BtcConvertTokenWizard.svelte
```diff
@@ -68,6 +69,14 @@
 	// Docs: https://github.com/dfinity/ic/blob/master/rs/bitcoin/ckbtc/minter/README.adoc#bitcoin-to-ckbtc
 	$: destinationAddress = $btcAddressStore?.[$destinationToken.id]?.data;
 
+	onMount(() => {
+		loadBtcPendingSentTransactions({
```
peterpeterparker, line 73: Is it correct to do it that way UX wise, shouldn't the information be loaded before the modal open under the "Hold tight..." backdrop loader?

peterpeterparker: If for some reason the call would actually takes time - think low end device or slow(er) internet connection - would the UI shift and would the UX be impacted by the data being not yet loaded?

peterpeterparker: I see, well not sure. On one hand, as you say, if it's fast, it's fast, on the other, a disabled button without any hint of the why is not necessarely user friendly. Don't know what's best. I approve the PR and let you decide ;)

peterpeterparker: That said, can you add a test that validates the button is disabled until loaded?

DenysKarmazynDFINITY: Tests have been added.

I would say, for the very first version, it should be acceptable. I will keep an eye on it while testing and if I notice something, I'll come back to this topic.

peterpeterparker: Alright. Thanks for the test

peterpeterparker (approved):
LGTM, thx

PR #3716, opened 2024-11-21
style(frontend): special format for positive amount in Transaction

AntonioVentilii:
# Motivation

For the transaction history (all of them), we would like to show the positive amount in green and with the plus sign.

# Changes

- Add optional prop showP`l`usSign to util `formatToken`.
- Add optional prop `formatPositiveAmount` (better name is welcomed) to component `Amount`, since it is used in other parts, not only transactions.
- Adjust `Transaction` component to format positive amounts.

# Tests

Screenshot 2024-11-21 at 10 53 26
Screenshot 2024-11-21 at 10 53 33
Screenshot 2024-11-21 at 10 53 40


src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -29,8 +31,7 @@ export const formatToken = ({
 	if (trailingZeros) {
 		return formatted;
 	}
-
-	return formatted.replace(/\.0+$/, '');
+	return (showPlusSign && +res > 0 ? '+' : '') + formatted;
```
peterpeterparker, line 34: I don't think spontanously we should modify `formatToken` with this. I'm a bit afraid of the side effect. Also not fully convinced that it should find place here.

Are you sure we won't break anything?

src/frontend/src/lib/components/transactions/Transaction.svelte
```diff
@@ -41,7 +41,7 @@
 
 		<svelte:fragment slot="amount">
 			{#if nonNullish(amount)}
-				<Amount {amount} decimals={token.decimals} />
+				<Amount {amount} decimals={token.decimals} formatPositiveAmount />
```
peterpeterparker, line 44: Are you sure the `+` sign is always applicable? Like it's the same regardless of the type of transaction?

src/frontend/src/lib/utils/format.utils.ts
AntonioVentilii: I included some tests and the existing ones are quite extensive (I re-checked them). So, I think we are covered on this (`showPlusSign` defaults to `false` too)

However I understand your point: i was wondering indeed if it should stay all in the component or not; but I thought this could be a cool feature to have in the util, boh

Undecided eheh! Any strong opinion?

src/frontend/src/lib/components/transactions/Transaction.svelte
AntonioVentilii: Good question! According to code we have 7 types of transactions:

- send --> negative
- receive --> positive
- withdraw --> positive --> conversion from ck-token
- deposit --> negative --> conversion to ck-token
- approve --> null (so no new format)
- mint --> positive --> conversion from twin token
- burn --> negative --> conversion to twin token

So, I imagine that all the positive-amount cases fit the new format, right?

src/frontend/src/lib/utils/format.utils.ts
peterpeterparker: I don't like it but, can live with it (particurlaly as I don't have to live too much with it 😉).

peterpeterparker (approved):
LGTm, thx

src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -11,13 +11,15 @@ interface FormatTokenParams {
...
 }
 
 export const formatToken = ({
 	value,
 	unitName = ETHEREUM_DEFAULT_DECIMALS,
 	displayDecimals = DEFAULT_DISPLAY_DECIMALS,
-	trailingZeros = false
+	trailingZeros = false,
+	showPlusSign = false
 }: FormatTokenParams): string => {
```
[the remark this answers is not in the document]
AntonioVentilii: Yes, I like it! I went with

`export type AmountString = `${'+' | '-' | ''}${number}`;`

src/frontend/src/lib/utils/format.utils.ts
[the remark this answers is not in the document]
AntonioVentilii: strange, it didn't.... i will have to double check that rule, but maybe it's because there is some logic?

in any case, I remove the literal

PR #3717, opened 2024-11-21
feat(frontend): do not fetch UTXOS on amount errors

DenysKarmazynDFINITY:
# Motivation

We can avoid fetching utxos fee for a provided amount either `insufficientFunds` or `insufficientFundsForFee` is `true`.


src/frontend/src/btc/components/convert/BtcConvertTokenWizard.svelte
```diff
@@ -52,6 +52,9 @@
 
 	const progress = (step: ProgressStepsConvert) => (convertProgressStep = step);
 
+	let insufficientFunds: boolean;
```
peterpeterparker, line 55: Do we need to pass two property up and down? Or one would fit? I guess depends where the logic finds place best?

src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
```diff
@@ -20,7 +21,7 @@
 			return;
 		}
 
-		if (isNullish(networkId) || isNullish(amount) || amount === 0) {
+		if (amountError || isNullish(networkId) || isNullish(amount) || Number(amount) === 0) {
```
peterpeterparker, line 24: `Number(amount)` is related to the PR?

DenysKarmazynDFINITY: Not really, I didn't notice this change slipped through. Will create a separate PR for that.

src/frontend/src/btc/components/convert/BtcConvertTokenWizard.svelte
DenysKarmazynDFINITY: Just one prop does the job too. Updated.

peterpeterparker (approved):
LGTM, thx

PR #3720, opened 2024-11-21
feat(frontend): add network selector in Activity page

AntonioVentilii:
# Motivation

We would like to include the logic of the NetworkSwitcher in the Activity page.

Basically:

1. It will not appear if testnet is not enabled and show Chain Fusion tokens transactions.
2. It will appear when testnet is enabled and allows to filter transactions per network.

NOTE: By default all networks (Chain Fusion) shows everything except transactions of testnet tokens.

# Changes

- Add `NetworksSwitcher` in unified transaction component.
- Add route parameters breakdown in `+page` of Activity.
- Change tokens store in unified transaction list to be `enabledNetworkTokens`.

# Tests

https://github.com/user-attachments/assets/cf54b381-b2cc-4103-9311-959e602ed967


src/frontend/src/lib/components/transactions/AllTransactions.svelte
```diff
@@ -1,13 +1,21 @@
...
 	import PageTitle from '$lib/components/ui/PageTitle.svelte';
+	import { testnetsEnabled } from '$lib/derived/settings.derived';
 	import { i18n } from '$lib/stores/i18n.store';
 </script>
 
 <div class="flex flex-col gap-5">
 	<PageTitle>{$i18n.activity.text.title}</PageTitle>
 
+	{#if $testnetsEnabled}
+		<div>
```
peterpeterparker, line 14: You need this div?

peterpeterparker (approved):
LGTM, thx

Note: unsolicited question but design wise, have you try to have the switcher after the message box? This way it's closer visually to the data it actually modifies when modified.

src/frontend/src/lib/components/transactions/AllTransactions.svelte
AntonioVentilii: alas yes... otherwise it will use full width and to change it i need to put the div inside and make a prop

(unless you prefer)

AntonioVentilii:
> have you try to have the switcher after the message box?

Good suggestion! I talked with the design team and they told me that for now it is better to leave it there, since there is discussion to hide the message box when it does not concern (for example, hide the BTC disclaimer if you are in Ethereum).

So, visually, it is better to leave it below in that case.

In any case, they will discuss it, and if they decide to not hide the message box, then it makes sense to see if we can move it

AntonioVentilii:
[OISY-409]

src/frontend/src/lib/components/transactions/AllTransactions.svelte
peterpeterparker: I did not get it. Figma screenshots explained it. 😅

PR #3721, opened 2024-11-21
feat(frontend): parse "amount" before using it in UtxosFeeContext

DenysKarmazynDFINITY:
# Motivation

Make sure `amount` is `number` before using it.

### DIFF
--- a/src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
+++ b/src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
@@ -20,6 +20,7 @@
 			return;
 		}
 
+		// we need to make the value is not 0 because the utxos call fails if amount = 0
 		if (isNullish(networkId) || isNullish(amount) || Number(amount) === 0) {
 			store.reset();
 			return;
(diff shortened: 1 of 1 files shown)


src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
```diff
@@ -20,7 +20,7 @@
 			return;
 		}
 
-		if (isNullish(networkId) || isNullish(amount) || amount === 0) {
+		if (isNullish(networkId) || isNullish(amount) || Number(amount) === 0) {
```
peterpeterparker, line 23: Good question actually, is zero acceptable?

Just noticed we have a utils `invalidAmount` which asserts amount is now lower than zero and which seems used in various places.

DenysKarmazynDFINITY: In this particular case - not acceptable as we need to make the value is not 0 because the `utxos` call fails if we request it with `amount = 0`.

peterpeterparker: Thanks for the feedback. Can you add a comment about it?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
DenysKarmazynDFINITY: Yep, added.

PR #3724, opened 2024-11-21
fix(frontend): issue with converting very low numbers to satoshis

DenysKarmazynDFINITY:
# Motivation

I noticed that for some numbers the `convertNumberToSatoshis` was throwing `cannot be converted to a BigInt because it is not an integer` error. This PR fixes it as well as adds some explanations to the code.

### DIFF
--- a/src/frontend/src/tests/btc/utils/btc-send.utils.spec.ts
+++ b/src/frontend/src/tests/btc/utils/btc-send.utils.spec.ts
@@ -6,5 +6,7 @@ describe('convertNumberToSatoshis', () => {
 		expect(convertNumberToSatoshis({ amount: 0.00005 })).toEqual(5000n);
 		expect(convertNumberToSatoshis({ amount: 0.25 })).toEqual(25000000n);
 		expect(convertNumberToSatoshis({ amount: 0 })).toEqual(0n);
+		expect(convertNumberToSatoshis({ amount: 0.00004 })).toEqual(4000n);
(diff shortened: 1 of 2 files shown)


src/frontend/src/tests/btc/utils/btc-send.utils.spec.ts
```diff
@@ -6,5 +6,7 @@ describe('convertNumberToSatoshis', () => {
 		expect(convertNumberToSatoshis({ amount: 0.00005 })).toEqual(5000n);
 		expect(convertNumberToSatoshis({ amount: 0.25 })).toEqual(25000000n);
 		expect(convertNumberToSatoshis({ amount: 0 })).toEqual(0n);
+		expect(convertNumberToSatoshis({ amount: 0.00004 })).toEqual(4000n);
+		expect(convertNumberToSatoshis({ amount: 0.00000001 })).toEqual(1n);
```
AntonioVentilii, line 10: maybe a test for `0.0000399999999999999` ?

AntonioVentilii (approved):
Good catch! Tks! A comment but not blocker, just suggestion

src/frontend/src/tests/btc/utils/btc-send.utils.spec.ts
DenysKarmazynDFINITY: Good idea, done.

PR #3725, opened 2024-11-22
refactor(build): centralize variables for tokens JSON files

AntonioVentilii:
# Motivation

To have better control on them, we extract and centralize some variables that points to the tokens CkERC20 and SNS JSON files in a new file called `constants.mjs`

UNRELATED: there was a non-alias import in one of the scripts.

### DIFF
--- a/scripts/utils.mjs
+++ b/scripts/utils.mjs
@@ -7,7 +7,11 @@ import { readFile } from 'node:fs/promises';
 import { join, resolve } from 'node:path';
 import pemfile from 'pem-file';
 
-const SNS_JSON_FILE = join(process.cwd(), 'src', 'frontend', 'src', 'env', 'tokens.sns.json');
+const DATA_DIR = join(process.cwd(), 'src', 'frontend', 'src', 'env');
+
+export const CK_ERC20_JSON_FILE = join(DATA_DIR, 'tokens.ckerc20.json');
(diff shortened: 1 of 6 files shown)


scripts/utils.mjs
```diff
@@ -7,7 +7,11 @@ import { readFile } from 'node:fs/promises';
 import { join, resolve } from 'node:path';
 import pemfile from 'pem-file';
 
-const SNS_JSON_FILE = join(process.cwd(), 'src', 'frontend', 'src', 'env', 'tokens.sns.json');
+const DATA_DIR = join(process.cwd(), 'src', 'frontend', 'src', 'env');
+
+export const CK_ERC20_JSON_FILE = join(DATA_DIR, 'tokens.ckerc20.json');
```
peterpeterparker, line 12: should we create a `constants.mjs` or `env.mjs` or something? I would spontaneously not search in a utils file since we generally use constants module. WDYT?

AntonioVentilii: Yes, makes sense! Done!

I just didn't move this, because it seems like a non-constants, if you know what i mean... WDYT?

`export const SNSES = JSON.parse((await readFile(SNS_JSON_FILE)).toString(), jsonReviver);`

peterpeterparker: Agree!

peterpeterparker (approved):
LGTM; thx

PR #3726, opened 2024-11-22
style(frontend): fixes hero usd balance

BonomoAlessandro:
# Motivation

The `Hero` should always display the text `$ value is not available` if the usd balance can't be load.

# Changes

- usd balance text

# Tests

**before:**
image

**after:**
image


src/frontend/src/lib/components/tokens/TokenExchangeBalance.svelte
```diff
@@ -14,6 +14,8 @@
 		{formatUSD({ value: usdBalance })}
 	{:else if isNullish(balance)}
 		<span class="animate-pulse">{nullishBalanceMessage ?? '-'}</span>
+	{:else if isNullish(usdBalance) && nonNullish(nullishBalanceMessage)}
+		<span class="animate-pulse">{nullishBalanceMessage}</span>
```
peterpeterparker, line 18: Is it correct that it should be animated with a pulse effect?

AntonioVentilii: adding another comment here, I see that above you do `{nullishBalanceMessage ?? '-'}` and you don't use `nonNullish(nullishBalanceMessage)`

while here the contrary

but in the end, those are the same situations no? why not merging the check?

```typescript
{:else if isNullish(balance) || isNullish(usdBalance)}
		<span class="animate-pulse">{nullishBalanceMessage ?? '-'}</span>
```

BonomoAlessandro: @peterpeterparker yes, it should be animated with a pulse effect. We already use this pulse effect in the check above and here we want to display the same value.

@AntonioVentilii you are right, we can combine those two checks. I just did it and pushed it.

AntonioVentilii (commented):
I defer to @peterpeterparker to continue the review, since he was first. All good for the changes I requested on my side

peterpeterparker (approved):
LGTM, thx

peterpeterparker (commented):
Note that I assume we’re going to test this with some E2E tests soon. Otherwise, it would be nice to provide tests for the new feature and to reduce the test debt.

PR #3728, opened 2024-11-22
feat(frontend): update utxos fee store and context

DenysKarmazynDFINITY:
# Motivation

After testing the conversion flow, I noticed that UTXOS fees are being re-fetched when switching between WizardModal steps. After debugging the whole flow, I noticed that BtcConvertTokenWizard is re-mounted every time a step changes (it actually happens in all places where we use WizardModal). To fix it for the BTC -> ckBTC (where fetching utxos takes ages), I did the following things:

1. Refactored a bit utxos fee store and context, so the related is store is initialised outside of Svelte. It actually simplified all unit tests where it's used.
2. Started to save the inputed amount for which utxos were already fetched and to compare it with the new value to prevent re-fetch if it's the same (additional unit tests are provided).

Declaring utxos-fee context outside of WizardModal would also solve the problem, but the parent would need to know too much data that is set in children components (sendAmount, amountErrors, etc).

### DIFF
--- a/src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
+++ b/src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
@@ -31,7 +31,7 @@
 
 		// WizardModal re-renders content on step change (e.g. when switching between Convert to Review steps)
 		// To avoid re-fetching the fees, we need to check if amount hasn't changed since the last request
-		if (nonNullish($store) && $store.amount === parsedAmount) {
+		if (nonNullish($store) && $store.amountForFee === parsedAmount) {
 			return;
 		}
 
@@ -52,7 +52,7 @@
 
 		store.setUtxosFee({
 			utxosFee,
-			amount: parsedAmount
+			amountForFee: parsedAmount
 		});
 	};
 
--- a/src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
+++ b/src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
@@ -21,9 +21,17 @@
 			return;
 		}
 
+		const parsedAmount = nonNullish(amount) ? Number(amount) : undefined;
+
 		// we need to make the value is not 0 because the utxos call fails if amount = 0
-		if (amountError || isNullish(networkId) || isNullish(amount) || Number(amount) === 0) {
+		if (amountError || isNullish(networkId) || isNullish(parsedAmount) || parsedAmount === 0) {
 			store.reset();
+			return;
+		}
+
+		// WizardModal re-renders content on step change (e.g. when switching between Convert to Review steps)
+		// To avoid re-fetching the fees, we need to check if amount hasn't changed since the last request
+		if (nonNullish($store) && $store.amount === parsedAmount) {
 			return;
 		}
 
@@ -31,7 +39,7 @@
 
 		const utxosFee = nonNullish(network)
 			? await selectUtxosFeeApi({
-					amount,
+					amount: parsedAmount,
 					network,
 					identity: $authIdentity
 				})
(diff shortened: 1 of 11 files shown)


src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
AntonioVentilii, line 1: more of a doubt here: before you were passing non-parsed `amount`. Why parsing now? I mean, can you not use `amount` purely, and store it in the Utxos store as it is? then compare it?

src/frontend/src/btc/stores/utxos-fee.store.ts
```diff
@@ -32,3 +34,5 @@ export interface UtxosFeeContext {
 }
 
 export const UTXOS_FEE_CONTEXT_KEY = Symbol('utxos-fee');
+
+export const utxosFeeStore = initUtxosFeeStore();
```
peterpeterparker, line 38: Do I understand correctly that your proposed solution is to turn the utxosFeeStore into a global store instead of a context-specific store? Spontaneously, this feels like a shortcut, but I’m unsure of the implications. Is it appropriate to have UTXO fees stored globally when no wizard is open? If someone accesses this information in such cases, what would it represent or imply?

src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
DenysKarmazynDFINITY: IMO, comparing numberish values seems a bit safer when done with `number` type instead of `string`. 

Why having a new variable - it's mostly to parse the value just once, instead of having `Number(amount)` about 3-4 times.

src/frontend/src/btc/stores/utxos-fee.store.ts
DenysKarmazynDFINITY: Yes, correct, I switched it to be a global store, since I did not find any side effects or implications. True, it can be now accessed outside of the wizard, but the data stored there (`undefined` or `utxosFee` for `amount`) is not very meaningful outside of BTC conversion or send flows. Even though it would be definitely nicer to keep it scoped, the pattern I'm using in this PR is already in the codebase (other BTC/ETH/ICP stores), so I think it already proved to be working.

peterpeterparker: Thanks for the explanation. There is no side effects, for now 😉. In the future do you foreseen any use cases that would justify having this store being available globally? If not, is there a way to resolve the issue while keep a context store?

peterpeterparker: Cool, thanks! Added two minor comments but, looks good!

src/frontend/src/btc/stores/utxos-fee.store.ts
```diff
@@ -1,9 +1,11 @@
 import type { UtxosFee } from '$btc/types/btc-send';
+import type { OptionAmount } from '$lib/types/send';
 import type { Option } from '$lib/types/utils';
 import { writable, type Readable } from 'svelte/store';
 
 export type UtxosFeeStoreData = Option<{
 	utxosFee?: UtxosFee;
+	amount?: OptionAmount;
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Yep, makes sense: done and done.

peterpeterparker (approved):
LGTm, thx

src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
DenysKarmazynDFINITY: @AntonioVentilii really sorry, I forgot about this thread and merged the code. Please take a look at my answer, and if we need to apply some changes, I'll create a new PR.

PR #3734, opened 2024-11-25
feat(frontend): persist network params in the navigation

AntonioVentilii:
# Motivation

We would like to persist the selected network network during navigation. So, if users enable testnets and select network A in the assets page, when they move to activity they have the same network selected. But even moving around the site, the network is persisted.

However, once the network selector is disabled (testnet disabled), we should reset the behavior, since otherwise the user could be "stuck " with a network and could not change it.

Furthermore, there is the edge case:

- user DOES not select a network in the Assets page;
- user clicks on a token;
- user clicks on on Activity;
- the token's network is persisted, and that is not what we want.

# Changes

- Apply function to join a path with a selected network to components Menu and NavigationMenu.

# Tests

https://github.com/user-attachments/assets/abe8ff08-e41c-43e0-b61a-6ef53d778c78


AntonioVentilii:
@peterpeterparker as discussed, I reduced this PR, ready for a review now

src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -19,37 +20,57 @@
...
+	let isTransactionsRoute = false;
+	$: isTransactionsRoute = isRouteTransactions($page);
+
 	const hidePopover = () => (visible = false);
 
 	const goToTokens = async () => {
 		hidePopover();
-		await goto(AppPath.Tokens);
+		await goto(
+			networkUrl({ path: AppPath.Tokens, networkId: $networkId, isTransactionsRoute, fromRoute })
```
peterpeterparker, line 51: Thinking at it now, maybe the parameter `isTransactionsRoute` should not be called like that but be more generic like "resetNetwork` or something. Not related to this PR, just thinking out at loud.

src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -19,37 +20,57 @@
...
+		await goto(
+			networkUrl({ path: AppPath.Explore, networkId: $networkId, isTransactionsRoute, fromRoute })
+		);
 	};
 
 	const goToActivity = async () => {
 		hidePopover();
-		await goto(AppPath.Activity);
+		await goto(
+			networkUrl({ path: AppPath.Activity, networkId: $networkId, isTransactionsRoute, fromRoute })
```
peterpeterparker, line 72: Maybe we can extract a local function in the component.

```
const navigate = async (path: AppPath) => {
   await goto(
			networkUrl({ path: AppPath.Activity, networkId: $networkId, isTransactionsRoute, fromRoute })
   hidePopover();
```

That way the code has less code duplication

```
const goToActivity = async () => navigate(AppPath.Activity)

```

Not for this PR neither, thinking out at loud too.

peterpeterparker (approved):
LGTm, thx

src/frontend/src/lib/components/core/Menu.svelte
AntonioVentilii: good point! it should indicate that "it should use the network id that was in the page before"

src/frontend/src/lib/components/core/Menu.svelte
AntonioVentilii: Done with:

- PR #3763

PR #3735, opened 2024-11-25
feat(frontend): util to create URL with network param

AntonioVentilii:
# Motivation

Splitting PR #3734 into smaller parts, we use this one to create an util that defined the URL for a selected network.

We would like to persist the selected network network during navigation. So, if users enable testnets and select network A in the `Assets` page, when they move to `Activity`, they have the same network selected. But even moving around the site, the network should be persisted.

However, there is the edge case:

- user DOES not select a network in the `Assets` page;
- user clicks on a token;
- user clicks on on `Activity`;
- the token's network is persisted, and that is not what we want.

So for the `Transactions` page, we use the history instead of the selected network.

# Changes

- Create function to join a path with a selected network, given the history too in case of Transactions page.

# Tests

Provided some tests

### DIFF
--- a/src/frontend/src/lib/utils/nav.utils.ts
+++ b/src/frontend/src/lib/utils/nav.utils.ts
@@ -12,7 +12,7 @@ import type { OptionString } from '$lib/types/string';
 import type { Token } from '$lib/types/token';
 import type { Option } from '$lib/types/utils';
 import { isNullish, nonNullish } from '@dfinity/utils';
-import type { LoadEvent, Page } from '@sveltejs/kit';
+import type { LoadEvent, NavigationTarget, Page } from '@sveltejs/kit';
 
 export const transactionsUrl = ({ token }: { token: Token }): string =>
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -48,6 +48,25 @@ const tokenUrl = ({
...
+	isTransactionsRoute,
+	fromRoute
+}: {
+	path: AppPath;
+	networkId: Option<NetworkId>;
+	isTransactionsRoute: boolean;
+	fromRoute: NavigationTarget | null;
+}) =>
+	isTransactionsRoute
+		? nonNullish(fromRoute?.url.searchParams.get(NETWORK_PARAM))
```
peterpeterparker, line 63: `notEmptyString` instead of `nonNullish` maybe? not sure if that would be an issue to have an empty param but, the url could be a bit more cute?

peterpeterparker (approved):
LGTM, thx

A small comment, already approving.

PR #3736, opened 2024-11-25
feat(frontend): reset network param when turning testnets off

AntonioVentilii:
# Motivation

Since ,with PR #3734, we want to persist the selected network during navigation, it makes sense to reset the network param when the network selector is disabled. That means, when the testnets are disabled.

# Changes

- Delete NETWORK_PARAM when toggle is turned off.
- Replace state.
- Add TODO for future tests.

# Tests

https://github.com/user-attachments/assets/5ad53784-3c69-4e49-9dba-f6bf3639cdcd

### DIFF
--- a/src/frontend/src/lib/components/networks/NetworksTestnetsToggle.svelte
+++ b/src/frontend/src/lib/components/networks/NetworksTestnetsToggle.svelte
@@ -3,11 +3,25 @@
 	import { testnetsEnabled } from '$lib/derived/settings.derived';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { testnetsStore } from '$lib/stores/settings.store';
+	import { NETWORK_PARAM } from '$lib/constants/routes.constants';
+	import { goto } from '$app/navigation';
+
+	// TODO: create tests for this component once we have testId from GIX-CMP
+	// PR: https://github.com/dfinity/gix-components/pull/531
+
 
 	let checked: boolean;
 	$: checked = $testnetsEnabled;
 
-	const toggleTestnets = () => testnetsStore.set({ key: 'testnets', value: { enabled: !checked } });
+	const toggleTestnets = async () => {
+		testnetsStore.set({ key: 'testnets', value: { enabled: !checked } });
+		// Reset network param, since the network is selectable only when testnets are enabled
+		if (checked) {
+			const url = new URL(window.location.href);
+			url.searchParams.delete(NETWORK_PARAM);
+			await goto(url, { replaceState: true, noScroll: true });
+		}
+	};
 </script>
 
 <Toggle ariaLabel={$i18n.settings.alt.testnets_toggle} bind:checked on:nnsToggle={toggleTestnets} />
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/networks/NetworksTestnetsToggle.svelte
```diff
@@ -1,13 +1,27 @@
...
+	// TODO: create tests for this component once we have testId from GIX-CMP
+	// PR: https://github.com/dfinity/gix-components/pull/531
+
 	let checked: boolean;
 	$: checked = $testnetsEnabled;
 
-	const toggleTestnets = () => testnetsStore.set({ key: 'testnets', value: { enabled: !checked } });
+	const toggleTestnets = async () => {
+		// Reset network param, since the network is selectable only when testnets are enabled
+		if (checked) {
```
peterpeterparker, line 17: I would maybe do this after the store is updated, not sure what's the most important?

src/frontend/src/lib/components/networks/NetworksTestnetsToggle.svelte
```diff
@@ -1,13 +1,27 @@
...
+			const url = new URL(window.location.href);
```
peterpeterparker, line 18: Either catch exception or use `URL.parse`

src/frontend/src/lib/components/networks/NetworksTestnetsToggle.svelte
```diff
@@ -1,13 +1,27 @@
...
+			url.searchParams.delete(NETWORK_PARAM);
+			await goto(url, { replaceState: true, noScroll: true });
```
peterpeterparker, line 20: I ask myself if we should use `goto` or imperatively remove the param for the URL because, does `goto` implies that some stores will be reevaluted including `testnetsStore` twice with next line?

src/frontend/src/lib/components/networks/NetworksTestnetsToggle.svelte
AntonioVentilii: sure!

src/frontend/src/lib/components/networks/NetworksTestnetsToggle.svelte
AntonioVentilii: i think that doing it imperatively will not "refresh" some data of the `page` store; BUT, I am not sure we need the `page` store in our cases; I just wanted to give a clean "refresh", let's say

src/frontend/src/lib/components/networks/NetworksTestnetsToggle.svelte
AntonioVentilii: done!

src/frontend/src/lib/components/networks/NetworksTestnetsToggle.svelte
peterpeterparker: Yeah but that's the thing, a "clean refresh" is a really different behavior than the current one, which is just updating the testnet store that's why to be honest I lean to "just" updating the URL here.

peterpeterparker: Discussed offline, I did not understood. `goto` is a must here as we really want to update the store since we remove the network information.

AntonioVentilii: As discussed offline, we will got with goto because we actually need to trigger a refresh of the `page` store of svelte, even if it costs to refresh derived that we don't need

peterpeterparker (approved):
LGTM, thx

PR #3737, opened 2024-11-25
fix(frontend): burn token usd value

loki344:
# Motivation

Currently the BURN USD token value is wrong (and therefore all token values of the additional icrc token json). It does not reflect the token value. This is caused by two bugs:

1. The exchangeCoinId is hard coded to internet-computer for icrc tokens originating from the JSON. This is wrong.
2. The usd rate is not loaded for ICRC_TOKENS_LEDGER_CANISTER_IDS, which all additional ICRC token happen to be in (although they should not be in there).

# Changes

- remove hardcoded exchangeCoinId from [map-icrc-data.ts](https://github.com/dfinity/oisy-wallet/pull/3737/files#diff-41718944cc5a4f359f98739e8ffd31411dd00e18969b58470da113a699f9ca89)
- rename and correct the content of ICRC_TOKENS_LEDGER_CANISTER_IDS to ICRC_**CK**_TOKENS_LEDGER_CANISTER_IDS to make it more clear. This list is currently only used in the derived enabledIcrcTokensNoCk, which indicates that the original intention behind this list was to have only CK canisterIds

# Tests
before
image

after
image


peterpeterparker (commented):
Can you split the two different things addressed / implemented in this PR in two PRs? Assuming the "remove" and "rename" described in the PR description can be split.

loki344:
> Can you split the two different things addressed / implemented in this PR in two PRs? Assuming the "remove" and "rename" described in the PR description can be split.

Yes, I hope this does not complicate the hotfix part.

loki344:
https://github.com/dfinity/oisy-wallet/pull/3740

loki344:
an

PR #3739, opened 2024-11-25
style(frontend): improves transaction lists

BonomoAlessandro:
# Motivation

In the transaction lists the token ticket should also be displayed (e.g. 0.00123 -> 0.00123 ETH). In addition the ck-Token symbols should always start lowercase.

# Changes

- adds token ticket to the transactions in the transaction lists
- capitalizes the first letter of the token label only on fallback

# Tests

**before:**
activity:
image

token view:
image

**after:**
activity:
image

token view:
image


AntonioVentilii (commented):
Can you split into 2 PRs please? One for the amount and one for the label?

AntonioVentilii (approved):
As discussed offline, LGTM tks!

PR #3742, opened 2024-11-25
feat(frontend): use only groups in TokensDisplayHandler

AntonioVentilii:
# Motivation

We want to use util created in PR #3387 to group the tokens. The main change is that instead of returning a mix group of tokens and token-groups , we return only token groups.

Then, we treat differently the groups with only one token, showing them as single-token and not groups.

# Changes

- Use function `groupTokens` in component TokensDisplayHandler, instead of `groupTokensByTwin`.
- Change the name and typing of the return array.
- Change component `TokensList` to show the group component for groups and a single card for the single token.
- Remove deprecated functions and types.

# Tests

https://github.com/user-attachments/assets/ed7e627a-df90-4dae-8cb4-11bc0aff321f


src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
```diff
@@ -3,21 +3,21 @@
...
 	import { showZeroBalances } from '$lib/derived/settings.derived';
 	import type { TokenUi } from '$lib/types/token';
-	import type { TokenUiOrGroupUi } from '$lib/types/token-group';
-	import { groupTokensByTwin } from '$lib/utils/token-group.utils';
+	import type { TokenUiGroup } from '$lib/types/token-group';
+	import { groupTokens } from '$lib/utils/token-group.utils';
 
 	// We start it as undefined to avoid showing an empty list before the first update.
-	export let tokens: TokenUiOrGroupUi[] | undefined = undefined;
+	export let tokenGroups: TokenUiGroup[] | undefined = undefined;
```
peterpeterparker, line 10: It would be kind of nice to keep a variant group or token WDYT? Feel like it's a bit more more maintainable. Or am I overthinking it? 

I mean something like

```
export type TokenUiOrGroupUi = TokenUi | TokenUi[]

// or even

export type TokenUiOrGroupUi = {token: TokenUi} | {group: GroupUi }
```

You know what I mean?

Again maybe I'm overengeering...

AntonioVentilii: Ah, understood! However, there is no real application for now: the list passed to Token list is just `TokenGroups`

If we add it, should I decompose the groups?

For example `groupTokens(sortedTokens).reduce<{token: TokenUi} | {group: GroupUi }> ( ... )` ?

peterpeterparker: Maybe no real application but, this PR would introduces logic such as `{@const token = tokenGroup.tokens[0]}` and I'm not sure I'm absolutely into it...

> For example groupTokens(sortedTokens).reduce<{token: TokenUi} | {group: GroupUi }> ( ... ) ?

Yes I would say so, do you think it's overeenginered?

I'm afraid that we start iterating on this and there we start having such logic as `tokens[0]` all over the place.

AntonioVentilii: nono, i agree that tokens[0] is not good; however I need to re-think better how to do it, since we would be doing 2 loops in this way, 2 reduces. I need to think how to merge it

AntonioVentilii:
Superseded by:

- https://github.com/dfinity/oisy-wallet/pull/4981

PR #3743, opened 2024-11-25
fix(frontend): fix hardcoded exchangecoinid

loki344:
# Motivation

IcrcData from the json is wrongly mapped with an exchangeCoinId hardcoded to internet-computer. This is wrong and causes the fallback logic of exchange balance to default to the ICP price.

# Changes

remove hardcoded value

### DIFF
--- a/src/frontend/src/tests/icp/utils/map-icrc-data.spec.ts
+++ b/src/frontend/src/tests/icp/utils/map-icrc-data.spec.ts
@@ -10,7 +10,7 @@ describe('mapIcrcData', () => {
 	};
 
 	const expected = {
-		TESTTOKEN: { ...token.TESTTOKEN}
+		TESTTOKEN: { ...token.TESTTOKEN }
 	};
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/tests/icp/utils/map-icrc-data.spec.ts
```diff
@@ -10,7 +10,7 @@ describe('mapIcrcData', () => {
 	};
 
 	const expected = {
-		TESTTOKEN: { ...token.TESTTOKEN, exchangeCoinId: 'internet-computer' }
+		TESTTOKEN: { ...token.TESTTOKEN }
```
AntonioVentilii, line 13: shall this be simplified `const expected = { TESTTOKEN: token.TESTTOKEN }`

peterpeterparker (approved):
Good catch, thanks!

AntonioVentilii (approved):
LGTM tks

PR #3744, opened 2024-11-25
fix(frontend): improve the naming and fix the content of ICRC Tokens ledger canister id

loki344:
# Motivation

The original intention of the list ICRC_TOKENS_LEDGER_CANISTER_IDS implied by its only usage in enabledIcrcTokensNoCk was to filter out Ck canister ids. By adding additional ICRC tokens to ICRC_TOKENS and the following mapping of the canister ids to the list, causes a bug:
There will be no usd exchange rates for these tokens as they are not included in the enabledIcrcTokensNoCk store.
This bug, combined with the wrong fallback logic corrected in https://github.com/dfinity/oisy-wallet/pull/3740 led to a false usd value of additional ICRC tokens.
# Changes

Rename the variable ICRC_TOKENS_LEDGER_CANISTER_IDS to reflect its only usage of holding CK token ledger canister ids
Fix the content of this list to include CK tokens.

# Tests

image

### DIFF
--- a/src/frontend/src/env/networks.icrc.env.ts
+++ b/src/frontend/src/env/networks.icrc.env.ts
@@ -458,7 +458,7 @@ const ICRC_CK_TOKENS: IcInterface[] = [
 	...(nonNullish(CKUNI_IC_DATA) ? [CKUNI_IC_DATA] : []),
 	...(nonNullish(CKEURC_IC_DATA) ? [CKEURC_IC_DATA] : []),
 	...(nonNullish(CKXAUT_IC_DATA) ? [CKXAUT_IC_DATA] : [])
-]
+];
 
 const ADDITIONAL_ICRC_TOKENS: IcInterface[] = [...(nonNullish(BURN_IC_DATA) ? [BURN_IC_DATA] : [])];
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/env/networks.icrc.env.ts
```diff
@@ -464,9 +460,17 @@ export const ICRC_TOKENS: IcInterface[] = [
 	...(nonNullish(CKXAUT_IC_DATA) ? [CKXAUT_IC_DATA] : [])
 ];
 
-export const ICRC_TOKENS_LEDGER_CANISTER_IDS: LedgerCanisterIdText[] = ICRC_TOKENS.map(
-	({ ledgerCanisterId }) => ledgerCanisterId
-);
+const ADDITIONAL_ICRC_TOKENS: IcInterface[] = [...(nonNullish(BURN_IC_DATA) ? [BURN_IC_DATA] : [])];
+
+export const ICRC_TOKENS: IcInterface[] = [
+	...PUBLIC_ICRC_TOKENS,
```
loki344, line 466: Should we also rename this to PUBLIC_CK_ICRC_TOKENS as it only holds CKBTC, CKETH and CKUSDC at the moment?

AntonioVentilii (approved):
Approving with a note, please have a look

PR #3748, opened 2024-11-25
test(frontend): add tests for BtcSendReview components

DenysKarmazynDFINITY:
# Motivation

As agreed, before implementing updates to a component, covering it with unit tests first. In this case, it's about BtcSendReview component. In the next PRs, I will be adding an additional check for the "insufficient-fund-for-fee" error.

### DIFF
--- a/src/frontend/src/tests/btc/components/send/BtcSendReview.spec.ts
+++ b/src/frontend/src/tests/btc/components/send/BtcSendReview.spec.ts
@@ -46,7 +46,7 @@ describe('BtcSendReview', () => {
 		mockPage.reset();
 	});
 
-	it('should keep the next button clickable if all requirements are met', () => {
+	it('should enable the next button if all requirements are met', () => {
 		mockBtcPendingSendTransactionsStatusStore();
 
 		const { getByTestId } = render(BtcSendReview, {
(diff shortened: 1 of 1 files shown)


src/frontend/src/tests/btc/components/send/BtcSendReview.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+			.spyOn(btcPendingSendTransactionsStatusStore, 'initPendingSentTransactionsStatus')
+			.mockImplementation(() => readable(status));
+
+	const buttonTestId = REVIEW_FORM_SEND_BUTTON;
+
+	beforeEach(() => {
+		mockPage.reset();
+	});
+
+	it('should keep the next button clickable if all requirements are met', () => {
```
peterpeterparker, line 49: nitpick: not sure why you use "should keep" in the titles. 

e.g. 

"should keep the next button clickable if all requirements are met"

vs

"should enable the next button if all requirements are met"

peterpeterparker (approved):
Thanks!

src/frontend/src/tests/btc/components/send/BtcSendReview.spec.ts
DenysKarmazynDFINITY: Updated 👍

PR #3749, opened 2024-11-25
feat(frontend): create and use InsufficientFundsForFee component

DenysKarmazynDFINITY:
# Motivation

The same UI component can be re-used in both convert and send flow. In this PR, I'm extracting the code as well as using the new component in BtcConvertForm.


peterpeterparker:
A test is failing. Ping me when it passes but, looks good!

DenysKarmazynDFINITY:
@peterpeterparker should be better now.

PR #3755, opened 2024-11-26
feat(frontend): handle insufficientFundsForFee case in BtcSendReview

DenysKarmazynDFINITY:
# Motivation

This PR fixes issue described [here](https://github.com/dfinity/oisy-wallet/issues/3480#issuecomment-2491849981) - there was no check in the BTC send flow which makes sure that provided amount and related utxos fee do not exceed available balance. 

The component now checks `insufficientFundsForFee` case, disabled "send" button if it's true, as well as shows the following user message:

error

### DIFF
--- a/src/frontend/src/btc/components/send/BtcSendReview.svelte
+++ b/src/frontend/src/btc/components/send/BtcSendReview.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
-	import { isNullish } from '@dfinity/utils';
+	import { isNullish, nonNullish } from '@dfinity/utils';
+	import { getContext } from 'svelte';
 	import type { Readable } from 'svelte/store';
 	import BtcReviewNetwork from '$btc/components/send/BtcReviewNetwork.svelte';
 	import BtcSendWarnings from '$btc/components/send/BtcSendWarnings.svelte';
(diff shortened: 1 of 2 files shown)


src/frontend/src/btc/components/send/BtcSendReview.svelte
```diff
@@ -21,15 +25,30 @@
...
+					value: `${amount}`,
+					unitName: $sendTokenDecimals
+				})
+					.add(utxosFee.feeSatoshis)
+					.gt($sendBalance)
+			: false;
+
 	let disableSend: boolean;
 	// We want to disable send if pending transactions or UTXOs fee isn't available yet, there was an error or there are pending transactions.
 	$: disableSend =
```
peterpeterparker, line 47: Unrelated: In a follow up PR, should we extract this logic in some function? It is becoming quite complex. WDY?

src/frontend/src/btc/components/send/BtcSendReview.svelte
```diff
@@ -46,9 +65,11 @@
 
 	<BtcUtxosFee slot="fee" bind:utxosFee {networkId} {amount} />
 
-	<BtcSendWarnings
-		slot="info"
-		{utxosFee}
-		pendingTransactionsStatus={$hasPendingTransactionsStore}
-	/>
+	<svelte:fragment slot="info">
```
peterpeterparker, line 68: Should we add some fade effect?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/btc/components/send/BtcSendReview.svelte
DenysKarmazynDFINITY: Both components already have animations in place: `InsufficientFundsForFee` uses `slide`, and `BtcSendWarnings` - `fade`. I think it's something to be aligned when implementing the new Send flow.

src/frontend/src/btc/components/send/BtcSendReview.svelte
DenysKarmazynDFINITY: Yes, makes sense!

src/frontend/src/btc/components/send/BtcSendReview.svelte
peterpeterparker: Coolio 👍

PR #3758, opened 2024-11-26
feat(frontend): different warning for non-working index canister

AntonioVentilii:
# Motivation

For ICRC tokens that have the index canister but it is not working, we want to show a different message.

# Changes

- Add new icon and new warning message.
- Change `IcNoIndexPlaceHolder` to accept a prop that defines if the index is either `missing` or `not-working`.
- Adapt `IcTransactions` component.

# Tests

I simulated `ckSepoliaUSDC` with a wrong index canister,

Screenshot 2024-11-26 at 13 07 53


src/frontend/src/icp/components/transactions/IcNoIndexPlaceholder.svelte
```diff
@@ -1,21 +1,28 @@
 <script lang="ts">
+	import IconAlertTriangle from '$lib/components/icons/lucide/IconAlertTriangle.svelte';
 	import IconSearchClose from '$lib/components/icons/lucide/IconSearchClose.svelte';
 	import RoundedIcon from '$lib/components/ui/RoundedIcon.svelte';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+
+	export let issueType: 'missing' | 'not-working' = 'missing';
```
peterpeterparker, line 8: `placeholderType` maybe? Just thinking that this component is a placeholder so not strictly related to why if that makes sense?

src/frontend/src/icp/components/transactions/IcNoIndexPlaceholder.svelte
```diff
@@ -1,21 +1,28 @@
...
 	</div>
 
 	<div class="space-y-4">
 		<p class="m-0 text-center text-lg font-bold">
 			{$i18n.transactions.text.transaction_history_unavailable}
 		</p>
 		<p class="text-secondary m-0 text-center opacity-50">
-			{replaceOisyPlaceholders($i18n.transactions.text.missing_index_canister_explanation)}
+			{replaceOisyPlaceholders(
+				issueType === 'not-working'
```
peterpeterparker, line 22: Should we add a "unit" test of its something we can easily test with E2E?

src/frontend/src/icp/components/transactions/IcTransactions.svelte
```diff
@@ -101,6 +101,9 @@
 
 	let noTransactions = false;
 	$: noTransactions = nonNullish($token) && $icTransactionsStore?.[$token.id] === null;
+
+	let hasIndexCanister = false;
+	$: hasIndexCanister = nonNullish($tokenAsIcToken) && nonNullish($tokenAsIcToken.indexCanisterId);
```
peterpeterparker, line 106: Not sure that's a good idea but, maybe there is a Zod.safeParse available? But again not sure makes sense...

AntonioVentilii: Yes, there is `isIcTokenCanistersStrict` , I can use that, is that what you mean?

src/frontend/src/icp/components/transactions/IcNoIndexPlaceholder.svelte
AntonioVentilii: makes sense, changed

src/frontend/src/icp/components/transactions/IcTransactions.svelte
peterpeterparker: Yes it's what I meant but, makes sense to you?

src/frontend/src/icp/components/transactions/IcNoIndexPlaceholder.svelte
AntonioVentilii: I added a some unit tests, since the E2E tests are a bit more complex (for another PR).

I put the E2E test for this in the list of E2E tests that should be implemented

src/frontend/src/icp/components/transactions/IcTransactions.svelte
peterpeterparker: So we keep it as it or you forgot pushing before pinging me for a new review?

No strong opinion, just asking.

AntonioVentilii: Yes yes, makes sense! Done!

src/frontend/src/icp/components/transactions/IcTransactions.svelte
```diff
@@ -52,6 +54,9 @@
 
 	let noTransactions = false;
 	$: noTransactions = nonNullish($token) && $icTransactionsStore?.[$token.id] === null;
+
+	let hasIndexCanister = false;
```
peterpeterparker, line 58: Isn't `isIcTokenCanistersStrict($tokenAsIcToken ?? ICP_TOKEN)` returning `true` per default which is the contrary of `hasIndexCanister = false;` -> should it be `hasIndexCanister = true;`

AntonioVentilii: ah, you mean for the fallback `ICP_TOKEN`, right? I will add a nullish check before

AntonioVentilii: In general, I think the token should be defined here alway, but we still use the token store

peterpeterparker: No I mean, if `$tokenAsIcToken === null` then it's `ICP_TOKEN` that is used and `isIcTokenCanistersStrict(ICP_TOKEN)` equals `true` so when the component is mounted without the `tokenAsIcToken` we got true fals

```
let hasIndexCanister = false; // <---- False on mount
hasIndexCanister = isIcTokenCanistersStrict($tokenAsIcToken ?? ICP_TOKEN) // <---- True if undefined on mount
```

e.g. shouldnt it be

```
let hasIndexCanister = true; // <---- true 
hasIndexCanister = isIcTokenCanistersStrict($tokenAsIcToken ?? ICP_TOKEN) // <---- True
```

But maybe I'm overseeing something?

AntonioVentilii: yes yes! i had understood exactly that! i fixed it with the `nonNullish` before and removing `ICP_TOKEN`? Is that ok?

AntonioVentilii: now I did 😉

peterpeterparker (approved):
LGTM, thx

PR #3759, opened 2024-11-26
style(frontend): fixes activity transaction amount

BonomoAlessandro:
# Motivation

BTC transactions displayed on the activity page showed a wrong calculated amount while the same transaction on the asset page showed the correct amount.

# Changes

- uses the right token to calculate the amount to display

# Tests

**before:**
transaction page:
image

activity page:
image

**after:**
transaction page:
image

activitiy page:
image


peterpeterparker (approved):
LGTM, thx

If you want to add a test, not against

PR #3760, opened 2024-11-26
test(e2e): add send and receive test

inc-man:
# Motivation

To further improve testing, we want to check if the receive and send tokens functionality is provided as intended.

# Changes

configured a complete test that receives and sends tokens from a wallet.

# Tests

tested thoroughly on local replica


src/frontend/src/icp/components/send/IcSendForm.svelte
```diff
@@ -29,8 +34,26 @@
...
+		<SendSource token={$sendToken} balance={$balance} {source} />
+
+		<IcFeeDisplay {networkId} />
+
+		<ButtonGroup slot="toolbar">
+			<slot name="cancel" />
+			<ButtonNext disabled={invalid} testId={SEND_FORM_NEXT_BUTTON} />
+		</ButtonGroup>
+	</ContentWithToolbar>
+</form>
```
AntonioVentilii, line 56: what is this part?

AntonioVentilii (commented):
do we need `dockerode`? don't we have enough tools to run a docker image or something similar? and in case not, shouldn't that go among the devDependecies?

PR #3761, opened 2024-11-26
feat(frontend): fallback for non-working index canisters

AntonioVentilii:
# Motivation

It may happen that an ICRC token has an Index Canister but it is not working. For example, it is out of cycles.

In those cases, we don't want to show an error anymore, but leverage the same wallet service used for tokens without Index Canister.

That means, that in those cases, we will just fetch the balance, and show an explanatory warning for not having the transacitons.

# Changes

- Change function `onLoadTransactionsError` to accept a fallback function.
- If the error is raised during the ICRC transaction fetching, the fallback restarts the worker, but providing only the lLedger Canister, and not the Index Canister. This will allow the service to treat it in the same way of a token without index canister (i.e. fetching balance only).

# Note

The changes in the UI will be done with:

- PR #3758 

# Tests

I simulated a wrong index canister in the local replica for ckSepoliaUSDC. In the video we can see how I switch between the correct Index canister and the wrong one, back and forth.

https://github.com/user-attachments/assets/f007e592-bd26-4dcb-88b4-b6cbb7b19d33


src/frontend/src/icp/services/worker.icrc-wallet.services.ts
```diff
@@ -42,8 +51,14 @@ export const initIcrcWalletWorker = async ({
...
 				onLoadTransactionsError({
 					tokenId,
-					error: (data.data as PostMessageDataResponseError).error
+					error: (data.data as PostMessageDataResponseError).error,
+					silent: true
 				});
+
+				// In case of error, we start the listener again, but only with the ledgerCanisterId,
+				// to make it request only the balance and not the transactions
+				restartWorkerWithLedgerOnly();
```
AntonioVentilii, line 60: @peterpeterparker as discussed; do you think it is too much to have an internal function? I would not externalize it, just because I need to pass all props (worker, ledger, env)

peterpeterparker (approved):
LGTM, thx

PR discussed and reviewed offline as well, cool stuff!

PR #3768, opened 2024-11-27
feat(backend): add types for user dApp

AntonioVentilii:
# Motivation

We want to save in the backend some configurations for the dApps specific to each user. So that it is persisted throughout the accesses. The first one is the fact that the user would not want to see the dApp in the carousel anymore.

# Changes

We replicate the structure of the user tokens, but I am not sure if it is overkill or not.

- Create version class
- Create dApp classes

# Tests

No test needed for now.

### DIFF
--- a/src/shared/src/impls.rs
+++ b/src/shared/src/impls.rs
@@ -1,16 +1,19 @@
 use crate::types::custom_token::{CustomToken, CustomTokenId, Token};
+use crate::types::dapp::Dapp;
 use crate::types::token::UserToken;
 use crate::types::user_profile::{
     AddUserCredentialError, OisyUser, StoredUserProfile, UserCredential, UserProfile,
 };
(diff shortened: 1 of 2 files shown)


peterpeterparker (commented):
I let Max review the PR, but, out of curiosity, what's your plan: save those dapps within the profile, or as a separate entity with distinct endpoints?

AntonioVentilii:
My plan would be different endpoints, not in the user settings/profile, because I thought that in the future we could have more things for the dapps. But WDYT? Maybe simplify and add in the user profile a list of "disabled carousel apps"?

peterpeterparker:
New endpoints = new query+update (maybe ?) = more stress on the network and on the performance but, indeed if it's suppose to grow... Probably something to discuss offline if worth discussing. Thanks for the feedback.

AntonioVentilii:
As discussed with both @peterpeterparker and @bitdivine , I will follow another approach: I will store the information inside the user profile. So closing this one

PR #3769, opened 2024-11-27
build(frontend): npm audit cookie

peterpeterparker:
# Motivation

There is a low security issue about `cookie` which is reported by `npm audit fix` but, won't be addressed in current version of SvelteKit (I reported the security [issue](https://github.com/sveltejs/kit/security/advisories/GHSA-fcpm-cfj3-v9vr) which is set to private). That's why this PR resolves the audit by overring the version.

Security advisory: https://github.com/advisories/GHSA-pxg6-pf52-xh8x

# Notes

I pinned Alchemy and WalletConnect because we probably not want just to quickly update those but, test any update and if I don't pin those, npm i or npm audit updates those.

# Changes

- Pin `cookie`
- Run `npm audit --fix`
- Ping `walletconnect` and `alchemy` version


AntonioVentilii (approved):
LGTM tks!

PR #3771, opened 2024-11-27
ci(backend): auto-generate bindings

AntonioVentilii:
# Motivation

We would like to auto-generate the bindings if necessary.

# Note

I had to remove the checkout from the single actions, because they would go in conflict with the possibility to create PRs (different head_ref). This is not an issue, since all actions are called after a previous checkout in their usage in the code.

# Tests

We injected bad bindings in this PR and they were auto-fixed.


bitdivine (approved):
Thanks!

PR #3775, opened 2024-11-27
refactor(ci): separate release notes creation

AntonioVentilii:
# Motivation

Since it already happened to create version tags outside the CI workflows (and I suppose it will be not the last time), it is useful to separate the actions of tagging a release and the action of creating release notes.


.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,26 @@
...
+      uses: actions/checkout@v4
+
+    - name: Publish Release Notes
+      id: publish_release
+      uses: release-drafter/release-drafter@v6
+      with:
+        version: ${{ github.ref_name }}
+        tag: ${{ github.ref_name }}
+        name: ${{ github.ref_name }}
+        latest: true
```
DenysKarmazynDFINITY, line 25: Looks like `latest` is `true` by default - https://github.com/release-drafter/release-drafter?tab=readme-ov-file#action-inputs

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3776, opened 2024-11-27
fix(ci): correct version tag selector

AntonioVentilii:
# Motivation

According to [GItHub documentation](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore), the correct syntax to select a version tag is `v*`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3777, opened 2024-11-28
feat(backend): function to add hidden dapps

AntonioVentilii:
# Motivation

We want to add an endpoint to add dApps that the user wants hidden in the dApps Carousel.

# Changes

- Add function `add_hidden_dapp_id` to `user_profile` implementation.
- Create error types and returns types
- Propagate the function up to open endpoint in `lib`

# Tests

Included tests for new endpoint.


src/shared/src/impls.rs
```diff
@@ -164,6 +164,35 @@ impl StoredUserProfile {
         new_profile.updated_timestamp = now;
         Ok(new_profile)
     }
+
+    /// # Errors
+    ///
+    /// Will return Err if there is a version mismatch or if the dApp ID is already hidden.
```
bitdivine, line 170: Personally I wouldn't return an error if the app is already hidden.  It is often helpful if functions are idenpontent, so saying "hide x!  hide x!  hide x!" should get the answer "ok, ok, ok, calm down!".  But it's a matter of taste.  Up to you.

src/backend/src/lib.rs
```diff
@@ -518,6 +520,29 @@ pub fn add_user_credential(
...
 }
 
+/// Adds a dApp ID to the user's list of hidden dApp IDs.
+///
+/// # Errors
+/// - Returns `Err` if the ID is already in the `hidden_dapp_ids`, or the user profile is not found, or the user profile version is not up-to-date.
+#[update(guard = "may_write_user_data")]
+pub fn add_user_hidden_dapp_id(
+    request: AddHiddenDappIdRequest,
+) -> Result<(), AddDappSettingsError> {
```
bitdivine, line 530: Can I provide a 1Mb long app ID?  Sounds as if I might be able to do a DOS fairly cheaply.  See the comment at the end of the PR for a suggested sanity check to call.

bitdivine (approved):

src/shared/src/impls.rs
AntonioVentilii: eheh agreed! changed

src/backend/src/user_profile.rs
```diff
@@ -51,3 +52,17 @@ pub fn add_credential(
         Err(AddUserCredentialError::UserNotFound)
     }
 }
+
```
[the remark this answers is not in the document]
AntonioVentilii: right! included

src/backend/src/lib.rs
AntonioVentilii: added a test for the max length of the ID

shall I add something for the max number of IDs that can be saved?

bitdivine: Sounds like a good idea. :-) But doesn't have to be in this PR.

src/backend/src/user_profile.rs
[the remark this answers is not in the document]
bitdivine: Thanks!

src/shared/src/impls.rs
bitdivine: Thanks!

bitdivine (approved):
LGTM, thanks!

src/backend/src/lib.rs
AntonioVentilii: Done with:

- https://github.com/dfinity/oisy-wallet/pull/5248

PR #3779, opened 2024-11-28
feat(frontend): rename "swap" to "convert"

DenysKarmazynDFINITY:
# Motivation

The renaming is requested by the design and product team:

```
It turns out that the word ‘Swap’ might be misleading in our context of ChainFusion conversion. After discussing with 
Stefan, we agreed to stick with the ‘Convert’ naming until we find a better unique one (it's a marketing question tbh).
Can we please update the wording in the modal title and in the ‘Swap fee’ line?”
```


AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/i18n/en.json
```diff
@@ -537,7 +537,7 @@
 			"estimated_inter_network": "Estimated Inter-network Fee",
 			"estimated_eth": "Estimated Ethereum Fee",
 			"max_fee_eth": "Max fee <small>(likely in &lt; 30 seconds)</small>:",
-			"convert_fee": "Swap fee",
+			"convert_fee": "Convert fee",
```
AntonioVentilii, line 540: Maybe "Conversion fee"? Or does it sound better "Convert fee"?

DenysKarmazynDFINITY: "Conversion fee" is better, thanks for the tip!

PR #3780, opened 2024-11-28
style(frontend): improve mobile view for dapp explorer

loki344:
# Motivation

The design team improved the mobile view for the dapp explorer. The filter bar on mobile is horizontal and scrollable. Instead of just one dapp per column, we show two dapps.

# Changes

- add no scrollbar utility class to hide scrollbar overflow
- change mobile layout of dapp explorer

# Tests

before
desktop for reference:

image

mobile:

image

after
desktop for refrence:

image

mobile > 360px
image

mobile < 360px

image


src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
```diff
@@ -74,3 +74,11 @@
...
 
 <SubmitDappButton />
+
+<style lang="postcss">
+	@media (max-width: 360px) {
+		ul {
+			@apply grid-cols-1;
+		}
+	}
+</style>
```
AntonioVentilii, line 84: isn't the minimum size `375px` (for example iPhone SE)? what is the use case for `360px` ? if there is, can we move it to tailwind? maybe defining a new screen

```
screens: {
  xs: '360px',
  ...
 }
```

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
```diff
@@ -32,15 +32,15 @@
...
+	<div class="mb-6 md:mb-10">
 		<DappPromoBanner
 			on:click={() => modalStore.openDappDetails(featuredDapp)}
 			dAppDescription={featuredDapp}
 		/>
 	</div>
 {/if}
 
-<div class="flex flex-wrap gap-4">
+<div class="no-scrollbar flex gap-4 overflow-x-auto p-1 md:flex-wrap md:p-0">
```
AntonioVentilii, line 43: with `overflow-auto` (including y), does the scrollbar appear? we don't want the same behavior as before?

loki344: We explicitly do not want a scrollbar here on mobile => that's the reason for the no-scrollbar class. That's why for larger screen flex-wrap will wrap items below.

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
loki344: As far as I know, for mobile designs it would be best practice to use 320px as smallest possible size where your designs should make sense. Just to give one reference: https://www.browserstack.com/guide/responsive-design-breakpoints 
image

I added it for this special case because of a request from the design team:
image

I'm unsure if I should add it to the tw definitions because the rest of the application is fine with the current breakpoints. Seems like an exception to me for now.

AntonioVentilii: wow 320px! very small! thank you for the explanation!

I am just concerned, because the rest of the layout/interface doesn't respect this size, so I am not sure, if we should or not. Meaning, it seems strange to have this kind of check only in a single place in the entire OISY, you know what I mean?

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
AntonioVentilii: just to clarify, is this something purely for this component or in general? for the entire OISY? my question is the same as above, shouldn't we treat this issue at top-level?

loki344: Shouldn't we leave it up to the implementation to show a scrollbar if needed? At least until there is a clear pattern. As of now we only have this one horizontal scrollbar - I think it's too early to optimise and generalise. That's why I introduced the no-scrollbar class.

src/frontend/src/lib/components/dapps/DappsExplorerSignedIn.svelte
AntonioVentilii: Yes, yes, agreed! But that is exactly my point: if the rest is not focused on 320px-screens, why should this one be? If there is a reason for this one to be treated for smaller screens, for sure there is the same reason for all the other screens/layouts, no? Am I missing something?

loki344: Hm I try to outline the problem a bit. We normally style the application the same for 0-640px (sm breakpoint from tailwind). In this particular case, the requirement is to have a more fine-grained control over the grid. While two columns are fine for small devices > 360px, even smaller screens will break the cards content. Problem: 
image

As this is a very specific case, I do not want to add this to a general configuration.

DenysKarmazynDFINITY (approved):
I tested the branch locally and the explorer page does look good now for screens < 360px. I suggest we move one with it and adjust the theme file whenever a next usage for that `xs` breakpoint occurs.

PR #3783, opened 2024-11-29
refactor(frontend): extract mock user profile

AntonioVentilii:
# Motivation

For reusability, we extract the user profile mock and we use it in the relevant tests.


peterpeterparker (approved):
LGTM, thx

PR #3794, opened 2024-12-02
test(frontend): add tests for btc-wallet worker

DenysKarmazynDFINITY:
# Motivation

As I'm going to introduce "queryAndUpdate" approach in the btc-wallet worker in the next PR, we need to add some tests for it beforehand. The test structure is similar to what we have in the other wallet test file.


peterpeterparker (approved):
Nice, thx!

PR #3795, opened 2024-12-02
style(frontend): improves logout behaviour

BonomoAlessandro:
# Motivation

When i navigate on the activity page, logout and then login again, i should land on the home page instead of the activity page. 
When i use a link that brings me to the activity page and i then login i should land on the activity page.
When i get logged out automatically and the login again i should land on the page i was when i logged out.

# Changes

- changes url to '/' before logout

# Tests

**before:**

https://github.com/user-attachments/assets/c1116cde-7dc0-4875-882b-4e2c4a7ed234

**after:**

https://github.com/user-attachments/assets/b027b370-3310-481a-b797-5a2d7e48d370

### DIFF
--- a/src/frontend/src/lib/components/core/SignOut.svelte
+++ b/src/frontend/src/lib/components/core/SignOut.svelte
@@ -4,10 +4,12 @@
 	import { LOGOUT_BUTTON } from '$lib/constants/test-ids.constants';
 	import { signOut } from '$lib/services/auth.services';
 	import { i18n } from '$lib/stores/i18n.store';
+	import { replaceHistory } from '$lib/utils/route.utils';
 
 	const dispatch = createEventDispatcher();
 
 	const logout = async () => {
+		replaceHistory('/')
 		dispatch('icLogoutTriggered');
 		await signOut();
 	};


src/frontend/src/lib/components/core/SignOut.svelte
```diff
@@ -4,10 +4,12 @@
 	import { LOGOUT_BUTTON } from '$lib/constants/test-ids.constants';
 	import { signOut } from '$lib/services/auth.services';
 	import { i18n } from '$lib/stores/i18n.store';
+	import { replaceHistory } from '$lib/utils/route.utils';
 
 	const dispatch = createEventDispatcher();
 
 	const logout = async () => {
+		replaceHistory(new URL(window.location.origin));
```
peterpeterparker, line 12: 1. Should we actually replace the history?
2. Have you tried using `goto`?
3. Can you rename the PR, it's not a `style` change but, rather a `feature`.
4. The implementation should probably belong in the `auth.service` module rather than the component, as there might be other consumers that could use the function without relying on the component. WDYT?
5. Can you provide a test, either an E2E or a unit test?

BonomoAlessandro: @peterpeterparker 
1/2. You are right i guess, we should not replace the history. The `goto` might be the better solution.
3. done
4. I don't see exactly why we should move this into the `auth.service` module. It is only a simple `goto` command that is only used if the user clicks on the `logout` button.
5. Yes, i will provide a test

peterpeterparker: I believe this is the right approach, aligning with the method we've been applying over the past few weeks/months:

1. We should provide new tests that cover the service (in that case `auth.module`) to resolve the tech debt and facilitate the addition of the new feature.
2. Then, we should extend the `logout` feature with the "result url option to navigate to the root" and adapt the tests accordingly.

BonomoAlessandro: @peterpeterparker @AntonioVentilii I added the optional `resetUrl` parameter to the `SignOut` function and implemented some test cases.

src/frontend/src/lib/services/auth.services.ts
```diff
@@ -58,7 +59,7 @@ export const signIn = async (
 	}
 };
 
-export const signOut = (): Promise<void> => logout({});
+export const signOut = (resetUrl = false): Promise<void> => logout({ resetUrl });
```
peterpeterparker, line 62: Despite having a single parameter, in this case I would still use an object for paramater, because this makes the code more readable in my opinion. WDYT?

```
await signOut(true)
```

is less expressive than

```
await signOut({resetUrl: true})
```

src/frontend/src/tests/lib/services/auth.services.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+	Object.defineProperty(window, 'location', {
+		writable: true,
+		value: {
+			href: url,
+			reload: vi.fn()
+		}
+	});
+};
+
+describe('auth.services', () => {
```
peterpeterparker, line 18: Looks good. Once this PR is merged, would you be interested adding more tests for the other feature of the service to reduce the test debt?

peterpeterparker (commented):
Looking good. Just a minor comment.

src/frontend/src/lib/services/auth.services.ts
BonomoAlessandro: @peterpeterparker I think this is a good idea. `await signOut({resetUrl: true})` is more readable than `await signOut(true)`.

src/frontend/src/tests/lib/services/auth.services.spec.ts
peterpeterparker: Super, thx!

peterpeterparker (approved):
LGTM, thx

PR #3802, opened 2024-12-02
style(frontend): new modal styling

loki344:
# Motivation

We want a consistent layout for the modals: 
Header and footer are always visible. The content becomes scrollable if needed. The spacing of modals needs to be minimal for mobiles.

# Changes

- Adjust the styling to match the figma designs
- Remove existing styling where necessary
- Style the Loader differently, so that it still shows the application in the background.

# Tests

Apart from the extensive testing from the design team, here is a video showcasing the new modals.

https://github.com/user-attachments/assets/b93230fc-d3b5-4a75-9a51-8741eecb22e7

https://github.com/user-attachments/assets/24634802-4d3c-47ce-a23c-62efecd2a201


AntonioVentilii (approved):
LGTM tks

PR #3803, opened 2024-12-02
feat(frontend): IC transaction scroll for Activity

AntonioVentilii:
# Motivation

In the Activity page we are not loading the next transactions for IC tokens. So, we create a component that will load them for all the necessary tokens. It is basically an expansion of `IcTransactionScroll` component.

# Changes

- Create component AllTransactionsScroll (born from IcTransactionScroll).
- Loop all tokens and load the next transactions of IC tokens. In another PR, we will include Solana.
- Map all the tokens if they are either disabled or not.

# Tests

Apart from creating new tests, this is a replica in test canister:


AntonioVentilii:
closing and re-opening to stay among the recent ones

AntonioVentilii:
The PR's three goals (extract `AllTransactionsScroll`, paginate IC tokens on Activity, handle the enabled/disabled token mapping) have all been delivered, and the Solana extension that PR #3803 explicitly deferred is also already in place. The branch is also 23 commits old with many merges from `main`, so reviving it would be more work than reproposing any missing piece in a focused PR.

PR #3804, opened 2024-12-02
feat(frontend): query and update BTC wallet

DenysKarmazynDFINITY:
# Motivation

In this PR, the "query and update" strategy is implemented in the BTC wallet worker. Important notes:
* "Querying" balance only available for BTC mainnet
* At the moment, we don't have a way of certifying BTC transactions, therefore they are fetched only on `query`

P.S: the UI part of the feature can be found here - https://github.com/dfinity/oisy-wallet/pull/3806

### DIFF
--- a/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
+++ b/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
@@ -170,7 +170,7 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> 
 					shouldFetchTransactions: data?.shouldFetchTransactions,
 					minterCanisterId: data?.minterCanisterId
 				}),
-			onLoad: ({ certified: _, ...rest }) => this.syncWalletData({ ...rest }),
+			onLoad: ({ certified: _, ...rest }) => this.syncWalletData(rest),
 			identity,
 			resolution: 'all_settled'
 		});
--- a/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
+++ b/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
@@ -26,12 +26,12 @@ interface LoadBtcWalletParams extends QueryAndUpdateRequestParams {
 	minterCanisterId?: OptionCanisterIdText;
 }
 interface BtcWalletStore {
-	balance: CertifiedData<bigint> | undefined;
+	balance: CertifiedData<bigint | null> | undefined;
 	transactions: Record<string, CertifiedData<BitcoinTransaction[]>>;
 }
 
 interface BtcWalletData {
-	balance: CertifiedData<bigint> | null;
+	balance: CertifiedData<bigint | null>;
 	uncertifiedTransactions: CertifiedData<BtcTransactionUi>[];
 }
 
@@ -94,26 +94,25 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> 
 		btcAddress,
 		minterCanisterId,
 		certified = true
-	}: Omit<
-		LoadBtcWalletParams,
-		'shouldFetchTransactions'
-	>): Promise<CertifiedData<bigint> | null> => {
-		// Query BTC balance only if minterCanisterId and BITCOIN_CANISTER_IDS[minterCanisterId] are available
+	}: Omit<LoadBtcWalletParams, 'shouldFetchTransactions'>): Promise<
+		CertifiedData<bigint | null>
+	> => {
 		if (!certified) {
-			if (nonNullish(minterCanisterId) && BITCOIN_CANISTER_IDS[minterCanisterId]) {
-				return {
-					data: await getBalanceQuery({
-						identity,
-						network: bitcoinNetwork,
-						address: btcAddress,
-						bitcoinCanisterId: BITCOIN_CANISTER_IDS[minterCanisterId]
-					}),
-					certified: false
-				};
-			}
-
-			// corner case in which minter and/or BTC canister ids are not set
-			return null;
+			// Query BTC balance only if minterCanisterId and BITCOIN_CANISTER_IDS[minterCanisterId] are available
+			// These values will be there only for "mainnet", for other networks - balance on "query" will be null
+			return {
+				data:
+					nonNullish(minterCanisterId) && BITCOIN_CANISTER_IDS[minterCanisterId]
+						? await getBalanceQuery({
+								identity,
+								network: bitcoinNetwork,
+								address: btcAddress,
+								bitcoinCanisterId: BITCOIN_CANISTER_IDS[minterCanisterId],
+								minConfirmations: BTC_BALANCE_MIN_CONFIRMATIONS
+							})
+						: null,
+				certified: false
+			};
 		}
 
 		return {
@@ -183,10 +182,9 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> 
 		response: BtcWalletData;
 	}) => {
 		const newBalance =
-			nonNullish(balance) &&
-			(isNullish(this.store.balance) ||
-				this.store.balance.data !== balance.data ||
-				(!this.store.balance.certified && balance.certified));
+			isNullish(this.store.balance) ||
+			this.store.balance.data !== balance.data ||
+			(!this.store.balance.certified && balance.certified);
 		const newTransactions = uncertifiedTransactions.length > 0;
 
 		this.store = {
@@ -206,7 +204,7 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> 
 			})
 		};
 
-		if (isNullish(balance) || (!newBalance && !newTransactions)) {
+		if (!newBalance && !newTransactions) {
 			return;
 		}
 
--- a/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
+++ b/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
@@ -4,7 +4,7 @@ import type { BtcPostMessageDataResponseWallet } from '$btc/types/btc-post-messa
 import { mapBtcTransaction } from '$btc/utils/btc-transactions.utils';
 import { BITCOIN_CANISTER_IDS } from '$env/networks.btc.env';
 import { getBalanceQuery } from '$icp/api/bitcoin.api';
-import { queryAndUpdate } from '$lib/actors/query.ic';
+import { queryAndUpdate, type QueryAndUpdateRequestParams } from '$lib/actors/query.ic';
 import { getBtcBalance } from '$lib/api/signer.api';
 import { WALLET_TIMER_INTERVAL_MILLIS } from '$lib/constants/app.constants';
 import { btcAddressData } from '$lib/rest/blockchain.rest';
@@ -13,17 +13,14 @@ import { SchedulerTimer, type Scheduler, type SchedulerJobData } from '$lib/sche
 import type { BtcAddress } from '$lib/types/address';
 import type { BitcoinTransaction } from '$lib/types/blockchain';
 import type { OptionCanisterIdText } from '$lib/types/canister';
-import type { OptionIdentity } from '$lib/types/identity';
 import type { PostMessageDataRequestBtc } from '$lib/types/post-message';
 import type { CertifiedData } from '$lib/types/store';
 import { mapToSignerBitcoinNetwork } from '$lib/utils/network.utils';
 import { type BitcoinNetwork } from '@dfinity/ckbtc';
 import { assertNonNullish, isNullish, jsonReplacer, nonNullish } from '@dfinity/utils';
 
-interface LoadBtcWalletParams {
-	identity: OptionIdentity;
+interface LoadBtcWalletParams extends QueryAndUpdateRequestParams {
 	bitcoinNetwork: BitcoinNetwork;
-	certified: boolean;
 	btcAddress: BtcAddress;
 	shouldFetchTransactions?: boolean;
 	minterCanisterId?: OptionCanisterIdText;
@@ -34,8 +31,8 @@ interface BtcWalletStore {
 }
 
 interface BtcWalletData {
-	balance: CertifiedData<bigint> | undefined;
-	transactions: CertifiedData<BtcTransactionUi>[];
+	balance: CertifiedData<bigint> | null;
+	uncertifiedTransactions: CertifiedData<BtcTransactionUi>[];
 }
 
 export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> {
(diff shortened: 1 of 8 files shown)


peterpeterparker (commented):
Can you strip out the UI part (component) in a separate PR? 

I believe we should be able to provide those in two distinct PRs and it would better scope the goal of the feature - the web worker implementing the query+update, the UI consuming the improvement.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -46,100 +65,159 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
-			minConfirmations: BTC_BALANCE_MIN_CONFIRMATIONS
-		});
-		const certifiedBalance = {
-			data: balance,
+		bitcoinNetwork,
+		btcAddress,
+		minterCanisterId,
+		certified = true
+	}: Omit<LoadBtcWalletParams, 'shouldFetchTransactions'>): Promise<
+		CertifiedData<bigint> | undefined
```
peterpeterparker, line 103: Is it correct to return `undefined`? In what case is the balance undefined? And shouldn't the return be `{balance: undefined, certified: false}` or is there a case where the balance is undefined and certified? And should we use `null` to express the fact that it's a choice?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -46,100 +65,159 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
+	> => {
+		// Query BTC balance only if minterCanisterId is available
+		if (!certified) {
+			if (nonNullish(minterCanisterId) && isNullish(this.store.balance)) {
```
peterpeterparker, line 107: Regarding `isNullish(this.store.balance)` - the scheduler is not meant to support polling as well? We never update the value with a query?

peterpeterparker, line 107: Regarding `nonNullish(minterCanisterId)`, wouldn't it be more accurate to test if `nonNullish(minterCanisterId) && nonNullish(BITCOIN_CANISTER_IDS[minterCanisterId])` or we are 100% sure the minter canister id is always in the `BITCOIN_CANISTER_IDS`?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -46,100 +65,159 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
+			bitcoinNetwork,
+			certified,
+			btcAddress,
+			minterCanisterId
+		});
 
-		return certifiedBalance;
-	}
+		// TODO: investigate and implement "update" call for BTC transactions
+		const transactions =
```
peterpeterparker, line 151: nitpick: on main we used the name `uncertifiedTransactions`. Maybe it would be nice to use it again for readability reasons

```
// TODO: investigate and implement "update" call for BTC transactions
const. uncertifiedTransactions = ...
return{ balance, transactions: uncertifiedTransactions };

File: src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts [hunk 4]
```diff
@@ -46,100 +65,159 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
 
 		const btcAddress = data?.btcAddress.data;
 		assertNonNullish(btcAddress, 'No BTC address provided to get BTC transactions.');
 
-		const balance = await this.loadBtcBalance({ identity, bitcoinNetwork });
-		const { newTransactions, latestBitcoinBlockHeight } =
-			nonNullish(data) && data.shouldFetchTransactions
-				? await this.loadBtcTransactionsData({ btcAddress })
-				: { newTransactions: [], latestBitcoinBlockHeight: undefined };
+		// TODO: implement "onCertifiedError" to handle errors in update calls
```
peterpeterparker, line 166: You implement this in next PR right? i.e. feature should not go live without handling error as we reset the UI stores that way.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -46,100 +65,159 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
-			? newTransactions.map((transaction) => ({
-					data: mapBtcTransaction({ transaction, btcAddress, latestBitcoinBlockHeight }),
-					certified: false
-				}))
-			: [];
+	private syncWalletData = ({
+		response: { balance, transactions }
+	}: {
+		response: BtcWalletData;
+		certified: boolean;
```
peterpeterparker, line 187: Should we remove/exclude the `certified` parameter? I spent quite some time reviewing this so it would maybe be useful to make it clear we are considering another certified information/source.

peterpeterparker (commented):
I solely reviewed the code of `btc-wallet.scheduler.ts` which already took quite some time. Thanks in advance for splitting the PRs.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -46,100 +65,159 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
-						}),
-						{}
-					)
-				}
-			};
-
 			const latestBitcoinBlockHeight = await btcLatestBlockHeight();
 
-			return { newTransactions, latestBitcoinBlockHeight };
+			return nonNullish(latestBitcoinBlockHeight)
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: Actually, it's a good point - the `isNullish(latestBitcoinBlockHeight)` is redundant, the value is expected to always be a number. I think that check is left over from the time when we used another API for getting current block height and where it could have been nullish. Updated now.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: I checked the other wallets, and indeed we support re-querying too. I somehow had an impression that we poll only with "update". For consistency, made BTC wallet work the same way.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: Since `isNullish(this.store.balance)` is now gone, balance can be nullish only in a very specific case - when minter and/or BTC canister ids are not available. I left a comment in the code with a hint on this corner case. Also, updated the value to `null`.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: Yep, agree, updated.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: Yes, gonna be the next step.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: Removed.

DenysKarmazynDFINITY:
@peterpeterparker I moved the UI component part to a separate PR, and replied to your comments. Could you please take another look?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -46,95 +62,153 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
-		newTransactions: BitcoinTransaction[];
-		latestBitcoinBlockHeight: number | undefined;
-	}> {
+	private async loadBtcTransactionsData({
+		btcAddress
+	}: {
+		btcAddress: BtcAddress;
+	}): Promise<CertifiedData<BtcTransactionUi>[]> {
 		try {
 			const { txs: fetchedTransactions } = await btcAddressData({ btcAddress });
```
peterpeterparker, line 71: Not for this PR, just something I noticed: if you want you can "group" the calls here.

```
const [{ txs: fetchedTransactions }, latestBitcoinBlockHeight] = await Promise.all([
  btcAddressData({ btcAddress }),
  btcLatestBlockHeight()
```

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -46,95 +62,153 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
+						identity,
+						network: bitcoinNetwork,
+						address: btcAddress,
+						bitcoinCanisterId: BITCOIN_CANISTER_IDS[minterCanisterId]
+					}),
+					certified: false
+				};
+			}
+
+			// corner case in which minter and/or BTC canister ids are not set
```
peterpeterparker, line 115: Is that a corner case or should it actually throw an error? i.e. if those are not provided, should the scheduler retries the call later on?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -46,95 +62,153 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
+			return null;
+		}
+
+		return {
+			data: await getBtcBalance({
+				identity,
+				network: mapToSignerBitcoinNetwork({
+					network: bitcoinNetwork
+				}),
+				minConfirmations: BTC_BALANCE_MIN_CONFIRMATIONS
```
peterpeterparker, line 125: We do not pass `minConfirmations` to the query?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -46,95 +62,153 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
+		'shouldFetchTransactions'
+	>): Promise<CertifiedData<bigint> | null> => {
+		// Query BTC balance only if minterCanisterId and BITCOIN_CANISTER_IDS[minterCanisterId] are available
+		if (!certified) {
+			if (nonNullish(minterCanisterId) && BITCOIN_CANISTER_IDS[minterCanisterId]) {
+				return {
+					data: await getBalanceQuery({
+						identity,
+						network: bitcoinNetwork,
+						address: btcAddress,
```
peterpeterparker, line 108: Are we 100% sur this `btcAddress` is the same address as the one that will be use for the update call? Should we have a guard or port-checks if we might end with a balance for an address A followed by the balance for an address B?

peterpeterparker (commented):
Thanks for the changes, looks good!

Few new questions because I went through the code again since it changed a bit. Hope you don't mind.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: Added it - didn't notice that they accept it here too.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: I re-worked the code to always return CertifiedData (`bigint | null` in case of balance) as I realized it's not a corner case - we can only query balance for mainnet, therefore it would always end up there for testnet and regtest. Also, added a comment with a small explanation. WDYT?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: Before launching the worker, we wait for `btcAddress` to be loaded and I don't think it can be updated at a later point. In case certifying IDB address fails, the session will be ended. So, unless I'm missing something, it seems that such a case cannot happen.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: 👍

DenysKarmazynDFINITY:
@peterpeterparker thanks a lot for the comments, they did help to find a problem - there was an issue with skipping postMessage if balance is null. With new iteration, I had to touch some additional files, so the PR got a bit bigger, sorry for that. 

Ready for another look.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -46,95 +62,151 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
+			request: ({ identity: _, certified }) =>
+				this.loadWalletData({
+					certified,
+					identity,
+					btcAddress,
+					bitcoinNetwork,
+					shouldFetchTransactions: data?.shouldFetchTransactions,
+					minterCanisterId: data?.minterCanisterId
+				}),
+			onLoad: ({ certified: _, ...rest }) => this.syncWalletData({ ...rest }),
```
peterpeterparker, line 173: nitpick: you can just do `this.syncWalletData(rest);`

src/frontend/src/btc/services/btc-listener.services.ts
```diff
@@ -19,13 +19,15 @@ export const syncWallet = ({
...
 	} = data;
 
-	balancesStore.set({
-		tokenId,
-		data: {
-			data: BigNumber.from(balance),
-			certified
-		}
-	});
+	if (nonNullish(balance)) {
```
peterpeterparker, line 22: If the balance is nullish, shouldn't we reset the store otherwise we keep in memory the previous balance?

peterpeterparker (commented):
Thanks. Just one comment regarding what I assume you changed in the last PRs. After that sounds good to me.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DenysKarmazynDFINITY: Done.

src/frontend/src/btc/services/btc-listener.services.ts
DenysKarmazynDFINITY: Good point, updated.

peterpeterparker (approved):
LGTM, thx

Note: It was quite a big PR 😅, so I assume there will be some testing and prodsec will be informed about the query + as commented, it should not go live until certified errors are handled.

DenysKarmazynDFINITY:
@peterpeterparker absolutely, both points are incoming - I will prepare error handling today and Stefan is going to arrange a security review for this code. In the meantime, let's continue with the UI part https://github.com/dfinity/oisy-wallet/pull/3806 🙂

PR #3806, opened 2024-12-03
feat(frontend): query and update BTC wallet - UI

DenysKarmazynDFINITY:
# Motivation

The second part of the "query and update BTC wallet" feature in which UI wallet loader component is updated. The new worker code can be found here - https://github.com/dfinity/oisy-wallet/pull/3804.


peterpeterparker (approved):
LGTM, thx

PS: I was wondering if we could setup some test for this but, maybe something easier to test with E2E

PR #3809, opened 2024-12-03
feat(ci): Start solana before running e2e tests

bitdivine:
# Motivation
We need a local solana testnet in order to test solana integration

# Changes
- Add a dev-tools file to store tool versions.
- Add a tool to install the version of solana specified in the dev-tools file.
- Add a tool to start solana in the background
- Start solana before running the e2e tests

# Tests
After starting solana, there is a health check that will fail if solana failed to start.


scripts/setup
bitdivine, line 1: Standard setup command, taken from the chain fusion signer.  I have kept all the installation methods even though only `sh` is needed for this PR, as I believe that the installations can be simplified and standardized with this.

AntonioVentilii (approved):
LGTM tks!

PR #3810, opened 2024-12-03
style(frontend): use btn close modal in transaction modal

loki344:
# Motivation

Use existing button to close modal in TransactionModal instead of custom implementation. Goal is to achieve a uniform layout for modals. 

see as an example from figma (not all modals are showcased there:

image

# Changes

Change Button to ButtonCloseModal

# Tests

before:
image

after:
image


AntonioVentilii (approved):
LGTM, tks!

BTW, I noticed that `SendTokensList` component has the same button, will you do it in another PR?

loki344:
> LGTM, tks!
> 
> BTW, I noticed that `SendTokensList` component has the same button, will you do it in another PR?

Thanks, yes - will follow

PR #3811, opened 2024-12-03
feat(frontend): filter carousel dApps using user profile

AntonioVentilii:
# Motivation

With PR #3770 , we introduces a list of hidden dApps in the carousel, saved in the user profile.

Now we filter the daps accordingly.

# Changes

- Extract dapps filter function.
- Add filter by hidden dApps from the User Profile. 

# Tests

Tests provided.


src/frontend/src/lib/components/dapps/DappsCarousel.svelte
```diff
@@ -2,16 +2,29 @@
...
+		dAppDescriptions,
+		type OisyDappDescription
 	} from '$lib/types/dapp-description';
 	import { filterCarouselDapps } from '$lib/utils/dapps.utils';
 
 	export let styleClass: string | undefined = undefined;
 
+	$: {
+		if ($authSignedIn) {
+			loadUserProfile({ identity: $authIdentity });
```
peterpeterparker, line 19: - Is this expected, you want to re-fetch and re-load the profiles every time the dapps carousel component is mounted? Shouldn't the user profile be loaded once per session and when required only?

- Should the carousel be displayed only when the user profile is ready otherwise doesn't it feels like a glitch if slides are displayed but suddenly disappear?

AntonioVentilii: - I am undecided on this one: theoretically it is required ONLY when we have the carousel (or in the settings page). So I didn't want to load it at top level.
Settings page re-loads it every time the page is mounted, so i thought it would make sense to re-load it if the carousel is re-mounted (for example navigating back and forth). Otherwise I can make it load just once for the Assets page, and eventually trigger a reload if and when necessary?

- Yes, makes sense, will change! I will not care about certified or not, though, WDYT?

peterpeterparker: Given that we are talking about a "user profile," I would assume that loading it globally—for example optimized for performance by saving the data in IndexedDB and using a deferred update call for security—would be a natural fit. However, if the profile isn't needed immediately, we could load it for a specific feature. That said, we should ensure we avoid unnecessary calls. In this case, will we end up reloading the profile unnecessarily?

peterpeterparker: >  I will not care about certified or not,

IMO it's a call to the IC so unless we are told otherwise, I would implement query+update.

Plus, the calls load the entire user profile.

The `queryAndUpdate` function as also an option, I think, to just make a query. So already implementing this makes it also easy to be tweaked in the future.

AntonioVentilii: > In this case, will we end up reloading the profile unnecessarily?

For settings, it is reloaded every time the user accesses `Settings` page because we need  the latest user profile data (makes sense). For the carousel, we need something similar.

We could do one of these:

1. reload it every time the user loads the carousel (meaning always when the user navigates to the `Assets` page), because we want to check if there is any change relevant to the carousel

2. make it more passive: we re-load only when we know that OISY changed something in it (or when user profile is nullish)

3. do it like the Ethereum balances: a services that uploads every X seconds for example. But I don't think it is really necessary for now.

WDYT? I maybe like more the second option

peterpeterparker: Honestly, in my opinion, the correct approach would be to load and handle the user profile as a global feature because, it's a user profile hence tied with the authentication. Any other solution feels like a temporary workaround. So, to keep it simple, maybe something like 'if the profile is already loaded, skip it; otherwise, load it' would be enough?

AntonioVentilii: yes, make sense, will change to load it if not defined

AntonioVentilii: done

peterpeterparker: Thanks. Maybe a bit a picky feedback, so apologies in advance if it feels like it:

1. I feel like it would be cleaner to have the logic encapsulated within the service. Either we create a new function, like `loadUserProfileIfNeeded`, or we add a `reload` paramter to `loadUserProfile`. That way the component can be naive and we can also test the behavior at the service layer.

2. Maybe use `onMount(load)` because, if I get it right, we only want to load the profile when the component is mounted and not if any stores is reevaluated.

WDYT?

AntonioVentilii: Ah!!!

Is it something like this?

- PR #3842

AntonioVentilii:
@peterpeterparker ready for another review, but have a look, I changed a bit, according to recent PRs

peterpeterparker (approved):
LGTM, thx

PR #3816, opened 2024-12-03
feat(frontend): hide carousel when there is no dApp

AntonioVentilii:
# Motivation

We would like to hide the carousel when the list of dApps filtered is empty.


DenysKarmazynDFINITY (approved):
Nice, thanks!

PR #3817, opened 2024-12-03
fix(frontend): do not load ETH transactions without API key

AntonioVentilii:
# Motivation

It does not make sense to run the `loadTransactions` function to load Ethereum transactions if there is no Etherscan API key.


AntonioVentilii:
Superseded by:

- PR #3822

PR #3818, opened 2024-12-03
feat(frontend): handle onCertifiedError in BTC worker

DenysKarmazynDFINITY:
# Motivation

We need to handle onCertifiedError in BTC wallet worker: for mainnet, there will be a toast mesage displayed with the error message, for regtest and testnet - just a console message.

### DIFF
--- a/src/frontend/src/btc/services/btc-listener.services.ts
+++ b/src/frontend/src/btc/services/btc-listener.services.ts
@@ -43,11 +43,11 @@ export const syncWallet = ({
 export const syncWalletError = ({
 	tokenId,
 	error: err,
-	silent
+	hideToast = false
 }: {
 	tokenId: TokenId;
 	error: unknown;
-	silent?: boolean;
+	hideToast?: boolean;
 }) => {
 	const errorText = get(i18n).init.error.btc_wallet_error;
 
@@ -56,7 +56,7 @@ export const syncWalletError = ({
 	// Currently, the certified error can only happen while fetching balance, but we still want to reset transactions to avoid displaying incorrect data
 	btcTransactionsStore.reset(tokenId);
 
-	if (silent) {
+	if (!hideToast) {
 		console.warn(`${errorText}:`, err);
 		return;
 	}
--- a/src/frontend/src/btc/services/worker.btc-wallet.services.ts
+++ b/src/frontend/src/btc/services/worker.btc-wallet.services.ts
@@ -1,5 +1,6 @@
 import { syncWallet, syncWalletError } from '$btc/services/btc-listener.services';
 import type { BtcPostMessageDataResponseWallet } from '$btc/types/btc-post-message';
+import { LOCAL } from '$lib/constants/app.constants';
 import {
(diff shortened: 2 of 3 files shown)


src/frontend/src/btc/services/btc-listener.services.ts
```diff
@@ -36,3 +39,30 @@ export const syncWallet = ({
...
 };
+
+export const syncWalletError = ({
+	tokenId,
+	error: err,
+	silent
+}: {
+	tokenId: TokenId;
+	error: unknown;
+	silent?: boolean;
```
peterpeterparker, line 50: I would maybe rename `silent` here or rename it? This option was recently added by Antonio for the ICRC token with the Index canister, which fails to load but still returns a balance when fetched with the ledger. The use case is somewhat exceptional, so I spontaneously would thought of not extending the concept of silencing the error further, if that makes sense?

peterpeterparker (commented):
Can you also provide a test for `btc-listener.services`? We can do that in another PR, as you rather like.

src/frontend/src/btc/services/btc-listener.services.ts
DenysKarmazynDFINITY: I had 2 thoughts when I was adding this param:

1. Local env: since in 99% of the time we launch dfx without btc canister, we will be constantly getting error toasts because of regtest BTC balance "update" request. It will also affect e2e tests.
2. Other envs: BTC testnet still has some issue on the IC side - for instance on beta it constantly fails with `IC0503: Error from Canister g4xu7-jiaaa-aaaan-aaaaq-cai: Canister called `ic0.trap` with message: Panicked at 'Bitcoin API is disabled', canister/src/lib.rs:258:13.
Consider gracefully handling failures from this canister or altering the canister to handle exceptions`. This issue is known, but I'm not sure how long will it take to get it fixed. Therefore, I wanted to avoid showing error toasts to users with test networks settings on.

WDYT? I can add some comments to the code with the above info.

peterpeterparker: On one hand, I hear you, it makes sense to silence those errors, on the other, we should not have to silence the error and rather fix or adapt our test environments. For example, Bitcoin should be entirely disabled.

I just fear that we silence an error here, then later add another "if dev" there, and then another "if dev" there etc. ultimately we end with lots of exception instead of a clean solution. You know what I mean?

But again, I hear you, so if you want to move on with this option, not against for sure. Maybe then let's rename it, add a checks that it's only use on DEV and a comment.

DenysKarmazynDFINITY: Yes, totally agree. Ideally, we should skip worker launch locally if BTC canister is not deployed. Added a todo comment and renamed the param.

DenysKarmazynDFINITY:
@peterpeterparker I added test for btc-listener.services in this PR, please take a look.

peterpeterparker (approved):
LGTM, thx

PR #3841, opened 2024-12-04
style(frontend): use toolbar for btns in ic receive info ck eth

loki344:
# Motivation

Uniform layout for modals. This MR prepares for the new layout that depends on the usage of the toolbar for footer elements.

# Changes

- group both btns in toolbar

# Tests

before:
image

after:
image


peterpeterparker (approved):
LGTM, thx

PR #3842, opened 2024-12-04
feat(frontend): optional reload prop in loadUserProfile

AntonioVentilii:
# Motivation

We want to provide the choice of not reloading the user profile if not necessary.

# Changes

- Add optional prop `reload` to `loadUserProfile`
- Default to `true`, since it is the current behavior

# Tests

Included tests.


src/frontend/src/lib/services/load-user-profile.services.ts
```diff
@@ -60,10 +60,16 @@ export const loadCertifiedUserProfile = async ({
...
 
 export const loadUserProfile = async ({
-	identity
+	identity,
+	reload = true
 }: {
 	identity: OptionIdentity;
+	reload?: boolean;
 }): Promise<void> => {
+	if (nonNullish(get(userProfileStore)) && !reload) {
```
peterpeterparker, line 69: I was asking myself it we should pass `$userProfileStore` as a parameter or not. Where you hesitating too?

peterpeterparker (approved):
LGTM, thx

PR #3844, opened 2024-12-04
feat(frontend): extract User Profile loader

AntonioVentilii:
# Motivation

The user profile will be used not only for the Settings page, but for other components through OISY (for example the carousel settings, see PR #3770 ).

So, we first extract it into a loader component at the top level layout.

However, we still need to force a refresh sometimes, so we make it listen to a specific event.

# Changes

- Extract component `LoaderUserProfile`.
- Load user profile on `onMount`.
- Create custom event `oisyRefreshUserProfile`.
- Use the event to re-load the user profile in the new component.
- Adapt `Settings`.

# Tests

Included some unittest for the new component.


src/frontend/src/lib/components/loaders/LoaderUserProfile.svelte
```diff
@@ -0,0 +1,21 @@
...
+	import { onMount } from 'svelte';
+	import { authIdentity } from '$lib/derived/auth.derived';
+	import { loadUserProfile } from '$lib/services/load-user-profile.services';
+	import type { OptionIdentity } from '$lib/types/identity';
+
+	let identity: OptionIdentity;
+	$: identity = $authIdentity;
+
+	onMount(() => {
+		loadUserProfile({ identity });
```
peterpeterparker, line 11: It works because there are guards but, are we sure at this point that the autosubscriber really kicked in?

```
let identity: OptionIdentity;
$: identity = $authIdentity; // <--- Did this happen?
	
onMount(() => {
  loadUserProfile({ identity }); // <- Was is really set
```

To be on the safe side, I would listen to the auto subscription, reset the store if identity is undefined else load.

WDYT?

AntonioVentilii: I don't know if I understood correctly, have a look at latest commit, is that what you had in mind?

peterpeterparker: Almost, following is missing:

```
const load = async ({ reload = false }: { reload?: boolean }) => {

                // In this function I would just use the same store as the one subscribed, just in case of race condition
                // I doubt it happens, just because I'm paranoiac

			userProfileStore.reset();
			return;
		}
		
		loadUserProfile({ identity: $authIdentity, reload });
	};
	
// Remove onMount

// Auto subscribe to trigger load
$: $authIdentity, load()
```

AntonioVentilii: Ah! understood! A guard on the `authIdentity`!

AntonioVentilii: done, have a look please

src/frontend/src/lib/components/settings/Settings.svelte
```diff
@@ -44,9 +44,7 @@
 	};
 
 	$: {
-		if ($authSignedIn && POUH_ENABLED) {
-			loadUserProfile({ identity });
-		}
+		emit({ message: 'oisyRefreshUserProfile' });
```
peterpeterparker, line 47: - Do we actually need refresh? Settings isn't wrapper with the loader component?
 - If we have to keep "refresh", the we can still maybe reload only if `POUH_ENABLED`?

AntonioVentilii: you are right, it does not need necessarily a refresh, removed!

src/frontend/src/app.d.ts
```diff
@@ -24,6 +24,7 @@ declare namespace svelteHTML {
 		'on:oisyCkEthMinterInfoStatus'?: (event: CustomEvent<any>) => void;
 		'on:oisyCkEthereumPendingTransactions'?: (event: CustomEvent<any>) => void;
 		'on:oisyValidateAddresses'?: (event: CustomEvent<any>) => void;
+		'on:oisyRefreshUserProfile'?: (event: CustomEvent<any>) => void;
```
peterpeterparker, line 27: Now that we have removed the `emit`, this is still required`? Likewise for the `svelte:window` related listener?

AntonioVentilii: it will be for sure for the next PR: do you prefer I add it in the next one?

peterpeterparker: No I mean shouldn't it be removed - not added no? This PRs adds it but, it's not needed anymore no?

AntonioVentilii: indeed! removed! i will add it back when needed

peterpeterparker (approved):
LGTM, thx

PR #3848, opened 2024-12-04
feat(frontend): load utxos fee on convert modal open

DenysKarmazynDFINITY:
# Motivation

In this PR, 2 updates on UtxosFeeContext have been implemented:

1. We now start loading utxos fee directly when user opens the convert modal. To do that, we pass to the endpoint a default amount specified in the code. The fees are usually the same (no matter what "amount" is provided) for a long period of time, therefore it won't lead to an incorrect value being displayed. 
2. After we got a fee value, we don't instantly reset the store if user provides a new amount. Instead, we continue displaying the old value and fetch the new fee in the background.


peterpeterparker (approved):
LGTM, thx

Already approving, just one question if interesting: If the user enters an amount and then clear the input field, are we trigerring an unecessery load of the fee?

DenysKarmazynDFINITY:
@peterpeterparker good point! I added an additional check, as well as a test case for that.

PR #3851, opened 2024-12-04
feat(frontend): installing Solana SDK

AntonioVentilii:
# Motivation

We install Solana SDK that will come useful for the integration.

Documentation at: https://solana.com/docs/clients/javascript

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
 		"@dfinity/verifiable-credentials": "^0.0.4",
 		"@junobuild/analytics": "^0.0.28",
 		"@metamask/detect-provider": "^2.0.0",
+		"@solana/web3.js": "^2.0.0",
 		"@walletconnect/web3wallet": "1.14.0",
 		"alchemy-sdk": "3.4.1",
 		"buffer": "^6.0.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -65,6 +65,7 @@
 		"@dfinity/verifiable-credentials": "^0.0.4",
 		"@junobuild/analytics": "^0.0.28",
 		"@metamask/detect-provider": "^2.0.0",
+		"@solana/web3.js": "^2.0.0",
```
peterpeterparker, line 68: ⚠️ The solana web3js library has been literally compromised yesterday. From what I remember reading it on twitter late in the evening, I think it was older version, something like version v1.9.xxx but, can you please double check this version is safe?

AntonioVentilii: on it!

peterpeterparker: Prodsec monitors this but, I thought it was better to start on a good fit. Thanks for double checking.

peterpeterparker (approved):
LGTM, thx

Adding such a chunk will have an impact on the JS bundle. We gonna need to have a look at the chunking to see how it is impacted and what's the results in term of performance.

PR #3853, opened 2024-12-04
feat(frontend): show total convert fee only when all values available

DenysKarmazynDFINITY:
# Motivation

Instead of using 0 as a back up value for a fee that hasn't been loaded yet, we will now be showing a skeleton. Also, total fee will only be displayed when all values are available.


peterpeterparker (approved):
LGTM, thx

PR #3855, opened 2024-12-04
feat(frontend): do not show Utxos warning if no amount provided

DenysKarmazynDFINITY:
# Motivation

We now fetch utxos even if no amount is provided by the user, so it's possible that the no-utxos warning message is shown before user actually types something. To avoid that, we add a small check. 

P.S.: pending-tx message will still be shown on `hasPendingTransactionsStore` availability.

### DIFF
--- a/src/frontend/src/btc/components/convert/BtcConvertForm.svelte
+++ b/src/frontend/src/btc/components/convert/BtcConvertForm.svelte
@@ -10,6 +10,7 @@
 		initPendingSentTransactionsStatus
 	} from '$btc/derived/btc-pending-sent-transactions-status.derived';
 	import { UTXOS_FEE_CONTEXT_KEY, type UtxosFeeContext } from '$btc/stores/utxos-fee.store';
+	import type { UtxosFee } from '$btc/types/btc-send';
 	import ConvertForm from '$lib/components/convert/ConvertForm.svelte';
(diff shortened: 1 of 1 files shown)


src/frontend/src/btc/components/convert/BtcConvertForm.svelte
```diff
@@ -58,7 +58,7 @@
 		{:else if nonNullish($hasPendingTransactionsStore)}
 			<div class="mb-4" data-tid="btc-convert-form-send-warnings">
 				<BtcSendWarnings
-					utxosFee={$storeUtxosFeeData?.utxosFee}
+					utxosFee={nonNullish(sendAmount) ? $storeUtxosFeeData?.utxosFee : undefined}
```
peterpeterparker, line 61: Maybe create a local state to derive the information? This would make the logic a bit more readable. WDYT?

Do we need a test or it's kind of covered?

DenysKarmazynDFINITY: Sounds good, updated.

Re. tests - correct, we already have a test case that checks if `BtcSendWarnings` appears correctly on `nonNullish($hasPendingTransactionsStore)`.

peterpeterparker (approved):
LGTM, thx

PR #3861, opened 2024-12-05
style(frontend): hide unnecessary elements in Carousel

AntonioVentilii:
# Motivation

When there is only one slide, we hide the bottom part of the carousel.

Screenshot 2024-12-05 at 10 11 32
Screenshot 2024-12-05 at 10 11 40


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3862, opened 2024-12-05
feat(frontend): Schnorr functions in signer canister

AntonioVentilii:
# Motivation

Adding the Schnorr functions for public key and signing in signer canister.


AntonioVentilii:
@bitdivine , is it worth maybe creating a `sol_address` like `eth_address` directly in the signer?

peterpeterparker (commented):
Worth also adding the tests ;)

AntonioVentilii:
@peterpeterparker for sure! But this was just a branch so that @bitdivine can take over the correct settings of the parameters first. We will add the tests soon after

peterpeterparker:
Even in a branch, I would not merge this standalone PR without its tests. Just my two cents though, feel free to ignore.

bitdivine:
Re: Creating a sol address in the signer: I think not, the general experience of the first round of integration was that the canister functionality needs to be minimal and logic needs to be pushed out to client libraries.  The signer is meant to be super-stable and many of the decisions ended up being customized for Oisy, which isn't good.

PR #3864, opened 2024-12-05
chore(devops): Import setup command

bitdivine:
# Motivation
The setup command is a general purpose command for installing command line tools.  This is used in future PRs for installing Solana and other tooling.

# Changes
- Copy the `scripts/setup` from the chain-fusion-signer codebase.
- Use the `.sh` suffix to adhere to Oisy policy.

# Tests
The tool is exercised extensively elsewhere.  The only adjustments are formatting and naming to match local policies.


AntonioVentilii (approved):
LGTM tks!

Not really sure if worth it, but a mention in HACK.MD or HOW-TO? Or unnecessary?

bitdivine:
I think it's not necessary at this time.  Maybe when we have instructions for how to set up an environment we can add that.

PR #3865, opened 2024-12-05
feat(frontend): derived store for user settings

AntonioVentilii:
# Motivation

We create a derived store for the user settings, so that we can handle eventual `nullable` values.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3866, opened 2024-12-05
feat(test): Skeleton binding crates

bitdivine:
# Motivation
We would like comprehensive pocket-ic tests for Oisy.

To that end, we would like to generate canister bindings to be used in pocket-ic tests rather than maintaining the bindings manually.  We will start with the ledger bindings, then move on to the chain fusion signer and rewards canisters.

There already bindings for use inside a canister.  There is a TODO to autogenerate them.

However it is inconvenient and sometimes very confusing if the API types are generated twice, with identical names.  We will therefore make `didc` generate a file with just Rust types, and the two (pocket-ic and canister) bindings will use those shared types.

A further constraint is that pocket-ic and canister bindings cannot be compiled together into the same crate.  We therefore need three separate crates for types, pocket-ic bindings and internal canister bindings.

There is quite  a lot to do here, so we break it down into small steps.

# Changes
- Create the Rust crates needed to support pocket-ic and types bindings.

# Tests
- Existing CI should verify that the crates are well formed.
- The crates are currently empty, so there is nothing else to test at this time.


Cargo.toml
```diff
@@ -2,6 +2,8 @@
 members = [
     "src/backend",
     "src/cycles_ledger/client",
+    "src/cycles_ledger/pic",
+    "src/cycles_ledger/types",
     "src/shared"
```
bitdivine, line 7: Note:
- `backend` and `shared` are manually written crates.
- `client` is generated by `didc` and then manually modified; that manual modification will be automated in a future PR.
- `pic` and `types` are the new crates with type definitions and pocket-ic bindings.

All the code below is the boilerplate code needed to add these new crates.

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3867, opened 2024-12-05
feat(frontend): save carousel hidden slides to backend

AntonioVentilii:
# Motivation

We use the backend method `add_hidden_dapp_id` introduced with PR #3777  , to save the hidden slides/dApps in the backend.

# Changes

- Add event `oisyRefreshUserProfile` to refresh the user profile in `LoaderUserProfile`. It is useful to emit it when we know that we change the user profile.
- Add method `add_hidden_dapp_id` in `DappCarousel` when closing a slide.

# Tests

https://github.com/user-attachments/assets/c849ed22-dfa3-4f51-90e8-5d6a446b5dfb

### DIFF
--- a/src/frontend/src/lib/components/dapps/DappsCarousel.svelte
+++ b/src/frontend/src/lib/components/dapps/DappsCarousel.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { nonNullish } from '@dfinity/utils';
+	import { fromNullable, isNullish, nonNullish } from '@dfinity/utils';
 	import Carousel from '$lib/components/carousel/Carousel.svelte';
 	import DappsCarouselSlide from '$lib/components/dapps/DappsCarouselSlide.svelte';
 	import {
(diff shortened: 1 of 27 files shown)


src/frontend/src/lib/components/loaders/LoaderUserProfile.svelte
```diff
@@ -14,6 +14,11 @@
 	};
 
 	$: $authIdentity, load({});
+	const reload = () => {
```
AntonioVentilii, line 17:
```suggestion
	const reload = () => {
```

src/frontend/src/lib/components/dapps/DappsCarousel.svelte
```diff
@@ -30,6 +36,18 @@
 		if (idx !== -1) {
 			carousel.removeSlide(idx);
 		}
+
+		if (isNullish($authIdentity) || isNullish($userProfileStore)) {
```
DenysKarmazynDFINITY, line 40: For `isNullish($authIdentity)` we can use `await nullishSignOut();`.

DenysKarmazynDFINITY (approved):
Small comment, otherwise LGTM, thanks!

src/frontend/src/lib/components/dapps/DappsCarousel.svelte
AntonioVentilii: Good one!!! Included!

PR #3869, opened 2024-12-05
feat(frontend): function to remove slide for Carousel

AntonioVentilii:
# Motivation

The component Carousel is not reactive to its children in `<slot />` (it will be possible only in Svelte 5). But we need to have a way to "refresh" the slides. Our first use-case is to remove a specific slide.

Since the component modify its children on mounting, we create a function that can be called outside the component to remove a slide at a particular index position.

Suggested by @DenysKarmazynDFINITY .

# Changes

- Add function `removeSlide` to `Carousel` and export it.
- Re-initialise the `Carousel` once the slide is removed.
- Stop timer if there is only one slide.
- Add transitions.

# Tests

We still don't have unittest with custom components (like GIX Components), nor an E2E test for this. However, I simulated a random removal of one of the slides from the parent of Carousel.

```typescript
// every 5 seconds remove a dapp from the carousel
const interval = setInterval(() => {
  if (nonNullish(carousel)) {
    const dapp = dappsCarouselSlides[Math.floor(Math.random() * dappsCarouselSlides.length)];
    
    console.log('removing', dapp?.id);
    const idx = dappsCarouselSlides.findIndex((d) => d.id === dapp?.id);
    
    carousel.removeSlide(idx);
  }
}, 5000);
```

The results are:

https://github.com/user-attachments/assets/2acb4440-ba94-4879-ace0-c368035aa4c1

### DIFF
--- a/src/frontend/src/lib/components/carousel/Carousel.svelte
+++ b/src/frontend/src/lib/components/carousel/Carousel.svelte
@@ -246,8 +246,6 @@
 	};
 
 	export const removeSlide = (idx: number) => {
-
-
 		slides = slides.filter((_, i) => i !== idx);
 
 		initializeCarousel();
--- a/src/frontend/src/lib/components/carousel/Carousel.svelte
+++ b/src/frontend/src/lib/components/carousel/Carousel.svelte
@@ -1,12 +1,12 @@
 <script lang="ts">
 	import { isNullish, nonNullish } from '@dfinity/utils';
 	import { onMount } from 'svelte';
+	import { slide } from 'svelte/transition';
 	import Controls from '$lib/components/carousel/Controls.svelte';
 	import Indicators from '$lib/components/carousel/Indicators.svelte';
 	import { CAROUSEL_CONTAINER } from '$lib/constants/test-ids.constants';
-	import { moveSlider, extendCarouselSliderFrame } from '$lib/utils/carousel.utils';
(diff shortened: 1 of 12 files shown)


src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -244,6 +244,12 @@
 		const offset = (currentSlide + 1) * containerWidth;
 		moveSlider({ sliderFrame, animateTo: -offset, withTransition, duration, easing });
 	};
+
+	export const removeSlide = (idx: number) => {
+		slides = slides.filter((_, i) => i !== idx);
```
DenysKarmazynDFINITY, line 249: Was just curious - looks like for arrays with length > 100, [filter is faster than slice](https://medium.com/@justintulk/javascript-performance-array-slice-vs-array-filter-4573d726aacb) 🙂

DenysKarmazynDFINITY (approved):
Cool, this is how I also verified it locally - I was removing a random slide on carousel click, and it worked perfectly fine.

src/frontend/src/lib/components/carousel/Carousel.svelte
AntonioVentilii: yes!!! you never know how many dapps we will have there... hopefully not 100 eheh

PR #3872, opened 2024-12-05
feat(ci): auto-update package-lock file

AntonioVentilii:
# Motivation

it would be useful to have a check/action that auto-updates the `package-lock.json` file if there is a need for it if `package.json` has changed.

Note that we do not specify the change because if during a PR, the `package.json` is reverted, the `package-lock.json` risks that is not.


AntonioVentilii:
this PR is not really necessary, so we close it

PR #3873, opened 2024-12-05
test(frontend): Tests for certified store init

AntonioVentilii:
# Motivation

We create some base tests for `initCertifiedStore` and `initCertifiedSetterStore`

Note that we are not creating tests for the `update` function since it is replicated as it is from `writable` constructor of `svelte`. Nor we create a test for `reset` in `initCertifiedSetterStore` since it is a copy of the same function inherited from `initCertifiedStore`.


peterpeterparker (approved):
That's really cool, thanks!

PR #3874, opened 2024-12-05

AntonioVentilii:
# Motivation

Following PR #3689 , we replace PATs generated by service accounts with GitHub Apps.


.github/workflows/update-snapshots.yml
```diff
@@ -31,7 +39,7 @@ jobs:
         with:
           repository: ${{ github.event.pull_request.head.repo.full_name }}
           ref: ${{ github.event.pull_request.head.ref }}
-          token: ${{ secrets.GIX_CREATE_PR_PAT }}
+          token: ${{ steps.app-token.outputs.token }}
```
cgundy, line 42: So I'm wondering if this is needed at all here. It looks like the `oisy-backend` action only pushes a docker image, it doesn't create a PR, so I think you could remove the token here and add permissions to the workflow.

.github/workflows/update-snapshots.yml
```diff
@@ -21,6 +21,14 @@ jobs:
         os: [ubuntu-24.04, macos-14]
     if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-e2e-snapshots') }}
```
cgundy, line 22: Here you would add:
```
permissions:
    contents: write
    packages: write
```

cgundy (approved):

.github/workflows/update-snapshots.yml
AntonioVentilii: You are right! Removed

.github/workflows/update-snapshots.yml
AntonioVentilii: included, tks!

peterpeterparker (approved):
Rubber stamp.

PR #3875, opened 2024-12-05
refactor(frontend): add SVG schema to network schema

AntonioVentilii:
# Motivation

As suggested in the past, we should refine the schema of network type to be more specific for the icons.


peterpeterparker (approved):
Nice, thx!

PR #3876, opened 2024-12-06
feat(frontend): add close button to carousel slides

AntonioVentilii:
# Motivation

We add a close button to the carousel slides to hide them.

# Note

This affects only current navigation. Persisting that will be applied in another PR, using the backend method `add_hidden_dapp_id` introduced with PR #3777  

# Changes

- Add button close to component `DappsCarouselSlide`.
- Bubble event `icCloseCarouselSlide` that provides the dApp that is being closed.
- Create function in `DappsCarousel` to manage the event and remove the correspondent slide.
- Adjust `Carousel` function removeSlide to refresh all needed variables.
- Add transitions.

# Tests

https://github.com/user-attachments/assets/873fa742-5307-48a2-b81a-2276e54afdc7

### DIFF
--- a/src/frontend/src/lib/components/carousel/Carousel.svelte
+++ b/src/frontend/src/lib/components/carousel/Carousel.svelte
@@ -279,7 +279,7 @@
 	{#if nonNullish(slides) && slides.length > 1}
 		<div
 			class={`absolute bottom-2 right-0 flex justify-between px-3 ${controlsWidthStyleClass ?? 'w-full'}`}
-			transition:slide={SLIDE_PARAMS}
+			out:slide={SLIDE_PARAMS}
 		>
 			<Indicators {onIndicatorClick} {totalSlides} {currentSlide} />
 			<Controls {onNext} {onPrevious} />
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -250,9 +250,13 @@
...
 		slides = slides.filter((_, i) => i !== idx);
 
+		totalSlides = slides.length;
+
 		initializeCarousel();
 
-		if (slides.length === 1) {
+		goToNextSlide();
+
+		if (slides.length <= 1) {
```
DenysKarmazynDFINITY, line 259: nitpick: you can use `totalSlides` here

AntonioVentilii: good one! done!

src/frontend/src/lib/components/carousel/Carousel.svelte
```diff
@@ -274,7 +279,7 @@
 	{#if nonNullish(slides) && slides.length > 1}
 		<div
 			class={`absolute bottom-2 right-0 flex justify-between px-3 ${controlsWidthStyleClass ?? 'w-full'}`}
-			transition:slide={SLIDE_PARAMS}
+			out:slide={SLIDE_PARAMS}
```
DenysKarmazynDFINITY, line 282: Wondering - wasn't `transition` covering the `out` case?

DenysKarmazynDFINITY (approved):
Small questions, otherwise LGTM 👍

src/frontend/src/lib/components/carousel/Carousel.svelte
AntonioVentilii: yes yes, i removed the `in` transition in this way: I never noticed but it was a bit strange if you replay it slowly eheh! not noticeable too much, but you never know

PR #3877, opened 2024-12-06
style(frontend): calculates amount for btc transactions correctly

BonomoAlessandro:
# Motivation

On the transactions page the BTC send transactions are displayed as if the user is receiving tokens instead of sending them. The amount of a `send` transaction should be displayed with a leading `-` (minus) and the text should be black.

# Changes

- fixes calculation

# Tests

**before:**
image

**after:**
image

### DIFF
--- a/src/frontend/src/btc/components/transactions/BtcTransaction.svelte
+++ b/src/frontend/src/btc/components/transactions/BtcTransaction.svelte
@@ -22,7 +22,7 @@
 	let label: string;
 	$: label = type === 'send' ? $i18n.send.text.send : $i18n.receive.text.receive;
 
-	let amount: BigNumber;
+	let amount: BigInt | undefined;
 	$: amount = nonNullish(value) ? (type === 'send' ? value * BigInt(-1) : value) : undefined;
 </script>
 
--- a/src/frontend/src/btc/components/transactions/BtcTransaction.svelte
+++ b/src/frontend/src/btc/components/transactions/BtcTransaction.svelte
@@ -23,7 +23,7 @@
 	$: label = type === 'send' ? $i18n.send.text.send : $i18n.receive.text.receive;
 
 	let amount: BigNumber;
-	$: amount = nonNullish(value) ? type === 'send' ? value * BigInt(-1) : value : undefined;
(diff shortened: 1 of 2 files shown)


src/frontend/src/btc/components/transactions/BtcTransaction.svelte
```diff
@@ -21,11 +21,14 @@
 
 	let label: string;
 	$: label = type === 'send' ? $i18n.send.text.send : $i18n.receive.text.receive;
+
+	let amount: bigint | undefined;
+	$: amount = nonNullish(value) ? (type === 'send' ? value * BigInt(-1) : value) : undefined;
```
AntonioVentilii, line 26: just to be sure, better use `value.mul(BigNumber.from(-1))` here, you never know with `BigInt`

src/frontend/src/tests/btc/components/transactions/BtcTransaction.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+			transaction: mockTransaction,
+			token: DEFAULT_BITCOIN_TOKEN
+		};
+
+		const { getByText } = render(BtcTransaction, { props });
+
+		expect(getByText('Send')).toBeInTheDocument();
+		expect(getByText('-0.00040827')).toBeInTheDocument();
+	});
+	it('should calculate amount for receive transaction correctly', () => {
```
AntonioVentilii, line 27:
```suggestion

	it('should calculate amount for receive transaction correctly', () => {
```

AntonioVentilii (approved):
LGTM tks! Smaller comments before merging

src/frontend/src/btc/components/transactions/BtcTransaction.svelte
BonomoAlessandro: @AntonioVentilii `value` is of type `bigint` and does not provide a function `mul`.

AntonioVentilii: ah! then all good! I thought it was `BigNumber`, my mistake, sorry

PR #3878, opened 2024-12-06
fix(frontend): correct IconSchema

AntonioVentilii:
# Motivation

In PR #3875 , I introduced the IconSchema, considering only the environments where the local files were passed as string path. However, in production, staging, and in general in any non-local deployed environment, the element is the SVG file content.

That is why we fix the schema here.


DenysKarmazynDFINITY (approved):
Thanks!

PR #3883, opened 2024-12-06
ci(e2e): improved e2e snapshotting

inc-man:
# Motivation

Our E2E Tests often fail due to flakiness or instability of snapshots. With these added features we will ensure better stability for snapshots.

# Changes

added functionality to snapshotting

# Tests

ran workflows successfully


playwright.config.ts
```diff
@@ -38,9 +46,14 @@ export default defineConfig({
...
 	projects: [
+		/* Test against desktop browsers. */
 		{
 			name: 'Google Chrome',
-			use: { ...devices['Desktop Chrome'], channel: 'chrome' }
+			use: { ...devices['Desktop Chrome'] }
+		},
+		{
+			name: 'Firefox',
+			use: { ...devices['Desktop Firefox'] }
```
AntonioVentilii, line 56: this in another PR right?

AntonioVentilii (commented):
Shall we create a generic method inside Homepage so that we avoid repetition of code?

```typescript
// pseudo code
toHaveScreenshotWithTimeout = async (
  options: Parameters<typeof page.toHaveScreenshot>[0] = {}
) => {
  return await self.#page.toHaveScreenshot({
    ...options,
    timeout: 1000 * 60,
  });
};
```

playwright.config.ts
inc-man: absolutely, that was a copy mistake I did from the other PR. will be removed

playwright.config.ts
```diff
@@ -22,6 +22,14 @@ const TIMEOUT = 5 * 60 * 1000;
 export default defineConfig({
 	timeout: TIMEOUT,
 	workers: DEV ? 5 : 2,
+	expect: {
+		toHaveScreenshot: {
+			threshold: 0.25,
+			maxDiffPixelRatio: 0.025,
+			animations: 'disabled',
+			caret: 'hide'
```
AntonioVentilii, line 30: as discussed offline, let's not make the tests more lax, let's keep them as they are and focus to check where and how the differences arises

inc-man:
superseded by https://github.com/dfinity/oisy-wallet/pull/4019

PR #3886, opened 2024-12-06
fix(e2e): added conditial device usage

inc-man:
# Motivation

We want to ensure that macOS runners are ONLY used for Apple devices and that everything else is being tested on ubuntu runners.

# Changes

adjusted the configuration for handling conditional device usage based on os.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -17,6 +17,35 @@ dotenv.populate(
 
 const DEV = (process.env.NODE_ENV ?? 'production') === 'development';
 
+const MATRIX_OS = process.env.MATRIX_OS || '';
+const isMac = MATRIX_OS.includes('macos') || process.platform === 'darwin';
+
+const appleProjects = [
+	{
+		name: 'Safari',
+		use: devices['Desktop Safari'],
+	},
+	{
+		name: 'iPhone SE',
+		use: devices ['iPhone SE'],
+	},
+];
+
+const nonAppleProjects = [
+	{
+		name: 'Google Chrome',
+		use: devices['Desktop Chrome'],
+	},
+	{
+    name: 'Firefox',
+    use: devices['Desktop Firefox'],
+  },
+  {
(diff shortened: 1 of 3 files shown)


playwright.config.ts
```diff
@@ -17,6 +17,35 @@ dotenv.populate(
 
 const DEV = (process.env.NODE_ENV ?? 'production') === 'development';
 
+const MATRIX_OS = process.env.MATRIX_OS ?? '';
+const isMac = MATRIX_OS.includes('macos') ?? process.platform === 'darwin';
```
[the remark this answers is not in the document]
inc-man: we cannot do || because of Lint Check

e2e/utils/pages/homepage.page.ts
```diff
@@ -228,25 +228,23 @@ abstract class Homepage {
...
-		await this.clickByTestId(NAVIGATION_ITEM_TOKENS);
+		await this.clickByTestId(NAVIGATION_ITEM_HOMEPAGE);
 	}
 
-	async takeScreenshot(): Promise<void> {
+	async takeScreenshot(isMobile: boolean): Promise<void> {
 		await expect(this.#page).toHaveScreenshot({
 			// creates a snapshot as a fullPage and not just certain parts.
-			fullPage: true,
+			fullPage: !isMobile,
```
DenysKarmazynDFINITY, line 247: Would be great to capture full pages for mobile too. Any particular reason you disabled it?

DenysKarmazynDFINITY: Yeah, guessed so, but I think we still need to have keep it as `true` - on mobile, most of the content will be hidden if we don't do so, therefore we would be ignoring 60%+ of a page without a full snapshot.

DenysKarmazynDFINITY (approved):
As discussed offline, the ETH txs page test flakiness will be investigated and fixed in a separate PR.

PR #3888, opened 2024-12-06
fix(frontend): btc wallet error hideToast option

DenysKarmazynDFINITY:
# Motivation

Fix for incorrect `hideToast` param handling in `btc-listener.services`.

### DIFF
--- a/src/frontend/src/btc/services/btc-listener.services.ts
+++ b/src/frontend/src/btc/services/btc-listener.services.ts
@@ -56,7 +56,7 @@ export const syncWalletError = ({
 	// Currently, the certified error can only happen while fetching balance, but we still want to reset transactions to avoid displaying incorrect data
 	btcTransactionsStore.reset(tokenId);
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/btc/services/btc-listener.services.ts
```diff
@@ -56,7 +56,7 @@ export const syncWalletError = ({
 	// Currently, the certified error can only happen while fetching balance, but we still want to reset transactions to avoid displaying incorrect data
 	btcTransactionsStore.reset(tokenId);
 
-	if (!hideToast) {
+	if (hideToast) {
```
peterpeterparker, line 59: Since we have discussed quite a bit about it and the fact that I missed it in the review (😅), can you provide a test for it (next week)?

DenysKarmazynDFINITY: Absolutely, that was also my thought after I noticed it a few mins ago 🙃

PR #3889, opened 2024-12-06
feat(frontend): avoid showing toast on errors with BTC testnet

DenysKarmazynDFINITY:
# Motivation

Since BTC testnet is not working at the moment, we want to avoid showing toast for all the users that have "testnet" setting on. We will remove it after it is fixed on the canister side.

### DIFF
--- a/src/frontend/src/btc/services/worker.btc-wallet.services.ts
+++ b/src/frontend/src/btc/services/worker.btc-wallet.services.ts
@@ -1,6 +1,5 @@
 import { syncWallet, syncWalletError } from '$btc/services/btc-listener.services';
 import type { BtcPostMessageDataResponseWallet } from '$btc/types/btc-post-message';
-import { LOCAL } from '$lib/constants/app.constants';
 import {
 	btcAddressMainnetStore,
 	btcAddressRegtestStore,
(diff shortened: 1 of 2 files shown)


src/frontend/src/btc/services/worker.btc-wallet.services.ts
```diff
@@ -48,8 +47,12 @@ export const initBtcWalletWorker = async ({
...
 					tokenId,
 					error: (data.data as PostMessageDataResponseError).error,
-					// TODO: do not launch worker locally if BTC canister is not deployed, and remove the below param afterwards
-					hideToast: isRegtestNetwork && LOCAL
+					/**
+					 * TODOs:
+					 * 1. Do not launch worker locally if BTC canister is not deployed, and remove "isRegtestNetwork" afterwards.
+					 * 2. Wait for testnet BTC canister to be fixed on the IC side, and remove "isTestnetNetwork" afterwards.
+					 * **/
+					hideToast: isRegtestNetwork || isTestnetNetwork
```
DenysKarmazynDFINITY, line 55: I removed `LOCAL` since it is not really needed here - `isRegtestNetwork` can only be launched locally.

AntonioVentilii (approved):
LGTM tks

src/frontend/src/btc/services/worker.btc-wallet.services.ts
DenysKarmazynDFINITY: Just pinging @peterpeterparker to be in the loop - we discussed this part previously, and agreed not to hide errors outside of localhost, but it seems we are forced to have an additional check to avoid having toasts shown every time someone loads BTC testnet wallet.

PR #3891, opened 2024-12-06
style(frontend): make only close button clickable in DappsCarouselSlide

AntonioVentilii:
# Motivation

We reduce the clickable space for the close button in `DappsCarouselSlide`.


DenysKarmazynDFINITY (approved):
I assume the UI stays unchanged, right?

LGTM, thanks

PR #3892, opened 2024-12-07
feat(frontend): add signout when authIdentity is nullish

AntonioVentilii:
# Motivation

As suggested by @DenysKarmazynDFINITY , we include function `nullishSignOut` in the check for `authIdentity` inside `LoaderUserProfile`.


peterpeterparker (commented):
We should probably not do this change here as it can lead to a race condition where sign out might happens twice or more?

Use click on "Logout".
Logout function reset auth identity. (it reloads the window but let's assume here for some reason the window is not reloaded)
Auth identity is null.
Subscriber start.
Auth identity is null.
Logout function call.
Logout function reset identity.
etc.

AntonioVentilii:
Agreed! PR closed

PR #3893, opened 2024-12-07
fix(frontend): load Carousel when missing settings in user profile

AntonioVentilii:
# Motivation

The carousel was not being loaded for old accounts, because `userSettings` was always `undefined`.

However, we should distinguish between not having loaded the user profile (from which `userSettings` derives), and having a missing field `settings` in the user profile (it happens for old accounts, since we introduced the field recently).

# Changes

- Improve `userSettings` derived store to set `null` when the user profile exists but it is missing the `settings` field.
- Adapt tests.
- Adapt `DapsCarousel`.

# Tests

Provided.

### DIFF
--- a/src/frontend/src/lib/components/dapps/DappsCarousel.svelte
+++ b/src/frontend/src/lib/components/dapps/DappsCarousel.svelte
@@ -57,7 +57,7 @@
 	};
 </script>
 
-{#if $userSettings!==undefined && nonNullish(dappsCarouselSlides) && dappsCarouselSlides.length > 0}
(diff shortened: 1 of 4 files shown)


peterpeterparker (approved):
LGTM, thx

Note: I actually read the code for quite some minutes because I'm not entirely sure making a distinction `undefined` or `null` should be something that happens at the derive level. Another option I guess would have been having a dedicated `$userSettingsLoaded: Readable<boolean>` information but just idea so all good.

AntonioVentilii:
@peterpeterparker makes sense! Have a look at please, is that what you thought:

- PR #3895

PR #3894, opened 2024-12-07
ci(frontend): merge auto-format in checks

AntonioVentilii:
# Motivation

It would make sense to merge the auto-format CI into the frontend checks, since they are already checking the format. So that we spare one Ci and already fix the formatting if needed, during the checks.

Plus, it is quicker.

# Tests

I injected bad formatting in this same PR and it was fixed.


bitdivine (approved):
Warning: A GitHub actions event can not normally trigger another GitHub actions event. For example, if CI adds a commit, that will not normally trigger a re-run of CI. If this PR works, great, and I would be interested to see what EndBug does other than hit commit and push, but please be alert for CI no longer working correctly for PRs with automated changes. There have been many attempts to do things like this PR in a variety of reposts and they usually have to get reverted after causing flakiness. But if it works, great.

PR #3895, opened 2024-12-07
feat(frontend): derived store userProfileLoaded

AntonioVentilii:
# Motivation

According to @peterpeterparker 's suggestion in https://github.com/dfinity/oisy-wallet/pull/3893#pullrequestreview-2486419830 , we find a better way to distinguish if the user settings are not available because the user profile is not loaded, or if they are missing (old accounts for example).

The base difference is simply if the user profile is not `nullish`: if it is not `nullish` and the settings are nullish, we are sure that the `settings` are missing.

# Changes

- Create derived store `userProfileLoaded`.
- Use new store in `DappsCarousel`.

# Tests

Test created and adjusted.

### DIFF
--- a/src/frontend/src/lib/derived/user-profile.derived.ts
+++ b/src/frontend/src/lib/derived/user-profile.derived.ts
@@ -4,11 +4,12 @@ import type { Option } from '$lib/types/utils';
 import { fromNullable, nonNullish } from '@dfinity/utils';
 import { derived, type Readable } from 'svelte/store';
 
+export const userProfileLoaded: Readable<boolean> = derived([userProfileStore], ([$userProfile]) =>
+	nonNullish($userProfile)
+);
+
 export const userSettings: Readable<Option<Settings>> = derived(
 	[userProfileStore],
 	([$userProfile]) =>
-		// The user profile may or may not have the settings field.
-		// To distinguish between user profile not loaded and missing settings,
-		// we use `undefined` for the former and `null` for the latter.
-		nonNullish($userProfile) ? (fromNullable($userProfile.profile.settings) ?? null) : undefined
+		nonNullish($userProfile) ? fromNullable($userProfile.profile.settings) : undefined
 );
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/derived/user-profile.derived.ts
```diff
@@ -1,14 +1,18 @@
...
-		// To distinguish between user profile not loaded and missing settings,
-		// we use `undefined` for the former and `null` for the latter.
-		nonNullish($userProfile) ? (fromNullable($userProfile.profile.settings) ?? null) : undefined
+		nonNullish($userProfile) ? fromNullable($userProfile.profile.settings) : undefined
+);
+
+export const userSettingsLoaded: Readable<boolean> = derived(
+	[userProfileStore, userSettings],
+	([$userProfile, $userSettings]) =>
+		nonNullish($userSettings) ||
```
peterpeterparker, line 16: Would `fromNullable($userProfile?.profile.settings) !== undefined` do the tricks?

AntonioVentilii: the only case where it does not work that is when `userProfile` is loaded but `settings` are missing (old accounts). In that case it would be `fromNullable([]) !== undefined` that returns `false`, but we need `true`: it is ok that the `settings` are missing when the `userProfile` is loaded.

peterpeterparker: Gotcha. Then we can follow your approach but, can we resolve it without a dependency on `$userSettings`? I feel like we spare one evaluation if we solely use `$userProfileStore`

AntonioVentilii: in that case it will just be a `userProfileLoaded` check, is that ok? if the user profile is loaded, we don't care if settings are nullish or not, they are loaded

I had put that dependency just to not have that small amount of time where `userProfile` is not nullish and `userSettings` is still calculating; hopefully it does not impact that much, right?

peterpeterparker: Ah yeah actually that should do, there is no use case where the user profile is loaded but the settings arent, good idea 👍

peterpeterparker (approved):
LGTM, thx

PR #3909, opened 2024-12-09
ci(e2e): handling of mobile view on navigation items

inc-man:
# Motivation

While expanding the e2e tests we realized that certain elements are not visible on mobile view. For that we implemented a new method to handle this problem.

# Changes

added method to check if element is visible, and if so click it, if not visible locate different element and click it.


AntonioVentilii (approved):
LGTM tks!

PR #3914, opened 2024-12-10
feat(frontend): track dApp modal clicks

AntonioVentilii:
# Motivation

We start tracking the click on the hyperlink of the dApp modal.

# Changes

- Add tracking constants.
- Add `on:click` event in `ExternalLink` that uses a new prop `trackEventParams`.
- Add `trackEventParams` in `DappModal`.

### DIFF
--- a/src/frontend/src/lib/components/ui/ExternalLink.svelte
+++ b/src/frontend/src/lib/components/ui/ExternalLink.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { isNullish } from '@dfinity/utils';
 	import IconExternalLink from '$lib/components/icons/IconExternalLink.svelte';
-	import { trackEvent } from '$lib/services/analytics.services';
+	import { trackEvent as trackEventServices } from '$lib/services/analytics.services';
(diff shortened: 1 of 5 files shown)


AntonioVentilii:
@peterpeterparker as discussed offline, I changed the way we pass the tracking data to the `ExternalLink` component. It makes more sense to not save the complete URL as metadata, but even adding more freedom to the usage of `ExternalLink`.

src/frontend/src/lib/components/ui/ExternalLink.svelte
```diff
@@ -9,6 +12,15 @@
 	export let color: 'blue' | 'inherit' = 'inherit';
 	export let fullWidth = false;
 	export let styleClass = '';
+	export let trackEventParams: TrackEventParams | undefined = undefined;
```
peterpeterparker, line 15: nitpick: I would call the property `trackEvent` and rename the function `trackEvent as trackEventServices } from ...` to avoid conflicts. That way consuming the component is a bit less verbose.

AntonioVentilii: done!

PR #3916, opened 2024-12-10
style(frontend): use content with toolbar in AddTokenForms

loki344:
# Motivation

Uniform display of modals

# Changes

use ContentWitToolbar

# Tests

before:

image

image

after:

image

image


loki344:
FYI ContentWithToolbar is used in https://github.com/dfinity/oisy-wallet/blob/7184881408801322309270e4b611df66ce2db3fd/src/frontend/src/lib/components/manage/AddTokenByNetwork.svelte#L75

AntonioVentilii (approved):
LGTM tks

PR #3917, opened 2024-12-10
refactor(frontend): make list for ReceiveAddressModal

AntonioVentilii:
# Motivation

Maybe opinionated, but since we are going to add more Networks, it feels more manageable to create a list of `addresses` and their configs, to pass it to a loop inside `ReceiveAddressModal` to be shown.

# Changes

- Create interface of address configs.
- Create list of configs given by the existing components.
- Create loop.

# Tests

No changes.

### Before

Screenshot 2024-12-10 at 05 19 27
Screenshot 2024-12-10 at 05 19 33

### After

Screenshot 2024-12-10 at 05 30 57
Screenshot 2024-12-10 at 05 31 02

### DIFF
--- a/src/frontend/src/lib/components/receive/ReceiveAddresses.svelte
+++ b/src/frontend/src/lib/components/receive/ReceiveAddresses.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+	import { nonNullish } from '@dfinity/utils';
 	import { createEventDispatcher } from 'svelte';
 	import {
 		BTC_MAINNET_NETWORK,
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/receive/ReceiveAddresses.svelte
```diff
@@ -38,167 +39,137 @@
...
+	import type { Network } from '$lib/types/network';
 	import type { ReceiveQRCode } from '$lib/types/receive';
+	import type { Token } from '$lib/types/token';
 
 	const dispatch = createEventDispatcher();
 
 	const displayQRCode = (details: Omit<Required<ReceiveQRCode>, 'qrCodeAriaLabel'>) =>
 		dispatch('icQRCode', details);
+
+	interface ReceiveAddress {
```
peterpeterparker, line 52: There is a cmp name `ReceiveAddress`, maybe `ReceiveAddressProps` to avoid conflicts?

peterpeterparker (approved):
LGTM, thx

Just a naming suggestion.

PS: Note that I understand this components is already covered by E2E tests, if not the case please let know.

PR #3918, opened 2024-12-10
style(frontend): uses session storage instead of local storage for closable message

BonomoAlessandro:
# Motivation

In some parts of the application, we display closable message bloxes. When logged in, closed message boxes should not appear again. But, if the user logs out and then logs in again, the messages should be displayed again.

# Changes

- uses session storage instead of local storage.
- resets session storage on logout

# Tests

**before:**

https://github.com/user-attachments/assets/bf9c3b99-53fe-4a9c-a31e-33b744c5f6d6

**after:**

https://github.com/user-attachments/assets/601804fc-619d-466d-82e0-7a9ae8082c66


peterpeterparker:
> But, if the user logs out and then logs in again, the messages should be displayed again.

In the case of explicit logout (i.e., when the user clicks the 'Logout' button), or also in the case of automatic logout when the session expires?

If it's the former, I think we should stick with the current implementation and add a function to clear the local storage for this specific option when the user clicks 'Logout.' This functionality is already partially implemented in auth.service and would only require adding the clear action.

Side note: As far as I remember, the current implementation was designed this way because users can potentially have multiple identities and might find it annoying if this information always pops up when they switch accounts. Similarly, we also considered that on a single laptop, a user might not want to see this information repeatedly each time they sign up again, assuming they click 'Logout' to terminate their session. Of course, these were just assumptions.

BonomoAlessandro:
@peterpeterparker I just had a chat with @StefanBerger-DFINITY and he meant that it is ok that the messages are shown again if the session exires or the user changes his identity. If a user changes the identity some other tokens can be enabled and/or broken/ missing an Index canister. This means it would always makes sense to display the messages again if the identity has changed.

peterpeterparker:
Surprising, that's the kind of small thing that really annoys me as a user when I use an app but that's probably just me then.
Alright, let me take a look at the code in the PR.

src/frontend/src/lib/services/auth.services.ts
```diff
@@ -115,6 +115,11 @@ const clearTestnetsOption = async () => {
 	testnetsStore.reset({ key: 'testnets' });
 };
 
+// eslint-disable-next-line require-await
+const clearSessionStorage = async () => {
+	sessionStorage.clear();
```
peterpeterparker, line 120: Why are we clearing the session storage?

BonomoAlessandro: @peterpeterparker Because `sveltes` session storage is cleared when the page session ends. This only happens if the browser or the tab is closed. This means that we have to handle the session storage clear on a logout.

src/frontend/src/lib/services/auth.services.ts
```diff
@@ -128,7 +133,12 @@ const logout = async ({
 	busy.start();
 
 	if (clearStorages) {
-		await Promise.all([emptyIdbBtcAddressMainnet(), emptyIdbEthAddress(), clearTestnetsOption()]);
+		await Promise.all([
+			emptyIdbBtcAddressMainnet(),
+			emptyIdbEthAddress(),
+			clearTestnetsOption(),
+			clearSessionStorage()
```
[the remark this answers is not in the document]
BonomoAlessandro: @peterpeterparker you are right. I checked it with @StefanBerger-DFINITY again and it should clear the session storage. I will fix it.
thx

peterpeterparker (approved):
LGTM, thx

PR #3920, opened 2024-12-10
fix(frontend): Carousel glitch for last slide

AntonioVentilii:
# Motivation

When an user closes a slide in the dApp, the backend is updated and the user profile re-loaded. However it may happen that the two operations are concurrent. So we could have a non-updated user profile.

That causes basically to not really hide the slide that the user wants to close.

We avoid this glitch saving the closed dApps in temporary array that we use to filter, together with the user settings.

### Before

https://github.com/user-attachments/assets/0eeac073-e12c-4bf6-a392-85fc0818cc7f

### After

https://github.com/user-attachments/assets/807ecb73-6f00-4f39-98b3-84a1a00bb100


src/frontend/src/lib/components/dapps/DappsCarousel.svelte
```diff
@@ -30,7 +38,7 @@
 	}: CustomEvent<CarouselSlideOisyDappDescription['id']>) => {
 		const idx = dappsCarouselSlides.findIndex(({ id }) => id === dappId);
 
-		hiddenDappsIds = [...hiddenDappsIds, dappId];
+		temporaryHiddenDappsIds = [...temporaryHiddenDappsIds, dappId];
 
 		dappsCarouselSlides = filterCarouselDapps({ dAppDescriptions, hiddenDappsIds });
```
DenysKarmazynDFINITY, line 43: Do we still need to call it here? Because now it doesn't include the current `dappId`. I assume the `dappsCarouselSlides` value will be updated then because of line 32.

AntonioVentilii: We still do, unfort, to avoid that the reactive statement is not that quick, and the slide is removed and re-put...

And actually, now that you made me noticed it, I need to leave the update of `hiddenDappsIds` there...

DenysKarmazynDFINITY: Ok, I see, thanks for the explanation!

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3923, opened 2024-12-10
feat(build): add Solana feature flag

AntonioVentilii:
# Motivation

We start the integration of Solana creating a feature flag to enable or disabled the access to Solana network.


DenysKarmazynDFINITY (approved):
LGTM, thanks.

P.S.: if you think the flag will be with us for some time, let's also add it to `.env.example`.

AntonioVentilii:
@DenysKarmazynDFINITY I think it will not stay long to be honest, we are going at a good pace. But I will keep it in mind if we are delayed/stopped

PR #3924, opened 2024-12-10
feat(frontend): add Solana aliases for path

AntonioVentilii:
# Motivation

We include aliases for the new path `lib/sol`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3925, opened 2024-12-10
feat(frontend): add method getSchnorrPublicKey to signer interface

AntonioVentilii:
# Motivation

We add the method `schnorr_public_key` from the Signer canister to our interface.

# Note

Differently from the BTC and ETH methods, the Schnorr methods of the Signer take as input the `key_id`, that itself needs the root key name.
We would typically provide this as input to the Signer configurations, in the cases of BTC and ETH. However, in Schnorr case, we are hard-coding it in the codebase (documentation [here](https://internetcomputer.org/docs/current/developer-docs/smart-contracts/signatures/signing-messages-t-schnorr#signing-messages)).

Ideally, in one of the next version of the Signer, we will have the same way of settings the root key for the other methods too.

# Changes

- Add variables for root key, depending on the environment.
- Add method `schnorr_public_key` both in the signer canister and in the API.

# Tests

Included tests.


src/frontend/src/lib/canisters/signer.canister.ts
```diff
@@ -195,4 +197,27 @@ export class SignerCanister extends Canister<SignerService> {
 
 		throw mapSignerCanisterSendBtcError(response.Err);
 	};
+
+	getSchnorrPublicKey = async (derivationPath: string[]): Promise<Uint8Array | number[]> => {
```
bitdivine, line 201: This will also need to pass the algorithm.  Maybe move this function into the solana directory?  It  looks as if this is trying to set default values but those default values are specific to solana.  If you move this into the solana directory you don't need to pass the derivation path as an argument as you'll be using the same derivation path every time.  Until you have subaccounts that is!

src/frontend/src/env/signer.env.ts
```diff
@@ -0,0 +1,12 @@
+import type { SchnorrKeyId } from '$declarations/signer/signer.did';
+import { BETA, PROD, STAGING } from '$lib/constants/app.constants';
+
+// According to:
+// https://internetcomputer.org/docs/current/references/samples/rust/threshold-ecdsa/#deploying-the-canister-on-the-mainnet
+// https://internetcomputer.org/docs/current/developer-docs/smart-contracts/signatures/signing-messages-t-schnorr#signing-messages
+const SIGNER_ROOT_KEY_NAME = PROD || BETA ? 'key_1' : STAGING ? 'test_key_1' : 'dfx_test_key';
```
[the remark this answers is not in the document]
AntonioVentilii: the `STAGING` variable includes the `test.fe.*` envs

`export const STAGING = MODE === 'staging' || MODE.startsWith('test_fe_');`

at least, that is how we do in the rest of the code

src/frontend/src/env/signer.env.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const SCHNORR_KEY_ID: SchnorrKeyId = {
```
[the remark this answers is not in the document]
AntonioVentilii: Moved it to Solana specific folder and renamed

src/frontend/src/lib/canisters/signer.canister.ts
AntonioVentilii: Good one! Done!

AntonioVentilii:
@bitdivine ready for another review, I think

src/frontend/src/lib/api/signer.api.ts
```diff
@@ -86,6 +87,17 @@ export const sendBtc = async ({
...
+	identity,
+	...rest
+}: CanisterApiFunctionParams<{ derivationPath: string[]; keyId: SchnorrKeyId }>): Promise<
+	Uint8Array | number[]
+> => {
+	const { getSchnorrPublicKey } = await signerCanister({ identity });
+
+	return await getSchnorrPublicKey(rest);
+};
+
```
bitdivine, line 100: It would be nice to be able to test that this gives the same result as a command line call such as:
```
# Authorize payment
dfx canister call cycles_ledger icrc2_approve "
  record {
    amount = 10000000000000;
    spender = record {
      owner = principal \"${SIGNER_CANISTER_ID}\";
    };
  }"

# Get the public key of the web user
dfx canister call signer schnorr_public_key --with-cycles 1T --wallet $WALLET '(
  record {
    key_id = record { algorithm = variant { ed25519 }; name = "dfx_test_key" };
    canister_id = opt principal "the-web-principal";
    derivation_path = vec {
      blob "SOL";
      blob "\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00";
    };
  },
  opt variant { CallerPaysIcrc2Cycles },
)'
```

bitdivine: I'll deploy this branch locally and check.

src/frontend/src/sol/services/sol-balance.services.ts
```diff
@@ -0,0 +1,11 @@
+import { SOLANA_KEY_ID } from '$env/networks/networks.sol.env';
+import { getSchnorrPublicKey } from '$lib/api/signer.api';
+import type { CanisterApiFunctionParams } from '$lib/types/canister';
+
+export const getSolanaPublicKey = async (
+	params: CanisterApiFunctionParams<{ derivationPath: string[] }>
+): Promise<Uint8Array | number[]> =>
```
bitdivine, line 7: This should prepend "SOL" to the derivation path.

bitdivine (approved):

src/frontend/src/lib/api/signer.api.ts
bitdivine: Once this is available in the UI I can check.  The derivation path should be at least ["SOL" ] and probably something like ["SOL", "testnet/mainnet", subaccount_number], with the subaccount number set to `[0u8;32]`.

src/frontend/src/sol/services/sol-balance.services.ts
loki344: like this? https://github.com/dfinity/oisy-wallet/pull/3925/commits/6974e3852c09742e758aaf872637345781ca2f1c

PR #3930, opened 2024-12-11
fix(frontend): span cannot be a child of option

peterpeterparker:
# Motivation

It's probably a warning currently but, following becomes an error when migrating to Svelte v5 in #3929:

> `<span>` cannot be a child of `<option>`. `<option>` only allows these children: `<#text>`. The browser will 'repair' the HTML (by moving, removing, or inserting elements) which breaks Svelte's assumptions about the structure of your components.

# Changes

- Remove `span` in `option`.


AntonioVentilii (approved):
LGTM tks!

PR #3933, opened 2024-12-11
fix(frontend): self-closing HTML tags for non-void elements are ambiguous

peterpeterparker:
# Motivation

Following is a warning but becomes an error when upgrading the dependencies for Svelte v5 in https://github.com/dfinity/oisy-wallet/pull/3929:

> ...src/lib/components/ui/StaticSteps.svelte
  22:3  error  Self-closing HTML tags for non-void elements are ambiguous — use `<div ...></div>` rather than `<div ... />`
https://svelte.dev/e/element_invalid_self_closing_tag(element_invalid_self_closing_tag)  svelte/valid-compile

# Changes

- Update self closing iframe and div


AntonioVentilii (approved):
LGTM tks

PR #3936, opened 2024-12-11
refactor(frontend): split address services into networks

AntonioVentilii:
# Motivation

Since we are going to add more networks, it seems more manageable to split the address.services into smaller parts: one for each network and a collective one.

# Changes

There is no change in the implementation per se, apart from moving and exposing functions.

# Tests

Included tests for exposed functions.


src/frontend/src/btc/services/btc-address.services.ts
```diff
@@ -0,0 +1,121 @@
+// We use the Testnet address for Regtest.
```
DenysKarmazynDFINITY, line 1: This comment seems to be outdated, could you please remove it since you're doing changes here?

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/btc/services/btc-address.services.ts
AntonioVentilii: done

PR #3937, opened 2024-12-11
feat(frontend): listener templating

peterpeterparker:
# Motivation

Migration to Svelte v5 (#3929) involves replacing the ComponentType. While updating the Listener, a minor TypeScript issue was identified. We use a `svelte:component` that maps a token property, which may or may not be present, and can sometimes be set or unset. This approach works because the component is assigned to a state variable, but it is not fully TypeScript-compliant. To address this, we can leverage the template to correctly mount the listener.

### DIFF
--- a/src/frontend/src/lib/components/core/Listener.svelte
+++ b/src/frontend/src/lib/components/core/Listener.svelte
@@ -11,11 +11,19 @@
 </script>
 
 {#if isNullish(token) || $authNotSignedIn}
-	<NoListener />
+	<NoListener>
+		<slot />
+	</NoListener>
 {:else if isNetworkIdICP(token.network.id)}
-	<NoListener />
+	<NoListener>
+		<slot />
+	</NoListener>
 {:else if isNetworkIdBitcoin(token.network.id)}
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/core/Listener.svelte
```diff
@@ -9,18 +8,22 @@
...
-					: EthListener;
 </script>
 
-<svelte:component this={cmp} {token}>
-	<slot />
-</svelte:component>
+{#if isNullish(token) || $authNotSignedIn}
+	<NoListener>
+		<slot />
+	</NoListener>
```
AntonioVentilii, line 16: shall we delete `NoListener` at this point? and leave just `<slot />`

AntonioVentilii (approved):
LGTM tks!

Minor question, non blocker

src/frontend/src/lib/components/core/Listener.svelte
peterpeterparker: I was thinking doing that in another PR.

PR #3940, opened 2024-12-11
feat(frontend): remove no listener

peterpeterparker:
# Motivation

Follow-up of #3937. We do not need the `NoListener` anymore and can simplify with a slot.


AntonioVentilii (approved):
LGTM tks!

PR #3942, opened 2024-12-11
fix(frontend): add token to bitcoin listener

peterpeterparker:
# Motivation

It's currently not an issue but, while migrating to Svelte v5 #3929 I noticed that the `BitcoinListener` was missing the property `token`. This is required because in the listener utils function `mapListeners` we are mapping the component to a `TokenToListener` with `{ token, listener: BitcoinListener }` however, this `token` does not exist effectivelly. 

# Changes

- Add `export const token` unused to `BitcoinListener`


AntonioVentilii (approved):
LGTM tks

PR #3950, opened 2024-12-11

cgundy:
# Motivation

Follow-up from: https://github.com/dfinity/oisy-wallet/pull/3689

### DIFF
--- a/.github/workflows/update-snapshots.yml
+++ b/.github/workflows/update-snapshots.yml
@@ -28,7 +28,6 @@ jobs:
         with:
           repository: ${{ github.event.pull_request.head.repo.full_name }}
           ref: ${{ github.event.pull_request.head.ref }}
-          token: ${{ secrets.GIX_CREATE_PR_PAT }}
       
       - name: Build oisy-backend WASM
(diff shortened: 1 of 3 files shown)


.github/workflows/update-snapshots.yml
```diff
@@ -28,7 +32,6 @@ jobs:
         with:
           repository: ${{ github.event.pull_request.head.repo.full_name }}
           ref: ${{ github.event.pull_request.head.ref }}
-          token: ${{ secrets.GIX_CREATE_PR_PAT }}
```
cgundy, line 31: Better to keep minimal permissions as I don't think the PAT is required here. It seems like this action is just pushing a doctor image, not creating a PR, so the above permissions should be enough.

AntonioVentilii:
@cgundy hi! I don't know if useful, but I submitted this:

- PR https://github.com/dfinity/oisy-wallet/pull/3874

following up on your first PR. I even tagged you as reviewer if you want to have a look

cgundy:
Totally missed that! I'm closing this in favor of yours

PR #3954, opened 2024-12-11
ci(e2e): improve update snapshots task

inc-man:
# Motivation

The update snapshot workflow seems to be flaky and it cancels itself if one of the two runners is finished faster.

# Changes

added handling that both workflows finish and then the complete changes get commited.

# Tests

tested on branch


.github/workflows/update-snapshots.yml
```diff
@@ -91,31 +90,91 @@ jobs:
...
+
+      - name: Upload Snapshots Artifact
+        if: env.NO_CHANGES == 'false'
+        uses: actions/upload-artifact@v4
+        with:
+          name: snapshots-${{ matrix.os }}
+          path: e2e/**/*.spec.ts-snapshots/**/*-${{ steps.os_suffix.outputs.os_suffix }}.png
+          retention-days: 30
+
+  finalize_snapshots_update:
```
AntonioVentilii, line 121: I would not create another step, keep it all in the same step, so you don't need to upload/download artifacts, nor checkout again

We do thew same for formatting I think

inc-man: I feel like its necessary to have a new step for the finalizing task. Otherwise it would be the same as it is right now no?

With the new step we make sure that the two runs wait for eachother instead of instantly updating and commiting the updated snapshots

AntonioVentilii (approved):
LGTM tks!

.github/workflows/update-snapshots.yml
```diff
@@ -107,28 +119,79 @@ jobs:
...
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.ref }}
+
+      - name: Download Ubuntu Snapshots
+        uses: actions/download-artifact@v4
+        with:
+          name: snapshots-ubuntu-24.04
```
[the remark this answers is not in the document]
inc-man: since we are defining the os inside the same yaml, we don't really need to put it as variable.

.github/workflows/update-snapshots.yml
```diff
@@ -107,28 +119,89 @@ jobs:
...
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.ref }}
+
+      - name: Check if snapshots changed
+        run: |
+          if [ -n "$(git status --porcelain e2e)" ]; then
+            echo "NO_CHANGES=false" >> $GITHUB_ENV
+          else
+            echo "NO_CHANGES=true" >> $GITHUB_ENV
+          fi
```
AntonioVentilii, line 161: if you are using it here, they should be specified for each single environment, otherwise you could have false negatives, no?

imagine macOs finished and save it as true, then linux finishes but save it as false. In that case you would have a false negative

inc-man: agreed but in the end it wouldnt have worked in any way. So I did something new.

AntonioVentilii (approved):
LGTM tks

PR #3955, opened 2024-12-11
feat(frontend): include Solana in the tokens derived

AntonioVentilii:
# Motivation

We include the Solana tokens in the list of the derived tokens, pinning it among the native ones.

# Changes

- Create derived tokens list store for Solana.
- Add the new store to the current list of tokens.
- Include Solana mainnet token among the pinned ones.

# Tests

Created some tests. But here there is the result.

Screenshot 2024-12-11 at 20 53 52
Screenshot 2024-12-11 at 20 54 10
Screenshot 2024-12-11 at 20 54 18


src/frontend/src/tests/sol/derived/tokens.derived.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+describe('tokens.derived', () => {
+	describe('enabledSolanaTokens', () => {
+		beforeEach(() => {
+			vi.resetAllMocks();
+
+			testnetsStore.reset({ key: 'testnets' });
+
+			vi.spyOn(solEnv, 'SOLANA_NETWORK_ENABLED', 'get').mockImplementation(() => true);
+
+			vi.spyOn(appContants, 'LOCAL', 'get').mockImplementation(() => false);
```
DenysKarmazynDFINITY, line 22: With the PR I merged today, I think this line is no longer needed.

src/frontend/src/tests/sol/derived/tokens.derived.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+			testnetsStore.set({ key: 'testnets', value: { enabled: true } });
+
+			expect(get(enabledSolanaTokens)).toEqual([
+				SOLANA_TOKEN,
+				SOLANA_TESTNET_TOKEN,
+				SOLANA_DEVNET_TOKEN
+			]);
+		});
+
+		it('should return localnet token when in local env', () => {
```
DenysKarmazynDFINITY, line 44: Also, this test doesn't make sense anymore since we removed the `LOCAL` flag usage from `testnet` store. We can remove this test suite completely since the previous one already validates the needed behavior.

loki344: Hmm, I think they still make sense:
First case: If we are not local => return SOL, SOL testnet and SOL devnet token
Second case: If we are local => return SOL, SOL testnet, SOL devnet and SOL local token

DenysKarmazynDFINITY: Ah, true, didn't check the test properly

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/tests/sol/derived/tokens.derived.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+		});
+
+		it('should return emtpy array if feature flag false', () => {
+			vi.spyOn(solEnv, 'SOLANA_NETWORK_ENABLED', 'get').mockImplementation(() => false);
+			expect(get(enabledSolanaTokens)).toEqual([]);
+		});
+
+		it('should return testnet tokens when they are enabled', () => {
+			testnetsStore.set({ key: 'testnets', value: { enabled: true } });
+			vi.spyOn(appContants, 'LOCAL', 'get').mockImplementationOnce(() => false);
```
DenysKarmazynDFINITY, line 34: Before you merge it, could you please remove this line from here too? Didn't notice it initially

loki344: I can't this will cause this test to fail in a local run

PR #3958, opened 2024-12-12
feat(frontend): transaction ui with variant and templating

peterpeterparker:
# Motivation

Svelte v5 seems to be stricter in terms of typing, which causes the linter to fail in #3929 when the `TransactionUI` is used to create a `svelte:component`. After reviewing the issue, I also noticed that we are spreading the `transaction` within the `transactionUi` while mixing it with a `token`. From a typing perspective, I believe it would be safer to maintain a clear separation between these two pieces of information.

This PR addresses both issues.

# Changes

- Replace `svelte:component` by a direct usage of the component in the template
- Do not spread the transaction in the transaction UI
- Adapt types and consumers accordingly


AntonioVentilii (approved):
LGTM tks

PR #3960, opened 2024-12-12
feat(frontend): no Listener if no known network

AntonioVentilii:
# Motivation

We are integrating more networks. To be on the safe side and avoiding raising possible mismatches, we make the listener fallback to nothing, if the network is not known.


peterpeterparker (commented):
Should we add a test?

AntonioVentilii:
@peterpeterparker I included some tests, but only for the cases that render nothing. That is because we would need to pass some children to the rendered listeners to test it completely.

And either I do it with a mock component (like GIX components), or I wait for Svelte 5

WDYT?

peterpeterparker:
Thanks!!!

> That is because we would need to pass some children to the rendered listeners to test it completely.

What we do in NNS dapp or Gix cmp for those use case, we create a test Svelte component. That way we can assert such behavior. Like for example [Collapsible.spec.ts](https://github.com/dfinity/gix-components/blob/main/src/tests/lib/components/Collapsible.spec.ts) which performs test again `CollapsibleTest` which in turns implement `Collapsible`. Would that be an interesting pattern here?

That said I'm fine with a TODO until Svelte v5 is here. Can take few weeks but, will happen.

Your call.

AntonioVentilii:
@peterpeterparker Thanks! I actually have more examples where we could use that type of tests (test component). But if it is possible to test it with Svelte5, then I would prefer to wait (I would change it either way when we bump eheh)

Just to double confirm, my understanding is that, with Svelte 5, I can pass the children as a prop when rendering, right? Did you already encounter an example of this in the migrations you did?

peterpeterparker:
> Did you already encounter an example of this in the migrations you did?

The pattern I suggested does not need to be migrated and even with Svelte v5, you probably still gonna need a "Test component for the test" to create the snippet but, as I said, sure, I'm fine if we wait.

peterpeterparker (approved):
LGTM, thx

PR #3961, opened 2024-12-12
feat(frontend): add all token store

loki344:
# Motivation

- Consolidate the logic of all available tokens into a derived so that it is reusable
- This derived will later be used to determine if a token is available to the user

# Changes

Extract logic of merging a list of all tokens from ManageTokens.svelte into a separate derived

# Tests

Basic unit testing added.

comparison video (sepolia eth on the right screen was added manually, see screenshot below with fresh account)

https://github.com/user-attachments/assets/cb04f57f-b578-4569-b66b-2e6384ea75bd

image


DenysKarmazynDFINITY (approved):
The refactoring LGTM. Probably the biggest functional change I noticed is that `buildIcrcCustomTokens` is now called outside of `onMount`, but it should not affect anything, maybe it can be even considered as improvement - previously, it was called every time the modal gets open, and now only when the related data changes 🙂

PR #3963, opened 2024-12-12
feat(frontend): align testnets behaviour between envs

DenysKarmazynDFINITY:
# Motivation

Currently, we force `testnets` store to be always on locally. On one hand, it's kinda convenient to have it always enabled there, but on the other hand, it's a bit confusing that the related toggle has absolutely no effect locally. Since `LOCAL` is also `true` in the e2e environment, it also affects the tests and makes it impossible to validate the setting toggle functionality.

### DIFF
--- a/src/frontend/src/lib/derived/testnets.derived.ts
+++ b/src/frontend/src/lib/derived/testnets.derived.ts
@@ -1,8 +1,7 @@
-import { LOCAL } from '$lib/constants/app.constants';
 import { testnetsStore } from '$lib/stores/settings.store';
 import { derived, type Readable } from 'svelte/store';
 
 export const testnets: Readable<boolean> = derived(
 	[testnetsStore],
-	([$testnetsStore]) => $testnetsStore?.enabled ?? LOCAL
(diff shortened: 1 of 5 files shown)


peterpeterparker (commented):
Currently on `main`, when running `npm run test`, is `LOCAL` equals to true or false?
With this PR, does that change?
And if yes, do we want to globally preserve the behavior for `npm run test`?

peterpeterparker:
But in that case why did you had to modify few tests to add `testnetsStore.set({ key: 'testnets', value: { enabled: true } });`?

DenysKarmazynDFINITY:
Oh, you're right, for `npm run test` we keep `LOCAL` as `true` by setting a default value here - https://github.com/dfinity/oisy-wallet/blob/main/vite.utils.ts#L98. This is why I had to adjust some tests. Still, the behavior of `LOCAL` var in all envs stay untouched with this PR.

peterpeterparker:
Alright, thanks!

peterpeterparker (approved):
LGTM, thx

PR #3976, opened 2024-12-16
test(frontend): use .textContent instead of .innerHTML

peterpeterparker:
# Motivation

Svelte v5 #3929 wraps rendered elements with `<!----><!---->` around it. If we use `.innerHTML` for comparison purpose we would need to amend those comments or trim those, which isn't nice. Therefore, instead of it, we can use `.textContent` for comparison purposes.


DenysKarmazynDFINITY (approved):
Nice, thanks

PR #3939, opened 2024-12-11
style(frontend): black and white icons in network switcher

AntonioVentilii:
# Motivation

For all networks, we will use the black and white icon in the `NetworksSwitcher` component.

### Before

Screenshot 2024-12-11 at 06 18 12

### After

Screenshot 2024-12-11 at 06 14 39

PR #3977, opened 2024-12-16
test(frontend): handle new lines

peterpeterparker:
# Motivation

Using Svelte v5 #3939 in tests renders the spaces as defined in the component while in v4 we got just one.

```
- Expected
+ Received

- 0.22 BTC
+ 0.22
+ 		BTC
```

for

```
{formatToken({
			value: balance ?? ZERO,
			unitName: token.decimals,
			displayDecimals: token.decimals
})}
{token.symbol}
```

So we can split the expected results.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3978, opened 2024-12-16
test(frontend): mocks multiple times and restore

peterpeterparker:
# Motivation

In the Svelte v5 branch the tests aren't working because the dApps data are mocked only once, I assume Svelte v5 reads the data constant more than once, and also because the mocks are not restored once a test is over.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #3847, opened 2024-12-04
style(frontend): add ContentWithToolbar to AddTokenByNetwork.svelte

loki344:
# Motivation

Unfiform modal layout, use contentWithToolbar in AddTokenByNetwork. This MR prepares the future styling of the toolbar as a footer of the modal.

# Changes

- use ContentWithToolbar

# Tests

direct comparison

https://github.com/user-attachments/assets/70d13194-de63-4245-bd54-206a3db1d068

PR #3987, opened 2024-12-16
fix(frontend): fix add token modal form

loki344:
# Motivation

https://github.com/dfinity/oisy-wallet/pull/3847 introduced a bug because the buttons in AddTokenByNetwork were moved outside the <form> tag into the new modal toolbar. The form can no longer be submitted.

# Changes

- Use the same pattern that is already used for the back button also for the next button

# Tests

before

https://github.com/user-attachments/assets/67938117-2e46-4456-a751-289d11902eb2

after, eth token

https://github.com/user-attachments/assets/22b0fbd1-7e1c-4d9a-a4fe-520ef489a09b

after ic token (ignore the error)

https://github.com/user-attachments/assets/b8cf93df-0292-477d-87c7-7ab8b34b9d22


peterpeterparker (commented):
Can't we address the bug by retaining the use of the form and avoiding event bubbling?

- All current usages of `ButtonNext` rely on the `form` element. This PR would create an inconsistency.
- Semantically, using a form is appropriate since it contains input fields, and this aligns with standard HTML practices.

peterpeterparker:
Note:  A new E2E test suite for this feature will be added in PR #3993, therefore no specific tests were included in this PR.

loki344:
> Can't we address the bug by retaining the use of the form and avoiding event bubbling?
> 
> * All current usages of `ButtonNext` rely on the `form` element. This PR would create an inconsistency.
> * Semantically, using a form is appropriate since it contains input fields, and this aligns with standard HTML practices.

Thanks for asking this! Yes, I think it's possible - I don't know why I did not thought about this. Reverted my changes and simply changed the structure of the DOM so that the form still contains the buttons. I tested it again:

https://github.com/user-attachments/assets/42199bf3-25d4-4b93-9888-1f14fedca4fc

peterpeterparker (approved):
LGTM, thx

PR #3988, opened 2024-12-16
feat(frontend): imperative transaction network guard

peterpeterparker:
# Motivation

With Svelte v5 #3949 we are not able to access the `/transactions` page because the `$page` store is not initalized when the route is mounted. This happens because we guard the page for case where the `network` query param would not be set. To resolve the issue we can imperatively access the query parameter.

# Changes

- Use `URLSearchParams` instead of `$page` to read the `network` param on mount.


e2e/guards-page.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+		// We go to transactions without network
+		await page.goto(`${TRANSACTIONS_URL}/?token=Internet%20Computer`);
+
+		// We should be redirected to the home screen.
+		const homepageLoggedIn = new HomepageLoggedIn({ page, iiPage });
+
+		await homepageLoggedIn.waitForContentReady();
+
+		await expect(page).toHaveScreenshot({ fullPage: true });
```
DenysKarmazynDFINITY, line 23: I think when we have more test cases like this one, we can create an additional method that validates which page is active without capturing a screenshot.

DenysKarmazynDFINITY (approved):
LGTM, thanks

peterpeterparker:
Require #3989 to pass - i.e. CI will fail until related PR is merged.

PR #3990, opened 2024-12-16
feat(frontend): adds vip qr code item to the menu

BonomoAlessandro:
# Motivation

Vip users should be able to see the `VIP QR codes` item into the menu.

# Changes

- adds item to menu

# Tests

**before:**
image

**after:**
image


src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -42,6 +48,17 @@
 
 	let fromRoute: NavigationTarget | null;
 
+	let isVip = false;
+	onMount(async () => {
```
AntonioVentilii, line 52: will this run everytime we open the `Menu`? should it be maybe a state somewhere?

src/frontend/src/lib/i18n/en.json
```diff
@@ -41,7 +41,8 @@
 			"source_code": "Source code",
 			"changelog": "Changelog",
 			"submit_ticket": "Submit a ticket",
-			"confirm_navigate": "Are you sure you want to navigate away?"
+			"confirm_navigate": "Are you sure you want to navigate away?",
```
AntonioVentilii, line 44: "navigate away" sounds like a bad translation honestly eheh! What is the idea? Like a closing question? "Are you sure you want to close this page?"

BonomoAlessandro: @AntonioVentilii I haven't changed this text in this PR. I just added the `,` to add the `vip_qr_code`

src/frontend/src/lib/components/core/Menu.svelte
BonomoAlessandro: hmm this might be a good idea

src/frontend/src/lib/i18n/en.json
BonomoAlessandro: @AntonioVentilii I haven't changed this text in this PR. I just added the `,` to add the `vip_qr_code`

src/frontend/src/lib/components/core/Menu.svelte
BonomoAlessandro: hmm it might be a good idea to but it into a state instead of calling every time

BonomoAlessandro: @AntonioVentilii ok nevermind. I don't think that it is necessary to store the state somewhere. The Menu item is only loaded one time and the vip data are loaded in the `onMount` function.

src/frontend/src/lib/i18n/en.json
AntonioVentilii: ah! right! tks

AntonioVentilii (approved):
LGTM

PR #3991, opened 2024-12-16
feat(frontend): adds vip qr code modal

BonomoAlessandro:
# Motivation

Vip users should be able to create reward codes and get qr codes to show them to other clients.

# Changes

- adds vip reward modal

# Tests

**qr code modal:**
image


src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+			<span class="mb-4 block w-full pt-3 text-center text-sm text-tertiary">
+				{replacePlaceholders($i18n.vip.invitation.text.regenerate_countdown_text, {
+					$counter: counter.toString()
+				})}
+			</span>
+		{:else}
+			<span class="w-full"><SkeletonText /></span>
+		{/if}
+
+		<ButtonGroup styleClass="flex-col sm:flex-row" slot="toolbar">
```
AntonioVentilii, line 94: not really sure about this: we don't do it in any of the modals, nor in any usage of the Buttons... if we are going to change it, better to do it in another PR for all the `ButtonGroup` components, for consistency

src/frontend/src/tests/lib/components/qr/VipQrCodeModal.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+	const mockAuthStore = (value: Identity | null = mockIdentity) =>
+		vi.spyOn(authStore, 'authIdentity', 'get').mockImplementation(() => readable(value));
+
+	const mockedNewRewardResponse: NewVipRewardResponse = {
+		VipReward: {
+			code: '1234567890'
+		}
+	};
+
+	it('renders the vip qr code modal items', async () => {
```
AntonioVentilii, line 26:
```suggestion
	it('should render the vip qr code modal items', async () => {
```

src/frontend/src/tests/lib/components/qr/VipQrCodeModal.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+			expect(qrCodeURL).toBeInTheDocument();
+			expect(qrCodeURL?.textContent?.includes(mockedNewRewardResponse.VipReward.code));
+
+			expect(copyButton).toBeInTheDocument();
+
+			expect(regenerateButton).toBeInTheDocument();
+		});
+	});
+
+	it('regenerates reward code', async () => {
```
AntonioVentilii, line 59:
```suggestion
	it('should regenerate reward code', async () => {
```

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
+	import ContentWithToolbar from '$lib/components/ui/ContentWithToolbar.svelte';
+	import SkeletonText from '$lib/components/ui/SkeletonText.svelte';
+	import { authIdentity } from '$lib/derived/auth.derived';
+	import { getNewReward } from '$lib/services/reward-code.services';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+
+	const secondsToRegenerate = 45;
```
peterpeterparker, line 18: Should we moved this as a global `constants`?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+	let counter = secondsToRegenerate;
+	let countdown: NodeJS.Timeout;
```
peterpeterparker, line 20: `| undefined`?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+
+	let code: string;
+	const generateCode = async () => {
+		if (nonNullish($authIdentity)) {
```
peterpeterparker, line 24: I think we have a guard that signout the user if not authenticated, maybe better to use this here?

```
			await nullishSignOut();
			return;
		}
```

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+			const vipReward = await getNewReward($authIdentity);
```
peterpeterparker, line 25: You gonna resolve the `TODO` within the service to return a result afterwards?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+			if (nonNullish(vipReward)) {
```
peterpeterparker, line 26: We continue to poll if generating a reward fails?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+	const intervalFunction = async () => {
+		counter--;
+
+		if (counter === 0) {
+			await regenerateCode();
+		}
+	};
+
+	onMount(() => {
+		countdown = setInterval(intervalFunction, 1000);
```
peterpeterparker, line 48: shouldn't `countdown` be cleared on destroy?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+	});
+
+	generateCode();
```
peterpeterparker, line 51: Why does this finds place at the root of the module? Shouldn't it start when the component is mounted?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+		clearInterval(countdown);
+		await generateCode();
+		counter = secondsToRegenerate;
+		countdown = setInterval(intervalFunction, 1000);
+	};
+
+	const intervalFunction = async () => {
+		counter--;
+
+		if (counter === 0) {
```
peterpeterparker, line 42: UI wise we display 0 for the few seconds that regenerating a code least?

```
3 -> displayed for 1 second
2 -> displayed for 1 second
1 -> displayed for 1 second
0 -> displayed for few seconds
```

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+	};
+
+	onMount(() => {
+		countdown = setInterval(intervalFunction, 1000);
+	});
+
+	generateCode();
+
+	let qrCodeUrl;
+	$: qrCodeUrl = `${window.location.origin}/?code=${code}`;
```
peterpeterparker, line 54: the `string` generated for the `code` procudes always a valid URL?

peterpeterparker (commented):
Is it expected to have no overall limit on polling? For example, a VIP user could open this particular screen, forget the tab, and continuously generate codes every 45 seconds for the entire duration of the session.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+<Modal on:nnsClose={modalStore.close}>
+	<svelte:fragment slot="title"
+		><span class="text-xl">{$i18n.vip.invitation.text.title}</span>
+	</svelte:fragment>
+
+	<ContentWithToolbar>
+		<div class="mx-auto mb-4 aspect-square h-80 max-h-[44vh] max-w-[100%] p-4">
+			{#if nonNullish(code)}
+				<QRCode value={qrCodeUrl}>
+					<svelte:fragment slot="logo">
```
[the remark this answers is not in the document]
BonomoAlessandro: @AntonioVentilii The QrCode element provides a slot for the logo and we want to use this slot.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,108 @@
...
+			<div class="flex items-center justify-between gap-4 rounded-lg bg-brand-subtle px-3 py-2">
+				<output>{qrCodeUrl}</output>
+				<ReceiveCopy
+					address={qrCodeUrl}
+					copyAriaLabel={$i18n.vip.invitation.text.invitation_link_copied}
+					testId="vip-qr-code-copy-button"
+				/>
+			</div>
+
+			<span class="mb-4 block w-full pt-3 text-center text-sm text-tertiary">
```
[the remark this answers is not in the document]
BonomoAlessandro: No, it's not default

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: @AntonioVentilii What exactly do you mean with 

>  for all the ButtonGroup components, for consistency

I mean like this we just enable to add some custom styling to the element

BonomoAlessandro:
> Is it expected to have no overall limit on polling? For example, a VIP user could open this particular screen, forget the tab, and continuously generate codes every 45 seconds for the entire duration of the session.

Good input. I will change it that way that it only generates new codes as long as the tab is active.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: @peterpeterparker Yes i will resolve the `TODO` in one of the next PRs

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: I guess it is ok to try it every 45 sec even if it failed before.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: Thats a good point, thx

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: hmm yes this should be moved into the `onMount` function

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
peterpeterparker: And how is a code generated without a valid identity? The backend support generating codes for anonymous call?

peterpeterparker: Also isn't this guard - i.e. an edge case - in the sense that a none authenticated user should never reaches this point?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: it depends on how long it takes to generate a new code. If the generation of the code is done instantly, the timer is also reseted instantly. Or what do you mean?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
peterpeterparker: Should retries continue indefinitely, or should there be a maximum limit? I'm asking because update calls incur costs.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: Yes, the qrCodeUrl is only displayed on the UI if a code exists and in this case the URL is always valid

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
peterpeterparker: It won't be instantly. In best case 1 sec I guess but, I don't know what the backend does exactly. If the subnet is busy, I guess it can least few seconds until a new code is generated.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: In case of a not authenticated user tries to generate a code nothing will happen. That's why i have this check `if (nonNullish($authIdentity)) {`

Yes exactly. It should not be possible for a none authenticated user to reach this modal.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: I mean in the end it is only trying to generate codes as long as the vip user is on this modal. So i guess in this case it is conscious decision of the vip user

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: Would this be a bad behaviour? 
In my opinion the counter should only be reseted if there is a new code. Do you think it would be a better user experience if the counter resets and then maybe take some more seconds until the URL is updated?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
peterpeterparker: > nothing will happen.

Not exactly if I understand the implementation correct an error would be thrown by the backend and a toast would be displayed.

For consistency and above reason I think using the guard would work best.

```
   await nullishSignOut();
   return;
}
```

If you rather avoid the "nullishSignout", as a nitpick, I would suggest to not wrap the all code in the if for readability but return early in the function.

BonomoAlessandro: I do not fully understand.
Yes if a none authenticated user tries to generate a code an error toast is displayed but in the `VipQrCodeModal` component i do have following implementation: 
```
const generateCode = async () => {
		if (nonNullish($authIdentity)) {
			const vipReward = await getNewReward($authIdentity);
			if (nonNullish(vipReward)) {
				code = vipReward.code;
			}
		}
	};
```
And in this implementation I do explicitly check if the user is authenticated.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
peterpeterparker: But of course, if the team is fine with this, feel free to ignore my feedback.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
peterpeterparker: Not sure, spontaneously I was thinking following:

```
3 -> displayed for 1 second
2 -> displayed for 1 second
1 -> displayed for 1 second
"Generating new code" with a pulse animation
3
2
1
"Generating new code" with a pulse animation
etc.
```

i.e. not displaying "0" but something else like a text with animation that way it does not hand on "0". Would that make sense?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: ok in this case it would be better to be consistent

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: this sounds like a good idea

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
AntonioVentilii: indeed, but why only for this modal? why not for all?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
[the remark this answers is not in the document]
AntonioVentilii: indeed, my question is: why the slot is not directly in the `div`, but in a `svelte:fragment`?

`<div slot="logo" class="flex items-center justify-center rounded-lg bg-white p-2">`

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: It's defined like this in the design. The buttons should swap the alignment if the page gets smaller. To achieve this on this modal we need something like this.
Other modals do not need to swap the alignment

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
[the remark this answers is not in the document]
BonomoAlessandro: ahaa yes, that's a fair point

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: @peterpeterparker I implemented it now. I guess you are right, it's better to have it.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,137 @@
...
+	let retries = 0;
+
+	let code: string;
+	const generateCode = async () => {
+		if (isNullish($authIdentity)) {
+			await nullishSignOut();
+			return;
+		}
+
+		if (retries !== maxRetries) {
```
peterpeterparker, line 36: To avoid an edge case issue, `<` maybe? Otherwise it might retries if the number of retries is bigger than the max retries.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,137 @@
...
+			const vipReward = await getNewReward($authIdentity);
+			if (nonNullish(vipReward)) {
+				code = vipReward.code;
+			} else {
+				retries++;
```
peterpeterparker, line 41: Shouldn't this counter be incremented when the call fails as well?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,137 @@
...
+	let code: string;
+	const generateCode = async () => {
+		if (isNullish($authIdentity)) {
+			await nullishSignOut();
+			return;
+		}
+
+		if (retries !== maxRetries) {
+			const vipReward = await getNewReward($authIdentity);
+			if (nonNullish(vipReward)) {
```
peterpeterparker, line 38: nitpick: we generally avoid wrapping code within if statements. Instead, we return early to improve readability. If you want to align with the code style of the codebase, you might consider updating your code accordingly.

```
const generateCode = async () => {
		await nullishSignOut();
		return;
	}

	if (retries !== maxRetries) {
		const vipReward = await getNewReward($authIdentity);
		if (nonNullish(vipReward)) {
			code = vipReward.code;
		} else {
			retries++;
		}
	}
};
```

can be adapted as following:

```
const generateCode = async () => {
		await nullishSignOut();
		return;
	}
	
	if (retries >= maxRetries) {
		return;
	}

	const vipReward = await getNewReward($authIdentity);

	if (nonNullish(vipReward)) {
		code = vipReward.code;
	} else {
		retries++;
	}
};
```

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,137 @@
...
+	import { nullishSignOut } from '$lib/services/auth.services';
+	import { getNewReward } from '$lib/services/reward-code.services';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+
+	let counter = VIP_CODE_REGENERATE_INTERVAL;
+	let countdown: NodeJS.Timeout | undefined;
+	const maxRetries = 3;
+	let retries = 0;
```
peterpeterparker, line 27: For better readability, could you either add a comment indicating that this is for retries in case of errors or rename the variable to make it more explicit?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,137 @@
...
+				code = vipReward.code;
+			} else {
+				retries++;
+			}
+		}
+	};
+
+	const regenerateCode = async () => {
+		clearInterval(countdown);
+		await generateCode();
```
peterpeterparker, line 48: houldn't the retries be asserted here? With the current implementation, we might indeed prevent any further calls, but the countdown would still proceed. Not a strong opinion, just a thought.

src/frontend/src/lib/constants/app.constants.ts
```diff
@@ -109,3 +109,6 @@ export const ZERO = BigNumber.from(0n);
 // Wallets
 export const WALLET_TIMER_INTERVAL_MILLIS = (SECONDS_IN_MINUTE / 2) * 1000; // 30 seconds in milliseconds
 export const WALLET_PAGINATION = 10n;
+
+// VIP
+export const VIP_CODE_REGENERATE_INTERVAL = 45;
```
peterpeterparker, line 114: For readability I would either add a comment about the fact that those are seconds or rename to `VIP_CODE_REGENERATE_INTERVAL_IN_SECONDS`.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,137 @@
...
+			clearInterval(countdown);
+		} else {
+			countdown = setInterval(intervalFunction, 1000);
+		}
+	};
+
+	onMount(() => {
+		generateCode();
+		countdown = setInterval(intervalFunction, 1000);
+		document.addEventListener('visibilitychange', onVisibilityChange);
```
peterpeterparker, line 72: `<svelte:window />` would work as well instead of handling the event listener manually?

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,137 @@
...
+	const onVisibilityChange = () => {
+		if (document.hidden) {
+			clearInterval(countdown);
+		} else {
+			countdown = setInterval(intervalFunction, 1000);
+		}
+	};
+
+	onMount(() => {
+		generateCode();
```
peterpeterparker, line 70: To avoid code duplication, could we call `regenerateCode` on mount? Would that work?

```
onMount(regenerateCode);
```

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: @peterpeterparker sounds good to me

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: Yes, this would also work

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: Ohh yes, this would be a cleaner way to solve this

peterpeterparker (approved):
LGTM, thanks

Given that I stricly only reviewed the `script` tag of `VipQrCodeModal.svelte`, please also wait for @AntonioVentilii approval to merge this PR.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -0,0 +1,134 @@
...
+<Modal on:nnsClose={modalStore.close}>
+	<svelte:fragment slot="title"
+		><span class="text-xl">{$i18n.vip.invitation.text.title}</span>
+	</svelte:fragment>
+
+	<ContentWithToolbar>
+		<div class="mx-auto mb-4 aspect-square h-80 max-h-[44vh] max-w-full p-4">
+			{#if nonNullish(code)}
+				<QRCode value={qrCodeUrl}>
+					<div slot="logo" class="flex items-center justify-center rounded-lg bg-white p-2">
```
AntonioVentilii, line 90: shouldn't `slot="logo"` go in the parent `div` two levels above?

```typescript
<div slot="logo" class="mx-auto mb-4 aspect-square h-80 max-h-[44vh] max-w-full p-4">
  {#if nonNullish(code)}
    <QRCode value={qrCodeUrl}>
	    <div class="flex items-center justify-center rounded-lg bg-white p-2">
```

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
AntonioVentilii: still not convinced about this... may you please double-check with design? furthermore, will the "Close" button go up or down when it is vertical?


Screenshot 2025-01-09 at 14 45 45

BonomoAlessandro: @AntonioVentilii On movile, the `Close` button should move to the top. In the attached picture there is the `Share this link` button but this feature will be implemented later and not in this PR.

image

BonomoAlessandro: @AntonioVentilii @artkorotkikh-dfinity just wrote me that he updated the designs.

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: @AntonioVentilii No it is a part of the QRCode

AntonioVentilii (approved):
LGTM tks!!

PR #3993, opened 2024-12-17
test(e2e): manually import tokens test

inc-man:
# Motivation

We realized that there was an issue with manually importing tokens. So for reference we want to test this feature so we get notified if its broken or working.

# Changes

created new e2e test
added new test constants
set test-id on elements


AntonioVentilii:
~16-month-old e2e test against an older version of the manual-import-tokens flow. Closing — please reopen with a fresh test if still needed.

PR #3994, opened 2024-12-17
fix(e2e): correct btc testnet usage in e2e test

inc-man:
# Motivation

Since we changed the handling of the btc testnet, the current tests fail because of faulty configuration. Now the test should first activate the testnet setting, then run all the tests and snapshots.

# Changes

commented out faulty code, added todo


DenysKarmazynDFINITY:
@inc-man it seems something went wrong with `e2e/explorer-page.spec.ts-snapshots/should-display-explorer-page-1-Google-Chrome-linux.png` - explorer page got replaced with the token page.

peterpeterparker (commented):
Generally looks good expect the `should-display-explorer-page-1-Google-Chrome-linux.png` which does not display the explorer anymore. Not sure if the issue is related to this PR or not? If not we can fix that in another PR, let me know.

peterpeterparker (approved):
LGTM, thx

As discussed offline, let's see if the new snapshots resolve the issue with the "explorer". There are few PRs open. Otherwise let's tackle this separatly.

PR #3996, opened 2024-12-17
feat(frontend): implements reward canister api

BonomoAlessandro:
# Motivation

We want to use the `reward canister api` and the `reward service` so that we can send some rewards to new or existing users.

# Changes

- implements reward canister api
- implements reward service


peterpeterparker:
I randomly noticed this PR. Not sure if it's a general work in progress or effective implementation of a feature so just a heads-up: canister, API, and services should be provided in separate PRs, each covered by tests.

BonomoAlessandro:
@peterpeterparker atm it's just work in progress. But good to know, if it is finished, i will split them into different PRs

BonomoAlessandro:
@peterpeterparker now this PR is open and is ready to be reviewed

src/frontend/src/lib/api/reward.api.ts
```diff
@@ -0,0 +1,58 @@
...
+	certified
+}: CanisterApiFunctionParams<QueryParams>): Promise<UserData> => {
+	const { getUserInfo } = await rewardCanister({ identity });
+
+	return getUserInfo({ certified });
+};
+
+export const getNewVipReward = async ({
+	identity
+}: CanisterApiFunctionParams): Promise<NewVipRewardResponse> => {
```
peterpeterparker, line 26: Should it be `CanisterApiFunctionParams<never>`?

src/frontend/src/lib/services/reward-code.services.ts
```diff
@@ -0,0 +1,63 @@
...
+	claimVipReward as claimVipRewardApi,
+	getNewVipReward as getNewVipRewardApi,
+	getUserInfo as getUserInfoApi
+} from '$lib/api/reward.api';
+import { i18n } from '$lib/stores/i18n.store';
+import type { Identity } from '@dfinity/agent';
+import { fromNullable } from '@dfinity/utils';
+import { get } from 'svelte/store';
+
+export const isVipUser = async ({
```
peterpeterparker, line 12: Are you planning to handle potential errors from the services on the UI side?

It makes sense to provide all the APIs, but should we perhaps provide the service wrappers only when necessary? I assume we’ll first review, for example, the usage of `isVipUser` before implementing `claimVipReward`, etc. I’m concerned about adding dead code or code that might require several iterations to refine, you know what I mean?

src/frontend/src/lib/services/reward-code.services.ts
```diff
@@ -0,0 +1,63 @@
...
+	identity,
+	certified
+}: {
+	identity: Identity;
+	certified: boolean;
```
peterpeterparker, line 17: Thinking at it now, what's the idea here, it's the UI that decides if it's a certified call or not? You are not planing to use `queryAndUpdate` - which in that case would probably better implemented within the service - or the security does not matter here and were are find with a query?

src/frontend/src/lib/api/reward.api.ts
BonomoAlessandro: @peterpeterparker no it should be `CanisterApiFunctionParams` like it is.

peterpeterparker: Why?

src/frontend/src/lib/services/reward-code.services.ts
BonomoAlessandro: @peterpeterparker I guess it would be better to handle the errors in the service instead of the UI.

I am not sure what exactly you mean with this. The feature we are planning to roll out contains all this functionality.

src/frontend/src/lib/api/reward.api.ts
peterpeterparker: > because never is a special type and implies that there is no value.

Yes it's what I mean, there is no value no?

```
export type CanisterApiFunctionParams<T = unknown> = T & {
	nullishIdentityErrorMessage?: string;
	identity: OptionIdentity;
	canisterId?: CanisterIdText;
};
```

in this case - `getNewVipReward` - there is no additional value, no addition params, no `T`.

src/frontend/src/lib/services/reward-code.services.ts
peterpeterparker: > I guess it would be better to handle the errors in the service instead of the UI.

Great. You can have a look at error handling in other existing services and implement a similar pattern.

src/frontend/src/lib/api/reward.api.ts
BonomoAlessandro: There are values. I do call the `getNewVipReward` function with the params `identity` and `nullishIdentityErrorMessage`. Thats why it cannot be `CanisterApiFunctionParams<never>`.

src/frontend/src/lib/services/reward-code.services.ts
BonomoAlessandro: As you mentioned, i will implement it in the service instead of the UI that makes more sense.
I think for this case we will always just use a query instead of an update

src/frontend/src/lib/api/reward.api.ts
peterpeterparker: `never` apply to the additional parameters not `identity`. Anyway, not that important, we can move on.

peterpeterparker (approved):
LGTM, thx

Reviewed offline together with @BonomoAlessandro. Few minor improvements will be provided within this PR after this approval as discussed to move forward.

PR #4008, opened 2024-12-17
feat(frontend): new year assets

peterpeterparker:
# Motivation

We want to update the welcome banner and OISY wallet logo for the Xmas season.

# Notes

I couldn't find a way to avoid using these assets for E2E testing purposes, and it feels like doing so would add unnecessary noise to the code.  
Similarly, I don't want to update the consuming code—only the assets (which are already shipped in the JS).  
For this reason, I am creating this PR instead of following up on #4007.

# Changes

- Updated assets as provided by @artkorotkikh-dfinity in PR #4007.

# Tests

No tests, solely assets changes. Snapshots will be recreated afterwards in a separate PR.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4010, opened 2024-12-17
feat(frontend): revert new year assets

peterpeterparker:
# Motivation

For new year we updated the assets. Once the period over we will have to revert this change with the previous, standard assets.

# Notes

E2E snapshots need to be updated once the change reverted.

# Changes

- Revert PR #4008


DenysKarmazynDFINITY (approved):
Thanks!

PR #4016, opened 2024-12-18
feat(frontend): implements reward canister

BonomoAlessandro:
# Motivation

We want to use the reward canister so that we can send some rewards to new or existing users.

# Changes

- implements reward canister


src/frontend/src/lib/canisters/reward.canister.ts
```diff
@@ -0,0 +1,50 @@
...
+			},
+			idlFactory: idlFactoryReward,
+			certifiedIdlFactory: idlCertifiedFactoryReward
+		});
+
+		return new RewardCanister(canisterId, service, certifiedService);
+	}
+
+	getUserInfo = (): Promise<UserData> => {
+		const { user_info } = this.caller({ certified: true });
```
peterpeterparker, line 34: Is this correct, `getUserInfo` will always be requested with an update call?

src/frontend/src/lib/canisters/reward.canister.ts
```diff
@@ -0,0 +1,50 @@
...
+
+		return user_info();
+	};
+
+	getNewVipReward = (): Promise<NewVipRewardResponse> => {
+		const { new_vip_reward } = this.caller({ certified: true });
```
peterpeterparker, line 40: ditto

src/frontend/src/tests/lib/canisters/reward.canister.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+		RewardCanister.create({
+			canisterId: Principal.fromText('tdxud-2yaaa-aaaad-aadiq-cai'),
+			identity: mockIdentity,
+			certifiedServiceOverride: serviceOverride,
+			serviceOverride
+		});
+	const service = mock<ActorSubclass<RewardService>>();
+	const mockResponseError = new Error('Test response error');
+
+	it('returns true if user is vip', async () => {
```
peterpeterparker, line 35: should we update the description of the test to be more generic or should we add more tests?

When I read `returns true if user is vip` I kind of expect other test, like "user is not a vip", but since we mock the results not sure what's best.

src/frontend/src/tests/lib/canisters/reward.canister.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+vi.mock(import('$lib/actors/agents.ic'), async (importOriginal) => {
+	const actual = await importOriginal();
+	return {
+		...actual,
+		// eslint-disable-next-line require-await
+		getAgent: async () => mock<HttpAgent>()
+	};
+});
+
+describe('reward.canister', () => {
```
peterpeterparker, line 22: can you group each function with a `describe` ?

```
describe('reward.canister', () => {
    describe('getUserInf', () => {
        it(....
        
    describe('getNew...
        etc
```

src/frontend/src/tests/lib/canisters/reward.canister.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+			sprinkles: []
+		};
+		service.user_info.mockResolvedValue(mockedUserData);
+
+		const { getUserInfo } = await createRewardCanister({
+			serviceOverride: service
+		});
+
+		const userData = await getUserInfo();
+		expect(userData.is_vip).toBeTruthy();
```
peterpeterparker, line 48: Isn't this test incorrect? `is_vip` is an array containing a boolean not a boolean no?

src/frontend/src/tests/lib/canisters/reward.canister.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+	it('should be possible to claim a vip reward', async () => {
+		const mockedClaimResponse = { Success: null };
+		service.claim_vip_reward.mockResolvedValue(mockedClaimResponse);
+
+		const { claimVipReward } = await createRewardCanister({
+			serviceOverride: service
+		});
+
+		const claimResponse = await claimVipReward({ code: '1234567890' });
+		expect(claimResponse).toEqual(mockedClaimResponse);
```
peterpeterparker, line 104: None of your positive tests assert that the call is executed with the expected parameters, can you add those `toHaveBeenCalledWith`?

src/frontend/src/lib/canisters/reward.canister.ts
BonomoAlessandro: @peterpeterparker I guess I do not understand exactly what you mean by this.

src/frontend/src/tests/lib/canisters/reward.canister.spec.ts
BonomoAlessandro: hmm yes. we should also provide a test for the "user is not a vip" case.

src/frontend/src/tests/lib/canisters/reward.canister.spec.ts
BonomoAlessandro: No it is not. I don't know why the test runs successful...

src/frontend/src/tests/lib/canisters/reward.canister.spec.ts
BonomoAlessandro: I do provide tests like this but not in this PR. I already implemented them but the are coming in the next PR together with the implementation of the `reward-code.service`

src/frontend/src/lib/canisters/reward.canister.ts
peterpeterparker: Are you planning to always call those getters using update calls (certified), or will you also use query calls (not certified)?

src/frontend/src/tests/lib/canisters/reward.canister.spec.ts
peterpeterparker: JavaScript weirdness. `if ([]) console.log('this is printed')`

src/frontend/src/tests/lib/canisters/reward.canister.spec.ts
peterpeterparker: It would be nice to have those included in this PR already. Those parameters matter at this level. Plus, it's consistent with other test suite - e.g. `backend.canister.spec` or `signer.canister.spec`.

src/frontend/src/lib/canisters/reward.canister.ts
BonomoAlessandro: To be honest this was not a conscious decision. How can i find out what the best solution would be?

peterpeterparker: Take a look at the DID file to determine if it's technically feasible and supported (e.g., `['query']` at the end of a function's declaration). If it is supported, consider whether you need to display the information to the end user as quickly as possible. If the answer is yes, you likely want to support it. Does that help?

src/frontend/src/lib/canisters/reward.canister.ts
peterpeterparker: According the comment https://github.com/dfinity/oisy-wallet/pull/4016#discussion_r1890344193 I understand it's expected that this call is strictly certified.

src/frontend/src/tests/lib/canisters/reward.canister.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+			service.user_info.mockResolvedValue(mockedUserData);
+
+			const { getUserInfo } = await createRewardCanister({
+				serviceOverride: service
+			});
+
+			const userData = await getUserInfo(queryParams);
+			expect(service.user_info).toHaveBeenCalledWith();
+			expect(userData.is_vip.length).toBe(1);
+			expect(userData.is_vip[0]!).toBeTruthy();
```
peterpeterparker, line 54: Please avoid the use of `!` 

cc we need an eslint rule to prevent this bad practice @AntonioVentilii , thought we already had one

peterpeterparker: Unrelated, you can also use the `fromNullable` utility to decode variant.

Your assertion can look like following for example:

```
expect(fromNullable(userData.is_vip) === true).toBeTruthy();

// or

const isVip = romNullable(userData.is_vip);
expect(isVip ?? false).toBeTruthy();
```

peterpeterparker (approved):
LGTM, thx

PR #4017, opened 2024-12-18
feat(frontend): add successful reward modal

BonomoAlessandro:
# Motivation

# Changes

# Tests


src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -177,6 +198,13 @@
 	</div>
 {/if}
 
+{#if $modalSuccessfulRewardModal}
+	<SuccessfulRewardModal />
+{/if}
+{#if $modalFailedRewardModal}
```
AntonioVentilii, line 204: for concurrency, I would use `else if`

src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -51,13 +57,28 @@
...
 		// A small delay for display animation purpose.
 		setTimeout(() => loading.set(false), 1000);
+
+		if (!$loading && $page.url.searchParams.has('code') && nonNullish($authIdentity)) {
+			const rewardCode = $page.url.searchParams.get('code');
+			if (nonNullish(rewardCode)) {
+				const result = await claimVipReward({ identity: $authIdentity, code: rewardCode });
+
+				$page.url.searchParams.delete('code');
+				window.history.pushState({}, '', $page.url);
```
AntonioVentilii, line 74: may you please move the navigation logic in a separate util in `nav.utils` and provide a few tests?

AntonioVentilii (commented):
In general, I would not put this logic in `Loader` component: its logic should be just to manage the initial loading part

Maybe a separate component?

AntonioVentilii (approved):
LGTM tks

src/frontend/src/tests/lib/utils/nav.utils.spec.ts
```diff
@@ -138,6 +139,21 @@ describe('nav.utils', () => {
...
+			const pushStateMock = vi.spyOn(appNavigation, 'pushState').mockImplementation(vi.fn());
+			const url = new URL(`https://example.com/`);
+			const searchParams = new URLSearchParams({
+				code: '123'
+			});
+			url.search = searchParams.toString();
+
+			removeSearchParam({ url, searchParam: 'code' });
+
+			expect(pushStateMock).toHaveBeenCalledWith(url, {});
```
AntonioVentilii, line 153: here maybe a check that the URL that is left is the expected one?

PR #4018, opened 2024-12-18
feat(frontend): implements failed review modal

BonomoAlessandro:
# Motivation

If a user wants to get some rewards and it fails we want to display a modal with an error message.

# Changes

- implements reward failure modal

# Tests

image


src/frontend/src/lib/assets/failed-reward.svg
AntonioVentilii, line 1: may you please add the source for this one?

src/frontend/src/lib/i18n/en.json
```diff
@@ -745,6 +745,16 @@
 			}
 		}
 	},
+	"vip": {
+		"reward": {
+			"text": {
+				"open_wallet": "Take me to the wallet",
```
AntonioVentilii, line 751: Why not simply a "Close" button?

src/frontend/src/lib/assets/failed-reward.svg
BonomoAlessandro: @AntonioVentilii There is not a specific reason why this text is used but it looks better than "close". Also it is possible that a user just registered to the application and never saw the wallet before.

src/frontend/src/lib/i18n/en.json
BonomoAlessandro: @AntonioVentilii There is not a specific reason why this text is used but it looks better than "close". Also it is possible that a user just registered to the application and never saw the wallet before.

AntonioVentilii: thanks for checking

AntonioVentilii (approved):
LGTM tks

PR #4019, opened 2024-12-18
test(e2e): improve quality of e2e snapshots

inc-man:
# Motivation

The current snapshots seem to be a bit flaky, so to improve that, we adjusted the snapshot config to make sure we get the best possible screenshots.

# Changes

adjusted toHaveScreenshot configuration
updated snapshots

# Tests

ran locally without any issues


e2e/utils/pages/homepage.page.ts
```diff
@@ -243,6 +243,13 @@ abstract class Homepage {
 		await this.clickByTestId(NAVIGATION_ITEM_TOKENS);
 	}
 
+	async takeScreenshot(): Promise<void> {
+		await expect(this.#page).toHaveScreenshot({
+			fullPage: true,
+			timeout: 5 * 60 * 1000
```
peterpeterparker, line 249: It's a pretty long timeout, we need it to be that long?

playwright.config.ts
```diff
@@ -22,6 +22,12 @@ const TIMEOUT = 5 * 60 * 1000;
 export default defineConfig({
 	timeout: TIMEOUT,
 	workers: DEV ? 5 : 2,
+	expect: {
+		toHaveScreenshot: {
```
peterpeterparker, line 26: Can you add some references for those options or some comments about what are their effects?

playwright.config.ts
```diff
@@ -38,9 +44,10 @@ export default defineConfig({
 		...(DEV && { headless: false })
 	},
 	projects: [
+		/* Test against desktop browsers. */
 		{
 			name: 'Google Chrome',
-			use: { ...devices['Desktop Chrome'], channel: 'chrome' }
+			use: { ...devices['Desktop Chrome'] }
```
peterpeterparker, line 50: Is this change related to this PR?

e2e/utils/pages/homepage.page.ts
inc-man: Currently our standard timeout for tests is set on 5 Minutes. That is because we have certain tests that do actually need 2-3 minutes sometimes. 

This timeout here only tells playwright that it has 5 minutes to try and get the correct screenshot. So if the first one fails, it will try again. I can lower it but does not hurt at all.

playwright.config.ts
inc-man: Will do that in my next commit.

playwright.config.ts
inc-man: Eh, I changed this some time ago, and I am pretty sure it "improves" the quality of the screenshots. So when I realized it got changed back I reverted it to what I did before.

e2e/utils/pages/homepage.page.ts
peterpeterparker: If it's standardized, should we use a constant or some constants instead of "hardcoding" the value here then?

playwright.config.ts
peterpeterparker: You mean in this PR right?

playwright.config.ts
peterpeterparker: Does it improves the quality because we get a more recent or older chrome by using another channel? And if so, should we pin the version of chrome we are actually using?

e2e/utils/pages/homepage.page.ts
inc-man: Well my initial idea was setting this option in the playwright config aswell. Somehow that is not supported, so the only other option I see is hardcoding it here. Basically its already reduced usage from mentioning it in every test file to now having it only on homepage.

peterpeterparker: Should we create an environment variable for the timeout that way we can use the same value in both configuration and code? (obviously if yes in another PR)

playwright.config.ts
inc-man: By defining a "channel" for a browser you make sure that you use a specific version, installed on the device that runs the test. By not using a "branded browser" we use playwrights device library and open source browsers. With that we remove the potential danger of having browser issues because not everything is installed on the github runner.

Its not a complete wow effect but I feel like if we are using playwright and they offer a wide range of devices, why not just use what they have to offer.

peterpeterparker: The ci will be a bit slower I guess because it will have to fetch Chrome. Alright, not sure how it's related but, indeed if it improves the screenshots, all good. Thanks for the explanation.

playwright.config.ts
inc-man: comments added for the new configs

e2e/utils/pages/homepage.page.ts
inc-man: I feel like a constant would be good enough for this case. But then again I wonder if its not a bit too much just for a timeout. Lets say in a few weeks we want to change the timeout in the config. Then we would need to create two constants for both cases or adjust the one in the test too.

peterpeterparker: Works for me. Let’s see how it evolves. I just fear that we might end up with multiple hardcoded timeouts, which could contradict each other at some point if we don’t pay attention.

peterpeterparker (approved):
LGTM, thx

PR #4020, opened 2024-12-18
feat(frontend): implements successful reward modal

BonomoAlessandro:
# Motivation

If a user wants to get some rewards and it is successful we want to display a modal with a congratulations message.

# Changes

- implements reward successful modal

# Tests

image


src/frontend/src/lib/assets/successful-reward.svg
AntonioVentilii, line 1: PLease add source to this SVG

src/frontend/src/lib/i18n/en.json
```diff
@@ -745,6 +745,16 @@
 			}
 		}
 	},
+	"vip": {
+		"reward": {
+			"text": {
+				"open_wallet": "Take me to the wallet",
```
AntonioVentilii, line 751: Why not a simply "Close" button?

BonomoAlessandro: @AntonioVentilii There is not a specific reason why this text is used but it looks better than "close". Also it is possible that a user just registered to the application and never saw the wallet before.

AntonioVentilii: tks for checking

AntonioVentilii (approved):
LGTM tks!

PR #4021, opened 2024-12-18
feat(frontend): add loader for solana balances

loki344:
# Motivation

Implement solana balance service.

# Changes

- implement solana balance service

# Tests

Unit testing added for solana balance service

PR #4022, opened 2024-12-18
feat(frontend): add sol loader

loki344:
# Motivation

load sol balances in application

# Changes

use balance service in component to load sol balances, consolidate existing eth loader in common component.

# Tests

see unit test for solLoader

additionally: 
currently receiving a 403 (rate limit on mainnet)

https://github.com/user-attachments/assets/3aa88395-3b32-44ff-afeb-b58016cfc911


loki344:
https://github.com/dfinity/oisy-wallet/pull/4021 has to be merged first

loki344:
For SOL, we follow the same approach as for ic and btc, loading the data via a worker/scheduler approach. No Loader needed

PR #4023, opened 2024-12-18
feat(cyclesledger): Generate Rust types

bitdivine:
# Motivation
We would like to auto-generate the bindings used when calling other canisters. We are doing this a step at a time, starting with the cycles ledger, which is an easy choice as it doesn't change much.

# Changes
* Generate types for the cycles ledger client.
  * The types have been removed from the manually generated and modified client.

# Tests
The binding command is run as part of CI, so we will see if the bindings are out of date.


src/cycles_ledger/client/src/lib.rs
bitdivine, line 1: Types are moved from here to the new file below.

src/cycles_ledger/client/Cargo.toml
bitdivine, line 1: The cycles ledger client now gets its types from the types crate, so we have to add that crate as a dependency.

scripts/generate.sh
bitdivine, line 1: This is the command run by `npm run generate` and by CI when CI checks whether the bindings are up to date.  Here we generate the Rust bindings as well.

peterpeterparker:
> with the cycles ledger,

There is a [crate](https://crates.io/crates/icrc-ledger-types) for ICRC types, just saying.

bitdivine:
@peterpeterparker I'm aware.  It is limited though, whereas this machinery works for all canisters and provides things such as pocket-ic bindings, that the ICRC crate doesn't provide.

DenysKarmazynDFINITY (approved):
LGTM

P.S.: thanks for the comments, they are really helpful 🙂

PR #4026, opened 2024-12-19
feat(frontend): handpick update Sns tokens

peterpeterparker:
# Motivation

We cannot merge automatic PR #4025 because the CTS case has still not being handled. This PR handpick the changes that have to be integrated.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4029, opened 2024-12-19
feat(frontend): add sol to receive addresses

loki344:
# Motivation

Add solana to receive addresses if feature is enabled.

# Changes

Integrate SOL addresses into receiveAddresses

# Tests

Uploading Bildschirmaufnahme 2024-12-19 um 11.52.43.mov…


src/frontend/src/lib/constants/test-ids.constants.ts
```diff
@@ -24,6 +24,10 @@ export const RECEIVE_TOKENS_MODAL_BTC_MAINNET_SECTION = 'receive-tokens-modal-bt
 export const RECEIVE_TOKENS_MODAL_BTC_TESTNET_SECTION = 'receive-tokens-modal-btc-testnet-section';
 export const RECEIVE_TOKENS_MODAL_BTC_REGTEST_SECTION = 'receive-tokens-modal-btc-regtest-section';
 export const RECEIVE_TOKENS_MODAL_ETH_SECTION = 'receive-tokens-modal-eth-section';
+export const RECEIVE_TOKENS_MODAL_SOL_MAINNET_SECTION = 'receive-tokens-modal-sol-mainnet-section';
```
DenysKarmazynDFINITY, line 27: Since you already added these vars, can you please add them to e2e/receive-tokens-modal.spec.ts? Otherwise, when we enable sol in the e2e env, the receive modal test will be failing due to different sol addresses in the modal on every run.

loki344: sure, do the e2e test have the LOCAL flag to true? I added this one too

DenysKarmazynDFINITY: Yes, LOCAL is true in the e2e env since we build and deploy the app the same way as we do it locally.

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/lib/constants/test-ids.constants.ts
loki344: I will have to comment them out until the SOLANA_NETWORK_ENABLED flag is set to true, otherwise e2e will fail

PR #4033, opened 2024-12-19
feat(frontend): adjust SOL typing

loki344:
# Motivation

adjust sol typing to conform with continous implementation.

# Changes

- separate rpc urls from token network
- introduce mapping function to map from networkId to network
- improve enum typing with zod


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4035, opened 2024-12-19
feat(frontend): add sol worker types

loki344:
#Note

review first:  https://github.com/dfinity/oisy-wallet/pull/4033

# Motivation

We add types for sol wallet workers

# Changes

- add types
- add messages

PR #4036, opened 2024-12-19
feat(frontend): add sol listener

loki344:
# Note

review first: https://github.com/dfinity/oisy-wallet/pull/4035

# Motivation

Implement sol listener for balances. Transactions will follow.

# Changes

Implement sol listener similar to btc listener.

# Tests

unit tests provided


src/frontend/src/sol/services/sol-listener.services.ts
```diff
@@ -0,0 +1,61 @@
...
+			data: {
+				data: BigNumber.from(balance),
+				certified
+			}
+		});
+	} else {
+		balancesStore.reset(tokenId);
+	}
+
+	// Transaction handling will be added later
```
DenysKarmazynDFINITY, line 35: Something to consider - after sol wallet worker supports transactions, maybe we can combine the btc and sol listeners into one single service as they seem to be completely identical. Transaction stores and error text from the function below can be passed as params. It obviously only one of many places where networks things can be combined, but in this case it seems to be very low hanging fruit which can spare 150 lines of code for us 🙂

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/sol/services/sol-listener.services.ts
loki344: Yes, totally agree. Also it is planned to migrate eth to the worker approach, which then could benefit from this shared implementation.

peterpeterparker: Note that having it separated has advantages too, notably a clear separation of concerns and the assurance that if one network fails, it does not impact the other—this without the need to handle the error.

PR #4037, opened 2024-12-19
feat(frontend): add sol wallet worker

loki344:
# Motivation

We integrate SOL with the same worker approach as btc and ic. 

# Changes

Implement worker and scheduler.

# Tests

Unit tests provided, same as for btc worker.

Additionally:

https://github.com/user-attachments/assets/7eca0ee7-dc69-4bd5-8597-c2aafc1b6c33

# Note
During the development there were some error messages for mainnet (even though we use alchemy now).

image

image

Not reproducable at the moment. Will address this with @StefanBerger-DFINITY to check if the alchemy dashboard shows something.


src/frontend/src/sol/schedulers/sol-wallet.scheduler.ts
```diff
@@ -0,0 +1,115 @@
...
+			job: this.syncWallet,
+			data
+		});
+	}
+
+	private loadBalance = async (
+		params: LoadSolWalletParams
+	): Promise<CertifiedData<bigint | null>> => ({
+		data: await loadSolLamportsBalance({ network: params.solanaNetwork, address: params.address }),
+		certified: true
```
DenysKarmazynDFINITY, line 57: The balance you're fetching here (from 3rd party API) is not certified. Same as with Infura and ETH.

loki344: Hm ok, what does certified mean? By default the SDK uses commitment level 'finalized', so the transaction is definite. https://solana.com/docs/rpc#default-commitment

src/frontend/src/sol/schedulers/sol-wallet.scheduler.ts
```diff
@@ -0,0 +1,115 @@
...
+		try {
+			const balance = await this.loadBalance({
+				address: data.address.data,
+				solanaNetwork: data.solanaNetwork
+			});
+
+			//todo implement loading transactions
+
+			this.syncWalletData({ response: { balance } });
+		} catch (error: unknown) {
```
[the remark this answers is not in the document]
loki344: Hmm, I think without this catch there will be no trigger to broadcast the 'syncSolWalletError'

src/frontend/src/sol/schedulers/sol-wallet.scheduler.ts
```diff
@@ -0,0 +1,115 @@
...
+			this.postMessageWalletError({ error });
+		}
+	};
+
+	private syncWalletData = ({ response: { balance } }: { response: SolWalletData }) => {
+		const newBalance =
+			isNullish(this.store.balance) ||
+			this.store.balance.data !== balance.data ||
+			(!this.store.balance.certified && balance.certified);
```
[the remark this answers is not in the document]
loki344: Hmm, yeah sounds reasonable - The btc-wallet.scheduler.ts has the same logic ATM. I'll adjust it here and add a todo there for another PR

src/frontend/src/sol/schedulers/sol-wallet.scheduler.ts
```diff
@@ -0,0 +1,115 @@
...
+
+		this.store = {
+			...this.store,
+			balance
+		};
+
+		this.postMessageWallet({
+			wallet: {
+				balance,
+				newTransactions: ''
```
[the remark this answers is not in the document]
loki344: 1) Let's keep them - They're not hurting atm and will follow shortly. They are already defined in the schema and I don't want to go back and forth.
2) I don't think so, the certified flag is included in the balance and set to false above.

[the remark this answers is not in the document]
peterpeterparker: 1. It's hurting I think because the message are parsed with JSON on the UI side and `''` is not a valid value for `JSON.parse('')`

2. Coolio

src/frontend/src/sol/schedulers/sol-wallet.scheduler.ts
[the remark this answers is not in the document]
peterpeterparker: Thanks for double checking.

src/frontend/src/sol/schedulers/sol-wallet.scheduler.ts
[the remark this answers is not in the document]
loki344: removed txns, note that this will touch an additional file

loki344:
> I did not review the two `...Wallets.svelte` components. Can you please provide those in a separate PR with their tests.

opened https://github.com/dfinity/oisy-wallet/pull/4080

src/frontend/src/sol/schedulers/sol-wallet.scheduler.ts
```diff
@@ -0,0 +1,116 @@
...
+	private syncWallet = async ({ data }: SchedulerJobData<PostMessageDataRequestSol>) => {
+		assertNonNullish(data, 'No data provided to get Solana balance.');
+
+		try {
+			const balance = await this.loadBalance({
+				address: data.address.data,
+				solanaNetwork: data.solanaNetwork
+			});
+
+			//todo implement loading transactions
```
peterpeterparker, line 70: nitpick: while we do not have an eslint rule for it, maybe something we can setup (?), we generally use `// TODO:`. Maybe easier to retrieve if you use the same pattern?

src/frontend/src/sol/services/worker.sol-wallet.services.ts
```diff
@@ -0,0 +1,95 @@
...
+					tokenId,
+					data: data.data as SolPostMessageDataResponseWallet
+				});
+				return;
+
+			case 'syncSolWalletError':
+				syncWalletError({
+					tokenId,
+					error: (data.data as PostMessageDataResponseError).error,
+					hideToast: isTestnetNetwork || isDevnetNetwork || isLocalNetwork
```
peterpeterparker, line 53: Can you add a `TODO` to review this at some point. As far as I remember neither @DenysKarmazynDFINITY nor myself were super happy with this pattern `hideToast`. Right Denys?

src/frontend/src/tests/sol/schedulers/sol-wallet.scheduler.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+import * as solanaApi from '$sol/api/solana.api';
+import { SolWalletScheduler } from '$sol/schedulers/sol-wallet.scheduler';
+import { SolanaNetworks } from '$sol/types/network';
+import { mockIdentity } from '$tests/mocks/identity.mock';
+import { type MockInstance } from 'vitest';
+
+describe('sol-wallet.scheduler', () => {
+	let spyLoadBalance: MockInstance;
+
+	const mockBalance = 100n;
```
peterpeterparker, line 13: Unrelated to this review, I just asked myself if the Solana balance was effectively a `bigint`, which it is. Then I had a look at `loadSolLamportsBalance`, which contains a comment about lamports.

Long story short, if I get it right, what do you think of creating a type `type Lamports = bigint`? It might be handy to retrieve those kinds of values instead of searching for `bigint`. Since we often have a type for values, maybe it makes sense in terms of consistency too.

Again, unrelated to this PR.

peterpeterparker (approved):
LGTM, thx

src/frontend/src/sol/services/worker.sol-wallet.services.ts
DenysKarmazynDFINITY: Yep, we wanted to revisit the approach at some point (e.g. by avoiding requests (or launching wallets in this case) that we know in advance are not gonna succeed, rather than making a call + failing silently).

src/frontend/src/tests/sol/schedulers/sol-wallet.scheduler.spec.ts
loki344: Yeah I can follow up with a PR for this.

PR #4039, opened 2024-12-19
feat(frontend): add types for spl tokens

loki344:
# Motivation

SPL Tokens are what ERC20 is in Ethereum. Ultimately the goal is to not only handle SOL native tokens, but also SPL tokens. To handle them in the application, we add typing as we did for ERC20 Tokens.

https://spl.solana.com/#:~:text=The%20Solana%20Program%20Library%20(SPL,and%20deployed%20to%20its%20mainnet.

# Changes

- Add types for spl and user tokens

### DIFF
--- a/src/frontend/src/sol/types/spl.ts
+++ b/src/frontend/src/sol/types/spl.ts
@@ -1,0 +1,8 @@
+import type { ContractAddress } from '$eth/types/address';
+import type { Exchange } from '$lib/types/exchange';
+import type { Token } from '$lib/types/token';
+
+export type SplToken = SplContract & Token;
+
+export type SplContractAddress = ContractAddress;
+export type SplContract = SplContractAddress & { exchange: Exchange };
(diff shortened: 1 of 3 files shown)


src/frontend/src/sol/types/spl.ts
```diff
@@ -0,0 +1,8 @@
+import type { ContractAddress } from '$eth/types/address';
+import type { Exchange } from '$lib/types/exchange';
+import type { Token } from '$lib/types/token';
+
+export type SplToken = SplContract & Token;
+
+export type SplContractAddress = ContractAddress;
+export type SplContract = SplContractAddress & { exchange: Exchange };
```
loki344, line 8: Does someone know if we need exchange here?

peterpeterparker (commented):
Are you aware that adding support for tokens that users can toggle or add manually requires extending the backend? I would recommend starting with the backend extension before implementing the UI, this is generally our usual approach.

loki344:
Nevermind, I'll just convert it to draft - the feature for Usertokens will follow, until then I'll render a disabled toggle for solana without adding this logic

PR #4040, opened 2024-12-19
feat(frontend): make solana native tokens requiredtokens

loki344:
# Motivation

To align with the other implementations, let's make Solana native tokens of type RequiredToken.

# Changes

adjust type


AntonioVentilii (approved):
LGTM tks

PR #4042, opened 2024-12-19
feat(frontend): add solana to exchangeCoinId check

loki344:
# Motivation

TBH I just stumbled across this. I think it makes sense to add solana there for consistency - I'm unsure if it's necessary.

# Changes

- add solana to exchangeCoinId check in exchange.derived.ts


AntonioVentilii (approved):
Tks! Good catch

PR #4057, opened 2024-12-23
build(frontend): bump ic-js

peterpeterparker:
# Motivation

`ledger-icrc@next` cannot solely be bumped in #4055 because the ic-js dependencies were released and we did not upgraded those. So this PR upgrade ic-js libraries to their most recent version.

It also set `ic-management` as dev dependency since we do not use it at runtime.


AntonioVentilii (approved):

peterpeterparker:
Since we just released ic-js, I bumped the libs again as we did not merged this PR yet. Please re-review @AntonioVentilii .

AntonioVentilii (approved):

PR #4058, opened 2024-12-23
chore(frontend): clean unnecessary extension in imports

AntonioVentilii:
# Motivation

A small cleanup of unnecessary `.js` extensions in some imports.


peterpeterparker (approved):
Webstorm auto-import 💙😉

(looks like they changed something in their last version of their IDE, that's why I assume it's related)

PR #4061, opened 2024-12-23
feat(frontend): use utility isEmptyString

peterpeterparker:
# Motivation

There is a new utility `isEmptyString` in `@dfinity/utils` which can be use instead of `!notEmptyString`. So this PR replace the one occurence of such code we have in our code base.

# Changes

- Replace function usage


AntonioVentilii (approved):
Nice!

We have `isNullishOrEmpty` util too, can it be replaced with this one?

peterpeterparker:
> We have isNullishOrEmpty util too, can it be replaced with this one?

Maybe but, we need to clean thing?

ic-js:

- notEmptyString -> boolean
- isEmptyString -> boolean

oisy:

- isNullishOrEmpty -> value is undefined | null

PR #4077, opened 2025-01-04
fix(dependabot): Continue even if a token cannot be generated.

bitdivine:
# Motivation
Dependabot PRs now all fail.  A typical example:

Screenshot from 2025-01-04 06-37-22

On investigation, this is because these workflows now have functionality to add commits to the PR, but this requires getting an authorization token from GitHub secrets, which Dependabot PRs are not allowed to do.  These workflows do check whether the token was null, and if so disable the committing feature, but workflow never gets that far because the step that tries to get the token fails and kills the run.

It looks as if allowing that step to fail (with [`continue-on-error`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error)) should fix this, but it is hard to be sure without letting Dependabot run.

# Changes
- Allow failure.

# Tests


bitdivine:
Merged as part of #4048. (Commit copied to that PR.  That PR was blocked by this problem.)

PR #4087, opened 2025-01-06
feat(frontend): review reward-code services "code style"

peterpeterparker:
# Motivation

I reviewed the `reward-code.services` after merging #3996 and noticed a few areas that could be better aligned with the codebase's style and patterns for consistency.

# Changes for Consistency

- Add JSDoc comments for exported functions.  
- Use specific error classes instead of the generic `Error` when the error type is known.  
- Bubble the `err` in the `ResultSuccess` return to potentially inform the consumer about which error occurred.  
- Align the pattern of `updateVipReward` with the pattern of other internal functions in the service, `updateReward`, which throws errors instead of returning values.  
- Define `err` as `err: unknown`.

# Opinionated Changes

- Destructure parameters.  

# TODO

- Add a TODO for `getNewReward` to use a `ResultSuccess` return type as well.


src/frontend/src/tests/lib/services/reward-code.services.spec.ts
```diff
@@ -7,14 +7,19 @@ import * as rewardApi from '$lib/api/reward.api';
 import { claimVipReward, getNewReward, isVipUser } from '$lib/services/reward-code.services';
 import { i18n } from '$lib/stores/i18n.store';
 import * as toastsStore from '$lib/stores/toasts.store';
+import { AlreadyClaimedError, InvalidCodeError } from '$lib/types/errors';
 import en from '$tests/mocks/i18n.mock';
 import { mockIdentity } from '$tests/mocks/identity.mock';
 import { get } from 'svelte/store';
-import { vi } from 'vitest';
+import { expect, vi } from 'vitest';
```
AntonioVentilii, line 14: should we already remove these imports? I think the IDE is auto-importing them

AntonioVentilii (approved):
LGTM tks!!!

A minor comment:

src/frontend/src/tests/lib/services/reward-code.services.spec.ts
peterpeterparker: Yes, absolutely, thanks!

Here's a cleanup PR for similar imports 👉 #4091

src/frontend/src/lib/services/reward-code.services.ts
```diff
@@ -50,22 +60,34 @@ const updateReward = async (identity: Identity): Promise<VipReward> => {
...
-	} catch (err) {
+	} catch (err: unknown) {
 		const { vip } = get(i18n);
 		toastsError({
 			msg: { text: vip.reward.error.loading_reward },
 			err
 		});
 	}
+
+	// TODO: should return a ResultSuccess
```
BonomoAlessandro, line 90: Why should a ResultSuccess object be returned?

BonomoAlessandro (approved):

PR #4092, opened 2025-01-07
feat(frontend): use lamports type for SOL instead of bigint

loki344:
# Motivation

As pointed out in https://github.com/dfinity/oisy-wallet/pull/4037 it matches the style of the repo to use a separate type for Lamports. The Solana rpc already offers such types, so it makes sense to use it from there.

# Changes

Use Typing from solana RPC npm package.


src/frontend/src/sol/schedulers/sol-wallet.scheduler.ts
```diff
@@ -10,18 +10,19 @@ import { loadSolLamportsBalance } from '$sol/api/solana.api';
...
+import type { Lamports } from '@solana/rpc-types';
 
 interface LoadSolWalletParams {
 	solanaNetwork: SolanaNetworkType;
 	address: SolAddress;
 }
 
 interface SolWalletStore {
-	balance: CertifiedData<bigint | null> | undefined;
+	balance: CertifiedData<Lamports | null> | undefined;
```
peterpeterparker, line 21: Unrelated but, here you can use `Option<Lamports>`

src/frontend/src/tests/sol/schedulers/sol-wallet.scheduler.spec.ts
```diff
@@ -5,12 +5,13 @@ import * as solanaApi from '$sol/api/solana.api';
...
 import { SolanaNetworks } from '$sol/types/network';
 import { mockIdentity } from '$tests/mocks/identity.mock';
+import { lamports, type Lamports } from '@solana/rpc-types';
 import { type MockInstance } from 'vitest';
 
 describe('sol-wallet.scheduler', () => {
 	let spyLoadBalance: MockInstance;
 
-	const mockBalance = 100n;
+	const mockBalance: Lamports = lamports(100n);
```
peterpeterparker, line 14:
```
// in this test
const mockBalance: Lamports = lamports(100n);

// vs in the other test of this PR
const mockBalance = lamports(1000n);
```

Do we need `: Lamports` or it can be omitted?

peterpeterparker (approved):
LGTM, thx

Two minor comments.

src/frontend/src/sol/schedulers/sol-wallet.scheduler.ts
loki344: Hmm, it won't be exactly the same but I don't see a problem with it

peterpeterparker: Ah sorry, I read it wrong, though it was `Lamports | null | undefined`. Nvm my feedback here.

PR #4094, opened 2025-01-07
fix(frontend): remove reference to svelte in sol-rpc.providers.ts

loki344:
# Motivation

Via sol-wallet.worker.ts there was a reference to svelte. This broke the build via sol-wallet.worker.ts as sveltekit is not available there (see vite config).

# Changes

Remove assertion as it's unnecessary anyway, it's already secured via TS that there cannot be passed another network.

# Tests

image


AntonioVentilii (approved):

AntonioVentilii:
Thank you!

PR #4095, opened 2025-01-07
fix(frontend): treat nullish amount in IcTransaction and BtcTransaction

AntonioVentilii:
# Motivation

The function `BigNumber.from` could be receiving a nullish value from `amount`, and lead to an unexpected error.

To fix it, we move the function directly in the `amount` definition.


loki344:
Would it make sense to align BtcTransaction and EthTransaction also accordingly?

AntonioVentilii:
@loki344 good point! `EthTransaction` is good, but I adjusted `BtcTransaction` too (and changed the title)

loki344 (approved):
LGTM, thx

PR #4097, opened 2025-01-07
feat(frontend): add Solana WebSocket URL for subscritption

AntonioVentilii:
# Motivation

To sing and send Solana transaction we will need to use subscription RPC (documentation [here](https://github.com/anza-xyz/solana-web3.js/blob/8111c88f3b98503ce445bac9c761f33d584fc5de/examples/transfer-lamports/src/example.ts#L73)). So, we add the respective URL to each environment.

Note: since the subscription is mostly for simple query calls to verify the status of a transaction during the sending process, we are using the public Solana RPC for it. We can always change to a private one in the future if needed.

# Changes

- Install [`@solana/rpc-subscriptions`](https://www.npmjs.com/package/@solana/rpc-subscriptions).
- Add websocket URL to the schema of RPCs.
- Add URLs to the map of RPCs.
- Create function to extract from the map.


loki344 (approved):
LGTM, thanks

PR #4098, opened 2025-01-07
refactor(frontend): explicit type in solanaHttpRpc

AntonioVentilii:
# Motivation

We put the explicit type for the util `solanaHttpRpc` and we decompose the logic.


loki344 (approved):
LGTM, thanks

PR #4099, opened 2025-01-07
feat(frontend): add SolSendAmount component

AntonioVentilii:
# Motivation

Like `EthSendAmount` and `BtcSendAmount`, we create `SolSendAmount` to be used in the send form for Solana.

Note: we still miss the max button and the check for fees.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4104, opened 2025-01-07
fix(frontend): adds dfx network config

BonomoAlessandro:
# Motivation

In a previous PR [(link)](https://github.com/dfinity/oisy-wallet/pull/4122) we replaced an error toast with an error log for stages running in MODE `local`. Since the e2e tests do not run in MODE `local` at the moment, this needs to be changed.

# Changes

- updates `MODE` for e2e tests


AntonioVentilii (approved):
LGTM tks

PR #4107, opened 2025-01-07
refactor(frontend): use SendForm in IcSendForm

AntonioVentilii:
# Motivation

To start standardizing the Send forms, we use the component `SendForm` inside the `IcSendForm`, with a few adjustments.


DenysKarmazynDFINITY (approved):
Nice, thanks.

It's probably worth to quickly test the updated flow on staging, as in the past we had a few issues popping up after similar refactoring.

PR #4116, opened 2025-01-08
feat(devops): Use DFX_NETWORK as the default ENV

bitdivine:
# Motivation
Commands such as this in the hacking file:
```
ENV=ic dfx deploy frontend --network ic --wallet yit3i-lyaaa-aaaan-qeavq-cai
```
This command is both non-standard for `dfx` projects and odd because `ic` is specified twice:  Once as an env var and once as the `dfx` network.  But the `dfx` network is already exposed as an environment variable.  See `DFX_NETWORK` in [the dfx env var documentation](https://internetcomputer.org/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-envars).

# Changes
* If `ENV` is not set explicitly, use `DFX_NETWORK`.

# Tests


AntonioVentilii (commented):
Shall the documentation file be updated too?

bitdivine:
Good catch.  I updated the docs but that commit was lost somehow. :-( There was an automatic formatting commit that went in at the same time and the two tangled.  I'll redo the changes.

AntonioVentilii (approved):
LGTM tks!

PR #4119, opened 2025-01-08
fix(frontend): fixes confirmation handling

BonomoAlessandro:
# Motivation

After a frontend calculation which sets the page into a `dirty` state (which means that a confirm message is displayed if the user tries to navigate) the state is not reseted correctly. This leads to the unwanted behaviour that every time a user tries to reload the page, a confirm message is displayed.

# Changes

- fixes state handling

# Tests

Every time the `browser refresh` button is clicked and the page is not reloading, following message is displayed (it's not visible in the video):
image

This should only happen if the sending process is going on and in every other case the page should reload instead of showing this message.

**before:**

https://github.com/user-attachments/assets/3ae4fd04-f9ac-45ee-9eb0-91c578878d89

**after:**

https://github.com/user-attachments/assets/ea41e773-31af-41a9-996d-a1af96286f30


AntonioVentilii (approved):
Thank you! Good one!

src/frontend/src/lib/components/ui/InProgressWizard.svelte
```diff
@@ -15,10 +15,14 @@
 	export let steps: ProgressSteps;
 	export let warningType: 'transaction' | 'manage' = 'transaction';
 
-	const startConfirmToClose = () => dirtyWizardState.set(true);
-	const stopConfirmToClose = () => dirtyWizardState.set(false);
-
-	$: confirmToCloseBrowser($dirtyWizardState);
+	const startConfirmToClose = () => {
```
peterpeterparker, line 18: I don't understand this fix. What was the bug technically speaking?

BonomoAlessandro: @peterpeterparker When the `InProgressWizard` component gets destroyed, the `confirmToCloseBrowser` function was never called with the value `false`. This means that the `beforeunload` listener we added was never removed.

BonomoAlessandro: With this change we make sure that the event listener is removed correctly if the component gets destroyed

peterpeterparker: Gotcha thanks for the explanation. We missed that when we tested this [feedback](https://github.com/dfinity/oisy-wallet/pull/3859#discussion_r1877775599), all clear.

peterpeterparker (approved):
LGTM, thx

PR #4122, opened 2025-01-08
fix(frontend): write error log instead of showing an error toast

BonomoAlessandro:
# Motivation

If the wallet is loaded the application tries to determine if the logged in user is a vip user or not. This displays an error toast if the application is run locally because we do not run the `rewards canister` locally at the moment. Cause of this the e2e tests fail. As long as we do not run the `rewards canister` locally the error toast should not be displayed but an error log should be made.

# Changes

- replaces error toast with error log

# Tests

**before:**

https://github.com/user-attachments/assets/c76f0eda-ffaf-457d-acb7-c14d5f45513e

**after:**

https://github.com/user-attachments/assets/bd298c9a-576a-486f-af77-52d67967deb7


src/frontend/src/lib/services/reward-code.services.ts
```diff
@@ -41,10 +41,7 @@ export const isVipUser = async (params: { identity: Identity }): Promise<ResultS
 		return await queryVipUser({ ...params, certified: false });
 	} catch (err: unknown) {
 		const { vip } = get(i18n);
-		toastsError({
-			msg: { text: vip.reward.error.loading_user_data },
-			err
-		});
+		console.error(vip.reward.error.loading_user_data, err);
```
AntonioVentilii, line 44: let make it dependent on `LOCAL`, and add a `TODO` that it is a temporary fix

AntonioVentilii (approved):
LGTM tks!

PR #4123, opened 2025-01-08
feat(frontend): add edge case for SOL txn from and to my wallet

loki344:
# Motivation

Add the edge case where a user sends an amount from his wallet, to his wallet. Due to the hard coded values of the from and to in the mapper, this is handled wrongly. In these cases, the SOL transaction only includes 3 Accountkeys: My wallet, the SOL System Program, and the SOL Compute Budget program. This case needs special handling which is introduced by a ternary in the mapper.

# Changes

Add SystemAccountKeys from SOL to filter them out. Add edge case in mapping functionality to have correct recipient.

# Tests

Unit test added for this case


AntonioVentilii (approved):
LGTM tks

PR #4124, opened 2025-01-08
style(frontend): Updated banner for Initializing state

artkorotkikh-dfinity:
# Motivation

We’d like to replace the “welcome” image to better fill the space and to use it as a preview of which networks OISY supports (for example, the Solana announcement).

# Changes

Replaced one image.
Note that the image is somewhat stretched on desktop, but this issue existed beforehand and can be fixed separately. This PR only updates the image.

# Tests

**Before:** 
image

**After:**
image
image


AntonioVentilii (approved):
THank you! LGTM!

PR #4125, opened 2025-01-08
feat(frontend): sol txn api

loki344:
# Motivation

Implement the method to receive SOL transactions with the rpc. This method fetches transactions in batches until the defined limit is reached. The proposed way to fetch transactions in the solana network is:
1. Fetch Signatures for a specific wallet
2. Fetch transaction details for the signature
3. Process transaction

We filter out signatures with errors.

# Changes

Implement getSolTransactions in solana.api.ts

# Tests

Unit tests provided


src/frontend/src/sol/api/solana.api.ts
```diff
@@ -19,3 +27,87 @@ export const loadSolLamportsBalance = async ({
...
+
+		transactions.push(...uiTransactions);
+
+		const hasNoMoreSignaturesLeft = signatures.length < limit;
+		const hasLoadedEnoughTransactions = transactions.length >= limit;
+
+		if (hasNoMoreSignaturesLeft || hasLoadedEnoughTransactions) {
+			break;
+		}
+	}
```
AntonioVentilii, line 85: any way of making it more functional?

`const transactions = ...`

src/frontend/src/sol/api/solana.api.ts
```diff
@@ -19,3 +27,87 @@ export const loadSolLamportsBalance = async ({
...
+	return transactions.slice(0, limit);
+};
+
+const getTransactionDetailForSignature = async ({
+	signature: { confirmationStatus, signature },
+	network
+}: {
+	signature: SolSignature;
+	network: SolanaNetworkType;
+}) => {
```
AntonioVentilii, line 96: missing return type?

src/frontend/src/sol/api/solana.api.ts
```diff
@@ -19,3 +27,87 @@ export const loadSolLamportsBalance = async ({
...
+	}).send();
+
+	if (isNullish(rpcTransaction)) {
+		return null;
+	}
+
+	return {
+		...rpcTransaction,
+		version: rpcTransaction.version,
+		confirmationStatus: confirmationStatus,
```
AntonioVentilii, line 110:
```suggestion
		confirmationStatus,
```

src/frontend/src/sol/api/solana.api.ts
```diff
@@ -19,3 +27,87 @@ export const loadSolLamportsBalance = async ({
...
+		maxSupportedTransactionVersion: 0
+	}).send();
+
+	if (isNullish(rpcTransaction)) {
+		return null;
+	}
+
+	return {
+		...rpcTransaction,
+		version: rpcTransaction.version,
```
AntonioVentilii, line 109: no need for this, right? it is already in `rpcTransaction`, no?

src/frontend/src/sol/api/solana.api.ts
```diff
@@ -19,3 +27,87 @@ export const loadSolLamportsBalance = async ({
...
+		const transactionDetails = await Promise.all(
+			signatures
+				.filter(({ err }) => isNullish(err))
+				.map(async (signature) => await getTransactionDetailForSignature({ signature, network }))
+		);
+
+		const uiTransactions: SolCertifiedTransaction[] = transactionDetails
+			.filter(nonNullish)
+
+			.map((transaction) => ({
```
AntonioVentilii, line 69: I would separate the mapping from the raw transaction to the UI transaction, in another function, so that it is more maneagable

src/frontend/src/sol/api/solana.api.ts
loki344: that's what I thought too, but there is a linting error if i just spread rpcTransaction.
image

src/frontend/src/sol/api/solana.api.ts
loki344: Ok, I thought it makes sense to have it here since this is the entry point from outside world into => Oisy domain. I'll change it though as the patterns of the other integrations are also with a later mapping

src/frontend/src/sol/api/solana.api.ts
loki344: Do you mean chaining the promises instead of awaiting them, or introducing lambdas with reduce?

AntonioVentilii: nono, I meant, remove the `while` loop and use `reduce` for example, or `map`? Is that possible?

loki344: I replaced the while loop with recursion combined with reduce. Better?

loki344: Note that this loads first all signatures in batches, and then fetches all transactions for the signatures. Vs. before: load a batch of signatures => load transactions for it.

AntonioVentilii: Code-wise I like it more, but your point is that is less performant?

AntonioVentilii (approved):
LGTM tks

PR #4131, opened 2025-01-08
feat(frontend): add SolSendReview component

AntonioVentilii:
# Motivation

Similar to Bitcoin, we create component `SolSendReview`.


src/frontend/src/sol/components/send/SolSendReview.svelte
```diff
@@ -0,0 +1,34 @@
...
+	let invalid = true;
+	$: invalid = invalidSolAddress(destination) || invalidAmount(amount);
+</script>
+
+<SendReview on:icBack on:icSend {source} {amount} {destination} disabled={disableSend}>
+	<ReviewNetwork sourceNetwork={network} slot="network" />
+
+	<svelte:fragment slot="info">
+		{#if insufficientFundsForFee}
+			<InsufficientFundsForFee testId="sol-send-form-insufficient-funds-for-fee" />
```
DenysKarmazynDFINITY, line 31: We'll need to make use of these testIds at some point ;) But I guess I will be doing the same with the Swap feature - merging components first, and coming back to unit tests after the feature is on main just so we can get a security review sooner.

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4135, opened 2025-01-08
feat(frontend): sol txn listener and worker

loki344:
# Note
first: https://github.com/dfinity/oisy-wallet/pull/4125

# Motivation

After adding the api for fetching SOL transactions, we want to use it in the SOL wallet scheduler.

# Changes

Implement transaction handling in sol wallet scheduler.

# Tests

Extended existing unit tests.


AntonioVentilii (approved):
LGTM tks

PR #4136, opened 2025-01-08
feat(frontend): sol txn service

loki344:
# Note
first: https://github.com/dfinity/oisy-wallet/pull/4125

# Motivation

We use the SOL api to fetch transactions and store them in the txn store. As we are fetch SOL native transactions, the tokenId is determined via the network parameter. The loadNextSolTransactions will be the entry point for our SolTransactionsScroll which will follow in the next MR.

# Changes

Implement sol-transaction service

# Tests

Unit tests provided


AntonioVentilii (approved):
LGTM tks

PR #4140, opened 2025-01-09
feat(frontend): update SNS tokens

peterpeterparker:
# Motivation

We cannot merge #4139 because we are still not handling properly the CTS deprecation. Therefore this PR that cherry pick the changes we have to integrate.


AntonioVentilii (approved):
Thank you!

PR #4142, opened 2025-01-09
build(frontend): hardcode AUTH_DERIVATION_ORIGIN in code instead of ENV files

AntonioVentilii:
# Motivation

We aim to remove the `AUTH_DERIVATION_ORIGIN` from the env files.

# Changes

- Set `AUTH_DERIVATION_ORIGIN` as hard-coded values in `app.constant`, instead of using env variable.
- Integrate `hasDerivationOrigin` logic inside `getOptionalDerivationOrigin` to explicitly have the type hint that `AUTH_DERIVATION_ORIGIN` is not `undefined`.
- Remove `VITE_AUTH_DERIVATION_ORIGIN` from `.env.example`.
- Remove `VITE_AUTH_DERIVATION_ORIGIN` from deploy CI.

# Tests

Current tests works.


src/frontend/src/lib/utils/auth.utils.ts
```diff
@@ -38,12 +39,10 @@ const isAlternativeOrigin = (): boolean => {
 	return knownAlternativeOrigins.includes(origin);
 };
 
-const hasDerivationOrigin = (): boolean => notEmptyString(AUTH_DERIVATION_ORIGIN);
-
 export const getOptionalDerivationOrigin = ():
 	| { derivationOrigin: string }
 	| Record<string, never> =>
-	isAlternativeOrigin() && hasDerivationOrigin()
+	isAlternativeOrigin() && !isNullishOrEmpty(AUTH_DERIVATION_ORIGIN)
```
peterpeterparker, line 45: `notEmptyString` instead of `!isNullishOrEmpty`?

peterpeterparker (approved):
LGTM, thx

A suggestion maybe for simplification.

src/frontend/src/lib/utils/auth.utils.ts
AntonioVentilii: the problem is that `notEmptyString` does not define the type of the variable as non-nullish, so I have an issue with the statement after


```typescript
Type { derivationOrigin: string | undefined; } is not assignable to type
{   derivationOrigin: string; }

```

  while `isNullishOrEmpty` does
  
  Shall I modify those utils `isNullishOrEmpty` and `isEmptyString` in `@dfinity/utils` to define the type? WDYT?

peterpeterparker: > Shall I modif...

Maybe not, I'm afraid of the side effect and `value is string` doesn't exactly mean not empty, I guess that's why we use a boolean.

All good, thanks for the explanation.

PR #4143, opened 2025-01-09
Backend build

bitdivine:
# Motivation

# Changes

# Tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -85,7 +85,6 @@ RUN ls scripts
 RUN mkdir target
 COPY ./target/tags target/tags
 COPY ./target/commit target/commit
-RUN touch src/*/src/*.rs
 # The Wasm build is slow and typically cached
 RUN scripts/build.backend.wasm.sh
 # The network may be overridden with arguments to the docker call:  --env DFX_NETWORK=whatever 
--- a/Dockerfile
+++ b/Dockerfile
@@ -86,7 +86,11 @@ RUN mkdir target
 COPY ./target/tags target/tags
 COPY ./target/commit target/commit
 RUN touch src/*/src/*.rs
-RUN dfx build --ic backend
+# The Wasm build is slow and typically cached
+RUN scripts/build.backend.wasm.sh
+# The network may be overridden with arguments to the docker call:  --env DFX_NETWORK=whatever 
+ENV DFX_NETWORK=ic
+RUN dfx build backend --network "$DFX_NETWORK"
 
 FROM scratch AS backend
 COPY --from=build_backend out/ /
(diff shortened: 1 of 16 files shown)


.github/actions/docker-build-backend/action.yml
```diff
@@ -18,6 +18,10 @@ runs:
     - name: Set up docker buildx
       uses: docker/setup-buildx-action@v3
 
+    - name: Prepare docker build inputs
+      shell: bash
+      run: scripts/docker-build.pre
```
bitdivine, line 23: The docker build gets commit and tag data as input now.  The command above puts this information into files given to docker.

Dockerfile
```diff
@@ -82,7 +81,7 @@ COPY src/cycles_ledger/client/Cargo.toml src/cycles_ledger/client/Cargo.toml
 COPY src/cycles_ledger/pic/Cargo.toml src/cycles_ledger/pic/Cargo.toml
 COPY src/cycles_ledger/types/Cargo.toml src/cycles_ledger/types/Cargo.toml
 COPY src/shared/Cargo.toml src/shared/Cargo.toml
```
bitdivine, line 83: We use the same target directory as for local builds, so that we can have the same code paths and eliminate branches.

Dockerfile
```diff
@@ -92,19 +91,26 @@ RUN mkdir -p src/backend/src \
     && mkdir -p src/cycles_ledger/types/src \
     && touch src/cycles_ledger/types/src/lib.rs \
     && mkdir -p src/shared/src \
```
bitdivine, line 93: The command below is essentially the same, just updated.  The goal: We build, so that the cache is warm.

Dockerfile
```diff
@@ -92,19 +91,26 @@ RUN mkdir -p src/backend/src \
...
+COPY ./in/tags in/tags
+COPY ./in/commit in/commit
+ENV DFX_NETWORK=ic
+COPY scripts/commit-metadata scripts/
 RUN touch src/*/src/lib.rs src/*/*/src/lib.rs
+RUN dfx build backend --network "$DFX_NETWORK"
 
-RUN ./docker/build --backend
-
-RUN sha256sum /backend.wasm.gz
```
bitdivine, line 107: The `dfx build backend` command prints these hashes, both for local and docker builds, so this does not need to be repeated.

HACKING.md
```diff
@@ -54,11 +54,26 @@ dfx deploy frontend --network beta --wallet yit3i-lyaaa-aaaan-qeavq-cai
 
 ### IC
 
-> To perform production development, you'll need a `.env.production` file.
+Ensure that you have `dfx-orbit` installed and are using the correct station:
```
AntonioVentilii, line 57: here I would add some link to how to install orbit

bitdivine: Good idea.  Done: Linked to the dfx-orbit README that includes installation instructions.

AntonioVentilii (approved):

PR #4144, opened 2025-01-09
feat(frontend): expose schnorr_sign from signer canister

AntonioVentilii:
# Motivation

To sign a Solana transaction we will need to use the `schnorr_sign` function from the signer canister.

# Changes

- Create function in `SignerCanister` class to expose `schnorr_sign`. We are using the same parameters as `schnorr_public_key`.
- Create util in signer API module.

# Tests

Created unittest.


src/frontend/src/lib/api/signer.api.ts
```diff
@@ -98,6 +98,19 @@ export const getSchnorrPublicKey = async ({
 	return await getSchnorrPublicKey(rest);
 };
 
+export const signSchnorrPublicKey = async ({
+	identity,
+	...rest
+}: CanisterApiFunctionParams<{
+	derivationPath: string[];
```
DenysKarmazynDFINITY, line 105: Small suggestion: I usually put this type to `src/frontend/src/lib/types/api.ts`, it can be then reused by API and canister files.

DenysKarmazynDFINITY (approved):
Just a small comment, LGTM otherwise

src/frontend/src/lib/api/signer.api.ts
AntonioVentilii: Good one! Done! Tks

PR #4147, opened 2025-01-09
fix(frontend): add Solana feature flag in tokens list

AntonioVentilii:
# Motivation

Just for insurance, we add the feature flag for Solana in the list of tokens to pin too.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4150, opened 2025-01-09
fix(frontent): updates in progress wizard

BonomoAlessandro:
# Motivation

The `onDestroy` function of the `InProgressWizard` component is not called every time. To ensure that still every listener is removed and the application is left in a clean state we do have a workaround. When the `dirtyWizardState` was introduced the workaround function was not updated correctly which sometimes left the application in a weard state. To fix this, the workaround needs to be called every time the `$modalStore` AND the `$dirtyWizardState` changes.

# Changes

- updates workaraound cleanup function


AntonioVentilii (approved):
LGTM tks

PR #4152, opened 2025-01-09
style(frontend): update footer

loki344:
# Motivation

Update copyright 2024 => 2025
Use Heart svg instead of emoji to have consistent rendering across browsers.

# Tests

before:

image

image

after:

image

image


src/frontend/src/lib/components/core/Footer.svelte
```diff
@@ -78,6 +79,8 @@
 					class="text-center md:text-left"
 				>
 					{replaceOisyPlaceholders($i18n.footer.text.incubated_with)}
+					<IconHeart styleClass="inline-flex mb-1" />
+					{replaceOisyPlaceholders($i18n.footer.text.by)}
```
DenysKarmazynDFINITY, line 83: Both `$i18n.footer.text.by` and `$i18n.footer.text.incubated_with` can now be used without `replaceOisyPlaceholders` as they don't have any placeholders anymore.

DenysKarmazynDFINITY (approved):
Small comment, otherwise LGTM

src/frontend/src/lib/components/core/Footer.svelte
loki344: Yes, TBH it seems all of the texts would have worked without this.

loki344:
just for reference after removing replacePlaceholder (they were unnecessary anyways)
image

PR #4156, opened 2025-01-09
feat(frontend): remove todo and explicit undefined return

peterpeterparker:
# Motivation

Alessandro noticed that we can actually not use `ResultSuccess<VipReward>` because that would be a misuage of `ResultSuccess`. Something I mislead the type and thought about ` {result: T, err?: Err} ` or `{result: T, err?: Err}` but, we do not have those patterns. So, for pragamatical reason, we decided to remove the TODO.

As a nipick, since I removed the TODO, I added an explicit `return undefined` for the function.


BonomoAlessandro (approved):
LGTM thx

PR #4159, opened 2025-01-09
feat(frontend): add SolSendTokenWizard component

AntonioVentilii:
# Motivation

Similar to Bitcoin we create the component `SolSendTokenWizard` and add it to the generic send wizard.


src/frontend/src/sol/components/send/SolSendTokenWizard.svelte
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		if (isNullish($sendToken)) {
+			toastsError({
+				msg: { text: $i18n.tokens.error.unexpected_undefined }
+			});
+			return;
+		}
+
+		if (isNullish($authIdentity)) {
```
DenysKarmazynDFINITY, line 94: Let's move it to the very top of the function

DenysKarmazynDFINITY (approved):
Small comment, but approving already.

src/frontend/src/sol/components/send/SolSendTokenWizard.svelte
AntonioVentilii: done! good point

PR #4160, opened 2025-01-09
feat(frontend): icons update

DenysKarmazynDFINITY:
# Motivation

In this PR, 2 new icons have been added, as well as an existing got an additional `size` prop.


AntonioVentilii (approved):
I suppose that including the usage in this PR would have been too much?

Anyway, if yes, approved, since any non-used component will be removed but the CI

DenysKarmazynDFINITY:
@AntonioVentilii yeah, these icons are gonna be used across several Swap components, that's why I wanted to merge them separately. We'll anyway have some unused components pending for the time being, since I'm going to merge the feature from bottom to top (same as we did with other massive feature branches).

PR #4161, opened 2025-01-09
style(frontend): Updated styles for large info boxes

artkorotkikh-dfinity:
Replaced leagacy stylings and colors for the "Convert to ckXXX" token helper info boxes 

# Motivation
These boxes used old styles which are differrent from what we have now. 

# Changes

1. Removed border 
2. Smaller close button (same that we use for info boxes on Activity page)
3. Changed description text to match new colors 

# Tests

**Before**

image

**After** 
image

image

image


AntonioVentilii (approved):
LGTM tks!

PR #4163, opened 2025-01-10
ci(build): add rust and cargo to bot approved files

AntonioVentilii:
# Motivation

Missing files from the approved for the Bots: cargo and rust files.


DenysKarmazynDFINITY (approved):
Thanks

PR #4167, opened 2025-01-10
feat(frontend): create service to send SOL tokens

AntonioVentilii:
# Motivation

Solely based in the Solana web3.js [documentation](https://solana.com/developers/cookbook/transactions/send-sol), we create a function to send SOL tokens.

# Note

The code is an actually copy of the documentation one, with the difference that the signer function is the one provided by the Signer canister.


DenysKarmazynDFINITY (commented):
The code itself looks good to me, though I would really love to have such a core function covered with unit tests. I'm fine with adding a TODO comment and tackling it in a separate PR.

AntonioVentilii:
@DenysKarmazynDFINITY you are right, I forgot to commit them. But in all honesty, they are very mocked, meaning that it is difficult to test them like this. I added a TODO to test it in the local environment in another PR, deploying Solana local node

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4168, opened 2025-01-10
feat(frontend): adapt pageToken to match network too

AntonioVentilii:
# Motivation

We will start to have the same token on different networks. For example USDC both on Ethereum and Solana. To avoid any mistake or mismatch, we change the store `pageToken` to compare the network name too.

# Changes

Add network name logic in `pageToken` based on store `routeNetwork`.

# Tests

Modified existing tests.


src/frontend/src/tests/lib/derived/page-token.derived.spec.ts
```diff
@@ -32,22 +32,22 @@ describe('page-token.derived', () => {
...
+		mockPage.mock({ token: mockToken.name, network: mockToken.network.name });
 
 		expect(get(pageToken)?.symbol).toBe(mockToken.symbol);
 	});
 
 	it('should find ICRC token', () => {
 		const mockToken = { ...mockIcrcCustomToken, enabled: true };
 		icrcCustomTokensStore.setAll([{ data: mockToken, certified: true }]);
-		mockPage.mock({ token: mockToken.name });
+		mockPage.mock({ token: mockToken.name, network: mockToken.network.name });
```
DenysKarmazynDFINITY, line 50: Let's add at least 1 test case when `name` matches and `networkName` doesn't

DenysKarmazynDFINITY: (and maybe vice versa)

AntonioVentilii: Right! Done!

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4173, opened 2025-01-10
feat(frontend): add devnet spl tokens eurc and usdc

loki344:
# Motivation

We introduce SPL solana tokens USDC and EURC for devnet.

# Changes

add/adjust typing
add devnet tokens


AntonioVentilii (approved):
LGTM tks

PR #4174, opened 2025-01-10
feat(frontend): implement Swap component and use it in Actions

DenysKarmazynDFINITY:
# Motivation

This is the last step of the Swap feature - Swap component + its usage.


AntonioVentilii (approved):
LGTM tks

PR #4180, opened 2025-01-11
feat(frontend): use a single web worker

AntonioVentilii:
# Credits

Copied from @peterpeterparker 's PR https://github.com/junobuild/juno/pull/1035 . The motivation below is the same as the Juno's PR, written by @peterpeterparker .

# Motivation

I don't know why I didn't realize this earlier, but each worker in our setup is bundled by vite and rollup into a standalone chunk without reusing chunks across workers. As a result, we are shipping quite a bit of redundant code since almost every worker includes both Buffer, agent-js and other inherited stuffs.

I couldn't find any non-hacky bundling option to improve this. However, I did come up with a potential solution: using a single worker entry point.

Since we are already splitting each worker into different modules, we can easily integrate a single worker entry point that dispatches messages to the appropriate modules. Furthermore, because the context is created at usage, even with a single entry point, we still maintain separate contexts for each worker.

# Results

### Before

Screenshot 2025-01-11 at 10 20 28

### After

Screenshot 2025-01-11 at 10 21 16

### DIFF
--- a/src/frontend/src/btc/services/worker.btc-wallet.services.ts
+++ b/src/frontend/src/btc/services/worker.btc-wallet.services.ts
@@ -26,7 +26,7 @@ export const initBtcWalletWorker = async ({
 	token: Token;
 	minterCanisterId?: OptionCanisterIdText;
 }): Promise<WalletWorker> => {
-	const WalletWorker = await import('$btc/workers/btc-wallet.worker?worker');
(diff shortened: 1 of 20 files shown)


peterpeterparker (approved):
LGTM, thx

Let's see if there are improvements.

Production:
Capture d’écran 2025-01-13 à 08 19 30

Staging:
Capture d’écran 2025-01-13 à 08 20 16

PR #4181, opened 2025-01-11
feat(frontend): add SPL balance in wallet worker and service

AntonioVentilii:
# Motivation

We include the SPL balance service in the wallet workers. To do so, we pass the SPL token address to the workers IF the Solana token is an SPL token and not a native one.

# Changes

- Adapt worker balance to be either `Lamports` (for native SOL tokens) or `bigint` (for SPL tokens).
- Add `tokenAddress` as an optional parameter to the data passed to the worker.
- Create util to check if it is a SPL token (plus tests).
- Use the util to decide if passing the token address or `undefined`.
- Adapt the worker to use `loadSplTokenBalance` if there is `tokenAddress` provided.

# Tests

Test adapted and results below:

Screenshot 2025-01-11 at 12 34 24

### DIFF
--- a/src/frontend/src/lib/schema/post-message.schema.ts
+++ b/src/frontend/src/lib/schema/post-message.schema.ts
@@ -88,7 +88,8 @@ export const PostMessageDataRequestBtcSchema = z.object({
 export const PostMessageDataRequestSolSchema = z.object({
 	// TODO: generate zod schema for CertifiedData
 	address: z.custom<CertifiedData<SolAddress>>(),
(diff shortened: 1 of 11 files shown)


loki344 (approved):
LGTM. This approach differs  from the ICP implementation where we have different workers for the worker for icp and a different one for icrc. In general I think it makes sense to revisit the worker implementations and consolidate the approach. Looking at the tight timeline for this feature, let's get this in and revisit.

peterpeterparker:
> we have different workers for the worker for icp and a different one for icrc

Those ledgers have different APIs.

PR #4190, opened 2025-01-13
feat(frontend): add Solana API call to load token account

AntonioVentilii:
# Motivation

When transferring SPL tokens we deal with the specific token account associated to an address. So we create the call to fetch such a value, given the wallet address, the token address and the network.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4191, opened 2025-01-13
feat(frontend): add service to send SPL tokens

AntonioVentilii:
# Motivation

To send SPL tokens, we need to add the function that creates the message instructions correct (that is provided by the library [`@solana-program/token`](https://www.npmjs.com/package/@solana-program/token)). Furthermore we will use the token accounts associated to `source` and `destination`.

# Changes

- Install [`@solana-program/token`](https://www.npmjs.com/package/@solana-program/token).
- Add Token Program Address constant.
- Create `createSplTokenTransactionMessage` that will return the correct instruction message for SPL tokens.

# Tests

Included unitest.


src/frontend/src/sol/services/sol-send.services.ts
```diff
@@ -127,12 +187,20 @@ export const sendSol = async ({
...
-		amount,
-		network: solNetwork
-	});
+	const transactionMessage = isTokenSpl(token)
+		? await createSplTokenTransactionMessage({
+				signer,
+				destination,
+				amount,
+				network: solNetwork,
+				tokenAddress: token.address
```
DenysKarmazynDFINITY, line 196: Since you are already destructing some other properties from `token`, let's do it the same for `address`.

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4195, opened 2025-01-13
chore(devops): Specify the rewards canister to use for beta

bitdivine:
# Motivation
The rewards functionality needs to be tested in beta.  beta will use the mainnet rewards canister.

# Changes
- Add an entry in dfx.json specifying the canister id of the rewards canister for beta.

# Tests
N/A


peterpeterparker (approved):
Rubber stamp

PR #4196, opened 2025-01-13
feat(frontend): extract InputSearch component

DenysKarmazynDFINITY:
# Motivation

Since Tokens list view is now gonna be available within the Swap feature, the idea is to re-use as match things from ManageTokens component as possible. One of such things - `InputSearch`.


AntonioVentilii (approved):
Good one! Tks!

PR #4197, opened 2025-01-13
fix(frontend): rename kong_backend API param

DenysKarmazynDFINITY:
# Motivation

There was an incorrectly named kong_backend API param - it should have been `receiveAddress` instead of `destinationAddress`.


AntonioVentilii (approved):
Tks

PR #4198, opened 2025-01-13
feat(frontend): expose sendIcp and sendIcrc methods

DenysKarmazynDFINITY:
# Motivation

We need to expose sendIcp and sendIcrc methods in order to initiate transactions and get their block indexes before calling the swap API method of kong_backend.

### DIFF
--- a/src/frontend/src/icp/services/ic-send.services.ts
+++ b/src/frontend/src/icp/services/ic-send.services.ts
@@ -93,13 +93,15 @@ const send = async ({
 	});
 };
 
-const sendIcrc = ({
+export const sendIcrc = ({
 	to,
 	amount,
 	identity,
 	ledgerCanisterId,
 	progress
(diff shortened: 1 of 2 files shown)


DenysKarmazynDFINITY:
Update: according to Kong, using icrc1 send is fine for the MVP version of the feature since they to send the assets back in case swap is unsuccessful. In the next version, we will start using icrc2 approval for tokens that support it, plus we will additionally implement the claiming functionality (which again is there just in case, according to Kong).

peterpeterparker:
Discussed offline, the team decided to proceed with an "ICRC-1 transfer + swap" method.

AntonioVentilii (approved):
As @peterpeterparker mentioned [above](https://github.com/dfinity/oisy-wallet/pull/4198#issuecomment-2588060122), it was discussed offline that this is the strategy decided until one of the next close releases.

Approving with the disclaimer that we will implement ICRC2 in one of the close release after the next one.

PR #4200, opened 2025-01-13
fix(frontend): fix hide zeros filter

BonomoAlessandro:
# Motivation

If a user who is not invested in ck tokens activates the `Hide zeros` option, these tokens disappear. But the user should always know, that there is an option to invest in cks. This means if there is a token group and at least one of the tokens do have a balance, display the whole group. 

# Changes

- updates filter logic

# Tests

**before:**
image

image

**after:**
image

image


src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
```diff
@@ -4,24 +4,36 @@
...
+	$: {
+		const hasBalance = (token: TokenUiOrGroupUi) => {
+			const checks = [
+				() => Number(token.balance ?? 0n) !== 0,
+				() => token.usdBalance && token.usdBalance !== 0,
+				() => $showZeroBalances
+			];
+
+			return checks.some((check) => check());
+		};
```
AntonioVentilii, line 25: can we not use the `Number(balance ?? 0n) || (usdBalance ?? 0) || $showZeroBalances` logic? Will it not work?

BonomoAlessandro: @AntonioVentilii you are right. This is much better

src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
```diff
@@ -4,24 +4,29 @@
...
+	$: groupedTokens = groupTokensByTwin($combinedDerivedSortedNetworkTokensUi);
+
+	let sortedTokensOrGroups: TokenUiOrGroupUi[];
+	$: {
+		const hasBalance = (token: TokenUiOrGroupUi) =>
+			Number(token.balance ?? 0n) || Number(token.usdBalance ?? 0n) || $showZeroBalances;
+
+		sortedTokensOrGroups = groupedTokens.filter((t: TokenUiOrGroupUi) =>
+			isTokenUiGroup(t) ? t.tokens.some((tok: TokenUi) => hasBalance(tok)) : hasBalance(t)
+		);
```
AntonioVentilii, line 22: can the above be put in an util, using `reduce` with a few tests please?

BonomoAlessandro: @AntonioVentilii i can put the code into an util and write some tests, but why should i use `reduce` instead of `filter`? Filtering is exactly what i want to do in this case

AntonioVentilii (approved):

PR #4201, opened 2025-01-13
refactor(frontend): network id in test ID for TokenCard

AntonioVentilii:
# Motivation

Since we will start to have the same symbols for some ERC20 and SPL tokens, we diversify them in tests where the symbols are used as strings.

NOTE: when the symbols remains `symbol` type, it is not an issue, since they are always different.


bitdivine (approved):
Is there a way of using say ETH and SPL instead of the description?  I don't see a better way with the data available now but I may be missing something.

AntonioVentilii:
@bitdivine we use the network short-name as id, so for example: SOL, SOL(Testnet), SOL(Local), ETH, etc

And since in the future we probably have more, I was thinking that it is ok like this for now, no? WDYT?

PR #4202, opened 2025-01-13
chore(e2e): remove unused constant

AntonioVentilii:
# Motivation

Removing unused Activity page URL, that it is now accessed directly in the E2E tests like the other pages.


bitdivine (approved):
Confirmed that ACTIVITY_URL appears here only.

PR #4210, opened 2025-01-13
feat(frontend): add SPL tokens to pageToken store

AntonioVentilii:
# Motivation

The `pageToken` store will require to have access to SPL tokens too.

# Tests

Modified existing tests.

### DIFF
--- a/src/frontend/src/lib/derived/all-tokens.derived.ts
+++ b/src/frontend/src/lib/derived/all-tokens.derived.ts
@@ -6,6 +6,7 @@ import { icrcTokens } from '$icp/derived/icrc.derived';
 import { buildIcrcCustomTokens } from '$icp/services/icrc-custom-tokens.services';
 import { sortIcTokens } from '$icp/utils/icrc.utils';
 import { parseTokenId } from '$lib/validation/token.validation';
+import { splTokens } from '$sol/derived/spl.derived';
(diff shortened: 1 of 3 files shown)


src/frontend/src/tests/lib/derived/page-token.derived.spec.ts
```diff
@@ -18,6 +20,11 @@ describe('page-token.derived', () => {
 		mockPage.reset();
 		vi.spyOn(btcEnv, 'BTC_MAINNET_ENABLED', 'get').mockImplementation(() => true);
 		vi.spyOn(solEnv, 'SOLANA_NETWORK_ENABLED', 'get').mockImplementation(() => true);
+
+		vi.spyOn(splTokens, 'subscribe').mockImplementation((fn) => {
+			fn([{ ...JUP_TOKEN, enabled: true }]);
+			return () => {};
+		});
```
bitdivine, line 27: Why is the event different from BTC and SOLANA?

bitdivine (approved):

src/frontend/src/tests/lib/derived/page-token.derived.spec.ts
AntonioVentilii: As discussed offline, for now we don't have any SPL fetching from the backend, that is why for now we mock directly the main derived store.

PR #4213, opened 2025-01-13
feat(frontend): add Max button for Solana

AntonioVentilii:
# Motivation

Similar to the other networks, we enable the Max button using the `calculateMax` function.

Screenshot 2025-01-13 at 16 59 59
Screenshot 2025-01-13 at 17 02 10


bitdivine (approved):
I still really like Lamports as a name.

PR #4217, opened 2025-01-13
feat(frontend): fix transactions that not touch sol balance showing up in txns

loki344:
# Motivation

Signatures that do not change SOL balances were shown in SOL txns. https://solscan.io/tx/JJU8JARp4WZV3oq5JmSujSRepq3czPo1vVkoRJrwyrQDbCsGzYUB4efbmVtMGZ9wNeHwkic3qh89ycpaZwRjktU?cluster=devnet

# Changes

We check for the SOL native balance change and only include txns that change the balance.

# Tests

unit test added

after:
image

before:

image


loki344:
> Maybe just add a `TODO` that specify that we should show anyway self-sent transactions.
> 
> For the rest, LGTM

they will show up as they have fees and therefore change the balance

PR #4222, opened 2025-01-13
feat(frontend): allDisabledIcrcTokens derived

DenysKarmazynDFINITY:
# Motivation

This PR includes:
1. Expose `allIcrcTokens` store to be used in SwapTokensList.
2. Create a new `allDisabledIcrcTokens` store in order to have all disabled icrc tokens for which we need to fetch balances and USD rates.


AntonioVentilii (commented):
Just a doubt: isn't it the same as checking if the prop `enabled` is `nonNullish` and `true`? Or am I confusing things?

DenysKarmazynDFINITY:
@AntonioVentilii it's a good question, but unfortunately it's not the same, you can see the comments that we have for `allIcrcTokens` store - https://github.com/dfinity/oisy-wallet/blob/main/src/frontend/src/lib/derived/all-tokens.derived.ts#L13. And then, after using `buildIcrcCustomTokens` to have all the available icrc tokens, I'm filtering already enabled tokens to avoid fetching data that is already known.

AntonioVentilii (approved):
Thank you for the explanation! Approved

PR #4223, opened 2025-01-13
feat(frontend): add SPL token DRIFT

StefanBerger-DFINITY:
Adding the SPL token DRIFT


StefanBerger-DFINITY:
Note: SVG is pending. Needs to be created by us.

PR #4224, opened 2025-01-13
refactor(frontend): move WalletConnect to lib

AntonioVentilii:
# Motivation

We will use `WalletConnect` for more networks (for example Solana). So, first, we move the generic components (and related types) to `lib`.


loki344 (approved):
LGTM, thanks

PR #4230, opened 2025-01-13
feat(frontend): fetch SPL prices from Coingecko

AntonioVentilii:
# Motivation

Including SPL token list in Coingecko services.

# Changes

# Tests


src/frontend/src/lib/components/exchange/ExchangeWorker.svelte
```diff
@@ -32,7 +34,8 @@
 	$: worker,
 		$enabledMergedErc20TokensAddresses,
 		$enabledIcrcLedgerCanisterIdsNoCk,
-		debounceSyncTimer();
+		$enabledSplTokenAddresses;
```
DenysKarmazynDFINITY, line 37: I think you added `;` instead of `,` after `$enabledSplTokenAddresses` and because of that `debounceSyncTimer()` is not a part of this reactive statement anymore.

AntonioVentilii: ah right!!!! good catch!

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4231, opened 2025-01-13
feat(frontend): add spl txns

loki344:
# Motivation

We want to display SPL token transactions.

# Changes

- Implement getSplTransactions
=> We get all token accounts of a token we're interested in, we get all signatures, and for them the transactions.

# Tests

Regarding the tight timeline, todos were added for unit test.

image


AntonioVentilii (approved):
LGTM! Small lint and test checks before merging

PR #4233, opened 2025-01-13
refactor(frontend): move WalletConnect components that have generic use

AntonioVentilii:
# Motivation

After https://github.com/dfinity/oisy-wallet/pull/4224 , we can move the other components that will be used for all the networks for WalletConnect.


loki344 (approved):
lgtm, thanks

PR #4234, opened 2025-01-14
feat(frontend): SwapInputCurrency component

DenysKarmazynDFINITY:
# Motivation

SwapInputCurrency - a small wrapper around existing InputCurrency, adds styling according to the design specs.

Screenshot 2025-01-14 at 09 24 30

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
+++ b/src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
@@ -1,0 +1,32 @@
+<script lang="ts">
+	import InputCurrency from '$lib/components/ui/InputCurrency.svelte';
+	import type { OptionAmount } from '$lib/types/send';
+	export let value: OptionAmount;
+	export let decimals: number;
+	export let name = 'swap-amount';
+	export let disabled = false;
+	export let placeholder = '0';
+	export let error = false;
+</script>
+
+<div class="swap-input-currency h-full w-full font-bold" class:text-error={error}>
+	<InputCurrency bind:value {name} {placeholder} {disabled} {decimals} on:focus on:blur>
+		<slot name="inner-end" slot="inner-end" />
+	</InputCurrency>
+</div>
+
+<style lang="postcss">
+	:global(.swap-input-currency div.input-block) {
+		display: block;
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
```diff
@@ -0,0 +1,32 @@
...
+		--padding: 0;
+		--input-width: 100%;
+	}
+	:global(.swap-input-currency div.input-field input[id]) {
+		height: 100%;
+		border: none;
+		border-radius: 0;
+		padding: 0 0 0 0.75rem;
+	}
+</style>
```
AntonioVentilii, line 32: why not putting the above in the CSS files?

AntonioVentilii (approved):
LGTM!

src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
DenysKarmazynDFINITY: These styles are only related to the Swap inputs, that's why I decided not to define them globally, but rather keep close to the component (same as it's done for Convert input). Both Convert and Swap will be reusing in the future the same component, this can be a good opportunity to move the styles to CSS files.

src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
```diff
@@ -0,0 +1,32 @@
...
+	export let error = false;
+</script>
+
+<div class="swap-input-currency h-full w-full font-bold" class:text-error={error}>
+	<InputCurrency bind:value {name} {placeholder} {disabled} {decimals} on:focus on:blur>
+		<slot name="inner-end" slot="inner-end" />
+	</InputCurrency>
+</div>
+
+<style lang="postcss">
```
peterpeterparker, line 18: @DenysKarmazynDFINITY postcss?

DenysKarmazynDFINITY: Indeed not needed as I'm not applying any tailwind classes. Will clean it up, thanks for the heads up.

PR #4237, opened 2025-01-14
feat(frontend): SwapAmountExchange component

DenysKarmazynDFINITY:
# Motivation

The PR includes the SwapAmountExchange component.

Screenshot 2025-01-14 at 09 24 30

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapAmountExchange.svelte
+++ b/src/frontend/src/lib/components/swap/SwapAmountExchange.svelte
@@ -1,0 +1,23 @@
+<script lang="ts">
+	import { nonNullish } from '@dfinity/utils';
+	import IconArrowUpDown from '$lib/components/icons/lucide/IconArrowUpDown.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import type { OptionAmount } from '$lib/types/send';
+	import { formatUSD } from '$lib/utils/format.utils';
+
+	export let amount: OptionAmount;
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/swap/SwapAmountExchange.svelte
```diff
@@ -0,0 +1,23 @@
...
+	import IconArrowUpDown from '$lib/components/icons/lucide/IconArrowUpDown.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import type { OptionAmount } from '$lib/types/send';
+	import { formatUSD } from '$lib/utils/format.utils';
+
+	export let amount: OptionAmount;
+	export let exchangeRate: number | undefined;
+
+	let amountUSD: number | undefined;
+	$: amountUSD = nonNullish(amount) && nonNullish(exchangeRate) ? Number(amount) * exchangeRate : 0;
```
AntonioVentilii, line 12: whouldn't be better to explicitly say when `exchangeRate` is nullish? something like n/a or exchange rate not available, or loading

AntonioVentilii: Ah wait, I just saw that you already guard for `nonNullish` exchange rate... so here you put it just for safe of the multiplier operation, so all good, thank you!

AntonioVentilii (approved):
LGTM tks

PR #4238, opened 2025-01-14
feat(frontend): sol wallet connect

loki344:
# Motivation

we want to reuse the wallet connect logic across multiple chains

# Changes

introduce base class for wallet connect
implement eth with it

# Tests

https://github.com/user-attachments/assets/bf04787d-b5fc-453e-9b7e-ad4cf0daa6be


AntonioVentilii (approved):
Approved with a small comment about changing the file name

loki344:
found a more correct / simpler solution

PR #4240, opened 2025-01-14
feat(frontend): SwapValue component

DenysKarmazynDFINITY:
# Motivation

SwapValue component:
Screenshot 2025-01-14 at 11 06 48

P.S.: I also used this PR to fix a small issue with SwapInputCurrency.

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
+++ b/src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
@@ -16,7 +16,7 @@
 	</InputCurrency>
 </div>
 
-<style>
+<style lang="scss">
 	:global(.swap-input-currency div.input-block) {
 		display: block;
 		height: 100%;
--- a/src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
+++ b/src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import InputCurrency from '$lib/components/ui/InputCurrency.svelte';
 	import type { OptionAmount } from '$lib/types/send';
+
 	export let value: OptionAmount;
 	export let decimals: number;
 	export let name = 'swap-amount';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
```diff
@@ -15,14 +16,15 @@
 	</InputCurrency>
 </div>
 
-<style lang="postcss">
+<style>
```
AntonioVentilii, line 19: should it be `<style lang="scss">`, no?

src/frontend/src/lib/components/swap/SwapValue.svelte
```diff
@@ -0,0 +1,20 @@
...
+
+	let secondaryValue: boolean;
+	$: secondaryValue = nonNullish($$slots['secondary-value']);
+</script>
+
+<div class="mb-2 flex w-full justify-between last:mb-0 md:items-center">
+	<label for={ref} class="mr-1 text-sm text-tertiary sm:mr-2"><slot name="label" /></label>
+
+	<div class="flex flex-col items-end sm:flex-row sm:items-center">
+		<span class={`mb-1 text-sm font-bold sm:mb-0 ${secondaryValue ? 'sm:mr-2' : ''}`}
```
AntonioVentilii, line 14: why not move `sm:mr-2` directly into the slot `secondary-value` ? does it work the same?

`<span class="text-sm text-tertiary sm:ml-2"><slot name="secondary-value" /></span>`

AntonioVentilii (approved):
Already approved with 2 small observations

src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
DenysKarmazynDFINITY: You're right, added

src/frontend/src/lib/components/swap/SwapValue.svelte
DenysKarmazynDFINITY: In that case, the wrapper will have `ml-2` no matter if `secondary-value` slot is provided or not, which will break the layout a little bit

PR #4244, opened 2025-01-14
refactor(frontend): move WalletConnect reject function to lib

AntonioVentilii:
# Motivation

The `reject` function and its sub-function `execute` will be used for more networks than Ethereum, so we move them to `lib`.


loki344 (approved):
lgtm, thx

PR #4247, opened 2025-01-14
feat(frontend): extend wallet connect listener with sol

loki344:
# Motivation

we want to include solana mainnet in the wallet connect listener if the feature flag is on.

# Changes

add namespace for solana mainnet, other will follow.

# Tests

https://github.com/user-attachments/assets/97636e32-584d-4e16-a033-0a2e2c95e9f0


AntonioVentilii (approved):
LGTM tks

PR #4253, opened 2025-01-14
Updated description for welcome drops

artkorotkikh-dfinity:
# Motivation

Since rewards might come with a few sec delay we'd like to adjust the description. 

# Changes

Only copy changes 

# Tests

No tests were done


StefanBerger-DFINITY (approved):
LGTM

PR #4256, opened 2025-01-14
feat(frontend): SwapTokensList component

DenysKarmazynDFINITY:
# Motivation

SwapTokensList component:
Screenshot 2025-01-14 at 14 50 54

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapTokensList.svelte
+++ b/src/frontend/src/lib/components/swap/SwapTokensList.svelte
@@ -1,0 +1,56 @@
+<script lang="ts">
+	import { createEventDispatcher, getContext } from 'svelte';
+	import { ICP_TOKEN } from '$env/tokens/tokens.icp.env';
+	import TokenCardContent from '$lib/components/tokens/TokenCardContent.svelte';
+	import TokenCardWithOnClick from '$lib/components/tokens/TokenCardWithOnClick.svelte';
+	import ButtonCancel from '$lib/components/ui/ButtonCancel.svelte';
+	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
+	import InputSearch from '$lib/components/ui/InputSearch.svelte';
+	import { allIcrcTokens } from '$lib/derived/all-tokens.derived';
+	import { exchanges } from '$lib/derived/exchange.derived';
+	import { balancesStore } from '$lib/stores/balances.store';
+	import { i18n } from '$lib/stores/i18n.store';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/swap/SwapTokensList.svelte
```diff
@@ -0,0 +1,56 @@
...
+	const dispatch = createEventDispatcher();
+
+	let filter = '';
+
+	let noTokensMatch = false;
+	$: noTokensMatch = filteredTokens.length === 0;
+
+	let tokens: Token[];
+	$: tokens = [ICP_TOKEN, ...$allIcrcTokens].filter(
+		(token: Token) => token.name !== $sourceToken?.name && token.name !== $destinationToken?.name
```
AntonioVentilii, line 29: isn't it better to compare ID?

src/frontend/src/lib/components/swap/SwapTokensList.svelte
```diff
@@ -0,0 +1,56 @@
...
+	noMatch={noTokensMatch}
+	placeholder={$i18n.tokens.placeholder.search_token}
+/>
+
+<div class="my-6 flex flex-col overflow-y-hidden sm:max-h-[26rem]">
+	<div class="tokens-scroll flex flex-col gap-6 overflow-y-auto overscroll-contain">
+		{#each filteredTokens as token (token.id)}
+			<TokenCardWithOnClick on:click={() => dispatch('icSelectToken', token)}>
+				<TokenCardContent
+					data={mapTokenUi({ token, $exchanges: $exchanges, $balances: $balancesStore })}
```
AntonioVentilii, line 47: ideally, this could be done as variable above?

```typescript
let tokenUi : TokenUI
$: mapTokenUi({ token, $exchanges: $exchanges, $balances: $balancesStore })
```

just for even more clarity?

AntonioVentilii (approved):
Approving, two small observations

src/frontend/src/lib/components/swap/SwapTokensList.svelte
DenysKarmazynDFINITY: yep, of course, forgot to update, doing it now

src/frontend/src/lib/components/swap/SwapTokensList.svelte
DenysKarmazynDFINITY: It's rendered inside an `each` loop, therefore `token` would not be available there. We can of course create a function that accepts it as param, but using `mapTokenUi` directly looks a bit cleaner.

PR #4258, opened 2025-01-14
feat(frontend): SwapFees component

DenysKarmazynDFINITY:
# Motivation

SwapFees component:

Screenshot 2025-01-14 at 14 59 11


src/frontend/src/lib/components/swap/SwapFees.svelte
```diff
@@ -0,0 +1,118 @@
...
+					{$destinationToken.symbol}
+				</svelte:fragment>
+			</SwapValue>
+
+			<SwapValue>
+				<svelte:fragment slot="label">{$i18n.swap.text.lp_fee}</svelte:fragment>
+
+				<svelte:fragment slot="main-value">
+					{liquidityProvidersFee}
+					{$destinationToken.symbol}
```
AntonioVentilii, line 107: nitpick: if you are going to use the symbol more time, put it in the SWAP context too?

AntonioVentilii (approved):
LGTM

src/frontend/src/lib/components/swap/SwapFees.svelte
DenysKarmazynDFINITY: I was thinking about it, but then decided not to do it since `$destinationToken` can be `undefined`, therefore after I did one check `nonNullish($destinationToken)` I can freely access all of its properties. If I expose token params separately, it can potentially increase the number of checks. In any case, after everything is merged, I'll re-validate this assumption.

PR #4260, opened 2025-01-14
feat(frontend): add price for twin tokens on Solana

AntonioVentilii:
# Motivation

For consistency, we use the twin token prices for the SPL tokens of Solana that have such a token.

### DIFF
--- a/src/frontend/src/lib/derived/exchange.derived.ts
+++ b/src/frontend/src/lib/derived/exchange.derived.ts
@@ -67,6 +67,17 @@ export const exchanges: Readable<ExchangesData> = derived(
 					}),
 					{}
 				),
+			...$splTokens
+				.filter(({ twinToken }) => nonNullish(twinToken))
+				.reduce((acc, { id, twinToken }) => {
+					const { address } = (twinToken as Partial<Erc20Token>) ?? { address: undefined };
+					const price = nonNullish(address) ? $exchangeStore?.[address.toLowerCase()] : undefined;
+
+					return {
+						...acc,
+						[id]: price
+					};
+				}, {}),
 			...$icrcTokens.reduce((acc, token) => {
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/derived/exchange.derived.ts
```diff
@@ -67,6 +67,17 @@ export const exchanges: Readable<ExchangesData> = derived(
 					}),
 					{}
 				),
+			...$splTokens
+				.filter(({ twinToken }) => nonNullish(twinToken))
```
DenysKarmazynDFINITY, line 71: No need to do it now as I think I saw this approach in a few different places, but I think we can avoid looping through the array twice and achieve the same result inside a single `reduce`.

DenysKarmazynDFINITY (approved):
Small comments, otherwise LGTM

src/frontend/src/lib/derived/exchange.derived.ts
AntonioVentilii: Added a TODO

src/frontend/src/lib/derived/exchange.derived.ts
```diff
@@ -67,6 +67,17 @@ export const exchanges: Readable<ExchangesData> = derived(
...
+				.reduce((acc, { id, twinToken }) => {
+					const { address } = (twinToken as Partial<Erc20Token>) ?? { address: undefined };
```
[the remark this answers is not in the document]
AntonioVentilii: done! tks

src/frontend/src/lib/derived/exchange.derived.ts
AntonioVentilii: Done with:

- https://github.com/dfinity/oisy-wallet/pull/5391

PR #4261, opened 2025-01-14
refactor(frontend): case of nullish balance in BtcSendAmount

AntonioVentilii:
# Motivation

As @DenysKarmazynDFINITY pointed out in https://github.com/dfinity/oisy-wallet/pull/4215#discussion_r1914730201 , we should consider the case that `$sendBalance` can be nullish too.


DenysKarmazynDFINITY (approved):
Thanks!

PR #4262, opened 2025-01-14
feat(frontend): SwapValueDifference component

DenysKarmazynDFINITY:
# Motivation

SwapValueDifference component:
Screenshot 2025-01-14 at 16 12 49


src/frontend/src/lib/components/swap/SwapValueDifference.svelte
```diff
@@ -0,0 +1,46 @@
...
+			: undefined;
+</script>
+
+{#if nonNullish(valueDifference)}
+	<span
+		class="underline"
+		class:text-success={valueDifference > SWAP_VALUE_DIFFERENCE_WARNING_VALUE}
+		class:text-warning={valueDifference <= SWAP_VALUE_DIFFERENCE_WARNING_VALUE &&
+			valueDifference > SWAP_VALUE_DIFFERENCE_ERROR_VALUE}
+		class:text-error={valueDifference <= SWAP_VALUE_DIFFERENCE_ERROR_VALUE}
```
AntonioVentilii, line 42: does it make sense to put these comparisons above, to have them triggered exactly at the same time? I think they are mirrored right?

or no risk?

P.S. is it ok to compare `undefined` with a `number` right?

AntonioVentilii (approved):
LGTM tks

src/frontend/src/lib/components/swap/SwapValueDifference.svelte
DenysKarmazynDFINITY: I think logic-wise there is no difference in this case whether we create vars for those checks or use them directly, since just one of those can be `true` at the same time.

Re. your second question - there is `nonNullish` check, so this part will only be rendered if valueDifference is not undefined.

PR #4275, opened 2025-01-14
fix(frontend): ckBTC minter info loaded on ckETH (and the contrary)

peterpeterparker:
# Motivation

We merged the workers into a single entry in #4180, which is great for performance reasons but implies that messages should be unique. It turns out this is not the case when loading the minter info—for example, the same messages were used for ckBTC and ckETH. As a result, when a user navigates to a ck token, the page attempts to load information for both ckETH and ckBTC. Given that their APIs are different, the responses cannot be decoded correctly for one or the other.

# Notes

There might be better solutions; this is just a quick fix.

# Changes

- Duplicated the messages to ensure uniqueness.
- Added a `postMessage` key to interact with the worker, as the UI-side code is shared.
- Moved the code to load the worker, making it more explicit that there is a single entry.

# Tests

No tests or debugging were performed.

### DIFF
--- a/src/frontend/src/icp/services/worker.ck-minter-info.services.ts
+++ b/src/frontend/src/icp/services/worker.ck-minter-info.services.ts
@@ -19,50 +19,41 @@ import type {
 } from '$lib/types/post-message';
 import type { SyncState } from '$lib/types/sync';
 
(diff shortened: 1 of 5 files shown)


AntonioVentilii (approved):
Thank you for the fix!

PR #4276, opened 2025-01-15
fix(e2e): set unique artifacts in CI

AntonioVentilii:
# Motivation

After PR https://github.com/dfinity/oisy-wallet/pull/4232 , there is the issue that since all the CI workflows of the matrix save an artifact, they can conflicts with their name.

That is why we assign a name unique for each workflow matrix element.


inc-man (approved):
LGTM, 

thought I had that implemented already but might be in one of the unmerged PRs

PR #4281, opened 2025-01-15
feat(frontend): sort SwapTokensList

DenysKarmazynDFINITY:
# Motivation

The idea is to apply to SwapTokensList the same sorting as we have in the main tokens list (tokens with balance should be pinned to the top).


AntonioVentilii (approved):
LGTM tks!

PR #4289, opened 2025-01-15
refactor(frontend): Extract function to create default transaction in Solana

AntonioVentilii:
# Motivation

Just to be more clear, we extract the common part of creating the default transaction for Solana, where:

1. the version is defined;
2. the payer is specified;
3. it is decided that the lifetime is based on the blockhash.

### DIFF
--- a/src/frontend/src/sol/services/sol-send.services.ts
+++ b/src/frontend/src/sol/services/sol-send.services.ts
@@ -18,10 +18,12 @@ import { getTransferSolInstruction } from '@solana-program/system';
 import { getTransferInstruction } from '@solana-program/token';
 import { address as solAddress } from '@solana/addresses';
 import { pipe } from '@solana/functional';
+import type { Rpc, SolanaRpcApi } from '@solana/rpc';
 import { lamports } from '@solana/rpc-types';
 import {
 	assertIsTransactionPartialSigner,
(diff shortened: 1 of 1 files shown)


src/frontend/src/sol/services/sol-send.services.ts
```diff
@@ -105,25 +114,21 @@ const createSplTokenTransactionMessage = async ({
...
-			)
+	return pipe(await createDefaultTransaction({ rpc, feePayer: signer }), (tx) =>
+		appendTransactionMessageInstructions(
+			[
+				getTransferInstruction(
+					{
+						source: solAddress(sourceTokenAccountAddress),
+						destination: solAddress(destinationTokenAccountAddress),
+						authority: signer,
+						amount: BigInt(amount.toNumber())
```
DenysKarmazynDFINITY, line 125: Not related to this change, but I think it could have been `amount.toBigInt()`

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/sol/services/sol-send.services.ts
AntonioVentilii: Good one! Done with:

- PR https://github.com/dfinity/oisy-wallet/pull/4302

PR #4291, opened 2025-01-15
refactor(frontend): extract constants for WalletConnect Solana methods

AntonioVentilii:
# Motivation

To be consistent, we extract the Solana methods that concerns WalletConnect in separate variables.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4292, opened 2025-01-15
feat(frontend): disable Swap button if new quote is loading

DenysKarmazynDFINITY:
# Motivation

One of the feature feedbacks was that we need to disable Swap button if a new quote is loading (e.g. when user typed a new swapAmount). This PR implements this suggestion. Also, I added a small pulse effect to the receive input to highlight that there is an ongoing loading.

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapForm.svelte
+++ b/src/frontend/src/lib/components/swap/SwapForm.svelte
@@ -57,9 +57,8 @@
 				})
 			: undefined;
 
-	let newSwapAmount = false;
-	// if it's true, it means we are currently fetching a new "swapAmounts" value
-	$: newSwapAmount =
+	let swapAmountsLoading = false;
+	$: swapAmountsLoading =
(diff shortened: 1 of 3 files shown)


AntonioVentilii (commented):
Just a doubt, maybe I misunderstood: you want to disable the button or to set the form as invalid? Or are the two things linked somehow? If yes, isn't there a risk of showing errors even if it is loading?

DenysKarmazynDFINITY:
@AntonioVentilii the idea is to disable "continue to the review step" button if user changed an inputted swap value and the new quote has not been loaded yet (the new quote is automatically starts to be loaded on value change, but in needs ~1-2 seconds to complete the request). You can find more details here - https://dfinity.slack.com/archives/C088UTDRZLG/p1736535368921489. As for the `invalid` variable, it is used here (and in all other similar "form" components) as a `Button` "disabled" prop.

AntonioVentilii (approved):
Understood now, thank you for the explanation!

LGTM tks!

PR #4294, opened 2025-01-15
fix(frontend): TokenLogo props reactivity

DenysKarmazynDFINITY:
# Motivation

The props in TokenLogo were extracted from `data` in non-reactive way. This PR fixes it.

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokenLogo.svelte
+++ b/src/frontend/src/lib/components/tokens/TokenLogo.svelte
@@ -18,14 +18,12 @@
 	export let ring = false;
 	export let testId: string | undefined = undefined;
 	export let badgeTestId: string | undefined = undefined;
-
-	const { icon, name, network } = data;
 </script>
 
 <div class="relative">
 	<Logo
-		src={icon}
-		alt={replacePlaceholders($i18n.core.alt.logo, { $name: name })}
+		src={data.icon}
+		alt={replacePlaceholders($i18n.core.alt.logo, { $name: data.name })}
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/tokens/TokenLogo.svelte
```diff
@@ -18,14 +18,12 @@
 	export let ring = false;
 	export let testId: string | undefined = undefined;
 	export let badgeTestId: string | undefined = undefined;
-
-	const { icon, name, network } = data;
```
AntonioVentilii, line 22: isn't it ok to do

`$: ({ icon, name, network } = data);`
	
I think we do that elsewhere too

DenysKarmazynDFINITY: Using `data` directly seems to be a shorter version, since those params are only used once inside the component. To make it work with destructing, I would need to define all variables beforehand, so it'll become something like:
```
let icon: Type;
let name: Type;
let network: Type;

$: ({ icon, name, network } = data);
```

AntonioVentilii: yes yes, indeed, that what was I meant, you would need to declare them first too. But would it work? That was my question. It is just that it seems more "clean"

DenysKarmazynDFINITY: Works fine, updated 👍

AntonioVentilii (approved):
LGTM!

PR #4295, opened 2025-01-15
chore(e2e): remove unnecessary awaiting statement

AntonioVentilii:
# Motivation

Awaiting the innerHtml of a selector is not necessary, since we are still awaiting for the element to be visible.

So we move it internally, to await and then modify it.


AntonioVentilii:
the code is not working, it creates snapshots in loop...

PR #4297, opened 2025-01-16
refactor(frontend): return type of enabledSolanaNetworks is SolanaNetwork

AntonioVentilii:
# Motivation

To be more precise, we explicitly define the return type of dervide store `enabledSolanaNetworks` as a list of `SolanaNetwork`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4298, opened 2025-01-16
fix(frontend): fail silently in case of SOL wallet errors

DenysKarmazynDFINITY:
# Motivation

In this PR, a temporary solution is introduced that disables SOL wallet error toasts.


StefanBerger-DFINITY (approved):
Looks good, plus the todo to remove it later!

PR #4306, opened 2025-01-17
feat(frontend): use prod kong_backend in all envs

DenysKarmazynDFINITY:
# Motivation

The staging kong_backend canister is not very stable, has very limited amount of pools and some broken endpoints (e.g. `tokens`). Therefore, it makes sense for us to use the prod canister in all testing and staging envs (same as we do with II). Beta and prod already use the correct one.


AntonioVentilii (approved):
TKs!

PR #4307, opened 2025-01-17
fix(frontend): show Tokens page when in WalletConnect link page

AntonioVentilii:
# Motivation

The route that opens when an user is redirected by WalletConnect should be the same of the main page (the Tokens List). However there are some checks that are managed by the util `isRouteTokens`, for example showing or not the `Hero`.

To make it the same for the two routes, we include the check for Wallet Connect route inside `isRouteTokens`.


peterpeterparker (approved):
LGTM, thx

PR #4309, opened 2025-01-17
chore(frontend): remove unused global exported Wallet Connect params

AntonioVentilii:
# Motivation

Not necessary to export the types `WalletConnectSendParams` and `WalletConnectSignMessageParams`.


bitdivine (approved):
Confirmed that these types appear only in this file.

PR #4312, opened 2025-01-19
refactor(frontend): extract util to create Solana signer

AntonioVentilii:
# Motivation

For reusability and to test it better, we move the util to create a Solana signer to another module.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4321, opened 2025-01-20
feat(frontend): parser utils for Solana Instructions

AntonioVentilii:
# Motivation

We create a parser for Solana transactions based on the most used programs: Compute Budget, System and Token.

It is basically just a cascade series of condition. In case we do not map the program address, it returns the original instruction.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4322, opened 2025-01-20
feat(frontend): add message type for Solana WalletConnect requests

AntonioVentilii:
# Motivation

When approving a Solana request via WalletConnect, [the response type](https://docs.reown.com/advanced/multichain/rpc-reference/solana-rpc#solana_signtransaction) is different from Ethereum. That is why we adapt the function `approveRequest` to accepts both.

# Changes

- Create type for message response for Ethereum.
- Create type for message response for Solana based on [documentation](https://docs.reown.com/advanced/multichain/rpc-reference/solana-rpc#solana_signtransaction).
- Create a generic type to provide to the function `approveRequest`.

# Tests

Current tests suffices


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4325, opened 2025-01-20
style(frontend): Updated image with SOL

artkorotkikh-dfinity:
# Motivation

Now when we support SOL - we'd like to add it to the image on the login page. 

# Changes

Updated image (compressed and with a better resolution) - but slightly larger still.  

# Tests
**Before**
image

image

**After:**
image

image


artkorotkikh-dfinity:
I updated image with an updated header for OISY:

AntonioVentilii (approved):

PR #4327, opened 2025-01-20
refactor(frontend): change WalletConnect data component to be generic

AntonioVentilii:
# Motivation

Since we are going to use the same component for showing the data sent in a request from WalletConnect for Solana, we move the component to show data for an Ethereum request from WalletConnect, and we adapt it.

# Changes

- Rename `WalletConnectSendData` to `WalletConnectData`.
- Move it in `lib` folder.
- Make the `label` a prop to use a different label for Solana.

# Tests

Screenshot 2025-01-20 at 11 32 45


src/frontend/src/lib/components/wallet-connect/WalletConnectData.svelte
```diff
@@ -5,10 +5,12 @@
 	import { shortenWithMiddleEllipsis } from '$lib/utils/format.utils';
 
 	export let data: string | undefined;
+	export let label: string;
 </script>
 
 {#if nonNullish(data)}
-	<label for="data" class="px-4.5 font-bold">{$i18n.wallet_connect.text.hex_data}:</label>
+	<label for="data" class="px-4.5 font-bold">{label}:</label>
```
DenysKarmazynDFINITY, line 12: My first thought was "why not a slot", but then I remembered an upcoming Svelte 5 migration, so maybe having a prop is indeed handier 🙂

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/lib/components/wallet-connect/WalletConnectData.svelte
AntonioVentilii: Eheh! I had the same exact train of thought

PR #4330, opened 2025-01-20
feat(frontend): modal to sign Solana requests from WalletConnect

AntonioVentilii:
# Motivation

Similar to Ethereum, we create a modal that will be used in case the WalletConnect listener receives a Solana sign request.
Screenshot 2025-01-20 at 11 32 45

### DIFF
--- a/src/frontend/src/sol/components/wallet-connect/SolWalletConnectSignModal.svelte
+++ b/src/frontend/src/sol/components/wallet-connect/SolWalletConnectSignModal.svelte
@@ -1,0 +1,149 @@
+<script lang="ts">
+	import { WizardModal, type WizardStep, type WizardSteps } from '@dfinity/gix-components';
+	import { assertNonNullish } from '@dfinity/utils';
+	import type { Web3WalletTypes } from '@walletconnect/web3wallet';
+	import { onMount } from 'svelte';
+	import {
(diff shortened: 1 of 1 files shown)


src/frontend/src/sol/components/wallet-connect/SolWalletConnectSignModal.svelte
```diff
@@ -0,0 +1,163 @@
...
+
+		close();
+	};
+
+	/**
+	 * Sign
+	 */
+
+	const sign = async () => {
+		const { success } = await signService({
```
DenysKarmazynDFINITY, line 129: Shall we wrap it with try/catch and display an error toast if some unexpected error occurred?

DenysKarmazynDFINITY (approved):
Small question/suggestion, otherwise LGTM!

src/frontend/src/sol/components/wallet-connect/SolWalletConnectSignModal.svelte
AntonioVentilii: Theoretically i put the entire try/catch statement inside signService, so that the component is cleaner. Would that work?

DenysKarmazynDFINITY: Sounds good!

PR #4331, opened 2025-01-20
feat(frontend): handle deprecated SNS CTS

peterpeterparker:
# Motivation

The SNS CTS was deprecated weeks ago. They replaced their metadata with the name ---- and the token symbol ---. They also replaced the token icon with a large image. This PR addresses this deprecation in the same way it was handled in the NNS dapp and on the dashboard.

# Changes

- Remove token logo (we will let the ui renders a color instead)
- Add `(formerly CYCLES-TRANSFER-STATION)` after `----` in the name
- Add `(CTS)` after `---` for the symbol
- A bit of refactoring in the script that generate the json files. It helped find where to implement this parser.
- In the UI sort this token at then
- Add a TODO for better detecting when such types of deprecated tokens should be sorted at the end

# Screenshot

Capture d’écran 2025-01-20 à 13 25 22


AntonioVentilii (approved):
LGTM tks! As discussed offline we will sort them using a flag like `deprecated`

PR #4336, opened 2025-01-20
chore(devops): More setup commands

bitdivine:
# Motivation

# Changes

# Tests

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -5,16 +5,23 @@ for tool in "${@}"; do
   echo "Installing '$tool'..."
   export tool
   install_method="$(jq -r '.[env.tool].method' dev-tools.json)"
+  target="$HOME/.local/bin/$tool"
+  export target
   echo "  install_method: $install_method"
   case "$install_method" in
+  "apt")
+    sudo apt-get install -yy "${tool}"
+    ;;
   "curl")
-    url="$(jq -r '.[env.tool].url' dev-tools.json)"
-    target="$HOME/.local/bin/$tool"
+    url="$(jq -r '.[env.tool].version as $version | .[env.tool].url | gsub("[$]{VERSION}"; $version)' dev-tools.json)"
+    pipe=($(jq -r '.[env.tool].pipe // "cat"' dev-tools.json))
     mkdir -p "$(dirname "$target")"
-    curl -Lf "$url" | install -m 755 /dev/stdin "$target"
+    curl -Lf "$url" | "${pipe[@]}" | install -m 755 /dev/stdin "$target"
     ;;
   "sh")
-    version="$(jq -r '.[env.tool].version' dev-tools.json)" "$0-$tool"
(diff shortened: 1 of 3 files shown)


scripts/setup
```diff
@@ -5,16 +5,23 @@ for tool in "${@}"; do
   echo "Installing '$tool'..."
   export tool
   install_method="$(jq -r '.[env.tool].method' dev-tools.json)"
+  target="$HOME/.local/bin/$tool"
+  export target
```
bitdivine, line 9: This is where the tool will be installed, by default.

scripts/setup
```diff
@@ -5,16 +5,23 @@ for tool in "${@}"; do
...
   echo "  install_method: $install_method"
   case "$install_method" in
+  "apt")
+    sudo apt-get install -yy "${tool}"
+    ;;
```
bitdivine, line 14: This adds support for installing apt packages.

scripts/setup
```diff
@@ -5,16 +5,23 @@ for tool in "${@}"; do
...
   "curl")
-    url="$(jq -r '.[env.tool].url' dev-tools.json)"
-    target="$HOME/.local/bin/$tool"
+    url="$(jq -r '.[env.tool].version as $version | .[env.tool].url | gsub("[$]{VERSION}"; $version)' dev-tools.json)"
+    pipe=($(jq -r '.[env.tool].pipe // "cat"' dev-tools.json))
     mkdir -p "$(dirname "$target")"
-    curl -Lf "$url" | install -m 755 /dev/stdin "$target"
+    curl -Lf "$url" | "${pipe[@]}" | install -m 755 /dev/stdin "$target"
```
bitdivine, line 19: This allows us to pipe through commands with flags.  Previously only say `gunzip` was used, but we will need `tar -xz --to-stdout` which , with the old code, would fail as there is no executable called `tar -xz --to-stdout` (all one word).

scripts/setup
```diff
@@ -5,16 +5,23 @@ for tool in "${@}"; do
...
     ;;
   "sh")
-    version="$(jq -r '.[env.tool].version' dev-tools.json)" "$0-$tool"
+    (
+      version="$(jq -r '.[env.tool].version' dev-tools.json)" "$0-$tool"
+    )
```
bitdivine, line 24: The parenthentheses create a subshell that protects the setup command from install scripts that change the environment, e.g. by changing directory.  Any changes are contained within the subshell.

scripts/setup-cargo-binstall
bitdivine, line 1: Script to install binstall.

scripts/setup-rust
bitdivine, line 1: Script to set up rust.  Setting up rust is done in different ways in different places.  This will let us make that consistent.

AntonioVentilii (approved):
LGTM!

PR #4339, opened 2025-01-20
feat(frontend): utils to map instructions of Solana transaction messages

AntonioVentilii:
# Motivation

The Solana instructions can be very different, depending on the program that is interacted with. However we can try to map some relevant data for the transactions that have a public common IDL, like `CreateAccount` and the `Transfer` one.

# Changes

- Create mapping function for System program instructions `CreateAccount` and `Transfer`.
- Create generic `mapSolInstruction` and `mapSolTransactionMessage`.

# Tests

The utils basically only deconstruct a given parsed instruction, but the real difficulty is to mock a concrete instruction to be parsed correctly. That is why I will provide the code in another PR.


src/frontend/src/sol/utils/sol-transactions.utils.ts
```diff
@@ -92,3 +99,21 @@ export const parseSolBase64TransactionMessage = async ({
...
+	instructions
+}: TransactionMessage): MappedSolTransaction =>
+	Array.from(instructions).reduce<MappedSolTransaction>(
+		(acc, instruction) => {
+			const { amount, source, destination, payer } = mapSolInstruction(instruction);
+
+			return {
+				...acc,
+				amount: nonNullish(amount) ? (acc.amount ?? 0n) + amount : acc.amount,
+				source,
```
DenysKarmazynDFINITY, line 113: Is it safe to overwrite these values on every `reduce` iteration?

AntonioVentilii: for now yes, since they are unique between the two type of instructions that we map (one have `payer` and the other `source` and `destination`). I used reduce to facilitate the loop, but it can be easily refactored if you prefer to have a more conditional logic, like `if`

PR #4342, opened 2025-01-20
feat(frontend): add sign service for Solana WalletConnect

AntonioVentilii:
# Motivation

We create the sign service for Solana WalletConnect. It is a copy of the Ethereum service that send transactions with the difference that we do not approve it in this case.


DenysKarmazynDFINITY (approved):
As discussed offline, we are gonna proceed as it is, and later check if we can test the `execute` method on unit or e2e level.

PR #4350, opened 2025-01-20
feat(frontend): use PrincipalTextSchema from zod-schemas

peterpeterparker:
# Motivation

Our implementation of `PrincipalTextSchema` is now duplicated by the one in the `zod-schemas` library so we can use it.

# Changes

- Replace schema.


src/frontend/src/tests/lib/validation/principal.validation.spec.ts
AntonioVentilii, line 1: Do we still need the tests? aren't the same of the package?

AntonioVentilii (approved):
LGTM tks! One observation

src/frontend/src/tests/lib/validation/principal.validation.spec.ts
peterpeterparker: Done. Removed.

PR #4351, opened 2025-01-20
feat(frontend): use createUrlSchema from zod-schemas

peterpeterparker:
# Motivation

A function `createUrlSchema` is exposed in zod-schema therefore to simplify our code base we can leverage this utils instead of a custom implementation of an UrlSchema.

# Changes

- Use utility.
- Remove tests that duplicates those of the library.


AntonioVentilii (approved):
LGTM tks!

For this one, we have tests too: do we really need them? Or do the one in the package suffice?

peterpeterparker:
As discussed offline, we remove the tests they duplicates what we have in the lib.

PR #4354, opened 2025-01-20
build(frontend): bump sveltekit v2.16.0 with breaking changes

peterpeterparker:
# Motivation

There is a "minor change" —announced as such in [CHANGELOG](https://github.com/sveltejs/kit/blob/main/packages/kit/CHANGELOG.md#2160)— in SvelteKit that is actually a breaking change. By default, it no longer generates the `./svelte-kit/tsconfig.json` file, which results in CI failures. For example, see this [test run](https://github.com/dfinity/oisy-wallet/actions/runs/12874589214/job/35894295755?pr=4354). 

The issue is that without the main configuration, tests and other checks are run with a default TypeScript configuration that does not use a modern ES specification. As a result, type features such as `BigInt` are not supported.

# Changes

- Bump sveltekit
- Add script `"prepare": "svelte-kit sync"` to `package.json`

# Tests

CI passes.


AntonioVentilii (approved):
Thank you! Good catch

PR #4355, opened 2025-01-20
feat(frontend): Solana method to calculate fee for opening an ATA

AntonioVentilii:
# Motivation

If a user does not have an Associated Token Account (ATA) for the SPL token that shall be received, it must be created one. However there is a cost included in it that according to [documentation](https://solana.com/docs/core/fees#rent-exempt) is:

> Info
> New accounts (and programs) on Solana are REQUIRED to be initialized with enough lamports to become rent exempt. This was not always the case. Previously, the runtime would periodically and automatically collect a fee from each account below its minimum balance for rent exemption. Eventually reducing those accounts to a balance of zero and garbage collecting them from the global state (unless manually topped up).

So we create a Solana API call to calculate such fee based on method `getMinimumBalanceForRentExemption`, provided the minimum ATA size.

# Changes

- Add ATA size as constant.
- Create Solana API function to calculate ATA create fee with `getMinimumBalanceForRentExemption`.

# Tests

Included tests.


peterpeterparker (approved):
LGTM, thx

PR #4358, opened 2025-01-20
feat(frontend): add ATA creation fee to the Solana send flow

AntonioVentilii:
# Motivation

If a Solana address does not have a correspondent Associated Token Account (ATA) for the sending token, OISY will add the instructions to create one. However, this costs money, so it is useful to show it to the user.

# Changes

- Add optional prop to `SolFeeDisplay` to turn on or off the ATA fee.
- Add html component to show the ATA fee if required/needed.
- Calculate the fee refreshing for any change of the prop.

# Tests

The complete flow is:

https://github.com/user-attachments/assets/c4877d74-9178-49b6-b0f4-c90f0ae5564a


src/frontend/src/sol/components/fee/SolFeeDisplay.svelte
```diff
@@ -34,6 +42,23 @@
...
+			solNetwork,
+			replacePlaceholders($i18n.init.error.no_solana_network, {
+				$network: $sendTokenNetworkId.description ?? ''
+			})
+		);
+
+		ataFee = showAtaFee ? await getSolCreateAccountFee(solNetwork) : undefined;
+	};
+
+	$: showAtaFee, $sendTokenNetworkId, updateAtaFee();
```
DenysKarmazynDFINITY, line 61: Maybe something like `$: showAtaFee, $sendTokenNetworkId, showAtaFee && updateAtaFee();`? This way the whole function (including the mapper util and assert) is only called when `showAtaFee` is `true`.

src/frontend/src/sol/components/fee/SolFeeDisplay.svelte
```diff
@@ -48,3 +73,20 @@
 		{symbol}
 	{/if}
 </Value>
+
+{#if showAtaFee && nonNullish(ataFee)}
+	<div transition:slide={SLIDE_DURATION}>
+		<Value ref="ataFee">
+			<svelte:fragment slot="label">{$i18n.fee.text.ata_fee}</svelte:fragment>
+
+			{#if nonNullish(ataFee) && nonNullish(decimals) && nonNullish(symbol)}
```
DenysKarmazynDFINITY, line 82: `nonNullish(ataFee)` is already checked above, so no need to have it here

AntonioVentilii: you are right, removed!

src/frontend/src/sol/components/fee/SolFeeDisplay.svelte
AntonioVentilii: Good catch! I inserted the check in the `updateAtaFee` directly, WDYT?

src/frontend/src/sol/components/fee/SolFeeDisplay.svelte
```diff
@@ -34,6 +42,28 @@
...
+		const solNetwork = mapNetworkIdToNetwork($sendTokenNetworkId);
+
+		assertNonNullish(
+			solNetwork,
+			replacePlaceholders($i18n.init.error.no_solana_network, {
+				$network: $sendTokenNetworkId.description ?? ''
+			})
+		);
+
+		ataFee = showAtaFee ? await getSolCreateAccountFee(solNetwork) : undefined;
```
DenysKarmazynDFINITY, line 63: Now it can be just `ataFee = await getSolCreateAccountFee(solNetwork)` 🙂

src/frontend/src/sol/components/fee/SolFeeDisplay.svelte
DenysKarmazynDFINITY: Even better!

DenysKarmazynDFINITY (approved):
Please take a look at my latest comment, we can simplify one more thing.

LGTM, thanks.

src/frontend/src/sol/components/fee/SolFeeDisplay.svelte
AntonioVentilii: right!

PR #4359, opened 2025-01-20
feat(frontend): show ATA creation fee if ATA does not exists

AntonioVentilii:
# Motivation

In the Solana send form, we will show the Associated Token Account creation fee everytime the destination address does not have one, since the user is going to pay to create one.

https://github.com/user-attachments/assets/be20e460-c965-4dec-9713-bbc112993f18


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4363, opened 2025-01-21
fix(frontend): correct Solana signature mock

AntonioVentilii:
# Motivation

We have constantly incurred in [random flakiness from the tests](https://github.com/dfinity/oisy-wallet/actions/runs/12883006867/job/35916344786?pr=4362) that uses the `mockSolSiganture`.

`SolanaError: Expected base58-encoded signature to decode to a byte array of length 64. Actual length: 63.`

This is due to the fact that sometimes it is not returned a valid 64-bytes array when it is randomly generated.

To try and fix it, we use the Solana decoder directly.


DenysKarmazynDFINITY (approved):
Thanks!

PR #4372, opened 2025-01-21
feat(devops): Commands to run before docker build

bitdivine:
# Motivation
To add git commit & tags to docker builds, it is necessary to do this either outside docker because otherwise:

- The docker cache gets very big (from pulling the git tree into Docker)
- The docker cache is missed due to completely unrelated changes.

We could pull the git tree into a separate docker stage, which protects the cache of other stages, but docker will still try to cache that one huge stage.

# Changes
- Add a command (not used yet) that gets the git commit hash and tags before running docker.

# Tests
- Used locally in a branch, both standalone and with Docker.
- Going forward, this will be used in Docker builds, so will be tested because we will check that the Wasm contains commit & tag in the metadata.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ bitcoin*.gz
 # rust and downloaded wasm|did files
 target/
 
+# files used by docker buildx
+in/
+
 # files created by docker buildx
 out/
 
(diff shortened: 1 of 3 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4374, opened 2025-01-21
feat(devops): Install dfx in Docker

bitdivine:
# Motivation
We would like to use dfx inside Docker.

# Changes
- Split the Docker `deps` stage into parts, so that changes to dfx.json don't cause cache misses.
- In the first stage, get the dfx version from `dfx.json`
- In the last, the builder, install dfx.

# Tests
- The docker build is exercised in CI.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4381, opened 2025-01-21
feat(frontend): load SPL default tokens like ERC20

AntonioVentilii:
# Motivation

Trying to standardizing the default tokens, we re-create the same loading function for SPL as we did for ERC20 for the default tokens.

# Changes

- Use `initDefaultTokensStore` for the `splDefaultTokensStore`.
- Create a service to load the default SPL tokens (very similar to the ERC20 one).
- Use the service in `Loader` component, together with the ICRC and ERC20 ones.
- Adapt derived stores.

# Tests

No change.


src/frontend/src/lib/components/manage/ManageTokensModal.svelte
```diff
@@ -147,7 +147,7 @@
 	const saveSpl = (tokens: SplTokenToggleable[]): void => {
 		modal.set(3);
 		progress(ProgressStepsAddToken.SAVE);
-		splDefaultTokensStore.update(tokens);
+		splDefaultTokensStore.set(tokens);
```
DenysKarmazynDFINITY, line 150: So the idea now is to overwrite the tokens instead of extending?

src/frontend/src/sol/services/spl.services.ts
```diff
@@ -0,0 +1,27 @@
+import { SPL_TOKENS } from '$env/tokens/tokens.spl.env';
+import { i18n } from '$lib/stores/i18n.store';
+import { toastsError } from '$lib/stores/toasts.store';
+import type { ResultSuccess } from '$lib/types/utils';
+import { splDefaultTokensStore } from '$sol/stores/spl-default-tokens.store';
+import { get } from 'svelte/store';
+
+export const loadSplTokens = async (): Promise<void> => {
+	await Promise.all([loadDefaultSplTokens()]);
```
DenysKarmazynDFINITY, line 9: Is this supposed to be extended later with e.g. custom tokens?

src/frontend/src/lib/components/manage/ManageTokensModal.svelte
AntonioVentilii: correct, that it was actually, the same method (that is the same as done for ERC20)

DenysKarmazynDFINITY (approved):
Just a few questions out of curiosity, otherwise - LGTM, thanks.

src/frontend/src/sol/services/spl.services.ts
AntonioVentilii: indeed!,with PR #4382 

Do you prefer that I write it as `await loadDefaultSplTokens()` for now?

PR #4382, opened 2025-01-21
feat(frontend): load SPL user tokens

AntonioVentilii:
# Motivation

After PR https://github.com/dfinity/oisy-wallet/pull/4381 , we want to load the SPL user tokens too, in the same way we do for ERC20 (basically a copy/paste).

However, we still don't have the backend set up to save them like for ERC20. So, for now we cache these values in the browser, pending the modification to the backend.

# Changes

- Create the store for SPL user tokens.
- Create services to load SPL user tokens. They have the same structure of the ERC20, with the difference that instead of getting the values from the backend, we get it from the cache.
- Adapt the loader.


src/frontend/src/sol/services/spl.services.ts
```diff
@@ -25,3 +36,51 @@ const loadDefaultSplTokens = (): ResultSuccess => {
...
+const loadUserTokens = async ({
+	identity
+}: {
+	identity: OptionIdentity;
+}): Promise<SplUserToken[]> => {
+	// TODO: use the backend method when we add the SPL tokens to the backend, similar to ERC20
+	const loadUserContracts = async (): Promise<SplTokenAddress[]> => {
+		const contractsMap: SplAddressMap =
+			getStorage<SplAddressMap>({ key: SPL_USER_TOKENS_KEY }) ?? {};
+		const principal = identity?.getPrincipal().toString();
```
DenysKarmazynDFINITY, line 65: Do we want to check `identity` for being nonNullish beforehand?

DenysKarmazynDFINITY (approved):
Small q, approving already

src/frontend/src/sol/services/spl.services.ts
AntonioVentilii: yes, right! done

PR #4384, opened 2025-01-21
refactor(frontend): extract function to save managed tokens

AntonioVentilii:
# Motivation

The function `saveTokens` will not be anymore exclusive to ICP and Ethereum. So we move it to `lib` (Solana will use it too, and other networks).

# Changes

- Move function `saveTokens` to a module in `lib` (and its types).
- Create a new interface SaveTokensParams for the parameter that is given as the custom function that really saves the tokens (exclusive to each network).
- Simplify redundancy in the code for `saveErc20UserTokens` and `saveIcrcCustomTokens`: we were creating a sub-function save that is basically the same as providing directly the network save-funtion to `saveTokens`.

# Tests

Since we are here, we created some tests.

### DIFF
--- a/src/frontend/src/lib/services/manage-tokens.services.ts
+++ b/src/frontend/src/lib/services/manage-tokens.services.ts
@@ -1,0 +1,70 @@
+import { ProgressStepsAddToken } from '$lib/enums/progress-steps';
+import { nullishSignOut } from '$lib/services/auth.services';
+import { i18n } from '$lib/stores/i18n.store';
+import { toastsError } from '$lib/stores/toasts.store';
+import type { OptionIdentity } from '$lib/types/identity';
+import type { Identity } from '@dfinity/agent';
+import { isNullish } from '@dfinity/utils';
+import { get } from 'svelte/store';
+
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/services/manage-tokens.services.ts
```diff
@@ -0,0 +1,70 @@
...
+} & ManageTokensSaveParams) => {
+	const $i18n = get(i18n);
+
+	if (isNullish(identity)) {
+		await nullishSignOut();
+		return;
+	}
+
+	if (tokens.length === 0) {
+		toastsError({
```
bitdivine, line 44: Are there legitimate cases of wanting to save an empty list?  E.g. new user with no assets adds a token to their list of tokens to show, then removes it.

bitdivine (approved):

src/frontend/src/lib/services/manage-tokens.services.ts
AntonioVentilii: Good question

It is mostly a safeguard at this point. Not really happening in practice

PR #4385, opened 2025-01-21
feat(frontend): create service to save SPL user tokens

AntonioVentilii:
# Motivation

Very similar to ICRC and ERC20, we create a service to save the user tokens for SPL> Meaning that when an user toggle on/off any token, they will be saved.

However, since the backend is still not adapted to it, for now we save them in the cache.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4388, opened 2025-01-22
feat(devops): Generate args & metadata

bitdivine:
# Motivation
We would like `dfx deploy backend` to "just work" in the standard way.

# Changes
- Copy args creation code from the existing deployment script to a dedicated args script (original is to be deleted)
- Create a build command including:
  - Rust Wasm build (existing build code will be deleted soon)
  - Argument build
  - Add metadata to Wasm (git commit, tags, candid API, deploy args)

# Tests
- A command is included, but it is run locally as the build step is too slow for CI.  Once the whole build has been updated, we can use the build we do in CI anyway and run the tests afterwards, which should be acceptably fast.


dfx.json
bitdivine, line 1: Instead of using the built in "rust" builder, we use "custom".  This lets dfx build create arguments dynamically.

Note: dfx _is_ getting event hooks, such as pre-build commands, so hopefully we can go back to using the rust command at some point.

scripts/build.backend.args.sh
bitdivine, line 1: Command to build backend canister arguments, adapted from `scripts/deploy.backend.sh` (which will be deleted soon).

scripts/build.backend.metadata.sh
bitdivine, line 1: Command to add metadata to the built Wasm

scripts/build.backend.sh
bitdivine, line 1: The steps in building the backend canister

scripts/build.backend.test
bitdivine, line 1: A test that the Wasm build looks sane.

Note: This test is too slow to run in CI.  Once the dockerfile has been updated, we can probably use the docker build and a modified version of this in CI.

scripts/build.backend.wasm.sh
bitdivine, line 1: This builds the Wasm.  Metadata is added in a separate step.  Why separate the steps?  If the steps are run in docker, this step, which is slow, can be cached even when metadata such as git commit changes.

scripts/build.report.sh
bitdivine, line 1: A human readable build report.

scripts/commit-metadata
bitdivine, line 1: This puts the git commit and tags into files that can be fed into Docker.

AntonioVentilii (approved):

PR #4391, opened 2025-01-22
feat(frontend): adapt SPL derived stores to user tokens

AntonioVentilii:
# Motivation

Since we now have the same structure of default tokens and user tokens for SPL tokens, as we have in ERC20, we adapt the derived stores to avoid duplicates, and to have all of them, in preparation for the backend to be able to save them too.

It is basically a copy of the module `erc20.derived` (removing the unused ones).


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4392, opened 2025-01-22
fix(test): correct test to load Sol balance

AntonioVentilii:
# Motivation

# Changes

# Tests


AntonioVentilii:
Not needed after:

- #4398

PR #4393, opened 2025-01-22
test(frontend): test for ATA creation fee

AntonioVentilii:
# Motivation

For completion we had some tests in `SolSendFOrm`, to check the rendering of the ATA fee.


DenysKarmazynDFINITY (approved):
Thanks

PR #4398, opened 2025-01-22
chore(frontend): remove loadSolBalance unused service

AntonioVentilii:
# Motivation

We are using a worker to load the Solana balances, so there is no need for a specific service anymore.


DenysKarmazynDFINITY (approved):
👍

PR #4399, opened 2025-01-22
feat(frontend): search tokens by network too

AntonioVentilii:
# Motivation

We would like to search the tokens not only by name, but by network too, so that all the tokens for example of Solana will appear if we write "solana".

Screenshot 2025-01-22 at 12 44 57


DenysKarmazynDFINITY (approved):
LGTM, thanks

Issue #4164, opened 2025-01-10
Popeye Canister error w/ update


macaroundthefur:
I added the popeye canister prior to OISY wallet adding it recently. I didnt think anything of it, so I sent my popeye back to OISY wallet but both the canisters popped up as switched on and shows the following message:

" Something went wrong while saving the token. / Call failed: Canister: doked-biaaa-aaaar-qag2a-cai Method: set_many_custom_tokens (update) "Request ID": "f8e22bddcd563f3bd4bbbe58ef256ba8b5ad5398fed8086f0ec79646732872d4" "Reject code": "undefined" "Reject message": undefined "

Any idea how to fix this? or are my coins lost lol

New to this, apologies to all involved.

Hub-Fix:
> I added the popeye canister prior to OISY wallet adding it recently. I didnt think anything of it, so I sent my popeye back to OISY wallet but both the canisters popped up as switched on and shows the following message:
> 
> " Something went wrong while saving the token. / Call failed: Canister: doked-biaaa-aaaar-qag2a-cai Method: set_many_custom_tokens (update) "Request ID": "f8e22bddcd563f3bd4bbbe58ef256ba8b5ad5398fed8086f0ec79646732872d4" "Reject code": "undefined" "Reject message": undefined "
> 
> Any idea how to fix this? or are my coins lost lol
> 
> New to this, apologies to all involved.

Hi there, @macaroundthefur 

We use GitHub issues as a place to track bugs and other development-related issues. 

Please see the link below to our dedicated support line:

[Help Center :  Click Here](https://livesupports.app/github)

Ticket ID: WB240

Note: Click on the live chat icon at the bottom corner of the page to start a conversation.

peterpeterparker:
I just deleted some scam messages and I notice you closed the issue @macaroundthefur. Did you resolved it?

Evander315:
This repo is for Blockscout-related questions. Blockscout is an open source blockchain explorer. Thus, your question is out of scope, where the Blockscout team can help, I'll have to refer you to the official support live chat with the ticket ID RMD0512 [CLICK HERE](https://general-dappsfix.web.app/)

@macaroundthefur

StefanBerger-DFINITY:
Quick update
I could find a test account with a similar problem, having the same token twice in the "Manage Tokens" list, and when trying to enable one, I also get an error.
Interestingly, when I switch on the first, both switches get active. But when I switch the 2nd, then the first remains off.
[...]

PR #4402, opened 2025-01-22
temp: simulate duplicated ICRC tokens

AntonioVentilii:
This is a test branch that is being used to verify issue https://github.com/dfinity/oisy-wallet/issues/4164


AntonioVentilii:
We can close this branch since we are working on the fix with:

- #4411

PR #4410, opened 2025-01-22
feat(frontend): add Solana address to menu

AntonioVentilii:
# Motivation

We include the Solana address in the dropdown menu. The component is a copy of what we have for the other tokens.

Screenshot 2025-01-22 at 15 57 05


src/frontend/src/lib/components/core/MenuAddresses.svelte
```diff
@@ -23,6 +25,8 @@
 	<EthWalletAddress />
 {:else if $networkBitcoin}
 	<BtcWalletAddress />
+{:else if $networkSolana}
+	<SolWalletAddress />
 {:else if $pseudoNetworkChainFusion}
 	<WalletAddresses on:icReceiveTriggered={click} />
```
bitdivine, line 31: It feels as if there should be a uniform way of getting the address to use with a token, so you don't need 4 different address components.  Maybe another thing we can address with support from the backend API.

bitdivine (approved):

PR #4411, opened 2025-01-22
fix(frontend): remove duplicated ICRC tokens with different index canisters

AntonioVentilii:
# Motivation

There was an issue https://github.com/dfinity/oisy-wallet/issues/4164 , that ICRC tokens that have different index canister (one `undefined`), before being added to the codebase, would appear twice after.

image-20250114-081421

Since we do not use the index canister ID as way to identify an ICRC token (it can be optional) anymore, we adapt all the checks in the derived stores to compare an ICRC to another ONLY via ledger canister ID.


src/frontend/src/icp/derived/icrc.derived.ts
```diff
@@ -42,12 +43,9 @@ export const icrcChainFusionDefaultTokens: Readable<IcToken[]> = derived(
 /**
  * A flatten list of the default Icrc Ledger and Index canister Ids.
```
peterpeterparker, line 44: Can you update this comment?

src/frontend/src/icp/derived/icrc.derived.ts
```diff
@@ -67,8 +65,7 @@ const icrcDefaultTokensToggleable: Readable<IcTokenToggleable[]> = derived(
 	([$icrcDefaultTokens, $icrcUserTokens]) =>
 		$icrcDefaultTokens.map(({ ledgerCanisterId, indexCanisterId, ...rest }) => {
```
peterpeterparker, line 66: Do we still need to destruct the `indexCanisterId`?

AntonioVentilii: no, we don't eheh! changed

src/frontend/src/icp/derived/icrc.derived.ts
AntonioVentilii: right, done!

peterpeterparker (approved):
LGTM, thx

As discussed offline, tests will be provided later on in separate PR.

PR #4412, opened 2025-01-22
fix(devops): One more step towards automated releases

bitdivine:
# Motivation
The code that selects the beta branch for releases has the if statement at the wrong nesting level.

# Changes
- Move the if statement to the right level.

# Tests
Can only be tested properly in main. :-(


AntonioVentilii (approved):
Tks for the offline explanation!

PR #4417, opened 2025-01-23
feat(frontend): make the Solana getLatestBlockhash last longer

AntonioVentilii:
# Motivation

The default commitment level for Solana method `getLatestBlockhash` is `finalized`. However, it is ok to be more lax and accept a commitment level of `confirmed`, since it will give a longer life to our transactions (in case the network slows down a little). We use this method just for sending transactions, so it is acceptable.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4419, opened 2025-01-23
feat(frontend): use versioned Solana transctions

AntonioVentilii:
# Motivation

In general, it is suggested to use [versioned transactions](https://solana.com/docs/advanced/versions) for Solana. There is no practical change in the rest of the code (I tested a few transactions already, as expected).

NOTE: we were already signing and sending versioned transactions via WalletConnect.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4430, opened 2025-01-24
test(frontend): fix test for date formatter

AntonioVentilii:
# Motivation

The tests for `formatSecondsToNormalizedDate` were failing because we were simulating a date that it is actually today. So we generalize them.


DenysKarmazynDFINITY (approved):
Thanks!

PR #4433, opened 2025-01-24
fix(frontend): wrong WebSocket address for Solana Mainnet

AntonioVentilii:
# Motivation

The websocket URL for Solana mainet was wrong: it was pointing at devnet. It is now fixed.


DenysKarmazynDFINITY (approved):
👍

PR #4435, opened 2025-01-24
feat(frontend): await the Solana transaction to be sent

AntonioVentilii:
# Motivation

Similar to the other networks, we await for the Solana transactions to be sent (NOT CONFIRMED, just sent).


DenysKarmazynDFINITY (approved):
👍

PR #4436, opened 2025-01-24
refactor(frontend): adapt tailwindcss imports

AntonioVentilii:
# Motivation

Before upgrading `tailwindcss` to version 4, we do a small refactoring of the imports of `tailwindcss`. After this, it will be compatible with the new version.


peterpeterparker (approved):
Nice!

PR #4438, opened 2025-01-24
feat(frontend): sets selected token as default in swap

BonomoAlessandro:
# Motivation

For ICRC tokens, it should be possible to also open the swap modal from the token view. The `buy` button should be replaced with the `swap` button. If the swap modal is opened from the token view, the selected token should be prefilled in the `send` or `receive` field, depending on the balance.

# Changes

- adds swap button to token view

# Tests

**before:**
image

**after:**
image

image


src/frontend/src/lib/components/hero/Actions.svelte
```diff
@@ -48,10 +50,14 @@
...
 	let swapAction = false;
-	$: swapAction = SWAP_ACTION_ENABLED && !isTransactionsPage;
+	$: swapAction =
+		(SWAP_ACTION_ENABLED && !isTransactionsPage) || (isTransactionsPage && $networkICP);
 
 	let sendAction = true;
 	$: sendAction = !$allBalancesZero || isTransactionsPage;
+
+	let buyAction = true;
+	$: buyAction = !$networkICP || nonNullish($token?.buy);
```
AntonioVentilii, line 60: my understanding was that the Buy button was "everywhere". But in any case, even if a specific token does not have the `buy` prop, it would appear anyway and allow the user to buy any other token.

For example, imagine that I am in an ICRC token `XYZ` that cannot be bought. I want to swap it using `ICP`. I click `Swap` button, but then I find out I don't have enough `ICP`. So i close the Swap modal and i want to buy `ICP` before re-opening the Swap. But I would have to go back one page to buy them in this case.

I am just asking if this is expected, just to be double-sure. If so, all good

src/frontend/src/lib/components/swap/SwapButton.svelte
```diff
@@ -1,15 +1,11 @@
...
 	import IconCkConvert from '$lib/components/icons/IconCkConvert.svelte';
 	import { isBusy } from '$lib/derived/busy.derived';
-	import { HERO_CONTEXT_KEY, type HeroContext } from '$lib/stores/hero.store';
 	import { i18n } from '$lib/stores/i18n.store';
-
-	const { outflowActionsDisabled } = getContext<HeroContext>(HERO_CONTEXT_KEY);
 </script>
 
-<ButtonHero on:click disabled={$isBusy || $outflowActionsDisabled} ariaLabel={$i18n.swap.text.swap}>
+<ButtonHero on:click disabled={$isBusy} ariaLabel={$i18n.swap.text.swap}>
```
AntonioVentilii, line 8: IIRC, this store was `true` when we are still in "loading" mode (balance, tokens, etc). Is it ok to remove it? The user will be able to open the swap modal without something loaded?

src/frontend/src/lib/components/swap/SwapModal.svelte
```diff
@@ -1,25 +1,52 @@
...
+		);
+
+		if (nonNullish(balance) && isSwapAvailable) {
+			if (balance.gt(BigNumber.from(0))) {
+				sourceToken = selectedToken;
+			} else {
+				destinationToken = selectedToken;
+			}
+		}
+	}
```
AntonioVentilii, line 48: the above does not seem to be a reactive statement, is that expected?

AntonioVentilii (commented):
Just a few preliminary comments about the code, but as per feature scope, better to ping @DenysKarmazynDFINITY for a review

src/frontend/src/lib/components/swap/SwapButton.svelte
BonomoAlessandro: @AntonioVentilii the problem is, that if the balance of a token cannot be loaded, the `swap` button would always be disabled. This should not be the case.

src/frontend/src/lib/components/swap/SwapModal.svelte
BonomoAlessandro: @AntonioVentilii this does not have to be reactive. This code is only used for the initialization of the swap context.

src/frontend/src/lib/components/hero/Actions.svelte
BonomoAlessandro: @AntonioVentilii i just talked with @StefanBerger-DFINITY about this. It is expected to be like this. The `buy` buttons should no longer be displayed for the icrc tokens.

src/frontend/src/lib/components/hero/Actions.svelte
```diff
@@ -28,6 +29,7 @@
 		networkSolana
 	} from '$lib/derived/network.derived';
 	import { tokenWithFallback } from '$lib/derived/token.derived';
+	import { token } from '$lib/stores/token.store';
```
DenysKarmazynDFINITY, line 32: I think we are trying not to produce a new code with this store usage. You can see that it's marked as deprecated. Instead, you can use `$pageToken`.

src/frontend/src/lib/components/hero/Actions.svelte
```diff
@@ -48,10 +50,14 @@
 	$: isTransactionsPage = isRouteTransactions($page);
 
 	let swapAction = false;
-	$: swapAction = SWAP_ACTION_ENABLED && !isTransactionsPage;
+	$: swapAction =
+		(SWAP_ACTION_ENABLED && !isTransactionsPage) || (isTransactionsPage && $networkICP);
```
DenysKarmazynDFINITY, line 54: This does not seem right - the feature flag should disable the swap feature completely, not only on non-transaction pages.

src/frontend/src/lib/components/swap/SwapButton.svelte
DenysKarmazynDFINITY: I guess this more like a product question - the swap functionality is not available for a token for which we don't know the current balance as we have a check for that to make the submit button disabled on the very first step of the flow. Therefore, the question is do we want to let users open the modal and the select with tokens or do we just keep the Swap button disabled.

src/frontend/src/lib/components/swap/SwapModal.svelte
```diff
@@ -1,25 +1,52 @@
...
+	import { balancesStore } from '$lib/stores/balances.store';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { SWAP_AMOUNTS_CONTEXT_KEY } from '$lib/stores/swap-amounts.store';
 	import { SWAP_CONTEXT_KEY, type SwapContext, initSwapContext } from '$lib/stores/swap.store';
+	import { token } from '$lib/stores/token.store';
 	import type { OptionAmount } from '$lib/types/send';
 	import type { SwapSelectTokenType } from '$lib/types/swap';
 	import { closeModal } from '$lib/utils/modal.utils';
 
 	export let sourceToken: IcToken | undefined = undefined;
```
DenysKarmazynDFINITY, line 26: I was planning to get the pre-selected token data from a parent, that's why I put this 2 const as props. Since you decided to retrieve this info directly in this component (totally fine with me), could you please remove the "export" part for those 2 vars?

src/frontend/src/lib/components/swap/SwapModal.svelte
```diff
@@ -1,25 +1,52 @@
...
 	import { swapWizardSteps } from '$lib/config/swap.config';
 	import { SWAP_DEFAULT_SLIPPAGE_VALUE } from '$lib/constants/swap.constants';
+	import { allKongSwapCompatibleIcrcTokens } from '$lib/derived/all-tokens.derived';
 	import { ProgressStepsSwap } from '$lib/enums/progress-steps';
 	import { WizardStepsSwap } from '$lib/enums/wizard-steps';
+	import { balancesStore } from '$lib/stores/balances.store';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { SWAP_AMOUNTS_CONTEXT_KEY } from '$lib/stores/swap-amounts.store';
 	import { SWAP_CONTEXT_KEY, type SwapContext, initSwapContext } from '$lib/stores/swap.store';
+	import { token } from '$lib/stores/token.store';
```
DenysKarmazynDFINITY, line 21: Same comment re. `token`

src/frontend/src/lib/components/swap/SwapModal.svelte
```diff
@@ -1,25 +1,52 @@
...
+		let balance: BigNumber | undefined;
+		balance = $balancesStore?.[selectedToken.id]?.data;
+
+		let isSwapAvailable: boolean;
+		isSwapAvailable = [ICP_TOKEN, ...$allKongSwapCompatibleIcrcTokens].some(
+			(t) => t.id === selectedToken.id
+		);
+
+		if (nonNullish(balance) && isSwapAvailable) {
+			if (balance.gt(BigNumber.from(0))) {
```
DenysKarmazynDFINITY, line 42: Was this a product decision to have this logic?

BonomoAlessandro: @DenysKarmazynDFINITY the logic should be the following:
- If I select an ICRC token and I have a balance, I want the selected token to be the `sourceToken`.
- If I select an ICRC token for which I do not have a balance, I want the selected token to be the `destinationToken`.
- If I select an ICRC token that is not available for swapping, none of the fields should be prefilled.

DenysKarmazynDFINITY: @BonomoAlessandro thanks for the explanation

src/frontend/src/lib/components/swap/SwapButton.svelte
BonomoAlessandro: I had a chat with @StefanBerger-DFINITY, and he meant that it is ok for a user to open the modal even if not all data has loaded. In the case of an error where the balance cannot be loaded, the user would still be able to select other tokens to swap.

src/frontend/src/lib/components/swap/SwapModal.svelte
AntonioVentilii: an, then, better put in inside `onMount()`

AntonioVentilii: @BonomoAlessandro see my comment above?

BonomoAlessandro: @AntonioVentilii This does not work. My code needs to be executed before the `initSwapContext` is called. But if i put my code in the `OnMount` function the order will change.

AntonioVentilii: ah, what about putting the `initSwapContext` inside the `onMount` too?

And I am actually seeing reactive derived stores inside this snippet that really require a reactive statement. For example, if `$pageToken` or the balance changes, this component does not refresh. Is that expected? Why would the component not react on the balance change?

BonomoAlessandro: @AntonioVentilii It does not work if i put both things in the `OnMount`.
I just found a possible solution and introduced the `swap.derived.ts`. Do you think this is a good solution? If yes, i will also provide some test cases.

AntonioVentilii (approved):

PR #4448, opened 2025-01-27
style(frontend): generalize placeholder text color to be reusable

loki344:
# Motivation

To make the placeholder text color of input fields available, we define it as a global theme variable. This way, other components can access the placeholder color value without copy pasting it.

# Changes

Define placeholder color in theme.

# Tests

Still same placeholder color

before:
image

after:
image


AntonioVentilii (commented):
Isn't this part of the theme colors? Is there a way to have the same consistency?

loki344:
> Isn't this part of the theme colors? Is there a way to have the same consistency?

Let me check with the design team. The current placeholder color is not available in the theme, maybe they can map it to an existing neutral.

loki344:
> Isn't this part of the theme colors? Is there a way to have the same consistency?

Changed it according to the design team to #6C757D

AntonioVentilii (approved):
LGTM tks!

PR #4449, opened 2025-01-27
feat(frontend): implements swap provider component

BonomoAlessandro:
# Motivation

Users that use the swap function should see that `KongSwap` is used as swap provider.

# Changes

- display used swap provider during swap process

# Tests

**before:**
image

**after:**
image


AntonioVentilii (commented):
Just a few preliminary comments about the code, but as per feature scope, better to ping @DenysKarmazynDFINITY  for a review

src/frontend/src/lib/components/swap/SwapProvider.svelte
```diff
@@ -0,0 +1,60 @@
...
+	let displayURL: OptionString;
+	$: {
+		if (nonNullish(kongSwapDApp)) {
+			try {
+				const validatedWebsiteUrl = safeParse({
+					schema: UrlSchema,
+					value: kongSwapDApp?.website
+				});
+				if (nonNullish(validatedWebsiteUrl)) {
+					websiteURL = new URL(validatedWebsiteUrl);
```
DenysKarmazynDFINITY, line 26: The `websiteURL` const  is not used anywhere except this line, is it correct? Isn't its idea is to be used as an anchor href?

src/frontend/src/lib/components/swap/SwapProvider.svelte
```diff
@@ -0,0 +1,60 @@
...
+				<div class="mt-1">
+					<Logo
+						src={kongSwapDApp.logo}
+						alt={replacePlaceholders($i18n.dapps.alt.logo, { $dAppName: kongSwapDApp.name })}
+					/>
+				</div>
+				<div class="mr-auto">
+					<div class="text-lg font-bold">{kongSwapDApp.name}</div>
+					{#if nonNullish(websiteURL) && nonNullish(displayURL)}
+						<div class="text-sm text-misty-rose">{displayURL}</div>
```
DenysKarmazynDFINITY, line 54: I mean this place, shouldn't it be a link?

src/frontend/src/lib/components/swap/SwapProvider.svelte
BonomoAlessandro: @DenysKarmazynDFINITY At first i implemented it as an anchor href, but in this case it should only be a plain text.

src/frontend/src/lib/components/swap/SwapProvider.svelte
BonomoAlessandro: @DenysKarmazynDFINITY no, it should not be a link

src/frontend/src/lib/components/swap/SwapProvider.svelte
```diff
@@ -0,0 +1,60 @@
...
+			<div class="flex gap-2">
+				<div class="mt-1">
+					<Logo
+						src={kongSwapDApp.logo}
+						alt={replacePlaceholders($i18n.dapps.alt.logo, { $dAppName: kongSwapDApp.name })}
+					/>
+				</div>
+				<div class="mr-auto">
+					<div class="text-lg font-bold">{kongSwapDApp.name}</div>
+					{#if nonNullish(websiteURL) && nonNullish(displayURL)}
```
DenysKarmazynDFINITY, line 53: Since `websiteURL` is not going to be used, just checking `displayURL` should be enough

DenysKarmazynDFINITY (approved):
Please take a look if we can remove that additional check, but otherwise LGTM, approving already.

PR #4450, opened 2025-01-27
refactor(frontend): rename Ethereum components to add a new token

AntonioVentilii:
# Motivation

Since we are going to create similar components for Solana network, it makes sense to rename the Ethereum components that  treat a new additional token.


DenysKarmazynDFINITY (approved):
👍

PR #4451, opened 2025-01-27
refactor(frontend): rename prop regarding ERC20 in Manage Tokens modal

AntonioVentilii:
# Motivation

For more clarity, we specify that the prop `contractAddress` is for ERC20, when adding a new token. It will be useful since Solana uses a similar nomenclature.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4452, opened 2025-01-27
refactor(frontend): remove duplicate codes in AddTokenByNetwork

AntonioVentilii:
# Motivation

To make it more readable (and since we are going to add Solana), we make the component `AddTokenByNetwork` simpler, by checking just once what network are we dealing with.


DenysKarmazynDFINITY (approved):
👍

PR #4453, opened 2025-01-27
test(frontend): clean console noise in unittests

AntonioVentilii:
# Motivation

The tests are providing too much noise as feedback. Once first thing to tackle is to remove the `console` feedback when we mock errors and warnings.

Furthermore, we normalize the way we mock them.


peterpeterparker (approved):
LGTM, thx

Maybe (in another PR) we can add the same configuration as in NNS dapp but, for the CI only - i.e. a console log makes the tests fail in the CI but, allowed locally which is handy for development purpose?

AntonioVentilii:
@peterpeterparker done with:

- PR #4455

PR #4454, opened 2025-01-27
refactor(frontend): give props to Iconly icons

AntonioVentilii:
# Motivation

The Iconly icons were giving some small warning because they were being called without the correct props. We fix it providing `styleClass` and `size` as the other icons.

No change in the UI

Screenshot 2025-01-27 at 10 51 41


peterpeterparker (approved):
LGTM, thx!

PR #4455, opened 2025-01-27
test(frontend): fail when tests logs to console

AntonioVentilii:
# Motivation

Under @peterpeterparker 's suggestion in https://github.com/dfinity/oisy-wallet/pull/4453#pullrequestreview-2574990676: taking inspiration from [NNS Dapp](https://github.com/dfinity/nns-dapp/blob/ac40ed19f67ce583f6ec0094aafbe3bc15d0ca0b/frontend/vitest.setup.ts#L134), we want the tests to fail in case they log to console. That is done to simplify our readability and the consistent writing of unittests.

However, we allow the developer to avoid such restriction when debugging, if necessary.

# Changes

- Copy utils from NNS dApp to throw errors when loggin, and to allow logging in debugging.
- Adapt vitest setup to use such functions.
- Create script `test:debug` that allows logging, via an env variable.

# Tests

The CI ran several times on this PR when there were still errors being logged to console, and it failed, as expected. Furthermore, I injected an error to run the `test:debug` script locally and it worked.


peterpeterparker (commented):
Looks good, indeed that's what I meant. However, I would just suggest an improvement:  I find this check super useful for CI but also really annoying when used locally while I'm developing tests. So was thinking would be cool to have `npm run test` with it and `npm run test:somethingelse` without it.

AntonioVentilii:
@peterpeterparker as discussed offline, good suggestion! I included a script `test:debug`

peterpeterparker (approved):
Cool. LGTM, thanks!

PR #4456, opened 2025-01-27
feat(frontend): sets autofocus on select field on desktop

BonomoAlessandro:
# Motivation

If the `select token modal` of the `swap` process is opened, the search field should automatically receive focus, but only on desktop. The exact same should happen if the `manage list modal` is opened.

# Changes

- auto focus search fields on desktop

# Tests

**before:**
image

**after:**
image


BonomoAlessandro:
@peterpeterparker What do you think about the `device.utils.ts` class? Do you see a better or cleaner solution than my implementation?

src/frontend/src/lib/components/ui/InputTextWithAction.svelte
```diff
@@ -1,11 +1,23 @@
...
 	export let value = '';
 	export let name: string;
 	export let placeholder: string;
 	export let required = true;
 	export let testId: string | undefined = undefined;
+	export let autofocus = false;
+
+	onMount(() => {
+		if (autofocus) {
+			let component = document.querySelector<HTMLInputElement>(`input[name="${name}"]`);
```
peterpeterparker, line 15: Is there a particular reason why you're using vanilla JavaScript instead of Svelte here?

If not, could you switch to Svelte? I generally prefer to avoid mixing both approaches, as I'm not sure that it cannot lead to unexpected behavior.

Additionally, this query selector seems optimistic, as it selects any matching element from the root. There seems to be nothing preventing another component from reusing the exact same selector in the future, which could lead to unexpected behavior or malfunctions.

src/frontend/src/lib/components/ui/InputTextWithAction.svelte
```diff
@@ -1,11 +1,23 @@
...
+			if (nonNullish(component)) {
+				component.focus();
```
peterpeterparker, line 17: nitpick: nullish coallish would not work?

src/frontend/src/lib/utils/device.utils.ts
```diff
@@ -0,0 +1,16 @@
+import { nonNullish } from '@dfinity/utils';
+
+interface UserAgentData {
+	mobile?: boolean;
+}
+
+export const isMobile = () => {
```
peterpeterparker, line 7: nitpick: return type `: boolean`

src/frontend/src/lib/utils/device.utils.ts
```diff
@@ -0,0 +1,16 @@
+import { nonNullish } from '@dfinity/utils';
+
+interface UserAgentData {
```
peterpeterparker, line 3: Is there a typedef (`@types/...`) available for this? If not, should we declare the type globally in a `.d.ts` file in case it ever become available in NodeJS?

Like declaring it for the `navigator` this way you do not have to explicitely declare the types when you use it.

Strictly a question, not necessarely a call to action.

src/frontend/src/lib/utils/device.utils.ts
```diff
@@ -0,0 +1,16 @@
...
+	mobile?: boolean;
+}
+
+export const isMobile = () => {
+	if ('userAgentData' in navigator && nonNullish(navigator.userAgentData)) {
+		const userAgentData: UserAgentData = navigator.userAgentData;
+		return nonNullish(userAgentData.mobile) && userAgentData.mobile;
+	}
+	const isTouchScreen = window.matchMedia('(any-pointer:coarse)').matches;
```
peterpeterparker, line 12: I just double checked, in one of my old piece of code I asserted `any-pointer:fine` in addition to `any-pointer:coarse` - i.e. "Is touch screen without mouse".

I'm not sure if this still works at all or if it even makes sense, but sharing in case it's useful.

```
const isTouchScreen = window.matchMedia('(any-pointer:coarse)').matches;
const isMouseScreen = window.matchMedia('(any-pointer:fine)').matches;

  return isTouchScreen && !isMouseScreen;
```

src/frontend/src/lib/components/ui/InputTextWithAction.svelte
```diff
@@ -1,11 +1,23 @@
...
 	import { Input } from '@dfinity/gix-components';
+	import { nonNullish } from '@dfinity/utils';
+	import { onMount } from 'svelte';
 
 	export let value = '';
 	export let name: string;
 	export let placeholder: string;
 	export let required = true;
 	export let testId: string | undefined = undefined;
+	export let autofocus = false;
```
peterpeterparker, line 11: I think we still need the JavaScript anyway but, for the state of the art, should this property be reflected as an attribute of the `input` element in the DOM?

peterpeterparker (commented):
- Can you provide the utility in a separate PR?

- No tests?

src/frontend/src/lib/components/ui/InputTextWithAction.svelte
```diff
@@ -1,11 +1,23 @@
...
+
+	onMount(() => {
```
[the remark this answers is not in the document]
BonomoAlessandro: @peterpeterparker yes, the behavior of always focus the search field is expected, even if you are returning to the first screen.

src/frontend/src/lib/components/ui/InputTextWithAction.svelte
BonomoAlessandro: @peterpeterparker yes there is. If i use the svelte `bind`, i will get the svelete component `Input` from `gix-components`, which does not provide an option to set focus. To be able to set the focus, i use the JavaScript to directly access the `input` element, allowing me to set the focus.

src/frontend/src/lib/components/ui/InputTextWithAction.svelte
BonomoAlessandro: @peterpeterparker i don't think so. The nullish coallish is an operator that returns the right-hand side operand when its left-hand side operand is nullish, but in this case we want to make sure that `component` is not nullish.

src/frontend/src/lib/utils/device.utils.ts
BonomoAlessandro: @peterpeterparker nice thx. It does still work and i think it does make sense to implement it here in the same way.

src/frontend/src/lib/components/ui/InputTextWithAction.svelte
BonomoAlessandro: @peterpeterparker this does not work. The reason is that here we do use the `Input` component from `gix-components` and not the default HTML `input` element. The `Input` component from `gix-components` does not provide an autofocus property.

peterpeterparker:
Thanks for the clarification. So, why not enhancing the Gix components directly instead of working around in OISY?

This feature could be integrated into the design kit, allowing the community and other foundation dapps, like NNS dapp, to benefit from it as well.

Alternatively, why not just exposing the HTML input element in the Gix component?

BonomoAlessandro:
@peterpeterparker I also think the cleanest solution would be to enhance the Gix components, but this will take some time. In the meantime, we want to implement a workaround on the OISY side.

As you requested in your first comment, i have provided some test cases and created a separate PR for the `device.utils`. You can find it here: https://github.com/dfinity/oisy-wallet/pull/4493

peterpeterparker:
I reviewed and approved #4493, thanks. It was a draft, but since you mentioned it, I guess it was ready.

Regarding Gix components, I don't think it takes that much more time, and I believe this workaround, as it is in this PR, should not be implemented this way. Hence, I’m personally not going to approve this PR in this form.

package.json
```diff
@@ -57,7 +57,7 @@
 		"@dfinity/candid": "^2.1.2",
 		"@dfinity/ckbtc": "^3.1.6",
 		"@dfinity/cketh": "^3.4.3",
-		"@dfinity/gix-components": "^5.0.0-next-2024-12-10",
+		"@dfinity/gix-components": "^5.0.0-next-2025-01-31",
```
peterpeterparker, line 60: Does this `next` of Gix components version contains solely the changes for this PR or does it contains other changes? If the later, it's cleaner I think to provide the upgrade in a dedicated PR. If the former, nvm this comment.

BonomoAlessandro: @peterpeterparker This provides all the changes made since december 10th. I will create a separate PR for this

BonomoAlessandro: @peterpeterparker I just opened the PR: https://github.com/dfinity/oisy-wallet/pull/4540

peterpeterparker: Thx!

src/frontend/src/lib/components/ui/InputTextWithAction.svelte
```diff
@@ -1,11 +1,22 @@
...
 	export let placeholder: string;
 	export let required = true;
 	export let testId: string | undefined = undefined;
+	export let autofocus = false;
+
+	let inputElement: HTMLInputElement | undefined;
+
+	onMount(() => {
+		if (autofocus && nonNullish(inputElement)) {
+			inputElement.focus();
```
peterpeterparker, line 17: Just a note—I assume you tested this after the change in Gix components.

Spontaneously, I would rather use an auto-subscriber, as I'm not sure the reference to the DOM element would always be established when the component is mounted. But I'm probably just being paranoid. ;)

peterpeterparker (approved):
LGTM, thx

PR #4459, opened 2025-01-27
refactor(frontend): create Solana Fee context

AntonioVentilii:
# Motivation

We will need to have the Solana fee through a few other components, especially to freeze it when sending the transaction.

So, we move the fee calculation (with the prioritization fee), in a separate component that defines the context. It is very similar to what it is done with Ethereum.

# Changes

- Create fee store and fee context: this is basically a simplified version of the Ethereum one.
- Move `prioritizationFee` calculation to another component `SolFeeContext`.
- Modify the functions to update the context instead of local variables.
- Wrap the components in `SolSendTokenWizard`.
- Initialize the context in `SolSendTokenWizard`.
- Adapt `SolFeeDisplay` to call the context.

# Tests

No changes to the replica:

Screenshot 2025-01-27 at 13 26 58


DenysKarmazynDFINITY (approved):
Looks good with the context!

PR #4460, opened 2025-01-27
feat(frontend): use Solana prioritization fee in sending flow

AntonioVentilii:
# Motivation

We can now use the calculated prioritization fee for the Solana transactions, when it is calculated. We base the implementation on the [documentation](https://solana.com/developers/guides/advanced/how-to-use-priority-fees) provided by the Solana foundation.

# Changes

- Provide `prioritizationFee` to service `sendSol` using the current context.
- Compute the approximate Compute Units necessary for the transaction message we are trying to send.
- Derive the Compute Unit Price using both the estimation and the `prioritizationFee`.
- Adjust the transaction message to set the Compute Unit price.
- Remove hard-coded buffer of `100n` block height: that is not necessary anymore after this PR.

# Tests

Adapted current tests. Plus I tested in the local environment some transactions, for example https://solscan.io/tx/56o8KNhwA45xZJ4n4zThwLt9VpBYDHYCjUfdkN71dhPVkrURzGGWBDHYuLxTVAku7qtEvHEnAmScbiaaFTa2GS9t


src/frontend/src/sol/services/sol-send.services.ts
```diff
@@ -65,7 +70,7 @@ export const setLifetimeAndFeePayerToTransaction = async ({
 
 	const correctedLatestBlockhash = {
 		...latestBlockhash,
-		lastValidBlockHeight: latestBlockhash.lastValidBlockHeight + 100n
+		lastValidBlockHeight: latestBlockhash.lastValidBlockHeight
```
DenysKarmazynDFINITY, line 73: I was this 🤏 close to ask how to interpret this change and then saw my answer in the PR description. Thanks 🙂

DenysKarmazynDFINITY (approved):
LGTM

PR #4461, opened 2025-01-27
feat(frontend): disable switch token in swap if loading

loki344:
# Motivation

We want to disable the SwitchTokensButton in the Swap context if the amounts are loading.

# Changes

Add optional disabled prop and use it in SwapForm

# Tests

Unit tests provided for new functionality in SwapForm.

Additionally, after and before:

https://github.com/user-attachments/assets/ecc38371-d2b2-4693-ac46-952d3f7411dd


AntonioVentilii (commented):
Is it possible to put in this PR the practical usage of the new prop, or is it too much code after?

src/frontend/src/lib/components/swap/SwapSwitchTokensButton.svelte
```diff
@@ -5,11 +5,13 @@
 	import { i18n } from '$lib/stores/i18n.store';
 
 	const dispatch = createEventDispatcher();
+	export let disabled = false;
```
DenysKarmazynDFINITY, line 8: nitpick: can you please define the prop right after the imports? So it follows our "informal" structure (imports -> props -> rest of the things).

loki344:
> Is it possible to put in this PR the practical usage of the new prop, or is it too much code after?

Yes. I'll add the usage with tests, hoping the PR does not become too big.

DenysKarmazynDFINITY (approved):
Good one, thanks.

P.S.: please wait for Antonio too, since he already left some comments.

AntonioVentilii (approved):

PR #4462, opened 2025-01-27
feat(backend): Support Solana SPL as a custom token type

bitdivine:
# Motivation
Users may select a Solana token as a custom token they wish to see.  This information needs to be stored in the user profile.

# Changes
- Add SPL as a custom token type
- Add a validator trait, to prevent a user from storing a  gigantic string as a token ID.

# Tests
- Unit tests are included for the validator.
- TODO:  Persistence integration tests


src/shared/src/types.rs
```diff
@@ -181,6 +196,8 @@ pub mod custom_token {
     pub enum CustomTokenId {
         /// An ICRC-1 compliant token on the Internet Computer mainnet.
         Icrc(LedgerId),
+        /// A Solana token on the Solana mainnet.
+        SolMainnet(SplTokenId),
```
AntonioVentilii, line 200: Can you add `SolDevnet` too, here? It will be useful for me

src/shared/src/impls.rs
```diff
@@ -336,10 +341,46 @@ fn next_matches_strum_iter() {
...
+                    "Invalid Solana address: not 32 bytes when decoded",
+                ));
+            }
+        } else {
+            return Err(candid::Error::msg("Invalid Solana address: not base58"));
+        }
+        Ok(())
+    }
+}
+
```
AntonioVentilii, line 378: is it worth to validate the length of `symbol` too? Just for insurance

bitdivine: The symbol length is checked below.  The symbol in in `SplToken` but not `SplTokenId`.

src/shared/src/types.rs
bitdivine: Done.

PR #4465, opened 2025-01-27
feat(frontend): move ckBtcUpdateBalance listener up to tokens page

DenysKarmazynDFINITY:
# Motivation

Previously, the update_balance call that is required for receiving ckBTC was only triggered when user opened related transactions page. Since this step is not obvious for users, we've got a few reports in which they thought the converted from BTC tokens never arrived. To solve this, the idea is to move the ckBtcUpdateBalance worker from the ckBTC txs page to the tokens page. It will be only launched if user has ckBTC enabled.


src/frontend/src/icp/components/core/CkBTCUpdateBalanceListener.svelte
```diff
@@ -0,0 +1,25 @@
...
+	import { BTC_MAINNET_TOKEN } from '$env/tokens/tokens.btc.env';
+	import IcCkListener from '$icp/components/core/IcCkListener.svelte';
+	import { initCkBTCUpdateBalanceWorker } from '$icp/services/worker.ckbtc-update-balance.services';
+	import type { IcCkToken } from '$icp/types/ic-token';
+	import { enabledIcTokens } from '$lib/derived/tokens.derived';
+	import { findTwinToken } from '$lib/utils/token.utils';
+
+	let ckBtcToken: IcCkToken | undefined;
+	$: ckBtcToken = findTwinToken({
+		tokenToPair: BTC_MAINNET_TOKEN,
```
AntonioVentilii, line 12: I imagine that this is because ckBTC works ONLY with BTC mainnet, right?

AntonioVentilii (approved):
LGTM, tks!

PR #4467, opened 2025-01-27
feat(frontend): enhance swap amount exchange

loki344:
# Motivation

We want to switch between token and usd unit in the swap context. As a first step, we introduce this function to the SwapAmountExchange. Note that the incorporation into the SwapForm will follow in another PR, to not clutter this PR.

# Changes

- add displayUnit type
- implement switching between display units in SwapAmountExchange
- disable this in the review view for swaps

# Tests

Unit test provided.

Additionally:

https://github.com/user-attachments/assets/579c7dac-5789-40e2-963a-9c695b447f77


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4468, opened 2025-01-27
feat(frontend): metadata API call to Quicknode

AntonioVentilii:
# Motivation

We will require to fetch an SPL token metadata when an user provides the token address and it is not among the ones already listed in our code base.

For that we will use method [`getAsset`](https://www.quicknode.com/docs/solana/getAsset) of Quicknode (of which we already have the API).


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4469, opened 2025-01-27
feat(frontend): Solana API function to get token decimals

AntonioVentilii:
# Motivation

To get an SPL token decimals there is an available method in the Solana RPC: [getAccountInfo](https://solana.com/docs/rpc/http/getaccountinfo).

PR #4470, opened 2025-01-27
feat(frontend): component SolAddTokenReview to add an SPL token

AntonioVentilii:
# Motivation

To add an SPL token, we need a few components, that we will replicate from the ERC20 similar flow.

The one approached in this PR is `SolAddTokenReview` that is responsible to show the fetched metadata based on the user input, before including the SPL token-to-be among the ones in the current list.

# Changes

- Create a single util `getSplMetadata` that merges the functions [`getTokenDecimals`](https://github.com/dfinity/oisy-wallet/pull/4469) and [`splMetadata`](https://github.com/dfinity/oisy-wallet/pull/4468).
- Copy `EthAddTokenReview` and adapt to the small parts of Solana: the component is basically identical, but a few error strings and variables change their name, so it is not trivial to re-use a common component.

# Tests

Screenshot 2025-01-27 at 21 31 12


src/frontend/src/sol/components/tokens/SolAddTokenReview.svelte
```diff
@@ -0,0 +1,155 @@
...
+	import type { SplTokenAddress } from '$sol/types/spl';
+	import { mapNetworkIdToNetwork } from '$sol/utils/network.utils';
+
+	export let tokenAddress: SplTokenAddress | undefined;
+	export let metadata: TokenMetadata | undefined;
+	export let network: Network;
+
+	onMount(async () => {
+		if (isNullish(tokenAddress)) {
+			toastsError({
```
DenysKarmazynDFINITY, line 29: Those checks are mostly for TS, right? I assume it's not possible to get to the review step without that data.

PR #4472, opened 2025-01-27
refactor(frontend): create custom interface for Solana save-user-tokens

AntonioVentilii:
# Motivation

Similar to ERC20, we require a minimal interface that would accept tokens added by the user (so without a lot of data, and without ID, since it is a `Symbol` and thus unique). Right now we provide the entire SPL token, but it is not necessary.

# Changes

- Create interface `SaveSplUserToken` that is a minimal version of `SplUserToken` (this is done in a very similar way in ERC20).
- Adapt usage of save functions to the new interface.

# Tests

Current tests are sufficient.


src/frontend/src/sol/services/spl-user-tokens.services.ts
```diff
@@ -7,10 +8,16 @@ import {
 	type SplAddressMap
 } from '$sol/stores/spl-user-tokens.store';
 import type { SplTokenAddress } from '$sol/types/spl';
-import type { SplTokenToggleable } from '$sol/types/spl-token-toggleable';
+import type { SplUserToken } from '$sol/types/spl-user-token';
 import type { Identity } from '@dfinity/agent';
 import { nonNullish } from '@dfinity/utils';
 
+export type SaveSplUserToken = Pick<
```
DenysKarmazynDFINITY, line 15: Since the type is being re-used in different places, let's move it to the `types` folder? We can do it in a separate PR to tackle erc20 `SaveUserToken` too.

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/sol/services/spl-user-tokens.services.ts
AntonioVentilii: Done!

PR #4473, opened 2025-01-27
fix(ci): Clean up deployment

bitdivine:
# Motivation
`update-snapshots.yaml` is failing.  Reading through the code there are quite a few places that are in need of tidying up.  Simplifying the code is quite effective at exposing where the bug is.

# Changes
- Specify canister IDs in dfx.json rather than with flags in deploy.sh.  (Note: There are a few more places where this still needs to be done).
- When getting cached Wasms, put them where dfx expects them, so that no additional flags are needed.
- Delete duplicate software installation.
- When using `dfx deploy backend`, support build strategies that just use the existing artefacts.  This is just for speed.

# Tests
- See CI.  update-snapshots is triggered by this PR.  Formerly the canister installation failed.  Now it succeeds.
- But updated snapshots are added to the PR, which then runs again and creates different snapshots, so it runs again ad nauseam.  I'll remove the `run-e2e-snapshots` label.


.github/actions/deploy-backend/action.yml
```diff
@@ -31,4 +31,4 @@ runs:
 
     - name: Deploy backend
       shell: bash
-      run: npm run deploy
+      run: BACKEND_BUILD_STRATEGY=args npm run deploy
```
bitdivine, line 34: When deploying, this rebuilds only the arguments.  It checks that the Wasm already exists.

Note: The arguments are built as part of the docker build but for some reason we keep only the Wasm and throw away the arguments, so we need to recreate them.

.github/actions/docker-build-backend/action.yml
bitdivine, line 1: When building the backend we really should provide the target network, as it isn't clear whether the network should default to local, as for e2e tests, or production, as for releases.

.github/actions/prepare-macos/action.yml
bitdivine, line 1: We need sha256sum, which is not installed by default on mac and is provided by coreutils.

.github/workflows/update-snapshots.yml
```diff
@@ -74,6 +76,7 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: backend.wasm.gz
+          path: out/
```
bitdivine, line 79: `dfx deploy` expects the Wasm to be at `out/backend.wasm.gz`, so if we put it there we can deploy without overriding the Wasm path.

.github/workflows/update-snapshots.yml
```diff
@@ -85,9 +88,6 @@ jobs:
       - name: Deploy the backend
         uses: ./.github/actions/deploy-backend
 
-      - name: Prepare
-        uses: ./.github/actions/prepare
-
```
bitdivine, line 90: The prepare action is already called by the deploy-backend action in the previous step.  No need to do it twice.

dfx.json
```diff
@@ -91,6 +91,7 @@
 			"candid": "https://github.com/dfinity/internet-identity/releases/download/release-2024-10-25/internet_identity.did",
 			"wasm": "https://github.com/dfinity/internet-identity/releases/download/release-2024-10-25/internet_identity_dev.wasm.gz",
 			"init_arg": "(opt record { captcha_config = opt record { max_unsolved_captchas= 50:nat64; captcha_trigger = variant {Static = variant {CaptchaDisabled}}}})",
+			"specified_id": "rdmx6-jaaaa-aaaaa-aaadq-cai",
```
bitdivine, line 94: There are lots of canisters that include `--specified-id xxx` in their canister creation or deploy commands.  This is unnecessary as the canister ID can be specified in `dfx.json`.  If we keep on simplifying like this we can do `dfx deploy` without arguments, and that will deploy all the canisters.

scripts/build.backend.sh
bitdivine, line 1: Here we support multiple strategies to get the deploy artefacts:
- Built everything from scratch, as before.  Simple but painfully slow.
- Assume that everything has already been prebuilt.  Useful when docker has already created the arguments and we want `dfx deploy` to just deploy them without change.
- Use the existing Wasm but rebuild the arguments.

scripts/deploy.sh
bitdivine, line 1: This script was failing, but didn't abort because `set -euo pipefail` was not set.

.github/actions/docker-build-backend/action.yml
bitdivine, line 1: When buildingthe backend, we should really provide the network, as otherwise it's not clear whether to build for local, as needed for tests, or for the ic, as needed for releases.

.github/actions/deploy-backend/action.yml
inc-man: maybe we can try and fix it in a way that we keep wasm and arguments. But agreed, for now the recreation makes sense.

.github/actions/prepare-macos/action.yml
inc-man: good catch

.github/workflows/update-snapshots.yml
inc-man: dito

inc-man (approved):
Overall a nice fix for the current issue. I like the way it both fixes and improves the backend deployment. 

As for the update snapshot task that runs in a limbo, that is an issue with the flakiness. Eventually it should stop after a few runs.

In general, LGTM! Thanks a lot!

PR #4474, opened 2025-01-27
feat(frontend): possible SPL token address in inputs in AddTokenByNetwork

AntonioVentilii:
# Motivation

In component AddTokenByNetwork there will be the possibility to add an SPL token. However, we first need to adapt the flow, before allowing the user to provide a data that OISY does not know how to treat.

# Changes

- Include SPL token address among the possible token data provided.
- Create type utils to choose only one type of token data.
- Adapt `AddTokenByNetwork` to the Solana data.

# Tests

No changes in local replica.


src/frontend/src/lib/utils/ts.utils.ts
```diff
@@ -6,3 +6,9 @@ export type Only<T, U> = {
 };
 
 export type Either<T, U> = Only<T, U> | Only<U, T>;
+
+export type OneOf<T extends readonly unknown[]> = T extends [infer First, ...infer Rest]
```
DenysKarmazynDFINITY, line 10: fancy!

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4475, opened 2025-01-27
feat(frontend): include Solana token review in ManageTokensModal

AntonioVentilii:
# Motivation

After https://github.com/dfinity/oisy-wallet/pull/4474 , we can include the component `SolAddTokenReview` in `ManageTokensModal`, since it can be possible that the user inputs an SPL token as additional token.

# Changes

- Create small sub-function `saveSplToken`, basically identical to the ERC20 equivalent.
- Include SolAddTokenReview in the if condition to check which network we are selecting.

# Tests

Screenshot 2025-01-27 at 21 30 57
Screenshot 2025-01-27 at 21 31 12


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4477, opened 2025-01-27
feat(frontend): use swap display unit in SwapForm

loki344:
# Note

First review: https://github.com/dfinity/oisy-wallet/pull/4467

# Motivation

Users want to change easily between usd and token amounts in the SwapForm. 

# Changes

- add DisplayUnit handling in SwapInputCurrency field
- include display unit state in SwapForm
- add and improve tests in SwapForm

# Tests

Unit tests provided.

additionally:

https://github.com/user-attachments/assets/8b30be97-8511-42fc-becd-cc1681db9bed


src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
```diff
@@ -1,22 +1,106 @@
...
+				? Number(displayValue) / exchangeRate
+				: Number(displayValue)
+		).toFixed(decimals);
+	};
+
+	const changeDirection = () => {
+		if (isNullish(exchangeRate) || isNullish(value)) {
+			return;
+		}
+		displayValue = isUSDDisplayUnit ? (Number(value) * exchangeRate).toFixed(2) : Number(value);
```
DenysKarmazynDFINITY, line 49: I'm wondering it would be cleaner to have the following structure:
1. We keep the `SwapInputCurrency` agnostic of the current display unit.
2. Instead, we create `SwapInputCurrencyUsd` and `SwapInputCurrencyToken`, where they handle just 1 display unit, which means we don't need so many `if`s.
3. In the `SwapForm`, we do something like `isUSDDisplayUnit ? <SwapInputCurrencyUsd decimals={2) ....... /> : <SwapInputCurrencyToken {decimals} ...... />`.

I understand the current approach does the job (kudos for the tests) and that splitting the logic into separate component will increase the amount of code, but I'd go for it to increase this part's readability.

src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
```diff
@@ -1,22 +1,106 @@
...
+		previousDisplayValue = displayValue;
+	};
+
+	const updateDisplayValue = () => {
+		if (isNullish(value)) {
+			value = undefined;
+			displayValue = undefined;
+			previousDisplayValue = displayValue;
```
DenysKarmazynDFINITY, line 57: It seems that `previousDisplayValue` almost always equals to `displayValue` since you update it after `displayValue` got a new value. Still don't understand fully what's the purpose of this const.

src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
```diff
@@ -1,22 +1,106 @@
...
+		<span
+			class="pl-3 transition-colors"
+			class:text-placeholder={isNullish(displayValue)}
+			data-tid={SWAP_INPUT_CURRENCY_USD_SYMBOL}
+		>
+			$
+		</span>
+	{/if}
+	<InputCurrency
+		bind:value={displayValue}
```
DenysKarmazynDFINITY, line 94: I may miss some context, but would the structure I suggested above help to avoid maintaining both `displayValue` and `value` at the same time?

src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
```diff
@@ -34,6 +118,14 @@
...
 		border-radius: 0;
+		padding: 0;
+	}
+
+	:global(.swap-input-currency.padding div.input-field input[id]) {
 		padding: 0 0 0 0.75rem;
 	}
+
+	.transition-colors {
+		transition: color var(--animation-time-short);
```
DenysKarmazynDFINITY, line 129: Would tailwind's `transition-colors` + `transition-duration-...`  not work?

src/frontend/src/lib/components/swap/SwapInputCurrency.svelte
loki344: I like this approach, will try this!

DenysKarmazynDFINITY (approved):
Thanks, I like how it looks now 👍

PR #4479, opened 2025-01-28
feat(backend): Validate custom tokens

bitdivine:
# Motivation
We would like to validate token types when they are parsed.

# Changes
- Add more "validate on parsing" machinery and apply it to the existing `IcrcToken`.

# Tests
- Tests for the `IcrcToken` validation are included.
- I  have verified manually that if test vectors are wrong (e.g. claim that an entry is valid when it is not), the test fails.
- I have verified manually that if I omit the deserilaization code for IcrcToken, the code either fails to compile or the tests fail, depending on what parts of the implementation I comment out.


src/shared/src/impls.rs
```diff
@@ -331,3 +335,54 @@ fn next_matches_strum_iter() {
         "Once completed, it should stay completed"
     );
 }
+
+impl Validate for CustomTokenId {
+    fn validate(&self) -> Result<(), candid::Error> {
+        match self {
+            CustomTokenId::Icrc(_) => Ok(()), // This is a principal.  In principle we could check the exact type of principal.
```
bitdivine, line 342: @AntonioVentilii   What exactly is this principal?  Is it the ledger canister ID?

src/shared/src/impls.rs
bitdivine, line 1: Create validators and call them on deserialization.

src/shared/src/types.rs
```diff
@@ -149,26 +152,34 @@ pub mod custom_token {
     pub type LedgerId = Principal;
     pub type IndexId = Principal;
 
+    /// An ICRC-1 compliant token on the Internet Computer.
     #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
+    #[serde(remote = "Self")]
```
bitdivine, line 157: To validate on deserialization we need this "remote" setting.

src/shared/src/types/tests.rs
bitdivine, line 1: Tests for IcrcTokens.  We check that the ledger and index canister IDs are valid and not, e.g. user or anonymous principals.

src/shared/src/validate.rs
bitdivine, line 1: This generalizes the test "validate on deserialize":

src/shared/src/validate/tests.rs
bitdivine, line 1: This test is updated to use the macro.

src/shared/src/impls.rs
AntonioVentilii: theoretically an ICRC token that ends up in the backend has a ledger canister ID and an optional index canister ID. But the identifier is always the ledger canister ID

AntonioVentilii (approved):
LGTM tks

src/shared/src/impls.rs
bitdivine: I'll validate it accordingly then.

PR #4480, opened 2025-01-28
feat(frontend): enables token after swap

BonomoAlessandro:
# Motivation

If a user swaps tokens and the newly received token is not enabled, it should be enabled automatically.

# Changes

- enables token on swap

# Tests

**before:**

https://github.com/user-attachments/assets/cf407ccb-a62a-412e-bc6b-10eb9ac312f1

**after:**

https://github.com/user-attachments/assets/b65e43a2-be02-4e2e-a3dd-d519928e6238


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -82,6 +88,18 @@ export const swap = async ({
 
 	progress(ProgressStepsSwap.UPDATE_UI);
 
+	const customTokenToSave: SaveCustomToken = destinationToken as unknown as SaveCustomToken;
```
AntonioVentilii, line 91: when we provide `destinationToken`, does it have the `enabled` prop? Meaning, will the cast be correct? It seems forced, since we pass through `unknown`.

src/frontend/src/icp/validation/custom-token.validation.ts
```diff
@@ -0,0 +1,7 @@
+import { SaveCustomTokenSchema } from '$icp/schema/custom-token.schema';
+import type { IcToken } from '$icp/types/ic-token';
+
+export const isSaveCustomToken = (token: IcToken) => {
```
AntonioVentilii, line 4:
```suggestion
export const isSaveCustomToken = (token: IcToken): token is SaveCustomToken => {
```

AntonioVentilii (approved):
LGTM tks!

PR #4483, opened 2025-01-28
feat(frontend): hide error toasts on BTC balance update

DenysKarmazynDFINITY:
# Motivation

Update BTC balance still sometimes throws "ingress_expiry" issue. Previously, there was a problem with busy subnets (https://forum.dfinity.org/t/subnets-with-heavy-compute-load-what-can-you-do-now-next-steps/35762/19), but it is supposed to be fixed some time ago. 

For the time being, we were asked to hide related error toasts (same as in the Solana wallet worker). I added a TODO comment to remove this change after a proper investigation is completed and the error is fixed. For the debugging purposes, we will still show toast in staging env.

### DIFF
--- a/src/frontend/src/btc/services/worker.btc-wallet.services.ts
+++ b/src/frontend/src/btc/services/worker.btc-wallet.services.ts
@@ -1,5 +1,6 @@
 import { syncWallet, syncWalletError } from '$btc/services/btc-listener.services';
 import type { BtcPostMessageDataResponseWallet } from '$btc/types/btc-post-message';
+import { STAGING } from '$lib/constants/app.constants';
 import {
 	btcAddressMainnetStore,
 	btcAddressRegtestStore,
@@ -31,6 +32,7 @@ export const initBtcWalletWorker = async ({
 
 	const isTestnetNetwork = isNetworkIdBTCTestnet(networkId);
 	const isRegtestNetwork = isNetworkIdBTCRegtest(networkId);
+	const isMainnetNetwork = isNetworkIdBTCMainnet(networkId);
 
 	worker.onmessage = ({ data }: MessageEvent<PostMessage<BtcPostMessageDataResponseWallet>>) => {
 		const { msg } = data;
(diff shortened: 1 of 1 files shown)


peterpeterparker:
Maybe none of my business but, the issue you mention has been resolved for some time now, are you sure it's related and the error should just be made hidden?

DenysKarmazynDFINITY:
@peterpeterparker hi David, thanks for taking a look. I adjusted the PR description and TODO comment in the code since I'm not sure if the issue I mentioned is indeed the root cause. For now, I only checked the OISY part, but to really find out where does the error come from we'll need to take a deeper look at the dependencies and/or canisters code. For now, we were suggested to reduce the number of toasts that do not affect user flow - e.g. in case of BTC wallet, the balance will be updated on the next worker trigger (30s).

peterpeterparker:
Weird way to resolve an issue but, none of my business. Thanks for the feedback.

src/frontend/src/btc/services/worker.btc-wallet.services.ts
```diff
@@ -51,8 +53,9 @@ export const initBtcWalletWorker = async ({
 					 * TODOs:
 					 * 1. Do not launch worker locally if BTC canister is not deployed, and remove "isRegtestNetwork" afterwards.
 					 * 2. Wait for testnet BTC canister to be fixed on the IC side, and remove "isTestnetNetwork" afterwards.
+					 * 3. Investigate the "ingress_expiry" error that is sometimes thrown by update BTC balance call, and remove "isMainnetNetwork" afterwards.
```
peterpeterparker, line 56: You might want to add a `// TODO:` per line / item here. This can be useful in IDE which have features to list the remaining todos.

peterpeterparker (approved):
Rubber stamp.

src/frontend/src/btc/services/worker.btc-wallet.services.ts
DenysKarmazynDFINITY: good point, thanks

StefanBerger-DFINITY:
Denys, I will add a follow up on that to deal with the remaining todos

PR #4484, opened 2025-01-28
chore(test): use scss modern-compiler for tests

peterpeterparker:
# Motivation

To resolve the following tests warning we want to use the CSS `modern-compiler` for testing purpose as well:

> Deprecation Warning [user-authored]: The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.
>
> More info: https://sass-lang.com/d/legacy-js-api

# Changes

- Extract vite CSS options to a constant utils
- Use constant for vitest config


AntonioVentilii (approved):
Thank you for the fix!

PR #4485, opened 2025-01-28
feat(frontend): load Solana user tokens that were included manually

AntonioVentilii:
# Motivation

Awaiting for the backend methods to be completed, to show better the SPL tokens that the user added manually, we can now split the loaded user tokens between the ones that we have hard-coded and the ones that require to load the metadata. 

# Changes

- Split loaded token addresses between existing tokens in the `SPL_TOKENS` list and addresses without matching token.
- Map the unmatched addresses using method `getSplMetadata`: this is very similar to how we map the ERC20 tokens.

# Tests

https://github.com/user-attachments/assets/82e913d0-9733-4268-b83a-07d8e08856e7

### DIFF
--- a/src/frontend/src/sol/services/spl.services.ts
+++ b/src/frontend/src/sol/services/spl.services.ts
@@ -1,3 +1,4 @@
+import { SOLANA_MAINNET_NETWORK } from '$env/networks/networks.sol.env';
 import { SPL_TOKENS } from '$env/tokens/tokens.spl.env';
 import { queryAndUpdate } from '$lib/actors/query.ic';
 import { nullishSignOut } from '$lib/services/auth.services';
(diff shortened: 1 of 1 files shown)


src/frontend/src/sol/services/spl.services.ts
```diff
@@ -65,25 +69,64 @@ export const loadUserTokens = async ({
...
+					solNetwork,
+					replacePlaceholders(get(i18n).init.error.no_solana_network, {
+						$network: network.id.description ?? ''
+					})
+				);
+
+				const metadata = await getSplMetadata({ address, network: solNetwork });
+
+				return {
+					...{
```
DenysKarmazynDFINITY, line 113: nitpick: looks like this spreading is redundant, you can directly put those properties into return object

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/sol/services/spl.services.ts
AntonioVentilii: good catch! adjsuted

PR #4487, opened 2025-01-28
refactor(frontend): use Solana fee context for validation

AntonioVentilii:
# Motivation

We have a Solana fee context, so it makes sense to use such a store to validate the balance.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4488, opened 2025-01-28
refactor(frontend): move ATA fee calculation in Solana fee context

AntonioVentilii:
# Motivation

Since we are going to use the ATA fee in the validation of balance, we move the calculation to the Solana fee context component and we save it in the fee context.

# Changes

- Add ATA fee store to Solana fee context.
- Move ATA fee calculation to `SolFeeContext`.
- Assign logic to the ATA fee store.
- Adapt the components to use the new store of the context.

# Tests

No changes in the local replica

Screenshot 2025-01-28 at 15 38 26


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4490, opened 2025-01-28
refactor(devops): Delete duplicate specified-id

bitdivine:
# Motivation
We would like to be able to deploy the local environment simply and reliably. Ideally with no more than `dfx deploy`.  To that end I am simplifying deploy commands one step at a time until they are trivial.

Specifically, for this PR:  The preferred way of specifying canister IDs is now in dfx.json, so passing `--specified-id` as a deploy argument is really only for overrides.  Most canister IDs are already specified in `dfx.json` so in most cases we can simply delete the flags.

# Changes
- Delete `--specified-id` almost everywhere.

# Tests
Existing CI should suffice.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -321,12 +321,24 @@
 		"ckusdc_ledger": {
 			"type": "custom",
 			"candid": "target/ic/cketh_ledger.did",
-			"wasm": "target/ic/cketh_ledger.wasm"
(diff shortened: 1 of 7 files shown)


dfx.json
AntonioVentilii, line 1: I think you can remove them from `deploy.ckerc20.sh` script too, no?

PR #4493, opened 2025-01-29
feat(frontend): introduce the detection of mobile devices

BonomoAlessandro:
# Motivation

We want to be able to determine if the application is running on a desktop or mobile device.

# Changes

- adds device utils


peterpeterparker (approved):
Nice!

LGTM, thx

peterpeterparker:
Maybe just review the title of the PR to introduce the detection of mobile devices before merge? It could be useful for reference if and when we search through the git history.

PR #4494, opened 2025-01-29
feat(frontend): update ETH txs and erc20 tokens error messages

DenysKarmazynDFINITY:
# Motivation

The idea is to make ETH txs and erc20 tokens error messages user-friendlier by removing the "err" part from the error toasts. The `err` object will be logged in the console instead.

a6575c24-a073-4cde-9728-98eb472ce5ab

### DIFF
--- a/src/frontend/src/lib/stores/toasts.store.ts
+++ b/src/frontend/src/lib/stores/toasts.store.ts
@@ -5,13 +5,11 @@ import { nonNullish } from '@dfinity/utils';
 
 export const toastsShow = (msg: ToastMsg): symbol => toastsStore.show(msg);
 
-export const toastsError = ({
-	msg: { text, ...rest },
-	err
-}: {
+interface ToastsErrorParams {
 	msg: Omit<ToastMsg, 'level'>;
 	err?: unknown;
-}): symbol => {
+}
+export const toastsError = ({ msg: { text, ...rest }, err }: ToastsErrorParams): symbol => {
 	if (nonNullish(err)) {
 		console.error(err);
 	}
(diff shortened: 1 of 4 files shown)


peterpeterparker:
Unsolicited question, if I may: How do you provide support for users on mobile devices?

DenysKarmazynDFINITY:
@peterpeterparker instead of just logging errors in the console, the plan is to start using Sentry (hopefully, ASAP).

peterpeterparker:
If the plan is to use Sentry (web2 for the win 😉), I guess there will be other features that will require silencing the error in the toast, therefore, would it makes already sense to add a utility that does "console error + toast without error details"?

Just in this PR, the same pattern is already repeated three times.

DenysKarmazynDFINITY:
@peterpeterparker yes, makes sense. Updated.

src/frontend/src/lib/stores/toasts.store.ts
```diff
@@ -23,6 +21,14 @@ export const toastsError = ({
 	});
 };
 
+export const toastErrorGeneric = ({ msg, err }: ToastsErrorParams) => {
```
peterpeterparker, line 24: - naming idea, sharing if interesting: `toastErrorWithoutDetails` or `toastErrorNoTrace`

- all other toasts, for some reason, are plural - e.g. `toastsShow`. Should we use plural here as well or rename all toast afterwards in another PR to singular? WDYT?

DenysKarmazynDFINITY: For consistency, renamed it to plural. Also, I like the `NoTrace` suggestion. So, the final version is `toastsErrorNoTrace`.

peterpeterparker (approved):
LGTM, thx

PR #4500, opened 2025-01-29
feat(frontend): include Solana testnet tokens in user tokens

AntonioVentilii:
# Motivation

Since we are filtering only by contract, it makes sense to get all the SPL tokens with the same contract saved (including the testnet ones).

### DIFF
--- a/src/frontend/src/sol/services/spl.services.ts
+++ b/src/frontend/src/sol/services/spl.services.ts
@@ -84,10 +84,15 @@ export const loadUserTokens = async ({
 			[SplUserToken[], SplTokenAddress[]]
 		>(
 			([accExisting, accUser], address) => {
-				const existingToken = SPL_TOKENS.find((token) => token.address === address);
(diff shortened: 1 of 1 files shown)


src/frontend/src/sol/services/spl.services.ts
```diff
@@ -84,10 +84,15 @@ export const loadUserTokens = async ({
 			[SplUserToken[], SplTokenAddress[]]
 		>(
 			([accExisting, accUser], address) => {
-				const existingToken = SPL_TOKENS.find((token) => token.address === address);
+				const existingTokens = SPL_TOKENS.filter((token) => contracts.includes(token.address)).map(
```
DenysKarmazynDFINITY, line 87: Maybe a single `reduce` instead of `filter` + `map`?

AntonioVentilii: Right! Done!

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4506, opened 2025-01-29
fix(devops): Deploy only frontend in frontend step

bitdivine:
# Motivation
The step "Deploy Frontend to Environment" deploys `$CANISTER` but it looks as if it may be possible for that not to be the frontend.  This is _probably_ not intentional, and if it is, it should be documented.

# Changes
- Deploy `frontend` instead of `$CANISTER` in the "Deploy Frontend to Environment" step.

# Tests


inc-man (approved):
No idea what came to my mind when I wrote that but yeah there is either an "else" missing or youre completely right. Since the step is called Deploy frontend, the variable $CANISTER makes no sense.

Good catch.

PR #4507, opened 2025-01-30
refactor(frontend): move agent IC mock to Vitest setup

AntonioVentilii:
# Motivation

Just for more manageability, we move the `agents.ic` mock to `vitest.setup`.


vitest.setup.ts
```diff
@@ -1,13 +1,24 @@
...
+import { HttpAgent } from '@dfinity/agent';
 import '@testing-library/jest-dom';
 import { configure } from '@testing-library/svelte';
 import 'fake-indexeddb/auto';
 import { vi } from 'vitest';
-import { mockPage } from './src/frontend/src/tests/mocks/page.store.mock';
+import { mock } from 'vitest-mock-extended';
 
 vi.mock('$app/stores', () => ({
 	page: mockPage
```
bitdivine, line 10: Does this line above need to be updated?

bitdivine (approved):
I think you need to update one line and resolve a merge conflict but otherwise looks good.  Thanks!

vitest.setup.ts
AntonioVentilii: Ah, this was just a shift due to formatting: i adde `import { mock } from 'vitest-mock-extended';` (that is needed) and it adjusted the import of `mockPage` to use the aliases

PR #4508, opened 2025-01-30
feat(frontend): use AgentManager from @dfinity/utils

AntonioVentilii:
# Motivation

We would like to use the `AgentManager` class introduced in https://github.com/dfinity/ic-js/pull/731 that is basically the same as the agent utils we currently have.

# Changes

- Substitute `agents` with a new declaration of class `AgentManager`.
- Deconstruct getAgent method thanks to PR https://github.com/dfinity/ic-js/pull/836

# Tests

Current tests are sufficient.


src/frontend/src/lib/actors/agents.ic.ts
```diff
@@ -1,39 +1,4 @@
...
-}): Promise<HttpAgent> =>
-	createAgentUtils({
-		identity,
-		fetchRootKey: LOCAL,
-		host: REPLICA_HOST,
-		verifyQuerySignatures
-	});
-
-export const clearAgents = () => (agents = null);
+export const agents = AgentManager.create({ fetchRootKey: LOCAL, host: REPLICA_HOST });
```
peterpeterparker, line 4: - `verifyQuerySignatures = true` is the default in the library, right?

- I was wondering, have you thought about destructuring the agents within this module to reduce the number of changes implied by this PR?

```typescript
const agents = AgentManager.create({ fetchRootKey: LOCAL, host: REPLICA_HOST });
export const { getAgent } = agents;
```

Something like that would work?

peterpeterparker (commented):
Nice! Just two minor questions

src/frontend/src/lib/actors/agents.ic.ts
AntonioVentilii: - yes, correct!

- good suggestion! I didn't think of that! but I tried it now and it seems to be prone to errors... i cannot deconstruct it apparently. It is possible that, being `getAgent` a method of the instance (not a property), destructuring it will result in losing `this` context?

AntonioVentilii: As discussed offline, for the second part, there is PR https://github.com/dfinity/ic-js/pull/836

Once merged, we can try again here and check if it works. I will put this PR in draft for now

AntonioVentilii: @peterpeterparker all that needed to be updated it updated now. May you please re-review this one?

src/frontend/src/lib/actors/agents.ic.ts
```diff
@@ -1,39 +1,6 @@
...
-	verifyQuerySignatures?: boolean;
-}): Promise<HttpAgent> =>
-	createAgentUtils({
-		identity,
-		fetchRootKey: LOCAL,
-		host: REPLICA_HOST,
-		verifyQuerySignatures
-	});
-
-export const clearAgents = () => (agents = null);
```
peterpeterparker, line 39: This was unused right?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/actors/agents.ic.ts
peterpeterparker: Reduced to a single module change, that's my jam!

src/frontend/src/lib/actors/agents.ic.ts
AntonioVentilii: yes, correct!

PR #4510, opened 2025-01-30
refactor(frontend): re-use Value component between swap and convert

DenysKarmazynDFINITY:
# Motivation

The first step of aligning Swap and Conversion modals is to start re-using Value component.

Screenshot 2025-01-30 at 12 25 57

Screenshot 2025-01-30 at 12 25 48


AntonioVentilii (approved):
LGTM tks

PR #4511, opened 2025-01-30
refactor(frontend): re-use ExpandableValues between swap and convert

DenysKarmazynDFINITY:
# Motivation

The next step of aligning Convert and Swap flow is to extract ModalExpandableValues into a separate component. The convert form usage will be added in the following PR.

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapFees.svelte
+++ b/src/frontend/src/lib/components/swap/SwapFees.svelte
@@ -1,8 +1,7 @@
 <script lang="ts">
-	import { Collapsible } from '@dfinity/gix-components';
 	import { nonNullish } from '@dfinity/utils';
 	import { getContext } from 'svelte';
-	import { fade } from 'svelte/transition';
+	import ModalExpandableValues from '$lib/components/ui/ModalExpandableValues.svelte';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/swap/SwapFees.svelte
```diff
@@ -58,27 +57,26 @@
...
 {#if nonNullish($destinationToken) && nonNullish($sourceToken)}
-	<div in:fade class="swap-fees">
-		<Collapsible>
-			<!-- The width of the item below should be 100% - collapsible expand button width (1.5rem) -->
-			<div class="flex w-[calc(100%-1.5rem)] items-center" slot="header">
-				<ModalValue>
-					<svelte:fragment slot="label">{$i18n.swap.text.total_fee}</svelte:fragment>
+	<ModalExpandableValues>
+		<svelte:fragment slot="list-header">
+			<ModalValue>
```
AntonioVentilii, line 62: can you not do directly `<ModalValue slot="list-header">` ?

DenysKarmazynDFINITY: I definitely can, update 🙂

AntonioVentilii (approved):
LGTM tks

PR #4513, opened 2025-01-30
fix(devops): dfx deploy icp_ledger works

bitdivine:
# Motivation
In `dfx` projects, it is normal that `dfx deploy X` deploys the canister with the default configuration.  This is not currently the case for `icp_ledger`, as the default arguments are not set.

# Changes
- Define a build command. (None was defined)
- Move the argument building code so that it is called as part of the canister build step.
- Likewise call the download as part of the canister build step.
- Update `scripts/deploy.sh`

# Tests
Locally, this works:
```
# Clean environment
git clean -dfx
dfx start --clean --background
# Deploy
dfx deploy icp_ledger
```

CI: 
The only interesting case here is the local deployment for e2e tests.  That is exercised in CI.


scripts/deploy.icp_ledger.sh
bitdivine, line 1: Not needed any more as `dfx deploy icp_ledger` just works.

scripts/build.icp_ledger.args.sh
bitdivine, line 1: This is moved from `deploy.icp_ledger.sh`

inc-man (approved):
LGTM

I was wondering if we should just run the update-snapshot task on such PRs / changes. By that we guarantee that our changes basically won't break any CI. What do you think about that?

bitdivine:
Regarding running the update snapshot task; if it could be affected then, yes, definitely.  Maybe it can be run manually.

PR #4517, opened 2025-01-31
refactor(frontend): move Custom Token mapper to lib utils

AntonioVentilii:
# Motivation

Since now we can save different network tokens as Custom Tokens in the backend, we need to adapt the respective functions.

But as first step, we move the util that maps a Custom Token to `lib`. We will include more tokens/networks after.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4520, opened 2025-01-31
feat(frontend): use RPC parser for Solana transactions

AntonioVentilii:
# Motivation

This is a small addition but it will be quite helpful: when we request a transaction to the Solana RPC, we can specific the type of return. One of those is a parsed JSON, meaning that the RPC will try to decode all the instructions, as much as possible, with the known IDLs.

# NOTE

There is a [known bug](https://github.com/anza-xyz/solana-web3.js/issues/80) in `@solana/web3.js` version 2.0.0 that causes a wrong type casting in one of the parsed object. It has already been fixed, but not yet released.
To adapt temporary to such a bug, we cast the object to the correct type, so that the compiler does not flag it.


src/frontend/src/sol/types/sol-transaction.ts
```diff
@@ -21,7 +23,28 @@ export interface SolTransactionUi extends TransactionUiCommon {
 	fee?: bigint;
 }
 
-export type SolRpcTransaction = NonNullable<ReturnType<GetTransactionApi['getTransaction']>> & {
+export type SolRpcTransactionRawWithBug = NonNullable<
```
DenysKarmazynDFINITY, line 26: I think we don't need to export this one

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/sol/types/sol-transaction.ts
AntonioVentilii: Right! Done

PR #4523, opened 2025-02-02
chore(script): remove fallback on sns root logo

peterpeterparker:
# Motivation

Noticed today that we actually do not use the fallback to fetch the SNS root icon instead of the SNS ledger icon in the `build.sns-tokens` script. This because we skip the ledger empty icon.

I hesitated re-enabling the fallback or removing it but, given that we explicitely do not want to fetch any logo for the deprecated CTS SNS, I felt like that the cleanest is just to remove the fallback.

Even if it means that currently, if started from scratch, the script would fetch no logos for few SNSes (see [forum](https://forum.dfinity.org/t/icrc-metadata-inconsistency/40793/2)).

# Changes

- Remove fallback to fetch root image in sns script


AntonioVentilii (approved):
LGTM tks!

PR #4525, opened 2025-02-03
feat(devops): fix minter PEM issue

inc-man:
# Motivation

The recent changes in the backend deployment introduced an error with the PEM of the minter. 

# Changes

when creating the minter identity we use plaintext so it does not ask for an encrypted PEM


peterpeterparker (approved):
Rubber stamp

PR #4528, opened 2025-02-03
test(e2e): test ids for send receive test

inc-man:
# Motivation

For the new send and receive e2e test we need to set the correct testIds

# Changes

set testIds on all the required components

# Tests

tested on PR: https://github.com/dfinity/oisy-wallet/pull/3760


src/frontend/src/lib/constants/test-ids.constants.ts
```diff
@@ -53,6 +55,13 @@ export const NAVIGATION_ITEM_EXPLORER = 'navigation-item-explore';
 export const NAVIGATION_ITEM_SETTINGS = 'navigation-item-settings';
 export const NAVIGATION_ITEM_HOMEPAGE = 'navigation-item-homepage';
 
+export const AMOUNT_INPUT = 'amount-input';
+export const DESTINATION_INPUT = 'destination-input';
+
+export const IN_PROGRESS_MODAL = 'in-progress-modal';
+export const AMOUNT_DATA = 'amount-data';
+export const TOKEN_CARD_ICP = 'token-card-ICP-ICP';
```
AntonioVentilii, line 63: where is this one used?

inc-man: as I wrote to you on friday, by splitting up the PR a lot of things will not make sense until everything is merged. In this case, these will be used inside the page objects or test files that are currently not yet there because there will be another PR for this.

AntonioVentilii: in that case, better to add this specific test ID, when it will be used. The rest make sense that are in this PR

inc-man: I understand your perspective, but I feel that this approach might be a bit excessive. Since there are only three testIds, which are relevant to the next PR in line, splitting it up could add unnecessary overhead without significant benefit. In a perfect world these testIds would already be set before the test is created. So in that case I think its not necessary to bind them to the test.

PR #4540, opened 2025-02-03
feat(frontend): updates version of gix components

BonomoAlessandro:
# Motivation

The application should use the latest version of the `gix-components` lib.

# Changes

- `gix-components` lib version


peterpeterparker (commented):
Did you trigger/re-generate the screenshots for the PR to ensure that updating the library does not cause any changes?

BonomoAlessandro:
@peterpeterparker It updated the "chagned" screens now. I don't think that the one screen that has changed has something to do with the updated `git-components` lib but still the updated screen look correct to me.

peterpeterparker (approved):
Agree without. The single screenshot modified rather looks like a race condition in the E2E tests than changes related to the UI components.

LGTM, thx

PR #4545, opened 2025-02-03
feat(frontend): save cached SPL tokens to backend

AntonioVentilii:
# Motivation

Before starting to use the SPL Custom tokens backed, we first need to save the current browser-cached SPL tokens to the backend, to avoid any any disruptions to the user experience, as the user tokens are not yet stored in the backend.
However, it is acceptable to remove it after a certain amount of time, since we expect that most of the users that consider this part relevant will log in before that time.
So, as timeline, we define a period of 1 month, more or less, to remove this function and its usage. That is enough time for the users to log in and have their tokens saved in the backend.

# Changes

- Create temporary util to load the SPL cached in the browser (this part of code was already there, just moved) and upload them to the backend (this is a copy of how we do for ICRC Custom tokens).
- Note: we are using a for loop instead of a functional approach to avoid the async/await issue due to the limit of the Quicknode API calls. This is acceptable since this is just a temporary solution that shall be removed soon.
- Adapt service `loadUserTokens`.

# Tests

Current tests worked well.

### DIFF
--- a/src/frontend/src/sol/services/spl.services.ts
+++ b/src/frontend/src/sol/services/spl.services.ts
@@ -1,6 +1,8 @@
+import type { CustomToken } from '$declarations/backend/backend.did';
 import { SOLANA_MAINNET_NETWORK } from '$env/networks/networks.sol.env';
 import { SPL_TOKENS } from '$env/tokens/tokens.spl.env';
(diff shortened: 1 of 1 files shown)


DenysKarmazynDFINITY (approved):
Discussed the approach and the further plans offline. 

LGTM

PR #4549, opened 2025-02-03
feat(frontend): save SPL custom tokens to backend

AntonioVentilii:
# Motivation

Since we can now save SPL tokens in the backend, we use the common service `setManyCustomTokens` to set SPL Custom tokens.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4550, opened 2025-02-03
feat(frontend): re-use swap components in ConvertForm

DenysKarmazynDFINITY:
# Motivation

In this PR, ConvertForm starts using components that were extracted from the Swap flow. 

Screenshot 2025-02-03 at 15 49 58


src/frontend/src/lib/components/convert/ConvertForm.svelte
```diff
@@ -23,10 +29,12 @@
 		bind:receiveAmount
 		bind:insufficientFunds
 		bind:insufficientFundsForFee
+		bind:exchangeValueUnit
+		{inputUnit}
```
AntonioVentilii, line 33: does it make sense to put the `inputUnit` logic inside the component instead of passing it here? it seems you use it only there, no?

AntonioVentilii (approved):
Just one small note, for the rest LGTM

src/frontend/src/lib/components/convert/ConvertForm.svelte
DenysKarmazynDFINITY: totally, moving it down to `ConvertAmount`, thanks!

PR #4552, opened 2025-02-03
fix(script): redo fallback SNS governance logo if no ledger logo

peterpeterparker:
# Motivation

Deprecating the CTS project (#4331) led to skipping the fetching of icons for governance if no ledger icon was available in the SNS script. This is why we removed the fallback (see #4523). However, as discussed today in the [forum](https://forum.dfinity.org/t/icrc-metadata-inconsistency/40793/5?u=peterparker) post, while there might be an issue in the tooling, older SNSes (OC, Ghost, Yarl, Kinic, Yuku, Motoko, Alice, Gold, Catalyze, and Boom) do not have effectively logos in their ledger for historical reasons. Therefore, we should redo the fallback.

# Changes

- If no icon and not a deprecated SNS, then use the SNS (root) project logo

### DIFF
--- a/scripts/build.tokens.sns.mjs
+++ b/scripts/build.tokens.sns.mjs
@@ -54,8 +54,11 @@ const querySnsAggregator = async (page = 0) => {
 };
 
 const saveLogos = async (logos) => {
-	const writeLogo = async ({ icon, ledgerCanisterId }) => {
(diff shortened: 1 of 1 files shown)


AntonioVentilii (approved):
LGTM tks

PR #4553, opened 2025-02-03
feat(frontend): use SPL custom tokens from backend instead of the cached ones

AntonioVentilii:
# Motivation

We can switch to using the SPL custom tokens loaded from the backend instead of using the browser-cached one. However we keep loading them to save them with the temporary solution provided in https://github.com/dfinity/oisy-wallet/pull/4545

# Changes

- Loop through `splCustomTokens` loaded from backed.
- Split them into existing tokens (the ones hard-coded in our codebase) and non-existing one (the ones that the user add manually).
- Refresh the metadata of the second group, for insurance.

# Tests

Current tests are sufficient, but we will provide a more comprehensive one, once removed the temporary solution.

### DIFF
--- a/src/frontend/src/sol/services/spl.services.ts
+++ b/src/frontend/src/sol/services/spl.services.ts
@@ -1,5 +1,6 @@
 import type { CustomToken } from '$declarations/backend/backend.did';
-import { SOLANA_MAINNET_NETWORK } from '$env/networks/networks.sol.env';
+import { SOLANA_DEVNET_NETWORK, SOLANA_MAINNET_NETWORK } from '$env/networks/networks.sol.env';
+import { SOLANA_DEFAULT_DECIMALS } from '$env/tokens/tokens.sol.env';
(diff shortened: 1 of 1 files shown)


src/frontend/src/sol/services/spl.services.ts
```diff
@@ -155,32 +156,61 @@ export const loadUserTokens = async ({
...
+				const version = fromNullable(versionNullable);
+
+				const {
+					network: tokenNetwork,
+					symbol,
+					decimals,
+					token_address: tokenAddress
+				} = 'SplDevnet' in token
+					? { ...token.SplDevnet, network: SOLANA_DEVNET_NETWORK }
+					: { ...token.SplMainnet, network: SOLANA_MAINNET_NETWORK };
```
bitdivine, line 178: Do you want some kind of fail-safe, in case this is missed when more Spl networks are added?

bitdivine (approved):

PR #4559, opened 2025-02-03
feat(frontend): fetch metadata for SPL devnet/testnet tokens too

AntonioVentilii:
# Motivation

We will fetch SPL token data for Solana devnet and testnet tokens too.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4560, opened 2025-02-03
refactor(frontend): keep caching only SPL mainnet tokens

AntonioVentilii:
# Motivation

Just to have a clean and smooth cache, we will be saving in the browser cache only SPL mainnet tokens, since the devnet and the testnet are not really relevant for the user experience, so they can be "lost" in the token list when the new version is released.


DenysKarmazynDFINITY (approved):
👍

PR #4565, opened 2025-02-04
feat(frontend): create Solana transaction mapper for parsed instructions

AntonioVentilii:
# Motivation

We are receiving parsed instructions from Solana RPC (as much as it is possible). So it makes sense that we start mapping them.

NOTE: if we cannot correctly map it, we will return undefined, since it is a transaction that we are not interested in showing.

# Changes

- Create mapper for System program instructions: this is just the `transfer` instruction, pretty simple.
- Create mapper for Token program Instructions: this requires a call to the RPC to identify what is the wallet address owner of the ATA in the `source` and in the `destination`, and the token address.
- Create mapper for ATA program instruction: this is only for the creation of ATA, and it requires a mapper inside the inner instructions, to identify both the payer and the fee paid.
- Merge all mappers into one.

# Tests

Added tests.


src/frontend/src/sol/utils/sol-instructions.utils.ts
```diff
@@ -66,8 +73,173 @@ import {
...
+				data: {
+					parsed: { info: destinationAccoutInfo }
+				}
+			} = destinationResult;
+
+			const { owner: to } = destinationAccoutInfo as {
+				owner: SolAddress;
+			};
+
+			return { value: BigInt(value), from, to, tokenAddress };
```
DenysKarmazynDFINITY, line 160: Will `value` always be a stringified integer? Asking because of the potential non-integer conversion error if it can be a float.

DenysKarmazynDFINITY (approved):

src/frontend/src/sol/utils/sol-instructions.utils.ts
AntonioVentilii: for this specific transaction type that is what it is provided: a stringified integer referring to the minimal unit of the token. I tested with a transaction of 1.23456 JUP and it came back as string integer indeed.

DenysKarmazynDFINITY: cool, thanks for the explanation!

PR #4566, opened 2025-02-04
refactor(frontend): use Solana signature to hash a transaction

AntonioVentilii:
# Motivation

A Solana signature can be associated to a number of transactions. So, we will change the ID of the transaction objects that we use in OISY. But before that, we need to provide the signature to all the Solana transactions, so that we can open correctly the explorer.

# Changes

- Add `signature` prop to `SolRpcTransaction` type.
- Provide the prop when the transaction is built.
- Adapt the usages.

# Tests

No changes in the replica and the current tests are sufficient.

Screenshot 2025-02-04 at 10 52 25


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4570, opened 2025-02-04
refactor(frontend): set absolute amount for Solana transaction

AntonioVentilii:
# Motivation

To be consistent with the other networks, we save the amount of a Solana transaction in absolute value and we put the sign ONLY when showing it in the transaction list.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4573, opened 2025-02-04
docs(security): Add the default Dfinity security policy

bitdivine:
# Motivation
Dfinity has developed a default security policy for GitHub repositories.

# Changes
- Add the SECURITY.md from the security template: https://github.com/dfinity/security-policy-template/blob/main/SECURITY.md

# Tests
N/A


AntonioVentilii (approved):
Tks!

PR #4574, opened 2025-02-04
refactor(frontend): move getSolTransactions to services

AntonioVentilii:
# Motivation

It makes more sense that the method `getSolTransactions` is a service instead of being inside the API module.


DenysKarmazynDFINITY (approved):
The refactoring LGTM. There are some failing tests to be fixed, approving already though so you can proceed faster afterwards.

PR #4572, opened 2025-02-04
feat(frontend): service to fetch Solana transaction and map them

AntonioVentilii:
# Motivation

After PR https://github.com/dfinity/oisy-wallet/pull/4565 , we can create a service to fetch the transactions related to a specific signature and map them to `SolTransactionUi` type.

# Changes

- Create service `fetchSolTransactionsForSignature` that first fetches the transaction details to a signature.
- Break down the details into instructions and parse them back to transaction list.

# Tests

Included tests.

PR #4577, opened 2025-02-04
feat(frontend): use new Solana transaction service

AntonioVentilii:
# Motivation

We can use the service `fetchSolTransactionsForSignature` introduced in https://github.com/dfinity/oisy-wallet/pull/4572 , to correctly fetch and parse instructions into transactions.

# Changes

- Move service `getSplTransactions` in the same module as `getSolTransactions`.
- For both of them use fetchSolTransactionsForSignature instead of the old method.
- The step above will provide a return of types `SolTransactionUi` directly, so we adapt the types. 

# Tests

### DIFF
--- a/src/frontend/src/sol/services/sol-signatures.services.ts
+++ b/src/frontend/src/sol/services/sol-signatures.services.ts
@@ -1,10 +1,11 @@
 import { WALLET_PAGINATION } from '$lib/constants/app.constants';
-import { fetchSignatures, fetchTransactionDetailForSignature } from '$sol/api/solana.api';
+import type { SolAddress } from '$lib/types/address';
+import { fetchSignatures } from '$sol/api/solana.api';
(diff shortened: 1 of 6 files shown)


src/frontend/src/sol/services/sol-signatures.services.ts
```diff
@@ -15,24 +16,62 @@ export const getSolTransactions = async ({
...
+				network,
+				address
+			});
+
+			return [...acc, ...parsedTransactions];
+		},
+		Promise.resolve([] as SolTransactionUi[])
+	);
+
+	return transactions.slice(0, limit);
```
DenysKarmazynDFINITY, line 38: Is there a case where there will be more txs fetched than `limit`? I saw on the line 22 you're passing `limit` to `fetchSignatures`.

DenysKarmazynDFINITY (approved):

PR #4579, opened 2025-02-04
build(frontend): bump ic-js

AntonioVentilii:
# Motivation

We want to use `AgentManager` of `@dfinity/utils` but with the last version updated with https://github.com/dfinity/ic-js/pull/836 .
So, we bump all `ic-js` dependencies to the `next` version.


peterpeterparker (approved):
LGTM, thx

PR #4580, opened 2025-02-04
feat(frontend): map Solana transaction errors

AntonioVentilii:
# Motivation

We want to provide a different message for Solana tokens when the transaction expires. This happens because the transaction was not picked up by any block, most probably because the priority fee was not enough.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4581, opened 2025-02-04
ci(frontend): update SNS script is Typescript

AntonioVentilii:
# Motivation

Since we would like to have the deprecated SNS tokens as object usable both from the script and for the program, as first step, we transform the script `update.tokens.sns` into a Typescript file.

# Changes

- Make the script a Typescript file.
- Adapt the types.
- Adjust `package.json`

# Tests

I ran the CI on this PR and it worked correctly: https://github.com/dfinity/oisy-wallet/actions/runs/13147079102


scripts/build.tokens.sns.ts
```diff
@@ -21,6 +26,29 @@ const AGGREGATOR_URL = `${SNS_AGGREGATOR_CANISTER_URL}/${AGGREGATOR_CANISTER_VER
...
 const STATIC_FOLDER = join(process.cwd(), 'src', 'frontend', 'static', 'icons', 'sns');
 
+interface ResponseData {
+	canister_ids: {
+		ledger_canister_id: CanisterIdText;
+		index_canister_id: CanisterIdText;
+		root_canister_id: CanisterIdText;
+	};
+	icrc1_metadata: [[string, { Text: string } | { Nat: [number] }]];
+	meta: { name: string; url: z.infer<typeof UrlSchema> };
```
peterpeterparker, line 36: maybe for another PR, should we would not infer the type in this scripts but add a module to `$lib/types` with those types?

side note: we should create a new library for the SNS aggregator in ic-js as we start using types NNS dapp and OISY. We could for example type everything with Zod this way it would also be possible to validate the responses since we use fetch but, just a though, definitely a side quest.

scripts/build.tokens.sns.ts
```diff
@@ -104,30 +132,33 @@ const mapOptionalToken = (response) => {
...
-		isNullish(nullishToken.symbol) ||
-		isNullish(nullishToken.name) ||
-		isNullish(nullishToken.fee) ||
-		isNullish(nullishToken.decimals)
-	) {
-		return undefined;
+	const { symbol, name, fee, decimals, ...rest } = nullishToken;
+
+	if (isNullish(symbol) || isNullish(name) || isNullish(fee) || isNullish(decimals)) {
+		throw new Error(`Missing token metadata: ${JSON.stringify(response)}`);
```
peterpeterparker, line 138: Why are you changing the behavior - i.e. any particular reason to throw an error instead of returning the partial token?

One or the other, probably better to propose this in another PR given that this one has for goal to convert to typescript?

AntonioVentilii: Ah, yes, forgot to mention in the description: this is a funny one that Typescript showed.

We were returning `undefined` in case one of the props is `nullish`. However, type-wise, it is wrong when we use the function in 

```typescript
({
	ledgerCanisterId: ledger_canister_id,
	indexCanisterId: index_canister_id,
	rootCanisterId: root_canister_id,
	metadata: {
		...mapOptionalToken(icrc1_metadata),
		alternativeName: alternativeName.trim(),
		url
	}
})
```

because the type of this object is given by schema `EnvSnsTokenSchema` once we use Typescript. And in that schema, the `metadata` provided by `mapOptionalToken` MUST be `non-nullish`. So, one solution could e the one I did, or another one is to fallback on empty metadata, i.e. `mapOptionalToken(icrc1_metadata) ?? {symbol: '', decimals: 0, ... }`

WDYT? Because as it is now, it basically ignores that they could have no metadata? Shall I change the schema, or add a new schema with partial metadata? But it is a cascade after eheh

scripts/build.tokens.sns.ts
AntonioVentilii: yes, you are right... we actually have a folder in `$env/types` (not really sure why it's there). I used that in the last commit.

Shall i move it to `$lib/types`? (in another PR)

scripts/build.tokens.sns.ts
peterpeterparker: 👉 #4587

AntonioVentilii: Cool! So, after that PR, I just need to make `metadata` prop optional in `EnvSnsTokenSchema`, correct?

peterpeterparker: Yes :)

peterpeterparker: Merged

AntonioVentilii: Rebased and changed the types: have a look, i had to create a sub-function to filter because the type was not being recognized as non-nullish `metadata`

peterpeterparker: @AntonioVentilii it should return `undefined` instead of throwing an error no?

```
throw new Error(`Missing token metadata: ${JSON.stringify(response)}`);
```

=>

```
return undefined
```

Or am I missing something?

AntonioVentilii: yes, indeed: i see it there, no?

AntonioVentilii: my mistake... i forgot to push the last commit, sec

AntonioVentilii: pushed! unfortunately I still need to deconstruct... it is giving me some type issue saying that `nullishToken` is not what we required, even after the nullish checks... strange

peterpeterparker (approved):
LGTM, thx

PR #4584, opened 2025-02-05
chore(rewards): Use newer rewards

bitdivine:
# Motivation
We would like to add the option of registering users for an airdrop.

# Changes
- Use a rewards canister with a placeholder airdrop API registration.

# Tests
CI should update the bindings.


AntonioVentilii (approved):
LGTM tks!

PR #4585, opened 2025-02-05
chore(template): add contact to github issue options

peterpeterparker:
# Motivation

As discussed with @StefanBerger-DFINITY, we can add Discord in the list of option when one open an issue on GitHub.


StefanBerger-DFINITY (approved):
LGTM, as discussed in slack.

PR #4587, opened 2025-02-05
fix(script): handle undefined token metadata

peterpeterparker:
# Motivation

As discussed in https://github.com/dfinity/oisy-wallet/pull/4581#discussion_r1942355892, the was undefined token metadata is handled in the script that generate the SNS environment data is not exactly correct as it desctruct undefined token metadata but, would process further anyway.

# Changes

- If no token metadata then metadata is undefined


AntonioVentilii (approved):
Tks!!!

PR #4588, opened 2025-02-05
chore(frontend): remove incorrect type casting

AntonioVentilii:
# Motivation

We were providing the Solana networks IDs to service `loadSolAddress` and casting it as `TokenId`. However, it makes more sense that we pass the Token IDs of the native tokens for each network (as we do for the other networks.

This ID is simply used as verbose error, it has no use in the logic of loading the address per-se:

```typescript
		toastsError({
			msg: {
					$symbol: tokenId.description ?? ''
				})
			},
			err
		});
```


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4590, opened 2025-02-05
fix(e2e): correct the download path of the backend wasm

inc-man:
# Motivation

There was an issue in the e2e test workflow, where the backend wasm was downloaded to the wrong location.

# Changes

Fixed the location path for the wasm download.


bitdivine (approved):
Thanks!

PR #4592, opened 2025-02-05
docs(security): updates formatting of the security policy

jwndlng:
# Motivation

The "Important" block is not displayed properly.

# Changes

Adds a line-break to fix the formatting.

# Tests

Not needed. It only changes the documentation.


AntonioVentilii (approved):
LGTM tks

PR #4593, opened 2025-02-05
chore(frontend): add types to derived stores of Kong swap ICRC tokens

AntonioVentilii:
# Motivation

Just to have more manageability, we explicitly set the types of a few derived stores on ICRC tokens that pertains to Kong swap


BonomoAlessandro (approved):
LGTM!

PR #4594, opened 2025-02-05
feat(frontend): Prepared Components and Store for Color Mode Selection

daviddecentage:
# Motivation

As user I can set the color mode in settings. This PR serves as a prerequisite for a theme-system refactoring, focusing only on the selector elements and saving the selected color mode to a store.

As defined, system will be the default (defined in constants).

# Changes

- Added ThemeSelector Component
- Added ThemeSelectorCard Component
- Added themeStore
- Added constant for default theme
- Added enum for available themes
- Added method to get selected theme to derived settings store
- Added images used for the ThemeSelectorCard component
- Prepared i18n messages

# Tests

No tests yet. To avoid making the feature available, the settings page will be updated to include the switcher once the theme refactoring has been merged. This is how the components look:

image


src/frontend/src/lib/components/settings/ThemeSelector.svelte
```diff
@@ -0,0 +1,27 @@
...
+		<ThemeSelectorCard
+			label={$i18n.settings.text[`appearance_${theme}`]}
+			selected={selected === theme}
+			on:click={selectTheme(theme)}
+			on:keydown={selectTheme(theme)}
+			tabindex={-1}
+		>
+			{#await import(`$lib/assets/${theme}-theme.png`) then { default: src }}
+				<img {src} alt={theme} />
+			{/await}
```
AntonioVentilii, line 24: - this can be imported directly as images I think, no need to import them directly in the html part.

- we have the `Img` component that is useful in these cases

- maybe the aria label should be a little more verbose?

src/frontend/src/lib/components/settings/ThemeSelector.svelte
```diff
@@ -0,0 +1,27 @@
...
+	import { i18n } from '$lib/stores/i18n.store';
+	import { themeStore } from '$lib/stores/settings.store';
+	$: selected = $selectedTheme;
+	const selectTheme = (name: Themes) => () => {
+		themeStore.set({ key: 'theme', value: { name } });
+	};
+</script>
+
+<div class="flex flex-row">
+	{#each Object.values(Themes) as theme}
```
AntonioVentilii, line 14: I think we will need this list somewhere else: may you please put it somewhere as exported constant?

src/frontend/src/lib/components/settings/ThemeSelectorCard.svelte
```diff
@@ -0,0 +1,31 @@
...
+	export let tabindex: number;
+</script>
+
+<div
+	class="mr-4 cursor-pointer"
+	on:click
+	on:keydown
+	role="radio"
+	aria-checked={selected}
+	{tabindex}
```
AntonioVentilii, line 13: this seems to be called always as `{-1}` so maybe we don't need to have it as prop for now?

src/frontend/src/lib/components/settings/ThemeSelector.svelte
daviddecentage: I did it this way to keep it generically and have a single source for the available themes (Themes enum). I think it makes sense if we decide to add another theme we can just add the image, extend the enum and add a new SCSS file for the new theme.

I can also hardcode it, no problem either way..

src/frontend/src/lib/components/settings/ThemeSelector.svelte
daviddecentage: Not sure I understand you right here. The enum is already exported, you mean an additional constant which exports the enums values?

src/frontend/src/lib/components/settings/ThemeSelectorCard.svelte
daviddecentage: I think we need it because it is a div with role="radio". I will assign a unique value for it 👍

daviddecentage:
> May you put the store (and derived) creation in another PR, before this one?

Its nessessary to have for the selector component. I would have to move it out as well. Up to you

src/frontend/src/lib/components/settings/ThemeSelector.svelte
AntonioVentilii: All good for the import then, thank you for the explanation; however maybe let's use `Img` component that we have instead of `img`

src/frontend/src/lib/components/settings/ThemeSelector.svelte
AntonioVentilii: yes, correct! pretty sure we will use it more, so it makes sense to have it already somewhere

AntonioVentilii:
> Its nessessary to have for the selector component. I would have to move it out as well. Up to you


what i meant is: let's do first a PR that creates the store and the derived

and then, after merged, rebase this one and proceed

src/frontend/src/lib/components/settings/ThemeSelectorCard.svelte
AntonioVentilii: indeed! that is what I meant, instead of passing as prop, you can already define `tabindex={-1}` inside component `ThemeSelectorCard` directly, no?

daviddecentage: I have updated it to assign a unique value. I think that should be cleaner instead of using tabindex just for the sake of it

src/frontend/src/lib/components/settings/ThemeSelector.svelte
```diff
@@ -0,0 +1,29 @@
...
+<div class="flex flex-row">
+	{#each THEME_VALUES as theme}
+		<ThemeSelectorCard
+			label={$i18n.settings.text[`appearance_${theme}`]}
+			selected={selected === theme}
+			on:click={selectTheme(theme)}
+			on:keydown={selectTheme(theme)}
+			tabindex={THEME_VALUES.indexOf(theme)}
+		>
+			{#await import(`$lib/assets/${theme}-theme.png`) then { default: src }}
```
peterpeterparker, line 24: I'm not going to review the PR, but I have a question about this pattern—which looks cool, btw.

1. What happens if an image is missing or fails to load? Does Svelte’s `await import` handle the potential error, or is it unhandled? If it's the latter, shouldn't we explicitly handle it?

2. I don't remember—does a production build of SvelteKit/Vite inline assets imported within the code, or does the bundler copy them to the output directory and replace the reference with a string pointing to the file? If it's the latter, we're fine. But if it's the former, it would be an issue, as it would increase the bundle size. Since we're not chunking the main bundle at this time, we probably want to avoid that.

daviddecentage: `What happens if an image is missing or fails to load? Does Svelte’s await import handle the potential error, or is it unhandled? If it's the latter, shouldn't we explicitly handle it?`

The missing image will lead to no error, silently not rendering the image. I would need to check if we can handle the error in a way that lets us just not render the whole Card to select the theme at all and print a nice warning to the console as a hint for Devs.

`I don't remember—does a production build of SvelteKit/Vite inline assets imported within the code, or does the bundler copy them to the output directory and replace the reference with a string pointing to the file? If it's the latter, we're fine. But if it's the former, it would be an issue, as it would increase the bundle size. Since we're not chunking the main bundle at this time, we probably want to avoid that.`

I spent some time analyzing the behavior and discovered several things. It seems like SvelteKit/Vite handle the import like any other 'static' image import. However, at first I didnt see it in the build output.

The build output asset folder shows theres currently around 20 files (mostly svg and some png). I played around with the vite config and set "build.assetsInlineLimit" to 0 to not inline any assets at all.

Another build after changing this vite config now produces approx. 100 files in the asset folder, which indicates to me that a lot of other assets (as well as these 3 that I added) are being inlined. I assume this is something we would need to address?

Tldr: Vite config "build.assetsInlineLimit" is responsible on how to handle asset inlining, and dynamic imports as per my PR don't behave different compared to normal image imports.

peterpeterparker: > The missing image will lead to no error, silently not rendering the image.

All good here then. No error, no problem. ✅

> I spent some time analyzing the behavior and discovered several things....

Are you sure that the way we are importing images in this PR using dynamic imports actually works in a production build?

I checked out your branch locally and ran `npm run build`. Analyzing the generated JS code, I found:

```
Object.assign({"../../assets/dark-theme.png":()=>gt(()=>Promise.resolve().then(()=>HTe),void 0,import.meta.url),"../../assets/light-theme.png":()=>gt(()=>Promise.resolve().then(()=>jTe),void 0,import.meta.url),"../../assets/system-theme.png":()=>gt(()=>Promise.resolve().then(()=>KTe),void 0,import.meta.url)}),`../../assets/${r[4]}-theme.png`
```

However, unlike the statically imported SVG - which are inlined - and WEBP files - which are copied over - referenced from the assets folder, there is no related PNG copied to the build output. I double checked visually and run a `find . | grep png` in the folder.

I might have missed something, so it might be worth deploying the branch to a test environment but, if this finding is correct, we might need to either fix the dynamic import to make it compatible with the production build or use static imports.

daviddecentage: To me it seeems the code snippet from the output you showed is just a reference to the manifest, as the hashes arent included and would be needed to correctly reference the file. Even with a static import these 3 images dont make it to the built output either, with the assets being referenced exactly as in the snippet you shared. I think we would really need to define the vite config on when to inline.

However, youre right that I cannot say with certainty that it behaves exactly like a static import. I guess we could just try it out on a test env as you suggested.

peterpeterparker: Right, I agree on the import—good point. 👍

Alright, I think they are not copied to the output build folder because they are smaller than a certain size and therefore are automatically inlined within the JS code. No need to deploy to a test environment; I just tested it locally (`npm run build + npx serve build`), and indeed, the result is a `data:image` (see screenshot).

This answers my questions. So technically, it's all good.

Personally, even if it's really just 3-5 KB, given that we currently ship a single chunk, along with the fact that these images are only used on the settings page - i.e. not useful for most sessions, I would prefer not to have them within the bundle. But I'll let you guys decide what you prefer.

Thanks for the answers.

Capture d’écran 2025-02-06 à 09 56 48

peterpeterparker: And also because having all images inlined in a single chunk kind of defeats the purpose of using dynamic imports but, we plan to improve performance so we might start chunking.

src/frontend/src/lib/i18n/en.json
```diff
@@ -214,7 +214,14 @@
...
 			"present_pouh_credential": "Present",
-			"pouh_credential_verified": "Verified ✅"
+			"pouh_credential_verified": "Verified ✅",
+			"appearance": "Appearance",
+			"appearance_light": "Light",
+			"appearance_light_description": "Enabled light mode",
+			"appearance_dark": "Dark",
+			"appearance_dark_description": "Enables dark mode",
+			"appearance_system": "System",
+			"appearance_system_description": "Enables your operating systems color mode"
```
AntonioVentilii, line 224: just for consistency, let's put them all the `_description` in `alt` section

src/frontend/src/lib/components/settings/ThemeSelector.svelte
```diff
@@ -0,0 +1,29 @@
...
+	import Img from '$lib/components/ui/Img.svelte';
+	import { THEME_VALUES } from '$lib/constants/app.constants';
+	import { selectedTheme } from '$lib/derived/settings.derived';
+	import { Themes } from '$lib/enums/themes';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { themeStore } from '$lib/stores/settings.store';
+	$: selected = $selectedTheme;
+	const selectTheme = (name: Themes) => () => {
+		themeStore.set({ key: 'theme', value: { name } });
+	};
```
AntonioVentilii, line 12: - why the double-function here? can it be directly:

```typescript
const selectTheme = (name: Themes) => {
		themeStore.set({ key: 'theme', value: { name } });
	};
```


- why the variable selected? can the `$selectedTheme` be used directly?

daviddecentage: First function is to pass the theme name, second one is the event handler for the click.

AntonioVentilii: 1. Ah! I understand your point. Because the component calls `...={selectTheme(theme)}`. Then, for consistency, isn't it better to call `...={() => selectTheme(theme)}`? and consequently adjust the `selectTheme` function? No strong opinion, just asking

daviddecentage: 1. I inlined the event handler, youre right its more clear than the double func 👍
2. Regarding the selected variable, Im relatively new to svelte and copied this from some other place in the code. I removed it and reference it directly now.

src/frontend/src/lib/i18n/en.json
daviddecentage: Done 👍

AntonioVentilii (approved):
LGTM tks!

PR #4595, opened 2025-02-05
chore(frontend): add type parameter to pinTokensWithBalanceAtTop

AntonioVentilii:
# Motivation

Since util `pinTokensWithBalanceAtTop` is being used with lists of specific types of Tokens, it is useful to make the type of such object parametrizable, especially for the return.


BonomoAlessandro (approved):
LGTM!

PR #4596, opened 2025-02-05
chore(frontend): specify type of tokens in swap context

AntonioVentilii:
# Motivation

In all the Swap components, the correct type for the destination tokens and the source token is `IcTokenToggleable`.


BonomoAlessandro (approved):
LGTM

PR #4600, opened 2025-02-05
feat(frontend): map Solana transactions that are of type closeAccount

AntonioVentilii:
# Motivation

The instruction `closeAccount` provides a return of the rent-free amount paid when opening an Associated Token Account. So, it should be mapped among the relevant transactions that we want to show.

# Changes

- Create mapper for `closeAccount`.
- NOTE: there is still no current way of knowing exactly the amount received, so for now we will show zero, because at least we can provide a link to the transaction.

# Tests

Included tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4602, opened 2025-02-05
fix(frontend): use signature instead of ID for Solana loader

AntonioVentilii:
# Motivation

Since we changed the ID of a Solana transaction to be more complex, it is correct to use the `signature` to ask the loader for the next N transactions.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4604, opened 2025-02-05
feat(frontend): map Solana inner instructions too

AntonioVentilii:
# Motivation

It may happen that some Solana instructions inside a transaction cannot be easily decoded by the RPC. For example, when we use Solana programs whose IDL we don't know, like some DEX swap (for example [this Jupiter swap aggregator](https://solscan.io/tx/5nBws1cGkKBCsPE7AtavcXFs79gg22SYAKNDNf7VRoc1dvmJEF6G6CQLzFncp5G7H6WQZhvFj1NJhPPdtKRo7RX9)).

However, such instructions may have inner instructions that we can use and map: in the example provided there are a few useful `Transfer` instructions inside `Jupiter Aggregator v6: route`


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4606, opened 2025-02-05
feat(frontend): filter fetched signature by ATA account if necessary

AntonioVentilii:
# Motivation

To improve the performance of the requests and queries that we do to the Solana RPC, we can filter the requested signatures using directly the ATA account address in case of an SPL token.

Before, we were searching for all the signatures associated to the main (owner) account, but it makes more sense to filter down.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4611, opened 2025-02-06
fix(e2e): avoid automatic scrolling for modals and flakiness

AntonioVentilii:
# Motivation

When we test the Receive modal, and we try and click the Receive button, the method `click` auto-scrolls into view until it is able to click the element.
This can cause quite a volatility in the screenshots if the time to load is different and sometimes it scrolls up and down, giving different images.

In fact, an example of such flakiness is PR https://github.com/dfinity/oisy-wallet/pull/4607 . If we analyse the log of the trace that modified the initial screenshot (showing a scrolled OISY), we can see that indeed it scrolls trying to click:

```
waiting for getByTestId('receive-tokens-modal-open-button')
locator resolved to <button type="submit" aria-label="Receive" data-tid="receive-tokens-modal-open-button" class="tertiary flex flex-1 text-center padding-sm link">…</button>
attempting click action
waiting for element to be visible, enabled and stable
element is not stable
retrying click action
waiting for element to be visible, enabled and stable
element is not stable
retrying click action
waiting 20ms
waiting for element to be visible, enabled and stable
element is visible, enabled and stable
scrolling into view if needed
done scrolling
<html lang="en">…</html> intercepts pointer events
retrying click action
waiting 100ms
waiting for element to be visible, enabled and stable
element is visible, enabled and stable
scrolling into view if needed
done scrolling
performing click action
click action done
waiting for scheduled navigations to finish
navigations have finished
```

Screenshot 2025-02-06 at 01 17 48

# Solution

Apparently there is no parameter to avoid this, but we can use the suggestions from the playwright issues ([here](https://github.com/microsoft/playwright/issues/11716#issuecomment-1025743316) and [here](https://github.com/microsoft/playwright/issues/11716#issuecomment-1025743316)) that suggest to simply dispatch the `click` event instead of using the method.

# Changes

- Add optional parameter `scrollIntoView` (defaults to `true`, that is the current behaviour) to method `clickByTestId`.
- The parameter will decide if we either use the `click` method (as before) or dispatch the event `click` to the locator.
- Use the latter when we want to await the modal.
- Adapt the rest of the code.

# Tests

The E2E tests and snapshots CIs ran on this PR and they set correctly the screenshots. Furthermore, it was already tested a few times successfully in PR https://github.com/dfinity/oisy-wallet/pull/4607


inc-man (approved):
LGTM, tks

PR #4613, opened 2025-02-06
test(e2e): extend homepage for new test

inc-man:
# Motivation

For clarity this PR will only focus on the new methods and functions that are created inside the homepage page object. These will later be used in the new send and receive test.

# Changes

added multiple new methods and functions

### DIFF
--- a/e2e/utils/pages/homepage.page.ts
+++ b/e2e/utils/pages/homepage.page.ts
@@ -1,4 +1,5 @@
 import {
+	AMOUNT_DATA,
 	LOADER_MODAL,
 	LOGIN_BUTTON,
 	LOGOUT_BUTTON,
@@ -38,9 +39,15 @@ interface TestIdOperationParams {
 	testId: string;
 }
 
+interface NavigateToTokenParams {
+	token: string;
+	network: string;
+}
+
 interface WaitForModalParams {
 	modalOpenButtonTestId: string;
 	modalTestId: string;
+	state?: 'detached';
 }
 
 type TestModalSnapshotParams = {
@@ -70,6 +77,10 @@ abstract class Homepage {
 		await this.#page.getByTestId(testId).click();
 	}
 
+	protected async waitForByTestId(testId: string): Promise<void> {
+		await this.#page.getByTestId(testId).waitFor();
+	}
+
 	protected async isVisibleByTestId(testId: string): Promise<boolean> {
 		const element = this.#page.locator(`[data-tid="${testId}"]`);
 		return await element.isVisible();
@@ -91,6 +102,15 @@ abstract class Homepage {
 		if (await this.isSelectorVisible({ selector })) {
 			await this.#page.locator(selector).evaluate((element) => (element.innerHTML = 'placeholder'));
 		}
+	}
+
+	protected async mockSelectorAll({ selector }: SelectorOperationParams): Promise<void> {
+		const elementsLocator = this.#page.locator(selector);
+		await elementsLocator.evaluateAll((elements) => {
+			for (const element of elements) {
+				(element as HTMLElement).innerHTML = 'placeholder';
+			}
+		});
 	}
 
 	private async goto(): Promise<void> {
@@ -135,12 +155,17 @@ abstract class Homepage {
 
 	protected async waitForModal({
 		modalOpenButtonTestId,
-		modalTestId
+		modalTestId,
+		state
 	}: WaitForModalParams): Promise<Locator> {
-		await this.clickByTestId(modalOpenButtonTestId);
+		
 		const modal = this.#page.getByTestId(modalTestId);
+		if (state === 'detached') {
+			await modal.waitFor({ state: 'detached' });
+			return modal
+		}
+		await this.clickByTestId(modalOpenButtonTestId);
 		await modal.waitFor();
-
 		return modal;
 	}
 
@@ -190,6 +215,34 @@ abstract class Homepage {
 
 	async waitForLoggedInIndicator(): Promise<void> {
 		await this.#page.getByTestId(NAVIGATION_MENU_BUTTON).waitFor();
+	}
+
+	protected async elementExistsByTestId(testId: string): Promise<boolean> {
+		return await this.#page
+			.getByTestId(testId)
+			.isVisible()
+			.catch(() => false);
+	}
+
+	getBalance(): Locator {
+		return this.#page.getByTestId(AMOUNT_DATA);
+	}
+
+	async setInputValueByTestId({
+		testId,
+		value
+	}: TestIdOperationParams & { value: string }): Promise<void> {
+		await this.#page.getByTestId(testId).fill(value);
+	}
+
+	async getAccountIdByTestId(testId: string): Promise<string> {
+		const container = this.#page.locator('#icp-account-id');
+		const addressLocator = container.getByTestId(testId);
+		const addressText = await addressLocator.textContent();
+		if (!addressText) {
+			throw new Error('No address text found in container icp-account-id');
+		}
+		return addressText.trim();
 	}
 
 	async testModalSnapshot({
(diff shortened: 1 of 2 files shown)


e2e/utils/pages/homepage.page.ts
```diff
@@ -70,6 +77,10 @@ abstract class Homepage {
 		await this.#page.getByTestId(testId).click();
 	}
 
+	protected async waitForByTestId(testId: string): Promise<void> {
+		await this.#page.getByTestId(testId).waitFor();
+	}
```
AntonioVentilii, line 82: i see that this syntax is used all over, passing options too;

may you please do a separate PR where you introduce this function and use it in all the possible cases?

e2e/utils/pages/homepage.page.ts
```diff
@@ -192,6 +217,34 @@ abstract class Homepage {
 		await this.#page.getByTestId(NAVIGATION_MENU_BUTTON).waitFor();
 	}
 
+	protected async elementExistsByTestId(testId: string): Promise<boolean> {
+		return await this.#page
+			.getByTestId(testId)
+			.isVisible()
+			.catch(() => false);
+	}
```
AntonioVentilii, line 225: the above I am pretty sure is doable via wait for and passing the option `'visible'`

and there is no need to catch the error in that case

e2e/utils/pages/homepage.page.ts
```diff
@@ -192,6 +217,34 @@ abstract class Homepage {
...
+
+	async setInputValueByTestId({
+		testId,
+		value
+	}: TestIdOperationParams & { value: string }): Promise<void> {
+		await this.#page.getByTestId(testId).fill(value);
+	}
+
+	async getAccountIdByTestId(testId: string): Promise<string> {
+		const container = this.#page.locator('#icp-account-id');
```
AntonioVentilii, line 239: shouldn't this ID be a constant variable somewhere?

e2e/utils/pages/homepage.page.ts
inc-man: used all over in what manner? In this PR it is simply introduced as new.

e2e/utils/pages/homepage.page.ts
inc-man: the account id? I mean its most likely just going to be used here and I used the ID that was to hand. Otherwise one should create an id that is widely useable.

e2e/utils/pages/homepage.page.ts
```diff
@@ -151,12 +171,17 @@ abstract class Homepage {
...
+		state
 	}: WaitForModalParams): Promise<Locator> {
-		await this.clickByTestId({ testId: modalOpenButtonTestId, scrollIntoView: false });
+    
 		const modal = this.#page.getByTestId(modalTestId);
+		if (state === 'detached') {
+			await modal.waitFor({ state: 'detached' });
+			return modal
+		}
+		await this.clickByTestId({ testId: modalOpenButtonTestId, scrollIntoView: false });
```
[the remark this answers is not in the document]
inc-man: we dont click it after. Later in the test I will do waitForModal(state: detached) and if that is the case, the modal has already been opened. Thats why state is optional. So to begin when we want to open the modal we don't pass a state. so it will automatically go to the clickByTestId to open it.

e2e/utils/pages/homepage.page.ts
inc-man: is it really necessary? I dont see a problem with using it this way since its "safer" in my opinion.

e2e/utils/pages/homepage.page.ts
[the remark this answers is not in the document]
AntonioVentilii: ah! understood! thank you for the explanation!

e2e/utils/pages/homepage.page.ts
AntonioVentilii: i mean `icp-account-id` is an hardcoded string here and it doesn't seem correct, since it refers specifically to ICP and since it is a test ID, most probably used in some component

e2e/utils/pages/homepage.page.ts
AntonioVentilii: yes, better to avoid repetition of code and use the methods that already exists: they make for easier manageability

e2e/utils/pages/homepage.page.ts
AntonioVentilii: the syntax `#page.getByTestId(testId).waitFor()` is used in a lot of other pages/tests

since we are introducing a specific method for it, to avoid repetition of code, we should use it in all instances. In this case, better to do it in a dedicated PR

AntonioVentilii: This one:

- https://github.com/dfinity/oisy-wallet/pull/4614

e2e/utils/pages/homepage.page.ts
AntonioVentilii: in fact, i this you can use the method you created here:

protected async elementExistsByTestId(testId: string): Promise<boolean> {
		return await this.waitForByTestId({testId, options: {state:'visible'}})
	}

inc-man: lets just delete the elementExistsByTestId and use the already present:

	protected async isVisibleByTestId(testId: string): Promise<boolean> {
		const element = this.#page.locator(`[data-tid="${testId}"]`);
		return await element.isVisible();
	}
	
agreed?

e2e/utils/pages/homepage.page.ts
inc-man: I kinda had to do it that way because when using the testId for the address-label it didn't work. But if we can set a dynamic testId for the container no matter what token, that would be even better.

inc-man: and actually I am not quite sure if the script works for anything else. But in that scenario we want to copy the Adress under "ICP Account". This does not exist for any other token I think.

e2e/utils/pages/homepage.page.ts
AntonioVentilii: why not use directly `waitForByTestId({testId, options: {state:'visible'}})` ?

inc-man: same same really, no? Both already exist, so no code duplication. Both do the same thing, isVisibleByTestId just does it a bit more clearly imo.

AntonioVentilii: ah wow! i didn't know they both existed, my mistake! I thought you were introducing it here. But then no, you are right, same same, your choice. In another PR maybe we should check if it possible to use `waitForByTestId` inside `isVisibleByTestId`

e2e/utils/pages/homepage.page.ts
AntonioVentilii: What i meant is, why not having a constant? `export const ACCOUNT_ID_ICP = 'icp-account-id'` and then `const container = this.#page.locator('ACCOUNT_ID_ICP);` or something like this

inc-man: done

e2e/utils/pages/homepage.page.ts
```diff
@@ -215,6 +230,34 @@ abstract class Homepage {
...
+	protected async elementExistsByTestId(testId: string): Promise<boolean> {
+		return await this.#page
+			.getByTestId(testId)
+			.isVisible()
+			.catch(() => false);
+	}
+
+	getBalance(): Locator {
+		return this.#page.getByTestId(AMOUNT_DATA);
+	}
```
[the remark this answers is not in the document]
inc-man: correct, we later check if that locator has a certain text. I changed it in latest commit

e2e/utils/pages/homepage.page.ts
AntonioVentilii: I think I may have explained it wrong:

- is the value `icp-account-id` (not `#icp-account-id`) being used as test ID?
- if yes: which component is using it?
- if yes: it should be in the module `test-ids.constants` like the other test IDs

AntonioVentilii (approved):
LGTM tks

PR #4615, opened 2025-02-06
feat(fix): Removed slot assignment to outer-content

DecentAgeCoder:
# Motivation
Currently, the content of **SwapReview.svelte** appears to be vertically aligned at the bottom. This issue is caused by assigning the Swap Review content to the **outer-content** slot. Since the default slot is very greedy (see stretch statement) stretches its height when the window height increases, the **SwapReview** content is displayed beneath the stretched div.

The Svelte component `SwapWizard.svelte` defines:

- a default slot:

  ```svelte
  <div class={`stretch ${styleClass ?? ''}`}>
  	<slot />
  </div>
  ```

- a slot named "outer-content" underneath:

  ```svelte
  <slot name="outer-content" />
  ```

Furthermore the the margin bottom "mb-12" has been removed since there is no need for an additional bottom margin. 

# Changes

1. The content will be assigned to the default slot
2. The class definition "mb-12" has been removed

# Tests


DenysKarmazynDFINITY:
@DecentAgeCoder the proposed solution does fix the mobile appearance issue but at the same time the removal of "outer-content" slot changes the design on all kind of screens:

* with "outer-content", the content has a reduced horizontal padding (screenshot from prod):
Screenshot 2025-02-06 at 12 34 22

* with "outer-content", the content has an increased horizontal padding (screenshot from your branch) which makes it a bit unaligned with the buttons:
Screenshot 2025-02-06 at 12 34 17

I personally prefer the current version (the one where content borders are aligned with the button section), but if you think this update is really necessary to fix the mobile issue, I suggest you show it to the Design team and get their approval - they would most likely want to update Figma designs accordingly.

Also, whatever decision is made here, we would need to create a follow-up PR to update the conversion feature (`ConvertReview` component) accordingly, as this flow also reuses the same UI with Swap.

daviddecentage (approved):
LGTM

PR #4620, opened 2025-02-06
test(e2e): added testIds for upcoming test

inc-man:
# Motivation

for the upcoming new test we need testIds

# Changes

added all the necessary testIds on the component and in the constants library.

### DIFF
--- a/src/frontend/src/lib/constants/test-ids.constants.ts
+++ b/src/frontend/src/lib/constants/test-ids.constants.ts
@@ -31,6 +31,8 @@ export const RECEIVE_TOKENS_MODAL_SOL_DEVNET_SECTION = 'receive-tokens-modal-sol
 export const RECEIVE_TOKENS_MODAL_SOL_LOCAL_SECTION = 'receive-tokens-modal-sol-local-section';
 
 export const RECEIVE_TOKENS_MODAL_ADDRESS_LABEL = 'receive-tokens-modal-address-label';
+export const RECEIVE_TOKENS_MODAL_TRANSACTION_TIMESTAMP =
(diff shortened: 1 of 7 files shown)


src/frontend/src/lib/constants/test-ids.constants.ts
```diff
@@ -53,6 +55,12 @@ export const NAVIGATION_ITEM_EXPLORER = 'navigation-item-explore';
 export const NAVIGATION_ITEM_SETTINGS = 'navigation-item-settings';
 export const NAVIGATION_ITEM_HOMEPAGE = 'navigation-item-homepage';
 
+export const AMOUNT_INPUT = 'amount-input';
+export const DESTINATION_INPUT = 'destination-input';
+
+export const IN_PROGRESS_MODAL = 'in-progress-modal';
+export const AMOUNT_DATA = 'amount-data';
+export const TOKEN_CARD_ICP = 'token-card-ICP-ICP';
```
AntonioVentilii, line 63: `TOKEN_CARD_ICP` does not seem to be used

inc-man: its not on a component, correct. I use it to click on the icp card after logging in. Since I dislike the token-card+"string" combination I did this. Looks better and feels better.

AntonioVentilii: indeed, but it is not being used in this PR, so theoretically it does not pertain here, right? or am I missing something?

AntonioVentilii: Removed

AntonioVentilii (approved):
LGTM tks

PR #4621, opened 2025-02-06
feat(backend): add selected Theme to the User Settings

AntonioVentilii:
# Motivation

We want to persist the user's selected theme throughout any session. So, we will save it in the backend, in the user's settings.

# Changes

- Add types for `theme` in user profile.
- Create functions to save a selected theme, very similar to the ones that modify the carousel.

# Tests

Included some tests.


AntonioVentilii:
As discussed off-line, we will not persist the theme in the backed, so we will work on a different solution provided by:

- https://github.com/dfinity/oisy-wallet/pull/4635
- https://github.com/dfinity/oisy-wallet/pull/4636

PR #4622, opened 2025-02-06
chore(backend): reorder impl items

AntonioVentilii:
# Motivation

We add the rust formatting rule [`reorder_impl_items`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#reorder_impl_items)


bitdivine (approved):
Note that this is an unstable feature, so will only have any effect if we also switch on +nightly.  I can do that, format and add that to this PR.

PR #4625, opened 2025-02-07
feat(frontend): Apply color mode on theme change or system theme change

daviddecentage:
# Motivation
As user I can set the color mode in settings. This PR serves as a prerequisite for a theme-system refactoring, focusing only on initially setting the theme based on the setting or system setting as fallback.

As defined, system will be the default in case the setting has not been touched by the user.

**Note:** Changing the HTML theme attribute will not change any appearance, as currently no CSS rules exist that would match by htmls theme attribute. This is merely a prerequisite for the upcoming theme change

# Changes

Added mechanism to set theme to root html tag:
- It has been placed in the root layout to always apply when you get to the wallet
- Added a check that reads the system setting in case the theme store has no value set
- Added a listener that sets the theme additionally in case the user changes his system theme setting

# Tests

Tested on chrome 132.0.6834.159, firefox 134.0.2. HTML tag attribute "theme" successfully changed on theme change, or if "system" is selected, the OS default gets applied. OS theme change correctly works on tested machine running Ubuntu.


peterpeterparker (changes requested):
I don't think this implementation should be placed in bulk in the root layout. There are also questions regarding the patterns, consistency, and lack of tests. Additionally, I understand that @AntonioVentilii is currently analyzing whether the theme should be persisted on-chain. Therefore, I'm blocking this to prevent it from being merged over the weekend. Let's wait for Antonio's findings.

peterpeterparker:
Discussed offline. We are moving forward with a different approach for the theming.
I provided two PRs (#4635 and #4636) to align and restart in that direction.
Therefore, this PR can be closed.

daviddecentage:
> Discussed offline. We are moving forward with a different approach for the theming. I provided two PRs (#4635 and #4636) to align and restart in that direction. Therefore, this PR can be closed.

Thanks for providing/preparing the setup. Will close this now

PR #4632, opened 2025-02-07
build(dev-tools): ic-wasm v0.8.5

peterpeterparker:
# Motivation

Not sure it's a must but, it's been a while that II ([here](https://github.com/dfinity/internet-identity/blob/d3d11ec3ada74ed919a5c60f091301b638b7ffae/scripts/bootstrap#L25)) and Juno ([here](https://github.com/junobuild/juno/blob/efa064cc7b43230db59300623871719cc6db6562/docker/bootstrap#L25)) are using ic-wasm v0.8.5.

Given that OISY uses a relatively older version, it's maybe interesting to bump ic-wasm to this version for consistency.

# Changes

- update v0.3.6 -> v0.8.5 in dev-tools
- update guard in docker build


AntonioVentilii (approved):

docker/build
```diff
@@ -71,10 +71,10 @@ if [ ${#CANISTERS[@]} -eq 0 ]; then
...
 
 # Checking for dependencies
-if [[ ! "$(command -v ic-wasm)" || "$(ic-wasm --version)" != "ic-wasm 0.3.6" ]]; then
-  echo "could not find ic-wasm 0.3.6"
-  echo "ic-wasm version 0.3.6 is needed, please run the following command:"
-  echo "  cargo install ic-wasm --version 0.3.6"
+if [[ ! "$(command -v ic-wasm)" || "$(ic-wasm --version)" != "ic-wasm 0.8.5" ]]; then
+  echo "could not find ic-wasm 0.8.5"
+  echo "ic-wasm version 0.8.5 is needed, please run the following command:"
+  echo "  cargo install ic-wasm --version 0.8.5"
```
bitdivine, line 77: If you can assume that `jq` is installed, you could do:
```
IC_WASM_VERSION="$(jq -r '.["ic-wasm"].version' dev-tools.json)"
```
and use the variable to save the repetition.  But this would be a drive-by cleanup and would maybe be better done in another PR.

bitdivine (approved):

docker/build
peterpeterparker: Indeed the goal of the PR was really just to bring up the tooling to the same version we have in other dapps. It can for sure be improved in another PR, not against!

PR #4633, opened 2025-02-08
refactor(frontend): self-sent transactions after its chain one

AntonioVentilii:
# Motivation

It is just a visual change, but it seems more coherent: we display the counter transaction of a self-sent/self-received transaction, after the transaction that originated it.

For example, if I initiate a transaction SENDING to myself, the counter-transaction RECEIVING appears as most recent in the list (even if with the same timestamp).


DenysKarmazynDFINITY (approved):
Makes sense, thanks!

PR #4634, opened 2025-02-08
fix(frontend): Hide social media icons and beta badge on mobile on "secondary" pages

artkorotkikh-dfinity:
# Motivation

On mobile – OISY footer is a bit cluttered, especially on secondary pages when they are empty. We'd like to hide social icons and beta badge on all pages on mobile except the "home screen" (footer is not that visible and located below the tokens) 

# Changes

Hiding block with buttons in <Footer.svelte> for all screens where sidebar is not visible. 

# Tests

Home screen: 
image
image

Token screen: 
image

image

### DIFF
--- a/src/frontend/src/lib/components/core/Footer.svelte
+++ b/src/frontend/src/lib/components/core/Footer.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 	import { IconGitHub } from '@dfinity/gix-components';
+	import { page } from '$app/stores';
 	import IconDfinity from '$lib/components/icons/IconDfinity.svelte';
 	import IconHeart from '$lib/components/icons/IconHeart.svelte';
 	import IconTwitter from '$lib/components/icons/IconTwitter.svelte';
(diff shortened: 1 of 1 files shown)


AntonioVentilii (approved):
Good one! Tks!

src/frontend/src/lib/components/core/Footer.svelte
```diff
@@ -10,6 +11,9 @@
 	import { authNotSignedIn, authSignedIn } from '$lib/derived/auth.derived';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+	import { isRouteTokens } from '$lib/utils/nav.utils';
+
+	$: isHomePage = isRouteTokens($page);
```
peterpeterparker, line 16: @AntonioVentilii can you declare the type here?

```
let isHomePage: boolean
$: isHomePage = isRouteTokens($page);
```

PR #4599, opened 2025-02-05
feat(frontend): Prepare Stores, Enums and Constants for Color Mode Selection 

daviddecentage:
# Motivation
This [comment](https://github.com/dfinity/oisy-wallet/pull/4594#pullrequestreview-2595783559) in PR #4594.

# Changes

- Added theme store to persist selected color mode
- Added derived theme store to get selected color mode
- Added enum for available themes
- Added constants for default color mode and the enums values

# Tests
Working screenshot in PR #4594.

PR #4635, opened 2025-02-09
revert(frontend): #4599 - Prepare Stores, Enums and Constants for Color Mode Selection

peterpeterparker:
# Motivation

We don't strictly use only the design system; we also use Gix components as a UI library, which supports theming. While it is technically possible to duplicate theming stores in OISY and set the root attribute on the <html /> element using the same keyword, this carries a risk of desynchronization—where OISY uses theme X, which happens to be unsupported by Gix components.

It also feels somewhat hacky that two different layers of presentation potentially rely on separate stores for the same information, leading to possible inconsistencies. That’s why I find it somewhat inelegant and, to some extent, a bit hacky to duplicate this information.

I agree that Gix components currently do not support the concept of a System. However, given that decisions related to OISY are generally made with the motto "fast iteration," I would suggest dropping this notion for now. If it is truly necessary, it should be implemented within the UI kit while ensuring that no breaking changes are introduced.

Long story short, I believe it’s easier, faster, and cleaner to revert the logical components introduced last week and simply use Gix components.

# Notes

- If this PR gets merged, PR #4625 should be closed and deleted
- Gix components PR #578 will be necessary to resolve the TODO added in this PR

# Changes

- Revert #4599
- Add comments and TODOs in theme toggler

# Tests

No tests provided given there is no tests in place currently for those features.

### DIFF
--- a/src/frontend/src/lib/components/settings/ThemeSelector.svelte
+++ b/src/frontend/src/lib/components/settings/ThemeSelector.svelte
@@ -1,21 +1,32 @@
 <script lang="ts">
 	import ThemeSelectorCard from '$lib/components/settings/ThemeSelectorCard.svelte';
 	import Img from '$lib/components/ui/Img.svelte';
-	import { THEME_VALUES } from '$lib/constants/app.constants';
-	import { selectedTheme } from '$lib/derived/settings.derived';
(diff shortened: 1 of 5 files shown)


AntonioVentilii (approved):
LGTM tks!

PR #4636, opened 2025-02-09
feat(frontend): init theme as fast as possible

peterpeterparker:
# Motivation

If the team ultimately decides not to load the theme from the backend (PR #4621) due to unavoidable flashes in various scenarios—given that there is no SSR on the IC and the app cannot determine the user's theming preferences at the time of serving—this PR sets up the best approach for theming in an SSG app on the IC. It follows the setup defined in the Gix components and implemented in NNS-dapp and Juno, ensuring the fastest possible theming load by embedding the theme script directly within the index.html files.

# Changes

- Add code snippet in `app.html` as documented in Gix components: https://gix.design/styling/theming

# Tests

No test at this point but an E2E would be nice.


src/frontend/src/app.html
```diff
@@ -55,6 +55,27 @@
...
 
+		<!-- Init theme as fast as possible -->
+		<script theme-loader>
+			try {
+				const isDarkPreferred = window.matchMedia('(prefers-color-scheme: dark)').matches;
+
+				const currentTheme =
+					localStorage.nnsTheme === null ||
+					localStorage.nnsTheme === undefined ||
+					localStorage.nnsTheme === ''
```
AntonioVentilii, line 66: shall we use `localStorage.nnsTheme` or just `localStorage.theme`, or something else?

AntonioVentilii (approved):
LGTM tks!

just a question, but we can always change later

src/frontend/src/app.html
peterpeterparker: Gix components expect `nnsTheme` - it's a constant. You can maybe extend the kit to support a variable for such purpose but, not sure how.

PR #4637, opened 2025-02-09
fix(ci): solve false positive in E2E update snapshots

AntonioVentilii:
# Motivation

Following up comment https://github.com/dfinity/oisy-wallet/pull/3954#discussion_r1946430063 , we try and solve the false positive. If for instance there are changes in macOS (so the TXT is saved as TRUE) but no changes in Ubuntu (so the TXT is saved as FALSE), when the CI read the single TXT it will read FALSE, thus proceeding as if there were no changes. We change this.

# Changes

Assign a different path to each system txt.

# Tests

Run the E2E CI on this PR and worked.


inc-man (approved):
LGTM. tks

PR #4645, opened 2025-02-10
build(frontend): bump gix-cmp next

AntonioVentilii:
# Motivation

Bumping `@dfinity/gix-components` because we would like to use the theme features exposed in:

- https://github.com/dfinity/gix-components/pull/578


peterpeterparker (approved):
LGTM, thx

PR #4646, opened 2025-02-10
feat(frontend): merge Solana wallet signatures with ATA signatures

AntonioVentilii:
# Motivation

In Solana, some signatures may be associated ONLY to the main wallet address, some others to ONLY the ATA address, and ultimately some with both. To include them all in our view, we merge them removing the duplicates one.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4647, opened 2025-02-10
build(frontend): remove unsafe-inline from CSP script

AntonioVentilii:
# Motivation

Following up a [thread in the forum](https://forum.dfinity.org/t/i-accidentally-sent-2-icp-to-my-sign-in-providers-principal-id/40909) about recovering tokens by injecting custom scripts, even though it may come useful and in handy, we would like to remove such a possibility.


scripts/build.csp.mjs
```diff
@@ -169,7 +169,7 @@ const updateCSP = (indexHtml) => {
         img-src 'self' data:;
         frame-src 'self' ${walletConnectFrameSrc} ${onramperConnectFrameSrc};
         manifest-src 'self';
-        script-src 'unsafe-inline' 'strict-dynamic' ${indexHashes.join(' ')};
+        script-src 'strict-dynamic' ${indexHashes.join(' ')};
         base-uri 'self';
         form-action 'none';
         style-src 'self' 'unsafe-inline';
```
AntonioVentilii, line 175: not really sure if we should remove it from here too

peterpeterparker:
Have you deployed to a test environment and can you share the link for review purposes?

AntonioVentilii:
@peterpeterparker just deployed to [fe3.oisy.com](https://fe2.oisy.com/) : https://github.com/dfinity/oisy-wallet/actions/runs/13237871960

peterpeterparker:
Looks good.

This means we won’t support Safari versions older than 15.4. I assume that’s acceptable?

If so, could you update the comment accordingly?

> Safari 15.4 has been released recently - March 15, 2022 - that's why we add 'unsafe-inline' to the rules for backwards compatibility.

On another note, not related to this PR, I noticed a CORS issue with blockchain.info. Not sure if it's expected, but I wanted to share this finding just in case it's unknown.

Capture d’écran 2025-02-10 à 10 59 15

AntonioVentilii:
What about this?

# Safari 15.4 (March 15, 2022) previously required 'unsafe-inline' for compatibility.
# It has now been removed to improve security. Ensure that all scripts are explicitly allowed via nonces or hashes.

AntonioVentilii:
Ultimately, as discussed offline, we decided to not proceed with this PR for now, we will tackle this issue in another time.

PR #4648, opened 2025-02-10
feat(frontend): all non-related Solana transactions are ignored

AntonioVentilii:
# Motivation

With Solana transactions, it may happen that even if a transaction is not related to the wallet address, it may be strictly related to the ATA address. So, we remove the filter done by only-wallet transactions.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4649, opened 2025-02-10
refactor(frontend): move isSourceTokenIcrc2 to context

DenysKarmazynDFINITY:
# Motivation

In order to re-use the `isSourceTokenIcrc2` const, we need to move it up to the Swap context.


AntonioVentilii (approved):
LGTM tks

PR #4656, opened 2025-02-10
build(frontend): add zod-schemas to update scripts

AntonioVentilii:
# Motivation

Since now `zod-schemas` will be release as next too after PR https://github.com/dfinity/ic-js/pull/840, we can add it in the scripts that updates the `ic-js` packages.


peterpeterparker (approved):
LGTm, thx

PR #4658, opened 2025-02-10
style(frontend): show Solana transactions in order of execution

AntonioVentilii:
# Motivation

The Solana instructions are received in the order they were executed, meaning the first instruction in the list was executed first, and the last instruction was executed last.
However, since they all share the same timestamp, we want to display them in reverse order—from the last executed instruction to the first.
This ensures that when shown, the most recently executed instruction appears first, maintaining a more intuitive, backward-looking view of execution history.

### Before

Screenshot 2025-02-10 at 16 19 17

### After

Screenshot 2025-02-10 at 16 18 21


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4660, opened 2025-02-10
fix(frontend): Display the fee types in the Swap modal in the correct format

DecentAgeCoder:
# Motivation
The fees in the Swap UI are currently displayed in the scientific format.

# Changes
- Introduced a new function which mainly formats a given token amount to a String
- Introduced an additional reactive variable for the formatted token amount for each fee type

# Tests
Screenshot from 2025-02-10 16-47-49
Screenshot from 2025-02-10 16-47-13


src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -49,6 +49,17 @@ export const formatTokenBigintToNumber = ({
...
+	value,
+	...restParams
+}: Omit<FormatTokenParams, 'value'> & {
+	value: bigint;
+}): string =>
+	formatToken({
+		value: BigNumber.from(value),
+		...restParams
+	});
+
```
AntonioVentilii, line 62: instead of creating a new util, why not modifying the existing one to accept `bigint` too?

daviddecentage:
Could not reproduce anymore, seems to have been solved by another PR:

image

src/frontend/src/lib/components/swap/SwapFees.svelte
```diff
@@ -27,9 +27,29 @@
 				})
 			: 0;
 
+	let sourceTokenTransferFeeDisplay: string;
+	$: sourceTokenTransferFeeDisplay =
+		nonNullish($sourceToken) && nonNullish($icTokenFeeStore?.[$sourceToken.symbol])
+			? formatTokenAmount({
```
DenysKarmazynDFINITY, line 33: Instead of having 2 calculations per a fee, you can do the following:
1. Leave the `sourceTokenTransferFeeDisplay` calculation as it is now.
2. `sourceTokenTransferFee` can now be something like `Number(sourceTokenTransferFeeDisplay)`. It's the same as you would call `formatTokenBigintToNumber`.

src/frontend/src/lib/components/swap/SwapFees.svelte
```diff
@@ -27,9 +27,29 @@
...
+					value: $icTokenFeeStore[$sourceToken.symbol],
+					displayDecimals: $sourceToken.decimals,
+					unitName: $sourceToken.decimals
+				})
+			: '0';
+
 	let sourceTokenApproveFee: number;
 	$: sourceTokenApproveFee = isSourceTokenIcrc2 ? sourceTokenTransferFee : 0;
 
+	let sourceTokenApproveFeeDisplay: string;
```
DenysKarmazynDFINITY, line 43: No need to re-calculate this one - it's always the same value as `sourceTokenTransferFeeDisplay`. Therefore, you can just check if `isSourceTokenIcrc2` is true and use `sourceTokenTransferFeeDisplay` here.

src/frontend/src/lib/components/swap/SwapFees.svelte
daviddecentage: Good point, I changed it accordingly

src/frontend/src/lib/components/swap/SwapFees.svelte
DenysKarmazynDFINITY: What I meant is that it could be simplified even further:
```
$: sourceTokenApproveFeeDisplay = isSourceTokenIcrc2 ? sourceTokenTransferFeeDisplay : '0';
```

Just to give you a context why it's like this: if `isSourceTokenIcrc2` is true, we just need to double the original source token fee ($sourceToken.fee, or a more precise value - $icTokenFeeStore?.[$sourceToken.symbol]).

daviddecentage: I see, thanks for the explanation. Wasnt too sure on the calculation there. Fixed accordingly

PR #4661, opened 2025-02-10
fix(frontend): event bubbling in TokenInput

DenysKarmazynDFINITY:
# Motivation

While working on the conversion flows, I noticed that `nnsInput` event bubbling was not properly working after the latest changes in TokenInput components.


AntonioVentilii (approved):
LGTM tks!

PR #4665, opened 2025-02-10
chore(frontend): bump tailwindcss to v4

AntonioVentilii:
# Motivation

We want to bump `tailwindcss` to v4. However [the migration tool does not support SASS files](https://github.com/tailwindlabs/tailwindcss/issues/15716).

# Workaround

We changed all the `.scss` files to a copy in `.css` and use the migration tool, and then reverted. Same thing with the language in the `<style>` tag of Svelte components.
Then we added a few manual tweaks anyway.

# Disclaimer

Since a part of the PR was a manual process, it may very well be that I missed something or there is some error in migration.

# Note

Apparently `prettier-plugin-tailwindcss` was bumped too, so a few changes happened in the sorting order of the tailwind classes too. That is the reason for this many files changed in this PR, since we have an auto-formatter.

Furthermore, it seems that [a few classes will require a Vite update to work properly](https://github.com/vitejs/vite/issues/18974). For example `space-x-*` and `space-y-*`.

# Changes

- The migration tool:
  - installed `tailwindcss` v4;
  - changed `postcss.config`;
  - migrated the tailwind classes that required migration (see [upgrade guide](https://tailwindcss.com/docs/upgrade-guide));
  - correctly changed `global.css` (that we reverted to `global.scss` after).

- A few changes were missing, and we did it manually:
  - The `Hero` gradient does not need a `via-` color anymore, it can changed to `from-`.
  - The util `@screen foo` is deprecated and should be replaced with `@media (width >= theme(--breakpoint-foo))` (see [codebase of tailwindcss migration](https://github.com/tailwindlabs/tailwindcss/blob/121cf6bd807a35ad9814dd5444eeb3594601fff8/packages/%40tailwindcss-upgrade/src/codemods/migrate-media-screen.test.ts)).
  - It is [suggested](https://tailwindcss.com/docs/upgrade-guide#using-vite) to install and add `@tailwindcss/vite` in `vite` (and `vitest`) configs;
  - Move `@include meta.load-css(...)` inside `@layer base {...}` (suggestion in [tailwindcss discussion](https://github.com/tailwindlabs/tailwindcss/discussions/16421)).

- We create another branch to make the changes manually following the [upgrade guide](https://tailwindcss.com/docs/upgrade-guide) and they matched:
  - https://github.com/dfinity/oisy-wallet/pull/4666

# Tests

There is not much test to help us, apart update the E2E snapshots and manual navigation in local replica:

https://github.com/user-attachments/assets/d0ae2e36-7fff-4ff8-8829-3c93e900650a


peterpeterparker (approved):
LGTM, thx

PR #4666, opened 2025-02-10
chore(frontend): bump tailwindcss to v4 - manual branch

AntonioVentilii:
# Motivation

The migration tool of `tailwindcss` does not wirk with SASS files. So we did PR https://github.com/dfinity/oisy-wallet/pull/4665 that is using a workaround. However, to double-check, we will use this branch to do a manual migration based on the `documentation for the upgrade guide` and then rebase with the first PR to check if they match.


AntonioVentilii:
THe other rebasing branch is merged, so we can close this one

PR #4669, opened 2025-02-11
feat(frontend): use gix-cmp theme store in ThemeSelector

AntonioVentilii:
# Motivation

We want to use the `themeStore` store from `gix-components` to switch between themes.

Since it does not support System theme, we will make it possible by deleting the cached theme, triggering an automatic usage of the system theme.

# Changes

- Create constants and enum.
- Use `themeStore` of `gix-components` in `ThemeSelector`.
- Create specific logic for System theme: delete local storage.

# Tests

For now we provide only unittests.


src/frontend/src/lib/components/settings/ThemeSelector.svelte
```diff
@@ -1,35 +1,33 @@
...
-	// TODO: replace with @dfinity/gix-components Theme.Light and Theme.Dark
-	enum Themes {
-		LIGHT = 'light',
-		DARK = 'dark',
-		SYSTEM = 'system'
-	}
+	const selectTheme = (theme: Theme | SystemTheme) => {
+		// Since gix-components does not support System theme, a good solution is to delete the cached theme when the user selects the System theme.
+		if (theme === SystemTheme.SYSTEM) {
+			// TODO: use variable exposed from gix-components when it will be exposed.
```
peterpeterparker, line 13: Shouldn't we rather "// TODO: add a reset feature to the themeStore in gix-comp" ?

src/frontend/src/lib/components/settings/ThemeSelector.svelte
```diff
@@ -1,35 +1,33 @@
...
+			localStorage.setItem('nnsTheme', JSON.stringify(null));
+			return;
```
peterpeterparker, line 15: The implementation is partial. Either we remove the value in the local storage and we update the `themeStore` according the system theme or we do no nothing. I guess related to my previous comment.

src/frontend/src/lib/enums/themes.ts
```diff
@@ -0,0 +1,3 @@
+export enum SystemTheme {
```
peterpeterparker, line 1: Does `SystemThem` really needs to be global? Isn't this just a value that is stricly only use in the `ThemeSelector`?

src/frontend/src/lib/components/settings/ThemeSelector.svelte
```diff
@@ -1,35 +1,33 @@
...
-	// TODO: implement selected using $themeStore from @dfinity/gix-components
-	// selected={$selectedTheme === theme}
 </script>
 
 <div class="flex flex-row">
 	{#each THEME_VALUES as theme}
 		<ThemeSelectorCard
 			label={$i18n.settings.text[`appearance_${theme}`]}
-			selected={false}
+			selected={$themeStore === theme}
```
peterpeterparker: Is this correct? Isn't the logic the following

```
if local storage empty then it's system
else themeStore === theme
```

src/frontend/src/lib/components/settings/ThemeSelector.svelte
daviddecentage: @peterpeterparker Youre right. Theres currently no way for System to be a selected mode. And updating the store is only possible currently by using any values of the gix "Theme" enum, which only has light or dark.

peterpeterparker:
```
const theme = window.getComputedStyle(prefered-scheme)
themeStore.selectTheme(theme)
```

src/frontend/src/lib/components/settings/ThemeSelector.svelte
daviddecentage: This would work:
`
selected={theme === SystemTheme.SYSTEM ? localStorage.getItem('nnsTheme') === "null" : $themeStore === theme}`

src/frontend/src/lib/components/settings/ThemeSelector.svelte
AntonioVentilii: I am actually working on that eheh!

src/frontend/src/lib/components/settings/ThemeSelector.svelte
AntonioVentilii: Indeed! Working on the `reset` function there

src/frontend/src/lib/enums/themes.ts
AntonioVentilii: I put it there just because it is useful to use it in the tests

src/frontend/src/lib/components/settings/ThemeSelector.svelte
AntonioVentilii: I changed to another further iteration and added tests specific for the `selected` flag. Have a look please

src/frontend/src/lib/components/settings/ThemeSelector.svelte
AntonioVentilii: Should be this PR:

- https://github.com/dfinity/gix-components/pull/582

src/frontend/src/lib/components/settings/ThemeSelector.svelte
AntonioVentilii: In the end, I opted for refresh, let me know what you think:

- https://github.com/dfinity/gix-components/pull/582

src/frontend/src/lib/components/settings/ThemeSelector.svelte
peterpeterparker: IMO it does not make sense to iterate here until the enhancement in Gix cmp is implemented.

src/frontend/src/lib/enums/themes.ts
AntonioVentilii: Sure, we can leave it "internal"; in the tests we will have to hard-code the value, right?

src/frontend/src/lib/components/settings/ThemeSelector.svelte
AntonioVentilii: The correct PR that was merged with the reset function is:

- https://github.com/dfinity/gix-components/pull/585

src/frontend/src/lib/enums/themes.ts
AntonioVentilii: So, I did an iteration putting `SystemTheme` internally, but we are using it in `THEME_VALUES` constants too:

`export const THEME_VALUES = [...Object.values(Theme), ...Object.values(SystemTheme)];`

So, it becomes quite a mess, because the loop inside the component does not recognize the string `'system'` as part of the enum type.

And `THEME_VALUES` is used quite a few times, it is quite handy for the tests too.

Any idea? I tried to do even

`export const THEME_VALUES = [...Object.values(Theme), 'system' as const];`

but nothing...

AntonioVentilii: well, it really is for the tests ultimately. Meaning that I would duplicate the declarations, once in the component, and once in the test module, all good?

peterpeterparker: I could argue that the tests should not use the constant as modifying it would also means automatically modifying the test and per extension not necessarely prevent unwanted changes.

But ultimately, really no strong opinion, you can also keep it global.

peterpeterparker (approved):
LGTM; thx

src/frontend/src/lib/enums/themes.ts
peterpeterparker: Did you maybe forgot to delete `SystemTheme` and `THEME_VALUES` in the related constants if ultimately you integrated those within the component @AntonioVentilii ?

AntonioVentilii: Ah! Sorry! Forgot

Done with:

- https://github.com/dfinity/oisy-wallet/pull/4709

PR #4670, opened 2025-02-11
feat(backend): use wrapping_add to increment version

peterpeterparker:
# Motivation

We would like to rotate the version increment in the unlikely events - an absolute edge case as it would consumes zillion of cycles to get there - of the number - used to prevent overwrite - reached `u64::MAX`.

# Changes

- Replace plus one with `wrapping_add` as discussed offline.

# Tests

No particular tests given the above motivation.


bitdivine (approved):
LGTM, thanks!

PR #4671, opened 2025-02-11
feat(frontend): ledger transactionFee

DenysKarmazynDFINITY:
# Motivation

To always use the latest icrc token fee, we are going to fetch this info from the ledgers in the Swap flow.

### DIFF
--- a/src/frontend/src/icp/api/icrc-ledger.api.ts
+++ b/src/frontend/src/icp/api/icrc-ledger.api.ts
@@ -43,6 +43,31 @@ export const metadata = async ({
 };
 
 /**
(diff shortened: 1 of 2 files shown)


AntonioVentilii (approved):
LGTM tks!

My understanding is that we will not use the `fee` prop of `IcToken` anymore, right? If so, when the implementation is finished, shall we remove that (schema `IcFeeSchema`)? Or is it used for other scopes?

DenysKarmazynDFINITY:
@AntonioVentilii the idea is to use the `fee` prop as a fallback if `ledger.transferFee` fails. You will see it in the next PRs.

PR #4672, opened 2025-02-11
feat(frontend): adjust swap fees logic

DenysKarmazynDFINITY:
# Motivation

The idea of this PR is to improve the way of how we calculate the fees needed to complete a kong swap action.

# Changes

1. Start using `transferFee` API method to fetch the latest icrc token fee. As a fallback, the hardcoded fee value will be used.
2. Display an approval fee if the selected token supports icrc2 approve.
3. Adjust swap service to assume the approval fee.
4. Update Swap components with the `sourceTokenFee`.

# Tests

The existing tests have been adjusted.


src/frontend/src/lib/components/swap/SwapWizard.svelte
```diff
@@ -37,6 +39,31 @@
...
+		if (isNullish($sourceToken)) {
+			return;
+		}
+
+		try {
+			sourceTokenFee = await transactionFee({
+				identity: $authIdentity,
+				ledgerCanisterId: ($sourceToken as IcToken).ledgerCanisterId
+			});
+		} catch (e) {
```
AntonioVentilii, line 60:
```suggestion
		} catch (e: unknown) {
```

AntonioVentilii (commented):
Before I review it, I am wondering: have you thought of using the context for this too?

For example, you could create `sourceTokenFee` in the swap context, initialize it with the prop `token.fee` and then update it correctly with the function you created.
Or even better, create a fee context component with the logic to update the fee.

That would avoid the need for bubbling the value.

src/frontend/src/icp/stores/ic-token-fee.store.ts
```diff
@@ -0,0 +1,34 @@
...
+import { writable, type Readable } from 'svelte/store';
+
+export type IcTokenFeeStoreData = Option<Record<string, bigint>>;
+
+export interface IcTokenFeeStore extends Readable<IcTokenFeeStoreData> {
+	setIcTokenFee: (data: IcTokenFeeStoreData) => void;
+	reset: () => void;
+}
+
+export const initIcTokenFeeStore = (): IcTokenFeeStore => {
```
AntonioVentilii, line 11: do you need to `export` this one? and same questions for the types above

src/frontend/src/icp/stores/ic-token-fee.store.ts
```diff
@@ -0,0 +1,34 @@
...
+	const { subscribe, set } = writable<IcTokenFeeStoreData>(undefined);
+
+	return {
+		subscribe,
+
+		reset() {
+			set(null);
+		},
```
AntonioVentilii, line 19: for consistency, maybe make it in arrow function?

`reset: () => {set(null)}`

src/frontend/src/icp/stores/ic-token-fee.store.ts
```diff
@@ -0,0 +1,34 @@
...
+
+		setIcTokenFee(data: IcTokenFeeStoreData) {
+			set(data);
+		}
```
AntonioVentilii, line 23: to avoid using the logic

```typescript
store.setIcTokenFee({
  ...$store,
  [token.symbol]: token.fee
});
```

you could use `update`:

```typescript
// pseudo-code

const { subscribe, update } = writable<IcTokenFeeStoreData>(undefined);

return {
...
  setIcTokenFee: ({ tokeSymbol, data }: { tokenSymbol: string; data: bigint }) =>
    update((state) => ({
    ...(nonNullish(state) && state),
    [tokeSymbol]: data
  })),
  ...
}
```

src/frontend/src/icp/components/fee/IcTokenFeeContext.svelte
```diff
@@ -0,0 +1,45 @@
...
+	import type { IcToken } from '$icp/types/ic-token';
+	import { authIdentity } from '$lib/derived/auth.derived';
+	import { nullishSignOut } from '$lib/services/auth.services';
+
+	export let token: IcToken | undefined;
+
+	const { store } = getContext<IcTokenFeeContext>(IC_TOKEN_FEE_CONTEXT_KEY);
+
+	const loadIcTokenFee = async () => {
+		if (isNullish($authIdentity)) {
```
AntonioVentilii, line 15: if i remember correctly the previous iteration you would start defining it as undefined

Is it still the case needed, like `store.reset()`, or something similar?

Or is this context not worried about it?

src/frontend/src/icp/stores/ic-token-fee.store.ts
DenysKarmazynDFINITY: This one can be without export indeed, the types will be used in tests/components.

src/frontend/src/icp/stores/ic-token-fee.store.ts
DenysKarmazynDFINITY: done

src/frontend/src/icp/stores/ic-token-fee.store.ts
DenysKarmazynDFINITY: good idea, updating

AntonioVentilii (approved):
LGTM tks

PR #4674, opened 2025-02-11
feat(frontend): expose method register_airdrop_recipient of rewards canister

AntonioVentilii:
# Motivation

We expose method `register_airdrop_recipient` of rewards canister to use it and send snapshots of possible eligible users.


src/frontend/src/lib/canisters/reward.canister.ts
```diff
@@ -47,4 +48,10 @@ export class RewardCanister extends Canister<RewardService> {
 
 		return claim_vip_reward(vipReward);
 	};
+
+	registerAirdropRecipient = (userSnapshot: UserSnapshot): Promise<undefined> => {
```
DenysKarmazynDFINITY, line 52: Maybe let's not return anything - `Promise<void>`? Or is it planned to update this type in a follow up PR?

DenysKarmazynDFINITY (approved):

PR #4675, opened 2025-02-12
feat(frontend): service to register user snapshot

AntonioVentilii:
# Motivation

We are going to send user account snapshots to the Rewards canister.

To do so, we first create the service to collect the data, parse them, and then register.

# Note

The data are collected from the stores (balance, transactions, etc.) in an imperative way, since the service since the service it is not reactive. It "simply" needs to take a snapshot of the current user's balances and transactions and send it to the reward canister.

# Changes

- Create mapper for IC and Solana transactions.
- Create mapper for IC and Solana token accounts.
- Collect data and map them all but only for ICRC and SPL tokens (for now).
- Use existing service to send the data to the rewards canister.

# Tests

Added tests


src/frontend/src/lib/services/user-snapshot.services.ts
```diff
@@ -0,0 +1,223 @@
...
+		last_transactions: lastTransactions.map((transaction) =>
+			toSplTransaction({ transaction, address })
+		)
+	};
+
+	return isNetworkIdSOLDevnet(networkId) ? { SplDevnet: snapshot } : { SplMainnet: snapshot };
+};
+
+const takeAccountSnapshots = (timestamp: number): AccountSnapshotFor[] => {
+	``;
```
DenysKarmazynDFINITY, line 176: some leftover to remove 🙂

src/frontend/src/lib/services/user-snapshot.services.ts
```diff
@@ -0,0 +1,223 @@
...
+	value,
+	timestamp
+}: Pick<IcTransactionUi | SolTransactionUi, 'type' | 'value' | 'timestamp'>): Omit<
+	Transaction_Icrc | Transaction_Spl,
+	'counterparty'
+> => ({
+	transaction_type: toTransactionType(type),
+	timestamp: timestamp ?? 0n,
+	amount: value ?? 0n,
+	network: {}
```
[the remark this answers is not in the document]
AntonioVentilii: yes indeed, I discussed it with @bitdivine and for now we should send it empty

src/frontend/src/lib/services/user-snapshot.services.ts
AntonioVentilii: wow! removed

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4677, opened 2025-02-12
refactor(frontend): extract mock for authIdentity store

AntonioVentilii:
# Motivation

Since it is used a few times, we move the mock for `authIdentity` store to a mock file.


DenysKarmazynDFINITY (approved):
Nice, thanks

PR #4679, opened 2025-02-12
feat(frontend): Swap fee logic update

DenysKarmazynDFINITY:
# Motivation

In this PR, we start using the IcTokenFee store in the Swap flow, as well as update SwapFees component UI with the skeleton loading indicator and additional entry for the approval fee.

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapFees.svelte
+++ b/src/frontend/src/lib/components/swap/SwapFees.svelte
@@ -60,7 +60,7 @@
 			: 0;
 
 	let sourceTokenApproveFee: number;
-	$: sourceTokenApproveFee = sourceTokenTransferFee;
+	$: sourceTokenApproveFee = isSourceTokenIcrc2 ? sourceTokenTransferFee : 0;
 
 	let destinationTokenTotalFeeUSD: number;
 	$: destinationTokenTotalFeeUSD = nonNullish($destinationTokenExchangeRate)
--- a/src/frontend/src/lib/components/swap/SwapFees.svelte
+++ b/src/frontend/src/lib/components/swap/SwapFees.svelte
@@ -1,8 +1,12 @@
 <script lang="ts">
-	import { nonNullish } from '@dfinity/utils';
+	import { isNullish, nonNullish } from '@dfinity/utils';
 	import { getContext } from 'svelte';
+	import IcTokenFeeContext from '$icp/components/fee/IcTokenFeeContext.svelte';
+	import { IC_TOKEN_FEE_CONTEXT_KEY } from '$icp/stores/ic-token-fee.store';
+	import SwapFee from '$lib/components/swap/SwapFee.svelte';
 	import ModalExpandableValues from '$lib/components/ui/ModalExpandableValues.svelte';
 	import ModalValue from '$lib/components/ui/ModalValue.svelte';
+	import SkeletonText from '$lib/components/ui/SkeletonText.svelte';
 	import { SWAP_TOTAL_FEE_THRESHOLD } from '$lib/constants/swap.constants';
 	import { i18n } from '$lib/stores/i18n.store';
 	import {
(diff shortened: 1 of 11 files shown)


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -68,7 +70,7 @@ export const swap = async ({
 		(await approve({
 			identity,
 			ledgerCanisterId,
-			amount: parsedSwapAmount.toBigInt() + (sourceToken.fee ?? 0n),
+			amount: parsedSwapAmount.toBigInt() + (sourceTokenFee ?? 0n) * (isSourceTokenIcrc2 ? 2n : 1n),
```
AntonioVentilii, line 73: maybe some documentation comment about this two values `2n` and `1n`? Why are they hard-coded?

src/frontend/src/lib/components/swap/SwapFees.svelte
```diff
@@ -35,34 +49,41 @@
...
 
 	let sourceTokenTotalFeeUSD: number;
 	$: sourceTokenTotalFeeUSD = nonNullish($sourceTokenExchangeRate)
-		? sourceTokenFee * $sourceTokenExchangeRate
+		? (sourceTokenTransferFee + sourceTokenApproveFee) * $sourceTokenExchangeRate
 		: 0;
 </script>
 
-{#if nonNullish($destinationToken) && nonNullish($sourceToken)}
+{#if nonNullish($destinationToken) && nonNullish($sourceToken) && nonNullish(swapAmount) && $swapAmountsStore?.swapAmounts !== null}
```
AntonioVentilii, line 76: i imagine this cannot be `nonNullish($swapAmountsStore?.swapAmounts)`? what is the case when it is `undefined` and it is ok to proceed?

src/frontend/src/lib/components/swap/SwapFees.svelte
```diff
@@ -35,34 +49,41 @@
...
+
 	let destinationTokenTotalFeeUSD: number;
 	$: destinationTokenTotalFeeUSD = nonNullish($destinationTokenExchangeRate)
 		? (liquidityProvidersFee + gasFee) * $destinationTokenExchangeRate
 		: 0;
 
 	let sourceTokenTotalFeeUSD: number;
 	$: sourceTokenTotalFeeUSD = nonNullish($sourceTokenExchangeRate)
-		? sourceTokenFee * $sourceTokenExchangeRate
+		? (sourceTokenTransferFee + sourceTokenApproveFee) * $sourceTokenExchangeRate
```
AntonioVentilii, line 72: for my understanding, why now we have 2 fees?

src/frontend/src/lib/components/swap/SwapFees.svelte
```diff
@@ -1,22 +1,36 @@
...
 		SWAP_AMOUNTS_CONTEXT_KEY,
 		type SwapAmountsContext
 	} from '$lib/stores/swap-amounts.store';
 	import { SWAP_CONTEXT_KEY, type SwapContext } from '$lib/stores/swap.store';
+	import type { OptionAmount } from '$lib/types/send';
 	import { formatTokenBigintToNumber, formatUSD } from '$lib/utils/format.utils';
 
-	const { destinationToken, destinationTokenExchangeRate, sourceToken, sourceTokenExchangeRate } =
-		getContext<SwapContext>(SWAP_CONTEXT_KEY);
+	export let swapAmount: OptionAmount;
```
AntonioVentilii, line 20: is this value actively used? I cannot find any reference apart from a nullish check

src/frontend/src/lib/services/swap.services.ts
DenysKarmazynDFINITY: Good point, added a clarification comment

src/frontend/src/lib/components/swap/SwapFees.svelte
DenysKarmazynDFINITY: It will be `undefined` if it has never been loaded yet, we will be showing a loader (skeleton) in this case. If it's a `null`, it means Kong cannot offer a deal for the provided tokens (https://github.com/dfinity/oisy-wallet/blob/main/src/frontend/src/lib/components/swap/SwapAmountsContext.svelte#L70).

src/frontend/src/lib/components/swap/SwapFees.svelte
DenysKarmazynDFINITY: Transferring a token with the icrc2 "approve" costs a double fee - "approval" fee and "transfer" fee. Previously, we only calculated the "transfer" part and the "approval" one was missing.

src/frontend/src/lib/components/swap/SwapFees.svelte
DenysKarmazynDFINITY: The fees will only be loaded if user starts typing some value in the source token input (hence if `swapAmount` has some value). Therefore, that check prevents the "fee" entry to be shown when the fees are unknown and not being loaded.

AntonioVentilii (approved):
LGTM tks

PR #4680, opened 2025-02-12
feat(frontend): updates rewards service

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/services/reward-code.services.ts
+++ b/src/frontend/src/lib/services/reward-code.services.ts
@@ -12,6 +12,7 @@ import type { ResultSuccess } from '$lib/types/utils';
 import type { Identity } from '@dfinity/agent';
 import { fromNullable } from '@dfinity/utils';
 import { get } from 'svelte/store';
+import type {AirdropsResponse} from "$lib/types/airdrop";
 
 const queryVipUser = async (params: {
 	identity: Identity;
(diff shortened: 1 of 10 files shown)


src/frontend/src/lib/types/airdrop.ts
```diff
@@ -0,0 +1,6 @@
+import type { RewardInfo } from '$declarations/rewards/rewards.did';
+
+export interface AirdropsResponse {
+	airdrops: RewardInfo[];
+	last_timestamp: bigint;
```
AntonioVentilii, line 5:
```suggestion
	lastTimestamp: bigint;
```

src/frontend/src/lib/services/reward-code.services.ts
```diff
@@ -56,6 +57,47 @@ export const isVipUser = async (params: { identity: Identity }): Promise<ResultS
...
+ * @async
+ * @param {Object} params - The parameters required to load the user data.
+ * @param {Identity} params.identity - The user's identity for authentication.
+ * @returns {Promise<AirdropsResponse>} - Resolves with the received airdrops and the last timestamp of the user.
+ *
+ * @throws {Error} Displays an error toast and returns an empty list of airdrops if the query fails.
+ */
+export const getAirdrops = async (params: { identity: Identity }): Promise<AirdropsResponse> => {
+	try {
+		return await queryAirdrops({ ...params, certified: false });
```
AntonioVentilii, line 89: why not putting it already here? removing `queryAirdrops` I mean. Do you need it for something else later? But even so, it seems that you should use `getAirdrops` instead

src/frontend/src/lib/services/reward-code.services.ts
```diff
@@ -56,6 +57,47 @@ export const isVipUser = async (params: { identity: Identity }): Promise<ResultS
 	}
 };
 
+const queryAirdrops = async (params: {
+	identity: Identity;
+	certified: boolean;
+}): Promise<AirdropsResponse> => {
+	const userData = await getUserInfoApi({
```
AntonioVentilii, line 64:
```suggestion
	const { usage_awards, last_snapshot_timestamp} = await getUserInfoApi({
```

src/frontend/src/lib/services/reward-code.services.ts
```diff
@@ -56,6 +57,47 @@ export const isVipUser = async (params: { identity: Identity }): Promise<ResultS
...
+		...params,
+		nullishIdentityErrorMessage: get(i18n).auth.error.no_internet_identity
+	});
+
+	return {
+		airdrops: fromNullable(userData.usage_awards) ?? [],
```
AntonioVentilii, line 70: i think you don't need `fromNullable` for this one. But for its element yes, because you will receive a list of

```typescript
export interface RewardInfo {
	name: [] | [string];
	ledger: Principal;
	timestamp: bigint;
	amount: bigint;
}
```

and so you need to use `fromNullable` for `name` prop

AntonioVentilii: my mistake, I read it wrong. You MUST indeed use `fromNullable`. However still check the inside props.

AntonioVentilii (approved):

src/frontend/src/lib/services/reward-code.services.ts
BonomoAlessandro: I implemented it like this for consistency. All the other function are structured the same way.

AntonioVentilii (approved):

PR #4685, opened 2025-02-12
style(frontend): remove new class that was not there before

AntonioVentilii:
# Motivation

As noticed by @daviddecentage , we were using a tailwind class `px-4.5` that somehow [never existed in the previous version of `tailwindcss` v3](https://v3.tailwindcss.com/docs/padding#add-horizontal-padding).

However, with [`tailwindcss` v4](https://tailwindcss.com/docs/padding), any class `px-<number>` will be treated as `padding-inline: calc(var(--spacing) * <number>);`. That means that we have unexpected paddings.

So, we are going to remove them all.


DenysKarmazynDFINITY (approved):
The code LGTM, but please take a look at the failing test

PR #4687, opened 2025-02-12
chore(frontend): change icons to use currentColor

AntonioVentilii:
# Motivation

It makes more sense that these icons inherit the color instead of the hardcoded one.


DenysKarmazynDFINITY (approved):
The changes LGTM! Shall we run e2e before merging to make sure no unexpected color changes are introduced?

PR #4688, opened 2025-02-12
feat(frontend): EthFeeDisplay component

DenysKarmazynDFINITY:
# Motivation

In this PR, a new component created for displaying ETH gas fee. It has the same logic as the old `DisplayFee` has but a different underlying UI component (see screenshot). The old component is considered outdated now and will be removed after both send and convert flows have been migrated.

The PR also includes a small related refactoring - `ConvertFee` component is renamed to `FeeDisplay` and moved to the `fee` folder.

Screenshot 2025-02-12 at 17 37 00


AntonioVentilii (approved):
LGTM tks

PR #4692, opened 2025-02-13
test(e2e): enable and disable SPL tokens

AntonioVentilii:
# Motivation

We want to include SPL in the tests to toggle tokens.

# Changes

- Created common function of enable and disable tokens.
- Use the function in the existing tests.
- Created new tests for tokens in Solana mainnet and Solana devnet.

# Tests

Running snapshots on this PR.


inc-man (commented):
This PR needs #4719 to be merged first right? So the new "loop" method is in place. Then adding new tokens should be even easier.

PR #4697, opened 2025-02-13
feat(frontend): implements airdrop guard

BonomoAlessandro:
# Motivation

It should be possible to notify users on login if they have received an airdrop.

# Changes

- adds airdrops guard


src/frontend/src/lib/components/guard/AirdropGuard.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
+	import { isNullish } from '@dfinity/utils';
+	import { onMount } from 'svelte';
+	import { authIdentity } from '$lib/derived/auth.derived';
+	import { nullishSignOut } from '$lib/services/auth.services';
+	import { getAirdrops } from '$lib/services/reward-code.services';
+
+	onMount(async () => {
+		if (isNullish($authIdentity)) {
+			await nullishSignOut();
```
AntonioVentilii, line 10: not really sure you need the `nullishSignOut` since there is a guard before already checking this, you can just `return`

src/frontend/src/lib/components/guard/AirdropGuard.svelte
```diff
@@ -0,0 +1,28 @@
...
+			return;
+		}
+
+		const initialLoading = sessionStorage.getItem('initialLoading');
```
AntonioVentilii, line 14: let's define this key as constant inside this component, so you don't need to hardcode it twice. And a more airdrop-specific text?

src/frontend/src/lib/components/guard/AirdropGuard.svelte
```diff
@@ -0,0 +1,28 @@
...
+		if (isNullish(initialLoading)) {
+			const { airdrops, lastTimestamp } = await getAirdrops({ identity: $authIdentity });
+			const newAirdrops = airdrops.filter((airdrop) => airdrop.timestamp >= lastTimestamp);
+
+			if (newAirdrops.length > 0) {
+				newAirdrops.some((airdrop) => airdrop.name === 'jackpot');
+				// TODO open airdrop modal
+			}
+			sessionStorage.setItem('initialLoading', 'true');
+		}
```
AntonioVentilii, line 24: ideally, the entire function can become a service or utils, so you can create unittests and you don't need to do it directly for the component

BonomoAlessandro: @AntonioVentilii I just moved it into a utils class and provided some test cases

src/frontend/src/lib/utils/airdrops.utils.ts
```diff
@@ -0,0 +1,25 @@
...
+	if (isNullish(initialLoading)) {
+		const { airdrops, lastTimestamp } = await getAirdrops({ identity });
+		const newAirdrops: AirdropInfo[] = airdrops.filter(
+			(airdrop) => airdrop.timestamp >= lastTimestamp
+		);
+
+		sessionStorage.setItem(INITIAL_AIRDROP_RESULT, 'true');
+
+		if (newAirdrops.length > 0) {
+			const containsJackpot: boolean = newAirdrops.some((airdrop) => airdrop.name === 'jackpot');
```
AntonioVentilii, line 19:
```suggestion
			const containsJackpot: boolean = newAirdrops.some(({name}) => name === 'jackpot');
```

src/frontend/src/lib/utils/airdrops.utils.ts
```diff
@@ -0,0 +1,25 @@
...
+import { isNullish } from '@dfinity/utils';
+
+export const INITIAL_AIRDROP_RESULT = 'initialAirdropResult';
+
+export const loadAirdropResult = async (identity: Identity): Promise<AirdropResult> => {
+	const initialLoading: string | null = sessionStorage.getItem(INITIAL_AIRDROP_RESULT);
+	if (isNullish(initialLoading)) {
+		const { airdrops, lastTimestamp } = await getAirdrops({ identity });
+		const newAirdrops: AirdropInfo[] = airdrops.filter(
+			(airdrop) => airdrop.timestamp >= lastTimestamp
```
AntonioVentilii, line 13:
```suggestion
			({timestamp}) => timestamp >= lastTimestamp
```

AntonioVentilii (approved):
LGTM tks! Two small changes, but up to you, non-blockers

PR #4701, opened 2025-02-13
feat(frontend): Theme refactoring preparations

daviddecentage:
# Motivation

In order to reduce the scope of the theme refactoring which is needed for the new dark mode, I split up the big PR in preperation to the full refactoring.

This step shouldnt change anything appearance wise, and just adds classes and prepares some heavily refactored components for the change.

# Changes

- Added styleClass prop to several components to accept classes from a parent
- Added nav-item, nav-item-condensed classes to navigational components to unify the nav components as they are in figma (these classes dont have any styling applied yet)
- Added dropdown-button and dropdown-item classes to dropdown buttons analogous to the nav items
- Added id to main oisy logo so it can be targeted by a selector (needed for inverting it in dark mode later)
- Added network-logo-bw class to network logo items if black and white (needed for inverting it in dark mode later)
- Added new button class types to style.ts so it matches the ones in figma

# Tests

None, there should be no visual or functional changes with this PR

### DIFF
--- a/src/frontend/src/lib/components/icons/OisyWalletLogo.svelte
+++ b/src/frontend/src/lib/components/icons/OisyWalletLogo.svelte
@@ -15,7 +15,7 @@
 	<IconAstronautHelmet />
 </div>
 
-<picture aria-label={ariaLabel} class="w-24">
+<picture aria-label={ariaLabel} id="oisy-wallet-logo" class="w-24">
 	<source srcset={oisyLogoSmall} media="(max-width: 639px)" />
 	<Img src={oisyLogoLarge} alt={ariaLabel} />
 </picture>
--- a/src/frontend/src/icp/components/tokens/IcTokenMenu.svelte
+++ b/src/frontend/src/icp/components/tokens/IcTokenMenu.svelte
@@ -19,6 +19,7 @@
 	{#if nonNullish(transactionsExplorerUrl)}
 		<div in:fade>
 			<ExternalLink
+				styleClass="nav-item"
 				fullWidth
 				href={transactionsExplorerUrl}
 				ariaLabel={$i18n.tokens.alt.open_dashboard}
--- a/src/frontend/src/lib/components/about/AboutItem.svelte
+++ b/src/frontend/src/lib/components/about/AboutItem.svelte
@@ -1,10 +1,13 @@
 <script lang="ts">
 	export let asMenuItem = false;
 	export let testId: string | undefined = undefined;
+	export let styleClass = '';
 </script>
 
 <button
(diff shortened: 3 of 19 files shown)


AntonioVentilii (changes requested):
I am not really should of the scope of this PR, there are a few tailwind classes that are being introduced but do not exists. And there are mixed scopes

src/frontend/src/lib/components/icons/OisyWalletLogo.svelte
```diff
@@ -15,7 +15,7 @@
 	<IconAstronautHelmet />
 </div>
 
-<picture aria-label={ariaLabel} class="w-24">
+<picture aria-label={ariaLabel} id="oisy-wallet-logo" class="w-24">
```
peterpeterparker, line 18: Why is this `id` needed, or how will it be used?

daviddecentage: As per the changes section in the text above:

> Added id to main oisy logo so it can be targeted by a selector (needed for inverting it in dark mode later)

daviddecentage:
> I am not really should of the scope of this PR, there are a few tailwind classes that are being introduced but do not exists. And there are mixed scopes

lets maybe reduce the scope to just adding styleClass props to elements that need it later or what do you suggest how to proceed with this PR?

daviddecentage:
> I am not really should of the scope of this PR, there are a few tailwind classes that are being introduced but do not exists. And there are mixed scopes

@AntonioVentilii lets maybe reduce the scope to just adding styleClass props to elements that need it later or what do you suggest how to proceed with this PR?

src/frontend/src/lib/components/icons/OisyWalletLogo.svelte
peterpeterparker: What do you mean by this - i.e technically, how are you going to target the selector? Can you share the code snippet please.

AntonioVentilii:
@daviddecentage @peterpeterparker I suggest we close this PR and do more scoped changes, small, medium or large. The intention is to:

- make new code/feature available to other devs and reviewers (example, we are using `bg-primary` instead of `bg-white`, but new components are being added without knowing this)
- avoid missing leftover non-existing classes (example, `border-brand-sublte-alt`)
- avoid code that is either non necessary, or that could be refactored (example, custom classes that are used only once)
- separate changes in margin, spacing, or similar, so that we can analyse them correctly
- analyse correctly the changes

peterpeterparker:
Sounds good to me! My only input here is regarding the introduction of the ID, which doesn’t seem idiomatic. I'm happy to review this in a dedicated PR.

PR #4702, opened 2025-02-13
fix(frontend): fallback for SPL metadata

AntonioVentilii:
# Motivation

It may happen that the service to fetch SPL metadata fails (too many calls to Quicknode API). However, since it is not critical information, and they are anyway refreshed, we can make `name` and `symbol` fallback to the `address`


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4705, opened 2025-02-13
refactor(frontend): move dApp descriptions to env folder

AntonioVentilii:
# Motivation

For consistency, we move the dApp descriptions to the $env folder.

### DIFF
--- a/src/frontend/src/sol/services/spl.services.ts
+++ b/src/frontend/src/sol/services/spl.services.ts
@@ -255,14 +255,22 @@ export const getSplMetadata = async ({
 }): Promise<TokenMetadata> => {
 	const decimals = await getTokenDecimals({ address, network });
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/types/dapp-description.ts
```diff
@@ -36,7 +35,7 @@ const CarouselDappDescriptionSchema = z.object({
 	callToAction: z.string()
```
BonomoAlessandro, line 35: Should the schema not also be moved into its own file?

AntonioVentilii: yes indeed, but I will do it in another PR, just to be more scoped

BonomoAlessandro (approved):
LGTM, thx

PR #4706, opened 2025-02-13
refactor(frontend): bg-dust --> bg-disabled

AntonioVentilii:
# Motivation

It makes more sense that we use `bg-disabled` instead of `bg-dust` in all its usages. Apart from `bg-dust/30` that is `bg-secondary`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4710, opened 2025-02-13
feat(frontend): adjust Convert components for ETH fees case

DenysKarmazynDFINITY:
# Motivation

In case of ETH -> ckETH conversion, we need to operate with total (max) and min gas fee: the total fee needs to be used for calculating max transaction amount and the min fee - for validating user input. For the reference, this is how it currently works in the old conversion flow (https://github.com/dfinity/oisy-wallet/blob/main/src/frontend/src/eth/components/send/EthSendAmount.svelte#L52).


AntonioVentilii (approved):
LGTM tks!

PR #4714, opened 2025-02-14
fix(e2e): fix status directory issue

inc-man:
# Motivation

To fix the update snapshots task, we specify the paths of the file to stage and commit, since it was taking them all by default (so including the TXT files used by the CI).

# Tests

ran CI on this branch


AntonioVentilii (changes requested):
I think it is best to analyse when this error surged and maybe revert that? or modify it

.github/actions/create-pr/action.yml
```diff
@@ -15,6 +15,9 @@ inputs:
   branch:
     description: 'Branch name for the PR'
     required: true
+  path:
+    description: 'Path of the changes to be staged'
+    required: true
```
AntonioVentilii, line 20: is it possible to make it optional? so that it defaults to including everything? it would make it easier so that we don't change the other usages of this action

.github/actions/create-pr/action.yml
```diff
@@ -28,6 +31,7 @@ runs:
         committer: GitHub <noreply@github.com>
         author: gix-bot <gix-bot@users.noreply.github.com>
         branch: ${{ inputs.branch }}
+        path: ${{ inputs.path }}
```
AntonioVentilii, line 34: i think the variable to change is `add_path` no? https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#action-inputs

AntonioVentilii (changes requested):

.github/actions/create-pr/action.yml
inc-man: we can just do required: false

but then keep in mind that when it is not required anymore you need to stage the changes again. So I think no matter what adding the path is a better use case.

.github/actions/create-pr/action.yml
inc-man: youre right I saw path and thought if there is just one this is the syntax to use. But indeed add_paths is the one to use here.

.github/actions/create-pr/action.yml
AntonioVentilii: doesn't it include everything by default? meaning, if we don't pass `add_path` , it will stage all changed files, right?

inc-man: yeah right, then I will just remove it

AntonioVentilii (approved):
LGTM tks

PR #4686, opened 2025-02-12
feat(frontend): Dark Theme

daviddecentage:
# Motivation

# Changes

# Tests

PR #4717, opened 2025-02-14
refactor(frontend): rename tailwind border color classes

daviddecentage:
# Motivation
As part of the new theme refactoring https://github.com/dfinity/oisy-wallet/pull/4686 , we rename a few bordr-* and ring-* (because ring classes reference border colors) colors that were hardcoded colors and map them to theme variables.


AntonioVentilii (approved):
LGTM tks

PR #4718, opened 2025-02-14
test(e2e): new test for enabling all testnets

inc-man:
# Motivation

We want to have a new test that enables all the available testnets and checks if they show correctly.

# Changes

created the whole new test

# Tests

ran the CI on this branch


AntonioVentilii (commented):
would it be possible to use a `[...].forEach(() =>{ test....} )` structure?

e2e/utils/pages/testnets.page.ts
```diff
@@ -0,0 +1,68 @@
...
+	},
+	{
+		name: 'enable SOL (Devnet)',
+		networkSymbol: 'SOL (Devnet)',
+		tokenSymbol: 'SOL (Devnet)'
+	},
+	{
+		name: 'enable SOL (Local)',
+		networkSymbol: 'SOL (Local)',
+		tokenSymbol: 'SOL (Local)'
```
AntonioVentilii, line 41: are the `tokenSymbol` correct?

AntonioVentilii (commented):
I like it!!! I will re-use the logic

e2e/utils/pages/testnets.page.ts
inc-man: they are. since its always the "default token" in the corresponding network. this step can actually be left out, its just "additional" security that the testnet is showing correctly.

PR #4720, opened 2025-02-14
chore(ci): remove unnecessary staging of new files

AntonioVentilii:
# Motivation

When we use the internal `action/create-pr` it is unnecessary to stage the changed files beforehand, since we are using `peter-evans/create-pull-request` that [add new and changed files](https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#action-inputs).

However, sometimes we use the staged files to check what are the files changed, so we will keep that for now.


inc-man (approved):
LGTM, tks

PR #4728, opened 2025-02-14
feat(frontend): workers for user snapshots

AntonioVentilii:
# Motivation

Very similar to the Auth and the Exchange workers, we create workers for the User snapshot. They will be triggered manually (when needed) and every 5 minutes.


src/frontend/src/lib/services/worker.user-snapshot.services.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const initUserSnapshotWorker = async (): Promise<UserSnapshotWorker> => {
+	const UserSnapshotWorker = await import('$lib/workers/workers?worker');
+	const worker: Worker = new UserSnapshotWorker.default();
+
+	worker.onmessage = ({ data }: MessageEvent<PostMessage<PostMessageDataResponse>>) => {
+		const { msg } = data;
+
+		switch (msg) {
+			case 'syncUserSnapshotError':
```
DenysKarmazynDFINITY, line 18: Just to be sure - we don't need to listen to success messages in case of this worker, right?

DenysKarmazynDFINITY (approved):
Nicely done!

src/frontend/src/lib/services/worker.user-snapshot.services.ts
AntonioVentilii: correct, it works by itself and we really don't care what happens after (for now)

PR #4730, opened 2025-02-14
refactor(frontend): adjust nav items components styling

daviddecentage:
# Motivation

As part of the new theme refactoring https://github.com/dfinity/oisy-wallet/pull/4686 , we adjust all nav items to the new figma design, adjusting classes and css on elements


src/frontend/src/lib/components/navigation/ChangelogLink.svelte
```diff
@@ -3,12 +3,20 @@
 	import ExternalLink from '$lib/components/ui/ExternalLink.svelte';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+
+	export let asMenuItem = false;
+	export let asMenuItemCondensed = false;
+
+	let styleClass = '';
+	if (asMenuItem) styleClass += ' nav-item';
+	if (asMenuItemCondensed) styleClass += ' nav-item-condensed';
```
AntonioVentilii: let's do this directly in the html part with an if condition, and are they exclusive one another?

src/frontend/src/eth/components/core/EthWalletAddress.svelte
```diff
@@ -23,7 +23,11 @@
 </div>
 
 {#if nonNullish(explorerUrl)}
-	<ExternalLink href={explorerUrl} ariaLabel={$i18n.wallet.alt.open_etherscan}>
+	<ExternalLink
+		styleClass="nav-item"
+		href={explorerUrl}
+		ariaLabel={$i18n.wallet.alt.open_etherscan}
+	>
```
AntonioVentilii, line 30: since it is used a lot with `ExternalLink`, let's add a prop directly inside the component `asNavItem`

src/frontend/src/lib/components/navigation/ChangelogLink.svelte
```diff
@@ -3,12 +3,24 @@
...
+	export let asMenuItem = false;
+	export let asMenuItemCondensed = false;
+
+	let styleClass = '';
+	if (asMenuItem) {
+		styleClass += ' nav-item';
+	}
+	if (asMenuItemCondensed) {
+		styleClass += ' nav-item-condensed';
+	}
```
AntonioVentilii, line 16: let's put in if condtion in the html prop

src/frontend/src/lib/components/navigation/ChangelogLink.svelte
daviddecentage: Yes theyre exclusive. If we have this condition in HTML it will be a ternary in a ternary which isnt very pretty. What do you think?

src/frontend/src/eth/components/core/EthWalletAddress.svelte
daviddecentage: Done

src/frontend/src/lib/components/navigation/ChangelogLink.svelte
daviddecentage: Done

AntonioVentilii (approved):
LGTM tks

PR #4731, opened 2025-02-14
refactor(frontend): landing page dark mode images

daviddecentage:
# Motivation

As part of the new theme refactoring #4686, we add a dark variant and updated version for light landing page image


src/frontend/src/lib/components/auth/LandingPage.svelte
```diff
@@ -21,7 +21,9 @@
 		class="pt-12 md:m-0 md:flex md:h-full md:content-center md:items-center ml-auto min-w-[1127px]"
 	>
 		<div class="md:h-md:mt-auto w-full">
-			<Img src={preview} alt={ariaLabel} />
+			{#await import(`$lib/assets/main_image-${$themeStore}.webp`) then { default: src }}
```
AntonioVentilii, line 24: `$themeStore` may be `undefined`, we need a fallback i think

AntonioVentilii (changes requested):

src/frontend/src/lib/components/auth/LandingPage.svelte
daviddecentage: done

src/frontend/src/lib/components/auth/LandingPage.svelte
```diff
@@ -1,12 +1,14 @@
...
+	import { themeStore } from '@dfinity/gix-components';
 	import HeroSignIn from '$lib/components/hero/HeroSignIn.svelte';
 	import Img from '$lib/components/ui/Img.svelte';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
 
 	let ariaLabel: string;
 	$: ariaLabel = replaceOisyPlaceholders($i18n.auth.alt.preview);
+
+	let bassa = null;
```
AntonioVentilii: ?

AntonioVentilii (changes requested):

src/frontend/src/lib/components/auth/LandingPage.svelte
daviddecentage: sorry left the test in. Done now

peterpeterparker (commented):
Non gzipped, this PR adds 200% weight for the same banner (105kb instead of 85kb) to the JavaScript. Is the team aware of it and fine with it?

daviddecentage:
> Non gzipped, this PR adds 20% weight for the same banner (105kb instead of 85kb) to the JavaScript. Is the team aware of it and fine with it?

Its two images now though, before the preview was 82kb and now each image is a bit over 50kb. Since only one of the images is loaded at a time, Id assume it to load even faster than before.

peterpeterparker:
No strong opinion. As I said in my question, I was just asking if the team were aware of it and fine with it. For sure if the team is fine, all good.

Note that I assume, given the images' size, they are not inlined within the JS bundle. If that’s not the case, then I think the extra size would be an issue though.

AntonioVentilii (commented):
Since it was accepted the extra slave taken by the new images, approved

AntonioVentilii (approved):

PR #4734, opened 2025-02-14
refactor(frontend): adjust dropdowns according to new button styling

daviddecentage:
# Motivation
As part of the new theme refactoring https://github.com/dfinity/oisy-wallet/pull/4686 , we adjust dropdown buttons and items styling to match the new button styles


src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
```diff
@@ -68,7 +68,7 @@
 
 		<ul class="gap-4 font-normal flex list-none flex-col">
 			<li class="flex items-center justify-between">
-				<div class="gap-2 flex items-center">
+				<div class="gap-2 dropdown-item disabled flex items-center">
```
DenysKarmazynDFINITY, line 71: Are you adding some specific value via `disabled` class? If yes, maybe we can have another class with the same CSS rules? I would reserve the "disabled" keyword only for the disabled state purposes

src/frontend/src/lib/components/ui/DropdownButton.svelte
```diff
@@ -9,14 +9,14 @@
...
-	class="min-w-72 gap-2 rounded-xl px-4 py-3 leading-5 font-medium justify-between border border-tertiary bg-primary text-left text-inherit text-primary hover:border-brand-primary"
+	class="dropdown-button"
 	bind:this={button}
 	on:click
 	aria-label={ariaLabel}
 	data-tid={testId}
 	{disabled}
-	class:opacity-50={disabled}
-	class:border-brand-primary={opened}
+	class:disabled
```
DenysKarmazynDFINITY, line 18: If it doesn't expect any variables, let's move this and the next lines to `class="..."` property

daviddecentage: It expects a variable, but if the variable is called the same as the prop the formatter will shorten it:

`class:disabled={disabled} === class:disabled`

src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
daviddecentage: it should appear as a disabled select option, that wasnt like this before but specified by Figma. In figma it has "disabled" state

DenysKarmazynDFINITY: in that case, all good

PR #4735, opened 2025-02-14
refactor(frontend): adjust default backround and text colors

daviddecentage:
# Motivation
As part of the new theme refactoring https://github.com/dfinity/oisy-wallet/pull/4686 , we adjust the root backround and text color by referencing theme variables instead of hardcoded colors.


src/frontend/src/app.html
```diff
@@ -80,7 +80,8 @@
 
 		<style>
 			:root {
-				background: #e8f1ff;
+				background: var(--color-background-page);
```
peterpeterparker, line 83: This cannot work, `--color-background-page` and `--color-foreground-primary` are undefined when the `index.html` is rendered.

daviddecentage: This just points to variables and whenever they become available the colors are applied. You can check out the branch and change the varables to some popping color to see it work.

Also, these styles are loaded after the compiled SASS files so they're already available when this rule is set.

peterpeterparker: > This just points to variables and whenever they become available the colors are applied.

How does it look when the colors are not applied? Does the UI produce any flashes notably on first load? Can you provide screenshots or screen recording and add them to the PR?

> Also, these styles are loaded after the compiled SASS files so they're already available when this rule is set.

SASS is a build processor. CSS variables are loaded when their related CSS files are loaded, which, as you can observe in the network tab, happens after the `index.html` page is rendered. For example, see this screenshot on 3G.

Capture d’écran 2025-02-14 à 20 07 03

daviddecentage: I understood you wrong, I thought you meant the rule here doesnt work at all. Also, I guess I was already checked out on friday as youre completely right about it being loaded after the index (obviously).

There is a short flicker of the screen. Do you think we should rather hardcode the default background and text but have them under a `html[theme='themename']:root` rule? This would reduce the flicker to a minimum at cost of it being declared a second time (the other being in the dedicated theme SCSS itself).

peterpeterparker: Personally, I think there should be no flickering between themes. Just my two cents, of course.

Therefore, I would try the following:

- Move the hardcoded `script` in the HTML page before (above) the `style` tag that sets the background.
- In the `style` section of the HTML page, hardcode the background color according to `html[theme]`.

```
<html>
   <head>
      <script -> set html[theme]
      <style>
          :root background: #fff -> light background
          :root[theme=dark] background: #eee -> dark background
```

This way, even though using the script before the style may cause the browser’s default background color to appear for a few milliseconds more but, it should lead to no flickering between the themes which might be less visible.

Not 100% sure, I would give it a try.

daviddecentage: Not flickering between theme change, only until the variables are available/CSS is loaded.

Okay, so we will hardcode the theme values in index.html for best performance.

peterpeterparker (commented):
Looks good. Just two minor questions.

src/frontend/src/app.html
```diff
@@ -1,6 +1,37 @@
 <!doctype html>
 <html lang="en">
 	<head>
+		<!-- Init theme as fast as possible -->
+		<script theme-loader>
```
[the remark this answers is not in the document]
daviddecentage: Yes copied 100%, just moved it to right after start head tag

src/frontend/src/app.html
```diff
@@ -1,6 +1,37 @@
...
+
+				document.documentElement.setAttribute(
+					'theme',
+					currentTheme ?? (isDarkPreferred ? 'dark' : 'light')
+				);
+			} catch (error) {
+				console.error('Error initializing theme', error);
+			}
+		</script>
+		<style>
```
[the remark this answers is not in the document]
daviddecentage: Done :)

PR #4737, opened 2025-02-14
feat(frontend): add worker component for user snapshots

AntonioVentilii:
# Motivation

We include a worker that will start and trigger a user snapshot every 5 minutes and also triggered by changes in the following parameters (for now):

- **Addresses**: IC principal and Solana addresses.
- **Tokens**: ICRC tokens and Solana tokens.
- **Balances**: All balances (since we need to check if the user has any balance).
- **Exchanges**: All exchanges initialized (since we have no disclaimer specific for the tokens we are interested in).
- **Transactions**: IC and Solana transactions.

Furthermore, we trigger the first snapshot as soon as the user is logged in and everything is correctly loaded.


src/frontend/src/lib/components/airdrops/UserSnapshotWorker.svelte
```diff
@@ -0,0 +1,112 @@
...
+	// Exchanges: All exchanges (since we have no disclaimer specific for the tokens we are interested in).
+	// Transactions: IC and Solana transactions.
+	$: $authSignedIn,
+		$solAddressMainnet,
+		$solAddressDevnet,
+		$sortedIcrcTokens,
+		$splTokens,
+		$enabledSolanaTokens,
+		$balancesStore,
+		$exchanges,
```
DenysKarmazynDFINITY, line 106: I think this one will cause the most re-triggers, let's see how it works when we have all the pieces together

DenysKarmazynDFINITY (approved):
LGTM, thanks. Let's see what we can do with the worker problem, but for now a simple timer will do the job.

src/frontend/src/lib/components/airdrops/UserSnapshotWorker.svelte
AntonioVentilii: You are right! The should have been `exchangeNotInitialized`, good catch! Changed!

PR #4739, opened 2025-02-14
feat(frontend): create store exchangeNotInitialized

AntonioVentilii:
# Motivation

Since we are going to use it more and since it is already used twice, we create store `exchangeNotInitialized` as opposed to `exchangeInitialized`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4745, opened 2025-02-15
feat(ci): use job output instead of TXT files for E2E snapshots

AntonioVentilii:
# Motivation

We remove the usage of TXT files to check changes in E2E snapshots in the respective CI, in favor of using jobs outputs.


inc-man (approved):
Really good change, was afraid this wouldn't work, hence the txt file. But good catch!

LGTM

PR #4746, opened 2025-02-15
chore(frontend): remove unused SPL transactions services

AntonioVentilii:
# Motivation

Since we use a single service for both SOL and SPL transactions, we can remove the unused SPL transactions service.


DenysKarmazynDFINITY (approved):
👍

PR #4747, opened 2025-02-15
fix(e2e): avoid flakiness due to page scrolling

AntonioVentilii:
# Motivation

Some E2E snapshots are too flaky, in part because they scroll a bit randomly until they can find a specific element (for example the enable/disable tokens.

To avoid this, we make the internal function `takeScreenshot` to scroll back to top, since all of our manage-tokens test do not require a specific position for now. Just in case, this option can be avoided if needed.


inc-man (approved):
LGTM, tks

e2e/utils/pages/homepage.page.ts
```diff
@@ -53,6 +53,9 @@ interface WaitForModalParams {
 	state?: 'detached';
 }
 
+interface TakeScreenshotParams {
+	scrollToTop?: boolean;
+}
```
AntonioVentilii, line 58:
```suggestion
}




```

AntonioVentilii, line 58:
```suggestion
  }   





```

PR #4749, opened 2025-02-16
fix(frontend): adding SPL token raises the correct error

AntonioVentilii:
# Motivation

When we fetch the metadata of a possible custom SPL token, we show an error if the metadata are non-retrievable. However, the common behaviour for the other token types (for example ERC20) is to return `nullish` metadata and let the consumer handle the issue.

In our case, the existing `SolAddTokenReview` already handles it. The missing piece is ignoring errors raised by QuickNode API due to missing metadata and return `nullish` values.

# Changes

- Rest function `splMetadata` will pass `undefined` in case of defined errors in the QuickNode API (for example if the token address is not in the database, but even in case of failure in the connection, and so forth).
- Service `getSplMetadata` will return nullish value for the case above too.
- Adapt the rest of the usages to the possibility of nullish results from `getSplMetadata`.

# Tests

Tested correctly with a custom SPL token that has metadata and with a random SOL address (that obviously has no metadata):

Screenshot 2025-02-17 at 00 13 11

Screenshot 2025-02-17 at 00 20 07


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4750, opened 2025-02-16
fix(frontend): limit the symbol length for SPL tokens

AntonioVentilii:
# Motivation

The backend limits any symbol to max 20 characters. Since we use the address of a token as fallback in case of missing metadata, we limit the symbol there too, to a fair number of unique first characters.


AntonioVentilii:
I will solve the issue in a more consistent way

PR #4751, opened 2025-02-17
feat(frontend): add owner address to SPL tokens

AntonioVentilii:
# Motivation

Since we will start to support Solana Token-2022, it is important that we define what is the owner address for each SPL token.

# Changes

- Add `TOKEN_PROGRAM_ADDRESS` to existing SPL tokens.
- Fetch owner address in when creating an SPL.

# Tests

Existing tests are sufficient.


BonomoAlessandro (approved):
LGTM

PR #4752, opened 2025-02-17
feat(frontend): call to get SPL token's owner

AntonioVentilii:
# Motivation

We will start to support Solana Token-2022. So, it will be important to have a call to fetch the token owner address.


src/frontend/src/sol/api/solana.api.ts
```diff
@@ -225,3 +226,24 @@ export const getTokenDecimals = async ({
...
+}: {
+	address: SplTokenAddress;
+	network: SolanaNetworkType;
+}): Promise<SplTokenAddress | undefined> => {
+	const { getAccountInfo } = solanaHttpRpc(network);
+	const token = solAddress(address);
+
+	const { value } = await getAccountInfo(token, { encoding: 'jsonParsed' }).send();
+
+	if (nonNullish(value)) {
```
DenysKarmazynDFINITY, line 242: Maybe just `return value?.owner?.toString()`?

DenysKarmazynDFINITY (approved):
Approving already, but please take a look at the comment

src/frontend/src/sol/api/solana.api.ts
AntonioVentilii: done! thank you!

PR #4756, opened 2025-02-17
build(frontend): bump agent-js v2.3.0

peterpeterparker:
# Motivation

Agent-js v2.1.3 and v2.2.0 are buggy. A patch has been released in v2.2.1 and v2.3.0 which rolls back a feature introduced in v2.1.3. 

We upgrade to latest given there were no clean patch version provided and we also gonna bump the version in the signer.


AntonioVentilii (approved):
Thank you for the fix!

PR #4758, opened 2025-02-17
refactor(frontend): refactor progress steps styling to use new color vars

daviddecentage:
# Motivation

As part of the new theme refactoring #4686 , we adjust the progress steps styling to use the new color variables introduces with the theme refactoring.


src/frontend/src/lib/styles/global/gix.scss
```diff
@@ -244,13 +244,24 @@ div.segment-button {
 	}
 }
 
+div.step .state {
+	color: var(--color-foreground-brand-primary-alt) !important;
```
DenysKarmazynDFINITY, line 248: By any chance there is a css variable that we can use instead (similar as it's done below for `div.step.completed` `svg` and `line`)? We try to avoid `!important` as much as possible.

daviddecentage: Sadly here its not possible as the layer for GIXs own variables is higher up than the layer we define the variables here. Antonio told me to add a comment on top of the gix.scss on how to handle !important. Since the gix.scss file is the last instance where we would touch gix component styling we can tolerate it as long as theres no other way of overriding the style without it.

DenysKarmazynDFINITY: Ok, I see, thanks for the explanation and the usage comment.

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4761, opened 2025-02-17
fix(frontend): do not show toast errors on Quicknode failures

AntonioVentilii:
# Motivation

As continuation of https://github.com/dfinity/oisy-wallet/pull/4749 , in service `getSplMetadata`, we now avoid showing toast errors for QuickNode failures and just return `undefined` value that the consumer will treat as in https://github.com/dfinity/oisy-wallet/pull/4749.


src/frontend/src/sol/services/spl.services.ts
```diff
@@ -263,25 +263,30 @@ export const getSplMetadata = async ({
...
+		return {
+			decimals,
+			name,
+			symbol,
+			icon
+		};
+	} catch (err: unknown) {
+		// We care only for development purposes.
+		console.warn(`Failed to fetch SPL metadata for token ${address} on ${network} network`, err);
+	}
```
AntonioVentilii, line 291: i just wrapped the entire function in a try/catch statement

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4763, opened 2025-02-17
feat(frontend): implements airdrop modal image banner

BonomoAlessandro:
# Motivation

A user should be able to see the image banner of the airdrop campaing modal. If the user has already received some airdrops, they should see the total received amount. Otherwise, they should see an alternative message.

# Changes

- implements airdrop campaign details banner

# Tests

**New:**

no received airdrops:
image

already received airdrops:

https://github.com/user-attachments/assets/24acf8f5-fc5e-41c3-b616-adb03b001114


src/frontend/src/lib/components/airdrops/AirdropBanner.svelte
```diff
@@ -0,0 +1,48 @@
...
+	import { getAirdrops } from '$lib/services/reward-code.services';
+	import type { AirdropInfo } from '$lib/types/airdrop';
+	import { getAirdropsBalance } from '$lib/utils/airdrops.utils';
+	import { usdValue } from '$lib/utils/exchange.utils';
+
+	const token = ICP_TOKEN;
+
+	let airdrops: AirdropInfo[] | undefined;
+	onMount(async () => {
+		if (isNullish($authIdentity)) {
```
DenysKarmazynDFINITY, line 21: Usually, in such cases we call `nullishSignOut` in case `authIdentity` is nullish:
```
  await nullishSignOut();
  return;
}
```

src/frontend/src/lib/components/airdrops/AirdropBannerOverlay.svelte
```diff
@@ -0,0 +1,67 @@
...
+	import Amount from '$lib/components/ui/Amount.svelte';
+	import Button from '$lib/components/ui/Button.svelte';
+	import { AppPath } from '$lib/constants/routes.constants.js';
+	import { networkId } from '$lib/derived/network.derived.js';
+	import { i18n } from '$lib/stores/i18n.store.js';
+	import { isMobile } from '$lib/utils/device.utils.js';
+	import { formatUSD } from '$lib/utils/format.utils.js';
+	import { networkUrl } from '$lib/utils/nav.utils.js';
+
+	export let token;
```
DenysKarmazynDFINITY, line 15: type is missing, e.g. `token: Token | undefined`

src/frontend/src/lib/utils/airdrops.utils.ts
```diff
@@ -38,3 +39,6 @@ export const isUpcomingCampaign = (startDate: Date) => {
 
 	return startDiff > 0;
 };
+
+export const getAirdropsBalance = (airdrops: AirdropInfo[]) =>
+	airdrops.reduce((total, { amount }) => total.add(BigNumber.from(amount)), BigNumber.from(0));
```
DenysKarmazynDFINITY, line 44: Instead of `BigNumber.from(0)` you can use `ZERO` from `app.constants`

DenysKarmazynDFINITY:
@BonomoAlessandro just to be sure, can you please also check if the new svg banner can be compressed? It probably won't be as light the default `banner.svg` (size - 8kb) because of the content, but still maybe we can do better as 55kb.

BonomoAlessandro:
@DenysKarmazynDFINITY I was able to reduce the size of new svg banner from 55kB to 40kB.

PR #4764, opened 2025-02-17
feat(frontend): remove swap LP and gas fees from calculations

DenysKarmazynDFINITY:
# Motivation

The next iteration of SwapFee changes - we remove the LP and gas fees from being displayed and assumed while calculating the total fee amount as they are both already included into a swap deal. In the next step, we will start showing them again but we'll update UI the way it's clear those fees are not to be paid once again.

Screenshot 2025-02-17 at 12 26 32

### DIFF
--- a/src/frontend/src/lib/i18n/en.json
+++ b/src/frontend/src/lib/i18n/en.json
@@ -473,7 +473,7 @@
 			"not_available": "n/a",
 			"value_difference": "Value difference",
 			"total_fee": "Total estimated fee",
-			"token_fee": "Token fee",
+			"token_fee": "Network fee",
 			"approval_fee": "Approval fee",
 			"gas_fee": "Gas fee",
 			"lp_fee": "LP fee",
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/i18n/en.json
```diff
@@ -469,7 +469,7 @@
 			"not_available": "n/a",
 			"value_difference": "Value difference",
 			"total_fee": "Total estimated fee",
-			"token_fee": "Token fee",
+			"token_fee": "Network fee",
```
AntonioVentilii, line 472: maybe just for consistency we call it `network_fee`?

AntonioVentilii (approved):
LGTM tks! Small observation

src/frontend/src/lib/i18n/en.json
DenysKarmazynDFINITY: good point, done

PR #4765, opened 2025-02-17
feat(frontend): map Solana Token-2022 transferChecked instruction

AntonioVentilii:
# Motivation

Since we are starting to support Solana Token-2022, we map the instructions too. The first one (and most common) is `transferChecked` that is basically the same as the "old" Token.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4766, opened 2025-02-17
refactor(frontend): use token owner address for program ID

AntonioVentilii:
# Motivation

The correct usage of the program ID params in function `getTransferInstruction` is not giving an hard coded program, but the correct program address, that now we have for SPL tokens.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4773, opened 2025-02-17
refactor(frontend): Prepare Icons and Logos for dark mode

daviddecentage:
# Motivation

As part of the new theme refactoring https://github.com/dfinity/oisy-wallet/pull/4686, we prepare the logos and icons that need to be inverted or recolored for dark mode.


src/frontend/src/lib/components/ui/RoundedIcon.svelte
```diff
@@ -6,7 +6,7 @@
 </script>
 
 <div
-	class="p-3 relative flex items-center justify-center rounded-full bg-primary ring-2 ring-brand-subtle-10"
+	class="p-3 relative flex items-center justify-center rounded-full bg-primary ring-1 ring-brand-subtle-20"
```
DenysKarmazynDFINITY, line 9: Just to make - `ring-2` -> `ring-1` is intended?

DenysKarmazynDFINITY (approved):
Small q, otherwise looks good

src/frontend/src/lib/components/ui/RoundedIcon.svelte
daviddecentage: Yes, as in Figma

PR #4776, opened 2025-02-17
refactor(frontend): Add dark theme scss and related files

daviddecentage:
# Motivation
As part of the new theme refactoring https://github.com/dfinity/oisy-wallet/pull/4686, we add the scss for dark mode. This is just the mapping to the colors, as long as the selector is not in there will be no visible change from this commit


DenysKarmazynDFINITY (approved):
👍

PR #4777, opened 2025-02-17
refactor(frontend): Enable theme selector cards

daviddecentage:
# Motivation
As part of the new theme refactoring https://github.com/dfinity/oisy-wallet/pull/4686, we enable the theme selector on the settings page. This is the PR which enables dark mode effectively!


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4781, opened 2025-02-17
feat(frontend): update Swap max button calculation

DenysKarmazynDFINITY:
# Motivation

Similar as in the other Swap fees-related places, we need to multiply `sourceTokenFee` by two if it's an icrc2 token to cover transfer and approval fees while calculation the max available balance.

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapMaxBalanceButton.svelte
+++ b/src/frontend/src/lib/components/swap/SwapMaxBalanceButton.svelte
@@ -14,7 +14,8 @@
 	export let amountSetToMax = false;
 	export let errorType: ConvertAmountErrorType = undefined;
 
-	const { sourceTokenBalance, sourceToken } = getContext<SwapContext>(SWAP_CONTEXT_KEY);
+	const { sourceTokenBalance, sourceToken, isSourceTokenIcrc2 } =
+		getContext<SwapContext>(SWAP_CONTEXT_KEY);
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/swap/SwapMaxBalanceButton.svelte
```diff
@@ -30,7 +31,8 @@
 	$: maxAmount = nonNullish($sourceToken)
 		? getMaxTransactionAmount({
 				balance: $sourceTokenBalance,
-				fee: BigNumber.from(sourceTokenFee ?? 0n),
+				// multiply sourceTokenFee by two if it's an icrc2 token to cover transfer and approval fees
+				fee: BigNumber.from((sourceTokenFee ?? 0n) * (isSourceTokenIcrc2 ? 2n : 1n)),
```
AntonioVentilii, line 35: is it ok to multiply `BigNumber` with `BigInt`?

DenysKarmazynDFINITY: The value that is gonna be multiplied is `sourceTokenFee`, it's a `bigint`. The result will be then converted to BigNumber.

AntonioVentilii: insert facepalm... you are completly right, my mistake

AntonioVentilii (approved):
LGTM tks

PR #4782, opened 2025-02-17
fix(frontend): change selected theme when goes from system to others

AntonioVentilii:
# Motivation

Since the `themeStore` does not change if it goes from System to the correspondent equivalent theme, the reactive statement that updates `selectedTheme` is not triggered.

That means, for example, that if the default is light and we switch between `System` and `Light` (and viceversa) only one theme will be always selected.

We fix this triggering the update.


src/frontend/src/lib/components/settings/ThemeSelector.svelte
```diff
@@ -16,16 +16,23 @@
 	const selectTheme = (theme: Theme | typeof THEME_SYSTEM) => {
 		if (theme === THEME_SYSTEM) {
 			themeStore.resetToSystemSettings();
+			console.log($themeStore);
+			updateSelectedTheme()
 			return;
 		}
 
 		themeStore.select(theme);
+		updateSelectedTheme();
```
peterpeterparker, line 25: That's correct?

```
themeStore.select(theme);
// +
updateSelectedTheme();
```

Aren't we fireing twice `updateSelectedTheme()` because of the reactive statement

```
$: $themeStore,updateSelectedTheme()
```

?

AntonioVentilii: indeed, but I left the reactive element just in case the `themeStore` is updated elsewhere that is not in this component. Overkill? I could remove it

peterpeterparker: I would remove this `updateSelectedTheme()` here. This way `selectedTheme` is initalized on mount and on select.

AntonioVentilii: if I remove it from there, the case where the user goes from System to the system default will not change the selected theme card. That is because the value of `$themeStore` does not change:

- It starts as System --> `$themeStore === light`
- I select Dark --> `$themeStore === dark`
- I select System again --> `$themeStore === light`
- I select Light -- > `$themeStore === light` (no changes)


https://github.com/user-attachments/assets/41bf88c3-de0f-4a51-ac1d-0319b4c96f86

AntonioVentilii: another solution may be

`$: $themeStore, selectThemeToggled, updateSelectedTheme();`

where `selectThemeToggled` is changed everytime we call `selectTheme`

AntonioVentilii: another solution may be

```typescript
const selectTheme = (theme: Theme | typeof THEME_SYSTEM) => {
  if (theme === THEME_SYSTEM) {
    themeStore.resetToSystemSettings();
    updateSelectedTheme();
    return;
  }
  
  const oldTheme = $themeStore;
  themeStore.select(theme);
  if (oldTheme === theme) {
    updateSelectedTheme();
  }
};
```

AntonioVentilii: Same issue as the video. I included a test that switch back and forth between themes and checks if the cards are selected. I added the above to the commit "use onMount". But the latest commit is what I was thinking, WDYT?

peterpeterparker: I'll have a look tomorrow.

AntonioVentilii: good one! done!

peterpeterparker (approved):
LGTM, thx

PR #4784, opened 2025-02-17
refactor(frontend): component to lazy-load images

AntonioVentilii:
# Motivation

We have the snippet to lazy-load images used a few times already, so it makes sense to export it in a separate component.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4788, opened 2025-02-18
fix(frontend): Logo is inverteed on light mode with dark system theme

daviddecentage:
# Motivation

When dark system theme is selected but light mode in the app, tailwinds dark:* classes always go for the system setting. We made it a custom class to avoid unexpected results.

# Changes

Made a custom class for inverting logos on dark mode and switched the classes on the elements

# Tests
Before:
image

After:
image

### DIFF
--- a/src/frontend/src/lib/components/about/AboutFeatureItem.svelte
+++ b/src/frontend/src/lib/components/about/AboutFeatureItem.svelte
@@ -8,7 +8,7 @@
 
 <div>
 	<div class="gap-3 font-bold flex items-center text-primary">
-		<div class="dark:invert">
+		<div class="invert-on-dark-theme">
 			<slot name="icon" />
 		</div>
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/about/AboutFeatureItem.svelte
```diff
@@ -8,7 +8,7 @@
 
 <div>
 	<div class="gap-3 font-bold flex items-center text-primary">
-		<div class="dark:invert">
+		<div class="text-primary">
```
DenysKarmazynDFINITY, line 11: `text-primary` is correct here?

daviddecentage: yes, I realized we can color the SVGs directly with using text color instead of inverting it with CSS filter

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4791, opened 2025-02-18
fix(frontend): Fix disappeared footer

daviddecentage:
# Motivation

The footer displaying dfinity logo and copyright disappeared, most likely after tailwind 4 upgrade.

Reason for this seems that property "screens" in tailwind config doesnt apply when placed in extend section.

# Changes

Moved custom screen declaration to theme section and added the default ones as well.

# Tests

Behaves exactly like Prod again:

Extra Large screen
image

Large screen
image

Medium Screen
image

Small Screen
image


DenysKarmazynDFINITY (commented):
It looks like you downgraded locally to Tailwind 3 which caused so many unrelated changes - the order of classes is different between the 3rd and 4th version. Please double-check it.

daviddecentage:
> It looks like you downgraded locally to Tailwind 3 which caused so many unrelated changes - the order of classes is different between the 3rd and 4th version. Please double-check it.

Thats strange. I checked and I have v4 locally, it seems the changes would be correct

DenysKarmazynDFINITY (approved):
LGTM, thanks.

For a reference, the linter updated the order of CSS classes because of the missing screen sizes fix.

PR #4796, opened 2025-02-18
fix(frontend): Fix token buttons hover state in token list

daviddecentage:
# Motivation

Brought back the hover state for token buttons in the token list, and adjusted design slightly to make it match Figma.

# Tests

Bildschirmaufnahme_20250218_144427


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4801, opened 2025-02-18
chore(frontend): Cleanup theme refactoring

daviddecentage:
# Motivation
In order to get a seamless transition to the new theming system implemented for #4686 we added the new colors alongside the old ones. Goal of this PR is to cleanup all older and unused colors, or if theyre used, point them to the newly defined colors and styles.

# Changes
- Had to adjust multiple color styles to new color names
- The styled scrollbars were still using fix coded colors, adjusted them to use new alpha color definitions
- Removed colors.scss and each variable has been checked for usages, as well as the usages being now pointed at the new colors
- Styled toasts as they have been left unstyled due to the changes in the theme refactoring (or possibly tailwind upgrade?)
- Removed old base colors and each variable has been checked for usages, as well as the usages being now pointed at the new colors
- Removed default theme old variables and each variable has been checked for usages, as well as the usages being now pointed at the new variables

# Tests

No tests


src/frontend/src/lib/components/manage/ManageTokens.svelte
```diff
@@ -212,13 +212,15 @@
...
+	@use '../../styles/theme/base-colors' as *;
+
 	.tokens {
 		padding: var(--padding-1_5x) 0;
 	}
 
 	.tokens-scroll {
 		&::-webkit-scrollbar-thumb {
-			background-color: rgba(var(--color-black-rgb), 0.2);
+			background-color: #{$a-dark-20};
```
DenysKarmazynDFINITY, line 223: Wondering if we can safely remove this rule - `::-webkit-scrollbar-thumb` defined here has the exact same background color as in the global theme file.

daviddecentage: Good point, removed

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4804, opened 2025-02-18
build(frontend): use Svelte bundle strategy single

peterpeterparker:
# Motivation

Instead of manually hacking / chunking the bundle we can try to rely on the SvelteKit `bundleStrategy` option which supports `single`.

https://svelte.dev/docs/kit/configuration#output

# Changes

# Tests


AntonioVentilii:
Superseded by #9720, which explores the same single-bundle strategy with the newer SvelteKit option. Closing as duplicate.

PR #4818, opened 2025-02-19
chore(repo): gitignore tmp

peterpeterparker:
# Motivation

The `./scripts/test.frontend.reproducibility.sh` generates data in `tmp` folder which is not excluded by git. Particularly when running the script locally, it would be more handy to ignore those changes that should not be committed anyway.


DenysKarmazynDFINITY (approved):
Thanks!

PR #4823, opened 2025-02-19
build(frontend): tailwindcss import instead of use

peterpeterparker:
# Motivation

We faced reproducibility when releasing v0.0.24. See CI job results [here](https://github.com/dfinity/oisy-wallet/actions/runs/13393064153).

After few (many) iterations, we figured out that the root cause of the issue was embededding Tailwind through Sass - using `@use` - instead of using a plain CSS `@import` as described in their documentation.

To validate the fix, we manually run the reproducibility job which succeeded. See results [here](https://github.com/dfinity/oisy-wallet/actions/runs/13408431227).

# Format

Importing Tailwind leads to multiple changes because somehow it modifies the order of the classes.

# Sass warning

Due to this PR and because we do not set up (yet?) a rule to ignore the warning, Sass will log the following warning at build time:

```
More info and automated migrator: https://sass-lang.com/d/import

  ╷
4 │ @import 'tailwindcss';
  │         ^^^^^^^^^^^^^
  ╵
    src/frontend/src/lib/styles/global.scss 4:9  root stylesheet
```

This warning can be ignored. The following is specified in their documentation:

> Eventually, all @import rules will be treated as [plain CSS @imports](https://sass-lang.com/documentation/at-rules/import/#plain-css-imports), likely after an intermediate period where anything that used to be a Sass @import throws an error.

Which is exactly what we want here.

# Changes

- Use `@import` to load Tailwind
- `npm run format`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ backend-v*.wasm.gz
 pocket-ic
 
 coverage/
+
+# Frontend reproducibility script output folder
+tmp
--- a/src/frontend/src/btc/components/core/BtcWalletAddress.svelte
+++ b/src/frontend/src/btc/components/core/BtcWalletAddress.svelte
@@ -18,7 +18,7 @@
 </script>
 
 <div class="p-3">
-	<label class="text-sm font-bold block" for="btc-wallet-address"
+	<label class="block text-sm font-bold" for="btc-wallet-address"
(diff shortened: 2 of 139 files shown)


DenysKarmazynDFINITY (approved):
Amazing finding!

PR #4790, opened 2025-02-18
chore(frontend): migrate to eslint v9

peterpeterparker:
# Motivation

We have to migrate before eslint v8 is fully deprecated and I'm also sick of the warnings.

# Changes

- Bump our library `@dfinity/eslint-config-oisy-wallet`
- Adapt configuration accordingly
- Run `npm run lint --fix` for few minor changes

PR #4825, opened 2025-02-19
build(frontend): bump ic-js to fix agents undefined when destructed

peterpeterparker:
# Motivation

The migration to eslint (#4790) incorrectly revered `@dfinity/utils` to its main release instead of using the `next` version which contained a patch to be able to destruct the agent manager. As a result, the app fails at initializing an agent because the reference is `undefined`.

Instead of bumping to next as the problem might re-surface, we released a new version of ic-js ([2025.02.19-1030Z](https://github.com/dfinity/ic-js/releases/tag/2025.02.19-1030Z)) and this PR update the libraries accordingly.

# Changes

- `npm run update:agent:ic-js`

# Tests

image


DenysKarmazynDFINITY (approved):
Thanks!

PR #4827, opened 2025-02-19
feat(frontend): abort opening Swap modal if kong_backend is down

DenysKarmazynDFINITY:
# Motivation

After recent incidents with kong_backend, we need to improve the swap modal opening behavior to avoid app being stuck if the dapp canister is not responsive.

The updated implementation of Swap component handles the following cases when user clicks the swap button:
1. If `loadKongSwapTokens` succeeds within 10s - show modal.
2. If `loadKongSwapTokens` does not succeed within 10s - show toast, do not show modal.
3. If `loadKongSwapTokens` throws - show toast, do not show modal.

### DIFF
--- a/src/frontend/src/lib/components/swap/Swap.svelte
+++ b/src/frontend/src/lib/components/swap/Swap.svelte
@@ -37,7 +37,7 @@
 		store: icTokenFeeStore
 	});
 
-	const isDisabled = () => isNullish($kongSwapTokensStore);
+	const isDisabled = (): boolean => isNullish($kongSwapTokensStore);
 
 	const loadKongSwapTokens = async (): Promise<'ready' | undefined> => {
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/swap/Swap.svelte
```diff
@@ -36,21 +37,57 @@
 		store: icTokenFeeStore
 	});
 
-	const onOpenSwap = async (tokenId: symbol) => {
+	const isDisabled = () => isNullish($kongSwapTokensStore);
```
peterpeterparker, line 40: - nitpick: `isDisabled = (): boolean =>`
- should we differenciate `undefined` and `null` or in both of its usage checking for both is correct?

peterpeterparker (approved):
LGTM, thx

PR #4828, opened 2025-02-19
style(frontend): Fix onramper widget flickering

daviddecentage:
# Motivation

Onramper light mode has a bug where it doesnt apply the card background color to dropdown options, even though everything is colored correctly according to our dark mode.

To fix this, we always applied dark mode and color it accordingly (light or dark depending on setting of our app).

The problem now, is that in our light mode the widget loads in dark mode first until the colors are applied, leading to not so nice flickering. I tried:
- Passing params directly in iframe src as onramper also supports
- setting the themeName to the theme we currently have selected in oisy wallet
- Combined solution of the previous two
All of those solutions still leave some amount of flickering.

To avoid it as best as possible, I overlaid a spinner covering the widget that disappears in a nice animation once the iframe is loaded. This way we can also have our own spinner instead of the unstyled spinner from onramper.

# Changes

Added spinner overlay

# Tests

Before:
before

After:
after_light
after_dark


src/frontend/src/lib/components/onramper/OnramperWidget.svelte
```diff
@@ -71,6 +72,7 @@
 			themeName: 'dark' // we always pass dark, as some card elements arent styled correctly (white text on white background) in light theme / onramper bug?
 		}));
 
+	$: themeLoaded = false;
```
DenysKarmazynDFINITY, line 75: nitpick: please add `let themeLoaded: boolean` before this line

DenysKarmazynDFINITY (approved):
Small comment, otherwise LGTM

src/frontend/src/lib/components/onramper/OnramperWidget.svelte
daviddecentage: done

PR #4830, opened 2025-02-20
build(backend): enable canister backtraces

peterpeterparker:
# Motivation

We would like to enable adding backtraces to the logs of the backend.

# Documentation

- https://github.com/dfinity/portal/pull/5185
 - https://forum.dfinity.org/t/canister-backtraces/35298/9?u=peterparker
 
 # Notes
 
Support for `--keep-name-section` was added to `ic-wasm` since june 2023 for the `shrink` and `optimize` commands.

`metadata` was added later. While there is no release notes for version v0.8.5, running the following commands we can assume it is supported.

```
❯ ic-wasm --version
ic-wasm 0.8.5
❯ ic-wasm metadata --help
Manage metadata in the Wasm module

Usage: ic-wasm <INPUT> metadata [OPTIONS] [NAME]

Arguments:
  [NAME]  Name of metadata. If not provided, list the current metadata sections

Options:
  -d, --data <DATA>              Content of metadata as a string
  -f, --file <FILE>              Content of metadata from a file
  -v, --visibility <VISIBILITY>  Visibility of metadata [default: private] [possible values: public, private]
  -k, --keep-name-section        
  -h, --help                     Print help
  ```
  
  # Requires
  
  `ic-wasm` needs to be bumped. This is provided in another PR: https://github.com/dfinity/oisy-wallet/pull/4632 

# Changes

- Add `--keep-name-section` to `ic-wasm` commands


AntonioVentilii (approved):
We have another script build.backend.metadata that has the same logic, I am assuming you don't want it there. Or if you want it there, we can always add it.

In any case, LGTM, tks!

bitdivine (approved):
Approving as this follows the documentation but I'm not 100% sure it has all the strings for backtraces:

- Bumped the tool version as in the earlier PR
- Compiled this branch
- Compiled a fresh wasm2wat and related wabt tools, as my exiting installed 2 year old tools didn't parse the Wasm.
- Converted the backend wasm to wat, so that it is nice and readable
- Looked for a couple of random function names.  E.g. `grep next_user_token_chunk out/backend.wat`  I didn't find them in the wat or the Wasm.

I haven't tried causing a panic and it could be that I'm missing something or that I was unlucky and picked some inlined functions.  It could also be that this is a necessary but not sufficient step.

peterpeterparker:
I get spammed by notifications about this PR. Do you plan to approve and merge it or not?

If not, let's close the PR.

PR #4836, opened 2025-02-20
feat(frontend): improves fee display

BonomoAlessandro:
# Motivation

Currently, only the token fee is displayed on the send form, but the USD value is not shown. To improve user-friendliness, the USD value should also be displayed.

# Changes

- displays fee in usd

# Tests

**IC Send:**
before
image

after
image

**ETH Send:**
before
image

after
image

**SOL Send:**
before
image

after
image


src/frontend/src/icp-eth/components/fee/FeeAmountDisplay.svelte
```diff
@@ -28,12 +44,23 @@
...
+		{#if usdFee < SWAP_TOTAL_FEE_THRESHOLD}
+			{`( < ${formatUSD({
+				value: SWAP_TOTAL_FEE_THRESHOLD
+			})} )`}
+		{:else}
+			{`( ${formatUSD({ value: usdFee })} )`}
+		{/if}
+	</div>
 </div>
 {#if insufficientFeeFunds && nonNullish(balance)}
```
DenysKarmazynDFINITY, line 65: By using this component to display IC, BTC and SOL fees, you also added this error message that is not supposed to be displayed anywhere except for ETH (current case in `main`). 

Considering that we tend to put all things that are re-used across at least 2 networks into the `lib` folder, I suggest the following approach:
1. Extract the code related to displaying fee and its USD value into a separate component created in `lib/components/fee` folder. `tokenExchangeRate` needs to be a prop instead of using context. This step can be done as a separate PR.
2. Use the newly created component for SOL, ICP and IC fees.
3. Update `icp-eth/components/fee/FeeAmountDisplay.svelte` to also use it, the error message can be passed as a slot.

BonomoAlessandro: @DenysKarmazynDFINITY Why is this error message not supposed to be displayed for IC, BTC and SOL fees?

I will create a new component in the `lib` folder and use this in the different implementations.

BonomoAlessandro: @DenysKarmazynDFINITY I don't really understand. The `insufficientFeeFunds` is used to check if the `balance` is large enough to cover the `fees`. The amount of tokens i want to send should be irrelevant, right?

DenysKarmazynDFINITY: Nope, it is only working like this with ETH, where `sendAmount` can be in some erc20 token and fee is always paid in ETH. 

With all other tokens, the logic is following:
1. If balance < sendAmount - "insufficient-funds"
2. If balance > sendAmount but balance < sendAmount + fee - "insufficient-funds-for-fee"

Examples for in the code: https://github.com/dfinity/oisy-wallet/blob/main/src/frontend/src/lib/utils/convert.utils.ts#L36 or https://github.com/dfinity/oisy-wallet/blob/main/src/frontend/src/btc/components/send/BtcSendReview.svelte#L35

BonomoAlessandro: @DenysKarmazynDFINITY hmm ok. In this case i will provide those messages in a separate PR and focus this one on displaying the USD value for the token amount.

BonomoAlessandro: @DenysKarmazynDFINITY I have also updated the screenshots for this PR. It now only uses the new `ExchangeAmountDisplay` component and displays the fees on the review form.

src/frontend/src/icp/components/fee/EthereumEstimatedFee.svelte
```diff
@@ -32,11 +35,11 @@
 
 			<div>
 				{#if nonNullish(maxTransactionFee)}
-					<FeeAmountDisplay
-						fee={BigNumber.from(maxTransactionFee)}
-						feeSymbol={feeToken.symbol}
-						feeTokenId={feeToken.id}
-						feeDecimals={feeToken.decimals}
+					<ExchangeAmountDisplay
```
DenysKarmazynDFINITY, line 38: This one should still use `FeeAmountDisplay`, shouldn't it?

BonomoAlessandro: @DenysKarmazynDFINITY yess you are right

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4845, opened 2025-02-21
feat(frontend): implements ExchangeAmountDisplay.svelte

BonomoAlessandro:
# Motivation

There should be a component that displays both the token amount and its corresponding USD value.

# Tests

**New:**

image

image

### DIFF
--- a/src/frontend/src/lib/components/exchange/ExchangeAmountDisplay.svelte
+++ b/src/frontend/src/lib/components/exchange/ExchangeAmountDisplay.svelte
@@ -1,0 +1,45 @@
+<script lang="ts">
+    import { nonNullish } from '@dfinity/utils';
+    import type { BigNumber } from '@ethersproject/bignumber';
+    import { fade } from 'svelte/transition';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/exchange/ExchangeAmountDisplay.svelte
```diff
@@ -0,0 +1,45 @@
...
+	<div transition:fade class="flex gap-4">
+		{formatToken({
+			value: amount,
+			unitName: decimals,
+			displayDecimals: EIGHT_DECIMALS
+		})}
+		{symbol}
+
+		<div class="text-tertiary">
+			{#if usdAmount < SWAP_TOTAL_FEE_THRESHOLD}
```
DenysKarmazynDFINITY, line 36: Not for this PR, but it now makes sense to rename this const to something more general, since it's used outside of swap flow

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4846, opened 2025-02-21
feat(frontend): use correct constant

BonomoAlessandro:
# Motivation

The correct constant should be used.


daviddecentage (approved):
LGTM

PR #4848, opened 2025-02-21
style(frontend): Fix transparent toasts

daviddecentage:
# Motivation

Since we styled toasts like alerts and alerts use alpha background colors, it can be very hard to see the text in some cases. After checking with Dan, we switch to using error/warning/success-secondary background color for toasts.

# Changes

Changed colors for toast bg

# Tests
Before:
image

After:
image

### DIFF
--- a/src/frontend/src/lib/styles/global/gix.scss
+++ b/src/frontend/src/lib/styles/global/gix.scss
@@ -349,11 +349,11 @@ div.select {
 
 // TOASTS
 div.error .toast {
-	--overlay-background: var(--color-background-error-subtle-20);
+	--overlay-background: var(--color-background-error-secondary);
 }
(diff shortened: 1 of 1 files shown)


BonomoAlessandro (approved):
LGTM

peterpeterparker:
Unrelated and just a random observation—sharing in case it's useful. Based on your screenshot, the icon on the left of the toast might not have enough contrast.

daviddecentage:
> Unrelated and just a random observation—sharing in case it's useful. Based on your screenshot, the icon on the left of the toast might not have enough contrast.

Good point. I think for now Ill just merge it so its at least readable. Then, as part of the Figma alignment task we can adjust the final styling

PR #4854, opened 2025-02-22
feat(frontend): changes swap slippage exceeded message

StefanBerger-DFINITY:
# Motivation

Pierre requested the message shown when an in-wallet fails due to exceeded slippage is more user frfiendly.

# Tests
before:
image

after:
image

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapWizard.svelte
+++ b/src/frontend/src/lib/components/swap/SwapWizard.svelte
@@ -101,6 +101,7 @@
 			const errorDetail = errorDetailToString(err);
 
 			if (nonNullish(errorDetail) && errorDetail.startsWith('Slippage exceeded.')) {
+				// Expected slippage is currently not shown in the message anymore, but I'm leaving in its derivation in case it's needed again.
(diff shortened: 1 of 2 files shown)


AntonioVentilii (approved):
LGTM tks

src/frontend/src/lib/components/swap/SwapWizard.svelte
```diff
@@ -101,15 +101,8 @@
 			const errorDetail = errorDetailToString(err);
 
 			if (nonNullish(errorDetail) && errorDetail.startsWith('Slippage exceeded.')) {
-				const expectedSlippageMatch = errorDetail.match(/(\d+(\.\d+)?)% slippage/);
```
peterpeterparker, line 104: We can drop the regex, and there isn’t a single test that needs to be updated.

StefanBerger-DFINITY: I checked for the i18n string and the explicit text. Didn't find another occurrence.
And since the test went through, I assume there is none yet.
Will check with @inc-man to see that we get an E2E test for this case.

PR #4855, opened 2025-02-23
refactor(frontend): use OISY placeholders in i18n

AntonioVentilii:
# Motivation

There are a few places I missed the OISY placeholders in the i18n files. For consistency we use them.


src/frontend/src/lib/components/swap/SwapWizard.svelte
```diff
@@ -102,9 +102,11 @@
 
 			if (nonNullish(errorDetail) && errorDetail.startsWith('Slippage exceeded.')) {
 				failedSwapError.set(
-					replacePlaceholders($i18n.swap.error.slippage_exceeded, {
-						$maxSlippage: slippageValue.toString()
-					})
+					replaceOisyPlaceholders(
```
DenysKarmazynDFINITY, line 105: I think we need to change the order of the functions here - `replacePlaceholders(replaceOisyPlaceholders(...))`. Not sure if it affects the result, but at least we will be consistent with other similar places.

DenysKarmazynDFINITY (approved):
Small comment, approving already

PR #4857, opened 2025-02-23
fix(frontend): timestamp in nanosecond for user snapshot

AntonioVentilii:
# Motivation

The correct value that the Rewards canister accepts is the timestamp in nanoseconds, so we adapt the code where necessary.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4835, opened 2025-02-20
style(frontend): Figma Sync adjust placeholder cards styling

daviddecentage:
# Motivation
We need to synchronize new Figma designs with our application. In this PR we focus on styling the 3 placeholder cards we have for no transactions, no tokens and failed index canister.

# Changes

- Adjusted styling for the cards
- Added new color for alpha 5 oisy blue which is used as card background

# Tests

image
image

PR #4871, opened 2025-02-24
refactor(frontend): Revert placeholder cards

daviddecentage:
# Motivation

Apparently the placeholder styles implemented in #4835 were a test in Figma and not intended to be implemented.

# Changes

Reverted placeholder cards to what it looked like before.

# Tests

image


BonomoAlessandro (approved):
LGTM

PR #4878, opened 2025-02-24
feat(frontend): updates rewards banner

BonomoAlessandro:
# Motivation

On IPhone the rewards banner appears blurry, so we needed to update the image.


daviddecentage (approved):
LGTM, thx

PR #4879, opened 2025-02-24
fix(frontend): iOS background bugfix

daviddecentage:
# Motivation

On iOS devices when switching from light mode to dark mode via the top right icon switcher, a part of the page remained light when scrolling down.

# Changes

To avoid this, we apply the background colors not only to the root/html element, but also to the position: fixed background image container, as the fixed property of that element would stay in place when scrolling.

# Tests

Tested on iPhone 14 Pro Max iOS 17.5.1, iPhone 14 Pro Max iOS v18.3.0 and iPhone 15 iOS v18.3.1 and the visual bug doesnt appear anymore.

Android seemed unaffected and works fine


BonomoAlessandro (approved):
LGTM

PR #4880, opened 2025-02-24
feat(frontend): add Airdrop slide to the carousel

DenysKarmazynDFINITY:
# Motivation

This PR includes the basic changes to include the feature airdrop as a slide to the carousel on the token page. The second PR of this feature should be a refactoring of `DappsCarousel` to make it more general. For that, I left a TODO comment.

Also, I had to move a place where `AirdropModal` was defined - it has to be moved to the global `Modals` component so it can be called from all parts of the app.

# Tests
Adjusted existing tests.


daviddecentage (approved):
LGTM, Thx!

PR #4896, opened 2025-02-25
feat(frontend): updates TokenInput.svelte

BonomoAlessandro:
# Motivation

The `TokenInput` field should also be used in the `SendForms`. Since the custom validation for the `SendAmount` components works with `errors` instead of `errorTypes`, the `TokenInput` should support this as well.

# Changes

- adds support for errors

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokenInput.svelte
+++ b/src/frontend/src/lib/components/tokens/TokenInput.svelte
@@ -72,9 +72,9 @@
 	<div class="mb-2 text-sm font-bold"><slot name="title" /></div>
 
 	<TokenInputContainer
-			{focused}
-			styleClass="h-14 text-3xl"
-			error={nonNullish(errorType) || nonNullish(error)}
+		{focused}
+		styleClass="h-14 text-3xl"
+		error={nonNullish(errorType) || nonNullish(error)}
 	>
 		<div class="flex h-full w-full items-center">
 			{#if token}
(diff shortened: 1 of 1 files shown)


DenysKarmazynDFINITY:
The suggested approach would do the job, but it means we would have 2 different approaches for handling one particular thing. My suggestion would be to adjust the Send components to also use the errorType approach (other way around would be also absolutely fine, but I think it would require more effort). I quickly looked at the existing Send code, shouldn't be a big problem to do so. WDYT?

BonomoAlessandro:
@DenysKarmazynDFINITY I agree. I thought updating the current validation would be risky, but you might be right. I'll take a look at it.

BonomoAlessandro:
@DenysKarmazynDFINITY To be honest, i think it's too much work and too risky to modify all those custom validations. We have 4 different implementations of them and it's definitely more secure to simply enable error support in the `TokenInput` component.

src/frontend/src/lib/components/tokens/TokenInput.svelte
```diff
@@ -25,10 +25,12 @@
 	export let disabled = false;
 	export let placeholder = '0';
 	export let errorType: ConvertAmountErrorType = undefined;
+	export let error: Error | undefined = undefined;
```
DenysKarmazynDFINITY, line 28: Can you please add a todo comment with a short explanation on the refactoring we'd like to do later.

DenysKarmazynDFINITY:
@BonomoAlessandro ok, thanks for checking. Let's add a small comment and proceed for now with the current approach.

DenysKarmazynDFINITY (approved):

PR #4900, opened 2025-02-25
feat(frontend): updates SwapMaxBalanceButton.svelte

BonomoAlessandro:
# Motivation

The `SwapMaxBalanceButton` component should be more generic so that it can also be used in the `SendForms`. Since the custom validation for the `SendAmount` components works with `errors` instead of `errorTypes`, the `MaxBalanceButton` should support this as well.

# Changes

- renames SwapMaxBalanceButton to MaxBalanceButton
- makes component more generic

### DIFF
--- a/src/frontend/src/lib/components/common/MaxBalanceButton.svelte
+++ b/src/frontend/src/lib/components/common/MaxBalanceButton.svelte
@@ -5,7 +5,6 @@
 	import IcTokenFeeContext from '$icp/components/fee/IcTokenFeeContext.svelte';
 	import { IC_TOKEN_FEE_CONTEXT_KEY } from '$icp/stores/ic-token-fee.store';
 	import { i18n } from '$lib/stores/i18n.store';
-	import { SWAP_CONTEXT_KEY, type SwapContext } from '$lib/stores/swap.store';
 	import type { ConvertAmountErrorType } from '$lib/types/convert';
 	import type { OptionAmount } from '$lib/types/send';
 	import { getMaxTransactionAmount } from '$lib/utils/token.utils';
@@ -32,7 +31,7 @@
 	let maxAmount: number | undefined;
 	$: maxAmount = nonNullish(token)
 		? getMaxTransactionAmount({
-				balance: $sourceTokenBalance,
+				balance: balance ?? undefined,
 				// multiply sourceTokenFee by two if it's an icrc2 token to cover transfer and approval fees
 				fee: BigNumber.from((sourceTokenFee ?? 0n) * (isIcrc2Token ? 2n : 1n)),
 				tokenDecimals: token.decimals,
--- a/src/frontend/src/lib/components/common/MaxBalanceButton.svelte
+++ b/src/frontend/src/lib/components/common/MaxBalanceButton.svelte
@@ -1,29 +1,18 @@
 <script lang="ts">
 	import { debounce, isNullish, nonNullish } from '@dfinity/utils';
 	import { BigNumber } from '@ethersproject/bignumber';
-	import { getContext } from 'svelte';
-	import IcTokenFeeContext from '$icp/components/fee/IcTokenFeeContext.svelte';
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/common/MaxBalanceButton.svelte
```diff
@@ -5,43 +5,44 @@
...
-	$: maxAmount = nonNullish($sourceToken)
+	$: maxAmount = nonNullish(token)
 		? getMaxTransactionAmount({
-				balance: $sourceTokenBalance,
+				balance: balance ?? undefined,
 				// multiply sourceTokenFee by two if it's an icrc2 token to cover transfer and approval fees
-				fee: BigNumber.from((sourceTokenFee ?? 0n) * (isSourceTokenIcrc2 ? 2n : 1n)),
-				tokenDecimals: $sourceToken.decimals,
-				tokenStandard: $sourceToken.standard
+				fee: BigNumber.from((sourceTokenFee ?? 0n) * (isIcrc2Token ? 2n : 1n)),
```
DenysKarmazynDFINITY, line 36: We need to receive `fee` value as a prop to make it general enough for the send form. Also, this way you can avoid passing `isIcrc2Token` down to this component.

src/frontend/src/lib/components/common/MaxBalanceButton.svelte
```diff
@@ -5,43 +5,44 @@
...
+	import type { OptionBalance } from '$lib/types/balance';
 	import type { ConvertAmountErrorType } from '$lib/types/convert';
 	import type { OptionAmount } from '$lib/types/send';
+	import type { Token } from '$lib/types/token';
 	import { getMaxTransactionAmount } from '$lib/utils/token.utils';
 
-	export let swapAmount: OptionAmount;
+	export let amount: OptionAmount;
 	export let amountSetToMax = false;
 	export let errorType: ConvertAmountErrorType = undefined;
```
DenysKarmazynDFINITY, line 16: I think in this case we don't need `errorType` - we only check if it's non empty to decide whether an error CSS class should be rendered. Therefore, we can remove it and accept only a boolean `error` instead.

src/frontend/src/lib/components/swap/SwapForm.svelte
```diff
@@ -148,7 +148,14 @@
 
 				<svelte:fragment slot="balance">
 					{#if nonNullish($sourceToken)}
-						<SwapMaxBalanceButton bind:amountSetToMax bind:swapAmount {errorType} />
+						<MaxBalanceButton
+							bind:amountSetToMax
+							bind:amount={swapAmount}
+							{errorType}
```
DenysKarmazynDFINITY, line 154: `error={nonNullish(errorType)`

src/frontend/src/lib/components/common/MaxBalanceButton.svelte
```diff
@@ -5,43 +5,44 @@
...
-
-	const { sourceTokenBalance, sourceToken, isSourceTokenIcrc2 } =
-		getContext<SwapContext>(SWAP_CONTEXT_KEY);
+	// TODO: We want to be able to reuse this component in the send forms. Unfortunately, the send forms work with errors instead of error types. For now, this component supports errors and error types but in the future the error handling in the send forms should be reworked.
```
DenysKarmazynDFINITY, line 17: The comment can be removed after the update

src/frontend/src/lib/components/swap/SwapForm.svelte
```diff
@@ -111,7 +115,7 @@
 					decimals: $sourceToken.decimals,
 					balance: $sourceTokenBalance,
 					// multiply sourceTokenFee by two if it's an icrc2 token to cover transfer and approval fees
```
DenysKarmazynDFINITY, line 117: Can you please remove this comment? We already have it above in this file.

DenysKarmazynDFINITY (approved):
Nice!

PR #4902, opened 2025-02-26
feat(frontend): updates eth send form

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/eth/components/send/EthSendAmount.svelte
+++ b/src/frontend/src/eth/components/send/EthSendAmount.svelte
@@ -1,11 +1,10 @@
 <script lang="ts">
-	import { debounce, isNullish, nonNullish } from '@dfinity/utils';
+	import { isNullish, nonNullish } from '@dfinity/utils';
 	import { BigNumber } from '@ethersproject/bignumber';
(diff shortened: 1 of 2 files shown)


src/frontend/src/eth/components/send/EthSendAmount.svelte
```diff
@@ -73,44 +69,40 @@
...
-		? undefined
-		: (): number =>
-				getMaxTransactionAmount({
-					balance: $sendBalance ?? ZERO,
-					fee: $maxGasFee,
-					tokenDecimals: $sendTokenDecimals,
-					tokenStandard: $sendTokenStandard
-				});
-
-	const onInput = () => evaluateFee?.();
```
DenysKarmazynDFINITY, line 88: Has this part been replaced in the new implementation? Meaning is fee being re-fetched on user input? As far as I can see, it's missing. 

P.S.: functionally speaking, it won't break anything as we keep updating gas values every 30s. Just making sure that we keep the existing features in the updated forms.

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4904, opened 2025-02-26
feat(frontend): updates ic send form

BonomoAlessandro:
# Motivation

The **IC send form** should be updated and match the updated design.

# Changes

- updates form

# Tests

**before:**
image

**after:**
image


BonomoAlessandro:
@DenysKarmazynDFINITY The e2e-tests are expected to fail for the moment. This is because we are not able to update the e2e-snapshots now. I try to find a solution as soon as possible.

DenysKarmazynDFINITY (approved):
Approving already so you can proceed when the e2e issue is resolved.

PR #4905, opened 2025-02-26
fix(frontend): Dont use subscribe for listening to store change when changing theme

daviddecentage:
# Motivation

This is a fix following #4901. We change updating the selector cards from using subscribe on the store to using a reactive statement as suggested by @AntonioVentilii and to stay consistent.

# Changes

Removed themeStore.subscribe(initSelectedTheme) and replaced with $: initSelectedTheme($themeStore)

# Tests

ezgif-664e1b57189e91


BonomoAlessandro (approved):
LGTM

src/frontend/src/lib/components/settings/ThemeSelector.svelte
```diff
@@ -28,15 +28,15 @@
...
 		selectedTheme = isNullish(localStorage.getItem(THEME_KEY))
 			? THEME_SYSTEM
-			: ($themeStore ?? THEME_SYSTEM);
+			: (themeStore ?? THEME_SYSTEM);
 	};
 
-	onMount(initSelectedTheme);
-	themeStore.subscribe(initSelectedTheme);
-	afterNavigate(initSelectedTheme);
+	onMount(() => initSelectedTheme($themeStore));
```
peterpeterparker, line 37: Do you still `onMount` if you add an auto-subscriber? Isn't it triggered on mount?

peterpeterparker (changes requested):
I did not check how the button in the navbar has been implemented but, aren't you doubling the calls to the theme store by subscring to changes?

Blocking the PR because it was approve to prevent it to move forward, just in case my questions make sense.

src/frontend/src/lib/components/settings/ThemeSelector.svelte
daviddecentage: It is, I removed onMount and afterNavigate. I dont think either of them is needed anymore

peterpeterparker (approved):
Discussed offline, there is no double update of the store nor infinite loop. We add a comment for a bit of readability.

LGTM, thx

PR #4909, opened 2025-02-26
feat(frontend): adjust convert utils for ERC20 case

DenysKarmazynDFINITY:
# Motivation

This is the first step of preparing EthConvertTokenWizard to handle ERC20 conversions - convert utils should be updated to tackle ERC20 case where the fee is always paid in ETH.


AntonioVentilii (approved):
LGTM tks!

PR #4911, opened 2025-02-26
chore(frontend): remove unused snapshot worker

AntonioVentilii:
# Motivation

Since we are ok using a worker inside a component directly, for the user snapshots we remove all the scheduler and workers that were created.


DenysKarmazynDFINITY (approved):
👍

PR #4912, opened 2025-02-26
feat(frontend): update ERC20 conversion flow

DenysKarmazynDFINITY:
# Motivation

In this PR, the new conversion flow is enabled for the erc20 tokens. To do that, we needed to update EthConvertTokenForm to cover both ETH and ERC20 cases (similar as it was before with EthSendTokenForm).

Screenshot 2025-02-26 at 14 08 24


AntonioVentilii (approved):
LGTM tks

PR #4915, opened 2025-02-26
fix(frontend): Detect system theme changes and adjust app theme accordingly

daviddecentage:
# Motivation

If "System" theme is selected and the user changes his system theme, the app does update the theme only after a reload. We fix that by utilizing the new gix component SystemThemeListener which updates the store once system theme changes.

# Changes

Added SystemThemeListener component to root layout

# Tests


AntonioVentilii (commented):
Shouldn't we remove/change the `<ThemeSelector>` component too?

daviddecentage:
> Shouldn't we remove/change the `<ThemeSelector>` component too?

Why that? This is just a listener for updating the themeStore when system theme changes. The themeSelector is the manual selection of a fixed theme "light" or "dark" or default system theme

AntonioVentilii:
Indeed, referring to that PR shouldn't something be changed? Because for example there is the comment

`// even though we call initSelectedTheme on store change, changing to system theme will not update the store so we manually update the selectedCard`

BUT the logic is not that anymore after this PR, right?

daviddecentage:
> Indeed, referring to that PR shouldn't something be changed? Because for example there is the comment
> 
> `// even though we call initSelectedTheme on store change, changing to system theme will not update the store so we manually update the selectedCard`
> 
> BUT the logic is not that anymore after this PR, right?

No this logic is still needed.

What you mean is the logic to display which ThemeSelectorCard is selected. This is isolated logic only for the <ThemeSelector/> component.

The change in this PR is only for updating themeStore when OS theme changes.

AntonioVentilii (approved):
Thank you for the explanation! LGTM tks!

PR #4921, opened 2025-02-26
fix(e2e): freeze carousel for flaky test

AntonioVentilii:
# Motivation

There are some E2E tests that are a bit flaky because the carousel is not frozen. Since we are using the same pattern to freeze in some other instances, we can make it a prop of the method `takeScreenshot`.


e2e/utils/pages/homepage.page.ts
```diff
@@ -359,7 +360,16 @@ abstract class Homepage {
...
 	}
 
-	async takeScreenshot({ centeredElementTestId }: TakeScreenshotParams = {}): Promise<void> {
+	async takeScreenshot(
+		{ freezeCarousel = false, centeredElementTestId }: TakeScreenshotParams = {
+			freezeCarousel: false
+		}
+	): Promise<void> {
+		if (freezeCarousel) {
+			await this.setCarouselFirstSlide();
```
bitdivine, line 369: Can the carousel rotate while waiting for load state?

bitdivine (approved):

e2e/utils/pages/homepage.page.ts
AntonioVentilii: no, it does not rotate: `setCarouselFirstSlide` really sets it to a single slide and "stop" the sliding. At least that is how we have used it until now

PR #4922, opened 2025-02-26
fix(frontend): Navigation consistency

daviddecentage:
# Motivation

On settings page, the main page links disappear which is a problem on mobile, since the main navigation to the left disappears. Theres then no way for a mobile user to navigate away from the settings page except clicking on the logo, which might not be clear for all users.

# Changes

Removed conditions that hide the links if youre on the settings page.

Also changed the order of Airdrops menu item and explore dapps menu item so it is the same as on the desktop only navigation to the left.

# Tests
image

### DIFF
--- a/src/frontend/src/lib/components/core/Menu.svelte
+++ b/src/frontend/src/lib/components/core/Menu.svelte
@@ -129,14 +129,14 @@
 			<MenuAddresses on:icMenuClick={hidePopover} />
 		{/if}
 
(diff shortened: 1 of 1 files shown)


AntonioVentilii (approved):
Ideally this should be split into 2 PRs, but in the interest of time, approved

BonomoAlessandro (approved):
LGTM

PR #4925, opened 2025-02-26
refactor(frontend): extract filter for snapshot transactions

AntonioVentilii:
# Motivation

To avoid repetition of code, it makes sense to concentrate filtering and slicing of the transactions used in the snapshots.

# Changes

- Create internal function to remove nullish transaction and slice the array.
- Apply its usage to the mapper `toIcrcTransaction` and `toSplTransaction`.

# Tests

Current tests are sufficient.

### DIFF
--- a/src/frontend/src/lib/services/user-snapshot.services.ts
+++ b/src/frontend/src/lib/services/user-snapshot.services.ts
@@ -66,6 +66,10 @@ interface ToSnapshotParams<T extends Token> {
 
 const LAST_TRANSACTIONS_COUNT = 5;
 
+const filterTransactions = <T extends Transaction_Icrc | Transaction_Spl>(
+	transactions: (T | undefined)[]
+) => transactions.filter(nonNullish).slice(0, LAST_TRANSACTIONS_COUNT);
+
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/services/user-snapshot.services.ts
```diff
@@ -66,6 +66,10 @@ interface ToSnapshotParams<T extends Token> {
 
 const LAST_TRANSACTIONS_COUNT = 5;
 
+const filterTransactions = <T extends Transaction_Icrc | Transaction_Spl>(
+	transactions: (T | undefined)[]
+) => transactions.filter(nonNullish).slice(0, LAST_TRANSACTIONS_COUNT);
```
bitdivine: Most recent transaction is at the front, I assume?

bitdivine (approved):

src/frontend/src/lib/services/user-snapshot.services.ts
AntonioVentilii: correct, at least that is the usual logic: whenever new transaction(s) appear, we prepend them

PR #4931, opened 2025-02-27
fix(frontend): missing SPL token owner address in worker message data

AntonioVentilii:
# Motivation

When I included the token owner address to the SPL tokens, I forgot to add it in the POST messages sent by the worker.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4933, opened 2025-02-27
fix(frontend): pass directly token owner address in the Solana transaction scroll

AntonioVentilii:
# Motivation

We should pass the token owner address to the `SolTransactionScroll` too.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4935, opened 2025-02-27
feat(frontend): fix reapply of max value

BonomoAlessandro:
# Motivation

In the `ETH send form` the fees change every few seconds. If the user clicks the `max` button the amount to send is also updated every time. This automatic update of the send amount after clicking the `max` button should stop as soon as the user enters a new amount.

# Tests

**before:**

https://github.com/user-attachments/assets/837f9f76-da81-4c55-ab04-83a437c01ffc

**after:**

https://github.com/user-attachments/assets/9bd09c37-7212-4e05-b44b-76462f54212b


AntonioVentilii (approved):
LGTM tks

PR #4937, opened 2025-02-27
chore(ci): adjust canister for deployment when triggered by workflow_run

AntonioVentilii:
# Motivation

Just to be more explicit, we adjust the canister of the CI that deploys to beta when there is a `workflow_run` event.


inc-man (approved):
LGTM, tks

PR #4943, opened 2025-02-27
feat(frontend): calculate source ATA address instead of fetching

AntonioVentilii:
# Motivation

As suggestion by PRODSEC, we will not fetch the ATA address of an user's owner address. We will instead calculate it with the function created in https://github.com/dfinity/oisy-wallet/pull/4942 .


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4948, opened 2025-02-27
feat(frontend): shows usd values on send review

BonomoAlessandro:
# Motivation

On the review page of the send process, the USD values corresponding to the amount and balance should also be displayed.

# Changes

- displays USD values

# Tests

**before:**
image

**after:**
image


src/frontend/src/lib/components/exchange/ExchangeAmountDisplay.svelte
```diff
@@ -32,14 +32,16 @@
...
-		<div class="text-tertiary">
-			{#if usdAmount < EXCHANGE_USD_AMOUNT_THRESHOLD}
-				{`( < ${formatUSD({
-					value: EXCHANGE_USD_AMOUNT_THRESHOLD
-				})} )`}
-			{:else}
-				{`( ${formatUSD({ value: usdAmount })} )`}
-			{/if}
-		</div>
+		{#if nonNullish(exchangeRate)}
```
AntonioVentilii, line 35: should you maybe define `let usdAmount: number | undefined;` and use `{#if nonNullish(usdAmount)}` ?

Because I see that the `nullish` check for `usdAmount` are a bit more. Plus do we want to show zero if there is an exchange rate but no amount?

AntonioVentilii (approved):
LGTM tks! One small observation

src/frontend/src/lib/components/exchange/ExchangeAmountDisplay.svelte
BonomoAlessandro: @AntonioVentilii You are right, i just changed it

PR #4950, opened 2025-02-27
refactor(frontend): rename Ethereum Transaction UI type

AntonioVentilii:
# Motivation

We are trying to standardize the UI transactions that we have. In order to do so, one of rhte first step is to have the same props, at least by name. The interface `EthTransactionUi` had a prop `uiType` that should in fact be simply `type`.

# Changes

- Adjust interface `EthTransactionUi` to ignore `Transaction.type`, remove `uiType`, and use `type` prop directly.
- Adapt the usages.

# Tests

Adapt existing tests.


DenysKarmazynDFINITY (approved):
👍

PR #4951, opened 2025-02-27
feat(frontend): calculate ATA address only if not provided in destination input

AntonioVentilii:
# Motivation

It may happen that the user provides an destination address that is already an ATA address and not the owner one. In that case, we should check and using directly the destination input.

# Changes

- Create method to fetch account owner via Solana RPC.
- Create util to check if a Solana address is an ATA address: we basically check if it has an owner.
- Use the util when creating an SPL transaction message.

# Tests

I adjusted the tests to be more comprehensive of all the cases and to check more calls of functions.


DenysKarmazynDFINITY (approved):
👍

PR #4952, opened 2025-02-27
fix(frontend): use Token owner address to fetch SPL balances

AntonioVentilii:
# Motivation

We should start to fetch the SPL token balances using the program address ID too.

# Changes

- Included `tokenOwnerAddress` as params for `loadSplTokenBalance`.
- Adapt the usage.

# Tests

Current tests are adpated.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4956, opened 2025-02-27
fix(frontend): Close menu popover when navigating to external links

daviddecentage:
# Motivation

When clicking on Documentation, Support, Sourcecode or Changelog links in the menu to the top right, the popover stays open. This feels weird when closing the external window and going back to Oisy as the popover is still openend and the majority of the page consists of backdrop.

# Changes

Added optional onClick props with a call to hide the popover that gets passed to the external link component.

# Tests

Demo of the menu being closed after navigating away
ezgif-11df160c71f5f9


AntonioVentilii (changes requested):
Typically, we don't use `onClick` events with hyperlinks (it may cause some unexpected rush behaviour). My suggestion is to use the `onNavigate` clause directly in the `Menu` component, or something similar

`onNavigate(hidePopover)`

I remember in the past we had a similar issue/task.

daviddecentage:
> Typically, we don't use `onClick` events with hyperlinks (it may cause some unexpected rush behaviour). My suggestion is to use the `onNavigate` clause directly in the `Menu` component, or something similar
> 
> `onNavigate(hidePopover)`
> 
> I remember in the past we had a similar issue/task.

Yeah not too happy with this solution either. However, onNavigate won't work:

```
**onNavigate**
A lifecycle function that runs the supplied callback immediately before we navigate to a new URL **except during full-page navigations**.
```
To confirm I tested it doesnt get called. Other ideas?

AntonioVentilii:
What if we put the on:click in the <Popover> ?

`<Popover bind:visible anchor={button} direction="rtl"  on:click={hidePopover}>`

What are the cases where this is an issue? Or the cases where it does not work properly? For example, it is ok that it closes every time that the user clicks anywhere in it? I tested for a case where the menu has a scroll bar and it does not seems to trigger this edge case, luckily, but there could be more cases

daviddecentage:
> What if we put the on:click in the ?
> 
> `<Popover bind:visible anchor={button} direction="rtl" on:click={hidePopover}>`
> 
> What are the cases where this is an issue? Or the cases where it does not work properly? For example, it is ok that it closes every time that the user clicks anywhere in it? I tested for a case where the menu has a scroll bar and it does not seems to trigger this edge case, luckily, but there could be more cases

Doesnt close for copying wallet addresses either, but I guess we should be careful with that.

Couldnt find any case where it shouldnt close... For now I implemented it as suggested by you

AntonioVentilii (approved):

PR #4963, opened 2025-02-27
feat(frontend): remove unnecessary token owner address in function

AntonioVentilii:
# Motivation

When I was adding the token owner address to the services, I overshoot: there is no need to pass it in `loadSplTokenBalance`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4965, opened 2025-02-28
feat(frontend): enables send form usd value

BonomoAlessandro:
# Motivation

It should be possible to choose between token amount and USD amount while entering the balance to send.

# Tests

https://github.com/user-attachments/assets/49ee83e3-62a0-4be1-8d2d-645208093e14


src/frontend/src/lib/components/tokens/TokenInputAmountExchange.svelte
```diff
@@ -39,7 +39,7 @@
 		<button
 			class:hover:cursor-default={disabled}
 			{disabled}
-			on:click={handleUnitSwitch}
+			on:click|preventDefault={handleUnitSwitch}
```
AntonioVentilii, line 42: is this one needed?

BonomoAlessandro: @AntonioVentilii Yes, otherwise if the form is filled out and you press the `swap currency` button you will get directed to the review page.

PR #4969, opened 2025-02-28
feat(frontend): add assertion of destination ATA address equal to calculated one

AntonioVentilii:
# Motivation

To add an additional layer of security, we make sure that the destination ATA address (if provided) is the same as the calculated one. This should help avoid any possible injection of destination address.

# Changes

- We already have the two values required.
- Add assertion between destination ATA and the calculated one.

# Tests

Included tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4971, opened 2025-02-28
feat(backend): add Antonio's principal to the allowed callers

AntonioVentilii:
# !!! IMPORTANT !!!

This change allows a specific user (in this case @AntonioVentilii ) to access some data and endpoint of the OISY backend. Specifically all the endpoints with guard `caller_is_allowed`, and among them the `update` endpoints are:
- `bulk_up`
- `migrate_user_data_to`
- `migration_stop_timer`
- `set_guards`
- `step_migration`
- `top_up_cycles_ledger`

# Motivation

As recent requirement, we would like to quickly check the user statistics of OISY, such as endpoint `stats`. The quickest way is to allow some user. In this case, we will allow me.

# Suggestion

We should have a different list of allowed users (or a different guard) for the less sensitive endpoints, that are useful to the OISY devs team.

### DIFF
--- a/scripts/build.backend.args.ic.did
+++ b/scripts/build.backend.args.ic.did
@@ -1,7 +1,7 @@
 (variant {
     Init = record {
          ecdsa_key_name = "key_1";
-         allowed_callers = vec{ principal "nynz6-haaaa-aaaan-qzqda-cai" };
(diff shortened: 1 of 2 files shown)


BonomoAlessandro (approved):
LGTM

lmuntaner (approved):
LGTM. This could be removed after the information has been retrieved.

It's probably more a privacy issue than a security issue.

StefanBerger-DFINITY:
Also approved by Product

PR #4973, opened 2025-02-28
feat(backend): add list of partially allowed users

AntonioVentilii:
# Motivation

As recent requirement, we would like to check the user statistics of OISY, such as `query` endpoint `stats`.
However for now only controllers (that is none of OISY team) and a list of allowed principal (for example the Rewards canister) can use such endpoints, because they have the `caller_is_allowed` guard.
The same guard applies to some relevant `update` endpoints:

- `bulk_up`
- `migrate_user_data_to`
- `migration_stop_timer`
- `set_guards`
- `step_migration`
- `top_up_cycles_ledger`

So, instead of changing the guard or [adding some OISY team's principal in the allowed list](https://github.com/dfinity/oisy-wallet/pull/4971), we create a new list that has slightly less limitation than the normal users. For now, such list will be allowed to call the `stats` endpoint.

We keep the rest of the permissions the same.

# Changes

- Add `limited_callers` to the canister configs, as array of principals.
- Add the same to the initial arguments: both interface and building arguments files. For now it contains only my (@AntonioVentilii) principal.
- At the same time, we keep the old structure of initial argument for the `Update` tests.
- Create new guard `caller_is_limited_or_allowed` that permits both `allowed_callers` and `limited_callers`.
- Apply the guard to the `stats` endpoint.

# Tests

Provided tests for the `stats` endpoint.


src/shared/src/impls.rs
```diff
@@ -62,6 +62,7 @@ impl From<InitArg> for Config {
         let InitArg {
             ecdsa_key_name,
             allowed_callers,
+            limited_callers,
```
AntonioVentilii, line 65: other naming suggestions:

- partially_allowed_callers
- permitted_callers
- safe_callers
- query_only_callers

AntonioVentilii:
This solution will not be implemented. We will work the statistics directly through the Rewards canister.

AntonioVentilii:
We are revisiting this solution temporary

AntonioVentilii:
Closing it again since `ALLOWED_USERS` is enough

PR #4974, opened 2025-02-28
feat(fix): Fixed review convert alignement issue

daviddecentage:
# Motivation

Analogous to PR #4615, and pointed out by @DenysKarmazynDFINITY in https://github.com/dfinity/oisy-wallet/pull/4615#issuecomment-2639592138, we adjust the convert UI by not utilizing the `outer-content` slot

# Changes

Removed div that was intended to be placed in outer-content slot, effectively placing it in the default slot.

# Tests

image


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4975, opened 2025-02-28
fix(e2e): remove flakiness due to focus

AntonioVentilii:
# Motivation

Some E2E tests are still flaky because the screenshot taken may include the focus of certain elements. However sometimes it includes it and sometimes not. We tried to identify the cause and it seems that the real consistency is really given by an asynchronous movement of the pointer in playwright:

- https://github.com/dfinity/oisy-wallet/pull/4977 --> force change of focus --> NOT WORKING
- https://github.com/dfinity/oisy-wallet/pull/4978 --> blur active element --> NOT WORKING
- https://github.com/dfinity/oisy-wallet/pull/4979 --> move mouse to point zero --> WORKING
- https://github.com/dfinity/oisy-wallet/pull/4980 --> disable animations in screenshot --> NOT WORKING

So, we force the movement of the mouse to point zero before screenshot.

# Changes

- Add `mouse.move(0, 0)`
- Revert function included to prevent this in previous PRs, that are not working.

# Tests

Run tests on this PR and it seems the flakiness is gone.


inc-man (approved):
LGTM, tks

PR #4981, opened 2025-02-28
refactor(frontend): use variant for TokenUiOrGroupUi

AntonioVentilii:
# Motivation

As per suggestion https://github.com/dfinity/oisy-wallet/pull/3742/files#r1856983637 , it is cleaner to use a variant for the type `TokenUiOrGroupUi` so that it is `{ token: TokenUi } | { group: TokenUiGroup }`. That avoids syntax like `{@const token = tokenGroup.tokens[0]}`.

# Changes

- Change type of `TokenUiOrGroupUi` for a variant of either a token or a group.
- Adapt component `TokenList`.
- Adapt all the grouping tokens utils.

# Tests

Test adapted but below there are some screenshots that show no changes.

Screenshot 2025-03-01 at 01 04 32
Screenshot 2025-03-01 at 01 04 41
Screenshot 2025-03-01 at 01 04 53
Screenshot 2025-03-01 at 01 05 39


peterpeterparker (approved):
Cool. LGTM, thx

PR #4982, opened 2025-03-01
refactor(frontend): remove redundant BTC tokens store

AntonioVentilii:
# Motivation

Since all Bitcoin tokens are enabled by default, we treat them as the other native tokens (for example Ethereum and Solana), and we use directly the `enabledBitcoinTokens` derived store.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #4999, opened 2025-03-03
chore(devops): Create another test frontend & backend

bitdivine:
# Motivation
Trailofbits needs a test frontend & paired backend.  The only full test pair we have is fe-1-{be/fe}, and both canisters are used quite a lot.  The simplest solution is to spin up an additional pair.

# Changes

- Add `test-trailofbits` as a network
- Create canisters for the frontend & backend for the trailofbits network
- Point other canisters on that network to the same values as staging.

# Tests
## Canisters have cycles
- Canisters have been added to cycleops.  Search for "trailofbits" here: https://cycleops.dev/app/team/oisy-wallet-developers/canisters/

## Canisters are controllable by the team
- The trailofbits frontend & backend canisters are controlled by the staging Orbit (Orbit station ID yo6ep-dyaaa-aaaal-ajs6q-cai)

## Deployments work
Deployed and login works: https://audit.oisy.com


AntonioVentilii (approved):
LGTM tks!

PR #5003, opened 2025-03-03
refactor(e2e): change snapshotting of modals

inc-man:
# Motivation

Currently we have on single takeSnapshot function that we should use in all the possible e2e cases. In that manner we change the snapshotting method when it comes to modals.

# Changes

Adjust the current snapshotting function and apply it to where only modals are snapshotted.

# Tests

CI runs on this PR


AntonioVentilii (approved):
LGTM tks!

PR #5004, opened 2025-03-03
feat(frontend): increase exchange load interval to 5 minutes

AntonioVentilii:
# Motivation

To reduce the number of calls to CoinGecko, we increase the interval of requesting exchange rates to 5 minutes. It is not so necessary for the user to have a very frequent rate.


inc-man (approved):
LGTM, tks

PR #5005, opened 2025-03-03
style(frontend): Change link colors in dark mode

daviddecentage:
# Motivation

Normal links have too little contrast in dark mode with dark blue against the dark background. In order to solve this Dan proposed using `text-brand-primary-alt` for the link colors. Since the current colors for `primary` and `primary-alt` are the same in light mode (where contrast is good) while in dark the `primary-alt` is slightly brighter this means we can just change the link classes without affecting light mode.

# Changes

Changed global link colors as well as special occurrences where colors were defined in svelte files.

# Tests

Before:
image
image
image
image

After:
image
image
image
image


DenysKarmazynDFINITY (approved):
LGTM, thanks.

After merging, let's double-check the e2e PR with the screenshots.

PR #5007, opened 2025-03-03
refactor(docker): Faster frontend docker builds

bitdivine:
# Motivation
Installing web dependencies is a time-consuming step in the frontend docker build.   Exactly how long depends on network speed and CPU but a typical number is 3 minutes, if installing from scratch in Docker.  The list of packages to install appears in the dockerfile after importing the code and setting the `DFX_NETWORK`, even though downloading the dependencies depends only on the dependency list.  This makes running docker builds for multiple networks, as needed when testing correctness of docker builds, unnecessarily slow.

# Changes
- Add a step early in the dockerfile that runs `npm ci` with just the `package*files`.

Note: I have left all other code in the Dockerfile exactly as it is, even though it looks as if further efficiencies are possible, e.g. skipping the second `npm ci`, using `--no-audit` and copying in only the files needed for the front-end build. This is to minimise the risk of anything in the build breaking or the output changing, for any reason.  Just the change in this PR is already a huge win.

# Tests
### Building all networks should be faster
```
build_all() { jq -r '.networks|keys[]' dfx.json | while read n ; do DOCKER_BUILDKIT=1 docker build -f Dockerfile.frontend --progress=plain --build-arg network="$n" -o target/ . || echo FAILED $n ; done ; }
```
With this PR:
```
$ time build_all
...
real	8m44,518s
user	0m3,503s
sys	0m3,846s
```
Without:
```
real	37m35,211s
user	0m4,910s
sys	0m5,543s
```

### Build artefacts should be unchanged
This should give the same output:
```
$ rm -fr target/frontend
$ DOCKER_BUILDKIT=1 docker build -f Dockerfile.frontend --progress=plain --build-arg network=ic -o target/ . && tar c ./target/frontend/ | sha256sum
```

With this PR:
```
ac4b6baba1771ed021747d38eac83824b7b77bf3f4c940571a368569ad502b37  -
```
Without:
```
ac4b6baba1771ed021747d38eac83824b7b77bf3f4c940571a368569ad502b37  -
```


AntonioVentilii (approved):
LGTM TKS!

Any chance of having some scripts for testing? Or non necessary?

PR #5008, opened 2025-03-03
fix(frontend): Svelte routing trailing slash never

daviddecentage:
# Motivation

Currently its defined that svelte always forces trailing slashes to routes. This was never a problem it seems but this leads to unexpected behaviours when navigating without passing trailing slashes on mobile (or mobile emulation through chrome dev tools!).

Example: 
Click nav `Assets` (`/`), path in browser is `oisy.com`
Click nav `Rewards` (`/rewards`), path in browser is now `oisy.com/rewards/` as the trailing slash is forced
When now clicking on `Rewards` again, the app recognizes it as a link to a different page because all routes are defined without trailing slash.

Note: to reproduce, select responsive mode in chrome dev tools!

For some reason, most browsers dont have a problem with that and the behavior is as expected. Some mobile devices however recognize this change in the routing and some routing mechanisms such as checking which route you are on (to check which menu item is highlighted for example) return inconsistent values or it just redirects to root route (`/`).

# Changes

- Set trailingSlashes to `never` so no trailing slashes are forced
- Adjust how link for token pages are build (removed slash being added)

Since all routes are defined without trailing slash in `routes.constants.ts`, this change would also be in line and consistent with never having/forcing trailing slashes.

### DIFF
--- a/src/frontend/src/lib/utils/nav.utils.ts
+++ b/src/frontend/src/lib/utils/nav.utils.ts
@@ -46,7 +46,7 @@ const tokenUrl = ({
 	token: Token;
 	path: AppPath.Transactions | undefined;
 }): string =>
-	`${path ?? ''}/?${TOKEN_PARAM}=${encodeURIComponent(
+	`${path ?? ''}?${TOKEN_PARAM}=${encodeURIComponent(
 		name.replace(/\p{Emoji}/gu, (m, _idx) => `\\u${m.codePointAt(0)?.toString(16)}`)
 	)}${nonNullish(networkId.description) ? `&${networkParam(networkId)}` : ''}`;
 
--- a/src/frontend/src/routes/+layout.ts
+++ b/src/frontend/src/routes/+layout.ts
@@ -1,6 +1,6 @@
 export const prerender = true;
 export const ssr = false;
-export const trailingSlash = 'always';
+export const trailingSlash = 'never';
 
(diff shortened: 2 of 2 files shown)


peterpeterparker:
Just in case it's interesting and unknown:

1. If you proceed with this PR, I advise clearing/resetting the frontend canister; otherwise, you might end up with conflicts in production between old and new routes.

2. An assets canister is not a web server; therefore, any changes you make here must be verified with a deployed frontend (not with a dev server or a local web server).

4. If the issue appears to be resolved, you should ensure that the new behavior does not actually break an existing functional behavior.

daviddecentage:
> Just in case it's interesting and unknown:
> 
> 1. If you proceed with this PR, I advise clearing/resetting the frontend canister; otherwise, you might end up with conflicts in production between old and new routes.
> 2. An assets canister is not a web server; therefore, any changes you make here must be verified with a deployed frontend (not with a dev server or a local web server).
> 3. If the issue appears to be resolved, you should ensure that the new behavior does not actually break an existing functional behavior.

I did a little rethinking on this topic. Now I think this change would probably be too risky and/or time consuming for the benefit it provides.

I am going to open a new PR that cleans up the routes only by explicitly defining the trailing slash. This will end up making the whole routing setup consistent while not changing any current routes.

peterpeterparker:
Side note for future readers: We discussed this offline, and it should work out, tests we promising. However, it required some more testing and a probable cleanup to avoid keeping stale HTML resources that point to outdated JS files when deploying. Hence, the decisions above.

PR #5009, opened 2025-03-03
chore(devops): Simplify signer deployment

bitdivine:
# Motivation
The signer is deployed with a script when `dfx deploy signer` will suffice.  Indeed, that is all the script does.

# Changes
- Delete the deploy script & call `dfx deploy` instead.

# Tests
CI should suffice, as this is called in the e2e tests.


inc-man (approved):
LGTM, tks

AntonioVentilii (approved):

PR #5012, opened 2025-03-03
fix(frontend): issues with ETH fees on send/convert

DenysKarmazynDFINITY:
# Motivation

This PR fixes a few related things:
1. ckERC20 tokens conversion forms (both send and review) are now properly calculating USD values using the ETH exchange rate. 
Screenshot 2025-03-03 at 17 38 24

2. Same issue fixed for the ERC20 tokens send flow.
Screenshot 2025-03-03 at 17 40 58

3. Added back the animation of the ETH fee value.

### DIFF
--- a/src/frontend/src/icp-eth/components/fee/FeeAmountDisplay.svelte
+++ b/src/frontend/src/icp-eth/components/fee/FeeAmountDisplay.svelte
@@ -1,14 +1,12 @@
 <script lang="ts">
 	import { debounce, nonNullish } from '@dfinity/utils';
 	import { BigNumber } from '@ethersproject/bignumber';
-	import { getContext } from 'svelte';
-	import { slide } from 'svelte/transition';
+	import { slide, fade } from 'svelte/transition';
 	import ExchangeAmountDisplay from '$lib/components/exchange/ExchangeAmountDisplay.svelte';
 	import { ZERO } from '$lib/constants/app.constants';
 	import { SLIDE_DURATION } from '$lib/constants/transition.constants';
 	import { balancesStore } from '$lib/stores/balances.store';
 	import { i18n } from '$lib/stores/i18n.store';
-	import { SEND_CONTEXT_KEY, type SendContext } from '$lib/stores/send.store';
 	import type { OptionBalance } from '$lib/types/balance';
 	import type { TokenId } from '$lib/types/token';
 	import { formatToken } from '$lib/utils/format.utils';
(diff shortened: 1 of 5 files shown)


src/frontend/src/icp-eth/components/fee/FeeAmountDisplay.svelte
```diff
@@ -33,12 +30,15 @@
...
 </script>
 
-<ExchangeAmountDisplay
-	amount={fee}
-	decimals={feeDecimals}
-	symbol={feeSymbol}
-	exchangeRate={$sendTokenExchangeRate}
-/>
+<div transition:fade>
+	<ExchangeAmountDisplay
```
AntonioVentilii, line 34: I see that inside `ExchangeAmountDisplay` there is already a `fade` effect, is it not going in conflict? Or it refers to another `IF` condition?

AntonioVentilii, line 34: I see that inside `ExchangeAmountDisplay` there is already a `fade` effect, is it not going in conflict? Or it refers to another `IF` condition?

src/frontend/src/icp/components/fee/EthereumEstimatedFee.svelte
```diff
@@ -23,6 +24,9 @@
 
 	let maxTransactionFee: Option<bigint> = undefined;
 	$: maxTransactionFee = $store?.maxTransactionFee;
+
+	let ethFeeExchangeRate: number | undefined;
+	$: ethFeeExchangeRate = $exchanges?.[feeToken.id]?.usd;
```
AntonioVentilii, line 29: wouldn't this be in the context store as `feeExchangeRateStore` ?

DenysKarmazynDFINITY: This component is a child of `IcSendTokenWizard` which does not use the required for that store context. The current structure is: `FeeContext` (with the store you mentioned) is for `EthSendTokenWizard`, `EthereumFeeContext` - for `IcSendTokenWizard`.

src/frontend/src/icp-eth/components/fee/FeeAmountDisplay.svelte
DenysKarmazynDFINITY: You're right - instead of adding additional transition, I checked why the one in `ExchangeAmountDisplay` does not work. It appeared to be because of several layers of `if`s that wrap this component. To solve this, instead of the default "local" transition, I made it "global", which means it is also triggered when any block containing the transitions is added or removed.

AntonioVentilii (commented):
LGTM tks

AntonioVentilii (approved):
LGTM tks!

PR #5013, opened 2025-03-03
feat(backend): endpoint to fetch list of creation timestamps

AntonioVentilii:
# Motivation

New endpoint `list_user_creation_timestamps` for OISY backend to provide a list of creation timestamps of the OISY user profiles.

# Changes

- Created new function `oisy_user_creation_timestamps` to list the creation timestamps from the OISY users map. This is very similar to what the endpoint `list_users` does through `oisy_users` function.
- Created endpoint `list_user_creation_timestamps` in the same structure of `list_users`.

# Tests

Included new tests.


src/backend/src/lib.rs
```diff
@@ -623,6 +626,21 @@ pub fn list_users(request: ListUsersRequest) -> ListUsersResponse {
     }
 }
 
+#[query(guard = "caller_is_allowed")]
+#[allow(clippy::needless_pass_by_value)]
+#[must_use]
+pub fn list_user_creation_timestamps(
```
bitdivine, line 632: Iterating over all users in stable memory will fail when the number of users is large enough.  However the number of users might still be small enough.  So let's just try it,  and if it fails we'll switch to a paginated endpoint.

src/backend/tests/it/utils/assertion.rs
```diff
@@ -49,3 +49,22 @@ pub fn assert_user_profiles_eq(results_users: Vec<OisyUser>, expected_users: Vec
         );
     }
 }
+
+pub fn assert_user_creation_timestamps_eq(
+    results_timestamps: Vec<Timestamp>,
+    expected_timestamps: Vec<Timestamp>,
+) {
+    assert_eq!(
```
bitdivine, line 57: This is fine and good, but be aware that there are "pretty assertions" that often do a  good job at making assertion failures readable: https://docs.rs/pretty_assertions/latest/pretty_assertions/

bitdivine (approved):
LGTM, thanks!

src/backend/tests/it/utils/assertion.rs
AntonioVentilii: Ah no, I didn't! Thank you for that! Changed

src/backend/src/lib.rs
AntonioVentilii: Yes, I thought so, it is not really scalable, right?

PR #5019, opened 2025-03-04
fix(frontend): Consistent routes adjustment to always include trailing slash matching trailingSlash="always" setting

daviddecentage:
# Motivation

This PR fixes a very specific and rare glitch (described in #5008). In order to keep the changes to a minimum, we change the routes by adding trailing slashes. This is then in line with svelte routings setting `trailingSlash="always"`, as this setting was conflicting with our route definitions used by anchor link tags href attributes (always defined without trailing slash).

This ultimately lead to issues on some mobile devices and responsive mode in chrome devtools.

# Changes

- Added trailing slashes to all route definitions
- Adjusted some conditions to keep UI routing related behavior:
- - Added `normalizePath` function to nav utils to always add trailing slashes to a URL if it doesnt have it already
- - Used that `normalizePath` function to normalize the paths returned by the $page objects (which return paths without trailing slash)

# Tests

ezgif-54dcc761e91350


AntonioVentilii (commented):
May you adapt the current tests to verify that they work without trailing slash?

For example

```typescript
		describe('isRouteActivity', () => {
			it('should return true when route id matches Activity path', () => {
				const mockPath = `${ROUTE_ID_GROUP_APP}${AppPath.Activity}`;

				expect(isRouteActivity(mockPage(mockPath))).toBe(true);
				expect(isRouteActivity(mockPage(`${mockPath.slice(0, -1)}`))).toBe(true);
			});
			...
```

AntonioVentilii (approved):
LGTM tks!

PR #5021, opened 2025-03-04
chore(frontend): remove unnecessary params from toCkMinterInfoAddresses

AntonioVentilii:
# Motivation

After PR https://github.com/dfinity/oisy-wallet/pull/5020 , the prop `networkId` is not really used in function `toCkMinterInfoAddresses`.

PR #5023, opened 2025-03-04
feat(frontend): add ck minter info guard in registerUserSnapshot

AntonioVentilii:
# Motivation

As mentioned by @peterpeterparker  in https://github.com/dfinity/oisy-wallet/pull/5021#discussion_r1979278043 , we are using the `ckEthMinterInfoStore` imperatively in `registerUserSnapshot` , it is better to add a guard for nullish values. It is ok to continue with the method, since it is just required to send snapshots of all the accounts.


src/frontend/src/lib/services/user-snapshot.services.ts
```diff
@@ -208,10 +208,16 @@ const toSplSnapshot = ({
 		minterInfo: get(ckEthMinterInfoStore)?.[ETHEREUM_TOKEN_ID],
 		networkId: ETHEREUM_NETWORK_ID
 	});
+	if (isNullish(ckEthMinterInfoAddressesMainnet)) {
+		return;
```
peterpeterparker, line 212: We do not display an error, just returning is fine? If it's fine, do we understand it by the context or does it needs a comment to improve readability?

AntonioVentilii: I included a comment there

PR #5026, opened 2025-03-04
fix(frontend): avoid using "global" transition

DenysKarmazynDFINITY:
# Motivation

We got an unexpected behaviour of the Send modal after adding `global` Svelte transition to the ETH fee component. It appeared that the `if` statement is not needed, which simplifies everything.


AntonioVentilii (approved):
LGTM tks!

peterpeterparker:
@DenysKarmazynDFINITY Did you double-check that the component is properly destroyed?

In Gix-CMP and the NNS dapp, and maybe here, we added the |global keyword to the transition to fix components that remain mounted.

PR #5031, opened 2025-03-04
style(frontend): Fix minor styling issue on an error alert

daviddecentage:
# Motivation

On Send Review the error alert displays if insufficient funds for fees displays with wrong text color.

# Changes

Removed span with wrong text color and let the text inherit MessageBox text class

# Tests

Before:
image

After:
image


DenysKarmazynDFINITY (approved):
👍

PR #5032, opened 2025-03-04
refactor(frontend): Variable height Dapp cards

daviddecentage:
# Motivation

The Dapp cards currently are fixed height cards. This can become a problem once other languages are introduced or a text changes to a longer one.

# Changes

To avoid this, we can simple remove some classes and let the content of the card grow the card. Additionally, equal height rows ensure that all cards are the same height per row, which seems already in place.

# Tests


AntonioVentilii (approved):
LGTM tks!

PR #5033, opened 2025-03-04
feat(frontend): remove unnecessary token address from Solana closeAccount mapping

AntonioVentilii:
# Motivation

It is not really necessary to fetch the token address for a `closeAccount` transaction: the ATA address will be closed (most of the times), and in general it is suggested to use directly the Solana native token for this type of transactions, since it concerns purely SOL transfers.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5036, opened 2025-03-05
feat(frontend): use schema and brand for generic Solana address

AntonioVentilii:
# Motivation

Ideally we want to start differentiating between type of addresses, but mostly not only in terms of how the string is composed, but really branding them. For example differentiate when it is a generic Solana address, an ATA address or an SPL address. Their structure is identical, but it is useful to know exactly which one we are using.

So, as first step, we create a brand for a generic Solana address.

# Changes

- Use `zod` for `Address` type.
- Use `zod` with brand for `SolAddress` type.
- Create parsing util for `SolAddress`.

# Tests

No changes in local replica.


src/frontend/src/tests/lib/derived/tokens.derived.spec.ts
```diff
@@ -106,6 +108,16 @@ describe('tokens.derived', () => {
...
 	};
 
+	const mockSplCustomToken: SplCustomToken = {
+		...mockValidSplToken,
+		id: parseTokenId('SplCustomTokenId2'),
+		symbol: 'SplCustomTokenId2',
+		address: parseSolAddress(`${mockValidSplToken.address}2`),
+		version: undefined,
+		enabled: true
+	};
```
AntonioVentilii:
```suggestion
	

```

AntonioVentilii:
this idea requires a lot of not-so-necessary refactoring: the cons outweigh the pros for now

PR #5037, opened 2025-03-05
refactor(frontend): use deconstruction of token data in AddTokenByNetwork

AntonioVentilii:
# Motivation

Small refactoring that it will come in handy when we will brand the type of addresses. Note that we remove the fallback to empty string that it was not really necessary


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5044, opened 2025-03-05
feat(frontend): reverse Solana transactions only after mapping

AntonioVentilii:
# Motivation

As pointed out by PRODSEC, we should reverse the list of mapped transactions ONLY after the mapping, to avoid mixing the indexes that match the inner instructions.


DenysKarmazynDFINITY (approved):
Small question/suggestion, otherwise LGTM

PR #5046, opened 2025-03-06
refactor(devops): Simplify ckETH deployment

bitdivine:
# Motivation
The ckETH ledger has a post-deployment step, moving come ckETH to a principal.  This can be put into a standard post-deploy hook, however it is not obvious why this step is needed at all.  Cannot the ledger be initialized with tokens in that account already?  Let's see.

# Changes
- Initialize the ckETH ledger with tokens for the x4.. user instead of transferring tokens after deployment.

# Tests
### Existing tests, including e2e tests, should work.
See CI


inc-man (commented):
LGTM, but e2e tests did not run on the PR it seems. Maybe for safety reasons we could add the update snapshot label? Or is it overkill?

inc-man (approved):
LGTM, tks

PR #5055, opened 2025-03-06
fix(frontend): show ATA addresses as destination and sources

AntonioVentilii:
# Motivation

Under PRODSEC suggestion: we will show the original addresses of a Solana transaction instead of their owner when we are parsing transactions


DenysKarmazynDFINITY (approved):
LGTM, thanks

Can you please rename the PR to feat/fix before merging? Just for the protocol 🙂

PR #5058, opened 2025-03-06
refactor(frontend): put Sol inner instructions at the respective index

AntonioVentilii:
# Motivation

Just to be more correct, we put the inner instructions right after the instruction they refer to: this will allow to have a proper cumulative balance during the parsing.

### DIFF
--- a/src/frontend/src/sol/services/sol-transactions.services.ts
+++ b/src/frontend/src/sol/services/sol-transactions.services.ts
@@ -58,13 +58,12 @@ export const fetchSolTransactionsForSignature = async ({
 	// Inside the instructions there could be some that we are unable to decode, but that may have
 	// simpler (and decoded) inner instructions. We should try to map those as well.
 	// They are inserted in the instructions array in the order they refer to the main instruction.
-	const allInstructions = putativeInnerInstructions.reduce(
-		(acc, { index, instructions }, offset) => {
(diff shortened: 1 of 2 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/sol/services/sol-transactions.services.ts
```diff
@@ -57,10 +57,13 @@ export const fetchSolTransactionsForSignature = async ({
 
 	// Inside the instructions there could be some that we are unable to decode, but that may have
 	// simpler (and decoded) inner instructions. We should try to map those as well.
-	const allInstructions = [
-		...instructions,
-		...putativeInnerInstructions.flatMap(({ instructions }) => instructions)
-	];
+	// They are inserted in the instructions array in the order they refer to the main instruction.
+	const allInstructions = [...putativeInnerInstructions]
+		.sort((a, b) => a.index - b.index)
```
DenysKarmazynDFINITY, line 62: Just to confirm - by adding `sort` by index here, we don't change the original order, do we? I'm just looking at the comment in the code which says that it should be "in the order they refer to the main instruction".

AntonioVentilii: yes yes, it does not change, it comes by default. It is just a good to have, since we just use the `index` prop. They could be mixed

PR #5059, opened 2025-03-06
refactor(frontend): make object for internal reduce of fetchSolTransactionsForSignature

AntonioVentilii:
# Motivation

We refactor the `reduce` function inside `fetchSolTransactionsForSignature` to return an object, so that we can prepare for further calculation: we will calculate the cumulative balance in another PR, to pass it to the mapping and calculate the `closeAccount` value (from PR https://github.com/dfinity/oisy-wallet/pull/5056).

# Changes

Instead of returning an array of parsed transactions `SolTransactionUi`, the reduce function returns an object that is `{ parsedTransactions: SolTransactionUi[] }`.

Nothing changed in the implementation of the functions.

# Tests

Current tests are enough.

### DIFF
--- a/src/frontend/src/sol/services/sol-transactions.services.ts
+++ b/src/frontend/src/sol/services/sol-transactions.services.ts
@@ -74,68 +74,79 @@ export const fetchSolTransactionsForSignature = async ({
 				})
 			: [undefined];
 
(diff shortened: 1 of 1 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5060, opened 2025-03-06
feat(frontend): reduce the number of calls of the Solana wallet

AntonioVentilii:
# Motivation

We pay the calls of the Solana RPC provider, and they are proportional to the number. It is becoming too unscalable.

So, as a temporary measure, we delay the calls of the Solana wallet scheduler to one minute (double the other wallets).

# Future improvement

This solution should be removed when we have a better one. For example, if we have a Solana RPC canister (similar to the EVM one), or when we change the code to not re-load the same transactions twice but guarantee that the data are correct.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5062, opened 2025-03-06
feat(frontend): calculate cumulative balances when parsing Solana transactions

AntonioVentilii:
# Motivation

The redeemed SOL value of a Solana `closeAccount` transaction is not explicitly indicated in the transaction data. So it must be inferred/calculated from the latest SOL balance of the ATA that is being closed.

After PR https://github.com/dfinity/oisy-wallet/pull/5056 , we can provide the accumulated balances to the mapper.

To calculate them we keep a record in the `reduce` function: every instruction inside a generic transaction will affect the cumulative SOL balance of each address.

It is guaranteed to be progressive since the RPC returns the list of instructions in the correct internal chronological order. 

The redeemed SOL value is the latest total SOL balance of the closed Associated Token Account.

NOTE: We include WSOL in the calculation, because it is used to affect the SOL balance of the ATA.

# Changes

# Tests

Apart from expanding the current tests, we verified with a [practical transaction](https://solscan.io/tx/inspector?txhash=4E88TD8dpeivGbAn83DQcfQYyBcPF869Q9Nw9xfRvbZMDB46EMwC8FEqidJ5PoCZRSZVrBJJz486Ncju7duD3kwn) comparing our results with Solscan.

Screenshot 2025-03-06 at 22 40 33
Screenshot 2025-03-06 at 22 41 56

### DIFF
--- a/src/frontend/src/sol/services/sol-transactions.services.ts
+++ b/src/frontend/src/sol/services/sol-transactions.services.ts
@@ -1,3 +1,4 @@
+import { WSOL_TOKEN } from '$env/tokens/tokens-spl/tokens.wsol.env';
 import {
 	SOLANA_DEVNET_TOKEN_ID,
 	SOLANA_LOCAL_TOKEN_ID,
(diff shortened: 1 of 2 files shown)


src/frontend/src/sol/services/sol-transactions.services.ts
```diff
@@ -107,23 +114,36 @@ export const fetchSolTransactionsForSignature = async ({
...
 			const { value, from, to, tokenAddress: mappedTokenAddress } = mappedTransaction;
 
+			// The cumulative balances are updated for every instruction, so we can keep track of the
+			// SOL balance of the address and its associated token account at any given time.
+			// It is useful when mapping for example a `closeAccount` instruction, where the redeemed value
+			// is not provided in the data and must be calculated as the latest total SOL balance of the Associated Token Account.
+			const cumulativeBalances = {
+				...accCumulativeBalances,
+				// We include WSOL in the calculation, because it is used to affect the SOL balance of the ATA.
+				...((isNullish(mappedTokenAddress) || mappedTokenAddress === WSOL_TOKEN.address) && {
```
DenysKarmazynDFINITY, line 124: Is it fine overwriting the previous values here?

DenysKarmazynDFINITY (approved):
Small q, otherwise LGTM

src/frontend/src/sol/services/sol-transactions.services.ts
AntonioVentilii: indeed! the scope is to update the balances, so yes

PR #5063, opened 2025-03-06
fix(e2e): wait for Activity page skeletons to disappear

AntonioVentilii:
# Motivation

After PR https://github.com/dfinity/oisy-wallet/pull/3493 , we need to adapt the E2E tests for the activity page: they should wait for the skeleton to disappear.


inc-man (commented):
Looks good in general, but the latest screenshots are still not "unified" right? One shows the skeleton the other one shows "Your transaction history will appear here". Shouldn't it all be the same if its the same page?

DenysKarmazynDFINITY (approved):
I think Nico's comment is now resolved (all screenshots are unified and display the empty state message), therefore approving

PR #5066, opened 2025-03-06
feat(frontend): map Solana transaction createAccount directly

AntonioVentilii:
# Motivation

We want to map directly the createAccount transaction from the Solana RPC.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5070, opened 2025-03-07
feat(frontend): New tag on rewards main nav item

daviddecentage:
# Motivation

We need to display a "New" tag on the rewards main navigation item

# Changes

- Added "New" tag to main nav item
- Added "New" translation item

# Tests

 Before:
image

After:
image
image


DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/lib/components/navigation/NavigationMenuMainItems.svelte
```diff
@@ -89,6 +89,10 @@
 	>
 		<IconGift />
 		{$i18n.navigation.text.airdrops}
+		<div
+			class="h-5.5 mt-0.25 text-xs/4.5 ml-1 rounded-md bg-error-primary px-2 py-0.5 font-bold text-white"
+			>{$i18n.core.text.new}</div
+		>
```
AntonioVentilii, line 95: The component `Tag` is not useful in this case?

AntonioVentilii: To be clear: it is not a review, just a question, that maybe we can even explore in another PR

daviddecentage: Certainly could use it, but it would have to be reworked a bit. Might do just that after in another PR

PR #5073, opened 2025-03-07
feat(e2e): wait correct URL when using navigateTo

AntonioVentilii:
# Motivation

Additional check in the E2E tests to verify that we navigated to the correct page.


BonomoAlessandro (approved):
LGTM

PR #5076, opened 2025-03-07
feat(frontend): user-amount utils

DenysKarmazynDFINITY:
# Motivation

In this PR, the user-amount utils are created. They are gonna replace the existing `convert.utils`. `validateUserAmount` is implemented the way that it can be used in swap, send and convert flow. 

In the next PR, I will remove the old file and update SwapForm and ConvertAmountSource to start using this method.

# Changes

New util is created.

# Tests

Provided.

### DIFF
--- a/src/frontend/src/lib/utils/user-amount.utils.ts
+++ b/src/frontend/src/lib/utils/user-amount.utils.ts
@@ -18,9 +18,9 @@ import {
 	assertErc20Amount
 } from '$lib/utils/assert-amount.utils';
 import { formatToken } from '$lib/utils/format.utils';
+import { parseToken } from '$lib/utils/parse.utils';
 import { nonNullish } from '@dfinity/utils';
 import { BigNumber } from '@ethersproject/bignumber';
-import { Utils } from 'alchemy-sdk';
 
 export const validateUserAmount = ({
 	userAmount,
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/utils/user-amount.utils.ts
```diff
@@ -0,0 +1,102 @@
...
+	balance?: BigNumber;
+	balanceForFee?: BigNumber;
+	fee?: bigint;
+	ethereumEstimateFee?: bigint;
+	minterInfo?: Option<CkBtcMinterInfoData | CkEthMinterInfoData>;
+	isSwapFlow?: boolean;
+}): ConvertAmountErrorType => {
+	// We should align balance and userAmount to avoid issues caused by comparing formatted and unformatted BN
+	const parsedSendBalance = nonNullish(balance)
+		? Utils.parseUnits(
```
AntonioVentilii, line 46: i think `parseToken` can be used here?

src/frontend/src/lib/utils/user-amount.utils.ts
```diff
@@ -0,0 +1,102 @@
...
+					unitName: token.decimals,
+					displayDecimals: token.decimals
+				}),
+				token.decimals
+			)
+		: ZERO;
+
+	// if the function called in the swap flow, we only need to check the basic assertAmount condition
+	// if convert or send - we identify token type and check some network-specific conditions
+	if (!isSwapFlow) {
```
AntonioVentilii, line 58: just for readability, i would invert the condition

```
if (isSwapFlow) {
return assertAmount({
		userAmount,
		balance: parsedSendBalance,
		fee
	});
	}
```
	
	and then the rest

AntonioVentilii (approved):
LGTM tks! Two small quesitons

src/frontend/src/lib/utils/user-amount.utils.ts
DenysKarmazynDFINITY: true, updated

src/frontend/src/lib/utils/user-amount.utils.ts
DenysKarmazynDFINITY: I tried to avoid using the `assertAmount` 2 times, but I agree it's more readable the way you're suggesting. Updated.

PR #5078, opened 2025-03-07
feat(frontend): Airdrops requirements component

daviddecentage:
# Motivation

We need to change how the requirments for airdrops are displayed and make it dynamic.

For this we prepare a dynamic requirements component that displays different states depending on what is passed to it.

# Changes

- Added new component
- Added translated texts

# Tests

image
image


src/frontend/src/lib/components/airdrops/AirdropsRequirements.svelte
```diff
@@ -0,0 +1,57 @@
...
+	export let requirementsFulfilled: boolean[];
+</script>
+
+{#if airdrop.requirements.length > 0}
+	{#if badgeOnly}
+		{#if isEligible}<span class="inline-flex pl-3"
+				><Badge variant="success">{$i18n.airdrops.text.youre_eligible}</Badge></span
+			>{/if}
+	{:else}
+		<Hr spacing="md" />
```
AntonioVentilii, line 20: maybe this should go in the parent component that will use this one, no?

src/frontend/src/lib/components/airdrops/AirdropsRequirements.svelte
```diff
@@ -0,0 +1,57 @@
...
+
+		<span class="text-md font-semibold"
+			>{$i18n.airdrops.text.requirements_title}
+		</span>{#if isEligible}<span class="inline-flex pl-3"
+				><Badge variant="success">{$i18n.airdrops.text.youre_eligible}</Badge></span
+			>{/if}
+		<ul class="list-none">
+			{#each airdrop.requirements as requirement, i}
+				<li class="flex gap-2 pt-1">
+					{#if requirementsFulfilled[i]}
```
AntonioVentilii, line 30: We should find a way to guarantee that there is a requirementsFulfilled value for each requirement. Better to treat this in a function, returning an object with `{requirement: Requirement, fulfilled:boolean}` or something similar

it can be done here or in the consumer, whichever. ideally in the consumer, so that we reduce the number of props here

src/frontend/src/lib/components/airdrops/AirdropsRequirements.svelte
```diff
@@ -0,0 +1,57 @@
...
+							class:transition={loading}
+							class:duration-500={loading}
+							class:ease-in-out={loading}
+							class:animate-pulse={loading}
+						>
+							<span class="-mt-0.5 mr-2 text-disabled"><IconCheckCircleFill size={32} /></span>
+							<span>
+								{requirement}
+							</span>
+						</span>
```
AntonioVentilii, line 51: if the condition just affects the color and the conditional classes, we can remove the `if`, no?


```svelte
<li class="flex gap-2 pt-1">
  <span
    class="flex w-full flex-row"
    class:transition={!fulfilled && loading}
    class:duration-500={!fulfilled && loading}
    class:ease-in-out={!fulfilled && loading}
    class:animate-pulse={!fulfilled && loading}
  >
    <span
      class="-mt-0.5 mr-2"
      class:text-success-primary={fulfilled}
      class:text-disabled={!fulfilled}
    >
      <IconCheckCircleFill size={32} />
    </span>
    <span>
      {requirement}
    </span>
  </span>
</li>
```

src/frontend/src/lib/components/airdrops/AirdropsRequirements.svelte
```diff
@@ -0,0 +1,57 @@
...
+					{/if}
+				</li>
+			{/each}
+		</ul>
+	{/if}
+{/if}
```
AntonioVentilii, line 57: maybe some tests for this component?

src/frontend/src/lib/components/airdrops/AirdropsRequirements.svelte
```diff
@@ -0,0 +1,52 @@
...
+	import Badge from '$lib/components/ui/Badge.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+
+	export let loading = true;
+	export let airdrop: AirdropDescription;
+	export let badgeOnly = false;
+	export let isEligible = false;
+	export let requirementsFulfilled: boolean[];
+
+	const isRequirementFulfilled = (index: number) => requirementsFulfilled[index] ?? false;
```
AntonioVentilii, line 13: this should be reactive, no?

AntonioVentilii (approved):

AntonioVentilii (changes requested):
I mistakenly approved.... sorry

src/frontend/src/lib/components/airdrops/AirdropsRequirements.svelte
BonomoAlessandro: No, it is not guaranteed that the order of the two arrays will be the same. However, at the moment, it is hardcoded, so it will always match. In the near future, all this information will come from the rewards service, and at that point, the order will be guaranteed.

src/frontend/src/lib/components/airdrops/AirdropsRequirements.svelte
BonomoAlessandro: @AntonioVentilii I just provided some test cases

AntonioVentilii (approved):
LGTM tks!

PR #5069, opened 2025-03-07
feat(frontend): Airdrops earnings components

daviddecentage:
# Motivation

For the airdrops refactoring, we need to display 3 cards displaying the earnings, as well as some other text and design changes. Here we focus on the cards and displaying the earnings.

# Changes

- Added new components to display earnings and earnings cards
- Added translations
- Added confetti bg image

# Tests

image
image

PR #5082, opened 2025-03-07
feat(frontend): USD amount utils function

daviddecentage:
# Motivation

For upcoming airdrops refactoring and according to @AntonioVentilii suggestion https://github.com/dfinity/oisy-wallet/pull/5069#discussion_r1985005747.

# Changes

- Added utils function `calculateTokenUsdAmount`
- Added unit test to verify that it works

# Tests

token.utils.spec.ts:
- should correctly calculate USD amount for the token and amount
- should return 0 if exchange rate is not available
- should return 0 if token is not available


src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -130,6 +130,32 @@ export const calculateTokenUsdBalance = ({
...
+ */
+export const calculateTokenUsdAmount = ({
+	amount,
+	token,
+	$exchanges
+}: {
+	amount: BigNumber;
+	token?: IcToken;
+	$exchanges: ExchangesData;
+}) => {
```
AntonioVentilii, line 149: return type missing

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -130,6 +130,32 @@ export const calculateTokenUsdBalance = ({
...
+	if (isNullish(amount) || isNullish(token)) {
+		return 0;
```
AntonioVentilii, line 151: i think you should return undefined here and let the caller treat it

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -130,6 +130,32 @@ export const calculateTokenUsdBalance = ({
...
+	}
+	const exchangeRate = $exchanges?.[token.id]?.usd;
+	return nonNullish(exchangeRate)
+		? usdValue({ decimals: token.decimals, balance: amount, exchangeRate })
+		: 0;
```
AntonioVentilii, line 156: same here

src/frontend/src/tests/lib/utils/token.utils.spec.ts
```diff
@@ -165,6 +166,45 @@ describe('calculateTokenUsdBalance', () => {
...
+
+		mockUsdValue.mockImplementation(
+			({ balance, exchangeRate }) => Number(balance ?? 0) * exchangeRate
+		);
+	});
+
+	it('should correctly calculate USD amount for the token and amount', () => {
+		const result = calculateTokenUsdAmount({
+			token: ICP_TOKEN,
+			amount: BigNumber.from(3),
```
AntonioVentilii, line 183:
```suggestion
			amount: bn3,
```

src/frontend/src/tests/lib/utils/token.utils.spec.ts
```diff
@@ -165,6 +166,45 @@ describe('calculateTokenUsdBalance', () => {
...
+			$exchanges: mockExchanges
+		});
+		expect(result).toEqual(bn3.toNumber());
+	});
+
+	it('should return 0 if exchange rate is not available', () => {
+		const result = calculateTokenUsdAmount({
+			token: ICP_TOKEN,
+			amount: BigNumber.from(3),
```
AntonioVentilii, line 192:
```suggestion
			amount: bn3,
```

src/frontend/src/tests/lib/utils/token.utils.spec.ts
```diff
@@ -165,6 +166,45 @@ describe('calculateTokenUsdBalance', () => {
...
+			$exchanges: {}
+		});
+		expect(result).toEqual(0);
+	});
+
+	it('should return 0 if token is not available', () => {
+		const result = calculateTokenUsdAmount({
+			token: undefined,
+			amount: BigNumber.from(3),
```
AntonioVentilii, line 201:
```suggestion
			amount: bn3,
```

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -119,15 +119,9 @@ export const calculateTokenUsdBalance = ({
...
+	const amount = $balances?.[token.id]?.data;
 
-	return nonNullish(exchangeRate)
-		? usdValue({
-				decimals: token.decimals,
-				balance: $balances?.[token.id]?.data,
-				exchangeRate
-			})
-		: undefined;
+	return calculateTokenUsdAmount({ amount, $exchanges, token });
```
AntonioVentilii, line 124: maybe all inlined?

`return calculateTokenUsdAmount({ amount:$balances?.[token.id]?.data, $exchanges, token });`

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -119,14 +119,34 @@ export const calculateTokenUsdBalance = ({
...
+ * @returns The USD balance or Number(0) in case the number cannot be calculated.
+ *
+ */
+export const calculateTokenUsdAmount = ({
+	amount,
+	token,
+	$exchanges
+}: {
+	amount?: BigNumber;
+	token?: Token;
```
AntonioVentilii, line 141: are there instances where this function is called with `token` nullish?

daviddecentage: Not yet but later token can be undefined if it hasnt loaded yet.

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -119,14 +119,34 @@ export const calculateTokenUsdBalance = ({
...
+	$exchanges: ExchangesData;
+}) => {
+	if (isNullish(amount) || isNullish(token)) {
+		return undefined;
+	}
```
AntonioVentilii, line 145: i would remove this condition, so that the behaviour is kept as it should: meaning that nullish `amount` provides `zero`

src/frontend/src/lib/utils/token.utils.ts
AntonioVentilii: then, this function should not be called. Maybe the condition should go in the consumer

src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -118,15 +118,30 @@ export const calculateTokenUsdBalance = ({
...
+ * @param $exchanges - The exchange rates data for the tokens.
+ * @returns The USD balance or Number(0) in case the number cannot be calculated.
+ *
+ */
+export const calculateTokenUsdAmount = ({
+	amount,
+	token,
+	$exchanges
+}: {
+	amount?: BigNumber;
```
AntonioVentilii, line 138:
```suggestion
	amount: Option<BigNumber>;
```

AntonioVentilii (approved):
LGTM tks, one last suggestion

PR #5083, opened 2025-03-07
feat(frontend): IcTokenFees component

DenysKarmazynDFINITY:
# Motivation

In this PR, I'm adding a new component for displaying all IC-tokens related fees. It will be used in both Convert and Send flows. After we updated the Send flow, we will remove the old fee components (for now, I added TODO comments there).

Screenshot 2025-03-07 at 14 48 08


AntonioVentilii (approved):
LGTM tks

PR #5085, opened 2025-03-07
feat(frontend): append can duplicate and not in transactions store

AntonioVentilii:
# Motivation

By default, the `append` method of a transaction store should remove the duplicated IDs (one never know). However the IC transactions store was behaving ok with (possibly) duplicating the transactions. So we keep that behaviour too, creating method `appendDuplicating`.

The changed `append` method will be useful for the Solana transactions store: mapping signatures to transactions in Solana is quite non-trivial. So, to add a further layer of "insurance", we guarantee that there are no duplicates transactions in the list, since the ID is unique.

# Changes

- Create sub-function to update the transaction list with a prop to remove duplicates or not.
- Change `append` method to remove duplicated ID.
- Create method `appendDuplicating` identical to append but with allowing duplicates.

# Tests

Current tests are adapted.

### DIFF
--- a/src/frontend/src/icp/services/ic-transactions.services.ts
+++ b/src/frontend/src/icp/services/ic-transactions.services.ts
@@ -65,7 +65,7 @@ export const loadNextTransactions = ({
 				return;
 			}
 
-			icTransactionsStore.append({
+			icTransactionsStore.appendDuplicating({
 				tokenId: token.id,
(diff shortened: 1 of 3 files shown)


AntonioVentilii:
Superseded by:

- https://github.com/dfinity/oisy-wallet/pull/5134

PR #5087, opened 2025-03-07
fix(backend): Document & fix top up return value

bitdivine:
# Motivation
The top up function lacks documentation and the return value is incorrect when there was no top-up.

# Changes
- Add documentation
- When the ledger balance is already sufficient, the return value was incorrect; this is fixed.

# Tests


AntonioVentilii (approved):
LGTM tks!

PR #5088, opened 2025-03-07
feat(frontend): create common type for transaction ID

AntonioVentilii:
# Motivation

We want to unify the transactions types, as much as possible. One of the first step is to have the same type of ID, in this case, a `string`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5098, opened 2025-03-10
feat(frontend): displays the usd value of the btc send fee

BonomoAlessandro:
# Motivation

In the BTC send form, the corresponding USD value of the fee is not displayed but should be.

# Changes

- displays usd value of BTC fee

# Tests

**before:**
image

**after:**
image


AntonioVentilii (approved):
LGTM tks!

PR #5101, opened 2025-03-10
fix(e2e): added wait for carousel fade out functionality

inc-man:
# Motivation

With recent changes to the carousel we also implemented a fade out effect when swapping pages. For the e2e tests that sometimes resulted in a screenshot where the carousel is still showing when it shouldnt.

# Changes

With this PR we implement a functionality for the relevant tests, so that we always wait for the carousel to completely fade out.

# Tests

CI runs on this PR


AntonioVentilii (approved):
LGTM tks! 

Nitpick: shall we create a function for this? so that we can reuse it? like `this.waitForCarouselToBeHidden`, boh

PR #5108, opened 2025-03-10
feat(frontend): IcConvertProgress component

DenysKarmazynDFINITY:
# Motivation

This PR include the new `IcConvertProgress` component. It is very similar to `IcSendProgress` at the moment, but may receive some changes later, therefore I kept them separate but re-used the main functionality.

### DIFF
--- a/src/frontend/src/icp/constants/steps.constants.ts
+++ b/src/frontend/src/icp/constants/steps.constants.ts
@@ -1,0 +1,50 @@
+import { isNetworkIdETH } from '$icp/utils/ic-send.utils';
+import { ProgressStepsSendIc } from '$lib/enums/progress-steps';
+import type { NetworkId } from '$lib/types/network';
+import type { ProgressSteps } from '$lib/types/progress-steps';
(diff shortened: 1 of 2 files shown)


src/frontend/src/icp/constants/steps.constants.ts
```diff
@@ -0,0 +1,50 @@
...
+		step: ProgressStepsSendIc.SEND,
+		text: i18n.send.text.sending,
+		state: 'next'
+	} as ProgressStep,
+	{
+		step: ProgressStepsSendIc.RELOAD,
+		text: i18n.send.text.refreshing_ui,
+		state: 'next'
+	} as ProgressStep
+];
```
AntonioVentilii, line 50: since we are here, does it work the same if you remove the cast for each one and put here `] as ProgressSteps;`

AntonioVentilii (approved):
LGTM tks!

Small observation below

src/frontend/src/icp/constants/steps.constants.ts
DenysKarmazynDFINITY: yes, indeed, updated!

PR #5115, opened 2025-03-10
feat(frontend): updates gitbook link

BonomoAlessandro:
# Motivation

The gitbook structure has slightly changed, which means the url needs to be updated.


AntonioVentilii (approved):
LGTM tks!

PR #5117, opened 2025-03-10
feat(frontend): IcConvertForm component

DenysKarmazynDFINITY:
# Motivation

This PR includes the new `IcConvertForm` component. Also, I realized `IcConvertProgress` can also use context to retrieve `networkId`, so I put this change here, too.

Screenshot 2025-03-10 at 14 05 41


src/frontend/src/icp/components/convert/IcConvertForm.svelte
```diff
@@ -0,0 +1,124 @@
...
+<ConvertForm
+	on:icNext
+	bind:sendAmount
+	bind:receiveAmount
+	bind:insufficientFunds
+	bind:insufficientFundsForFee
+	bind:amountLessThanLedgerFee
+	bind:minimumAmountNotReached
+	bind:unknownMinimumAmount
+	bind:minterInfoNotCertified
```
AntonioVentilii, line 99: not for this PR, but maybe there is a way to include some of these values in the context? or at least the amount analysis, boh, some sort of value that identifies if there is no funds, no fee enough, etc

AntonioVentilii (approved):
LGTM tks

src/frontend/src/icp/components/convert/IcConvertForm.svelte
DenysKarmazynDFINITY: totally, the plan is to create a context for these errors, it will be later re-used in all conversion/send components

PR #5120, opened 2025-03-10
refactor(frontend): Update airdrops page/card

daviddecentage:
# Motivation

This is a PR for updated airdrops card design.

# Changes

Updated airdrops card

# Tests

Old:
image

New:
image


src/frontend/src/lib/components/airdrops/AirdropCard.svelte
```diff
@@ -22,20 +24,26 @@
...
-			<span class="flex">
-				<AirdropDateBadge
-					date={airdrop.endDate}
-					testId={nonNullish(testId) ? `${testId}-badge` : undefined}
-				/></span
-			>
-		</section>
+			<div>
+				<Button colorStyle="primary" styleClass="py-2" paddingSmall
+					>{$i18n.airdrops.text.check_earnings}</Button
```
AntonioVentilii, line 44: is this button missing the call to action `onClick`? or did i miss something?

daviddecentage: The on:click is on the whole card

AntonioVentilii: Then this one should not be a button. Better another type, because an enabled button without a click action is a problem

daviddecentage: Fixed

PR #5121, opened 2025-03-10
feat(frontend): enables autofocus on send amount field

BonomoAlessandro:
# Motivation

When the send form is opened, the `Amount` field should be autofocused so that the user can start entering the amount directly.

# Tests

**before:**
image

**after:**
image


AntonioVentilii (approved):
LGTM tks!

PR #5131, opened 2025-03-10
build(scripts): include the script folder for the aliases

AntonioVentilii:
# Motivation

It is useful that the IDE recognize the aliases for the Typescript scripts. So we include those scripts in the tsconfig for the aliases.


DenysKarmazynDFINITY (approved):
👍

PR #5133, opened 2025-03-10
refactor(devops): Remove unnecessary arguments

bitdivine:
# Motivation
The arguments to `dfx deploy cketh_ledger` match the defaults so there is no point in providing them.

# Changes
- Remove `cketh_ledger` arguments.

# Tests
### Arguments match the defaults
See the comments in the PR.


scripts/deploy.cketh.sh
```diff
@@ -30,28 +30,7 @@ dfx deploy cketh_minter --specified-id jzenf-aiaaa-aaaar-qaa7q-cai --network "$D
 })"
 
 echo "Step 3: deploy ledger canister..."
-PRINCIPAL="$(dfx identity get-principal)"
-dfx deploy cketh_ledger --specified-id apia6-jaaaa-aaaar-qabma-cai --network "$DFX_NETWORK" --argument "(variant {
```
bitdivine, line 34: --specified-id is the same as that provided in dfx.json

--argument is the same as that created by the build command

AntonioVentilii (approved):

PR #5134, opened 2025-03-10
feat(frontend): transaction store append method cannot duplicate transactions

AntonioVentilii:
# Motivation

Since we made the transaction IDs unique for self-transacitons with https://github.com/dfinity/oisy-wallet/pull/5124, there is no reason to have duplicated IDs in the transaction store.

This will help us normalize all transaction stores.

# Changes

- Change append method of transaction store to non duplicate transactions, based on the ID.

# Tests

Adapted tests.


BonomoAlessandro (approved):
LGTM, thx

PR #5142, opened 2025-03-10
build(frontend): use Beta canisters ID for rewards, signer and Kongswap

AntonioVentilii:
# Motivation

Since we set the definition in the `dfx.json` file, we use the Beta canisters IDs for the Rewards canister, the Signer and KongSwap.

It will not change much, since their value is the same as prod (that was the variable they were using until now).


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5143, opened 2025-03-10
build(frontend): load metadata of additional ICRC tokens

AntonioVentilii:
# Motivation

Ideally we would like to have the additional ICRC metadata already in the code base, to avoid fetching them at runtime, for each user. So, we duplicate what we do for the SNS tokens and the ckERC20 tokens, and fetch and save the metadata in a script.

The next step will be to use such data and remove the fetching in the code.

# Changes

- Create script to fetch and save metadata very similar to the existing ones for SNS and ckERC20.
- The script will use the list of ledger canister IDs (and index canister IDs) already in the JSON file.

# Tests

Ran the script in this same PR.

### DIFF
--- a/src/frontend/src/env/tokens/tokens.icrc.json
+++ b/src/frontend/src/env/tokens/tokens.icrc.json
@@ -1,82 +1,82 @@
 {
-  "BURN": {
-    "ledgerCanisterId": "egjwt-lqaaa-aaaak-qi2aa-cai",
-    "name": "MSQ Cycle Burn",
-    "indexCanisterId": "nrant-tyaaa-aaaag-atsjq-cai",
-    "decimals": 8,
-    "symbol": "BURN",
-    "fee": {
-      "__bigint__": "10000"
-    }
-  },
-  "POPEYE": {
-    "ledgerCanisterId": "6fvyi-faaaa-aaaam-qbiga-cai",
-    "name": "Popeye The Sailor",
-    "indexCanisterId": "gg3c3-6iaaa-aaaah-aq6dq-cai",
-    "decimals": 8,
-    "symbol": "POPEYE",
-    "fee": {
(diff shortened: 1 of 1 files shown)


scripts/build.tokens.icrc.ts
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	// Open the file for writing only if it does not already exist (wx flag).
+	// This avoids a potential file system race condition warning.
+	const fd = openSync(file, 'wx');
+
+	if (!logoData.includes(';') || !logoData.includes(',')) {
+		console.error(`Invalid logoData format for ${name}: ${logoData}`);
+		return;
+	}
```
AntonioVentilii, line 108:
```suggestion
	}
	
```

scripts/build.tokens.icrc.ts
```diff
@@ -0,0 +1,130 @@
...
+	return await Object.entries(icrcTokens).reduce<Promise<TokensAndIcons>>(
+		async (acc, [key, token]) => {
+			if (isNullish(token)) {
+				throw new Error(`Data is missing for token symbol ${key}.`);
+			}
+
+			const { ledgerCanisterId: savedLedgerCanisterId, indexCanisterId: savedIndexCanisterId } =
+				token;
+
+			if (isNullish(savedLedgerCanisterId) || isNullish(savedIndexCanisterId)) {
```
DenysKarmazynDFINITY, line 37: Maybe it's worth handling these 2 cases separately (hence having dedicated error message per check)? So it's instantly clear whether the former or latter is missing.

DenysKarmazynDFINITY (approved):
LGTM, thanks

scripts/build.tokens.icrc.ts
AntonioVentilii: done!

PR #5144, opened 2025-03-10
refactor(frontend): switch ledger and index canisters ID in ICRC JSON file

AntonioVentilii:
# Motivation

A very small refactor that will help us compare changes, since [we are going to load ICRC metadata](https://github.com/dfinity/oisy-wallet/pull/5143) and save them in the file. When we load them, the ledger canister ID comes always before the index canister ID.


DenysKarmazynDFINITY (approved):
👍

PR #5145, opened 2025-03-10
feat(frontend): show ATA address in Receive modal for SPL tokens

AntonioVentilii:
# Motivation

As PRODSEC suggested, it is more secure to let the user manage the direct address where they will receive the tokens. In the SPL case, it is the ATA address. So we display it in the Receive modal.

# Changes

- Start passing the `token` prop to `SolReceive` (like we do for IC and Ethereum).
- Calculate the ATA address in `SolReceive` when necessary (in case of SPL token).
- Provide the address to the modal.

# Tests

See below the different addresses:

Screenshot 2025-03-10 at 22 05 34
Screenshot 2025-03-10 at 22 05 45
Screenshot 2025-03-10 at 22 05 53


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5146, opened 2025-03-10
feat(frontend): pass last signature to Sol wallet worker

AntonioVentilii:
# Motivation

Each Solana signature requires quite a lot of calls to the RPC to be parsed. To avoid too many unnecessary calls, we pass the last signature fetched to the worker, so that it can keep loading from there.

# Changes

- Get last signature imperatively from the Solana transaction store.
- Pass it to the worker as data.

# Tests

Current tests are sufficient.


AntonioVentilii:
The PR is no longer useful: its goal (avoiding redundant Solana RPC calls) is already solved better by the backend transaction cache (`USER_TRANSACTIONS_LOAD_FROM_BACKEND_ENABLED`), and applying it as-is would actually break the polling worker since pinning `before` to the oldest known signature at init means the worker would never fetch newer transactions.

PR #5149, opened 2025-03-10
feat(frontend): track list of address and tokens in Solana transactions loader

AntonioVentilii:
# Motivation

After https://github.com/dfinity/oisy-wallet/pull/5148 , we can track the list of address and their respective SPL tokens addresses, and pass it to the mapper when we are inside the same signature.

# Changes

- Add a mapper to the `reduce` function: it will track each address to its respective SPL token address.
- Pass the map to the mapper function.

# Tests

Current tests adapted.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5150, opened 2025-03-10
feat(frontend): Solana service to fetch all signatures for address

AntonioVentilii:
# Motivation

We are fetching and parsing the Solana signatures for each Solana token (native or SPL). That is quite inefficient, since the same signature may be fetched twice and consequently parsed twice. It is not an issue, since we parse the instructions inside and returns transactions that are unique per token. So no duplicates in the transactions that are returned.

However we do an unnecessary amount of calls to the Solana RPC for unnecessary mapping of the same signatures.

To avoid this, we would ideally fetch all the signature for a specific user first, remove the duplicates, and then parse each one.

So, as first step, we create a service that fetches all the transaction of a specific user: we provide a list of tokens too, so that we fetch the signatures for the respective ATA addresses. Furthermore, the function will filter all duplicates signatures.

# Changes

- Add service `getSolSignatures`, as described above.

# Tests

Added tests for the new service.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5151, opened 2025-03-11
feat(frontend): fixes spacing

BonomoAlessandro:
# Motivation

The spaces between the different fields on the send form are inconsistent. Sometimes the space is larger and sometimes it is smaller but the spacing should always be the same.

# Changes

- adjust spacing

# Tests

**before:**
image

**after:**
image


AntonioVentilii (approved):
LGTM tks

PR #5155, opened 2025-03-11
feat(frontend): add fee in Solana transaction mappers

AntonioVentilii:
# Motivation

A SOL fee is not assigned to a single transfer/instruction in a Solana transaction/signature. It is for the entire signature. However, it is useful to have it as data when we map the transactions and when we show it in the UI.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5160, opened 2025-03-11
ci(devops): deploy on workflow_run only if successful

AntonioVentilii:
# Motivation

The deployment to Beta is triggered by workflows `'Tag on Merge from Release Branch'`, when it completes. However Github Actions consider it completed even if it is `skipped`. So we make the condition more restrictive so that it is triggered only when the conclusion is a `success`.

This should help us avoid deploying to Beta for each trigger by `'Tag on Merge from Release Branch'`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5167, opened 2025-03-11
fix(frontend): fixes padding

BonomoAlessandro:
# Motivation

The padding between the two input fields on the convert form is missing and should be added.

# Changes

- fixes padding

# Tests

**before:**
image

**after:**
image


AntonioVentilii (approved):
LGTM tks!

PR #5168, opened 2025-03-11
feat(frontend): enables to customize onClose of ManageTokens/ManageTokensModal

BonomoAlessandro:
# Motivation

Bug:
A user that has a token in the token list, but the token is disabled. Clicking on a direct link to this token (e.g https://fe3.oisy.com/transactions/?token=ckEURC&network=ICP) will lead to an infinite loading state.

Fix:
Instead of falling into an infinite loading state, the manage token modal should be displayed to give the user the option to manually enable the token.


AntonioVentilii (commented):
Do you already have an use case for it? Can it be added in this PR?

If the use case is bigger, let's first do the PRs that lead to have this prop, and ultimately when we had this prop, it is being used in the same PR

BonomoAlessandro:
@AntonioVentilii The use case for this change will following in this PR [#5156](https://github.com/dfinity/oisy-wallet/pull/5156)

src/frontend/src/lib/components/transactions/Transactions.svelte
```diff
@@ -1,14 +1,37 @@
...
+	import { FALLBACK_TIMEOUT } from '$lib/constants/app.constants';
+	import { allTokens } from '$lib/derived/all-tokens.derived';
 	import { routeNetwork, routeToken } from '$lib/derived/nav.derived';
 	import { networkBitcoin, networkICP, networkSolana } from '$lib/derived/network.derived';
+	import { pageToken } from '$lib/derived/page-token.derived';
+	import type { OptionToken } from '$lib/types/token';
 	import SolTransactions from '$sol/components/transactions/SolTransactions.svelte';
+
+	let token: OptionToken;
+	$: token = $allTokens.find((token) => token.name === $routeToken);
```
AntonioVentilii, line 17: you should check for `$routeNetwork` too

src/frontend/src/lib/components/transactions/Transactions.svelte
```diff
@@ -1,14 +1,37 @@
...
+	const handleClose = async () => {
+		if (isNullish($pageToken)) {
+			await goto('/');
+		}
+	};
 </script>
 
-{#if nonNullish($routeNetwork)}
+{#if showTokenModal && nonNullish(token)}
+	<ManageTokensModal onClose={handleClose} />
```
AntonioVentilii, line 33: are you sure it is ok to put this component here? from what i recall it uses the `token` store that it is not that easy manageable. For example, what happens when you have the `token` store already set (for whatever reason) and you open this modal in a non-enabled-token transaction page. Does the `token` store resets? Is it ok for the manage modal? Same question for `selectedNetwork`

src/frontend/src/lib/components/transactions/Transactions.svelte
```diff
@@ -1,14 +1,37 @@
 <script lang="ts">
```
AntonioVentilii, line 1: in general it is quite an important/breaking change, so for sure we need tests (unittest and/or E2E) before merging

src/frontend/src/lib/components/transactions/Transactions.svelte
```diff
@@ -1,14 +1,37 @@
...
 	import SolTransactions from '$sol/components/transactions/SolTransactions.svelte';
+
+	let token: OptionToken;
+	$: token = $allTokens.find((token) => token.name === $routeToken);
+	let showTokenModal = false;
+
+	$: if (isNullish($pageToken) && nonNullish($routeToken) && nonNullish(token)) {
+		setTimeout(() => {
+			showTokenModal = true;
+		}, FALLBACK_TIMEOUT);
```
AntonioVentilii, line 23: i am not really sure I understood this part: by default it is in loading state, so at some point, after the timeout, we open the manage tokens modal. We don't care if it may take more time to load (more that the `FALLBACK_TIMEOUT`), right? Just for my understanding. If so, can you please provide some comment/docstring here to explain it?

src/frontend/src/lib/components/transactions/Transactions.svelte
```diff
@@ -1,14 +1,37 @@
...
+	}
+	const handleClose = async () => {
+		if (isNullish($pageToken)) {
+			await goto('/');
+		}
```
AntonioVentilii, line 28: Scenario:

- it is taking too long to load all the tokens
- the manage modals opens
- it finally loads all the token and we find out that the relevant token is already enabled
- in the microinstant that the $pageToken is not set BUT the user sees that the token is enabled, the user closes the modal

What is the result? May you please provide a test just to track this? Something like "should ... if the token is enabled but $pageToken is nullish"

src/frontend/src/lib/components/transactions/Transactions.svelte
BonomoAlessandro: This should not be a problem. If for example, i navigate to the following URL `/transactions/?token=Bonk&network=SOL` and the `Bonk` token is not enabled, the modal opens and i can see all tokens from the `SOL` network. So there is active network filtering but the token itself does not cause any issues.

src/frontend/src/lib/components/transactions/Transactions.svelte
BonomoAlessandro: Yess exactly. We wait for the time defined in `FALLBACK_TIMEOUT` and if the data is not loaded, we display the modal. This is because we do not have a change to check if the data fetching is complete.

src/frontend/src/lib/components/transactions/Transactions.svelte
BonomoAlessandro: In this case what will happen is, that the modal closes but the user does not get redirected to the activity page.

BonomoAlessandro: In this case the modal closes and the user gets redirected to the activity page.

AntonioVentilii (approved):
A few observation, but LGTM

src/frontend/src/lib/components/transactions/Transactions.svelte
AntonioVentilii: As discussed offline, we can revisit this edge case in the future

src/frontend/src/lib/components/transactions/Transactions.svelte
```diff
@@ -1,14 +1,52 @@
...
+
+	let token: OptionToken;
+	$: token = $allTokens.find(
+		(token) =>
+			token.name === $routeToken &&
+			$routeNetwork &&
+			token.network.id.toString().includes($routeNetwork)
+	);
+
+	let showTokenModal = false;
```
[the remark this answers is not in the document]
BonomoAlessandro: Otherwise after enabling the token the modal would reopen for a fraction of a second. Like this i can prevent this behavior.

src/frontend/src/lib/components/transactions/Transactions.svelte
```diff
@@ -1,14 +1,52 @@
...
+
+	const handleClose = async () => {
+		if (isNullish($pageToken)) {
+			await goto('/');
+		}
+	};
 </script>
 
-{#if nonNullish($routeNetwork)}
+{#if $modalManageTokens && showTokenModal && nonNullish(token)}
```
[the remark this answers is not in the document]
BonomoAlessandro: same as above

src/frontend/src/lib/components/transactions/Transactions.svelte
AntonioVentilii: as discussed offline, all good, forget my comment

PR #5169, opened 2025-03-11
refactor(devops): simplify checks for deployment CI

AntonioVentilii:
# Motivation

Just to organize better the logic of deploying, we put the general check of the canister directly in the step that sets the deployment network.

This avoids doubling the code too.

### DIFF
--- a/.github/workflows/deploy-to-environment.yml
+++ b/.github/workflows/deploy-to-environment.yml
@@ -7,8 +7,8 @@ on:
     tags:
       - v*
   workflow_run:
-    workflows: ['Tag on Merge from Release Branch']
-    types: [completed]
+    workflows: [ 'Tag on Merge from Release Branch' ]
+    types: [ completed ]
   workflow_dispatch:
     inputs:
       network:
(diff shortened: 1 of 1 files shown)


bitdivine (approved):
Not 100% sure about workflow dispatch but if there is a problem we can change it.

.github/workflows/deploy-to-environment.yml
```diff
@@ -66,12 +69,19 @@ jobs:
               exit 1
             fi
             echo "CANISTER=frontend" >> $GITHUB_ENV
+            echo "deploy_frontend=true" >> $GITHUB_ENV
+            echo "deploy_backend=true" >> $GITHUB_ENV
           elif [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
             echo "NETWORK=${{ github.event.inputs.network }}" >> $GITHUB_ENV
             echo "CANISTER=${{ github.event.inputs.canister }}" >> $GITHUB_ENV
+            echo "deploy_frontend=${{ github.event.inputs.canister == 'frontend' }}" >> $GITHUB_ENV
+            echo "deploy_backend=${{ github.event.inputs.canister == 'backend' }}" >> $GITHUB_ENV
```
bitdivine, line 78:
```suggestion
            echo "deploy_${{ github.event.inputs.canister}}=true" >> $GITHUB_ENV
```

PR #5172, opened 2025-03-11
feat(frontend): fix direct link to unexisting token

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/routes/(app)/transactions/+page.svelte
+++ b/src/frontend/src/routes/(app)/transactions/+page.svelte
@@ -1,10 +1,14 @@
 <script lang="ts">
-	import { isNullish, notEmptyString } from '@dfinity/utils';
+    import {isNullish, notEmptyString} from '@dfinity/utils';
 	import { onMount } from 'svelte';
 	import { goto } from '$app/navigation';
 	import Transactions from '$lib/components/transactions/Transactions.svelte';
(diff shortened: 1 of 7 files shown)


src/frontend/src/routes/(app)/transactions/+page.svelte
```diff
@@ -16,6 +20,14 @@
 		// Therefore, we cannot automatically select the network if it is not provided when the component mounts, and we cannot wait indefinitely.
 		// That's why, if no network is provided, we route to the root.
 		if (isNullish(routeNetwork) || !notEmptyString(routeNetwork)) {
+			if (nonNullish($routeToken)) {
```
AntonioVentilii, line 23: here better to do the same as `routeNetwork`

`const routeToken = urlParams.get(TOKEN_PARAM);`

AntonioVentilii: and I am actually not sure this message is needed, because if I got to this link

`http://localhost:5173/transactions/?token=Internet%20Computer`

It shows the error message, but the ICP token is enabled. The issue is that the link is wrong

BonomoAlessandro: As discussed offline we do display this message but with a more generalized message.

AntonioVentilii (approved):

PR #5175, opened 2025-03-11
feat(frontend): remove the warning for non-ATA destination address

AntonioVentilii:
# Motivation

The PRODSEC review suggested that we warn/inform the user when the destination input for Solana is not an ATA address, since we will derive it in our code.

This information can happen either in the form or in the confirmation review, or both, or in whichever form it is acceptable to have the user aware.

However, we are going to revisit how to show this information further on. For the current release we temporarily remove it.

### DIFF
--- a/src/frontend/src/lib/i18n/en.json
+++ b/src/frontend/src/lib/i18n/en.json
@@ -1,988 +1,988 @@
 {
-  "core": {
-    "text": {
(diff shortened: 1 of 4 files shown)


DenysKarmazynDFINITY (approved):
There is one test suite to adjust, approving already though

PR #5179, opened 2025-03-11
fix(frontend): correct inner instructions merging in fetchSolTransactionsForSignature

AntonioVentilii:
# Motivation

In service `fetchSolTransactionsForSignature` , we are merging the inner instructions into the main instructions, to map them all, as we should. However, it is more correct to map them at the index right after the main instruction they refer to.

This merging was wrong until now. No real effect, but it gives us the chance to fix it and modify the tests to be more extensive.

# Changes

- Correct the merging of the inner instructions to use an offset, since the reduce function modifies the final array length, and it shall be dynamic to insert the inner instructions in the correct original index.
- Add a mock transaction that is, in fact, a real (big) transaction from our test account.

# Tests

Adapt the tests to be more extensive, more detailed, more readable, and to use a practical case.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5186, opened 2025-03-12
feat(frontend): use enabled Bitcoin networks instead of all of them

AntonioVentilii:
# Motivation

Since we have a flag to enable/disable Bitcoin mainnet, we should use the correct list of Bitcoin networks, in case the mainnet is not present.

### DIFF
--- a/src/frontend/src/env/networks/networks.env.ts
+++ b/src/frontend/src/env/networks/networks.env.ts
@@ -6,6 +6,7 @@ import {
 	ETHEREUM_EXPLORER_URL,
 	SEPOLIA_EXPLORER_URL
 } from '$env/explorers.env';
+import { BTC_MAINNET_ENABLED } from '$env/networks/networks.btc.env';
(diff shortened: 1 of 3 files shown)


.env.test
```diff
@@ -1,4 +1,3 @@
-VITE_BITCOIN_MAINNET_DISABLED=true
```
DenysKarmazynDFINITY, line 1: Is there some particular reason why you decided to remove it? The flags for ETH and SOL (PR next to this one) are still there.

AntonioVentilii: ah sorry, not really sure why is there, maybe i was testing something. But now I am wondering, why is it disabled for the tests env? Do you have an idea?

DenysKarmazynDFINITY: Those env vars are used in the unit test environments. Since after the `VITE_BITCOIN_MAINNET_DISABLED ` removal your PR stayed green, it means it was not required. For instance, I tried to remove some other values from this file - some of the unit tests failed.

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5193, opened 2025-03-12
chore(frontend): bump axios in package-lock

AntonioVentilii:
# Motivation

We want to update `axios` since the current version raises [vulnerabilities](https://github.com/axios/axios/issues/6463).


AntonioVentilii:
A better solution is to wait for Alchemy to fix it on their side. I raised a ticket: https://github.com/alchemyplatform/alchemy-sdk-js/issues/486

So, for now, this PR is not necessary

PR #5195, opened 2025-03-12
chore(frontend): downgrade tailwind css to 4.0.11

AntonioVentilii:
# Motivation

There is an issue with `tailwindcss > 4.0.11` since it seems that it does not recognize the punctuation Dot in a custom screen breakpoint, causing for example the disappearance of the footer. I [raised a ticket](https://github.com/tailwindlabs/tailwindcss/issues/17148), but in the meanwhile we downgrade to `v4.0.11` that works

### Before

Screenshot 2025-03-12 at 16 29 20

### After

Screenshot 2025-03-12 at 16 27 57


peterpeterparker (approved):
LGTM, thx

Also had issue in oisy-wallet-signer with recent version of tailwind

https://github.com/dfinity/oisy-wallet-signer/issues/496

PR #5196, opened 2025-03-12
fix(frontend): Fix blurry token list

daviddecentage:
# Motivation

The token list on send, manage tokens and swap looks blurry sometimes. This was due to subpixel definitions in many places.

# Changes

To avoid using subpixels, were now rounding padding definitions, font sizes and line-height definitions.

# Tests
Running e2e tests


AntonioVentilii (approved):
TKs!!!

PR #5199, opened 2025-03-12
feat(frontend): show token name in transaction detail instead of address

AntonioVentilii:
# Motivation

When we show the transactions details of an Ethereum transaction, if `from` and/or `to` are token addresses, it would be better to show it, instead of its address. This is to avoid possible confusions on what that address refers too.

# Changes

- Create the `from` and `to` strings by searching among the known tokens.

# Tests

### Before

Screenshot 2025-03-12 at 20 15 10

### After

Screenshot 2025-03-12 at 20 12 13

### Rest is unchanged

Screenshot 2025-03-12 at 20 12 21
Screenshot 2025-03-12 at 20 12 32
Screenshot 2025-03-12 at 20 12 41


AntonioVentilii:
This PR is replaced by one that goes to a specific release-dedicated branch:

- PR #5205

PR #5200, opened 2025-03-12
feat(frontend): refresh ERC20 mined transaction details

AntonioVentilii:
# Motivation

There is an issue caused by a wrong data in the ERC20 transactions details: after the transaction is `mined`, we process it again from `pending`. However, we use an endpoint that returns the wrong source and/or destination: specifically, it shows the token address, that is the smart contract that was interacted with.

The endpoint we are using is [eth_getTransactionByHash](https://www.quicknode.com/docs/ethereum/eth_getTransactionByHash): for an ERC20 `transfer` transaction, it correctly returns the token address in the `to` field, and in the `input` field we could extract/parse the real destination.

However, we alreaady use another provider to list the correct parsed data of ERC20 transactions. So, to fix it, we decide to reload the transactions from the provider that returns the correct data, and to not process a mined transaction with the method `getTransaction`.

# Changes

- Use service `loadEthereumTransactions` in `processMinedTransaction`, instead of calling `getTransaction` method.

# Tests

### Before

Even after the transaction is in a mined state, the `to` field is the token address:

https://github.com/user-attachments/assets/1eb88b04-8065-4472-bf96-ebc67ef716d0

### After

The `to` field becomes the real destination after the transaction is mined:

https://github.com/user-attachments/assets/0c7059ea-64fc-4192-97ea-f91214600612

### DIFF
--- a/src/frontend/src/eth/providers/alchemy.providers.ts
+++ b/src/frontend/src/eth/providers/alchemy.providers.ts
@@ -8,13 +8,7 @@ import type { NetworkId } from '$lib/types/network';
 import { replacePlaceholders } from '$lib/utils/i18n.utils';
 import { assertNonNullish, nonNullish } from '@dfinity/utils';
 import type { Listener, TransactionResponse } from '@ethersproject/abstract-provider';
-import {
-	Alchemy,
-	AlchemySubscription,
-	type AlchemySettings,
-	type Network,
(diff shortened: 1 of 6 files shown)


AntonioVentilii:
This PR is replaced by one that goes to a specific release-dedicated branch:

- PR #5204

PR #5206, opened 2025-03-13
feat(frontend): hide the destination for pending ERC20 transactions

AntonioVentilii:
# Motivation

When we have a pending ERC20 transaction, we typically receive an "not-totally-correct" information in the `to` field: it shows the ERC20 token (smart contract) address.

This could lead to confusions on the user side. For example the user could copy such an address thinking it was the original destination address, and re-use the copied address in a new transaction.

So, we avoid showing the `to` address for ERC20 pending transactions.  

Screenshot 2025-03-13 at 08 41 55

Unchanged for ETH native token.

Screenshot 2025-03-13 at 08 42 31


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5210, opened 2025-03-13
feat(frontend): service loadEthereumTransactions can be update only

AntonioVentilii:
# Motivation

The service `loadEthereumTransactions` typically re-set the transactions, using the method `set` of the transaction store: it ignores the existing transactions and overrwrites everything with the new ones.

However, we prefer to have a "refresh" when using such a service in processing mined transaction (function `processMinedTransaction`), using method `update` of the transaction store. There is no practical difference, but it allows to just refresh the data of the pending transactions, instead of completing removing it and add the respective mined ones.

It seems more correct/coherent this way.

# Changes

- Add prop `updateOnly` to `loadEthereumTransactions` (defaults to `false`): it will use method `update` for each transaction, instead of `set` for all.
- Create new service `reloadEthereumTransactions` that has `updateOnly` as `true`.\
- Use new service when processing mined transactions.

# Tests

Adapted/added tests.

### DIFF
--- a/src/frontend/src/eth/services/eth-transaction.services.ts
+++ b/src/frontend/src/eth/services/eth-transaction.services.ts
@@ -120,7 +120,7 @@ const processMinedTransaction = async ({ token }: { token: Token }) => {
 	} = token;
 
 	// Reload transactions as a transaction has been mined
-	await loadEthereumTransactions({ tokenId, networkId });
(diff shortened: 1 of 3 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5213, opened 2025-03-13
feat(frontend): map CK minter contract name for ERC20 transactions

AntonioVentilii:
# Motivation

We are already mapping the known ERC20 token names to display instead of their addresses in a transaction detail. A further step is to map the CK Minter too: if we know their address, we can map it and display a better name instead of the address.

# Changes

- Export util `toCkMinterAddress`
- Add mapping of relevant CK minter addresses in util `mapAddressToName`.
- Provide CK Minter Info to the caller of `mapAddressToName` in component `EthTransactionModal`.

# Tests

Added tests.

### DIFF
--- a/src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
+++ b/src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
@@ -22,6 +22,10 @@
 		shortenWithMiddleEllipsis
 	} from '$lib/utils/format.utils';
 	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+	import type { OptionCertifiedMinterInfo } from '$icp-eth/types/cketh-minter';
+	import { ckEthMinterInfoStore } from '$icp-eth/stores/cketh.store';
+	import { isNetworkIdSepolia } from '$lib/utils/network.utils';
+	import { ETHEREUM_TOKEN_ID, SEPOLIA_TOKEN_ID } from '$env/tokens/tokens.eth.env';
 
 	export let transaction: EthTransactionUi;
 	export let token: OptionToken;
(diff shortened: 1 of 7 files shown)


src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
```diff
@@ -44,12 +48,19 @@
 	let toExplorerUrl: string | undefined;
 	$: toExplorerUrl = notEmptyString(to) ? `${$explorerUrlStore}/address/${to}` : undefined;
 
+	let ckMinterInfo: OptionCertifiedMinterInfo;
+	$: ckMinterInfo =
+		$ckEthMinterInfoStore?.[
```
DenysKarmazynDFINITY, line 53: I think you can use `$ckEthMinterInfoStore?.[$ethereumToken.id]` instead

src/frontend/src/eth/utils/transactions.utils.ts
```diff
@@ -33,17 +39,35 @@ export const decodeErc20AbiDataValue = (data: string): BigNumber => {
...
 	)?.name;
 
+	const ckEthHelperContractAddress = toCkEthHelperContractAddress(ckMinterInfo);
+	const ckErc20HelperContractAddress = toCkErc20HelperContractAddress(ckMinterInfo);
+	const ckMinerAddress = toCkMinterAddress(ckMinterInfo);
+
+	return (
+		putativeErc20TokenName ??
+		(address === ckEthHelperContractAddress
+			? 'ckETH Minter Helper Contract'
```
DenysKarmazynDFINITY, line 62: Do we want to i18n those strings?

src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
AntonioVentilii: the issue is with the activity page, where that store is not always correct, unfort

src/frontend/src/eth/utils/transactions.utils.ts
AntonioVentilii: good question: for now I would say no, because theoretically it is not translated, it is kind of the smart contract name (like an ERC20 token name). However we should have a dynamic way for the names... added a TODO

src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
DenysKarmazynDFINITY: got it

src/frontend/src/eth/utils/transactions.utils.ts
DenysKarmazynDFINITY: sounds good, thanks

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5215, opened 2025-03-13
fix(frontend): Fixate tailwind version to 4.0.9

daviddecentage:
# Motivation

Since the latest minor versions of tailwind introduce breaking changes to gradient and viewport classes (affecting token hero and footer), we fixate the version of tailwind to the last working minor version.

The last working minor version that still works well with gradients and viewports is 4.0.9

# Changes

Changed tailwind version to 4.0.9 and removed `^` to avoid going to the latest minor version.

# Tests

Working footer and gradient:
image


AntonioVentilii (approved):
LGTM tks!

PR #5216, opened 2025-03-13
feat(frontend): util toCkMinterInfoAddresses shall return non nullish values

AntonioVentilii:
# Motivation

In util `toCkMinterInfoAddresses` we are always returning an array. However it does not make sense to return an array with `undefined` values inside.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5221, opened 2025-03-13
style(frontend): Add hover state to transaction cards

daviddecentage:
# Motivation

Transactions on activity page should have a hover state like token cards on homepage.

# Changes

Added required classes for hover state and changed list gaps as it is on the token list.

# Tests
ezgif-12e67673d1e03c


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5222, opened 2025-03-13
feat(frontend): add timer to Ethereum transactions loader

AntonioVentilii:
# Motivation

When we are in an Ethereum transaction page, we are not refreshing the transaction list. We just listen to new transactions, but ideally we should have a timer to refresh every certain amount of time.

# Changes

- Create functions to reload transactions.
- Create timer utils to start and stop.
- Assign `onMount` and `onDestroy`.

# Tests

Added tests.

### DIFF
--- a/src/frontend/src/eth/components/loaders/LoaderEthTransactions.svelte
+++ b/src/frontend/src/eth/components/loaders/LoaderEthTransactions.svelte
@@ -41,19 +41,19 @@
 			return;
 		}
 
+
 		// We don't reload the same token in a row.
-		if (tokenIdLoaded === tokenId) {
+		if (tokenIdLoaded === tokenId && !reload) {
 			loading = false;
 			return;
 		}
 
+
 		tokenIdLoaded = tokenId;
 
+
 		const { success } = reload
-			? await reloadEthereumTransactions({
-					tokenId,
-					networkId
-				})
+			? await reloadEthereumTransactions({ tokenId, networkId })
 			: await loadEthereumTransactions({ tokenId, networkId });
 
 		if (!success) {
@@ -79,15 +79,18 @@
 		await reload();
 
 		timer = setInterval(reload, WALLET_TIMER_INTERVAL_MILLIS);
+
 	};
 
 	const stopTimer = () => {
+
 		if (isNullish(timer)) {
 			return;
 		}
 
 		clearInterval(timer);
 		timer = undefined;
+
 	};
 
 	onMount(startTimer);
--- a/src/frontend/src/eth/components/loaders/LoaderEthTransactions.svelte
+++ b/src/frontend/src/eth/components/loaders/LoaderEthTransactions.svelte
@@ -51,9 +51,9 @@
 
 		const { success } = reload
 			? await reloadEthereumTransactions({
-				tokenId,
-				networkId
-			})
+					tokenId,
+					networkId
+				})
 			: await loadEthereumTransactions({ tokenId, networkId });
 
 		if (!success) {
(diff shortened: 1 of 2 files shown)


src/frontend/src/eth/components/loaders/LoaderEthTransactions.svelte
```diff
@@ -22,24 +37,59 @@
...
+	};
+
+	const startTimer = async () => {
+		if (nonNullish(timer)) {
+			return;
+		}
+
+		await reload();
+
+		timer = setInterval(reload, WALLET_TIMER_INTERVAL_MILLIS);
```
DenysKarmazynDFINITY, line 78: 30s can be a bit too often? Just wondering, we can try it to see if it works for us

src/frontend/src/eth/components/loaders/LoaderEthTransactions.svelte
```diff
@@ -22,24 +37,59 @@
...
+	const stopTimer = () => {
+		if (isNullish(timer)) {
+			return;
+		}
+
+		clearInterval(timer);
+		timer = undefined;
+	};
+
+	onMount(startTimer);
```
DenysKarmazynDFINITY, line 90: We could potentially startTimer only when `$tokenNotInitialized` is initialized, but I guess it doesn't make too much difference

src/frontend/src/tests/eth/components/loaders/LoaderEthTransactions.spec.ts
```diff
@@ -3,26 +3,36 @@ import { SEPOLIA_PEPE_TOKEN } from '$env/tokens/tokens-erc20/tokens.pepe.env';
...
 	const mockLoadTransactions = loadEthereumTransactions as MockedFunction<
 		typeof loadEthereumTransactions
 	>;
+	const mockReloadTransactions = reloadEthereumTransactions as MockedFunction<
+		typeof reloadEthereumTransactions
+	>;
 
 	beforeEach(() => {
 		vi.clearAllMocks();
+		// vi.useFakeTimers();
```
DenysKarmazynDFINITY, line 32: Shall it be removed or uncommented?

src/frontend/src/tests/eth/components/loaders/LoaderEthTransactions.spec.ts
```diff
@@ -32,6 +42,10 @@ describe('LoaderEthTransactions', () => {
 		]);
 	});
 
+	// afterEach(() => {
```
DenysKarmazynDFINITY, line 45: ditto

src/frontend/src/tests/eth/components/loaders/LoaderEthTransactions.spec.ts
AntonioVentilii: Good catch! removed

src/frontend/src/tests/eth/components/loaders/LoaderEthTransactions.spec.ts
AntonioVentilii: same

src/frontend/src/eth/components/loaders/LoaderEthTransactions.svelte
AntonioVentilii: the problem is that i have to re-run it when the token is initialized. I thought it was cleaner like this since we already check for it inside the function that is called

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/eth/components/loaders/LoaderEthTransactions.svelte
DenysKarmazynDFINITY: ok, makes sense

src/frontend/src/eth/components/loaders/LoaderEthTransactions.svelte
DenysKarmazynDFINITY: agreed 👍

PR #5231, opened 2025-03-13
fix(frontend): assign the correct testnet tokens to the user snapshots

AntonioVentilii:
# Motivation

We want to correctly treat the testnet tokens when we are sending the user snapshots.

# Changes

- Ignore ICRC testnet tokens.
- Ignore BTC and SOL local networks.
- Use variant `SplDevnet` for Ethereum and Bitcoin testnets too.

# Tests

Added one test, but the rest should be sufficient.


DenysKarmazynDFINITY (commented):
LGTM, thanks

DenysKarmazynDFINITY (approved):

PR #5241, opened 2025-03-14
fix(e2e): check for Mac in Playwright consider empty strings

AntonioVentilii:
# Motivation

If the tests run locally in a MacBook, without the `MATRIX_OS` variable set (the one used in the CIs), it will give a false negative for the option `isMac`, running for devices that are not Mac-related.


inc-man (approved):
LGTM, tks

PR #5244, opened 2025-03-14
refactor(frontend): derive display local state

peterpeterparker:
# Motivation

Svelte v5 rendering differs in terms of tabs. That's why we inline the display of the amount into a derived local state.

# Changes

- Refactor code for Svelte v5 ([branch](https://github.com/dfinity/oisy-wallet/pull/3929))


AntonioVentilii (approved):
LGTM tks!

PR #5245, opened 2025-03-14
build(frontend): bump Node LTS to fix cjs/esm

peterpeterparker:
# Motivation

Node v22.11.0 appears to have an issue causing tests to misuse `require` instead of `esm`. Consequently, a test related to WalletConnect fails in the Svelte v5 branch. Using Node LTS resolves the issue.

# Resources

- Example of CI that fails: https://github.com/dfinity/oisy-wallet/actions/runs/13851360749/job/38759086161?pr=3929

# Changes

- Bump Node LTS 22.14.0


AntonioVentilii (approved):
Good catch! Thank you for that! I went on a rampage on this ahahah!

peterpeterparker:
I just got lucky! I always use Node LTS and test were passing locally. So I could compare what was the difference between the CI and my environment.

PR #5247, opened 2025-03-14
chore(frontend): explicitly install all Wallet Connect packages

AntonioVentilii:
# Motivation

Since we are going to migrate from `@walletconnect/web3wallet` to `@reown/walletkit` (PR #5246), we need to first explicitly install the packages that we use that for now are dependencies of `@walletconnect/web3wallet`.

They are:

- `@walletconnect/auth-client`
- `@walletconnect/core`
- `@walletconnect/jsonrpc-utils`
- `@walletconnect/types`
- `@walletconnect/utils`


AntonioVentilii:
This actually needs to be done in the same PR as the migration, since there are interlocked dependencies

PR #5250, opened 2025-03-14
refactor(build): simplify util loadMetadata

AntonioVentilii:
# Motivation

Util `loadMetadata` is overcomplicated returning both the token data and the metadata. So we simplify it to just return the metadata, and it will require only the Ledger canister ID.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5265, opened 2025-03-17
feat(frontend): display additional swap provider data

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/components/convert/ConvertAmountDisplay.svelte
+++ b/src/frontend/src/lib/components/convert/ConvertAmountDisplay.svelte
@@ -9,27 +9,49 @@
 	export let amount: OptionAmount = undefined;
 	export let symbol: string;
 	export let exchangeRate: number | undefined = undefined;
+	export let displayExchangeRate = true;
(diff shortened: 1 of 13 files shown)


src/frontend/src/lib/components/convert/ConvertAmountDisplay.svelte
```diff
@@ -9,27 +9,48 @@
 	export let amount: OptionAmount = undefined;
 	export let symbol: string;
 	export let exchangeRate: number | undefined = undefined;
+	export let displayExchangeRate = true;
```
DenysKarmazynDFINITY, line 12: Maybe instead of the prop you can pass `ConvertAmountExchange` as a slot from the parents that require it? It would simplify `ConvertAmountDisplay` as you will only have smth like this pseudo-code:
```
<ModalValue>
...

  <svelte:fragment slot="secondary-value">
    <slot name="amount-exchange" />
  </svelte:frament>
</ModalValue>
```

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5272, opened 2025-03-17
fix(e2e): flaky expansion of viewport

AntonioVentilii:
# Motivation

# Changes

# Tests


AntonioVentilii:
This PR was a test to check if we could solve the issue with the viewport, but it didn't work

PR #5274, opened 2025-03-17
refactor(devops): Move clippy flags into Cargo.toml

bitdivine:
# Motivation
A developer had issues getting different `cargo clippy` results from the scripts and when calling directly on the command line.

If we move the allow and deny flags into `Cargo.toml`, we can reduce the risk of different flags being used depending on how clippy is called.  This most likely doesn't solve the issue but it reduces the number of possible causes.

# Changes
- Move clippy lint flags into `Cargo.toml`

# Tests

### Lints should be the same
- Manual code review should show the flags being the same
- CI should show no additional lints
- Injecting an error into the code should cause a warning.


AntonioVentilii (approved):
LGTM tks!

PR #5275, opened 2025-03-17
feat(frontend): DestinationValue component

DenysKarmazynDFINITY:
# Motivation

A small UI component for displaying either a label for default address or a custom address string. The "change" button can be optionally displayed via slots (the following PRs).

Default address:
Screenshot 2025-03-18 at 00 11 57

Custom address:
Screenshot 2025-03-18 at 00 12 05

### DIFF
--- a/src/frontend/src/lib/components/address/DestinationValue.svelte
+++ b/src/frontend/src/lib/components/address/DestinationValue.svelte
@@ -1,0 +1,34 @@
+<script lang="ts">
+	import IconAstronautHelmet from '$lib/components/icons/IconAstronautHelmet.svelte';
+	import NetworkLogo from '$lib/components/networks/NetworkLogo.svelte';
+	import ModalValue from '$lib/components/ui/ModalValue.svelte';
+	import { logoSizes } from '$lib/constants/components.constants';
+	import { i18n } from '$lib/stores/i18n.store';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/address/DestinationValue.svelte
```diff
@@ -0,0 +1,34 @@
...
+</script>
+
+<ModalValue>
+	<svelte:fragment slot="label">{$i18n.core.text.destination}</svelte:fragment>
+
+	<div class="flex items-center gap-2" slot="main-value">
+		{#if !isDestinationCustom}
+			<div
+				class="flex items-center justify-center"
+				style={`width: ${logoSizes['xxs']}; height: ${logoSizes['xxs']};`}
```
AntonioVentilii, line 22: maybe some transition here?

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/components/address/DestinationValue.svelte
DenysKarmazynDFINITY: The container (ModalValue component) has a transition set, and adding one to the content is not necessary at the moment as the props can't be changed within one screen, therefore animation will never be visible. Should this behavior change, we'll definitely add one 🙂

PR #5277, opened 2025-03-18
feat(frontend): map Solana program Token2022 as legacy Token program

AntonioVentilii:
# Motivation

Solana program [Token-2022](https://spl.solana.com/token-2022) is a superset of the functionality provided by the [Token Program](https://spl.solana.com/token).

So, it makes sense (according to PRODSEC's suggestion too) that we map the Token-2022 instructions exactly as the legacy Token program, at least for the ones that are in common.

This will even help with avoiding duplicated code.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5278, opened 2025-03-18
feat(frontend): displays reward boxes even if the user is no eligible

BonomoAlessandro:
# Motivation

Currently the user can only see their received rewards if they are eligible, but the rewards should always be displayed if any have been received.

# Tests

**before:**
image

**after:**
image


src/frontend/src/lib/components/rewards/RewardEarnings.svelte
```diff
@@ -25,7 +25,7 @@
 	import { networkUrl } from '$lib/utils/nav.utils';
 	import { calculateTokenUsdAmount, findTwinToken } from '$lib/utils/token.utils';
 
-	export let isEligible = false;
+	export let amountOfRewards = 0;
```
AntonioVentilii, line 28: maybe a name like `rewardsNumber` ?

AntonioVentilii (approved):
LGTM tks!

Small observation

PR #5285, opened 2025-03-18
refactor(frontend): Add styling for buttons with prop link

daviddecentage:
# Motivation

I have noticed we have a `link` prop for buttons, but the styling didn't reflect that. In order to make it look like a link we refactor the styling a bit and adjust/cleanup occurrences where the link prop was applied incorrectly.

# Changes

- Updated/added link styles
- updated ocurrences with wrong link props

# Tests

image


BonomoAlessandro (approved):
LGTM

PR #5287, opened 2025-03-18
test(frontend): do not reset console mock

AntonioVentilii:
# Motivation

We are mocking the console directly in the `vitest.setup` file. So, we need to avoid resetting it when we have tests that could raise a console log.


peterpeterparker (approved):
LGTM, thx

PR #5291, opened 2025-03-18
test(frontend): simplify IcConvertTokenWizard tests

AntonioVentilii:
# Motivation

Similar to `EthConvertTokenWizard` (PR https://github.com/dfinity/oisy-wallet/pull/5290), the tests for `IcConvertTokenWizard` are flaky, because sometimes they raise a console log for missing data (that is correct, since the tests is done exactly for that).

However, we don't want console logs, we mock them before each test. So, we do not reset the mocks but only clear them. And refactor the code to make this work.

# Changes

- Move custom mocks in `beforeEach`.
- Substitute `resetAllMocks` with `clearAllMocks`.

# Tests

Current tests worked.


peterpeterparker (approved):
Thx!

PR #5292, opened 2025-03-18
feat(frontend): Settings card components

daviddecentage:
# Motivation

For the upcoming changes on the settings page, we prepare some components to be used on the redesigned settings.

# Changes

- Added Card component
- Added Card Item component

# Tests

image
image


src/frontend/src/lib/components/settings/SettingsCardItem.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
+	import { slide } from 'svelte/transition';
+	import IconHelp from '$lib/components/icons/IconHelp.svelte';
+	const SLOTS = $$props.$$slots;
```
BonomoAlessandro, line 4: Maybe you could use it something like this?

`$$slots['secondary-value']`

BonomoAlessandro (approved):
LGTM

PR #5293, opened 2025-03-18
fix(frontend): Check icon fixes

daviddecentage:
# Motivation

In order to repurpose the check icon, we refactor it a bit to reuse it in upcoming PRs regarding the Settings page refactoring.

# Changes

- Changed fixed white color for check icon to `currentColor` so we can dynamically set color
- Added `text-white` to only current occurrence of the icon so the color doesn't change there


BonomoAlessandro (approved):
LGTM

PR #5294, opened 2025-03-18
feat(frontend): Settings list components

daviddecentage:
# Motivation

For the upcoming changes on the settings page, we prepare some components to be used on the redesigned settings.

# Changes

- Added Settings List component
- Added Settings List Item component

# Tests

image
image


src/frontend/src/lib/components/settings/SettingsList.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
+	const SLOTS = $$props.$$slots;
```
BonomoAlessandro, line 2: Here you can use it like 

`let title: boolean;
	$: title = nonNullish($$slots['title']);`

daviddecentage: updated

BonomoAlessandro (approved):
LGTM

PR #5295, opened 2025-03-18
feat(e2e): take mobile screenshot both with and without navigation bar

AntonioVentilii:
# Motivation

It is useful that some mobile screenshots are shown both in a normal viewport mode (as the users sees it) and as full page (to check all the details).

So we create both tests, with and without the navigation bar.

# Changes

- We include a function to scroll the lateral navigation menu up top (this is due to a recent change too, that unfortunately MUST enter in this PR).
- We modify the code for the viewport adjuster to verify at intervals if the maximum height changed (code by @inc-man )
- Adapt the `takeScreenshot` method to create the aforementioned tests for mobiles.

# Tests

More screenshots were included in this PR since we accumulated a few past changes.


inc-man (approved):
LGTM, tks

PR #5300, opened 2025-03-18
refactor(frontend): rename Svelte type ComponentType to Component

AntonioVentilii:
# Motivation

Migrating to Svelte v5 will require to use `Component` instead of `ComponentType`. So, in preparation of it, we just rename the `ComponentType` imports to be `ComponentType as Component`.


AntonioVentilii:
As discussed offline, it is not necessary, since we can migrate to the new interface directly, without using `as`

PR #5301, opened 2025-03-18
feat(frontend): Settings modal

daviddecentage:
# Motivation

For the upcoming changes on the settings page, we prepare settings modals to be used on the redesigned settings.

# Changes

- Added settings modal to modal store
- Added derived store for new settings modal
- Added modal types enum
- Added translation


src/frontend/src/lib/components/settings/SettingsModal.svelte
```diff
@@ -0,0 +1,33 @@
...
+		type SettingsModalType,
+		SettingsModalType as SettingsModalEnum
+	} from '$lib/enums/settings-modal-types';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+
+	let settingsType: SettingsModalType;
+	$: settingsType = $modalSettingsData;
+
+	let modalTitle: string | undefined;
```
DenysKarmazynDFINITY, line 16: tiny nitpick: let's defined `modalTitle` as `string` and keep an empty string as fallback value:
```
let modalTitle: string;
	$: modalTitle =
		settingsType === SettingsModalEnum.ENABLED_NETWORKS
			? $i18n.settings.text.active_networks
			: '';
```
it will allow to just use `modalTitle` (w/o `?? ''`) on the line 24

src/frontend/src/lib/stores/modal.store.ts
```diff
@@ -84,6 +86,7 @@ export interface ModalStore<T> extends Readable<ModalData<T>> {
 	openVipRewardState: <D extends T>(data: D) => void;
 	openRewardDetails: <D extends T>(data: D) => void;
 	openRewardState: <D extends T>(data: D) => void;
+	openSettings: (data: SettingsModalType) => void;
```
DenysKarmazynDFINITY, line 89: Is there a reason why you didn't use the default pattern here? There are a lot of other "open" functions that expect some data, and the current pattern was always sufficient

src/frontend/src/lib/stores/modal.store.ts
```diff
@@ -137,6 +140,7 @@ const initModalStore = <T>(): ModalStore<T> => {
 		openVipRewardState: setTypeWithData('vip-reward-state'),
 		openRewardDetails: setTypeWithData('reward-details'),
 		openRewardState: setTypeWithData('reward-state'),
+		openSettings: <(data: SettingsModalType) => void>setTypeWithData('settings'),
```
DenysKarmazynDFINITY, line 143: same q re. the pattern

src/frontend/src/lib/stores/modal.store.ts
daviddecentage: I had a discussion with Antonio about this. Effectively data is always typed as unknown. We should do it as I do here and refactor the other ones according as I do here to make use of the generics and really type data.

src/frontend/src/lib/components/settings/SettingsModal.svelte
daviddecentage: Updated it

src/frontend/src/lib/stores/modal.store.ts
DenysKarmazynDFINITY: ok, thanks for the explanation, sounds good to me

src/frontend/src/lib/components/settings/SettingsModal.svelte
DenysKarmazynDFINITY: please update the type as well - just string is enough

DenysKarmazynDFINITY (approved):

PR #5302, opened 2025-03-18
test(frontend): POC: Base network

AntonioVentilii:
# Motivation

# Changes

# Tests


AntonioVentilii:
Base network is integrated, so we don't need this Proof of Concept anymore

PR #5306, opened 2025-03-18
feat(frontend): remove duplicated store for testnets

AntonioVentilii:
# Motivation

Derived stores `testnets` and `testnetsEnabled` have the same identical implementation. Even if the latter is a better name, we will substitute it with `testnets`, because it is the one that is less used.

In another PR, I will provide the renaming.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5307, opened 2025-03-18
feat(frontend): create derived stores for user profile

AntonioVentilii:
# Motivation

We are going to re-use the `UserProfile` structure. So, instead of repeating the code `userProfileStore?.profile`, we create a derived store.

Furthermore, we are going to re-use the `version` too, so we create a derived store for that too.

# Changes

- Create derived store `userProfile` that is `$userProfile?.profile`
- Use it in store `userSettings`
- Create derived store `userProfileVersion` that is `userProfile?.version`
- Use it in component `DappsCarousel`.

# Tests

Added tests for the new stores.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5308, opened 2025-03-18
test(frontend): extract setup function for testnets store

AntonioVentilii:
# Motivation

We are going to change the store of testnets. To facilitate this task (and to avoid repetition of code), we first concentrate the setup of the store for the tests in a single function.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5309, opened 2025-03-18
feat(frontend): use user's profile for testnets store

AntonioVentilii:
# Motivation

The store that checks if testnets are enabled or not was using the `testnetStore`. However, since we now save such setting into the backend, we can use directly the user profile.

# Changes

- Derived store `testnets` will now use derived store `userSettings`.
- Component `NetworksTestnetsToggle` will not await for the toggle to be executed (nor successful), to avoid possible delays for the user.

# Tests

Changed mock of testnet toggle. Current tests should work as expected. Below a practical test:

https://github.com/user-attachments/assets/1b1227e8-850a-4147-8fa3-52d8becae425


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5312, opened 2025-03-18
refactor(backend): split tests of settings in sub-parts

AntonioVentilii:
# Motivation

Just to keep it more organized, we split the `settings` tests into the ones related to the `networks` and the ones related to the `dapps`.


bitdivine (approved):
LGTM.  Thanks!

PR #5313, opened 2025-03-18
feat(backend): add map of Networks settings to user's profile interface

AntonioVentilii:
# Motivation

We would like to save some networks settings in the backend, for example if a network is active or not.

As first step, we just expand the structure of settings to include a map of networks with their respective settings.

After that, we will create a function to set it.


src/shared/src/types.rs
```diff
@@ -363,17 +363,27 @@ pub mod signer {
...
+    }
+
     #[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq, Default)]
     pub struct TestnetsSettings {
         pub show_testnets: bool,
     }
 
     #[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq, Default)]
     pub struct NetworksSettings {
+        pub networks: BTreeMap<String, NetworkSettings>,
```
AntonioVentilii, line 386: Do you think a map is the correct way to go? Or better a simple array?

bitdivine: Network settings will be stored within each user profile, yes?

And the key here is the network ID?

Note that each string will need a parsing validator, to prevent someone from providing milliion character strings.

src/shared/src/types.rs
```diff
@@ -363,17 +363,27 @@ pub mod signer {
...
 pub mod networks {
+    use std::collections::BTreeMap;
+
     use candid::{CandidType, Deserialize};
 
     use crate::types::Version;
 
+    #[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq, Default)]
+    pub struct NetworkSettings {
+        pub id: String,
```
bitdivine, line 374: Should the network ID be a string or something stricter, like an enum?

src/shared/src/types.rs
AntonioVentilii: Correct, it will be stored with each and the key is the network ID.

Agreed on the parsing validator: shall I already include it in this PR?

src/shared/src/types.rs
AntonioVentilii: Good question... Not really sure, but since we are starting to include more and more networks (hence this PR to try and manage them), it does not seem really scalable.

I mean, I could add a variant and just add new variants with each new network, WDYT? For sure, as you suggested, I will limit the `id` here too. or maybe remove it altogether, since we already have it in the Map as key

bitdivine: Does it make sense to get the snapshot API ready, then see whether we can use types from there?

AntonioVentilii: So, after the offline discussion, I introduced the variants that we talked about as keys. May you please check and see if it is ok? Or better to use the variants in another way?

bitdivine (approved):

PR #5314, opened 2025-03-18
chore(build): upgrade ethers to v6

AntonioVentilii:
# Motivation

We are starting to include more EVM-based Layer-2 networks (for example Base). But before that we need to stay up-to-date with the latest `ethers` package that is version 6 (repo: https://github.com/ethers-io/ethers.js).

Thankfully there is an [easy guide](https://docs.ethers.org/v6/migrating/) to migrate from v5 to v6.

# Note

1. Most of the changes are around the deprecation of `BigNumber` in favour of built-in `bigint`.
2. Package `alchemy-sdk` is still not updated to the newest version of `ethers` (issue https://github.com/alchemyplatform/alchemy-sdk-js/issues/489). That means that we have to adapt one to the other, until they bump it.

# Changes

- Install new versions of `ethers`.
- Change override of `ws` to be specific to the `@ethersproject/providers` package that is a dependency of `alchemy-sdk`. It can be removed when they fix the dependency issue https://github.com/alchemyplatform/alchemy-sdk-js/issues/491.
- Update imports and functions names according to the [guide](https://docs.ethers.org/v6/migrating/).
- Replace all `BigNumber` instances with `bigint`.
- Remove all `BigNumber` related operations (`.add`, `.gt`, etc) in favour of arithmetic ones.
- Use `ethers/utils` instead of utils from Alchemy.
- Create parsing logic to adapt Alchemy transactions to new `ethers` transactions (since `alchemy-sdk` is still not updated: https://github.com/alchemyplatform/alchemy-sdk-js/issues/489)
- In `EtherscanProvider` class we need to use a custom `getHistory` method since it was deprecated in the new version: https://github.com/ethers-io/ethers.js/issues/4303

# Tests

I had to mock the providers that comes from the `ethers/providers` imports, even when they are nested in the component/function that is being tested. Current tests are sufficient.

Furthermore, I deployed in a test canister that pointed to staging and the following manual tests worked:
- ETH balances are correct.
- ERC20 balances are correct.
- ERC20 metadata are correct.
- ETH transactions are correct.
- ERC20 transactions are correct.
- Send ETH.
- Send ERC20.
- Convert ETH to ckETH.
- Convert ERC20 to ckERC20.
- WalletConnect:
  - Conversion from USDC to ETH.
  - Conversion from ETH to USDC.


src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
```diff
@@ -30,11 +30,11 @@
...
 
 	let from: string;
-	let to: string | undefined;
+	let to: string | null;
 	let value: bigint;
 	let timestamp: number | undefined;
-	let hash: string | undefined;
-	let blockNumber: number | undefined;
+	let hash: string | null;
+	let blockNumber: number | null;
```
daviddecentage, line 37: Im assuming Ethers returns null in this new version, right?

src/frontend/src/eth/providers/alchemy.providers.ts
```diff
@@ -111,16 +111,49 @@ export class AlchemyProvider {
...
+			gasLimit,
+			maxPriorityFeePerGas,
+			maxFeePerGas,
+			nonce,
+			data,
+			chainId,
+			wait
+		} = transaction;
+
+		const possibleUndefinedToNull = <T>(value: T | undefined): T | null => value ?? null;
```
DenysKarmazynDFINITY, line 138: Maybe worth moving to utils, I can see this helper being used in other places too. We can do it in a separate PR.

DenysKarmazynDFINITY (approved):
Massive work! Didn't see anything suspicious on fe2, so I think we should be ready to proceed with testing on staging.

BonomoAlessandro (approved):
LGTM

daviddecentage (changes requested):

src/frontend/src/eth/services/fee.services.ts
```diff
@@ -48,8 +47,7 @@ export const getErc20FeeData = async ({
 		const { getFeeData: fn } = isNetworkIdICP(targetNetworkId)
```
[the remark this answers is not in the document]
AntonioVentilii: this was already in the previous implementation; I am not really sure it is needed, but I will have a look and tackle it in another PR, in case it is not. Is that ok?

src/frontend/src/lib/components/transactions/TransactionModal.svelte
```diff
@@ -27,7 +28,7 @@
 	let from: string;
 	let to: string | undefined;
 	let timestamp: bigint | undefined;
-	let blockNumber: number | undefined;
+	let blockNumber: Option<number>;
```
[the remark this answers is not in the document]
AntonioVentilii: Ah yes, we have a lint rule that when there is something like `T | null | undefined`, it suggests to use `Option<T>`

It is a standard that we wanted to enforce a few months ago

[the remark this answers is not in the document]
daviddecentage: I think for consistency we should maybe have a type for `T | undefined` but thats for another PR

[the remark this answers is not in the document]
peterpeterparker: It's a good point though. By setting it to `Option<number>` you are changing the behavior. Plus it's not related to ether. Can you preserve the code with an eslint-disable-next-line?

src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
```diff
@@ -30,11 +30,11 @@
 	export let token: OptionToken;
 
 	let from: string;
-	let to: string | undefined;
+	let to: string | null;
```
peterpeterparker, line 33: Is there a path where we could potentially avoid to migrate from `undefined` to `null`? We generally use our nullish utils so we should be find but, still, it's a change in terms of meaning for JS.

src/frontend/src/eth/providers/alchemy.providers.ts
```diff
@@ -27,7 +27,7 @@ const configs: Record<NetworkId, AlchemySettings> = {
 	}
 };
 
-const alchemyConfig = (networkId: NetworkId): AlchemySettings => {
+const alchemyConfig = (networkId: NetworkId): Pick<AlchemySettings, 'apiKey' | 'network'> => {
```
peterpeterparker, line 30: is that related to ether bigint change?

src/frontend/src/eth/providers/alchemy.providers.ts
```diff
@@ -111,16 +111,49 @@ export class AlchemyProvider {
 		});
 	}
 
-	getTransaction = async (hash: string): Promise<TransactionResponseWithBigInt | null> => {
+	getTransaction = async (hash: string) => {
```
peterpeterparker, line 114: We do not need a return type anymore here?

src/frontend/src/eth/providers/infura-erc20-icp.providers.ts
```diff
@@ -49,9 +46,9 @@ export class InfuraErc20IcpProvider implements Erc20Provider {
 		contract: { address: contractAddress },
 		to,
 		amount
-	}: PopulateTransactionParams & { amount: BigNumber }): Promise<PopulatedTransaction> => {
-		const erc20Contract = new ethers.Contract(contractAddress, ERC20_ICP_ABI, this.provider);
-		return erc20Contract.populateTransaction.burnToAccountId(amount, to);
+	}: PopulateTransactionParams & { amount: bigint }): Promise<ContractTransaction> => {
+		const erc20Contract = new Contract(contractAddress, ERC20_ICP_ABI, this.provider);
+		return erc20Contract.burnToAccountId.populateTransaction(amount, to);
```
peterpeterparker, line 51: All these contract changes have been tested right?

src/frontend/src/eth/providers/infura-erc20.providers.ts
```diff
@@ -39,7 +36,7 @@ export class InfuraErc20Provider implements Erc20Provider {
 		return {
 			name,
 			symbol,
-			decimals
+			decimals: Number(decimals)
```
peterpeterparker, line 39: This cast is safe right?

src/frontend/src/eth/utils/transactions.utils.ts
```diff
@@ -21,12 +22,12 @@ export const isErc20TransactionApprove = (data: string | undefined): boolean =>
 	nonNullish(data) && data.startsWith(ERC20_APPROVE_HASH);
 
 export const decodeErc20AbiDataValue = (data: string): bigint => {
-	const [_to, value] = ethers.utils.defaultAbiCoder.decode(
+	const [_to, value] = AbiCoder.defaultAbiCoder().decode(
```
peterpeterparker, line 25: You tested this one? The original code sounds like the types of code that still gives me PTSD of having developing or reviewing it originally.

peterpeterparker (commented):
Does the bundle(s) become lighter or heavier?

Since this PR bumps ws, did you double check that WalletConnect still works out?

src/frontend/src/eth/providers/infura-erc20-icp.providers.ts
AntonioVentilii: Yes indeed! I tested:

- ETH balances are correct.
- ERC20 balances are correct.
- ERC20 metadata are correct.
- ETH transactions are correct.
- ERC20 transactions are correct.
- Send ETH.
- Send ERC20.
- Convert ETH to ckETH.
- Covert ERC20 to ckERC20.

src/frontend/src/eth/providers/infura-erc20.providers.ts
AntonioVentilii: Yes: decimals is now returned as `BigInt`, it was `number` before. I actually tested it in the test canister.

src/frontend/src/eth/utils/transactions.utils.ts
AntonioVentilii: Ahahahahah! Yes, tested, but in general I was following the migration guide. There is an entire chapter for utils: https://docs.ethers.org/v6/migrating/#migrate-utils

src/frontend/src/eth/providers/alchemy.providers.ts
AntonioVentilii: The problem is that the return type is `TransactionResponse` from `ethers` v6 ; however, it still exists a `TransactionResponse` from `alchemy-sdk` linked to `ethers` v5.

I tried to import the first one in all ways, even renaming, another module, etc, but it always returns conflicts for types.

I wasn't able to fix this part, so I thought to remove it and work on `AlchemyProvider` in a separate PR, trying to see if we can use directly `ethers` v6 instead of `alchemy-sdk`

src/frontend/src/eth/providers/alchemy.providers.ts
AntonioVentilii: Actually, you are right, it can be done in a separate PR:

- https://github.com/dfinity/oisy-wallet/pull/5617

src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
AntonioVentilii: yes, we can: I can map/parse the `ethers` type `Transaction` to the old `EthTransaction` that accepts for example `to?: string` instead of `to: string | null` (and the rest too)

I can do it in another PR, before merging this one, if you prefer

peterpeterparker: Yes, I would prefer. I think it's safer and better align we the general conception and usage we have of `undefined` and `null`.

src/frontend/src/eth/providers/alchemy.providers.ts
peterpeterparker: Can you add a comment or todo about it?

src/frontend/src/eth/providers/infura-erc20.providers.ts
peterpeterparker: I see. I guess we do not have the choice. Thanks for the feedback.

src/frontend/src/lib/components/transactions/TransactionModal.svelte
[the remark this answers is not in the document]
AntonioVentilii: The problem is that same as https://github.com/dfinity/oisy-wallet/pull/5314#discussion_r2027299190 

The new types provide `null` instead of `undefined`, meaning that `blockNumber` can be `null` for Ethereum transactions. But as discussed in the thread, I will do another PR to map it to the "old" version

src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
AntonioVentilii: I thought of doing like this: I will first provide PRs to "freeze" these types and not being dependent on the `ethers` type. For example:

- https://github.com/dfinity/oisy-wallet/pull/5634

Then I will create a mapper based on this branch, before proceeding with this branch. So that ultimately, it will revert these changes

src/frontend/src/eth/providers/alchemy.providers.ts
AntonioVentilii: Actually! I may have found a solution, in order to leave the types. I will check

src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
AntonioVentilii: With the latest change in `main`, I was able to revert these changes in this PR. Please have a look

src/frontend/src/eth/providers/alchemy.providers.ts
AntonioVentilii: Please have a look, I left the type, but I changed a bit its definition: I am changing all the props that should be `BigInt`

daviddecentage (approved):

src/frontend/src/eth/services/fee.services.ts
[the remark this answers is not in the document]
AntonioVentilii: Done with:

- https://github.com/dfinity/oisy-wallet/pull/5973

PR #5317, opened 2025-03-19
refactor(frontend): extract specific type for ethers transaction

AntonioVentilii:
# Motivation

We want to migrate to `ethers` v6. However, the `Transaction` type changes quite a bit, so we first define exactly what type of interface we need from `ethers.Transaction`


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5320, opened 2025-03-19
feat(frontend): use bigint instead of BigNumber in util formatToken

AntonioVentilii:
# Motivation

To smooth the migration, we start replacing `BigNumber` with `bigint` where possible, starting from a very used util that is `formatToken`.

# Changes

- Make `formatToken` accept `bigint` instead `BigNumber`.
- Adapt the usage all throughout the code (quite a lot of changes).

# Tests


src/frontend/src/lib/components/tokens/TokenInputBalance.svelte
```diff
@@ -15,7 +14,7 @@
 
 	<div class="ml-1 font-semibold">
 		{formatToken({
-			value: balance ?? ZERO,
+			value: balance?.toBigInt() ?? 0n,
```
peterpeterparker, line 17: maybe keep ZERO? It's handy to find all places where we fallback to zero no?

peterpeterparker (approved):
LGTM, thanks!

I added one comment.

I was also wondering if the migration path would be easier by making `formatToken` accept both `BigNumber | bigint` and handling the type within the function—just a thought.

src/frontend/src/lib/components/tokens/TokenInputBalance.svelte
AntonioVentilii: Yes! I created `ZERO_BI` (I will rename it if and when we deprecate `ZERO`)

PR #5323, opened 2025-03-19
fix(devops): Always run final jobs

bitdivine:
# Motivation
Many of our workflows have a final job that checks whether all the other jobs succeeded.  This makes it much easier to manage which workflows need to pass for a PR to merge, than if all the individual jobs need to be listed in the ruleset.

However I noticed that many of the workflows omit the `always()` conditional on this final job.  Omitting always will make that final step pointless: If a job dependency, listed in `needs`, is skipped or fails, the dependent job will be skipped, and a skipped job is considered as not applicable when considering a merge rule.

Having PRs mergeable when required jobs fail is of course a bug.

# Changes
- Add `always()` as a conditional for all the final steps.

# Tests

### DIFF
--- a/.github/workflows/backend-checks.yml
+++ b/.github/workflows/backend-checks.yml
@@ -44,6 +44,7 @@ jobs:
 
   may-merge:
     needs: ['lint', 'workspace-dependencies']
+    if: ${{ always() }}
     runs-on: ubuntu-24.04
     steps:
       - name: Cleared for merging
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -75,6 +75,7 @@ jobs:
 
   may-merge:
     needs: ['tests']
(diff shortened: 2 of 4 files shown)


.github/workflows/backend-checks.yml
```diff
@@ -44,6 +44,7 @@ jobs:
 
   may-merge:
     needs: ['lint', 'workspace-dependencies']
+    if: ${{ always() }}
```
bitdivine, line 47: Note: `always()` also works without the curly braces, however in [the docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-jobs-in-a-workflow#example-not-requiring-successful-dependent-jobs), curly braces are used, so I use the same form as the docs just to me sure.

AntonioVentilii (approved):

PR #5324, opened 2025-03-19
fix(frontend): Text wrap and aligning settings card and list item

daviddecentage:
# Motivation

Wrong classes were used for w-full and the text didnt break correctly. Also, in the list items the text was centered if the text fills the full width.

# Changes

- Changed all w-100 to w-full
- Added text-left to force aligning to left

# Tests
Before
image

After
image

Before
image

After
image


BonomoAlessandro (approved):
LGTM

PR #5333, opened 2025-03-19
fix(frontend): Updated-main-images on the login page 

artkorotkikh-dfinity:
Updated image to reflect Rewards and mobile bottom navigation. 
Also added a bit of an alternation in screen - highlighting BTC and conversion to ckBTC
Also - Let's try slightly larger ones for better quality.

# Motivation

Old screenshots do not reflect the app updates

# Tests

**Before** 
image

image

**After**
image

image


AntonioVentilii (approved):
As discussed offline, we couldn't reduce the size more than this, so approved

PR #5339, opened 2025-03-19
feat(frontend): new context for validation errors on token action

DenysKarmazynDFINITY:
# Motivation

The next step of creating a context for validation errors in send/convert/swap flows is to actually create all required stores and context.


AntonioVentilii (approved):
LGTM tks!

PR #5342, opened 2025-03-19
feat(frontend): balance stores will use bigint instead of BigNumber

AntonioVentilii:
# Motivation

After PR https://github.com/dfinity/oisy-wallet/pull/5320, to smooth the migration, we replace `BigNumber` with `bigint` in the balance store (and in its usages).

# Changes

- Make type `Balance` as `bigint` instead of `BigNumber`.
- Use type `Balance` in the data for `balanceStore`.

# Tests


peterpeterparker (approved):
LGTM / rubber stamp. Seems okay, but there are too many changes to fully guarantee everything is correct—though it appears to be.

PR #5344, opened 2025-03-19
fix(frontend): issue with USD value not being displayed on convert

DenysKarmazynDFINITY:
# Motivation

This PR fixes the issue that recently happened in ConvertAmountDisplay - the USD values stoped being displayed in the convert modal. It is now properly working in both Swap modal only token values does not need to be displayed, as well as in the Convert flow.

Screenshot 2025-03-19 at 23 48 29

Screenshot 2025-03-19 at 23 48 44


AntonioVentilii (approved):
LGTM tks!

PR #5349, opened 2025-03-20
feat(frontend): CK assert utils will use bigint instead of BigNumber

AntonioVentilii:
# Motivation

To smooth the migration, we replace `BigNumber` with `bigint` in the utils for ck amount assertion.

# Changes

- Make utils accept `amount` as `bigint` instead of `BigNumber`.

# Tests


src/frontend/src/icp/utils/cketh.utils.ts
```diff
@@ -113,13 +112,13 @@ export const assertCkETHBalanceEstimatedFee = ({
 
 	const { decimals, symbol } = tokenCkEth;
 
-	const estimatedFee = BigNumber.from(feeStoreData?.maxTransactionFee ?? 0n);
+	const estimatedFee = feeStoreData?.maxTransactionFee ?? 0n;
```
DenysKarmazynDFINITY, line 115: shall we use `ZERO_BI` here?

DenysKarmazynDFINITY: I saw there are a lot of other places where it can be updated, so maybe your plan to do it afterwards in another PR?

DenysKarmazynDFINITY (approved):
Small q, otherwise LGTM

PR #5351, opened 2025-03-20
feat(backend): Snapshot types

bitdivine:
# Motivation
The backend canister is to store account snapshots.

# Changes
- Add account snapshot types.

# Tests
Unit tests are included, however there is very little functionality, hence few tests.

### DIFF
--- a/src/shared/src/types/network.rs
+++ b/src/shared/src/types/network.rs
@@ -1,0 +1,60 @@
+//! Level 0 networks
+
+pub mod marker_trait {
+    use candid::{CandidType, Deserialize};
+    use serde::Serialize;
+
+    /// A marker trait, used to indicate that something is a network.
+    pub trait Network {}
+
+    /// A marker trait, used to indicate that a type is to be used with the Internet Computer
+    /// mainnet.
+    #[derive(CandidType, Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+    pub struct InternetComputer {}
+    impl Network for InternetComputer {}
+
+    /// A marker trait, used to indicate that a type is to be used with the Solana mainnet.
+    #[derive(CandidType, Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+    pub struct SolanaMainnet {}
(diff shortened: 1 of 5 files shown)


src/shared/src/types/network.rs
```diff
@@ -0,0 +1,60 @@
...
+    use serde::Serialize;
+
+    /// A marker trait, used to indicate that something is a network.
+    pub trait Network {}
+
+    /// A marker trait, used to indicate that a type is to be used with the Internet Computer
+    /// mainnet.
+    #[derive(CandidType, Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+    pub struct InternetComputer {}
+    impl Network for InternetComputer {}
```
AntonioVentilii, line 14: I think i called it `Icp` in another type (see [here](https://github.com/dfinity/oisy-wallet/blob/1c6ff3316cb204be0681d518bfbaba4e2c0c445f/src/shared/src/types.rs#L381)), but this one is better: can you rename that other variable too please? It is ok, since I am not using it yet, no backward compatibility issues

AntonioVentilii (approved):
LGTM tks! One small favour to ask:

src/shared/src/types/network.rs
bitdivine: Done!

src/shared/src/types/token_id.rs
```diff
@@ -0,0 +1,37 @@
...
+    Icrc {
+        ledger: Principal,
+        index: Option<Principal>,
+    },
+}
+impl TokenId<InternetComputer> for IcrcTokenId {}
+
+pub type SolTokenId = SolPrincipal;
+
+pub type EthTokenId = EthAddress;
```
AntonioVentilii, line 37: i think there is an empty `transaction.rs` file left-over

AntonioVentilii (approved):

src/shared/src/types/token_id.rs
AntonioVentilii: It seemed left-over, so I removed. We can always add it again if needed

PR #5360, opened 2025-03-20
feat(frontend): validation in component TokenInput will use bigint instead of BigNumber

AntonioVentilii:
# Motivation

To smooth the migration, we replace `BigNumber` with `bigint` in the validation props of component `TokenInput`.

# Changes

- Make function props `customValidate` and `customErrorValidate` accept `bigint` instead of `BigNumber`.

# Tests


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5363, opened 2025-03-20
refactor(frontend): extract type for required transaction fee data

AntonioVentilii:
# Motivation

To avoid repetition of code and simplify manageability, we extract a type for the required transaction fee data (excluding nullish values).


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5370, opened 2025-03-20
fix(frontend): Solana fees are set only if user is payer

AntonioVentilii:
# Motivation

According to [Solana documentation](https://solana.com/docs/core/fees#base-transaction-fee), the first signer is the fee payer. So we set the fee in a `SolTransactionUi` only if the user is the fee payer.

# Changes

- Create small function to extract the fee payer from `accountKeys`: it filters all the signers and returns the first.
- Use the fee payer to set the fee prop in `fetchSolTransactionsForSignature`.

# Tests

Current tests are sufficient and works.

### DIFF
--- a/src/frontend/src/sol/services/sol-transactions.services.ts
+++ b/src/frontend/src/sol/services/sol-transactions.services.ts
@@ -15,6 +15,7 @@ import {
 import { SolanaNetworks, type SolanaNetworkType } from '$sol/types/network';
 import type { GetSolTransactionsParams } from '$sol/types/sol-api';
 import type {
+	ParsedAccount,
 	SolMappedTransaction,
 	SolSignature,
 	SolTransactionUi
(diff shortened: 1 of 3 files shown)


src/frontend/src/sol/services/sol-transactions.services.ts
```diff
@@ -52,12 +58,14 @@ export const fetchSolTransactionsForSignature = async ({
...
 		confirmationStatus: status,
 		transaction: {
-			message: { instructions }
+			message: { instructions, accountKeys }
 		},
 		meta
 	} = transactionDetail;
 
 	const { fee } = meta ?? {};
+	const { pubkey: feePayer } = extractFeePayer([...(accountKeys ?? [])]) ?? {};
```
DenysKarmazynDFINITY, line 67: nitpick: it can be a bit simplified:
```
const { pubkey: feePayer } = extractFeePayer(accountKeys ?? []) ?? {};
```

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/sol/services/sol-transactions.services.ts
AntonioVentilii: I actually tried it, but since accountKeys is ReadOnly, I cannot pass it freely, i need to deconstruct it to make it mutable

DenysKarmazynDFINITY: got it, thanks for the info

PR #5371, opened 2025-03-21
feat(frontend): explicit type of method getTransaction of Solana RPC

AntonioVentilii:
# Motivation

We want to have a more explicit type for the getTransaction method of Solana RPC.

However, according to [Solana dev forum](https://solana.stackexchange.com/questions/17643/how-to-infer-or-find-basic-types-for-return-values-in-solana-web3-js-v2-library/17673#17673), exporting ‘result types’ from `@solana/kit` is impractical, because the RPC is fluid to the point where its result types change depending on the particular inputs to the calls themselves.

So, we must first create a sub-function and extract the type from there, in a more fluid way.

# Changes

- Create function that just wraps the `getTransaction` method of the Solana RPC.
- Extract the type for such type.
- Use it in the code.

# Tests

Adapted a few mock, the rest is sufficient.

### DIFF
--- a/src/frontend/src/sol/api/solana.api.ts
+++ b/src/frontend/src/sol/api/solana.api.ts
@@ -90,7 +90,7 @@ export const fetchSignatures = async ({
 	return await fetchSignaturesBatch(before);
 };
 
-export const getTransactionFromSolanaRpc = async ({
+export const getRpcTransaction = async ({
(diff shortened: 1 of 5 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5373, opened 2025-03-21
refactor(frontend): remove unnecessary swap const

DenysKarmazynDFINITY:
# Motivation

The threshold const for the swap feature is redundant, since we can re-use the global one.


AntonioVentilii (approved):
LGTM tks

PR #5380, opened 2025-03-21
refactor(frontend): use ZERO_BI instead of 0n

AntonioVentilii:
# Motivation

Replacing instances of `0n` with `ZERO_BI`.


DenysKarmazynDFINITY (approved):
Thanks!

PR #5381, opened 2025-03-21
feat(frontend): Fee context will use bigint instead of BigNumber

AntonioVentilii:
# Motivation

To smooth the migration, we replace `BigNumber` with `bigint` in the fee context.

# Changes

- Make interface `TransactionFeeData ` accept `amount` as `bigint` instead of `BigNumber`. NOTE: when we will migrate to `ethers` v6 we can switch back to using the built-in `FeeData` interface.
- Transform the results of provider method `getFeeData` into `bigint`. NOTE: it will not be necessary anymore when we migrate to `ethers` v6.

# Tests

Existing tests were sufficient.


src/frontend/src/eth/components/fee/FeeContext.svelte
```diff
@@ -59,9 +59,19 @@
 
 			const { getFeeData } = infuraProviders(sendToken.network.id);
 
+			const { gasPrice, maxPriorityFeePerGas, maxFeePerGas } = await getFeeData();
+
+			const feeDataBigInt = {
```
DenysKarmazynDFINITY, line 64: This const is temporary, right? Otherwise, it would probably make sense to have this manipulations directly in the store (`setFee`), but if it's only until we upgrade to ethers 6 then all good.

AntonioVentilii: completely temporary: as soon as we update ethers, this one will give an error (`.toBigInt` does not exists anymore), and I will have to remove it, since it is redundant with the return of `getFeeData`

DenysKarmazynDFINITY: cool!

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5383, opened 2025-03-21
test(frontend): add missing ConvertWizard tests

DenysKarmazynDFINITY:
# Motivation

While working on another PR, I noticed that ConvertWizard.spec is missing some test suites.


AntonioVentilii (approved):
Yes, tests are always good! Tks!

PR #5388, opened 2025-03-21
feat(frontend): Logo button component

daviddecentage:
# Motivation

For the upcoming refactoring of several token lists, we create a super-generic logo button component that can be adjusted via props to fit all variants that we might need.

# Changes

New component added

# Tests

Title and titleEnd
image

With description
image

With description and descriptionEnd
image

Big logo
image

Selectable
image

No dividers
image

With subtitle
image

With toggle passed in action slot:
image

Example with badge and icons like a token card
image


AntonioVentilii (commented):
We have the component `NetworkLogo` , can that one be changed/modified?

daviddecentage:
> We have the component `NetworkLogo` , can that one be changed/modified?

This component would also be used for the token cards/buttons. Also, the goal is to replace the existing components and simplify the UI structure

src/frontend/src/lib/components/ui/LogoButton.svelte
```diff
@@ -0,0 +1,85 @@
...
+	export let rounded = true;
+
+	let hasTitleSlot: boolean;
+	$: hasTitleSlot = nonNullish($$slots['title']);
+
+	let hasSubtitleSlot: boolean;
+	$: hasSubtitleSlot = nonNullish($$slots['subtitle']);
+
+	let hasTitleEndSlot: boolean;
+	$: hasTitleEndSlot = nonNullish($$slots['titleEnd']);
```
DenysKarmazynDFINITY, line 20: In the codebase, we usually use dash-case for slot names, can you please adjust this file accordingly for the sake of consistency?

daviddecentage: sure, that was a remnant since I first had it as props but figured we might wanna pass elements too

DenysKarmazynDFINITY (approved):
LGTM, thanks

AntonioVentilii (approved):

PR #5390, opened 2025-03-21
refactor(frontend): use built-in type instead of BigInt()

AntonioVentilii:
# Motivation

It is unnecessary to use `BigInt(XYZ)` since we can use the built-in type `XYZn`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5398, opened 2025-03-24
fix(frontend): fixes token view page reload

BonomoAlessandro:
# Motivation

When a user opens the `token view` and the token cannot be loaded within 5 seconds, the `manage token modal` is opened. If a user opens the `token view` and then navigates away, the `manage token modal` gets still opened. To resolve this issue the timeout should be terminated when the page is closed.

# Tests

**before:**

https://github.com/user-attachments/assets/3d1f6718-5093-4911-857e-c81e348650b3

**after:**

https://github.com/user-attachments/assets/91a74509-a292-4c42-9dfa-b8e4408dbcf2


src/frontend/src/lib/components/transactions/Transactions.svelte
```diff
@@ -24,16 +24,20 @@
 			token.name === $routeToken && $routeNetwork && token.network.id.description === $routeNetwork
 	);
 
+	let timeout: NodeJS.Timeout | undefined;
```
AntonioVentilii, line 27: just for consistency

```suggestion
	let timer: NodeJS.Timeout | undefined;
```

src/frontend/src/lib/components/transactions/Transactions.svelte
```diff
@@ -24,16 +24,20 @@
...
+
 	onMount(() => {
 		// Since we do not have the change to check whether the data fetching is completed or not, we need to use this fallback timeout.
 		// After the timeout, we assume that the fetch has failed and open the token modal.
-		setTimeout(() => {
+		timeout = setTimeout(() => {
```
AntonioVentilii, line 32:
```suggestion
		timer = setTimeout(() => {
```

src/frontend/src/lib/components/transactions/Transactions.svelte
```diff
@@ -24,16 +24,20 @@
...
 			if (isNullish($pageToken) && nonNullish($routeToken) && nonNullish(token)) {
 				modalStore.openManageTokens();
 			}
 		}, FALLBACK_TIMEOUT);
 	});
 
+	onDestroy(() => clearTimeout(timeout));
```
AntonioVentilii, line 39:
```suggestion
	onDestroy(() => clearTimeout(timer));
```

AntonioVentilii (approved):
Good one! Tks!

PR #5401, opened 2025-03-24
feat(e2e): completely hide the carousel

AntonioVentilii:
# Motivation

The carousel is difficult to handle for the E2E tests, causing quite a bit of flakiness for each change that we do. That makes the E2E tests unreliable, and they lose their importance.

Since we need them more than ever, for now we just `mask` the carousel, looking for stable tests.

We will work to have a permanent state of the carousel as priority in another PR.


inc-man (approved):
LGTM, only thing I would add is a different mask color in the playwright config. Something as in:

maskColor: "#172846"

AntonioVentilii:
@inc-man  I feel like the more it is evident the more we will remembered that it is a temporary masking and we need to fix it, and there is some component there

PR #5407, opened 2025-03-24
feat(frontend): use bigint mocked balance instead of BigNumber

AntonioVentilii:
# Motivation

We have a few constants that are mocked balances as `BigNumber`. However, we are deprecating them in favour of `bigint`, for the coming upgrade to `ethers` v6.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5408, opened 2025-03-24
refactor(frontend): no type-derivation for TransactionUiCommon

AntonioVentilii:
# Motivation

The interface TransactionUiCommon should not depend on the transaction type of ethers package. It should be independent inside OISY since it involves all possible networks. So we explicitly define the `from`, `to` and `blockNumber` properties.

Note that `to` may be undefined for transactions that are approvals, or that do not require a counterparty.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5412, opened 2025-03-24
ix(frontend): added kongswap rest to obtain alternate icrc price source

roman-nazaruk:
# Motivation

# Changes

Integrated KongSwap as a secondary price source for ICRC tokens.

### DIFF
--- a/src/frontend/src/lib/rest/kongswap.rest.ts
+++ b/src/frontend/src/lib/rest/kongswap.rest.ts
@@ -1,0 +1,31 @@
+import { KONGSWAP_API_URL } from '$env/rest/kongswap.env';
+import type { LedgerCanisterIdText } from '$icp/types/canister';
+import type { KongSwapToken } from '$lib/types/kongswap';
+
+const fetchKongSwap = async <T>(endpoint: string): Promise<T | null> => {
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/rest/kongswap.rest.ts
```diff
@@ -0,0 +1,31 @@
...
+	});
+
+	if (!response.ok) {
+		throw new Error('Fetching KongSwap failed.');
+	}
+
+	return response.json();
+};
+
+export const getKongSwapTokenById = (id: LedgerCanisterIdText): Promise<KongSwapToken | null> =>
```
DenysKarmazynDFINITY, line 18: So, I guess in the end it appeared that kong does not provide an endpoint for fetching this data in a bundle?

roman-nazaruk: They didn't respond to my messages, but after investigating their Swagger documentation, I couldn’t find any way to fetch data using a batch of IDs.

DenysKarmazynDFINITY (approved):
LGTM, thanks

just a tiny thing: could you please change the PR title to “feat” instead of “fix”?

peterpeterparker:
No tests?

PR #5414, opened 2025-03-24
ci(e2e): new environment for e2e

inc-man:
# Motivation

Since our basic e2e setup is configured we have new tests coming up that require advanced handling. For testing purposes with tokens that we cannot mock it would be unwise to do that in a local environment. For that we want to create a new e2e environment.

# Changes

Necessary configuration for the e2e environment to work:
- canister creation
- dfx.json entries
- script files adjustment
- keys adjustment

# Tests

Tested locally.


bitdivine (approved):
LGTM.  Thanks!

PR #5415, opened 2025-03-24
refactor(backend): Move type submodules to a separate files

bitdivine:
# Motivation
The shared types file has become large and it is hard to see what is in there.  It contains submodules.  Nesting modules inside one file typically only make sense when the files are small.  This not being the case any more, it would be better to make them separate files.

# Changes
- Move all modules inside the shared types.rs into separate files.
- Move remaining types into modules such as migration.

# Tests
No functionality has changed.  If it compiles, it should be correct.


AntonioVentilii (approved):
Good one! TKs!

PR #5416, opened 2025-03-24
feat(frontend): implement ModalTokensList component

DenysKarmazynDFINITY:
# Motivation

This PR includes a new component that is going to be used in Send and Swap modals (potentially, also in ManageTokens, bigger refactoring will be needed there) for displaying clickable token cards. It is based on recently implement `LogoButton` component. Also, one of the next steps will be to extend ModalTokensList with a network selector.

Screenshot 2025-03-24 at 20 10 32


src/frontend/src/lib/components/tokens/ModalTokensList.svelte
```diff
@@ -0,0 +1,57 @@
...
+
+	const dispatch = createEventDispatcher();
+
+	let filter = '';
+
+	let filteredTokens: TokenUi[] = [];
+	$: filteredTokens = filterTokens({ tokens, filter });
+
+	let noTokensMatch = false;
+	$: noTokensMatch = filteredTokens.length === 0;
```
AntonioVentilii, line 23: I am not really sure, but maybe this should be 

`let noTokensMatch = filteredTokens.length === 0;`

because we are going to update the lint package. See here: https://github.com/dfinity/oisy-wallet/pull/5420

src/frontend/src/lib/components/tokens/ModalTokensListItem.svelte
```diff
@@ -0,0 +1,43 @@
...
+	import { i18n } from '$lib/stores/i18n.store';
+	import type { LogoSize } from '$lib/types/components';
+	import type { CardData } from '$lib/types/token-card';
+
+	export let data: CardData;
+	export let logoSize: LogoSize = 'md';
+</script>
+
+<LogoButton on:click dividers={true}>
+	<svelte:fragment slot="title">
```
AntonioVentilii, line 16: maybe already deconstruct `data`? if possible

src/frontend/src/lib/components/tokens/ModalTokensList.svelte
```diff
@@ -0,0 +1,57 @@
...
+				{:else}
+					<div class="flex flex-col">
+						{#each filteredTokens as token (token.id)}
+							<ModalTokensListItem
+								on:click={() => dispatch('icTokenButtonClick', token)}
+								data={token}
+							/>
+						{/each}
+					</div>
+				{/if}
```
AntonioVentilii, line 51: any transition animation necessary for this IF statement (both situations)? or it becomes too strange?

AntonioVentilii (approved):
LGTM tks!

A few observations, non-blockers

src/frontend/src/lib/components/tokens/ModalTokensList.svelte
DenysKarmazynDFINITY: I rebased and tested this suggestion - in this case, we do need the reactivity since `filteredTokens` is being updated on `tokens` and `filter` values change.

src/frontend/src/lib/components/tokens/ModalTokensListItem.svelte
DenysKarmazynDFINITY: Done!

src/frontend/src/lib/components/tokens/ModalTokensList.svelte
DenysKarmazynDFINITY: The transition is defined inside `TokensSkeletons`, so it should be enough for this particular case.

PR #5417, opened 2025-03-25
fix(devops): Require jobs to succeed

bitdivine:
# Motivation
There is a GitHub action that checks that all required jobs succeeded.

# Changes
- Vendor & use the action.

# Tests
This action is used extensively in the nns-dapp and other repos.


.github/actions/needs_success/action.yaml
bitdivine, line 1: Taken from: https://github.com/dfinity/nns-dapp/tree/main/.github/actions/needs_success

AntonioVentilii (approved):
LGTM tks!

PR #5422, opened 2025-03-25
fix(frontend): missing reactive symbol for stores in Swap components

AntonioVentilii:
# Motivation

# Changes

# Tests


DenysKarmazynDFINITY (approved):
Good catch, thanks

peterpeterparker (commented):
Good catch actually. So it never worked out or we are resolving a bug here?

DenysKarmazynDFINITY:
@peterpeterparker it was indeed a bug. The impact is rather moderate since out of all available for swap with kong tokens just one does not support icrc2.

peterpeterparker (approved):
LGTM, thx

Nice catch then!

PR #5418, opened 2025-03-25
chore(frontend): bump eslint-config-oisy-wallet

AntonioVentilii:
# Motivation

We bump package `@difnity/eslint-config-oisy-wallet` to stay up to date.

PR #5423, opened 2025-03-25
feat(frontend): remove unnecessary reactive declarations for functions

AntonioVentilii:
# Motivation

We want to bump library @dfinity/eslint-config-oisy-wallet (PR https://github.com/dfinity/oisy-wallet/pull/5418).

However, that causes some Svelte lint issues linked to new enabled rules.

In this PR, we tackle [no-reactive-functions](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-reactive-functions/) :  

> This rule reports whenever a function is defined in a reactive statement. This isn’t necessary, as each time the function is executed it’ll already have access to the latest values necessary. Redefining the function in the reactive statement is just a waste of CPU cycles.

# Changes

Replace all reactive-declared functions with constants. 

# Tests

Existing tests will suffice.

### DIFF
--- a/src/frontend/src/btc/components/send/BtcSendAmount.svelte
+++ b/src/frontend/src/btc/components/send/BtcSendAmount.svelte
@@ -24,7 +24,7 @@
 
 	// TODO: Enable Max button by passing the `calculateMax` prop - https://dfinity.atlassian.net/browse/GIX-3114
 
-	$: customValidate = (userAmount: bigint): Error | undefined => {
+	const customValidate = (userAmount: bigint): Error | undefined => {
 		// calculate-UTXOs-fee endpoint only accepts "userAmount > 0"
 		if (invalidAmount(Number(userAmount)) || userAmount === ZERO_BI) {
 			return new BtcAmountAssertionError($i18n.send.assertion.amount_invalid);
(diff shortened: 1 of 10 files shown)


DenysKarmazynDFINITY (approved):
These changes do make sense, though we'll definitely need to do some manual tests after all related PRs are in.

LGTM, thanks

peterpeterparker (commented):
Have you manually tested all those use cases? I would actually not be that surprised if the linter would be incorrect about few of those. I even wonder if it would not be safer to ignore the reported error.

AntonioVentilii:
@peterpeterparker I was actually wondering about these changes before the linter. We already have instances where we declare functions as constants in the svelte components, even if they use stores.

For example check the `send` function in the Ethereum send flow: https://github.com/dfinity/oisy-wallet/blob/83e5dd7c3d83016734a295bbc626e85ec45bb829/src/frontend/src/eth/components/send/EthSendTokenWizard.svelte#L124

In fact, I was trying to find a way to homologate them all, and apparently now it is built-in the the checks.

peterpeterparker:
Yes but, that's a different use case. I'm not sure that setting `customValidate` or `isInvalidDestination` as constant will not have side effects.

PR #5425, opened 2025-03-25
feat(frontend): use ModalTokensList in the Swap flow

DenysKarmazynDFINITY:
# Motivation

This is the next step of aligning tokens list view - we start using it in the Swap flow.


AntonioVentilii (approved):
LGTM tks!

PR #5427, opened 2025-03-25
feat(frontend): renames reward-code service to reward service

BonomoAlessandro:
# Motivation

The service should not be called the `reward-code.service` as it handles the communication between the application and the `reward canister`. A better name for this service would be `reward.service`, which reflects its functionality.

# Changes

- renames service from `reward-code.service` to `reward.service`


AntonioVentilii (approved):
LGTM tks!

Maybe even `rewards` instead of `reward`?

PR #5428, opened 2025-03-25
refactor(frontend): move reset of sliderFrame HTML inside the function extendCarouselSliderFrame

AntonioVentilii:
# Motivation

We want to bump library @dfinity/eslint-config-oisy-wallet (PR https://github.com/dfinity/oisy-wallet/pull/5418).

However, that causes some Svelte lint issues linked to new enabled rules.

In this PR, we tackle [no-reactive-functions](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-reactive-functions/) :  

> In general, DOM manipulating should delegate to Svelte runtime. If you manipulate the DOM directly, the Svelte runtime may confuse because there is a difference between the actual DOM and the Svelte runtime’s expected DOM. Therefore this rule reports where you use DOM manipulating function. We don’t recommend but If you intentionally manipulate the DOM, simply you can ignore this ESLint report.

# Changes

Move the `Carousel` resetting snippet of `sliderFrame` HTML inside the function `extendCarouselSliderFrame`. This is possible since the function `extendCarouselSliderFrame` is used only in this particular case.

# Tests

Existing tests will suffice. Plus manual tests in local replica showed no changes in the Carousel.


DenysKarmazynDFINITY (approved):
Thanks! Svelte 5 opens new horizons for this component, I'll try to find some time to look into how to simplify it 🙂

PR #5430, opened 2025-03-25
feat(frontend): derived store for user networks settings

AntonioVentilii:
# Motivation

We separate a different derived store specifically for the user networks settings. It will come in handy wince we are adding more networks settings.


src/frontend/src/lib/derived/user-networks.derived.ts
```diff
@@ -0,0 +1,8 @@
+import type { NetworksSettings } from '$declarations/backend/backend.did';
+import { userSettings } from '$lib/derived/user-profile.derived';
+import { derived, type Readable } from 'svelte/store';
+
+export const userNetworksSettings: Readable<NetworksSettings | undefined> = derived(
```
BonomoAlessandro, line 5: Wouldn't `userSettingsNetworks` be a better name? These are networks, not the settings of the networks.

AntonioVentilii: good point! I was just trying to keep the same convention as the backend. Shall I change?

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/lib/derived/user-networks.derived.ts
BonomoAlessandro: I think it would be worth to change the name

BonomoAlessandro (approved):

src/frontend/src/lib/derived/user-networks.derived.ts
AntonioVentilii: done!

PR #5431, opened 2025-03-25
feat(frontend): derived store for user networks maps

AntonioVentilii:
# Motivation

We are saving the networks settings in the backend, especially the enabled/disabled networks. So, we need to have a way to fetch them easily. To that scope we create a store derived from the user profile.

# Changes

- Create derived store to extract the networks map from the user settings.
- If nothing is set, by default it will show all mainnets as enabled.
- Otherwise it will map each network variant to its network ID and the settings.

# Tests

Added tests.


src/frontend/src/lib/derived/user-networks.derived.ts
```diff
@@ -1,8 +1,82 @@
...
+					...acc,
+					[id]: { enabled: true, isTestnet: false }
+				}),
+				{}
+			);
+		}
+
+		const keyToNetworkId = (key: NetworkSettingsFor): NetworkId => {
+			const networkId: NetworkId | undefined =
+				'InternetComputer' in key
```
DenysKarmazynDFINITY, line 45: Can we create a mapper for that? Something like:
```
const KEY_TO_NETWORK_ID = {
 InternterComputer: ICP_NETWORK_ID,
...
}
```

and then it can be used as `const networkId = KEY_TO_NETWORK_ID[Object.keys(key)[0]]`. 

WDYT?

DenysKarmazynDFINITY: Discussed offline with Antonio, we'll keep it as it is to keep it more type-safe (e.g. to avoid type casting)

src/frontend/src/lib/derived/user-networks.derived.ts
```diff
@@ -1,8 +1,82 @@
...
+			return SUPPORTED_MAINNET_NETWORKS_IDS.reduce<UserNetworks>(
+				(acc, id) => ({
+					...acc,
+					[id]: { enabled: true, isTestnet: false }
+				}),
+				{}
+			);
+		}
+
+		const keyToNetworkId = (key: NetworkSettingsFor): NetworkId => {
```
DenysKarmazynDFINITY, line 43: I guess it's fine to have this util here and extract it if/when we have a second usecase

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5432, opened 2025-03-25
feat(frontend): Full network list of supported networks

daviddecentage:
# Motivation

After refactoring and cleaning up the network env files, we now add a full list of all supported networks.

# Changes

- Added full supported networks list
- Exported supported btc networks to use it in the networks list


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5433, opened 2025-03-25
fix(frontend): added kongswap api and schema tests

roman-nazaruk:
# Motivation

# Changes

Added kongswap api and schema tests

# Tests

Provided new tests for kongswap API and zod schema


src/frontend/src/lib/rest/kongswap.rest.ts
```diff
@@ -21,11 +21,4 @@ export const getKongSwapTokenById = (id: LedgerCanisterIdText): Promise<KongSwap
...
 ): Promise<(KongSwapToken | null)[]> =>
-	Promise.all(
-		canisterIds.map((id) =>
-			getKongSwapTokenById(id).catch((error) => {
-				console.warn(`KongSwap fallback failed for ${id}`, error);
-				return null;
-			})
-		)
-	);
+	Promise.all(canisterIds.map((id) => getKongSwapTokenById(id).catch(() => null)));
```
AntonioVentilii, line 24: instead of `catch`, let's use a `try` statement (maybe for another PR)

src/frontend/src/tests/lib/validation/kongswap.validation.spec.ts
```diff
@@ -0,0 +1,166 @@
+import {
+	KongSwapTokenSchema,
+	KongSwapTokensSchema,
+	KongSwapTokenWithMetricsSchema
+} from '$lib/types/kongswap';
+import { createMockKongSwapToken } from '$tests/mocks/exchanges.mock';
+import { describe, expect, it } from 'vitest';
+
+describe('Schema: KongSwapTokenWithMetricsSchema', () => {
```
AntonioVentilii, line 9: if these are tests only for the schemas, let's call it `kongswap.schema.spec`

if there are functions that validates, then keep the name

src/frontend/src/tests/mocks/exchanges.mock.ts
```diff
@@ -7,3 +9,62 @@ export const mockExchanges: ExchangesData = mockTokens.reduce<ExchangesData>((ac
 	acc[token.id] = { usd: mockOneUsd };
 	return acc;
 }, {});
+
+export const createMockKongSwapToken = ({
```
AntonioVentilii, line 13: let's put this in `kongswap.mock`

roman-nazaruk: I will use this function to generate KongSwapToken in a few different places, so maybe it's better to name it createKongSwapMock?

src/frontend/src/tests/lib/validation/kongswap.validation.spec.ts
roman-nazaruk: I reused the name from the other files where we parse the schema, so maybe it's better to keep it consistent?

src/frontend/src/lib/rest/kongswap.rest.ts
roman-nazaruk: Done.

AntonioVentilii: as discussed offline, let's use `allSettled` and filter just the `fulfilled`

roman-nazaruk: Changed Promise.all to Promise.allSettled

src/frontend/src/lib/rest/kongswap.rest.ts
```diff
@@ -18,15 +18,14 @@ const fetchKongSwap = async <T>(endpoint: string): Promise<T | null> => {
...
-		})
-	);
+): Promise<KongSwapToken[]> => {
+	const results = await Promise.allSettled(canisterIds.map(getKongSwapTokenById));
+
+	return results
+		.filter(
+			(result): result is PromiseFulfilledResult<KongSwapToken> => result.status === 'fulfilled'
+		)
+		.map((result) => result.value);
```
AntonioVentilii, line 30: nitpick: you can use `reduce` to reduce one loop

```typescript
return results.reduce<KongSwapToken[]>(
	(acc, result) => result.status === 'fulfilled' ? [...acc, result.value] : acc,
	[]
);
```

roman-nazaruk: Agree. Nice tip

PR #5434, opened 2025-03-25
feat(frontend): Allow dropdown popover to be displayed as modal on small screens

daviddecentage:
# Motivation

Due to the upcoming change in the network dropdown we change th component slightly. We can pass `asModalOnMobile` prop so that the dropdown popover renders as a modal on small screens.

# Changes

Changed the way the popover renders and adjusted the network selector accordingly.

# Tests

On desktop with asModalOnMobile
image

Without the prop
image

On smaller screen with asModalOnMobile
image

Without the prop
image

On very small screen with asModalOnMobile
image

Without the prop
image

### DIFF
--- a/src/frontend/src/lib/components/ui/Dropdown.svelte
+++ b/src/frontend/src/lib/components/ui/Dropdown.svelte
@@ -1,9 +1,7 @@
 <script lang="ts">
 	import { Modal, Popover } from '@dfinity/gix-components';
-	import { nonNullish } from '@dfinity/utils';
 	import DropdownButton from '$lib/components/ui/DropdownButton.svelte';
 
-	export let title: string | undefined;
 	export let disabled = false;
 	export let asModalOnMobile = false;
 	export let ariaLabel: string;
@@ -45,7 +43,7 @@
 >
 	{#if visible}
 		<Modal on:nnsClose={() => close()}>
-			<svelte:fragment slot="title">{nonNullish(title) ? title : ''}</svelte:fragment>
+			<svelte:fragment slot="title"><slot name="title" /></svelte:fragment>
 
 			<slot name="items" />
 		</Modal>
--- a/src/frontend/src/lib/components/ui/Dropdown.svelte
+++ b/src/frontend/src/lib/components/ui/Dropdown.svelte
@@ -1,8 +1,11 @@
 <script lang="ts">
-	import { Popover } from '@dfinity/gix-components';
+	import { Modal, Popover } from '@dfinity/gix-components';
+	import { nonNullish } from '@dfinity/utils';
 	import DropdownButton from '$lib/components/ui/DropdownButton.svelte';
 
+	export let title: string | undefined;
 	export let disabled = false;
+	export let asModalOnMobile = false;
 	export let ariaLabel: string;
 	export let testId: string | undefined = undefined;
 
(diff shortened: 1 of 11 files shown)


src/frontend/src/lib/components/ui/Dropdown.svelte
```diff
@@ -1,8 +1,11 @@
 <script lang="ts">
-	import { Popover } from '@dfinity/gix-components';
+	import { Modal, Popover } from '@dfinity/gix-components';
+	import { nonNullish } from '@dfinity/utils';
 	import DropdownButton from '$lib/components/ui/DropdownButton.svelte';
 
+	export let title: string | undefined;
 	export let disabled = false;
+	export let asModalOnMobile = false;
```
BonomoAlessandro, line 8: I think `asModal` would be a better name. This parameter simply tells the component whether it should be displayed as a modal or not. The fact that the component should be displayed as a modal on mobile devices is irrelevant for the component.

src/frontend/src/lib/components/ui/Dropdown.svelte
```diff
@@ -23,6 +26,28 @@
...
+<!-- Mobile dropdown displayed as modal -->
+<div
+	class="absolute"
+	class:md:hidden={asModalOnMobile}
+	class:block={asModalOnMobile}
+	class:hidden={!asModalOnMobile}
+>
+	{#if visible}
+		<Modal on:nnsClose={() => close()}>
+			<svelte:fragment slot="title">{nonNullish(title) ? title : ''}</svelte:fragment>
```
DenysKarmazynDFINITY, line 48: Instead of the `title` prop, shall we pass it as a slot directly from consumers? This line would become `<slot name='title' slot='title' />`

DenysKarmazynDFINITY:
Also, the chain fusion label seems a bit off on the desktop screenshot. Can we maybe somehow decrease paddings, font size or even shorten the label to make it display in one line?

src/frontend/src/lib/components/ui/Dropdown.svelte
daviddecentage: But asModal only would imply that it is always rendered as a modal, which it is not.

src/frontend/src/lib/components/ui/Dropdown.svelte
```diff
@@ -23,6 +24,28 @@
...
+<!-- Mobile dropdown displayed as modal -->
+<div
+	class="absolute"
+	class:md:hidden={asModalOnMobile}
+	class:block={asModalOnMobile}
+	class:hidden={!asModalOnMobile}
+>
+	{#if visible}
+		<Modal on:nnsClose={() => close()}>
+			<svelte:fragment slot="title"><slot name="title" /></svelte:fragment>
```
DenysKarmazynDFINITY, line 46: This can be simplified to `<slot name='title' slot='title' />`, `svelte.fragment` part is not needed

src/frontend/src/lib/components/ui/Dropdown.svelte
```diff
@@ -23,6 +24,28 @@
...
+</div>
+<!-- Mobile dropdown displayed as modal -->
+<div
+	class="absolute"
+	class:md:hidden={asModalOnMobile}
+	class:block={asModalOnMobile}
+	class:hidden={!asModalOnMobile}
+>
+	{#if visible}
+		<Modal on:nnsClose={() => close()}>
```
DenysKarmazynDFINITY, line 45: nitpick: `<Modal on:nnsClose={close}>`

src/frontend/src/lib/components/ui/Dropdown.svelte
daviddecentage: nice, didnt know that!

src/frontend/src/lib/components/ui/Dropdown.svelte
daviddecentage: updated

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/lib/components/ui/Dropdown.svelte
DenysKarmazynDFINITY: With the current approach, IMO having `mobile` prefix in the prop name does make sense, as it's indeed provides more context on where it's gonna be displayed as modal. Another IMO - ideally, I would keep the current `Dropdown` exactly as it is now in `main`, and instead of the prop I'd add another component called smth like `DropdownWithMobileModal` (naming tbd) and add the modal functionality there to allow consumers deciding what type of dropdown to use. No need to change it now, since the prop approach works also fine.

PR #5441, opened 2025-03-25
style(frontend): Replace all text-md to text-base

daviddecentage:
# Motivation

Changing style classes because `text-md` is not a valid tailwind class.

# Changes

Changed `text-md` to `text-base`

# Test

Nothing changes since default text is base anyways. But for future uses it will then enforce the correct size


AntonioVentilii (commented):
Since we are here, can you replace it in the other instances too? I see there is one in `RewardsRequirements`

So that this PR becomes a generic replacement of `text-md` to `text-base`

daviddecentage:
> Since we are here, can you replace it in the other instances too? I see there is one in `RewardsRequirements`
> 
> So that this PR becomes a generic replacement of `text-md` to `text-base`

Sure, done

AntonioVentilii (approved):
thanks!

PR #5442, opened 2025-03-25
refactor(frontend): Remove chain fusion text just all networks

daviddecentage:
# Motivation

To save space in the redesigned network switcher we change the chain fusion text slightly.

# Changes

Changed text from `Chain Fusion (all networks)` to `All networks`

# Tests

Before:
image

After:
image


AntonioVentilii (approved):
LGTM tks!

PR #5452, opened 2025-03-26
feat(frontend): implemented fallback logic for icrcUsdPrices

roman-nazaruk:
# Motivation

# Changes

Implemented fallback logic in the exchangeRateICRCToUsd service to fetch missing token prices from KongSwap when not available via CoinGecko.

# Tests

Added test coverage for new utility methods.


AntonioVentilii (approved):
Tks!

PR #5453, opened 2025-03-26
fix(frontend): issue with modal content

DenysKarmazynDFINITY:
# Motivation

Debugging showed that the align-items rule is not really necessary for NetworksSelector. Additionally, it was causing issue for other modals. The solution is to remove it completely.


daviddecentage (approved):
LGTM, thanks!

PR #5455, opened 2025-03-26
fix(frontend): fix explorer icon animations

BonomoAlessandro:
# Motivation

# Changes

# Tests


AntonioVentilii (commented):
A few points:

1. Before this one, let's make it so `IconlyUfo` is ONLY the `svg`. And then create a wrapper component for the animation, something like `AnimatedIconUfo` in folder `components/icons` or even in `components/ui`, since it is not strictly an icon if it is wrapped in a `div`

2. I would prefer not to have a nested solution for the classes, it is a bit "smelly", don't you think? From what I see, this icon will need to be always animated on hovering (mobile version too I imagine, since it makes no difference), and only on selected in a mobile version, right? Can we simplify the classes to have that? We know when it is using in a way or another, since the parent navigation component changes

BonomoAlessandro:
It is not possible to simplify these classes in this way. I could create a component `AnimatedIconUfo` but the issue is that it should be animated as soon as the parent component gets hovered or selected. It is not possible to determine this state without javascript... That is why i created the global `animations` definition.

BonomoAlessandro:
@AntonioVentilii What do you think about the current implementation?

AntonioVentilii (approved):
LGTM tks!

PR #5456, opened 2025-03-26
feat(frontend): Responsive component

daviddecentage:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/components/ui/Responsive.svelte
+++ b/src/frontend/src/lib/components/ui/Responsive.svelte
@@ -1,58 +1,35 @@
 <script lang="ts">
-	import { themeVariables } from '$lib/styles/tailwind/theme-variables';
-
-	type ScreensKeyType = keyof typeof themeVariables.screens;
-
-	const MIN_SCREEN: ScreensKeyType = 'xs';
-	const MAX_SCREEN: ScreensKeyType = '2.5xl';
+	import {
+		type AvailableScreen,
+		filterScreens,
+		getActiveScreen,
+		getAvailableScreens,
+		MAX_SCREEN,
+		MIN_SCREEN,
+		type ScreensKeyType,
+		shouldDisplay
+	} from '$lib/utils/screens.utils';
 
 	export let up: ScreensKeyType = MIN_SCREEN;
 	export let down: ScreensKeyType = MAX_SCREEN;
 
 	$: innerWidth = 0;
 
-	let screens: { screen: ScreensKeyType; width: number }[];
-	$: screens = ((os) => {
-		return Object.entries(os)
-			.filter(([, v]) => typeof v === 'string') // warning is wrong since we have a custom element which is an object
-			.map(([k, v]) => {
-				console.log(v, typeof v);
-				return {
-					screen: k as ScreensKeyType,
-					width: Number((v as string).replaceAll('rem', '')) * 16
-				};
-			})
-			.sort((a, b) => a.width - b.width);
-	})(themeVariables.screens);
-
-	const getActiveScreen = (w: number) => {
-		for (const { width, screen } of screens) {
-			if (w < width) {
-				return screen;
-			}
-		}
-		return MAX_SCREEN;
-	};
+	let screens: AvailableScreen[];
+	$: screens = getAvailableScreens();
 
 	let activeScreen: ScreensKeyType;
-	$: activeScreen = getActiveScreen(innerWidth);
+	$: activeScreen = getActiveScreen({ screenWidth: innerWidth, availableScreens: screens });
 
-	const getBreakpoints = () => {
-		const upIndex = screens.findIndex((screen) => screen.screen === up);
-		const downIndex = screens.findIndex((screen) => screen.screen === down);
-
-		// If either key is invalid, return an empty array
-		if (upIndex === -1 || downIndex === -1) {
-			return [];
-		}
-
-		// Return the slice of screens between `up` and `down` inclusive
-		return screens.slice(upIndex, downIndex + 1).map((screen) => screen.screen);
-	};
+	let display = false;
+	$: display = shouldDisplay({
+		filteredScreens: filterScreens({ availableScreens: screens, up, down }),
+		activeScreen
+	});
 </script>
 
 <svelte:window bind:innerWidth />
 
-{#if getBreakpoints().includes(activeScreen)}
+{#if display}
 	<slot />
 {/if}
--- a/src/frontend/src/lib/components/ui/Responsive.svelte
+++ b/src/frontend/src/lib/components/ui/Responsive.svelte
@@ -1,0 +1,58 @@
+<script lang="ts">
+	import { themeVariables } from '$lib/styles/tailwind/theme-variables';
+
+	type ScreensKeyType = keyof typeof themeVariables.screens;
+
+	const MIN_SCREEN: ScreensKeyType = 'xs';
+	const MAX_SCREEN: ScreensKeyType = '2.5xl';
+
+	export let up: ScreensKeyType = MIN_SCREEN;
+	export let down: ScreensKeyType = MAX_SCREEN;
+
+	$: innerWidth = 0;
+
+	let screens: { screen: ScreensKeyType; width: number }[];
+	$: screens = ((os) => {
+		return Object.entries(os)
+			.filter(([, v]) => typeof v === 'string') // warning is wrong since we have a custom element which is an object
+			.map(([k, v]) => {
+				console.log(v, typeof v);
+				return {
+					screen: k as ScreensKeyType,
+					width: Number((v as string).replaceAll('rem', '')) * 16
+				};
+			})
+			.sort((a, b) => a.width - b.width);
+	})(themeVariables.screens);
+
+	const getActiveScreen = (w: number) => {
+		for (const { width, screen } of screens) {
+			if (w < width) {
+				return screen;
+			}
+		}
+		return MAX_SCREEN;
+	};
+
+	let activeScreen: ScreensKeyType;
+	$: activeScreen = getActiveScreen(innerWidth);
+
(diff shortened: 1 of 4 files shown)


tailwind.config.ts
```diff
@@ -8,17 +8,7 @@ export default {
...
-			// we need to use rem instead of px because the default tailwind values changed to rem,
-			// and mixing units breaks custom screen definitions
-			xs: '28rem', // 448px
-			'1.5md': '56rem', // 896px
-			'1.5lg': '72rem', // 1152px
-			'1.5xl': '88rem', // 1408px
-			'2.5xl': '108rem', // 1728px
-			'h-md': { raw: '(max-height: 68rem)' } // ~1090px
-		},
+		screens: themeVariables.screens,
```
AntonioVentilii, line 11: here i would leave

```
{
...defaultTheme.screens,
...themeVariables.screens
}
```

but I understand that you need all the variables... is there no other way? it's just that it seems wrong to put `defaultTheme.screens` in `themeVariables` (and in general put the `screens` there)

src/frontend/src/lib/utils/screens.utils.ts
```diff
@@ -0,0 +1,64 @@
...
+export const MIN_SCREEN: ScreensKeyType = 'xs';
+export const MAX_SCREEN: ScreensKeyType = '2.5xl';
+
+export const getAvailableScreens: () => AvailableScreen[] = () =>
+	Object.entries(themeVariables.screens)
+		.filter(([, v]) => typeof v === 'string') // warning is wrong since we have a custom element which is an object
+		.map(([k, v]) => ({
+			screen: k as ScreensKeyType,
+			width: Number((v as string).replaceAll('rem', '')) * 16
+		}))
```
AntonioVentilii, line 18: can we use a `reduce` here? Plus, why not creating a constant for it, instead of a function? it seems pretty constant

src/frontend/src/lib/components/ui/Responsive.svelte
```diff
@@ -0,0 +1,56 @@
...
+	$: {
+		if (nonNullish(timeoutHandle)) {
+			clearTimeout(timeoutHandle);
+		}
+		timeoutHandle = setTimeout(() => {
+			debouncedWidth.set(innerWidth);
+		}, 50); // debounce width on screen size change so we don't calculate all the time
+	}
+
+	let screens = getAvailableScreens();
```
AntonioVentilii, line 31: this can be a `const`, right?

src/frontend/src/lib/utils/screens.utils.ts
```diff
@@ -0,0 +1,64 @@
...
+	screenWidth: number;
+	availableScreens: AvailableScreen[];
+}) => {
+	for (const { width, screen } of availableScreens) {
+		if (screenWidth < width) {
+			return screen;
+		}
+	}
+	// if nothing matches it must be the largest screen size
+	return MAX_SCREEN;
```
AntonioVentilii, line 34: any way of making more functional, for example:

`const screen = availableScreens.find(({ width }) => screenWidth < width)?.screen ?? MAX_SCREEN`

would that work? And i would move the sorting here, since this function needs them already sorted, we are not sure that they will be always provided like that. Or you can change the input `availableScreens` --> `availableScreensSortedByWidth` (or something less verbose)

src/frontend/src/lib/utils/screens.utils.ts
```diff
@@ -0,0 +1,64 @@
...
+	const downIndex = availableScreens.findIndex((screen) => screen.screen === down);
+
+	// If either key is invalid, return an empty array
+	if (upIndex === -1 || downIndex === -1) {
+		return [];
+	}
+
+	// Return the slice of screens between `up` and `down` inclusive
+	return availableScreens.slice(upIndex, downIndex + 1).map((screen) => screen.screen);
+};
```
AntonioVentilii, line 56: any way of making this more functional?

src/frontend/src/lib/utils/screens.utils.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const shouldDisplayForScreen = ({
+	filteredScreens,
+	activeScreen
+}: {
+	filteredScreens: ScreensKeyType[];
+	activeScreen: ScreensKeyType;
+}) => filteredScreens.includes(activeScreen);
```
AntonioVentilii, line 64: unless you plan of re-use this one, I would put it directly inside the component, since it is a simple condition

src/frontend/src/tests/lib/utils/screens.utils.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+	getAvailableScreens,
+	shouldDisplayForScreen,
+	type ScreensKeyType
+} from '$lib/utils/screens.utils';
+import { describe, expect, it } from 'vitest';
+
+// Helper to convert rem to px for test comparison
+const remToPx = (rem: string) => Number(rem.replaceAll('rem', '')) * 16;
+
+describe('screens.utils tests', () => {
```
AntonioVentilii, line 13: can you create tests that should raise errors? for example, for `getAvailableScreens`, what if the values are in `px` ?

src/frontend/src/lib/utils/screens.utils.ts
daviddecentage: Can you give an example? Its a one liner already

tailwind.config.ts
daviddecentage: In tailwind screens are considered part of the theme, so I think its actually even better to put it where we have our theme config.

src/frontend/src/lib/components/ui/Responsive.svelte
daviddecentage: yes updated

src/frontend/src/lib/utils/screens.utils.ts
daviddecentage: nice I like that, updated

src/frontend/src/tests/lib/utils/screens.utils.spec.ts
daviddecentage: `getAvailableScreens` cannot receive `px` values because we define them in our config in `rem`. If we would define them in `px` the test would fail which is what we want

src/frontend/src/lib/utils/screens.utils.ts
AntonioVentilii: no need for that, all variables are static imports:

```typescript
export const AVAILABLE_SCREENS: AvailableScreen[] = Object.entries(themeVariables.screens).reduce<AvailableScreen[]>((acc, [k, v]) => {
	return [...acc, ...(typeof v === 'string' ? [{
			screen: k as ScreensKeyType,
			width: Number(v.replaceAll('rem', '')) * 16
		}] : [] )
}, []).sort((a, b) => a.width - b.width);
```

src/frontend/src/lib/utils/screens.utils.ts
AntonioVentilii: well, the condition is simply using `includes` , no need to test that. It can be tested directly inside the component. If at any moment it becomes more complex, we can revisit

src/frontend/src/tests/lib/utils/screens.utils.spec.ts
AntonioVentilii: well, the function does not know that we only give rem, and actually this is a very good point in favour of tests, so that we are sure that no value is `px`; for sure we need tests that raises errors

tailwind.config.ts
AntonioVentilii: yes, but my point was not about our custom screens, but about the default ones

src/frontend/src/lib/utils/screens.utils.ts
AntonioVentilii: For example, trying to use `availableScreens.reduce(....` and remove two loops

src/frontend/src/lib/utils/screens.utils.ts
daviddecentage: Yes but then we assume people use the array from filterScreens correctly, if they always go for this method it cant be done wrong. 

Discussed offline, we can leave it in utils

tailwind.config.ts
daviddecentage: Okay Ill move them back and import default screens where we need it additionally too

src/frontend/src/lib/utils/screens.utils.ts
daviddecentage: updated

src/frontend/src/lib/utils/screens.utils.ts
daviddecentage: updated

src/frontend/src/tests/lib/utils/screens.utils.spec.ts
daviddecentage: updated the tests with a lot of invalid input tests

src/frontend/src/lib/components/ui/Responsive.svelte
```diff
@@ -0,0 +1,54 @@
...
+	export let up: ScreensKeyType = MIN_SCREEN;
+	export let down: ScreensKeyType = MAX_SCREEN;
+
+	let innerWidth = 0;
+	let debouncedWidth = writable(0);
+	let timeoutHandle: ReturnType<typeof setTimeout>;
+
+	$: {
+		if (nonNullish(timeoutHandle)) {
+			clearTimeout(timeoutHandle);
```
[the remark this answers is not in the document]
daviddecentage: Huuh I wouldve said not needed but after testing, the clearTimeout indeed leaves the timer id on the variable. Looks like we do need to set it to undefined

PR #5458, opened 2025-03-26
refactor(frontend): move User profile loader to top level

AntonioVentilii:
# Motivation

We move the component `LoaderUserProfile` before the other loaders: we are going to user the enabled networks that are saved in the user profile in the backed. That will affect the rest: loading address, transactions, prices, and so forth.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5459, opened 2025-03-26
feat(frontend): add success return to service loadUserProfile

AntonioVentilii:
# Motivation

We will need to check if the `loadUserProfile` service had a successful run or not. So we make it return a value if successful or not.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5460, opened 2025-03-26
feat(frontend): load user profile during login

AntonioVentilii:
# Motivation

We are going to save the enabled/disabled networks in the backend, in the user profile. So, we need to load the user profile before all the other services, to avoid doing work for disabled networks.

# Changes

- Add `loadUserProfile` in the `onMount` of `Loader` component, before the addresses.

# Tests

No changes in local replica.


src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -126,6 +127,15 @@
 	const validateAddresses = () => emit({ message: 'oisyValidateAddresses' });
 
 	onMount(async () => {
+		// The user profile settings will define the enabled/disabled networks.
+		// So we need to load it first to enable/disable the rest of the services.
+		const { success: userProfileSuccess } = await loadUserProfile({ identity: $authIdentity });
+
+		if (!userProfileSuccess) {
```
DenysKarmazynDFINITY, line 134: We better make this endpoint bulletproof :)

DenysKarmazynDFINITY: Maybe it's worth adding some retry mechanism here? Not in this PR

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/lib/components/loaders/Loader.svelte
AntonioVentilii: Yes, makes sense! Will work on it in another PR (inside service `loadUserProfile`)

PR #5461, opened 2025-03-26
fix(frontend): scrollable content of ModalTokensList

DenysKarmazynDFINITY:
# Motivation

I noticed today that on some screen sizes the TokensList modal displays with unwanted scrollbar. It appeared to be an issue with the `stretch` class that was giving this unexpected effect to the list. The solution is to not use `ContentWithToolbar` in this particular case. Also, I cleaned up the component by removing some unnecessary HTML/CSS.

Screenshot 2025-03-26 at 16 46 22


AntonioVentilii (approved):
Good catch!

PR #5463, opened 2025-03-26
feat(frontend): add plausible analytics

roman-nazaruk:
# Motivation

# Changes

- Integrated plausible-tracker with initialization logic.

- Added support for tracking custom events in Plausible

# Tests

- Added unit tests for initPlausibleAnalytics and trackEvent

 - Verified tracker is initialized with the correct config

- Ensured trackEvent calls Plausible only when initialized


AntonioVentilii (commented):
Let's split this PR, please

And most important: how much does `plausible-tracker` adds to the bundle?

PR #5464, opened 2025-03-26
refactor(frontend): Dropdown and modal responsive

daviddecentage:
# Motivation

In order to avoid having duplicated DOM elements, we make use of the new Responsive component instead rely on pure CSS.

# Changes

- Removed a todo comment because we cannot have the modal use the modal store, so we need the z-index left to be z-3
- Changed modal and popover of dropdown to be using the new Responsive component


BonomoAlessandro (approved):
LGTM

PR #5468, opened 2025-03-27
feat(frontend): store to check if ICP network is enabled

AntonioVentilii:
# Motivation

We will start to enable and disable network according to the user preference. That means that we could have the Internet Computer network disabled.

To that scope we need to turn off some features that would not be possible if the Internet Computer network is turned off.

# Changes

- Create derived store to check if Internet Computer network is enabled.
- Use the store to disable buttons that convert CK tokens.
- Use the store to hide the ICP addresses.

# Next Steps

- Use the store to hide the ICP/ICRC tokens

# Tests

I simulated the disabled network:

https://github.com/user-attachments/assets/067d2cb1-6a14-4cb8-bb36-9f9ad262527f

### DIFF
--- a/src/frontend/src/icp-eth/components/core/CkEthLoader.svelte
+++ b/src/frontend/src/icp-eth/components/core/CkEthLoader.svelte
@@ -10,6 +10,7 @@
 	import { erc20ToCkErc20Enabled, ethToCkETHEnabled } from '$icp-eth/derived/cketh.derived';
 	import { loadCkEthMinterInfo } from '$icp-eth/services/cketh.services';
 	import { LOCAL } from '$lib/constants/app.constants';
+	import { networkICPDisabled } from '$lib/derived/networks.derived';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { toastsError } from '$lib/stores/toasts.store';
 	import type { TokenId } from '$lib/types/token';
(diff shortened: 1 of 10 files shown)


src/frontend/src/icp-eth/components/core/CkEthLoader.svelte
```diff
@@ -10,13 +10,18 @@
...
 	import { LOCAL } from '$lib/constants/app.constants';
+	import { networkICPDisabled } from '$lib/derived/networks.derived';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { toastsError } from '$lib/stores/toasts.store';
 	import type { TokenId } from '$lib/types/token';
 
 	export let nativeTokenId: TokenId;
 
 	const load = async () => {
+		if ($networkICPDisabled) {
```
DenysKarmazynDFINITY, line 21: I would probably move this condition to the `if` statement below, but if you wanted to specifically keep it separate it's fine too

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/icp-eth/components/core/CkEthLoader.svelte
AntonioVentilii: i think I prefer it visually, just so that we know of it explicitly, make sense?

DenysKarmazynDFINITY: yep, all good

PR #5469, opened 2025-03-27
refactor(frontend): store pageToken should use enabledEthereumTokens

AntonioVentilii:
# Motivation

In store `pageToken`, the native Ethereum tokens were hardcoded, but it makes sense that they are used with store `enabledEthereumTokens` (like the rest).


DenysKarmazynDFINITY (approved):
Good one, thanks

PR #5471, opened 2025-03-27
feat(frontend): include testnets in default of userNetworks

AntonioVentilii:
# Motivation

The `userNetworks` store should consider if `testnets` are enabled or not. So that the default (in case nothing is set in the backend for the user's profile) will include both mainnets and testnets.

# Changes

- Create list of testnet networks IDs (similar to mainnets).
- Include `testnets` in the default response of the store `userNetworks` (basically when there are no settings about networks in thew user profile).

# Tests

Included tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5472, opened 2025-03-27
refactor(frontend): move store userSettingsNetworks

AntonioVentilii:
# Motivation

Since we are going to use the `userSettingsNetworks` store in other modules and we risk to cross-import, it is better to move it to the user-profile module.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5473, opened 2025-03-27
fix(e2e): fix for overlapping navigation

inc-man:
# Motivation

Currently we run into an issue where the dynamic footer starts overlapping with the navigation menu on the left side. This happens, because of the screen size and because of the carousel space that should not be there.

# Changes

Until the frontend part for this is fixed, we implement a quick fix here. We check if the overlapping is happening and if so, we increase the screensize to 1080p.

# Tests

CI runs on this PR.


inc-man:
This PR is no longer needed since we want the E2E-Tests to show that the overlapping is happening so we can see that it is wrong. The PR is a good start if this handling should ever be used again. It will use existing elements to make sure that overlapping is not happening anymore and adjusts the viewport accordingly. Its not fully done yet and needs some refining but could be a good cornerstone.

PR #5474, opened 2025-03-27
fix(frontend): added safeParse and condition to nullish kongswap response

roman-nazaruk:
# Motivation

# Changes

Added new checks for response from kongswap

# Tests

Covered new logic with tests


src/frontend/src/lib/utils/exchange.utils.ts
```diff
@@ -32,7 +32,7 @@ export const formatKongSwapToCoingeckoPrices = (
 	tokens: KongSwapToken[]
 ): CoingeckoSimpleTokenPriceResponse =>
 	tokens.reduce<CoingeckoSimpleTokenPriceResponse>((acc, { token, metrics }) => {
-		if (isNullish(metrics.price)) {
+		if (isNullish(token) || isNullish(metrics) || isNullish(metrics.price)) {
```
DenysKarmazynDFINITY, line 35: Can be a bit shorter:
```
if (isNullish(token) || isNullish(metrics?.price)) {
```

src/frontend/src/tests/lib/rest/kongswap.rest.spec.ts
```diff
@@ -103,5 +103,19 @@ describe('KongSwap REST client', () => {
 			expect(fetch).toHaveBeenCalledTimes(2);
 			expect(result).toEqual([]);
 		});
+
+		it('skips tokens with null token or null metrics', async () => {
```
DenysKarmazynDFINITY, line 107: Would be great to split this test case into 2 suites: one for `token: null` and another one for `metrics: null`

src/frontend/src/lib/utils/exchange.utils.ts
roman-nazaruk: Done.

src/frontend/src/tests/lib/rest/kongswap.rest.spec.ts
roman-nazaruk: Done.

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5476, opened 2025-03-27
feat(frontend): stores to check if Ethereum networks are enabled

AntonioVentilii:
# Motivation

We will start to enable and disable network according to the user preference. That means that we could have the Ethereum networks disabled.

# Changes

- Create derived stores to check if Ethereum networks are enabled: one for Mainnet and one for Sepolia.
- Use the stores to hide buttons that convert CK tokens.
- Use the stores to hide the Ethereum address.
- Use the stores to hide the Info modal.

# Next Steps

- Use the store to hide the Ethereum tokens (native and ERC20).

# Tests

I simulated the disabled network:

https://github.com/user-attachments/assets/1b175479-8ebf-4ac9-b046-3ed6fd60430c


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5477, opened 2025-03-27
fix(frontend): Fix testnet title and transitions in network switcher

daviddecentage:
# Motivation

After the network switcher refactoring there has been the issue that transitions dont work properly for enabled/disabled testnet. Also, the testnet title is visible even if testnet is disabled.

# Changes

Fixed transitions and testnet title when testnet is disabled

# Tests

image
image


src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
```diff
@@ -59,10 +59,12 @@
...
 
-		<span class="my-5 flex px-3 font-bold">{$i18n.networks.test_networks}</span>
-
 		{#if $testnets}
-			<ul class="flex list-none flex-col">
+			<span class="my-5 flex px-3 font-bold" transition:slide={SLIDE_EASING}
+				>{$i18n.networks.test_networks}</span
+			>
+
+			<ul class="flex list-none flex-col" transition:slide={SLIDE_EASING}>
```
AntonioVentilii, line 67: just to be sure, have you checked if it glitches to have double transition? In the past a similar structure was quite glitchy. If it is still, i think you can remove it from here and leave it in the single item of the list

AntonioVentilii (approved):
Tks! One observation:

src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
daviddecentage: I dont notice any glitchyness.

PR #5480, opened 2025-03-27
feat(frontend): filter enabled networks using user networks

AntonioVentilii:
# Motivation

We are going to use the enabled/disabled networks from the user profile saved in the backend. The most direct way of doing it is by modifying the stores that lists all the enabled network.

# Changes

- Create util to check if a network is enabled as user network.
- Apply the filter to all the enabled networks stores.

# Next Improvements

Use the same logic for the ICP network.

# Tests

Created a mock for the user networks and included more test.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5481, opened 2025-03-27
feat(frontend): avoid having all networks disabled

AntonioVentilii:
# MISSING TESTS

# Motivation

Since we are going to use the networks that are enabled/disabled in the user profile, there may be a risk of having all networks disabled. In this case, we prefer to avoid any possible issue, and return at least one network.


AntonioVentilii:
since ICP network is never disabled, this PR does not make sense

PR #5483, opened 2025-03-27
refactor(frontend): Network switcher manage button and amount of activated networks

daviddecentage:
# Motivation

We need to be able to open the settings modal where we can enable/disable the active networks, as well as display how many are active out of the total.

# Changes

- Added settings modal to modals globals
- Added amount of networks active of total
- Added translation

# Tests

image


src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
```diff
@@ -72,5 +78,23 @@
...
+					link
+					on:click={() => {
+						dropdown?.close();
+						modalStore.openSettings(SettingsModalType.ENABLED_NETWORKS);
+					}}><IconManage />{$i18n.networks.manage}</Button
+				>
+			</span>
+			<span class="ml-4 mr-2 flex text-nowrap text-right text-base">
+				{replacePlaceholders($i18n.networks.number_of_enabled, {
+					$numNetworksEnabled: `${$networksMainnets.length + $networksTestnets.length}`,
```
BonomoAlessandro, line 94: Would it make sense to check if `$networksMainnets` and `$networksTestnets` are defined and not null?

daviddecentage: As it is now, the lists can not be undefined or null

BonomoAlessandro (approved):
LGTM

PR #5484, opened 2025-03-27
feat(frontend): make NetworksSwitcher re-usable for tokens list filters

DenysKarmazynDFINITY:
# Motivation

The idea of this refactoring/update is to make NetworksSwitcher re-usable for tokens list filters. To do so, we need to make the component agnostic of what happens next after a network has been selected. This logic is now moved to consumers (`Header` in case of this PR, `ModalTokensList` - my next step).


AntonioVentilii (commented):
As first impact, before reviewing, I would say: let's not put the `NetworkSwitcher` logic in the `Header`.

If the scope is to have a reusable UI component that is the Networks dropdown, let's create that and then use it in the existing `NetworkSwitcher` component, keeping its logic.

That can be done splitting the existing component.

DenysKarmazynDFINITY:
@AntonioVentilii very good point, updated accordingly - we now have a UI component for a dropdown with networks, and `NetworksSwitcher` that uses it.

PR #5486, opened 2025-03-27
feat(frontend): create feature flag to avoid using user networks

AntonioVentilii:
# Motivation

We will start to enable/disable networks based on the preferences in the user profile. However, we want to have more control on this new feature. So we create a flag that would default the user networks the value of all mainnets networks enabled, and testnets too if they are enabled with the toggle. This is as it is before the feautre.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5490, opened 2025-03-28
refactor(frontend): rename store testnets to testnetsEnabled

AntonioVentilii:
# Motivation

It makes more sense that the derived store `testnets` is renamed `testnetsEnabled`.


DenysKarmazynDFINITY (approved):
Agree. LGTM, thanks!

PR #5491, opened 2025-03-28
test(frontend): remove VITE_ETHEREUM_MAINNET_DISABLED from test env file

AntonioVentilii:
# Motivation

Having the `VITE_ETHEREUM_MAINNET_DISABLED` flag set in the `.env.test` file makes it quite troublesome to mock in the tests: in fact, it requires a series of sequential mock, instead of a single one on the variable `ETHEREUM_MAINNET_DISABLED`. We remove it, since it can always be mocked if needed, both for unittests and E2E tests.


DenysKarmazynDFINITY (approved):
That's a good one, thanks

PR #5495, opened 2025-03-28
refactor(frontend): use selectedEthereumNetwork store instead of network enabled store

AntonioVentilii:
# Motivation

Instead of using logic based on stores `networkEthereumEnabled` and `networkSepoliaEnabled`, we can use directly the store `selectedEthereumNetwork`: by definition, it is nullish if the relevant network is not enabled.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5496, opened 2025-03-28
feat(frontend): stores to check if Bitcoin networks are enabled

AntonioVentilii:
# Motivation

We will start to enable and disable network according to the user preference. That means that we could have the Bitcoin networks disabled.

# Changes

- Create derived stores to check if Bitcoin networks are enabled: Mainnet, Testnet and Regtest.
- Use the stores to hide buttons that convert CK tokens.
- Use the stores to hide the Bitcoin addresses.
- Use the stores to hide the Info modal.

# Next Steps

- Use the store to hide the Bitcoin tokens.

# Tests

I simulated the disabled network and went in ckBTC testnet:

Screenshot 2025-03-28 at 15 38 21


src/frontend/src/lib/components/hero/Actions.svelte
```diff
@@ -40,10 +41,11 @@
 	$: convertErc20 = $erc20ToCkErc20Enabled && $erc20UserTokensInitialized;
 
 	let convertCkBtc = false;
-	$: convertCkBtc = $tokenCkBtcLedger && $erc20UserTokensInitialized;
+	$: convertCkBtc =
+		$networkBitcoinMainnetEnabled && $tokenCkBtcLedger && $erc20UserTokensInitialized;
```
DenysKarmazynDFINITY, line 45: Adding this condition here means we hide the button completely. Do we want it to behave like this? Or do we want to show it in the disabled state (which you added to the ConvertToCkBTC component above)?

AntonioVentilii: It was decided that we do not show them at all. However, just for insurance, I am adding the condition to the `disabled` prop too

DenysKarmazynDFINITY: got it, thanks for confirming!

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5500, opened 2025-03-28
feat(frontend): filter enabled tokens using user networks

AntonioVentilii:
# Motivation

We are going to use the enabled/disabled networks from the user profile saved in the backend. So, we need to hide the tokens of the disabled stores.

To do that we use the same method that we used for the enabled networks in PR https://github.com/dfinity/oisy-wallet/pull/5480


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5505, opened 2025-03-30
fix(e2e): merge-e2e tag must resolve to true when no changes

AntonioVentilii:
# Motivation

In PR https://github.com/dfinity/oisy-wallet/pull/5045 , I need to change the final step to resolve the `merge-e2e` check to positive if there are no changes in the E2E files. This is the case for PRs that do not require E2E tests, otherwise it will not pass the CI checks (see https://github.com/dfinity/oisy-wallet/actions/runs/14154886004/job/39652848558?pr=5504).


peterpeterparker (approved):
LGTM, thx

PR #5510, opened 2025-03-31
feat(frontend): stores to check if Solana networks are enabled

AntonioVentilii:
# Motivation

We will start to enable and disable network according to the user preference. That means that we could have the Solana networks disabled.

# Changes

- Create derived stores to check if Solana networks are enabled: Mainnet, Testnet, Devnet and Local.
- Use the stores to hide the Solana addresses.

# Next Steps

- Use the store to hide the Solana tokens.

# Tests

I simulated the Solana disabled network:

Screenshot 2025-03-31 at 08 04 26


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5511, opened 2025-03-31
feat(frontend): remove unnecessary store to check if ICP network is enabled

AntonioVentilii:
# Motivation

With PR https://github.com/dfinity/oisy-wallet/pull/5468 , I created a store to check if Internet Computer network was disabled or enabled by the user. However, it was decided that the user will ALWAYS have ICP network enabled.

So, to avoid confusion and wrongly rely on a store that is not 100% integrated, I remove it completely, because it is not even necessary anymore.

We can always re-add it in the future if needed.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5520, opened 2025-03-31
refactor(frontend): extract parts of NetworksSwitcher into components

DenysKarmazynDFINITY:
# Motivation

After some iterations, we finally landed on how do we want to re-use NetworksSwitch in the ModalTokensList component - we only re-use particular parts of it. This PR extract them into their own components.


AntonioVentilii (approved):
LGTM tks!

PR #5521, opened 2025-03-31
feat(backend): shared-changes-for-pow-feature

DecentAgeCoder:
# :star: PR 2: Implement necessary changes in shared for PoW feature

# Motivation
Provide the shared enum, structs proof-of-work (PoW) structures and logic, including challenge creation, validation, and difficulty management.

# Changes
1. Introduced a new rust module `pow` in crate `backend`
2. Added the necessary canid structures
3. Added the constants currently used to configure the PoW Protector
4. Added PoW Store to store a challange
5. Moved `AllowSigningError` from src/backend/src/signer.rs to src/shared/src/types/signer.rs so it is exposed and can therefore be used by the canister integrations tests.
6. Since `AllowSigningError` has a dependency to `ApproveError` (defined in ) we have to add a new dependency for `ic-cycles-ledger-client` in the backend project

# Tests
Backend Tests:   All tests successfully completed
Frontend Tests:  2 tests failed (same result as on master)


bitdivine (approved):
LGTM.  Thanks!

PR #5522, opened 2025-03-31
feat(frontend): track click events for authentication button

AntonioVentilii:
# Motivation

We already track all the possible cases of results for an authentication event:

- success
- cancelled
- error

However, to be sure that we really DO track them all, we will add an event for the clicks on the authentication button. Theoretically the sum of the existing three events should match this new one.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5526, opened 2025-03-31
feat(frontend): implements qr code skeleton

BonomoAlessandro:
# Motivation

As a user i want to see a qr code skeleton if the data are loading.

# Changes

- adds qr code skeleton

# Tests
image


daviddecentage (approved):
LGTM

PR #5527, opened 2025-03-31
feat(frontend): implements receive copy skeleton

BonomoAlessandro:
# Motivation

As a user i want to see a qr receive copy skeleton if the data are loading.

# Changes

- adds skeleton

# Tests

image


daviddecentage (approved):
LGTM

PR #5530, opened 2025-03-31
feat(frontend): implements referral modal

BonomoAlessandro:
# Motivation

As a user i want to have a modal that displays my referral code and allows me to share it with friends or other people. It should also show how many people have registered using my referral code. 

# Changes

- add referral modal

# Tests

https://github.com/user-attachments/assets/791145d2-696b-43cc-a65c-58693de07395

when clicking on the share button:
image


BonomoAlessandro:
Currently the URL defined for `OISY_REFERRAL_URL` in `oisy.metadata.json` is the same as `OISY_REWARDS_URL`. This will change soon which is why it is defined as a separate value.

PR #5532, opened 2025-03-31
fix(frontend): Add test ids for new testnet toggler in network settings

daviddecentage:
# Motivation

We need to adjust the e2e tests to work with the adjusted testnet toggler (its a checkbox now which opens in the active networks settings modal)

# Changes

- Added test id for edit button in settings
- Added test id for enable testnet checkbox


inc-man (approved):
LGTM, tks

PR #5534, opened 2025-03-31
fix(e2e): snapshots tests should run on main

AntonioVentilii:
# Motivation

The E2E tests are running only for changes in E2e-related files. But we need it to always run on `main`, for each new push.


inc-man (commented):
I'd do something like: if github.branch](if: github.ref == 'refs/heads/main'

AntonioVentilii:
@inc-man  good catch! done

inc-man (approved):
LGTM, tks

PR #5540, opened 2025-03-31
test(frontend): add thresholds to the test CI

AntonioVentilii:
# Motivation

We want to start enforcing good test coverage and practices. A first small step is to set some thresholds in the `vitest` coverage. For now we set everything to 50%: it is very lax but we will slowly increase it, when we start creating more and more tests.


BonomoAlessandro:
Since our current code coverage is at 73% i don't think it makes sense to start with the required 50% coverage. Starting with 50% it will not affect anything.

AntonioVentilii:
@BonomoAlessandro you are right, it was above 70%. However, I just ran it again and we are way lower, around 57%... I checked version v1.3.4 and we were still above. I imagine we let ourselves go...

I will change it to 55%, and we will focus on increase it

BonomoAlessandro (approved):
LGTM, thx

PR #5541, opened 2025-03-31
feat(frontend): formatSecondsToNormalizedDate will work with Daylight Saving Time

AntonioVentilii:
# Motivation

There is a particular edge case for util formatSecondsToNormalizedDate: if it is not run in a UTC zone, it will return the wrong format for the "yesterday" date, since the difference in UTC dates will be less than one day.

Unfortunately it does not appear in the tests, because they are run in UTC timezone. And it is not possible to change the timezone for a specific test, unless we change it for all. So, for now, we need to rely on the new logic only and run the test locally.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5543, opened 2025-03-31
fix(devops): do not persist credentials in checkout action

AntonioVentilii:
# Motivation

An apparently known security issue of `actions/checkout` is the default of `persist-credentials` set to `true`. It was requested to the Github developers to change it (see https://github.com/actions/checkout/issues/485), but it does not seem that it is a plan in the short-terms.

So, for now we solve it our way.

# Changes

- Set `persist-credentials` to `false` in all the uses of `actions/checkout`.
- Created new action to `add-and-commit` since the existing `EndBug/add-and-commit` one does not work well if we do not persist the credentials according to [documentation](https://github.com/EndBug/add-and-commit?tab=readme-ov-file#about-actionscheckout) (see https://github.com/EndBug/add-and-commit/issues/673 too).

# Tests

CIs work as they are supposed to.


.github/actions/add-and-commit/action.yml
```diff
@@ -0,0 +1,31 @@
...
+  token:
+    description: GitHub token
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Set git remote silently
+      run: |
+        git config --global url."https://github-actions:$GITHUB_TOKEN@github.com/".insteadOf "https://github.com/"
```
bitdivine, line 21: This writes to the `$HOME/.gitconfig` file.  Maybe if the goal is to avoid persistence, the `.gitconfig` file should be restored to its former state at the end of this action?  Save, modify, use, restore.

AntonioVentilii: Good catch! Yes, I completely forgot it! Check last commit, WDYT?

bitdivine (approved):
LGTM.  Thanks!

PR #5544, opened 2025-03-31
refactor(frontend): extract service that initalize the loader

AntonioVentilii:
# Motivation

For better manageability, we extract the service called on mounting in the `Loader` component: this service is responsible for:
- load user profile;
- load mainnet addresses from IDB;
- validate such addresses;
- load the addresses from the backend in case of failures (initializing the signer allowance too).

# Changes

- Move entire internal function to new service `initLoader`.
- Use it in component `Loader`.

# Tests

Created tests specific for it.


src/frontend/src/lib/services/loader.services.ts
```diff
@@ -35,3 +40,80 @@ export const initSignerAllowance = async (): Promise<ResultSuccess> => {
...
+	progressAndLoad,
+	setProgressModal
+}: {
+	identity: OptionIdentity;
+	validateAddresses: () => void;
+	progressAndLoad: () => Promise<void>;
+	setProgressModal: (value: boolean) => void;
+}): Promise<void> => {
+	if (isNullish(identity)) {
+		await signOut({});
```
DenysKarmazynDFINITY, line 76: Let's use `nullishSignOut` here

DenysKarmazynDFINITY (approved):
One small comment, but LGTM already, thanks!

src/frontend/src/lib/services/loader.services.ts
AntonioVentilii: done!

PR #5550, opened 2025-04-01
style(frontend): hide ModalTokensList network label on small screens

DenysKarmazynDFINITY:
# Motivation

Similar to the global NetworksSwitcher, we'd like to hide the label on smaller screens in the ModalTokensList component.


daviddecentage (approved):
LGTM

PR #5554, opened 2025-04-01
fix(e2e): add SOL to receive modal mocking

inc-man:
# Motivation

The SOL addresses in the receive modal are not being mocked.

# Changes

added SOL to the list of placeholders.


AntonioVentilii (approved):
Good catch!

PR #5555, opened 2025-04-01
fix(e2e): add SOL QR codes to receive test

inc-man:
# Motivation

Since we have SOL displayed in the modal too, we can also check for the QR codes.

# Changes

added QR code tests for SOL.


AntonioVentilii (approved):
In another PR, why not creating a loop?

PR #5556, opened 2025-04-01
ci(e2e): E2E tests should be triggered when labelling a PR

AntonioVentilii:
# Motivation

To trigger the CI that does the E2E tests for the label, we need to specify the types in the trigger for the pull requests.


inc-man (approved):
LGTM, tks

PR #5558, opened 2025-04-01
fix(frontend): tokens filter input reset button

DenysKarmazynDFINITY:
# Motivation

We need to show the reset button when user start typing something and not only when there are no matching tokens found.

# Changes

1. Rename the input prop.
2. Adjust its usage in ManageTokens and ModalTokensList.

# Tests

No tests affected.


AntonioVentilii (approved):
LGTM Tks!

Nitpick: you can use `notEmptyString()` instead of `!isNullishOrEmpty`

PR #5561, opened 2025-04-01
refactor(frontend): define output type of util isNullishOrEmpty

AntonioVentilii:
# Motivation

The type definition of util `isNullishOrEmpty` should be an optional empty string instead of just a nullish value.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5568, opened 2025-04-01
feat(frontend): update testnets icons

DenysKarmazynDFINITY:
# Motivation

We need to update tesntnets icons so they all have a single style.


AntonioVentilii (commented):
Solana Devnet is the same as Solana Local: is that expected?

DenysKarmazynDFINITY:
@AntonioVentilii good question - they are indeed the same in Figma.

Screenshot 2025-04-02 at 09 00 04

AntonioVentilii (approved):
As discussed offline, design team approved the same icon, so approved

PR #5570, opened 2025-04-01
Implement necessary changes in backend for PoW feature

DecentAgeCoder:
⭐ PR 3: Implement necessary changes in backend for PoW feature
# Motivation

# Changes
The POW_ENABLED is currently set to false so it behaves as the current situation.

Even though this PR's intent is to provide the required changes 

# Tests
Backend Tests: All tests successfully completed
Frontend Tests: All tests successfully completed

### DIFF
--- a/scripts/test.backend.sh
+++ b/scripts/test.backend.sh
@@ -4,7 +4,8 @@ POCKET_IC_SERVER_VERSION=6.0.0
 OISY_UPGRADE_VERSIONS="v0.0.13,v0.0.19"
 BITCOIN_CANISTER_RELEASE="2024-08-30"
 BITCON_CANISTER_WASM="ic-btc-canister.wasm.gz"
-
+CYCLES_LEDGER_CANISTER_RELEASE="v1.0.1"
+CYCLES_LEDGER_CANISTER_WASM="cycles-ledger.wasm.gz"
 # If a backend wasm file exists at the root, it will be used for the tests.
 
 if [ -f "./backend.wasm.gz" ]; then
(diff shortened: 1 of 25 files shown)


src/backend/src/lib.rs
```diff
@@ -658,18 +686,116 @@ pub fn get_user_profile() -> Result<UserProfile, GetUserProfileError> {
...
+
+    let challenge_completion: ChallengeCompletion = pow::complete_challenge(request.nonce)
+        .map_err(|e| match e {
+            ChallengeCompletionError::InvalidNonce
+            | ChallengeCompletionError::MissingUserProfile
+            | ChallengeCompletionError::ExpiredChallenge
+            | ChallengeCompletionError::MissingChallenge
+            | ChallengeCompletionError::ChallengeAlreadySolved => {
+                AllowSigningError::PowChallenge(e)
+            }
```
bitdivine, line 778: This can be shortened to:
```suggestion
        .map_err(AllowSigningError::PowChallenge)
```

scripts/test.backend.sh
```diff
@@ -4,7 +4,8 @@ POCKET_IC_SERVER_VERSION=6.0.0
 OISY_UPGRADE_VERSIONS="v0.0.13,v0.0.19"
```
bitdivine, line 4: The PR is really quite big.  Normally we try to keep PRs to under 5 files and maybe 200 lines of code.  I would recommend putting the changes to `test.backend.sh` in a separate PR, for example.

src/backend/src/lib.rs
```diff
@@ -658,18 +686,116 @@ pub fn get_user_profile() -> Result<UserProfile, GetUserProfileError> {
...
+/// Errors are enumerated by: `CreateChallengeError`.
+///
+/// # Returns
+///
+/// * `Ok(CreateChallengeResponse)` - On successful challenge creation.
+/// * `Err(CreateChallengeError)` - If challenge creation fails due to invalid parameters or
+///   internal errors.
+#[update(guard = "may_write_user_data")]
+#[candid_method(update)]
+pub async fn create_pow_challenge() -> Result<CreateChallengeResponse, CreateChallengeError> {
```
bitdivine, line 701: I would recommend making a PR with these new canister methods with just `todo()` in the function body.  That PR would then also contain the .did file changes and the bindings.

src/backend/src/lib.rs
DecentAgeCoder: Yes, I will commit your suggestion. The reason I initially added a separate mapping for every ChallengeCompletionError was to make each one explicitly noticeable for anyone reading the code.

PR #5571, opened 2025-04-01
feat(frontend): add reactive statement for mainnets address loaders

AntonioVentilii:
# Motivation

Since we start to save the user settings for enabled and disabled networks, the component Loader should "listen" to the mainnets being enabled or disabled, in case it is needed to load their respective addresses.

So, we add such condition together with the reactive statement for the testnet addresses.

# Changes

- Load Etherum address if Ethereum mainnet is enabled.
- Load BItcoin mainnet address if Bitcoin mainnet is enabled.
- Load Solana mainnet address if Solana mainnet is enabled. 

# Tests

Added tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5573, opened 2025-04-01
feat(frontend): validate addresses only for enabled networks

AntonioVentilii:
# Motivation

We should validate addresses only for enabled networks.

# Changes

- Change component Loader to validate the addresses using stores of enabled networks.
- Add such store to the reactive statement.

# Tests

Added tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5574, opened 2025-04-01
feat(frontend): load addresses only for enabled networks

AntonioVentilii:
# Motivation

As final step for the usage of enabled/disabled networks taken from the backend, we disable the loading of addresses that are not included in the enabled networks.

For example, if the user disabled Ethereum, the next time it logs in, it does not need to load the Ethereum address. It will be loaded only when the user turn the network back on. 

# Changes

- Add list of enabled network to service `initLoader`. It will be used ONLY for the addresses that needs to be loaded from the signer.
- Provide the list of enabled networks in the `Loader` component.

# Tests

Added more tests, but a practical example in the local replica:

https://github.com/user-attachments/assets/3a8088b5-d537-4acf-a652-10f06ca61757


DenysKarmazynDFINITY (approved):
LGTM, thanks!

PR #5576, opened 2025-04-01
fix(devops): remove Github Env injection vulnerability

AntonioVentilii:
# Motivation

The CI workflow to deploy to frontend/backend presents a vulnerability pointed out by zizmor as [github-env](https://woodruffw.github.io/zizmor/audits/#github-env): we should avoid to write to `GITHUB_ENV` because it may allow code execution.

# Changes

Refactor the entire CI to pass the env variables as outputs in a step and use them as real env variables in the relevant subsequent steps.

# Tests

I ran the CI and it worked: https://github.com/dfinity/oisy-wallet/actions/runs/14206126043


bitdivine (approved):
LGTM.  Thanks!

PR #5579, opened 2025-04-02
feat(e2e): mask carousel in modal tests

AntonioVentilii:
# Motivation

To avoid flakiness, we should mask the carousel in the tests for the modals too.


AntonioVentilii:
It is not acceptable, since it is masking on the foreground... I will try to tackle the issue in another way:

- https://github.com/dfinity/oisy-wallet/pull/5581

PR #5580, opened 2025-04-02
feat(frontend): added new package plausible tracker

roman-nazaruk:
# Motivation

# Changes

Integrated plausible-tracker

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
 		"buffer": "^6.0.3",
 		"ethers": "^5.7.0",
 		"idb-keyval": "^6.2.1",
+		"plausible-tracker": "^0.3.9",
 		"svelte-confetti": "^2.3.1",
 		"zod": "^3.23.8"
 	},
--- a/scripts/build.csp.mjs
+++ b/scripts/build.csp.mjs
@@ -148,6 +148,8 @@ const updateCSP = (indexHtml) => {
 
 	const coingeckoApiConnectSrc = 'https://pro-api.coingecko.com';
 
+	const plausibleApiConnectSrc = 'https://plausible.io/api/event';
+
 	const walletConnectSrc =
 		'wss://relay.walletconnect.com wss://relay.walletconnect.org https://verify.walletconnect.com https://verify.walletconnect.org';
 	const walletConnectFrameSrc = 'https://verify.walletconnect.com https://verify.walletconnect.org';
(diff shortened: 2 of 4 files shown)


scripts/build.csp.mjs
```diff
@@ -165,7 +167,7 @@ const updateCSP = (indexHtml) => {
 	const csp = `<meta
         http-equiv="Content-Security-Policy"
         content="default-src 'none';
-        connect-src 'self' https://ic0.app https://icp0.io https://icp-api.io ${ethMainnetConnectSrc} ${ethSepoliaConnectSrc} ${walletConnectSrc} ${onramperConnectFrameSrc} ${blockstreamApiConnectSrc} ${blockchainApiConnectSrc} ${coingeckoApiConnectSrc} ${solanaApiConnectSrc};
+        connect-src 'self' https://ic0.app https://icp0.io https://icp-api.io ${ethMainnetConnectSrc} ${ethSepoliaConnectSrc} ${walletConnectSrc} ${onramperConnectFrameSrc} ${blockstreamApiConnectSrc} ${blockchainApiConnectSrc} ${coingeckoApiConnectSrc} ${solanaApiConnectSrc}  ${plausibleApiConnectSrc};
```
AntonioVentilii, line 170: let's do this in another PR

for this one we should only install the new package

package.json
```diff
@@ -87,6 +87,7 @@
 		"buffer": "^6.0.3",
 		"ethers": "^5.7.0",
 		"idb-keyval": "^6.2.1",
+		"plausible-tracker": "^0.3.9",
```
AntonioVentilii, line 90: how much does the bundles increase?

AntonioVentilii: you mean after the entire feature, right? usage etc

src/frontend/src/env/plausible.env.ts
```diff
@@ -0,0 +1 @@
+export const PLAUSIBLE_DOMAIN = 'oisy.com';
```
AntonioVentilii, line 1: this should not be here, right?

package.json
rnazaruk16: Yes, after entire feature

AntonioVentilii: tks for checking!

AntonioVentilii (approved):

PR #5581, opened 2025-04-02
feat(e2e): mask carousel with stylesheet

AntonioVentilii:
# Motivation

As per @inc-man 's suggestion, it is better to completely hide the carousel instead of mask it. This even helps with "foreground" masking (see https://github.com/dfinity/oisy-wallet/pull/5579)


inc-man (approved):
I like that! Good catch!

LGTM

PR #5583, opened 2025-04-02
feat(devops): check frontend bundle size

AntonioVentilii:
# Motivation

It is useful to have a non-blocking warning about the increase of size of the bundle for each PR.

# Changes

- Created CI workflow to build both base branch and PR branch bundles, and compare after.

# Tests

Tested on this PR, even injecting a new image: https://github.com/dfinity/oisy-wallet/actions/runs/14262508973?pr=5583


.github/workflows/frontend-bundle-check.yml
```diff
@@ -0,0 +1,109 @@
...
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+          persist-credentials: false
+
+      - name: Prepare
+        uses: ./.github/actions/prepare
+
+      - name: Install dependencies
+        run: npm ci
```
peterpeterparker, line 27: step above, prepare, does `npm ci` already

.github/workflows/frontend-bundle-check.yml
```diff
@@ -0,0 +1,109 @@
...
+
+      - name: Build project
+        run: npm run build
+
+      - name: Get bundle size
+        id: get-size
+        run: |
+          SIZE=$(du -sk "./.svelte-kit" | cut -f1)
```
peterpeterparker, line 35: Why `.svelte-kit` and not `build` which is the folder that gets deployed?

.github/workflows/frontend-bundle-check.yml
```diff
@@ -0,0 +1,109 @@
...
+          persist-credentials: false
+
+      - name: Prepare
+        uses: ./.github/actions/prepare
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build project
+        run: npm run build
```
peterpeterparker, line 59: Should we use a special `build` target that uses the same environment variables for both base and PR - just in case? Note if yes, maybe something to improve later

.github/workflows/frontend-bundle-check.yml
```diff
@@ -0,0 +1,109 @@
...
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    steps:
+      - name: Compare bundle sizes
+        run: |
+          BASE_SIZE=$BASE_SIZE
+          PR_SIZE=$PR_SIZE
+
+          DIFF=$((PR_SIZE - BASE_SIZE))
```
peterpeterparker, line 79: We do not make difference between JS and assets?

.github/workflows/frontend-bundle-check.yml
AntonioVentilii: right! removed

.github/workflows/frontend-bundle-check.yml
AntonioVentilii: not really any reason, I went back and forth between the two eheh! Makes more sense `build`, changed

.github/workflows/frontend-bundle-check.yml
AntonioVentilii: I was in doubt about that: my idea is that it would be useful even if we increase the size of some image above the threshold, we can be warned

.github/workflows/frontend-bundle-check.yml
AntonioVentilii: good suggestion! yes, I think it is better, but in that case shall we maybe use the same env variables of STAGING (because `main` is always deployed in staging)

peterpeterparker: Maybe custom?

.github/workflows/frontend-bundle-check.yml
peterpeterparker: Generally speaking, one typically reviews images manually, and I assume that the reviewer also checks the size of the images included in the PR. When it comes to size, images and JavaScript are somewhat similar but behave differently at runtime—for example, JavaScript, being heavier, can slow down the initial load.

Long story short, to get started at least, I would focus on observing JavaScript size.

AntonioVentilii: done! using wildcard `build/_app/immutable/**/*.js`

.github/workflows/frontend-bundle-check.yml
AntonioVentilii: mmm right! I will try to do it in another PR, and ping you

.github/workflows/frontend-bundle-check.yml
peterpeterparker: `.js.gzip` (or appropriate extension) for mirroring what's actually often delivered in prod?
of just `*`?

peterpeterparker (approved):
LGTM, thx

.github/workflows/frontend-bundle-check.yml
AntonioVentilii: I included `.js.gz` files

PR #5588, opened 2025-04-02
feat(frontend): added plausible to csp and created .env file

roman-nazaruk:
# Motivation

# Changes

Added plausible to csp scripts and created .env file for it.

### DIFF
--- a/src/frontend/src/env/plausible.env.ts
+++ b/src/frontend/src/env/plausible.env.ts
@@ -1,0 +1,1 @@
+export const PLAUSIBLE_DOMAIN = 'oisy.com';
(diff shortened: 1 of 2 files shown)


src/frontend/src/env/plausible.env.ts
```diff
@@ -0,0 +1 @@
+export const PLAUSIBLE_DOMAIN = 'oisy.com';
```
AntonioVentilii, line 1: let's do this in another PR, possibly, when you are using it

roman-nazaruk: Removed file

PR #5590, opened 2025-04-02
refactor(frontend): Network switcher should display enabled networks depending on if testnets are enabled

daviddecentage:
# Motivation

If we have no testnets enabled, we shouldnt show n of 10 networks enabled, but consider the total to be only the amount of mainnets.

Only if testnets and at least one of the them is enabled, we display n of 10 (all supported) networks.

Also, if all are selected we dont display the count at all.

# Changes

Changes logic on when and how to display.

# Tests

Case 1 - testnets disabled, all mainnets enabled
image
image

Case 2 - testnets disabled, 3/4 mainnets enabled
image
image

Case 3 - testnets enabled, but not individual testnets enabled:
image
image

Case 4 - testnets enabled, a few individual testnets enabled as well:
image
image

Case 5 - testnets enabled, all individual testnets enabled:
image
image


BonomoAlessandro (approved):
LGTM

PR #5593, opened 2025-04-02
refactor(frontend): replace duplicated util formatTokenAmount

AntonioVentilii:
# Motivation

Util `formatTokenAmount` is a duplicated of `formatToken`, so we replace the first with the latter.


BonomoAlessandro (approved):
LGTM thx

PR #5594, opened 2025-04-02
feat(frontend): use explicit type for etherscan provider

AntonioVentilii:
# Motivation

We want to bump `ethers` to v6. however, in v6, the function `getHistory` of class `EtherscanProvider` was deprecated ([see https://github.com/ethers-io/ethers.js/issues/4303](https://github.com/ethers-io/ethers.js/issues/4303#issuecomment-1677102615)).

So, we need to re-create getHistory as internal function (from the StackOverflow [suggestion](https://ethereum.stackexchange.com/questions/147756/read-transaction-history-with-ethers-v6-1-0/150836#150836)).

# Changes

- Create interface for method `txlist` of Etherscan API ([source](https://docs.etherscan.io/etherscan-v2/api-endpoints/accounts#get-a-list-of-normal-transactions-by-address)).
- Create method `getHistory` inside OISY class `EtherscanProvider` using the suggestion.

# Tests

Current tests are sufficient.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5595, opened 2025-04-02
refactor(frontend): explicit type for method transactions of class EtherscanRest

AntonioVentilii:
# Motivation

To be more type-safe, we explicitly set the type of the mapper in method `transactions` of class `EtherscanRest`, and clean the unnecessary values.


BonomoAlessandro (approved):
LGTM

PR #5597, opened 2025-04-03
fix(frontend): remove race condition from Loader component

AntonioVentilii:
# Motivation

The `Loader` component should await initialization before running the reactive statement. Otherwise there will be a race condition where the component is trying to load the addresses but the signer allowance is not initialized yet:

Screenshot 2025-04-03 at 08 31 01

### DIFF
--- a/src/frontend/src/lib/components/loaders/Loader.svelte
+++ b/src/frontend/src/lib/components/loaders/Loader.svelte
@@ -208,9 +208,9 @@
 {/if}
 
 <style>
-    :root:has(.login-modal) {
-        --alert-max-width: 90vw;
-        --alert-max-height: initial;
-        --dialog-border-radius: calc(var(--border-radius-sm) * 3);
-    }
+	:root:has(.login-modal) {
+		--alert-max-width: 90vw;
+		--alert-max-height: initial;
+		--dialog-border-radius: calc(var(--border-radius-sm) * 3);
+	}
 </style>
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -168,12 +170,16 @@
 	};
 
 	onMount(async () => {
+		initiatingLoader = true;
```
BonomoAlessandro, line 173: I guess this one is not needed since you already assign `true` during initialization

BonomoAlessandro (approved):
LGTM, thx

src/frontend/src/lib/components/loaders/Loader.svelte
AntonioVentilii: ah yes, you are right! Maybe just to be safe?

src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -119,7 +121,7 @@
 	const debounceLoadSolAddressDevnet = debounce(loadSolAddressDevnet);
 	const debounceLoadSolAddressLocal = debounce(loadSolAddressLocal);
 
-	$: {
+	$: if (!initiatingLoader) {
```
peterpeterparker, line 124:
```
if (progressStep !== ProgressStepsLoader.DONE) {
			return;
		}
```

AntonioVentilii: Good point! Changed

PR #5598, opened 2025-04-03
feat(frontend): added new alias for routes

roman-nazaruk:
# Motivation

We need to cover the routes with tests, so I added a new alias for the route paths.

# Changes

Added a new alias for the route paths.


AntonioVentilii (approved):
Thank you!

PR #5599, opened 2025-04-03
feat(frontend): added plausible service

roman-nazaruk:
# Motivation

# Changes

Integrated plausible-tracker

# Tests

Added unit tests for initPlausibleAnalytics and trackEvent

### DIFF
--- a/src/frontend/src/lib/services/analytics.services.ts
+++ b/src/frontend/src/lib/services/analytics.services.ts
@@ -1,7 +1,12 @@
+import { PLAUSIBLE_DOMAIN } from '$env/plausible.env';
 import { PROD } from '$lib/constants/app.constants';
 import type { TrackEventParams } from '$lib/types/analytics';
-import { isNullish } from '@dfinity/utils';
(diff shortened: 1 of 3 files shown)


AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/services/analytics.services.ts
```diff
@@ -1,7 +1,12 @@
+import { PLAUSIBLE_DOMAIN } from '$env/plausible.env';
 import { PROD } from '$lib/constants/app.constants';
 import type { TrackEventParams } from '$lib/types/analytics';
-import { isNullish } from '@dfinity/utils';
+import { isNullish, nonNullish } from '@dfinity/utils';
 import { initOrbiter, trackEvent as trackEventOrbiter } from '@junobuild/analytics';
+import Plausible from 'plausible-tracker';
+
+const domain = PLAUSIBLE_DOMAIN;
```
AntonioVentilii, line 8: maybe this can be done inline, no need to create a constant

PR #5607, opened 2025-04-03
feat(frontend): improves eth transaction loading errors

BonomoAlessandro:
# Motivation

Currently every time the request to load the ETH transactions fails, we display an error toast. In some cases this leads to multiple error toasts being displayed if the application has been idle. To improve the user experience we want to only show the error toast if the request fails 3 times in a row. In this way, the user will not be interrupted by unnecessary error messages when everything else works as expected.

# Changes

- improves error handling


src/frontend/src/eth/services/eth-transaction.services.ts
```diff
@@ -119,7 +119,7 @@ const processMinedTransaction = async ({ token }: { token: Token }) => {
 	} = token;
 
 	// Reload transactions as a transaction has been mined
-	await reloadEthereumTransactions({ tokenId, networkId });
+	await reloadEthereumTransactions({ tokenId, networkId, silent: false });
```
DenysKarmazynDFINITY, line 122: Since `silent` is `false` by default, we can leave this line as it was before

src/frontend/src/eth/services/eth-transactions.services.ts
```diff
@@ -36,16 +39,19 @@ export const loadEthereumTransactions = ({
 export const reloadEthereumTransactions = (params: {
 	tokenId: TokenId;
 	networkId: NetworkId;
+	silent: boolean;
```
DenysKarmazynDFINITY, line 42: `loadEthTransactions` has `silent` as optional, here it is required. Any reason for that? I would keep it sync between these 2 functions.

src/frontend/src/lib/constants/app.constants.ts
```diff
@@ -142,3 +142,6 @@ export const FALLBACK_TIMEOUT = 10000;
 // Git
 export const GIT_COMMIT_HASH = VITE_GIT_COMMIT_HASH;
 export const GIT_BRANCH_NAME = VITE_GIT_BRANCH_NAME;
+
+// Threshold
+export const FAILURE_THRESHOLD = 3;
```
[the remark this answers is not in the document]
BonomoAlessandro: Yes exactly. I just rebased it.

src/frontend/src/eth/services/eth-transactions.services.ts
BonomoAlessandro: You are right, that is not necessary

src/frontend/src/eth/services/eth-transaction.services.ts
BonomoAlessandro: I do not understand what you mean. The type of `silent` is boolean for the load and reload function. Did you mean why one was optional while the other was mandatory?

DenysKarmazynDFINITY: I meant that you do not need to pass `silent: false` explicitly if the param is optional and has a default value. You already applied this change, so all good.

src/frontend/src/tests/eth/services/eth-transactions.services.spec.ts
```diff
@@ -200,7 +200,8 @@ describe('eth-transactions.services', () => {
 
 			const result = await reloadEthereumTransactions({
 				networkId: mockNetworkId,
-				tokenId: mockTokenId
+				tokenId: mockTokenId,
+				silent: false
```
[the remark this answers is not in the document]
BonomoAlessandro: you are right. i removed it

src/frontend/src/eth/components/loaders/LoaderEthTransactions.svelte
```diff
@@ -50,11 +52,21 @@
...
+					networkId,
+					silent: failedReloadCounter + 1 <= FAILURE_THRESHOLD
+				})
 			: await loadEthereumTransactions({ tokenId, networkId });
 
 		if (!success) {
 			tokenIdLoaded = undefined;
+
+			if (reload) {
+				++failedReloadCounter;
```
DenysKarmazynDFINITY, line 66: Currently, the error will be shown on every failed request after the threshold is reached. Is it expected? Or do we want to reset counter after it reaches the threshold so the error message really appears on every 3rd failed request?

BonomoAlessandro: This is expected. We want to handle the errors the same way as we did in the PR [5601](https://github.com/dfinity/oisy-wallet/pull/5601)

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5608, opened 2025-04-03
feat(frontend): added feature flag plausible

roman-nazaruk:
# Motivation

# Changes

Added feature flag to test Plausible on staging.

# Tests

Updated tests.


src/frontend/src/lib/services/analytics.services.ts
```diff
@@ -47,12 +47,19 @@ export const initPlausibleAnalytics = () => {
...
 export const trackEvent = async ({ name, metadata }: TrackEventParams) => {
-	if (!PROD) {
-		return;
+	/**
+	 * We use the `PLAUSIBLE_ENABLED` feature flag instead of relying solely on `PROD`
+	 * to allow more flexibility—for example, disabling analytics in certain production builds.
+	 *
+	 * This is temporary and can be removed once testing is complete and Plausible
+	 * should only run in production.
+	 */
```
AntonioVentilii, line 56: Add a TODO here too, just so that it is flagged

AntonioVentilii (commented):
It needs to be included in the CI that deploys

src/frontend/src/lib/services/analytics.services.ts
roman-nazaruk: Added

PR #5614, opened 2025-04-03
feat(backend): provide implementation and tests for create_pow_challenge method

DecentAgeCoder:
# Motivation

The goal of this pull request is to implement and test the `create_pow_challenge` method, which is a crucial component of the backend's proof-of-work functionality. This ensures that the system generates valid challenges for clients to solve.

# Changes

- Added implementation for the `create_pow_challenge` method.
- Developed unit tests to validate the functionality and correctness of the method.

# Tests

- Verified correctness using unit tests covering various scenarios and edge cases.
- Ensured the method adheres to the expected functionality as defined in the requirements.


src/backend/src/lib.rs
```diff
@@ -659,6 +664,28 @@ pub fn get_user_profile() -> Result<UserProfile, GetUserProfileError> {
...
+/// Errors are enumerated by: `CreateChallengeError`.
+///
+/// # Returns
+///
+/// * `Ok(CreateChallengeResponse)` - On successful challenge creation.
+/// * `Err(CreateChallengeError)` - If challenge creation fails due to invalid parameters or
+///   internal errors.
+#[update(guard = "may_write_user_data")]
+#[candid_method(update)]
+pub async fn create_pow_challenge() -> Result<CreateChallengeResponse, CreateChallengeError> {
```
bitdivine, line 679: I would recommend checking that the caller is not anonymous.

src/backend/tests/it/utils/asserts.rs
```diff
@@ -0,0 +1,109 @@
...
+        max
+    );
+}
+
+#[allow(dead_code)]
+pub fn assert_deviation<T>(value: T, expected_value: T, max_deviation: T)
+where
+    T: PartialOrd + std::fmt::Debug + Sub<Output = T> + Add<Output = T> + Clone,
+{
+    let min_value = expected_value.clone() - max_deviation.clone();
```
bitdivine, line 42: Should this addition and subtraction be saturating?  What happens if expected_value + max_deviation > T::MAX?

bitdivine (approved):
LGTM.  Thanks!

src/backend/src/pow.rs
```diff
@@ -0,0 +1,333 @@
...
+#[allow(dead_code)]
+pub(crate) fn complete_challenge(
+    nonce: u64,
+) -> Result<ChallengeCompletion, ChallengeCompletionError> {
+    let principal = caller();
+    let stored_principal = StoredPrincipal(principal);
+
+    // we reject any request from a principle without a user profile
+    if !has_user_profile(stored_principal) {
+        ic_cdk::println!(
```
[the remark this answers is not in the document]
DecentAgeCoder: Good point! I think it make sense to leave the println statements in the code and have them prefixed with #[cfg(test)] as well as adding  "DEBUG: " to every message until this new feature runs stable

src/backend/src/lib.rs
DecentAgeCoder: The guard` may_write_user_data` already calls guard `caller_is_not_anonymous`

src/backend/src/pow.rs
```diff
@@ -0,0 +1,333 @@
...
+///     - The byte conversion fails.
+async fn get_random_u64() -> Result<u64, String> {
+    // Call raw_rand() and await the result
+    let (random_bytes,): (Vec<u8>,) = ic_cdk::api::management_canister::main::raw_rand()
+        .await
+        .map_err(|e| format!("raw_rand failed:  {e:?}"))?;
+
+    // Check if we have at least 8 bytes which are required for u64
+    if random_bytes.len() < 8 {
+        ic_cdk::println!(
```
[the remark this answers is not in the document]
DecentAgeCoder: I added a panic here just to be sure.

src/backend/tests/it/utils/asserts.rs
DecentAgeCoder: Yes, indeed this could lead to an issue. I will handle the saturation above and below using the num-traits library

DecentAgeCoder: Currently we do not have a direct dependency to crate num-traits. Should I implement the saturation manually?

PR #5615, opened 2025-04-03
fix(frontend): make SPL user tokens list dependant on user networks

AntonioVentilii:
# Motivation

Since we are using networks enabled/disabled by the user, I forgot to add the condition in `splUserTokens` to return an empty list in case Solana networks are disabled.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5616, opened 2025-04-03
feat(frontend): TransactionRecipient component

DenysKarmazynDFINITY:
# Motivation

As a preparation for handling multiple recipients addresses in the BtcTransactionModal component, we need to extract some logic into a separate component. Also, we add additional `transaction-custom-to` slot for displaying a custom recipient information (BTC case).


AntonioVentilii (approved):
LGTM tks

PR #5617, opened 2025-04-03
refactor(frontend): specify type for Alchemy configs

AntonioVentilii:
# Motivation

To be more precise, we specify the type of Alchemy configs, picking the props `apiKey` and `network` from type `AlchemySettings`.


peterpeterparker (approved):
LGTM, thx

PR #5621, opened 2025-04-03
fix(frontend): solve issue with not loading the ICP transactions when scrolling

AntonioVentilii:
# Motivation

When I standardized the transaction ID to be a string for all the networks in PR https://github.com/dfinity/oisy-wallet/pull/5088 , I forgot that there is logic in component `IcTransactionsScroll` that relies on the ID being a `BigInt`.

So, it would never load the next transacitons because there is a condition to exit the function if the ID is not a `BigInt`.

# Changes

- Change the condition to check the ID into a `try` statement that tries to parse the ID into a `BigInt`. If it fails, the function returns.
- Pass the ID as `BigInt` to function `loadNextTransactions`.

# Tests

Deployed in a test canister, it went back to load the next transactions:

https://github.com/user-attachments/assets/87e37185-0ad7-46b2-bcac-beb83c2d556b


src/frontend/src/icp/components/transactions/IcTransactionsScroll.svelte
```diff
@@ -27,7 +27,9 @@
 			return;
 		}
 
-		if (typeof lastId !== 'bigint') {
+		try {
+			BigInt(lastId);
```
DenysKarmazynDFINITY, line 31: Just curious: there can be some non-nullish string value that is not a bigint?

AntonioVentilii: in case of IC transactions, it would be about `Pseudo transactions`.

But according to docs:

there is not such use case in OISY.

> Pseudo transactions are displayed at the end of the list. There is not such use case in Oisy.
> Additionally, if it would be the case, that would mean that we display pseudo transactions at the end of the list and therefore we could assume all valid transactions have been fetched

AntonioVentilii: Sorry! My mistake, good catch! For "self-transactions" we add the suffix `-self` at the end now. Corrected the code.

DenysKarmazynDFINITY: Thanks for the explanation and the fix!

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5626, opened 2025-04-04
feat(ci): make deployment CI callable

AntonioVentilii:
# Motivation

We are trying to solve a [zizmor](https://woodruffw.github.io/zizmor/) issue that is caused by [dangerous triggers](https://woodruffw.github.io/zizmor/audits/#dangerous-triggers): in our case `workflow_run` in deployment CI, because we want to release to Beta automatically when the "Tag and Release" CI runs successfully.

According to [zizmor remediation documentation](https://woodruffw.github.io/zizmor/audits/#remediation_1), it is better to transform the deployment CI in a [resuable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows) and call it inside the "Tag and Release" CI.

# Changes

- Make deployment workflow callable/reusable, including the defaults inputs.
- Adapt the logic to set the correct variables in case of call from workflow_call.
  - NOTE: `github.event_name` will be the same as the caller of this workflow and not `workflow_call`, so we cannot use it. We need to use a [workaround](https://github.com/actions/runner/discussions/1884#discussioncomment-6377587), using a specific optional input variable.
- Call the deployment workflow in "Tag and Release" CI, freezing it to network Beta. 

# Tests

Tested manually and worked: https://github.com/dfinity/oisy-wallet/actions/runs/14260927365


.github/workflows/tag-release.yml
```diff
@@ -43,3 +45,49 @@ jobs:
...
+      VITE_POUH_ENABLED_BETA: ${{ secrets.VITE_POUH_ENABLED_BETA }}
+      VITE_AUTH_ALTERNATIVE_ORIGINS_BETA: ${{ secrets.VITE_AUTH_ALTERNATIVE_ORIGINS_BETA }}
+      VITE_ONRAMPER_API_KEY_DEV_BETA: ${{ secrets.VITE_ONRAMPER_API_KEY_DEV_BETA }}
+      VITE_SWAP_ACTION_ENABLED_BETA: ${{ secrets.VITE_SWAP_ACTION_ENABLED_BETA }}
+      VITE_USER_SNAPSHOT_ENABLED_BETA: ${{ secrets.VITE_USER_SNAPSHOT_ENABLED_BETA }}
+      VITE_AIRDROPS_ENABLED_BETA: ${{ secrets.VITE_AIRDROPS_ENABLED_BETA }}
+      VITE_PLAUSIBLE_ENABLED_BETA: ${{ secrets.VITE_PLAUSIBLE_ENABLED_BETA }}
+      VITE_USER_NETWORKS_FEATURE_ENABLED_BETA: ${{ secrets.VITE_USER_NETWORKS_FEATURE_ENABLED_BETA }}
+      VITE_ONRAMPER_API_KEY_PROD_BETA: ${{ secrets.VITE_ONRAMPER_API_KEY_PROD_BETA }}
+      DFX_DEPLOY_KEY_BETA: ${{ secrets.DFX_DEPLOY_KEY_BETA }}
```
bitdivine, line 93: How are these variables set when a deployment is triggered manually, with `workflow_dispatch`?

AntonioVentilii: I ran it triggering `workflow_dispatch`

https://github.com/dfinity/oisy-wallet/actions/runs/14302847315/job/40080253900

It seems they are set (even if anonymized), for example: `VITE_SWAP_ACTION_ENABLED: ***`

bitdivine (approved):
Makes sense.  I didn't see the secrets being passed in for manual deployments - because that code was unchanged.

PR #5627, opened 2025-04-04
feat(frontend): handle BTC tx multiple recipients case

DenysKarmazynDFINITY:
# Motivation

As suggested by the external audit, we need to properly handle BTC transactions with multiple valid outputs. This case is not reproducible via OISY at the moment, but we might extend the BTC functionality in the future. 

# Changes

1. Instead of using the first valid output to get a value, we calculate a sum from all the valid entries.
2. Instead of using the first valid output to get a recipient address, we collect all valid entries as an array.
3. Update UI to display all the recipients as a list of address. The UI can be improved later, if and when this case becomes real.

# Tests

Existing tests adjusted.

Screenshot 2025-04-03 at 19 45 28

### DIFF
--- a/src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
+++ b/src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
@@ -3,6 +3,7 @@
 	import type { BtcTransactionStatus, BtcTransactionUi } from '$btc/types/btc';
 	import type { BtcTransactionType } from '$btc/types/btc-transaction';
 	import { BTC_MAINNET_EXPLORER_URL, BTC_TESTNET_EXPLORER_URL } from '$env/explorers.env';
+	import TransactionAddress from '$lib/components/transactions/TransactionAddress.svelte';
 	import TransactionModal from '$lib/components/transactions/TransactionModal.svelte';
 	import Value from '$lib/components/ui/Value.svelte';
 	import { i18n } from '$lib/stores/i18n.store';
@@ -13,7 +14,7 @@
 	export let token: OptionToken;
 
 	let from: string;
-	let to: string | undefined;
+	let to: string[] | undefined;
 	let type: BtcTransactionType;
 	let value: bigint | undefined;
 	let timestamp: bigint | undefined;
@@ -34,9 +35,6 @@
 	let txExplorerUrl: string | undefined;
 	$: txExplorerUrl = nonNullish(explorerUrl) ? `${explorerUrl}/tx/${id}` : undefined;
 
-	let toExplorerUrl: string | undefined;
-	$: toExplorerUrl = nonNullish(explorerUrl) ? `${explorerUrl}/address/${to}` : undefined;
-
 	let fromExplorerUrl: string | undefined;
 	$: fromExplorerUrl =
 		nonNullish(explorerUrl) && nonNullish(to) ? `${explorerUrl}/address/${from}` : undefined;
@@ -44,12 +42,10 @@
 
 <TransactionModal
 	commonData={{
-		to,
 		from,
 		timestamp,
 		blockNumber,
 		txExplorerUrl,
-		toExplorerUrl,
 		fromExplorerUrl
 	}}
 	hash={id}
@@ -71,4 +67,24 @@
 		<svelte:fragment slot="label">{$i18n.transaction.text.status}</svelte:fragment>
 		{`${$i18n.transaction.status[status]}`}
 	</Value>
+
+	<svelte:fragment slot="transaction-custom-to">
+		{#if nonNullish(to)}
+			<Value ref="recipients">
+				<svelte:fragment slot="label">{$i18n.transaction.text.to}</svelte:fragment>
+
+				<ul class="list-none">
+					{#each to as address, index (`${address}-${index}`)}
+						<li>
+							<TransactionAddress
+								{address}
+								explorerUrl={`${explorerUrl}/address/${address}`}
+								copiedText={$i18n.transaction.text.to_copied}
+							/>
+						</li>
+					{/each}
+				</ul>
+			</Value>
+		{/if}
+	</svelte:fragment>
 </TransactionModal>
--- a/src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
+++ b/src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
@@ -78,7 +78,9 @@
 						<li>
 							<TransactionAddress
 								{address}
-								explorerUrl={`${explorerUrl}/address/${address}`}
+								explorerUrl={nonNullish(explorerUrl)
+									? `${explorerUrl}/address/${address}`
+									: undefined}
 								copiedText={$i18n.transaction.text.to_copied}
 							/>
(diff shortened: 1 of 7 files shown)


src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
```diff
@@ -71,4 +67,24 @@
 		<svelte:fragment slot="label">{$i18n.transaction.text.status}</svelte:fragment>
 		{`${$i18n.transaction.status[status]}`}
 	</Value>
+
+	<svelte:fragment slot="transaction-custom-to">
```
AntonioVentilii, line 71: can this `slot` goes directly with the child `<Value>` or do we have strange behaviours?

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
```diff
@@ -71,4 +67,24 @@
...
+		{#if nonNullish(to)}
+			<Value ref="recipients">
+				<svelte:fragment slot="label">{$i18n.transaction.text.to}</svelte:fragment>
+
+				<ul class="list-none">
+					{#each to as address, index (`${address}-${index}`)}
+						<li>
+							<TransactionAddress
+								{address}
+								explorerUrl={`${explorerUrl}/address/${address}`}
```
AntonioVentilii, line 81: should there be a nullish check here? `explorerUrl` can be `undefined` for what I see

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -28,33 +28,28 @@ export const mapBtcTransaction = ({
...
 				(isTypeSend && addr !== btcAddress) || (!isTypeSend && addr === btcAddress);
 
-			if (isNullish(acc.value) && isValidOutput) {
-				acc.value = (acc.value ?? 0) + value;
-			}
-
-			if (isNullish(acc.to) && isValidOutput) {
-				acc.to = addr;
-			}
+			totalValue = isValidOutput ? (totalValue ?? 0) + value : totalValue;
```
AntonioVentilii, line 42: why not putting `totalValue` inside the accumulator too?

```typescript
const { totalOutputValue, to, totalValue } = out.reduce<{
		to: string[];
		totalOutputValue: number;
		totalValue: number | undefined;
	}>(...
```

src/frontend/src/lib/services/user-snapshot.services.ts
```diff
@@ -132,7 +132,8 @@ const toSplTransaction = ({
 			value: BigNumber.from(value ?? ZERO_BI).toBigInt(),
 			timestamp: BigInt(timestamp ?? ZERO_BI)
 		}),
-		counterparty: address === from ? to : from
+		// in case it's a BTC tx, "to" is an array
+		counterparty: address === from ? (Array.isArray(to) ? to[0] : to) : from
```
AntonioVentilii, line 136: just for my understanding, it is ok to just give the first address? or better to pass some sort of string  that are the addresses separated with comma, boh..

in case, can you just add a TODO here and we will tackle in another PR?

src/frontend/src/tests/mocks/btc-transactions.mock.ts
```diff
@@ -7,7 +7,7 @@ export const mockBtcTransactionUi: BtcTransactionUi = {
 	id: 'e793cab7e155a0e8f825c4609548faf759c57715fecac587580a1d716bb2b89e',
 	status: 'confirmed',
 	timestamp: 1727175987n,
-	to: 'bc1qt0nkp96r7p95xfacyp98pww2eu64yzuf78l4a2wy0sttt83hux4q6u2nl7',
+	to: ['bc1qt0nkp96r7p95xfacyp98pww2eu64yzuf78l4a2wy0sttt83hux4q6u2nl7'],
```
AntonioVentilii, line 10: would it be possible to create a test (or more) to check what happens with multiple addresses?

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
DenysKarmazynDFINITY: We cannot move it to Value because the slot cannot be rendered conditionally - there is an `if` statement that decides whether the Value should be displayed or not

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
DenysKarmazynDFINITY: you're right! adding

src/frontend/src/lib/services/user-snapshot.services.ts
DenysKarmazynDFINITY: It's ok to do it now, since OISY does not support multiple recipients for BTC. Since this user-snapshot method is also marked as temporary, I think we can have it like this for now. When we improve the BTC send and/or refactor the user-snapshot approach, we can handle multiple recipients case.

src/frontend/src/tests/mocks/btc-transactions.mock.ts
DenysKarmazynDFINITY: The tests with single and multiple recipients are actually already there (btc-transactions.utils.spec.ts) - you can see that in some of the test suites we're overriding the `to` value of this mock object. Though, if you think we need more cases, I can definitely add them.

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
AntonioVentilii: thanks for checking!

src/frontend/src/btc/utils/btc-transactions.utils.ts
AntonioVentilii: indeed, can you not do `acc.totalValue = acc.totalValue + ....` ? Or the value is assigned outside the `reduce` statement too?

src/frontend/src/tests/mocks/btc-transactions.mock.ts
AntonioVentilii: perfect!

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: you're totally right, done!

src/frontend/src/btc/utils/btc-transactions.utils.ts
```diff
@@ -28,33 +28,29 @@ export const mapBtcTransaction = ({
 		}
 	);
 
-	const { totalOutputValue, value, to } = out.reduce<{
+	// let totalValue: number | undefined = undefined;
```
AntonioVentilii, line 31: some leftover?
```suggestion
```

AntonioVentilii (approved):
Thank you!

src/frontend/src/btc/utils/btc-transactions.utils.ts
DenysKarmazynDFINITY: oops, removed

PR #5628, opened 2025-04-04
feat(frontend): added try/catch for plausible init

roman-nazaruk:
# Motivation

# Changes

Added try-catch block to prevent app initialization from breaking if Plausible fails to load

# Tests

Covered new logic with the tests.

### DIFF
--- a/src/frontend/src/lib/services/analytics.services.ts
+++ b/src/frontend/src/lib/services/analytics.services.ts
@@ -45,7 +45,7 @@ export const initPlausibleAnalytics = () => {
 			});
 			plausibleTracker.enableAutoPageviews();
 		}
-	} catch (_err) {
+	} catch (_err: unknown) {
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/services/analytics.services.ts
```diff
@@ -36,13 +36,17 @@ export const initPlausibleAnalytics = () => {
...
+	try {
+		if (isNullish(plausibleTracker)) {
+			plausibleTracker = Plausible({
+				domain: PLAUSIBLE_DOMAIN,
+				hashMode: false,
+				trackLocalhost: false
+			});
+			plausibleTracker.enableAutoPageviews();
+		}
+	} catch (_err) {
```
AntonioVentilii, line 48:
```suggestion
	} catch (_err: unknown) {
```

AntonioVentilii (approved):
LGTM tks! Small change

PR #5630, opened 2025-04-04
chore(frontend): explicit type in test for Etherscan rest

AntonioVentilii:
# Motivation

To help us with the explicit Ethereum transaction types, we set it in the test for Etherscan rest.


daviddecentage (approved):
LGTM, thx!

PR #5634, opened 2025-04-04
refactor(frontend): explicitly set props of Ethers transaction interface

AntonioVentilii:
# Motivation

Furthermore we explicitly specify props `hash`, `from` and `to` in interface `EthersTransaction`, and props `blockNumber` and `timestamp` in interface `Transaction`.

It does not change anything, since the new definitions are the same as the imbedded ones. However, in `ethers` v6 they will change to be null instead of undefined, so this helps us with maintaining the behaviours during migration.


daviddecentage (approved):
LGTM, thx!

peterpeterparker (approved):
LGTM, thx

PR #5635, opened 2025-04-04
feat(frontend): use BigInt instead of BigNumber in Ethers transaction interace

AntonioVentilii:
# Motivation

Unfortunately, `alchemy-sdk` is still not updated to the newest version of `ethers` (issue https://github.com/alchemyplatform/alchemy-sdk-js/issues/489). That means that we have to adapt one to the other, until they bump it.

So, we set explicitly the types of `ethers` and adjust them in a mapper inside the `AlchemyProvider` class.

# Changes

- Explicitly set props `gasLimit`, `gasPrice` and chainId in interface `EthersTransaction`.
- Create mapper in class `AlchemyProvider`.
- Prepare type `TransactionResponseWithBigInt` to use `ethers` props instead of `alchemy-sdk`.

# Tests

Adapted tests.

### DIFF
--- a/src/frontend/src/eth/providers/alchemy.providers.ts
+++ b/src/frontend/src/eth/providers/alchemy.providers.ts
@@ -120,9 +120,15 @@ export class AlchemyProvider {
 			return transaction;
 		}
 
-		const { value, ...rest } = transaction;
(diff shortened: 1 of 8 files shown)


daviddecentage (approved):
LGTM, thx!

PR #5637, opened 2025-04-04
feat(frontend): do not update fees on ETH/ERC20 convert review step

DenysKarmazynDFINITY:
# Motivation

It was decided to stop updating fees when user is already on the conversion review step. To do that for ETH and ERC20, a few steps are required.

# Changes

1. Move fee context initialization from EthConvertTokenWizard to respective parent component (same as it's done for BTC) - ConvertToCkERC20 and ConvertToCkETH. The reason is that the wizard is being rerendered (and therefore fee context is reset) on each step change.
2. Stop observing when on the REVIEW step.
3. Update FeeContext to not trigger `updateFee` method on 

# Tests

Existing tests are updated. Also, tested both flows manually.

### DIFF
--- a/src/frontend/src/eth/components/convert/ConvertToCkETH.svelte
+++ b/src/frontend/src/eth/components/convert/ConvertToCkETH.svelte
@@ -1,50 +1,16 @@
 <script lang="ts">
 	import { nonNullish } from '@dfinity/utils';
-	import { setContext } from 'svelte';
-	import { writable } from 'svelte/store';
 	import { ETHEREUM_TOKEN } from '$env/tokens/tokens.eth.env';
+	import FeeStoreContext from '$eth/components/fee/FeeStoreContext.svelte';
 	import { ethereumToken, ethereumTokenId } from '$eth/derived/token.derived';
-	import {
-		FEE_CONTEXT_KEY,
-		type FeeContext as FeeContextType,
-		initFeeContext,
-		initFeeStore
-	} from '$eth/stores/fee.store';
 	import type { IcCkToken } from '$icp/types/ic-token';
 	import ConvertETH from '$icp-eth/components/convert/ConvertETH.svelte';
 	import ConvertModal from '$lib/components/convert/ConvertModal.svelte';
 	import IconCkConvert from '$lib/components/icons/IconCkConvert.svelte';
-	import { exchanges } from '$lib/derived/exchange.derived';
 	import { modalConvertToTwinTokenCkEth } from '$lib/derived/modal.derived';
 	import { tokens } from '$lib/derived/tokens.derived';
 	import { i18n } from '$lib/stores/i18n.store';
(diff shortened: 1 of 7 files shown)


AntonioVentilii (commented):
I am not really sure this is ok... We really should update them before approving the conversion, otherwise either the transaction may fail or the users pay different fees from what they see, no?

DenysKarmazynDFINITY:
@AntonioVentilii totally a valid point! However, the "Review" step kinda implies that it's only for confirming the values shown during the first step. If we want to allow fee updates on the overview step, then we’d also need to revalidate the inputted send/convert amount and display all the errors and warnings from the initial form (currently, not the case) - which would make the "Review" step redundant.

AntonioVentilii:
@DenysKarmazynDFINITY yes, it makes sense. My worry is if we increase the number of possible failed transactions, but I imagine we can check afterwards and add some margin to the fees

src/frontend/src/eth/components/fee/FeeContext.svelte
```diff
@@ -151,7 +153,10 @@
 
 	$: obverseFeeData(observe);
 
-	$: $ckEthMinterInfoStore, debounceUpdateFeeData();
+	$: $ckEthMinterInfoStore,
+		() => {
+			observe && debounceUpdateFeeData();
+		};
```
AntonioVentilii, line 159: i think you have to wrap it here to make it execute:

```typescript
$: $ckEthMinterInfoStore,
		(() => {
			observe && debounceUpdateFeeData();
		})();
```

src/frontend/src/eth/components/convert/ConvertToCkETH.svelte
```diff
@@ -1,16 +1,51 @@
...
+		FEE_CONTEXT_KEY,
+		initFeeContext({
+			feeStore,
+			feeSymbolStore,
+			feeTokenIdStore,
+			feeDecimalsStore,
+			feeExchangeRateStore
+		})
+	);
+
```
AntonioVentilii, line 48: to avoid repetition of code, does it make sense to create a component that does the initialization (like we do for example for `<SendTokenContext>`) and reuse it as wrapper?

src/frontend/src/eth/components/fee/FeeContext.svelte
DenysKarmazynDFINITY: Good catch, updated!

src/frontend/src/eth/components/convert/ConvertToCkETH.svelte
DenysKarmazynDFINITY: Great idea, updated. The new `FeeContextStore` expects to get `token` as a prop so we can re-use it in other places (added to my todo list to refactor them).

AntonioVentilii (approved):
LGTM tks!

PR #5639, opened 2025-04-04
refactor(frontend): use BigInt() instead of BigNumber.from().toBigInt()

AntonioVentilii:
# Motivation

It is cleaner to use directly `BigInt()` instead of `BigNumber.from().toBigInt()`


BonomoAlessandro (approved):
LGTM, thx

PR #5640, opened 2025-04-04
refactor(frontend): use Number() of Zero BigInt instead of Zero BigNumber.toNumber()

AntonioVentilii:
# Motivation

We are deprecating BigNumber, so it is better to use `Number(ZERO_BI)` instead of `ZERO.toNumber()`.


BonomoAlessandro (approved):
LGTM, thx

PR #5641, opened 2025-04-04
feat(scripts): remove unnecessary script format-imports

AntonioVentilii:
# Motivation

Script `format-imports` is not needed anymore (it was used to slowly refactor the svelte imports). The rule it was enforcing is already in the library [@dfinity/eslint-config-oisy-wallet](https://github.com/dfinity/eslint-config-oisy-wallet) that we use as linter.

So, we remove it and add `eslint --fix .` to the `format` script.


BonomoAlessandro (approved):
LGTM, thx

PR #5644, opened 2025-04-04
fix(frontend): Wallet Connect gas is returned as hex string

AntonioVentilii:
# Motivation

In PR https://github.com/dfinity/oisy-wallet/pull/5381, I forgot that the `gas` value returned by Wallet Connect is an hex string. So we need to convert it to a `BigInt`.


BonomoAlessandro (approved):
LGTM

PR #5646, opened 2025-04-05
refactor(frontend): SPL User tokens --> SPL Custom tokens

AntonioVentilii:
# Motivation

Technically speaking, it is not correct to call the SPL tokens that are saved in the backend as "User tokens": we are using the methods that are used for Custom tokens (for example the ones already used for ICRC).

So we rename everything related to "SPL User tokens" to "SPL Custom tokens".


BonomoAlessandro (approved):
LGTM, thx

PR #5649, opened 2025-04-06
feat(frontend): remove deprecated function that saves cached SPL tokens to backend

AntonioVentilii:
# Motivation

Enough time has passed to suppose that active users already logged in and that OISY saved the cached SPL tokens in the backend. So we can remove the temporary solution that was saving them in the backend.


BonomoAlessandro (approved):
LGTM

PR #5650, opened 2025-04-06
feat(devops): enforce non-empty pull requests

AntonioVentilii:
# Motivation

It happened in the past that we merged empty PRs: for example PR2 is a sub-set of PR1, but PR1 is merged first. So once PR2 is rebased, it has no file changed.

We would like to avoid such situations.


.github/workflows/pr-checks.yml
```diff
@@ -9,9 +9,12 @@ on:
       - synchronize
       - labeled
 
+permissions:
+  pull-requests: read
+  contents: read
+
```
AntonioVentilii, line 15:
```suggestion
permissions: {}

```

BonomoAlessandro (approved):
LGTM

PR #5651, opened 2025-04-06
feat(devops): set default permissions to none for all workflows

AntonioVentilii:
# Motivation

One of [zizmor security suggestions](https://woodruffw.github.io/zizmor/audits/#excessive-permissions) is to set the workflow permissions to a generic value (possibly all disabled) and set the permissions individually for each job.

So, we set them all to no permissions, and solve the permissions for each single job


bitdivine (approved):
Thanks!

PR #5653, opened 2025-04-06
refactor(devops): remove possible attacker-controllable code in CI to update next

AntonioVentilii:
# Motivation

One of [zizmor security suggestions](https://woodruffw.github.io/zizmor/audits/#template-injection) is to disallow possible injections via template expansions. So we remove them in CI `update-next`, in favour of env variables.


.github/workflows/update-next.yml
```diff
@@ -56,11 +56,15 @@ jobs:
             *) echo "Invalid package selection: $PACKAGE" && exit 1 ;;
           esac
           echo "PKG_NAME=$PKG_NAME" >> "$GITHUB_ENV"
+          echo "PKG_TAG=$NEXT_TAG" >> "$GITHUB_ENV"
```
bitdivine, line 59: I suspect that zizmor would prefer OUTPUT over ENV but I think either is fine in this case.

bitdivine (approved):

.github/workflows/update-next.yml
AntonioVentilii: Yes yes, you are completly right, but I wanted to tackle that entire step in another PR

PR #5663, opened 2025-04-07
feat(frontend): Tokenlist filter

daviddecentage:
# Motivation

# Changes

# Tests


AntonioVentilii (commented):
May you please split this PR? Maybe smaller scopes

src/frontend/src/lib/components/tokens/TokenGroupCard.svelte
```diff
@@ -37,12 +39,17 @@
...
-		);
+	$: filteredTokens = getFilteredTokenGroup({
+		filter: $tokenListStore.filter,
+		list: tokenGroup.tokens
+	});
+
+	// list of tokens that should display with a "show more" button for not displayed ones
+	let truncatedTokens: TokenUi[];
+	$: truncatedTokens = filteredTokens.filter((token) => {
+		const totalBalance = filteredTokens.reduce((p, c) => p + BigInt(c.balance ?? 0n), BigInt(0n));
```
AntonioVentilii, line 52: since we are here

```suggestion
		const totalBalance = filteredTokens.reduce((p, c) => p + BigInt(c.balance ?? 0n), ZERO_BI);
```

src/frontend/src/lib/components/tokens/TokensFilter.svelte
```diff
@@ -12,7 +12,7 @@
 	let {
 		inputValue = $bindable($tokenListStore.filter),
 		overflowableContent
-	}: { inputValue: string; overflowableContent?: Snippet } = $props();
+	}: { inputValue?: string; overflowableContent?: Snippet } = $props();
```
AntonioVentilii: just wondering, are you planning of using it as bind or prop in another PR? if no, it can be removed from the props (in another PR), no?

AntonioVentilii: ah ok, i think you already did it while i was writing eheh

AntonioVentilii (approved):
Small comment, LGTM! tks

PR #5674, opened 2025-04-07
fix(frontend): updated logic of loading resources on App init

roman-nazaruk:
# Motivation

Improve app stability by initializing all core services in parallel, without stopping the app if one of them fails.

# Changes

Replaced Promise.all with Promise.allSettled in the init() function to avoid early termination when non-critical services fail.

# Tests

Verified that all core services (authStore.sync, initAnalytics, initPlausibleAnalytics, i18n.init) are called once on app load.


src/frontend/src/tests/routes/layout.spec.ts
```diff
@@ -1,6 +1,9 @@
 import * as analytics from '$lib/services/analytics.services';
+import { authStore } from '$lib/stores/auth.store';
+import { i18n } from '$lib/stores/i18n.store';
+import { toastsError } from '$lib/stores/toasts.store';
 import App from '$routes/+layout.svelte';
-import { render } from '@testing-library/svelte';
+import { render, waitFor } from '@testing-library/svelte';
 import { beforeAll, describe, expect, it, vi } from 'vitest';
```
AntonioVentilii, line 7: since we are here

```suggestion
```

src/frontend/src/routes/+layout.svelte
```diff
@@ -28,8 +28,23 @@
...
+			 * Initialize all core services required for the app.
+			 *
+			 * These services are currently independent and can be initialized in parallel
+			 * to reduce startup time. If any future dependency is introduced between them,
+			 * consider switching to sequential loading.
+			 */
+			await Promise.all([syncAuthStore(), initAnalytics(), initPlausibleAnalytics(), i18n.init()]);
+		} catch (err: unknown) {
+			toastsError({
+				msg: { text: $i18n.auth.error.unexpected_issue_with_syncing },
```
AntonioVentilii, line 43: maybe let's show something more generic like `unexpected_issue_with_initializing` ? or something similar

AntonioVentilii (approved):
As discussed, we do not logout if any of the services fails.


LGTM tks! Two observations!

src/frontend/src/routes/+layout.svelte
peterpeterparker: - `syncAuthStore` already shows a toast on error, so this might result in two toasts.
- Also, I'm not sure initializing authentication should be treated with the same level of importance as initializing analytics—so it may not make sense to show the same error.
- Not sure why the user should even be notified if analytics fail to initialize btw.?
- And users with plugins that block analytics might always see an error toast when accessing OISY?
- If i18n fails to initialize, the user will likely notice. And if it's not initialized, what would the error message even say if there is no text available?
- Finally, the error message wouldn't be generic enough.

peterpeterparker (changes requested):
Just blocking because Antonio already approved it and I'm unsure about this..l

peterpeterparker:
As discussed with @AntonioVentilii maybe we should not catch but use `allSettled` instead of `all` on init

src/frontend/src/routes/+layout.svelte
roman-nazaruk: I agree, and have updated the PR to replace Promise.all with Promise.allSettled to ensure all services initialize in parallel without blocking the app if any of them fail.

peterpeterparker: Sounds. Does the description of the PR needs to be updated?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/routes/+layout.svelte
peterpeterparker: Thanks!

PR #5678, opened 2025-04-07
refactor(devops): pin softprops/action-gh-release to specific version

AntonioVentilii:
# Motivation

One of [zizmor security suggestions](https://woodruffw.github.io/zizmor/audits/#unpinned-uses) is to pin the third party actions to a specific commit, to avoid possible liabilities. So we pin [softprops/action-gh-release to version v2.2.1](https://github.com/softprops/action-gh-release/tree/v2.2.1).


BonomoAlessandro (approved):
LGTM

PR #5679, opened 2025-04-07
refactor(devops): pin dorny/paths-filter to specific version

AntonioVentilii:
# Motivation

One of [zizmor security suggestions](https://woodruffw.github.io/zizmor/audits/#unpinned-uses) is to pin the third party actions to a specific commit, to avoid possible liabilities. So we pin [dorny/paths-filter to version v3.0.2](https://github.com/dorny/paths-filter/tree/v3.0.2).


BonomoAlessandro (approved):
LGTM

PR #5680, opened 2025-04-07
refactor(devops): pin actions/setup-node to specific version

AntonioVentilii:
# Motivation

One of [zizmor security suggestions](https://woodruffw.github.io/zizmor/audits/#unpinned-uses) is to pin the third party actions to a specific commit, to avoid possible liabilities. So we pin [actions/setup-node to version v4.3.0](https://github.com/actions/setup-node/tree/v4.3.0).


BonomoAlessandro (approved):
LGTM

PR #5681, opened 2025-04-07
refactor(devops): pin release-drafter/release-drafter to specific version

AntonioVentilii:
# Motivation

One of [zizmor security suggestions](https://woodruffw.github.io/zizmor/audits/#unpinned-uses) is to pin the third party actions to a specific commit, to avoid possible liabilities. So we pin [release-drafter/release-drafter to version v6.1.0](https://github.com/release-drafter/release-drafter/tree/v6.1.0).


BonomoAlessandro (approved):
LGTM

PR #5683, opened 2025-04-07
refactor(devops): pin peter-evans/create-pull-request to specific version

AntonioVentilii:
# Motivation

One of [zizmor security suggestions](https://woodruffw.github.io/zizmor/audits/#unpinned-uses) is to pin the third party actions to a specific commit, to avoid possible liabilities. So we pin [peter-evans/create-pull-request to version v6.0.5](https://github.com/peter-evans/create-pull-request/tree/v6.0.5).


BonomoAlessandro (approved):
LGTM

PR #5687, opened 2025-04-07
fix(frontend): Do not parse environment variables

FlavioInacta:
# Motivation

The application should not crash when an environment variable is set to empty string.

# Changes

Compare all boolean environment variables to 'true' instead of parsing the content.

# Tests

FE still working.


FlavioInacta:
Tests do not seem to work. Too much effort I guess for the change.

PR #5688, opened 2025-04-07
feat(e2e): disable rewards modal test for no campaign

AntonioVentilii:
# Motivation

The E2E test for the rewards modal shall be stopped for now, since it assumes that there will always be an ongoing campaign. Since that is not always the case, we skip it and come back to improve it with a mock.


BonomoAlessandro (approved):
LGTM thx

AntonioVentilii (commented):
Edit

PR #5694, opened 2025-04-07
feat(frontend): Adds menuentry for address book

FlavioInacta:
# Motivation

User can navigate to the address book.

# Changes

Adds a menuentry to navigate to the address book

# Tests

- Tested if modal is working


DecentAgeCoder (changes requested):
Blocking this PR until dependent PR's have been approved to avoid accidental merge

PR #5697, opened 2025-04-07
feat(frontend): Adds address book component

FlavioInacta:
# Motivation

Adds a simple address book component with empty address book as a first implementation of the address book

# Changes

Adds address book component


daviddecentage (changes requested):
Blocking this PR: I would not just have an empty state modal, instead we could already initialize an empty list of contacts and display the empty state if the array is empty. Also add a todo comment which describes whats missing (just an example):

```
let contactList = []; // todo: get list from backend
// ...
{#if contactList.length === 0}
	//svg and empty state text
{:else}
	// todo: add not empty state UI where we can manage the contacts etc etc
{/if}
```

FlavioInacta:
@daviddecentage What's the point? Please provide a reason why I should add an empty array with an else statement that will never be reached.

It is work in progress, the feature is clearly not finished. I would like to add as minimal code as possible in order to not add useless code that has to be refactored later.

FlavioInacta:
Since there should be only finished thing go to main, I close this PR and will reopen PRs when functionlity (with design) are completely finished.

PR #5704, opened 2025-04-08
refactor(devops): change action prepare-macos to avoid possible code execution

AntonioVentilii:
# Motivation

One of [zizmor security suggestions](https://woodruffw.github.io/zizmor/audits/#github-env) is to avoid dangerous writes to the `GITHUB_ENV` and `GITHUB_PATH` environment variables.

When used in workflows with dangerous triggers (such as pull_request_target and workflow_run), `GITHUB_ENV` and `GITHUB_PATH` can be an arbitrary code execution risk.

So, we refactor action `prepare-macos` to avoid possible code injections.


bitdivine (approved):
I have to say, I think Zizmor is wrong about this.  But there is no harm in expressing the same code in a different way that has the same risks but that doesn't get Zizmor worried.

PR #5705, opened 2025-04-08
build(frontend): bump sveltekit and vitest

peterpeterparker:
# Motivation

There is yet again a new Vite server security disclosure ([CVE-2025-31486](https://github.com/advisories/GHSA-xcj6-pq6g-qj4x)). We are not affected because we do not use the Vite server and also the library is already up-to-date in the repo. Nevertheless, I go through all repo to update related dependencies and as a result this PR bumps SvelteKit and Vitest.

Nothing critical, just because I'm doing some maintenance.


AntonioVentilii (approved):
Tks for the fix!

PR #5708, opened 2025-04-08
refactor(frontend): IcWalletTransactionsScheduler --> IcWalletBalanceAndTransactionsScheduler

AntonioVentilii:
# Motivation

Technically speaking, the scheduler/worker for the ICRC wallets fetches both transactions and balance. Since we are going to change the logic a bit to have different functions internally, one for transactions and one for balance, it seems correct to rename `IcWalletTransactionsScheduler` to `IcWalletBalanceAndTransactionsScheduler`, even if a little verbose.


peterpeterparker (approved):
LGTM, thx

PR #5713, opened 2025-04-08
fix(frontend): Fix menu width with long texts

daviddecentage:
# Motivation

We fix the menu width of the top right menu when long texts are inside. 

Before we used 300px max-width on the popover wrapper SCSS, which applies it to all popovers. When we removed it for other popovers it lead to issues with this specific one. To achieve the same look we apply 272px max-width to the content wrapper, so it will be the same with all paddings etc as 300px on the popover wrapper.

# Changes

Added max width to menu content wrapper

# Tests

Before:
image

After:
image

### DIFF
--- a/src/frontend/src/lib/components/core/Menu.svelte
+++ b/src/frontend/src/lib/components/core/Menu.svelte
@@ -79,7 +79,7 @@
 </ButtonIcon>
 
 <Popover bind:visible anchor={button} direction="rtl" on:click={hidePopover}>
-	<div class="flex max-w-[300px] flex-col gap-1" data-tid={NAVIGATION_MENU}>
+	<div class="flex max-w-[272px] flex-col gap-1" data-tid={NAVIGATION_MENU}>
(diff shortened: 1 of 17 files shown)


src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -79,7 +79,7 @@
 </ButtonIcon>
 
 <Popover bind:visible anchor={button} direction="rtl" on:click={hidePopover}>
-	<div class="flex flex-col gap-1" data-tid={NAVIGATION_MENU}>
+	<div class="flex max-w-[272px] flex-col gap-1" data-tid={NAVIGATION_MENU}>
```
BonomoAlessandro, line 82: Wouldn't `max-w-68` be better?

daviddecentage: Sure

BonomoAlessandro (approved):
LGTM

PR #5714, opened 2025-04-08
fix(frontend): Network name in token group

daviddecentage:
# Motivation

In the homepage token list, grouped tokens should display the network name, not the token name since the group itself is displaying it already. Exception is the ck twin tokens, which should display the symbol.

# Changes

If token card displayed as network (`!isCkToken` condition), render network name instead of symbol.

# Tests

image


BonomoAlessandro (approved):
LGTM

PR #5728, opened 2025-04-09
feat(devops): deploy backend always with yes-flag

AntonioVentilii:
# Motivation

When the backend is auto-deployed, sometimes it asks to confirm for breaking changes:

> Upgrading code for canister backend, with canister ID d3nvo-aaaaa-aaaar-qagzq-cai
> WARNING!
> Candid interface compatibility check failed for canister 'backend'.
> You are making a BREAKING change. Other canisters or frontend clients relying on your canister may stop working.
> 
> Method allow_signing: func (opt AllowSigningRequest) -> (Result_2) is not a subtype of func () -> (Result_2/1)
> Do you want to proceed? yes/No

However, since it is a CI, nobody answers and it defaults to "No" (see https://github.com/dfinity/oisy-wallet/actions/runs/14339313922/job/40194404634).

So, instead of using the `--yes` flag only when we force the backend deployment, we use it always when we deploy it.


peterpeterparker:
Wouldn't it be safer not to proceed here? Otherwise, what safeguard is in place to detect preventively breaking changes when deploying to production?

bitdivine:
In the rewards canister, I compare the current candid file with production.  Maybe we can do something similar: https://github.com/dfinity/oisy-rewards-canister/blob/airdrop/src/oisy_rewards_canister/src/tests.rs#L38

A caveat is that sometimes the candid compatibility check leaks motoko concepts, and motoko compatibility is stricter than Candid compatibility, so I do get false positives sometimes, which is annoying.

bitdivine:
Maybe the important thing is to make it very visible if an API change is going to break in production, but to make this non-blocking.  

I think an issue in the current CI is that the changes are not obvious.  Staging just doesn't get updated, and that can be misleading.  As such, I think this PR is good, but I believe that we need an explicit check for compatibility with prod.  WDYT?

bitdivine (approved):

peterpeterparker:
> Maybe the important thing is to make it very visible if an API change is going to break in production

Agreed, that's my concern too. Right now, I feel like this check - deploying to staging - is the one that indicates it. The suggestions you shared seem interesting 👍.

AntonioVentilii:
@peterpeterparker @inc-man 

As first generic measure WDYT of blocking the deployment if the network is prod?

- https://github.com/dfinity/oisy-wallet/pull/5734

AntonioVentilii:
As discussed here and offline, I will not merge this CI for the moment. We will work on a solution that at least warns us about breaking changes

AntonioVentilii:
As discussed here and offline, I will not merge this CI for the moment. We will work on a solution that at least warns us about breaking changes.

Closing this PR.

PR #5732, opened 2025-04-09
feat(frontend): implements help form

BonomoAlessandro:
# Motivation

There should be a modal that helps the user to find the best solution for their problem with the login process.

# Changes

- implements help form

# Tests

image


daviddecentage (approved):
LGTM

PR #5734, opened 2025-04-09
feat(devops): create guard for deployments in prod

AntonioVentilii:
# Motivation

As safe measure (thread in PR https://github.com/dfinity/oisy-wallet/pull/5728), we want to avoid the deployment job to run for production environment.

### DIFF
--- a/.github/workflows/deploy-to-environment.yml
+++ b/.github/workflows/deploy-to-environment.yml
@@ -59,10 +59,11 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}-${{ github.event.inputs.network }}-${{ github.event.inputs.canister }}
   cancel-in-progress: false
 
-permissions: {}
+permissions: { }
 
 jobs:
   deployment:
+    if: "${{ github.event.inputs.network != 'ic' && github.event.inputs.network != 'prod' && inputs.network != 'ic' && inputs.network != 'prod' }}"
     runs-on: ubuntu-24.04
     permissions:
       contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-to-environment.yml
```diff
@@ -63,6 +63,7 @@ permissions: {}
 
 jobs:
   deployment:
+    if: "${{ github.event.inputs.network != 'ic' && github.event.inputs.network != 'prod' && inputs.network != 'ic' && inputs.network != 'prod' }}"
```
bitdivine, line 66: These situations cannot occur, as `ic` and `prod` are not in the list of values for network at the top of this file.

I even tested whether that list can be bypassed with the cli:
```
$ gh workflow run deploy-to-environment.yml -f network=bonkers
could not create workflow dispatch event: HTTP 422: Provided value 'bonkers' for input 'network' not in the list of allowed values (https://api.github.com/repos/dfinity/oisy-wallet/actions/workflows/122410405/dispatches)
```

AntonioVentilii: theoretically it could happen with

```yml
  workflow_call:
    inputs:
      workflow-call:
        description: 'To distinguish workflow_call from regular calls'
        required: false
        type: boolean
        default: true
      network:
        required: true
        type: string
      canister:
        required: false
        type: string
        default: frontend
```
        
since we cannot give a list here (AFAIK)

or you mean that this should simply be...?

if: "${{ inputs.network != 'ic' && inputs.network != 'prod' }}"

(that takes the inputs ONLT of `workflow_call`)

AntonioVentilii:
as discussed offline, I will refactor this workflow to be an action, with specific inputs, so that we remove workflow call

AntonioVentilii:
as discussed offline, I will refactor this workflow to be an action, with specific inputs, so that we remove workflow call

PR #5736, opened 2025-04-09
chore(devops): add merge_group trigger to all relevant workflows

AntonioVentilii:
# Motivation

Included `merge_group` trigger to all the workflows that are triggered by the `pull_request` (most of them were already set correctly).


bitdivine (approved):
Cool!  Shall we enable merge queues right after the release?

PR #5741, opened 2025-04-09
fix(frontend): Fix gray text in logo button

daviddecentage:
# Motivation
We need to fix gray texts for titles in the `LogoButton` component.

# Changes
Add classes correctly for applying text color

# Tests

Before:
image

After:
image


AntonioVentilii (approved):
Tks for the fix!

PR #5743, opened 2025-04-09
fix(e2e): increase threshold for screenshots to reduce icon rendering flakiness

AntonioVentilii:
# Motivation

When we run the E2E tests, sometimes, some emojis are rendered with different pixel ratio, sometimes blurry sometimes not, creating flakiness. So, we try to fix it with a slightly higher threshold as per suggestion in issue https://github.com/microsoft/playwright/issues/8161#issuecomment-1913353418


BonomoAlessandro (approved):
LGTM, thx

PR #5749, opened 2025-04-09
ci(e2e): playwright config for e2e-environment

inc-man:
# Motivation

Since the e2e environment is running we should configure tests to run against it and not locally.

# Changes

Adjusted the playwright config and first tests

# Tests


AntonioVentilii:
Stale Playwright e2e-environment config from April 2025; superseded by subsequent e2e config changes in main. Closing.

PR #5754, opened 2025-04-09
feat(frontend): show Index canister update status in `IcTransactions`

AntonioVentilii:
# Motivation

The balance of an ICRC token is taken from the Ledger canister, while the transactions are taken from the Index canister. That means that there could some lag between the two informations.

So, to be more user-friendly, we want to show a status message for pending syncing of the index canister (very similar to what we do with the ck-canisters).

# Changes

- Create event to define an out-of-sync Index canister, based on discordant balance with the Ledger canister.
- Emit such event in the ICRC worker, when fetching balance and transactions for ICRC tokens.
- Create component `IcIndexCanisterStatus` that catches the new event and show a message accordingly.
- Wrap existing status components with the new `IcIndexCanisterStatus`.

# Tests

Created tests.


roman-nazaruk (approved):
LGTM, thnx

PR #5758, opened 2025-04-09
feat(frontend): load internal transactions in Etherscan provider

AntonioVentilii:
# Motivation

When an user bridges an Ethereum-based token from an EVM network to another (for example USDC-Ethereum to USDC-Base), it creates what is defined as an ["internal" transaction](https://docs.etherscan.io/etherscan-v2/api-endpoints/accounts#get-a-list-of-internal-transactions-by-address).

Since we are starting to include other EVM-networks, we need to start showing these transactions.

# Changes

- Create type for Etherscan API internal transactions.
- Add method to fetch internal transactions history in `EtherscanProvider` provider.

# Tests

Adapted the unittests, creating specific mocks for different types of Etherescan API transactions.


src/frontend/src/eth/providers/etherscan.providers.ts
```diff
@@ -75,13 +79,61 @@ export class EtherscanProvider {
...
+				// Chain ID is not delivered by the Etherscan API so, we naively set 0
+				chainId: 0n
+			})
+		);
+	}
+
+	transactions = async (params: {
 		address: EthAddress;
 		startBlock?: BlockTag;
-	}): Promise<Transaction[]> => this.getHistory({ address, startBlock });
```
roman-nazaruk, line 84: Do we need destructuring params now?

AntonioVentilii: Good point! Changed

roman-nazaruk (approved):
LGTM

PR #5764, opened 2025-04-09
fix(frontend): remove track event which breaks sign-on on Safari

peterpeterparker:
# Motivation

Safari expects popups to be triggered directly by a user interaction, such as a click event. If anything interferes with that interaction—like an intermediate event handler—Safari considers it suspicious, blocks the popup, and requires the user to manually confirm its opening.

This is what's currently happening: the addition of a track event between the click action and the sign-in with Internet Identity (which opens in a popup) breaks the expected flow.

It's also worth noting that even if the user manually confirms the popup, the sign-up process will still fail, as communication between the dapps will be blocked.

The issue currently occurs on production.

# Changes

- Delete / remove track event for analytics

# Screenshot

Issue replicated locally.

Capture d’écran 2025-04-09 à 22 03 22


AntonioVentilii (approved):

AntonioVentilii (approved):
Thank you for the fix!

P.S. shall we remove the constant too, since it is not used?

peterpeterparker:
@AntonioVentilii I just removed the constant, thanks for pointing that out!

peterpeterparker:
Note that it might work with a passive function in between such as plausible track event, which would explain why it works on staging and beta but not prod where an additional async function is triggered.

PR #5767, opened 2025-04-09
chore(frontend): format tests removing unnecessary imports

AntonioVentilii:
# Motivation

We remove the not needed  `import { beforeEach } from 'node:test';`

UNRELATED: removed an unnecessary whitespace


daviddecentage (approved):
LGTM, thx

PR #5771, opened 2025-04-10
fix(frontend): remove track event for legacy sign-in

AntonioVentilii:
# Motivation

Follow-up of PR https://github.com/dfinity/oisy-wallet/pull/5764 : we remove the track event during "legacy" sign-in since it may cause the same issue solved in the other PR:

> Safari expects popups to be triggered directly by a user interaction, such as a click event. If anything interferes with that interaction—like an intermediate event handler—Safari considers it suspicious, blocks the popup, and requires the user to manually confirm its opening.
> 
> This is what's currently happening: the addition of a track event between the click action and the sign-in with Internet Identity (which opens in a popup) breaks the expected flow.
> 
> It's also worth noting that even if the user manually confirms the popup, the sign-up process will still fail, as communication between the dapps will be blocked.
> 
> The issue currently occurs on production.


peterpeterparker (approved):
LGTM, thx

PR #5772, opened 2025-04-10
refactor(e2e): remove unnecessary matrix in E2E tests final step

AntonioVentilii:
# Motivation

The final step of the E2E tests CI does not really need to run on different environments. It simply checks if the previous jobs worked as they should.


bitdivine (approved):
Thanks!

PR #5778, opened 2025-04-10
fix(frontend): Fixes auth help modal on Safari

daviddecentage:
# Motivation

On Safari the auth help modal was offset initially presumably due to a rendering bug.

# Changes

We could solve the issue by placing the modal on the root of the page instead of in a nested element.

# Tests

Tested on FE3 on Stefans machine.


roman-nazaruk (approved):
LGTM

AntonioVentilii (approved):

PR #5779, opened 2025-04-10
test(backend): Check candid compatibility

bitdivine:
# Motivation
We wish CI to fail if there is an unannounced breaking change in the backend canister API

# Changes
- Run an compatibility check with the prod API, when the backend candid file changes and no breaking change has been announced.

# Tests
### Test passes if a PR contains no changes to the backend API
https://github.com/dfinity/oisy-wallet/actions/runs/14379681392
Screenshot from 2025-04-10 14-02-02

### Test passes if a PR contains API changes and a conventional commit to match
I added a breaking change to the code and modified the PR title & description to declare the change.  The result was as expected:

Screenshot from 2025-04-10 14-05-53

Tests are skipped because the PR declares the breaking change:
Screenshot from 2025-04-10 14-06-58

### Test fails if there is a breaking API change
I kept the breaking change but ran without the PR breaking change conventions.  This failed, as desired:
Screenshot from 2025-04-10 14-45-39


.github/workflows/backend-tests.yml
```diff
@@ -87,8 +87,84 @@ jobs:
...
+          filters: |
+            api:
+              - 'src/backend/backend.did'
+      - name: 'Check whether the PR description & title mention "breaking interface"'
+        id: conventional-commit
+        run: |
+          if [[ "$TITLE" =~ ^([a-z]+)(\([-a-zA-Z0-9,]+\))\!\: ]]; then
+            echo "BREAKING_TITLE=true" | tee -a "$GITHUB_OUTPUT"
+          fi
+          if [[ "$DESCRIPTION" =~ "BREAKING CHANGE" ]]; then
```
AntonioVentilii, line 122: shall it be...?

`if [[ "$DESCRIPTION" =~ "# BREAKING CHANGE" ]]; then`

so that we force it to be a title? I am just wondering if somehow "breaking change" is mentioned in the description in another way that is not necessarily referred to the PR or OISY. WDYT?

No strong opinion, just wondering. I mean, worst case scenario: we don't run the check, and it is written very small inside the description, referring to something else, while in fact it causes breaking changes in OISY

.github/workflows/backend-tests.yml
```diff
@@ -87,8 +87,84 @@ jobs:
...
+            target
+          key: ${{ runner.os }}-cargo-backend-tests-${{ hashFiles('Cargo.lock', 'rust-toolchain.toml') }}
+
+      - name: Install dfx
+        if: steps.check-for-breaking-changes.outputs.EXPECT_NO_BREAKING_CHANGES == 'true'
+        uses: dfinity/setup-dfx@e50c04f104ee4285ec010f10609483cf41e4d365 # main
+
+      - name: Download backend.wasm.gz
+        uses: actions/download-artifact@95815c38cf2ff2164869cbab79da8d1f422bc89e # v4.2.1
+        with:
```
AntonioVentilii, line 146: theoretically this one can have the condition too, right?

` if: steps.check-for-breaking-changes.outputs.EXPECT_NO_BREAKING_CHANGES == 'true'`

.github/workflows/backend-tests.yml
```diff
@@ -87,8 +87,84 @@ jobs:
...
+      - name: 'Run backend tests'
+        if: steps.check-for-breaking-changes.outputs.EXPECT_NO_BREAKING_CHANGES == 'true'
+        working-directory: .
+        run: |
+          ./scripts/test.backend.sh -- --ignored candid || {
+            echo "The backend API is incompatible with production."
+            echo "If this is intentional, please use the conventional commit convention for indicating"
+            echo "a breaking change in this PR, then push another commit to re-run this job."
+            echo "E.g. title: feat(backend)!: Change the API to..."
+            echo "     body: BREAKING CHANGE: The backend API is now incompatible with production."
```
AntonioVentilii, line 159: and maybe here it could be a title? but no strong opinion

`body: # BREAKING CHANGE: The backend API is now incompatible with production."`

.github/workflows/backend-tests.yml
bitdivine: I'm wary of deviating from the conventional commit standard, however I can make the pattern stricter, making it hard to get false positives.  A line has to start with exactly the characters: `BREAKING CHANGE: `

.github/workflows/backend-tests.yml
bitdivine: Good catch!  Done.

.github/workflows/backend-tests.yml
bitdivine: As above, I'm wary of deviating from the standard.  Why have a standard if we go our own way anyway?

AntonioVentilii (approved):
LGTM tks!

PR #5781, opened 2025-04-10
fix(frontend): Update missing network logo

daviddecentage:
# Motivation

In the network icons refactoring we missed one of the logos.

# Changes

Updated so it uses the NetworkLogo component

#Tests 
image


AntonioVentilii (approved):
Tks!!!

PS. May you please add some screenshot to the description?

PR #5783, opened 2025-04-10
feat(frontend): the canister backend changes required for the pow feature

DecentAgeCoder:
# Motivation

This PR aims to enhance backend communication by introducing new API functions for challenge creation and updating signing methods while maintaining backward compatibility. It also refines existing service integrations and updates related tests accordingly.

# Changes

- Added createPowChallenge and allowSigningResult API methods, improving extensibility and functionality.
Updated implementations for allowSigning to include nonce and introduced createPowChallengeResult and allowSigningResult.
- Made changes to avoid breaking existing integrations.
- Modified loader.services.ts to include the nonce parameter in allowSigning calls.
- Added new types for API responses, including AllowSigningResult and CreateChallengeResult.
- Created tests for createPowChallengeResult and error-handling situations.
- Updated existing tests to validate behavior with the new nonce parameter.

# Tests
- New test cases for createPowChallenge to cover both success and various error conditions like missing profiles and randomness issues.
- Updated test coverage for the allowSigning API to validate new nonce parameter usage.
- This PR introduces functional enhancements while ensuring that existing features remain unaffected.


src/frontend/src/lib/canisters/backend.canister.ts
```diff
@@ -175,16 +177,32 @@ export class BackendCanister extends Canister<BackendService> {
...
 		}
+		return result.Ok;
+	};
+
+	// directly returning result and not the response
+	allowSigningResult = async (
+		allowSigningRequest: { request?: AllowSigningRequest } = {}
+	): Promise<AllowSigningResult> => {
+		const { allow_signing } = this.caller({ certified: true });
+		return await allow_signing(allowSigningRequest.request ? [allowSigningRequest.request] : []);
```
AntonioVentilii, line 200:
```suggestion
		return await allow_signing(toNullable(allowSigningRequest.request));
```

src/frontend/src/lib/canisters/backend.canister.ts
```diff
@@ -175,16 +177,32 @@ export class BackendCanister extends Canister<BackendService> {
...
-			return Ok;
+		if ('Err' in result) {
+			throw mapAllowSigningError(result.Err);
 		}
+		return result.Ok;
+	};
+
+	// directly returning result and not the response
+	allowSigningResult = async (
+		allowSigningRequest: { request?: AllowSigningRequest } = {}
```
AntonioVentilii, line 197: may you please create an explicit type for this function, similar to the other methods?

src/frontend/src/lib/canisters/backend.canister.ts
```diff
@@ -175,16 +177,32 @@ export class BackendCanister extends Canister<BackendService> {
 		throw mapBtcSelectUserUtxosFeeError(response.Err);
 	};
 
-	allowSigning = async ({ request }: AllowSigningParams): Promise<AllowSigningResponse> => {
+	// keeping the existing implementation to not beak the existing integration
+	allowSigning = async (
+		allowSigningRequest: { request?: AllowSigningRequest } = {}
```
AntonioVentilii, line 182: same

src/frontend/src/lib/canisters/backend.canister.ts
```diff
@@ -175,16 +177,32 @@ export class BackendCanister extends Canister<BackendService> {
...
+	): Promise<AllowSigningResponse> => {
 		const { allow_signing } = this.caller({ certified: true });
-		const response = await allow_signing(toNullable(request));
+		const result = await allow_signing(
+			allowSigningRequest.request ? [allowSigningRequest.request] : []
```
AntonioVentilii, line 186: `toNullable`

src/frontend/src/lib/canisters/backend.canister.ts
```diff
@@ -175,16 +177,32 @@ export class BackendCanister extends Canister<BackendService> {
...
+		);
 
-		if ('Ok' in response) {
-			const { Ok } = response;
-			return Ok;
+		if ('Err' in result) {
+			throw mapAllowSigningError(result.Err);
```
AntonioVentilii, line 190: preferably, we should invert the condition:

```typescript
if ('Ok' in response) {
....
}

throw...
```

src/frontend/src/lib/canisters/backend.canister.ts
```diff
@@ -175,16 +177,32 @@ export class BackendCanister extends Canister<BackendService> {
...
+		allowSigningRequest: { request?: AllowSigningRequest } = {}
+	): Promise<AllowSigningResult> => {
+		const { allow_signing } = this.caller({ certified: true });
+		return await allow_signing(allowSigningRequest.request ? [allowSigningRequest.request] : []);
+	};
 
-		throw mapAllowSigningError(response.Err);
+	createPowChallengeResult = (): Promise<CreateChallengeResult> => {
+		const { create_pow_challenge } = this.caller({ certified: true });
+		return create_pow_challenge();
```
AntonioVentilii, line 205: can you please add tests for all the new functions, similar to the other methods?

src/frontend/src/lib/api/backend.api.ts
```diff
@@ -148,6 +158,18 @@ export const allowSigning = async ({
 	return allowSigning(params);
 };
 
+export const allowSigningResult = async ({
+	request,
+	identity
+}: CanisterApiFunctionParams<{
+	request?: AllowSigningRequest;
```
AntonioVentilii, line 165: once you create a specific method for the function in the canister class, you can reuse it here

src/frontend/src/lib/api/backend.api.ts
```diff
@@ -148,6 +158,18 @@ export const allowSigning = async ({
 	return allowSigning(params);
 };
 
+export const allowSigningResult = async ({
+	request,
+	identity
```
AntonioVentilii, line 163:
```suggestion
	identity,
	...params
```

src/frontend/src/lib/api/backend.api.ts
```diff
@@ -148,6 +158,18 @@ export const allowSigning = async ({
...
+}: CanisterApiFunctionParams<{
+	request?: AllowSigningRequest;
+}>): Promise<AllowSigningResult> => {
+	const { allowSigningResult } = await backendCanister({ identity });
+
+	// Conditionally call allowSigning with request or provide default logic
+	return allowSigningResult(request ? { request } : {});
```
AntonioVentilii, line 170:
```suggestion
	return allowSigningResult(params);
```

src/frontend/src/lib/canisters/backend.canister.ts
DecentAgeCoder: Yes, I agree

DecentAgeCoder: This is very uncommon : Usually the error cases are handled immediately at the start of the function, allowing the rest of the code to focus on the successful path or `Ok` case. Why would you want to have it opposite?

AntonioVentilii: just for consistency, it is like that for all the other functions. We can for sure thinking of inverting it, i don't see why not.

Code-wise, I think the idea is that, if for whatever reason at some point the variant are not only `Ok` and `Error`, we are sure that we will handle correctly ONLY the `Ok` variant, and the rest we raise it, so we know what is happening

src/frontend/src/lib/canisters/backend.canister.ts
DecentAgeCoder: I will do that. Even-though from a design point of view why we would you use here (backend.canister.ts) a new and redundant model (which does not abstract anything). Why not use a more domain centric/driven approach for integrating external API services? Same for errors. :wink:

src/frontend/src/lib/canisters/backend.canister.ts
DecentAgeCoder: Yes I absolutely agree

src/frontend/src/lib/canisters/backend.canister.ts
AntonioVentilii: What happens is really a porting "issue":

- most of backend interfaces uses nullable, while frontend doesn't. Plus we need to adapt FE types to the BE, so it is always useful to have different interfaces for the FE methods and the BE (bindings) methods
- consistency: we use camel case in the FE, while the bindings (and the BE) use snake case. So, most of the time we just need to create a different interface, for example. So, we should do for them all

Plus, we reuse those types both in backend.canister and backend.api

src/frontend/src/lib/canisters/backend.canister.ts
```diff
@@ -175,16 +177,30 @@ export class BackendCanister extends Canister<BackendService> {
 		throw mapBtcSelectUserUtxosFeeError(response.Err);
 	};
 
-	allowSigning = async ({ request }: AllowSigningParams): Promise<AllowSigningResponse> => {
+	// keeping the existing implementation to not beak the existing integration
```
AntonioVentilii:
```suggestion
	// keeping the existing implementation to not break the existing integration
```

AntonioVentilii (approved):
As discussed offline, we will merge this one for now and check the usage of `allowSigningResult` in the worker in the next PR, to verify that we are not adding duplication of code.

PR #5787, opened 2025-04-10
feat(frontend): Token list filter store

daviddecentage:
# Motivation

For upcoming PR that adds filter functionality to the token list, we add the store for the filter value to be stored in.

# Changes

Added store


src/frontend/src/lib/stores/token-list.store.ts
```diff
@@ -0,0 +1,7 @@
+import { writable } from 'svelte/store';
+
+export interface TokenListStoreData {
```
AntonioVentilii, line 3: do you need to export it for the next PRs?

AntonioVentilii (approved):
LGTM tks!

One non-blocker question

src/frontend/src/lib/stores/token-list.store.ts
daviddecentage, line 3:
```suggestion
interface TokenListStoreData {
```

PR #5797, opened 2025-04-10
feat(frontend): include `vitest` lint rules

AntonioVentilii:
# Motivation

We want to start including the vitest rules provided by the library [eslint-config-oisy-wallet](https://github.com/dfinity/eslint-config-oisy-wallet).

However, we turn the rule [expect-expect](https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/expect-expect.md) off, because we use functions that calls tests and that is not yet supported (nor it will be probably): https://github.com/vitest-dev/eslint-plugin-vitest/issues/697


eslint.config.mjs
```diff
@@ -1,7 +1,17 @@
...
 
 export default [
-	...config,
+	...vitestConfig,
+	...svelteConfig,
+
+	{
+		rules: {
+			// TODO: re-enable this rule when it includes `expect` statements nested in callable functions.
+			'vitest/expect-expect': ['off']
```
peterpeterparker, line 11: Should we turn this off in the eslint-config-oisy-wallet lib?

AntonioVentilii: I was wondering the same, but it was actually a good exercise to have it enabled at first, because there were a few examples where the function was not needed and I moved the tests directly inside the `describe(...)` statements.

The issue for OISY codebase is mostly (90%) around test-function `testDerivedUpdates` that is used around several test modules. I can put ignore for each one, but I thought that maybe this could be a one-shot solution.

It is among the "recommended" rules for the `vitest` lint plugin.

Boh... ultimately no strong opinion on my side, so if you feel it stronger, I can proceed with the removal in the library

peterpeterparker: Makes sense, thanks for the feedback!

peterpeterparker (approved):
LGTM, thx

PR #5798, opened 2025-04-10
test(frontend): add threshold auto-update (ratchet) to vitest coverage

AntonioVentilii:
# Motivation

There is an useful [autoUpdate](https://vitest.dev/config/#coverage-thresholds-autoupdate) option for vitest coverage that updates all threshold values `lines`, `functions`, `branches` and `statements` to configuration file when current coverage is better than the configured thresholds. This option helps to maintain thresholds when coverage is improved.

It is basically a "surrogate" of a possible coverage ratchet suggested by @bitdivine .

# Changes

- Turn `autoUpdate` on in `vitest.config` file.
- Create new CI  workflow to create a PR that updates the coverage thresholds if they improve. It is triggered for any new commit in `main`, since we want to keep up.

# Concurrency Note

If PRs are merged before a new update-threshold-PR, we may lose some increase. For example:

1. Current coverage on `main` is 50% and thresholds are at 50%
2. `PR1` is merged and creates new `Threshold-PR1` of 80% (still being merged)
3. Coverage on `main` is at 80% but thresholds are still not updated
4. `PR2` is being merged too and would create a new `Threshold-PR2` of 70% (but the CI to create it is still running)
5. `PR2` is merged before `Threshold-PR1`
6. Coverage on `main` is somewhere between 70% and 80% (because of both `PR1` and `PR2`).
7. `Threshold-PR1` is rebased with `main` and shows an error, since the thresholds decreased
8. We close `Threshold-PR1` and await for the creation of `Threshold-PR2`

# Tests

Tested manually with https://github.com/dfinity/oisy-wallet/actions/runs/14389041441/job/40351470318 and it created PR https://github.com/dfinity/oisy-wallet/pull/5799


bitdivine (approved):
Cool!  I'm a bit surprised that vitest updates the config file itself and not some separate thresholds file, but whatever works!

PR #5800, opened 2025-04-10
fix(devops): callable deployment workflow is missing secrets as inputs

AntonioVentilii:
# Motivation

The workflow that deploys OISY has become callable (PR https://github.com/dfinity/oisy-wallet/pull/5626), so that the `tag-release.yml` workflow can trigger it.

However I forgot to create the input secrets for the `workflow_call` trigger in the deployment CI, causing an error (see https://github.com/dfinity/oisy-wallet/actions/runs/14361662192).

I included them in this PR.


bitdivine (approved):
Thanks!

PR #5807, opened 2025-04-11
feat(frontend): Adds address boook feature flag

FlavioInacta:
# Motivation

To turn on and off the address book feature in different environments. This will be removed once the address book implementation is complete.

# Changes

Adds `VITE_ADDRESS_BOOK_ENABLED` feature flag


AntonioVentilii (approved):
LGTM tks! Let's remember to add it to the deployment CIs when we have more of it developed

.env.example
```diff
@@ -15,4 +15,5 @@ VITE_AUTH_ALTERNATIVE_ORIGINS=
 VITE_ONRAMPER_API_KEY_DEV=pk_test_
 VITE_ONRAMPER_API_KEY_PROD=pk_prod_
 VITE_SWAP_ACTION_ENABLED=
+VITE_ADDRESS_BOOK_ENABLED=
```
AntonioVentilii, line 18: not necessary to add it here, since it is a temporary flag

PR #5808, opened 2025-04-11
refactor(frontend): use network IDs instead of token IDs for loading addresses

AntonioVentilii:
# Motivation

In the flow to load/verify/certify addresses, we are using the token ID to associate each address. However, it is more correct (or at least consistent), to associate it to the network ID.

# Changes

- Replace all instances of usage of token ID with network ID in the loading flow for addresses.
- Replace all instances of usage of token ID with network ID in the verify flow for addresses.

# Tests

Adapted current tests.


roman-nazaruk (approved):
LGTM

PR #5810, opened 2025-04-11
test(frontend): remove declaration folder form coverage analysis

AntonioVentilii:
# Motivation

It doesn't make real sense to include the `declarations` folder among the path analyzed for vitest coverage, since it is constituted of auto-built external files.


roman-nazaruk (approved):
LGTM

PR #5811, opened 2025-04-11
feat(frontend): Sliding input component

daviddecentage:
# Motivation

For the upcoming token list filter PR we need to create a sliding input that opens over its parent content.

# Changes

- Added condensed variant for inputs
- Export input value so it can be bound
- Added SlidingInput component
- Added SlidingInputBoundary component

# Tests

Sliding input showcasing passing `overflowable-content` slot, responsiveness, clear and outside click functionality:

https://github.com/user-attachments/assets/9984c5ca-c286-4de2-bd41-9918f8f84b51

Sliding input showcasing without `overflowable-content` slot, passing different icon, to be reused anywhere without adjusting to parent element:

https://github.com/user-attachments/assets/64838f7b-395a-4d06-9b75-cf25bbae4d24


src/frontend/src/lib/components/ui/SlidingInput.svelte
```diff
@@ -0,0 +1,112 @@
...
+	import { onMount } from 'svelte';
+	import { fade, slide } from 'svelte/transition';
+	import { afterNavigate } from '$app/navigation';
+	import IconClose from '$lib/components/icons/lucide/IconClose.svelte';
+	import ButtonIcon from '$lib/components/ui/ButtonIcon.svelte';
+	import InputTextWithAction from '$lib/components/ui/InputTextWithAction.svelte';
+	import { SLIDE_PARAMS } from '$lib/constants/transition.constants.js';
+	import { i18n } from '$lib/stores/i18n.store';
+
+	export let testIdPrefix: string;
```
AntonioVentilii, line 12: it can be `undefined` too right?

`export let testIdPrefix: string | undefined = undefined`

src/frontend/src/lib/components/ui/SlidingInput.svelte
```diff
@@ -0,0 +1,112 @@
...
+	export let ariaLabel: string;
+
+	let visible = false;
+
+	let button: HTMLButtonElement | undefined;
+	let inputElement: HTMLInputElement | undefined;
+
+	const handleOpen = (e?: Event) => {
+		e?.stopPropagation();
+		if (visible) {
```
AntonioVentilii, line 25: isn't it always defined?

```suggestion

	const handleOpen = (e: Event) => {
		e.stopPropagation();
		if (visible) {
```

src/frontend/src/lib/components/ui/SlidingInput.svelte
```diff
@@ -0,0 +1,112 @@
...
+			handleClose();
+		} else {
+			inputElement?.focus();
+			visible = true;
+		}
+	};
+
+	const handleClose = () => {
```
AntonioVentilii, line 33: maybe this should be `handleExternalClick`?

src/frontend/src/lib/components/ui/SlidingInput.svelte
```diff
@@ -0,0 +1,112 @@
...
+			}
+		}, 1);
+	});
+</script>
+
+<div class="z-3 absolute right-0 w-full">
+	{#if visible}
+		<div
+			in:slide={{ ...SLIDE_PARAMS, axis: 'x' }}
+			out:fade={SLIDE_PARAMS}
```
AntonioVentilii, line 68: can you give `SLIDE` params to `fade`? if so, better to explicit pass the params, instead of using `SLIDE_PARAMS`

src/frontend/src/lib/styles/global/gix.scss
```diff
@@ -122,6 +123,10 @@ div.input-field input[id] {
 	padding: var(--padding-1_5x);
 }
 
+.input-field.condensed input {
```
AntonioVentilii, line 126: can this go directly inside the new component?

src/frontend/src/lib/styles/global/gix.scss
```diff
@@ -42,6 +42,7 @@
 	--focus-border-color: var(---color-border-brand-primary);
 	--input-border-color: var(--color-border-primary);
 	--input-border-size: 1px;
+	--input-width: 100%;
```
AntonioVentilii, line 45: have you checked if this changes the other usages of input?

src/frontend/src/lib/components/ui/SlidingInputBoundary.svelte
```diff
@@ -0,0 +1,9 @@
+<!--
+	We need this boundary element to define how far out the SlidingInput can go.
+-->
+<div class="relative flex w-full">
+	<div class="flex pr-12">
+		<slot name="content" />
+	</div>
+	<slot name="input" />
+</div>
```
AntonioVentilii, line 9: this component can go directly in another PR, no? it is not related to the new component

Meaning, the new component should work by itself, otherwise you can merge this one there, or not?

src/frontend/src/lib/components/ui/SlidingInput.svelte
```diff
@@ -0,0 +1,112 @@
...
+		{ariaLabel}
+		testId={`${testIdPrefix}-open-btn`}
+	>
+		<slot name="icon" slot="icon" />
+	</ButtonIcon>
+</div>
+
+{#if visible && inputValue === ''}
+	<div class="fixed bottom-0 left-0 right-0 top-0 z-0">
+		<Backdrop invisible on:nnsClose={handleClose} />
```
AntonioVentilii, line 110: isn't the backdrop already `fixed bottom-0 left-0 right-0 top-0 z-0`? honest question

src/frontend/src/lib/components/ui/SlidingInput.svelte
```diff
@@ -0,0 +1,112 @@
...
+		// instead of dispatching an event from the parent component using this component
+		setTimeout(() => {
+			if (inputValue === '') {
+				handleClose();
+			}
+		}, 1);
+	});
+</script>
+
+<div class="z-3 absolute right-0 w-full">
```
AntonioVentilii, line 64: this `div` should not be `absolute` nor have specific positions on `z-index`, otherwise it is difficult to re-use

src/frontend/src/lib/components/ui/SlidingInput.svelte
```diff
@@ -0,0 +1,112 @@
...
+	{#if visible}
+		<div
+			in:slide={{ ...SLIDE_PARAMS, axis: 'x' }}
+			out:fade={SLIDE_PARAMS}
+			class="input-field condensed absolute right-0 -mt-[11px] mr-[1px] flex w-full overflow-hidden transition-all duration-300 md:w-[250px]"
```
AntonioVentilii, line 69:
```suggestion
			class="input-field condensed absolute right-0 -mt-[11px] mr-px flex w-full overflow-hidden transition-all duration-300 md:w-[250px]"
```

AntonioVentilii, line 69: `w-[250px]` is too specific i think... shouldn't it be more responsive? or you meant `max-w` ?

AntonioVentilii, line 69: why the transitions if you already have the svelte transitions?

src/frontend/src/lib/components/ui/SlidingInput.svelte
```diff
@@ -0,0 +1,112 @@
...
+				autofocus
+			/>
+			{#if inputValue !== ''}
+				<div
+					class="transition-bg duration-250 top-3.25 absolute right-12 bg-primary"
+					transition:fade
+				>
+					<ButtonIcon
+						ariaLabel={$i18n.core.text.clear_filter}
+						styleClass="text-tertiary"
```
AntonioVentilii, line 86: i think there is the prop `colorStyle` to be used here. Or the color pallet is another one?

AntonioVentilii (commented):
I really like this component! Some observations:

src/frontend/src/lib/components/ui/SlidingInput.svelte
daviddecentage: 250px is basically the minimum/default width. On smaller screens it goes to w-full

src/frontend/src/lib/components/ui/SlidingInput.svelte
daviddecentage: no, its also used in handleOpen, onMount and afterNavigate, not only on externalClick

src/frontend/src/lib/components/ui/SlidingInput.svelte
daviddecentage: not when we open it in the onMount. But we dont need the stopPropagation anymore I remove it

src/frontend/src/lib/components/ui/SlidingInput.svelte
daviddecentage: Its a prefix, if we dont pass the testIds will be "-open-btn" and "-clear-btn`"

src/frontend/src/lib/components/ui/SlidingInput.svelte
daviddecentage: updated

src/frontend/src/lib/components/ui/SlidingInput.svelte
daviddecentage: updated

src/frontend/src/lib/components/ui/SlidingInput.svelte
daviddecentage: No, without that it will just be the size of the wrapper element where this is placed in

src/frontend/src/lib/styles/global/gix.scss
daviddecentage: I checked and couldnt find any differences as all inputs are 100% width anyways. For some reason here it wasnt filling 100% though

src/frontend/src/lib/styles/global/gix.scss
daviddecentage: No, we need to force it over the gix components own styling

src/frontend/src/lib/components/ui/SlidingInput.svelte
daviddecentage: Its different

src/frontend/src/lib/components/ui/SlidingInput.svelte
AntonioVentilii: you can give a default that is not `undefined` or not set the test id if nullish. We do something similar in the other components whe

src/frontend/src/lib/components/ui/SlidingInput.svelte
```diff
@@ -0,0 +1,129 @@
...
+
+	afterNavigate(() => {
+		// By setting a timeout of 1ms we make sure it runs after any parent components afterNavigate that uses this component.
+		// This works because normal function calls take precedent over timeout functions
+		// Can be useful to trigger the condition "handleClose if empty inputValue" by setting the bound input value to empty string
+		// instead of dispatching an event from the parent component using this component
+		setTimeout(() => {
+			if (inputValue === '') {
+				handleClose();
+			}
```
AntonioVentilii, line 62: if i understand correctly the logic here you can do simply

```ts
if (inputValue === '') {
				visible=== false;
			}
```
			
no? since the `handleClose` checks for `inputValue` too. Or do you plan on expanding `handleClose`?

src/frontend/src/lib/components/ui/SlidingInput.svelte
```diff
@@ -0,0 +1,129 @@
...
+
+	let hasOverflowableSlot: boolean;
+	$: hasOverflowableSlot = nonNullish($$slots['overflowable-content']);
+
+	let visible = false;
+
+	let button: HTMLButtonElement | undefined;
+	let inputElement: HTMLInputElement | undefined;
+
+	const handleOpen = () => {
```
AntonioVentilii, line 27: `handleToggle`? or `handleOnClick`? (since you both close and open)

src/frontend/src/lib/components/ui/SlidingInput.svelte
daviddecentage: right, updated

src/frontend/src/lib/components/ui/SlidingInput.svelte
```diff
@@ -0,0 +1,129 @@
...
+	import { i18n } from '$lib/stores/i18n.store';
+
+	export let testIdPrefix: string;
+	export let inputValue: string;
+	export let inputPlaceholder: string | undefined;
+	export let disabled = false;
+	export let ariaLabel: string;
+
+	let hasOverflowableSlot: boolean;
+	$: hasOverflowableSlot = nonNullish($$slots['overflowable-content']);
```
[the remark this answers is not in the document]
daviddecentage: You cant mix reactive vars and new props, so I had to refactor the props definitions to svelte 5 as well

src/frontend/src/lib/components/ui/SlidingInput.svelte
daviddecentage: updated naming

AntonioVentilii (approved):
LGTM tks!

PR #5812, opened 2025-04-11
feat(frontend): extract function to parse boolean env variables

AntonioVentilii:
# Motivation

For better manageability, testing and to reduce repeated code, we extract an util to parse boolean env variables.

We typically use it for env variables that define a feature flag or the switch of a specific network/feature. So, by default (in most cases), we check if the variable is `true`.

For the flags that define a disabled feature, we check if the variable is `false`.

# Changes

- Create util to parse a putative string (that is how the env variables are passed in vite) into a boolean: the logic is the same as the existing one.
- Add an optional check that defaults to `true`. We set it to `false` for variables that define disabled features.
- Use the util in the code, where possible.

# Tests

Added tests.


roman-nazaruk (approved):
LGTM

PR #5815, opened 2025-04-11
feat(backend): add contact type and address type.

SYInacta:
⭐ PR 1: Implement necessary changes in shared for Contacts feature

# Motivation

Provide the shared type, contacts structures and logic, including challenge creation, update, and delete of contacts and addresses associated to those contact.

# Changes

Introduced a new rust module contacts in shared
Added the necessary candid structures

# Tests


src/shared/src/types/contact.rs
```diff
@@ -0,0 +1,118 @@
...
+            "AVAX" => AddressType::AVAX,
+            "BSC" => AddressType::BSC,
+            "MATIC" => AddressType::MATIC,
+            _ => AddressType::Other(s.to_string()),
+        }
+    }
+}
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct ContactAddress {
```
DecentAgeCoder, line 62: Since this structures are reflecting the various canister endpoints used for address book, they should therefore be prefixed with 'Request' or 'Response'. So here you would probably want to rename it to ContactAddressRequest. This applies to all structures in this source file.

src/shared/src/types/contact.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub contact_id: String,
+    pub network_type: AddressType,
+    pub address: String,
+}
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct UpdateAddress {
+    pub contact_id: String,
+    pub old_network_type: AddressType,
+    pub old_address: String,
```
DecentAgeCoder, line 105: This could imply that one has to provide the old and new value of the address. If the two fields are representing two different formats, than I would probably call it legacy_address.

DecentAgeCoder (changes requested):
Mostly minor issues that still need to be addressed

src/shared/src/types.rs
```diff
@@ -7,6 +7,7 @@ pub type Timestamp = u64;
 pub mod account;
 pub mod backend_config;
 pub mod bitcoin;
+pub mod contact;
```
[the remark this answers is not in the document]
SYInacta: Yes, the main data is contacts that have addresses.

src/shared/src/types/contact.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub new_address: ContactAddress,
+}
+
+#[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
+pub enum ContactError {
```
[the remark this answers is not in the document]
SYInacta: added ContactNameAlreadyExists.

src/shared/src/types/contact.rs
```diff
@@ -0,0 +1,118 @@
...
+    UserNotFound,
+    ContactNotFound,
+    ContactIdAlreadyExists,
+    InvalidContactData,
+    AddressAlreadyExists,
+    AddressNotFound,
+    InvalidAddressFormat,
```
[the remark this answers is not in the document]
SYInacta: added.

src/shared/src/types/contact.rs
SYInacta: added everywhere.

src/shared/src/types/contact.rs
SYInacta: current and new changed for update.

DecentAgeCoder (commented):
Looks good now. Max will have a final glance at the PR and than approve.

src/shared/src/types/contact.rs
```diff
@@ -0,0 +1,117 @@
...
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct Contact {
+    pub id: String,
+    pub name: String,
+    pub addresses: Vec<ContactAddressData>,
+    pub avatar: Option<String>,
+}
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub enum AddressType {
```
bitdivine, line 14: Note: You may want to sync with @AntonioVentilii on this list.  We could have a meeting with the three of us.  In particular:

-  We support testnets as well as mainnets for each network, so some clarity is needed: is this a list of protocols, or is it "where to send the message"?  E.g. what is the addressType for ckBTC (which is BTC, but using the ICP protocol), what is the address type for a Solana Testnet address, what is the address type for a Solana Devnet Account?

- There is a risk of having several lists of networks that are all slightly different.  One for snapshots, one for rewards, one for rewards, and so on.  It might be an idea to have a single list of networks and a single list of protocols.

bitdivine: As discussed, maybe move this to `shared/types/`?

SYInacta: Done.

src/shared/src/types/address_type.rs
```diff
@@ -0,0 +1,61 @@
+use std::fmt;
+
+use candid::{CandidType, Deserialize};
+
+use super::account::TokenAccountId;
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
```
bitdivine, line 7: This refers specifically to account addresses, not canister addresses or other sorts of address?  If so, something like `AccountIdType` would be less ambiguous.  A rustdoc would also help clarify the purpose.

```suggestion
/// Token account ID types.
#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
```

bitdivine (approved):
Thanks!  Some more rustdoc would help - we are trying to get the rustdoc coverage up to the point that we can enforce the missing rustdoc clippy. But otherwise looks good to me!

src/shared/src/types/address_type.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+}
+
+impl From<&TokenAccountId> for AddressType {
+    fn from(token_account_id: &TokenAccountId) -> Self {
+        match token_account_id {
+            TokenAccountId::Icrcv2(_) => AddressType::ICP,
+            TokenAccountId::Sol(_) => AddressType::SOL,
+            TokenAccountId::Btc(_) => AddressType::BTC,
+            TokenAccountId::Eth(_) => AddressType::ETH,
```
AntonioVentilii, line 43: why not using these `TokenAccountId` variants directly?

AntonioVentilii (changes requested):

src/shared/src/types/account.rs
```diff
@@ -199,3 +199,25 @@ impl AccountId<EthereumMainnet> for EthAddress {}
...
+        match self {
+            TokenAccountId::Icrcv2(_) => write!(f, "ICP"),
+            TokenAccountId::Sol(_) => write!(f, "SOL"),
+            TokenAccountId::Btc(_) => write!(f, "BTC"),
+            TokenAccountId::Eth(_) => write!(f, "ETH"),
+        }
+    }
+}
+
+impl From<&str> for TokenAccountId {
```
DecentAgeCoder, line 214: I do not recommend to misuse the TokenAccountId (by initializing it with an empty address) as an "network  enum". You also do not have to convert the TokenAccountId. Thede- serialization is already correctly handled by the  TokenAccountId itself.

src/shared/src/types/account.rs
```diff
@@ -199,3 +199,25 @@ impl AccountId<EthereumMainnet> for EthAddress {}
 impl AccountId<EthereumSepolia> for EthAddress {}
 impl TokenId<EthereumMainnet> for EthAddress {}
 impl TokenId<EthereumSepolia> for EthAddress {}
+
+impl std::fmt::Display for TokenAccountId {
```
DecentAgeCoder, line 203: Why do you need to convert the TokenAccountId to a network identifier?

DecentAgeCoder (changes requested):
Please update the PR according to the comments

src/shared/src/types/contact.rs
```diff
@@ -0,0 +1,67 @@
...
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct AddContactRequest {
+    pub contact: Contact,
+}
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct RemoveContactRequest {
+    pub contact_id: String,
+    pub network_type: TokenAccountId,
+    pub address: String,
```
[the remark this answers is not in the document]
FlavioInacta: As far as I understood, the `TokenAccountId` is sometimes a Map of a Map of a Map?

Eg:
```ts
{
  Icrcv2: {
    WithPrincipal: {
      owner: 'xxx',
    }
  }
}
```
It loks to me very over engineered. I know we have this type `TokenAccountId` but if it is not an enum and we know that all addresses are a simple string, why not just save a string together with the _real_ enum?

This type makes it really complicated. For every type of address we will need to find the right keys to peel out the simple `string` that we need.

src/shared/src/types/account.rs
SYInacta: No

src/shared/src/types/account.rs
SYInacta: Fixed

src/shared/src/types/contact.rs
```diff
@@ -0,0 +1,67 @@
...
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct AddAddressRequest {
+    pub contact_id: String,
+    pub address: ContactAddressData,
+}
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct UpdateAddressRequest {
+    pub contact_id: String,
+    pub current_network_type: TokenAccountId, // Changed from TokenAccountId
```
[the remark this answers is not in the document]
SYInacta: Fixed.

src/shared/src/types/contact.rs
```diff
@@ -0,0 +1,67 @@
...
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct UpdateContactRequest {
+    pub contact: Contact,
+}
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct AddAddressRequest {
+    pub contact_id: String,
+    pub address: ContactAddressData,
```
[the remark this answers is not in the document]
SYInacta: contact_address_data to be clearer.

DecentAgeCoder (approved):
Looks good from my side. Antonio and Max will now review.

src/shared/src/types/contact.rs
```diff
@@ -0,0 +1,67 @@
...
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct AddContactRequest {
+    pub contact: Contact,
+}
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct RemoveContactRequest {
+    pub contact_id: String,
+    pub network_type: TokenAccountId,
```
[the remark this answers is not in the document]
SYInacta: Fix

src/shared/src/types/contact.rs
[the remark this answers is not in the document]
SYInacta: Fix.

AntonioVentilii (approved):
On my side, I see that you are using `TokenAccointId`, that was my only note. Unblocking it, but please wait for @bitdivine 's review

AntonioVentilii:
Why not wait for @bitdivine 's review?! @SYInacta

AntonioVentilii:
Next time i will not approve

src/shared/src/types/contact.rs
```diff
@@ -0,0 +1,63 @@
+use candid::{CandidType, Deserialize};
+
+use super::account::TokenAccountId;
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct Contact {
+    pub id: String,
+    pub name: String,
+    pub addresses: Vec<ContactAddressData>,
```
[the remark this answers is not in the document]
DecentAgeCoder: Thanks for the excellent finding and proposal. I completely agree with your concern..I consider it critical.

We will add a validation function for all exposed models and register it with validate_on_deserialize() so that the validation step is injected.
Additionally, we will mark the struct itself with the serde attribute #[serde(remote = "Self")] to enable the default serde deserialization.

For instance the validator for the Contact struct will probably look like this: 
```
validate_on_deserialize(Contact);
impl Validate for Contact {
        // Validate id length
        if self.id.len() > 40 {
            return Err(candid::Error::String("Contact id too long".to_string()));
        }
        
        // Validate name length
        if self.name.len() > 100 {
            return Err(candid::Error::String("Contact name too long".to_string()));
        }
        
        # ... add more validations here
        
        // Also we need to restrict the length of the array
        for address in &self.addresses {
            if address.address.len() > 100 {
                return Err(candid::Error::String("Address too long".to_string()));
            }
            
            // probably its better to just call address.validate() and implement the validation directly in the address Address
            if let Some(label) = &address.label {
                if label.len() > 50 {
                    return Err(candid::Error::String("Address label too long".to_string()));
                }
            }
        }
        
        Ok(())
    }
}
```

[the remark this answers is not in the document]
bitdivine: LGTM.  You will probably want to use constants for the limits; you will need to know the limit in tests, for example.  There are some constants defined already such as `MAX_SYMBOL_LENGTH`.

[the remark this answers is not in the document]
DecentAgeCoder: Yes we will introduce constants for the length.

Regarding the comment I had added to the code in my previous message:
 // probably its better to just call address.validate() and implement the validation directly in the address Address
 
 Do I need to propagate the validation to the sub-structures manually or is this also automatically happening during de-serialization? If it happens automatically: are the arrays already initialized when testing the size of an  array? I would assume it de-serializes the structures from top to bottom..?
```
 
  for address in &self.addresses {
    address.validate(..)
  }
```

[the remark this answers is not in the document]
bitdivine: If sub-structures implement the Validate trait, they will run their own validator.

PR #5818, opened 2025-04-11
feat(frontend): removed junobuild/analytics worker

roman-nazaruk:
# Motivation

Removed the Juno analytics worker, as we are now using Plausible for tracking and no longer need Juno.

Removing JunoAnalytics from package.json will be handled in a separate PR.

# Changes

Removed the worker copying script and any associated JunoBuild workers in the build process.


AntonioVentilii (approved):
LGTM tks!

PR #5831, opened 2025-04-14
feat(frontend): add address book menu entry

FlavioInacta:
# Motivation

Adds a menu entry to be able to open the address book

# Changes

Adds the entry to the menu with translations and a new lucide icon.

# Tests
Screenshot_2025-04-14_09-03-46

### DIFF
--- a/src/frontend/src/lib/components/core/Menu.svelte
+++ b/src/frontend/src/lib/components/core/Menu.svelte
@@ -94,7 +94,7 @@
 				testId={NAVIGATION_MENU_ADDRESS_BOOK_BUTTON}
 				on:click={modalStore.openAddressBook}
 			>
-				<IconUserSquare size="20	"></IconUserSquare>
+				<IconUserSquare size="20" />
 				{$i18n.navigation.text.address_book}
 			</ButtonMenu>
 
--- a/src/frontend/src/lib/components/core/Menu.svelte
+++ b/src/frontend/src/lib/components/core/Menu.svelte
@@ -3,12 +3,14 @@
 	import { nonNullish } from '@dfinity/utils';
 	import { onMount } from 'svelte';
 	import { page } from '$app/stores';
+	import { ADDRESS_BOOK_ENABLED } from '$env/address-book.env';
 	import AboutWhyOisy from '$lib/components/about/AboutWhyOisy.svelte';
 	import MenuAddresses from '$lib/components/core/MenuAddresses.svelte';
 	import SignOut from '$lib/components/core/SignOut.svelte';
 	import IconGitHub from '$lib/components/icons/IconGitHub.svelte';
 	import IconVipQr from '$lib/components/icons/IconVipQr.svelte';
 	import IconShare from '$lib/components/icons/lucide/IconShare.svelte';
+	import IconUserSquare from '$lib/components/icons/lucide/IconUserSquare.svelte';
 	import LicenseLink from '$lib/components/license-agreement/LicenseLink.svelte';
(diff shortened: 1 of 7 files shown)


src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -82,6 +85,20 @@
...
+			<Hr />
+		{/if}
+
+		{#if ADDRESS_BOOK_ENABLED}
+			<ButtonMenu
+				ariaLabel={$i18n.navigation.alt.address_book}
+				testId={NAVIGATION_MENU_ADDRESS_BOOK_BUTTON}
+				on:click={modalStore.openAddressBook}
+			>
+				<IconUserSquare size="20	"></IconUserSquare>
```
AntonioVentilii, line 97:
```suggestion
				<IconUserSquare size="20" />
```

src/frontend/src/lib/components/icons/lucide/IconUserSquare.svelte
```diff
@@ -0,0 +1,20 @@
...
+	xmlns="http://www.w3.org/2000/svg"
+	width={size}
+	height={size}
+	viewBox="0 0 24 24"
+	fill="none"
+	stroke="currentColor"
+	stroke-width="2"
+	stroke-linecap="round"
+	stroke-linejoin="round"
+	class="lucide lucide-square-user-icon lucide-square-user"
```
AntonioVentilii, line 16: where are this classes defined?

src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -82,6 +85,20 @@
...
+				{$i18n.navigation.text.address_book}
+			</ButtonMenu>
+
+			<Hr />
```
AntonioVentilii, line 101: Maybe a good time to add some unittests for this component?

src/frontend/src/lib/components/icons/lucide/IconUserSquare.svelte
FlavioInacta: I just copied the svg from https://lucide.dev/icons/
There are other icons with same kind of classes. I think they are defaults if you want to select some icons.

src/frontend/src/lib/components/core/Menu.svelte
FlavioInacta: Didn't even see that there are any FE tests. Never worked on a FE project where the tests are in a spearate folder. ;)

I updated the tests.

src/frontend/src/lib/components/icons/lucide/IconUserSquare.svelte
AntonioVentilii: i think we don't use the lucide classes in OISY (they must be installed with the package), so, these can be removed

I will remove the rest too, for consistency:
- https://github.com/dfinity/oisy-wallet/pull/5845

src/frontend/src/lib/components/core/Menu.svelte
AntonioVentilii: Thank you! We can always move them in each respective folder if you prefer, but we should do it for all, for consistency. BTW, as reference, some repos that have a separate test folder:

- https://github.com/MetaMask/metamask-extension/tree/main/test
- https://github.com/dfinity/nns-dapp/tree/main/frontend/src/tests
- https://github.com/ethers-io/ethers.js/tree/main/src.ts/_tests

But I think there are quite more

PR #5842, opened 2025-04-14
feat(devops): round vitest coverage thresholds to floor integer

AntonioVentilii:
# Motivation

To have some wiggle room and not get bombed with PRs, we round the thresholds to the closest (lower) integer.

# Tests

Run with CI https://github.com/dfinity/oisy-wallet/actions/runs/14444846353 and created PR https://github.com/dfinity/oisy-wallet/pull/5843

### DIFF
--- a/.github/workflows/frontend-update-coverage-thresholds.yml
+++ b/.github/workflows/frontend-update-coverage-thresholds.yml
@@ -38,6 +38,10 @@ jobs:
       - name: Test
         run: npm run test:coverage
 
+      - name: Floor coverage thresholds (remove decimals)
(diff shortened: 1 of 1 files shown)


bitdivine (approved):
Might also consider:
```
awk -i inplace '/(lines|functions|statements|branches):/{$2=$2-1}{print}' vitest.config.ts
```
This will reduce each of the numbers by exactly 1 rather than rounding down, which might give no wiggle room if current coverage is say exactly 69.00%.  This way I can also consistently reduce test coverage, whereas with floor I can't.

PR #5849, opened 2025-04-14
chore(devops): Deploy ckbtc index

bitdivine:
# Motivation
`dfx deploy` should just work

# Changes
- Set default arguments for `ckbtc-index`, so that `dfx deploy ckbtc-index` works by default.

# Tests
Existing CI should suffice


AntonioVentilii (approved):
thank you!

PR #5853, opened 2025-04-15
feat(frontend): adds rowReverse flag to ExternalLink.svelte

BonomoAlessandro:
# Motivation

Instead of defining the `flex-row-reverse` of external links over the `styleClass` we added a `rowReverse` flag.

# Changes

- adds `rowReverse` flag on external link

### DIFF
--- a/src/frontend/src/lib/components/ui/ExternalLink.svelte
+++ b/src/frontend/src/lib/components/ui/ExternalLink.svelte
@@ -17,7 +17,7 @@
 	export let asMenuItem = false;
 	export let asMenuItemCondensed = false;
 	export let asButton = false;
-	export let rowReverse = false;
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/ui/ExternalLink.svelte
```diff
@@ -17,6 +17,7 @@
 	export let asMenuItem = false;
 	export let asMenuItemCondensed = false;
 	export let asButton = false;
+	export let rowReverse = false;
```
AntonioVentilii, line 20: nitpick: i would call it `iconAsLast` or something similar, to be extremely clear

AntonioVentilii (approved):
LGTM tks!

PR #5859, opened 2025-04-15
feat(frontend): load native Ethereum balances only for enabled tokens

AntonioVentilii:
# Motivation

Since we are now using enabled/disabled networks, it makes sense to load the Ethereum native tokens balances ONLY for the enabled tokens.

# Changes

- Add input `tokens` to service `loadEthBalances` that had instead the hardcoded list of native Ethereum tokens otherwise.
- Use enabled tokens derived store in component `LoaderEthBalances`.

# Tests

Since we are here, I created tests for both the service and the component.


roman-nazaruk (approved):
LGTM

PR #5861, opened 2025-04-15
feat(frontend): extract specific parsing function for mainnet-enabled env variables

AntonioVentilii:
# Motivation

To avoid repetition of code, we create an util for parsing env variables that are specific for enabling mainnet.


roman-nazaruk (approved):
LGTM

PR #5864, opened 2025-04-15
refactor(frontend): migrating `LoaderEthBalances` component to Svelte 5

AntonioVentilii:
# Motivation

Using the Svelte 5 [migration guide](https://svelte.dev/docs/svelte/v5-migration-guide) and the [migration script](https://svelte.dev/docs/svelte/v5-migration-guide#Migration-script), we migrate component `LoadeEthBalances` to v5 and improve its tests (especially for rendering the slot.

# Note

According to [documentation](https://svelte.dev/docs/svelte/$effect#Understanding-dependencies), the `$effect` rune does not consider the state changes of elements inside a timeout:

> Values that are read asynchronously — after an `await` or inside a `setTimeout`, for example — will not be tracked.

Meaning that the `debounce` function will not be triggered by changes of the derived stores inside the function that it is called:

```ts
const debounceLoad = debounce(load, 500);   // `load` function has some derived stores inside
```

So, we use a workaround, manually including the dependencies in the `$effect` statement.

### DIFF
--- a/src/frontend/src/icp/schema/ic-token.schema.ts
+++ b/src/frontend/src/icp/schema/ic-token.schema.ts
@@ -1,3 +1,4 @@
+import { TokenGroupPropSchema } from '$lib/schema/token-group.schema';
 import { TokenSchema } from '$lib/schema/token.schema';
 import { CanisterIdTextSchema } from '$lib/types/canister';
 import { CoingeckoCoinsIdSchema } from '$lib/validation/coingecko.validation';
@@ -40,4 +41,5 @@ export const IcTokenWithoutIdSchema = IcTokenSchema.omit({ id: true }).strict();
 
 export const IcCkTokenSchema = IcTokenSchema.merge(IcCkMetadataSchema.partial());
 
-export const IcCkInterfaceSchema = IcInterfaceSchema.merge(IcCkMetadataSchema);
+export const IcCkInterfaceSchema =
+	IcInterfaceSchema.merge(IcCkMetadataSchema).merge(TokenGroupPropSchema);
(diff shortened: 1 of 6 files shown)


src/frontend/src/eth/components/loaders/LoaderEthBalances.svelte
```diff
@@ -1,11 +1,22 @@
...
 	import { enabledErc20Tokens } from '$lib/derived/tokens.derived';
 
+	interface Props {
+		children?: Snippet;
+	}
+
+	let { children }: Props = $props();
+
 	const load = async () => {
+		if (isNullish($ethAddress)) {
```
DenysKarmazynDFINITY, line 16: Wondering why we now need to have this check?

AntonioVentilii: Good question: I included the tests for the nullish checks of the ETH address and it was failing with the $effect hook. While the old version apparently wasn't (tried both ways...). I am not really sure if maybe there is some concurrency race that wasn't there before. In any case it is even a good check to have.

Shall I add it in another PR?

AntonioVentilii: Actually you are right, better in another PR:

- https://github.com/dfinity/oisy-wallet/pull/5917

DenysKarmazynDFINITY: thanks for the explanation and splitting the changes!

DenysKarmazynDFINITY (approved):
Approving already so you can proceed after rebasing this PR with the changes from https://github.com/dfinity/oisy-wallet/pull/5917

PR #5878, opened 2025-04-15
refactor(frontend): use test util return for sol-wallet scheduler tests

AntonioVentilii:
# Motivation

We are going to add the vitest linter (PR https://github.com/dfinity/oisy-wallet/pull/5797), but it raises quite a few issues. So, in preparation, for now we apply the rule [vitest/no-duplicate-hooks](https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/no-duplicate-hooks.md) on the sol-wallet scheduler tests.

We will make the test util return three callback functions: one for the `beforeEach` hook, one for the `afterEach` hook and one that contains the tests.

# Changes

- Create interface for the test utils.
- Make `testWorker` in the `sol-wallet.scheduler` tests return `setup`, `teardown` and `tests` functions.
- Use the functions when calling `testWorker`.

# Tests

Current tests work as before.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5880, opened 2025-04-15
refactor(frontend): use test util return for ic-wallet worker tests

AntonioVentilii:
# Motivation

We are going to add the vitest linter (PR https://github.com/dfinity/oisy-wallet/pull/5797), but it raises quite a few issues. So, in preparation, for now we apply the rule [vitest/no-duplicate-hooks](https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/no-duplicate-hooks.md) on the ic-wallet-balance-and-transactions worker tests.

We will make the test util return three callback functions: one for the `beforeEach` hook, one for the `afterEach` hook and one that contains the tests.

# Changes

- Create interface for the test utils.
- Make all init test utils in the `ic-wallet-balance-and-transactions.worker` tests return `setup`, `teardown` and `tests` functions.
- Use the functions when calling each test.

# Tests

Current tests work as before.


src/frontend/src/tests/types/utils.ts
```diff
@@ -0,0 +1,5 @@
+export interface TestUtil {
```
DenysKarmazynDFINITY, line 1: This is the same interface we added in another PR, right?

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/tests/types/utils.ts
AntonioVentilii: yep, i had to duplicate it for all the similar PRs

PR #5882, opened 2025-04-15
refactor(frontend): Manage tokens modal

daviddecentage:
# Motivation

According to new figma designs, we adjust the manage tokens modal.

# Changes

- Updated `ManageTokens` and `ManageTokensModal` to use `ModalTokensList`
- Added translations
- Removed `testId` for removed button, added general close modal `testId` and adjusted tests

# Tests

https://github.com/user-attachments/assets/f1fd872d-38d7-4050-9f0b-8ae6704a1c14


src/frontend/src/lib/components/manage/ManageTokens.svelte
```diff
@@ -51,55 +50,42 @@
 		exchangesStaticData = nonNullish($exchanges) ? { ...$exchanges } : undefined;
 	});
 
-	let allTokensForSelectedNetwork: TokenToggleable<Token>[] = $derived(
-		filterTokensForSelectedNetwork([$allTokens, $selectedNetwork, $pseudoNetworkChainFusion])
-	);
-
-	let allTokensSorted: Token[] = $derived(
+	let allTokensSorted = $derived(
```
AntonioVentilii, line 53:
```suggestion
	let allTokensSorted: Token[] = $derived(
```

src/frontend/src/lib/components/manage/ManageTokens.svelte
```diff
@@ -1,44 +1,43 @@
...
-	import { filterTokensForSelectedNetwork } from '$lib/utils/network.utils';
-	import { filterTokens, pinEnabledTokensAtTop, sortTokens } from '$lib/utils/tokens.utils';
+	import { pinEnabledTokensAtTop, sortTokens } from '$lib/utils/tokens.utils';
 	import SolManageTokenToggle from '$sol/components/tokens/SolManageTokenToggle.svelte';
 	import type { SplTokenToggleable } from '$sol/types/spl-token-toggleable';
 	import { isTokenSplToggleable } from '$sol/utils/spl.utils';
 	import { isSolanaToken } from '$sol/utils/token.utils';
 
-	let { initialSearch, infoElement }: { initialSearch?: string; infoElement?: Snippet } = $props();
+	let { initialSearch, infoElement }: { initialSearch: string | undefined; infoElement?: Snippet } =
```
AntonioVentilii, line 39: curious on this change: you want that `initialSearch` is provided always even if undefined? isn't it the same for `initialSearch?: string` ? or there is no enforcement?

src/frontend/src/lib/constants/test-ids.constants.ts
```diff
@@ -117,7 +118,6 @@ export const SWAP_TOKENS_MODAL = 'swap-tokens-modal';
 export const MANAGE_TOKENS_MODAL = 'manage-tokens-modal';
 export const MANAGE_TOKENS_MODAL_BUTTON = 'manage-tokens-modal-button';
 export const MANAGE_TOKENS_MODAL_SAVE = 'manage-tokens-modal-save';
-export const MANAGE_TOKENS_MODAL_CLOSE = 'manage-tokens-modal-close';
```
AntonioVentilii, line 120: just to be sure, have you checked that the E2E tests still work with this change?

src/frontend/src/lib/constants/test-ids.constants.ts
```diff
@@ -9,6 +9,7 @@ export const LOGIN_BUTTON = 'login-button';
 export const MAX_BUTTON = 'max-button';
 
 export const LOADER_MODAL = 'loader-modal';
+export const BUTTON_MODAL_CLOSE = 'close-modal';
```
AntonioVentilii, line 12: isn't it better a more specific name? meaning, usually it is better to specify the "context" in which the modal is used, so that we don't use the same generic test id in multiple tests.

I would leave something like `MANAGE_TOKENS_MODAL_CLOSE`

AntonioVentilii (commented):
To facilitate the review, it would be better to split this PR in smaller scopes, for example, removing the first two elements in two separate PRs, then maybe wrapping in the `if` condition, or even the usage of `ModalTokensList`, or the introduction of the the context  with the change of the filter logic

src/frontend/src/lib/components/tokens/ModalTokensList.svelte
```diff
@@ -68,15 +68,13 @@
...
-						</p>
-					{/if}
-				</p>
+				{#if noResults}
+					{@render noResults()}
+				{:else}
+					<p class="text-primary">
+						{$i18n.core.text.no_results}
+					</p>
+				{/if}
```
AntonioVentilii, line 77: this change is related to this PR?

daviddecentage: No Ill fix it in another, it will be merged before this

src/frontend/src/lib/constants/test-ids.constants.ts
daviddecentage: Havent checked, we can run the e2e before merging

src/frontend/src/lib/constants/test-ids.constants.ts
daviddecentage: The thing is, new design doesnt have a button "Close" anymore. So to fix the tests that already existed, I close the modal by the generic Gix testId

src/frontend/src/lib/components/manage/ManageTokens.svelte
daviddecentage: I think `initialSearch?: string` is equivalent to `initialSearch: string | undefined`, but yeah lets keep it consistent and revert this change, wasnt intended

src/frontend/src/lib/constants/test-ids.constants.ts
daviddecentage: This is meant for the generic modal close icon on the top right. The close button in the bottom toolbar gets removed, to fix the tests I go for the Gix modal default close icon now

AntonioVentilii (approved):
LGTM tks!

PR #5886, opened 2025-04-16
feat(frontend): add crypto utilities for pow

DecentAgeCoder:
# feat(frontend): add crypto utilities for pow

## Motivation

Adds frontend hashing and encoding helpers used by PoW services and workers.

## Changes

Includes real code changes for scope 3 split from pow.changes.diff.

## Tests

Covered by scope-local or dependent integration tests.


src/frontend/src/lib/utils/crypto.utils.ts
```diff
@@ -0,0 +1,25 @@
...
+		.join('');
+}
+
+/**
+ * Combines the hashing and hex conversion of a string into a single function.
+ */
+export async function hashToHex(input: string): Promise<string> {
+	const hashBuffer = await sha256(input);
+	return bufferToHex(hashBuffer);
+}
```
roman-nazaruk, line 25: I think we can reuse methods from dfinity/utils instead of writing our own utility methods for these cases. Here's the documentation—hope it helps. @AntonioVentilii WDYT?

Documentation: https://github.com/dfinity/ic-js/blob/main/packages/utils/README.md

peterpeterparker (commented):
Agree with @roman-nazaruk 

1. Aren’t there already functions in `@dfinity/utils` that could be reused?

2. If not, aren’t there functions in `@dfinity/oisy-wallet-signer` that already fulfill those requirements and could be moved to `@dfinity/utils` for use in both the Wallet and the Signer libraries and generally speaking in the eco-system?

3. If none to both, shouldn't the new utils be added to `@dfinity/utils`?

src/frontend/src/lib/utils/crypto.utils.ts
```diff
@@ -0,0 +1,25 @@
...
+	return await crypto.subtle.digest('SHA-256', textEncoder.encode(input));
+}
+
+/**
+ * Converts an ArrayBuffer to its hexadecimal string representation.
+ */
+export function bufferToHex(buffer: ArrayBuffer): string {
+	return Array.from(new Uint8Array(buffer))
+		.map((byte) => byte.toString(16).padStart(2, '0'))
+		.join('');
```
AntonioVentilii, line 16: can this be done by `Buffer.toString('hex')` ?

src/frontend/src/lib/utils/crypto.utils.ts
AntonioVentilii: Agreed! i think there are some methods, but not sure which one to use. For example `uint8ArrayToHexString`, maybe have a look there

src/frontend/src/lib/utils/crypto.utils.ts
```diff
@@ -0,0 +1,25 @@
+const textEncoder = new TextEncoder();
```
AntonioVentilii, line 1: you can put this inside the function that uses it right? or is it better here for performance?

src/frontend/src/lib/utils/crypto.utils.ts
DecentAgeCoder: I agree that manually performing the serialization from an ArrayBuffer to a hex string might no be best approach. I will check:
1. whether there is a library available that directly generates a sha-256 as a string
2. or if there is a replacement for bufferToHex(..) i will use that

peterpeterparker:
@DecentAgeCoder @AntonioVentilii I don't think we should review the PR at this point. I suggest first to start with the comment I left above https://github.com/dfinity/oisy-wallet/pull/5886#pullrequestreview-2771621602

src/frontend/src/lib/utils/crypto.utils.ts
DecentAgeCoder: I just checked the function you proposed. You nailed it - uint8ArrayToHexString(..) is exactly what is required as a replacement.

DecentAgeCoder:
@peterpeterparker Yes, that was actually a good point. I checked the source code at: https://github.com/dfinity/ic-js/blob/main/packages/utils/src/utils/arrays.utils.ts#L60

That said, I believe the performance aspect shouldn't be underestimated here. Since this function is called in a loop thousands of times, any inefficiencies can quickly add up. 
The IC code uses string concatenation inside a loop, which is likely inefficient, if strings are also immutable in JavaScript are immutable. Additionally, there's a type conversion involved, which introduces some overhead. The reduce in this case, will most likely negatively impact the performance. Yes, we could create a PR to have that changed. I haven’t looked into @dfinity/oisy-wallet-signer, you probably know more about this implementation. What do you you recommend?

peterpeterparker (commented):
Woot? Sorry I don't get your answer.

A. Are you saying that a feature provided in this PR potentially duplicates one that already exists in `@dfinity/utils` or `@dfinity/oisy-wallet-signer`, but might be more performant as provided here?

Or

B. Are you saying that changing the implementation would result in lower performance?

If it's the former (A), cool! I'm always in favor of improvements. Let's enhance (or add) the utilities in the libraries so the entire IC ecosystem and our dapps can benefit from it.

If it's the latter (B), I never suggested modifying the current implementation. I suggested either using the existing utils or contributing the new ones to the libraries so the whole ecosystem can benefit from them.

DecentAgeCoder:
I meant B. Replacing my current implementation with uint8ArrayToHexString from @dfinity/utils would degrade performance. Let's note this as a TODO for later, so we proceed, ok?

peterpeterparker:
ChatGPT large sample:

> bufferToHex: 201.553 ms
> uint8ArrayToHexString: 193.907 ms
> Outputs are equal: true

```
const { performance } = require('perf_hooks');

function bufferToHex(buffer) {
  return Array.from(new Uint8Array(buffer))
    .map((byte) => byte.toString(16).padStart(2, '0'))
    .join('');
}

function uint8ArrayToHexString(bytes) {
  if (!(bytes instanceof Uint8Array)) {
    bytes = Uint8Array.from(bytes);
  }
  return bytes.reduce(
    (str, byte) => str + byte.toString(16).padStart(2, '0'),
    ''
  );
}

// Generate large buffer
const largeBuffer = new Uint8Array(256 * 10000).map((_, i) => i % 256);
const arrayBuffer = largeBuffer.buffer;

// Warm up
bufferToHex(arrayBuffer);
uint8ArrayToHexString(largeBuffer);

// Benchmark bufferToHex
const t1 = performance.now();
const res1 = bufferToHex(arrayBuffer);
const t2 = performance.now();
console.log(`bufferToHex: ${(t2 - t1).toFixed(3)} ms`);

// Benchmark uint8ArrayToHexString
const t3 = performance.now();
const res2 = uint8ArrayToHexString(largeBuffer);
const t4 = performance.now();
console.log(`uint8ArrayToHexString: ${(t4 - t3).toFixed(3)} ms`);

console.log(`Outputs are equal: ${res1 === res2}`);
```

peterpeterparker:
ChatGPT small sample:

> bufferToHex: 0.062333 ms
> uint8ArrayToHexString: 0.040791 ms
> Outputs are equal: true

```
const { performance } = require('perf_hooks');

function bufferToHex(buffer) {
  return Array.from(new Uint8Array(buffer))
    .map((byte) => byte.toString(16).padStart(2, '0'))
    .join('');
}

function uint8ArrayToHexString(bytes) {
  if (!(bytes instanceof Uint8Array)) {
    bytes = Uint8Array.from(bytes);
  }
  return bytes.reduce(
    (str, byte) => str + byte.toString(16).padStart(2, '0'),
    ''
  );
}

// Small buffer (e.g., 32 bytes)
const smallBuffer = new Uint8Array(32).map((_, i) => i);
const arrayBuffer = smallBuffer.buffer;

// Benchmark bufferToHex
const t1 = performance.now();
const res1 = bufferToHex(arrayBuffer);
const t2 = performance.now();
console.log(`bufferToHex: ${(t2 - t1).toFixed(6)} ms`);

// Benchmark uint8ArrayToHexString
const t3 = performance.now();
const res2 = uint8ArrayToHexString(smallBuffer);
const t4 = performance.now();
console.log(`uint8ArrayToHexString: ${(t4 - t3).toFixed(6)} ms`);

console.log(`Outputs are equal: ${res1 === res2}`);
```

peterpeterparker:
Long story short but happy to be provden wrong:

- This PR introduces code duplication
- It appears to be less performant than using the existing utilities
- Is not beneficial for the eco-system and our dapps (if any new utilities is introduced)

Therefore I think it should be updated.

src/frontend/src/lib/utils/crypto.utils.ts
DecentAgeCoder: I would not move the 'instantiation' into the function. This avoids repeated allocation and setup overhead - especially the context of the challenge loop

src/frontend/src/lib/utils/crypto.utils.ts
```diff
@@ -0,0 +1,22 @@
+import { uint8ArrayToHexString } from '@dfinity/utils';
+
+const textEncoder = new TextEncoder();
+
+/**
+ * Hashes the given input string using the SHA-256 algorithm.
+ */
+export async function sha256(input: string): Promise<ArrayBuffer> {
```
peterpeterparker, line 8: - This function doesn't need to be exposed.
- Nitpick: For consistency with the codebase, could you use an arrow function (`const sha256 = () => ...`) if you don't mind?
- The JSDoc doesn't describe the parameter—was that intentional?

src/frontend/src/lib/utils/crypto.utils.ts
```diff
@@ -0,0 +1,22 @@
...
+	return await crypto.subtle.digest('SHA-256', textEncoder.encode(input));
+}
+
+/**
+ * Combines the hashing and hex conversion of a string into a single function.
+ */
+export async function hashToHex(input: string): Promise<string> {
```
peterpeterparker, line 15: Same as previous comments plus:

- This duplicates existing code in the Oisy Wallet Signer library that has been recently added by @roman-nazaruk .  Therefore,  I think we should move the implementation to `@dfinity/utils` and reuse it both in the signer library and in the wallet.  I think it should either be resolved right after we merge this PR —in that case, please add a TODO — or already now. I prefer the former as it makes this PR irrelevant but, your call.

https://github.com/dfinity/oisy-wallet-signer/blob/ae546e39c3f209e875ca09edfb2affc790e32e73/src/utils/crypto.utils.ts#L7

src/frontend/src/lib/utils/crypto.utils.ts
DecentAgeCoder: Great input! Yeap, I fully agree - I’ll update the code accordingly. I missed that parameter in the JSDoc, I will add it as well.

src/frontend/src/lib/utils/crypto.utils.ts
DecentAgeCoder: I agree, the hash operation is a core, security-related functionality and should therefore be moved to @dfinity/utils. That said, I’d like to proceed for now and add a TODO. @roman-nazaruk, could you take on this task (providing a dedicated function for the outlined code line)  since you’re already working on @dfinity/utils?

peterpeterparker: For sure a TODO works for me, as long as we resolve it right afterwards.

DecentAgeCoder: @DecentAgeCoder Not sure whether I understood correctly. In this PR the sha256 is currently imported in  crypto.utils.spec.ts. So I will at least have to export this one.

peterpeterparker: We generally avoid leaking functions, objects, etc. when it's not necessary - i.e. we try not to expose code solely for testing purposes.

In this particular case, it's useful I think because the code will eventually shortly be deleted, so not exposing it helps prevent others from relying on a function that is already considered "deprecated".

DecentAgeCoder: Good, makes sense! I will remove the test and remove the export for hashToHex

peterpeterparker: Thx a lot 👍

DecentAgeCoder: @peterpeterparker Now the changes should be complete

src/frontend/src/lib/utils/crypto.utils.ts
peterpeterparker: For consistency with the code base, our general functional approach, and (admittedly opinionated) because I believe it's cleaner not to declare anything at the root of the module unless necessary, I would prefer, unless a benchmark proves it's significantly more efficient, to have it scoped, as Antonio mentioned.

src/frontend/src/lib/utils/crypto.utils.ts
DecentAgeCoder: Yes, I deleted them on purpose (as mentioned in previous conversation) :wink: Should I revert this change and remove the export again?

peterpeterparker: Ah, I missed that part. I'm a bit hesitant I have to say. This piece of code, this PR, will be deleted anyway, so we could potentially be flexible because of that but generally speaking, it would need a test.

By the way, @roman-nazaruk still hasn't confirmed whether he'll move the utils to ic-js.

DecentAgeCoder: Okey dokey: I will revert the crypto.utils.test file to test both functions defined in crypto.utils.spec.ts **and export both functions** accordingly.

roman-nazaruk: Yeah, no problem on my end. But just to confirm — am I understanding correctly from this conversation that we need to add two new utility methods to ic-js?

1) sha256

2) hashToHex

And then reuse these two utilities in oisy-wallet, updating the logic in oisy-wallet-signer for generateHash?

DecentAgeCoder: Yes, then I can use your corresponding function for hashToHex(..) directly, and therefore avoid having to call bufferToHex(..) This would provide a dedicated function for `sha256` optional on my side.

peterpeterparker: > and export both functions accordingly.

As suggested [here](https://github.com/dfinity/oisy-wallet/pull/5886#discussion_r2047187067), only what is used should be exposed.

> Yeah, no problem on my end.

Cool

> And then reuse these two utilities in oisy-wallet, updating the logic in oisy-wallet-signer for generateHash?

Let's wait for the final review of this PR. I added a comment [here](https://github.com/dfinity/oisy-wallet/pull/5886#discussion_r2048457659) that should be still discussed.

DecentAgeCoder: @peterpeterparker Ok: I have re-introduced a test for hashToHex in crypto.utils.ts and exported hashToHex (that will also used by the worker service).

src/frontend/src/tests/lib/utils/crypto.utils.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+
+describe('crypto.utils', () => {
+	describe('hashToHex', () => {
+		it('should hash a string and return its hex representation', async () => {
+			const input = 'hash-to-hex-test';
+			const hex = await hashToHex(input);
+
+			expect(typeof hex).toBe('string');
+
+			expect(hex.length).toBe(64);
```
peterpeterparker, line 12: Any particular reason not to assert the hex value itself?

```
expect(hex).toEqual("....")
```

DecentAgeCoder: You’re right. I’ll add a test to verify the expected hash for a given string, along with a few additional tests. Please note that I won’t use `sha256` to generate the hash dynamically,  I will hardcode the expected hex strings instead.

peterpeterparker:
Lots of noise here, so just to summarize:

1. We still have this question to address: https://github.com/dfinity/oisy-wallet/pull/5886#discussion_r2048457659.

I got the feeling there isn't enough performance reason to keep the text encoder at the root - but happy to be proven wrong.
This is an important question to close the loop, as the answer will shape our next steps.

2. Once it’s resolved, we can move forward:

a. Move forward and approve the PR
b. Finally define which utils need to be created or moved to `@dfinity/utils`

cc @DecentAgeCoder @roman-nazaruk @AntonioVentilii

AntonioVentilii:
@peterpeterparker agreed on both points; awaiting changes on the first one to proceed with the final review

src/frontend/src/lib/utils/crypto.utils.ts
DecentAgeCoder: @peterpeterparker  Sorry for the long delay. I have just tested both versions with a simple performance test (written in typescript):


```
Running SHA-256 performance test with 10,000 iterations...

Version 1 (TextEncoder at root): 185.72 ms  
Version 2 (TextEncoder inside function): 177.80 ms
```
It seems that the performance is slightly better when declaring the TextEncoder inside the function (which was surprising to me). @AntonioVentilii, your remark regarding the performance has been proven correct. Based on this observation, and for the sake of consistency, I will ensure that the TextEncoder is declared within the function body moving forward.

peterpeterparker (changes requested):
Thanks for the iterations. We've ultimately reached a point where the code provided in this PR is a duplication of the existing code in the OISY Wallet Signer’s [generateHash](https://github.com/dfinity/oisy-wallet-signer/blob/ae546e39c3f209e875ca09edfb2affc790e32e73/src/utils/crypto.utils.ts#L4) function**.

Therefore, I suggest we **don't** proceed with this PR.

Instead, we should move the utility from the signer to ic-js`@dfinity/utils`, which would require very little time investment.

**PS: excluding the additional `JSON.stringify` which has probably little performance overrhead given the input being a `string` and the use of the `jsonReplacer`, which to my knowledge has no performance impact if no keys match during stringification.

peterpeterparker:
There I copied the function to ic-js: https://github.com/dfinity/ic-js/pull/889

peterpeterparker:
- https://github.com/dfinity/ic-js/pull/890
- https://github.com/dfinity/ic-js/pull/891

peterpeterparker:
Replaced by #6006.

This PR can be closed and related branch deleted.

PR #5890, opened 2025-04-16
feat(frontend): integrate pow into loader and guard logic

DecentAgeCoder:
## Motivation

This Pull Request introduces a conditional feature toggle (POW_FEATURE_ENABLED) to support or bypass Proof-of-Work-related functionality. Modifications are made to reflect this toggle in address validation, component hierarchy, and loading logic.

## Changes

- AddressGuard.svelte:
Introduced POW_FEATURE_ENABLED usage to conditionally skip initSignerAllowance logic.
- Loaders.svelte:
Updated component hierarchy to include a new wrapper component, PowProtector, for Proof-of-Work protection.
- loader.services.ts:
Incorporated POW_FEATURE_ENABLED to bypass conditional logic for invoking initSignerAllowance in the loading process.


src/frontend/src/lib/components/guard/AddressGuard.svelte
```diff
@@ -18,11 +19,13 @@
 	let signerAllowanceLoaded = false;
 
 	const loadSignerAllowanceAndValidateAddresses = async () => {
```
AntonioVentilii, line 21: maybe now can be renamed?

AntonioVentilii (commented):
Before anything, may you please provide an explanation in the description on why it will be ok that the component AddressGuard is agnostic to being allowed to sign? Like we discussed offline.

Most important part is: we are removing the check for the signer-allowance-success. Why?

src/frontend/src/lib/components/guard/AddressGuard.svelte
DecentAgeCoder: Yes it had been renamed

AntonioVentilii: i mean, if you are going to remove the signer allowance, you can rename this function no?

My understanding is that once the feature is complete, the feature flag is gone, so that part inside the IF condition disappear. Or not?

DecentAgeCoder: I recommend to leave the Feature flag, so that we can disable it quickly if there are issues with the PoW in production (e.g. to much CPU overhead, or not enough Cycles per time period are allowance (fine-tuning of settings)).

AntonioVentilii:
Superseded by later PoW iterations and the current PoW code in main. Closing as obsolete.

PR #5891, opened 2025-04-16
feat(frontend): implement pow protector ui

DecentAgeCoder:
# Motivation

Adding a new Svelte component to manage Proof-of-Work (PoW) protection functionality.

## Changes

- Introduced a new PowProtector.svelte component in the frontend. Includes logic to initialize, start, and stop a PoW protection worker when POW_ENABLED is true

## Tests

No specific tests are provided in this PR for the new component.


src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -0,0 +1,21 @@
...
+			powWorker = await initPowProtectorWorker();
+			// Start the worker
+			powWorker.start();
+		});
+
+		onDestroy(() => powWorker?.stop());
+	}
+</script>
+
+<slot />
```
AntonioVentilii, line 21: can you do it in Svelte 5 ?

DecentAgeCoder: What do you want me exactly to migrate? Should I migrate the code to use $effect(..) instead of onMount(..) and onDestroy(..) or use $derived(..). I do not see that these runes are currently used in any other svelte component. Also I do not see any changes that would be required by svelte 5 for the slot element.

AntonioVentilii: I updated it with the last commit, have a check

AntonioVentilii (approved):
Ideally we should have tests too, but it can be for another PR, please make sure we create it

LGTM tks!

PR #5892, opened 2025-04-16
feat(frontend): extend message schemas for pow

DecentAgeCoder:
## Motivation
This PR introduces the required schema changes which are used to integrate the pow scheduler used  for the 'pow functionality'

## Changes

Added new enum values and schema registration for pow protection

## Tests

The PR does not provide new tests explicitly, but existing schema and types should align with unit tests if they are already implemented.


AntonioVentilii (commented):
I think in this particular case, it is better to do this changes together with their usage. For example, if you create a skeleton of the scheduler/worker, or if you develop each message flow in separate PRs

AntonioVentilii:
Superseded by:

- https://github.com/dfinity/oisy-wallet/pull/5887

PR #5893, opened 2025-04-16
feat(frontend): add pow environment config

DecentAgeCoder:
# Motivation

This PR introduces configurations for enabling Proof-of-Work (PoW) functionality on the frontend, likely as part of a security or rate-limiting feature.

# Changes

Added a new environment configuration file pow.env.ts in the frontend.
Defined constants:
- POW_ENABLED: Determines if Proof-of-Work is enabled based on environment variables.
- POW_CHALLENGE_INTERVAL_MILLIS: Sets the challenge interval in milliseconds, with a default of 120,000 ms.

# Tests

No test implementations were provided in the PR. Test validation for these configurations might be required, especially for edge cases like invalid or missing environment variables.


src/frontend/src/env/pow.env.ts
```diff
@@ -0,0 +1,4 @@
+export const POW_ENABLED = JSON.parse(import.meta.env.VITE_POW_ENABLED ?? false) === true;
```
AntonioVentilii, line 1: we have a new util called `parseBoolEnvVar`

P.S. let's call it `POW_FEATURE_ENABLED`, to be even more clear that it is a feature flag (i suppose is temporary)

src/frontend/src/env/pow.env.ts
```diff
@@ -0,0 +1,4 @@
...
+export const POW_CHALLENGE_INTERVAL_MILLIS: number = JSON.parse(
```
AntonioVentilii, line 2: why this as env variable? let's use it as constants in the `constants` folder

src/frontend/src/env/pow.env.ts
DecentAgeCoder: Yes, good point

src/frontend/src/env/pow.env.ts
DecentAgeCoder: Yes

src/frontend/src/env/pow.env.ts
```diff
@@ -0,0 +1 @@
+export const POW_FEATURE_ENABLED = JSON.parse(import.meta.env.VITE_POW_ENABLED ?? false) === true;
```
AntonioVentilii, line 1:
```suggestion
export const POW_FEATURE_ENABLED = parseBoolEnvVar(import.meta.env.VITE_POW_FEATURE_ENABLED)
```

AntonioVentilii (approved):

PR #5894, opened 2025-04-16
refactor(frontend): extract type class for IcWalletBalanceAndTransactionsScheduler

AntonioVentilii:
# Motivation

To help with the refactoring of `IcWalletBalanceAndTransactionsScheduler`, we need first to extract a single type for the result of the balance-and-transacitons query.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5897, opened 2025-04-16
refactor(frontend): Modal token list reusability

daviddecentage:
# Motivation

For the upcoming manage tokens modal refactoring, we adjust the `ModalTokenList` to make it reusable for the new use case.

# Changes

- Added slot for no result message
- Added slot for single token list item


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5901, opened 2025-04-16
feat(frontend): Add generic input component with reset button

FlavioInacta:
# Motivation

There are some non generic approach for `reset` functionality for input fields (see InputSearch and InputTextWithAction components). I need the reset funcitonality in another field so I decided to add a generic component. Maybe the mentioned components will be able to build on this new `Input` component.

# Changes

Adds a generic Input component based on the gix library which supports a `reset` button.

# Screenshots
Without value:
Screenshot_2025-04-16_14-54-14
With value:
Screenshot_2025-04-16_14-53-57


AntonioVentilii (commented):
We have something similar in `SlidingInput` created by @daviddecentage , would it be possible to re-use the same or have similar parts?

Furthermore, the `autofocus` is an interesting feature to have directly in GIX components I think, so let's try to add it there directly

FlavioInacta:
I would not mix the two. The new component should be very generic and the sliding input is very specific for the needed usecase.

AntonioVentilii:
@FlavioInacta Not really sure to be honest: why would the two of them be mixed? I see quite a few similarities... UI components are re-usable by definition. Since we risk to duplicate the code, I suggest to at least study if there is re-usablity between the two components.

RE the autofocus: it can find a good place to be a new feature in the GIX component input

FlavioInacta:
I thought you want to combine the sliding input and the new one into one component. 

And yes. I would like to add the new input to the Gix-Library but not now. I think there will be more changes to it in near future. If all the features are added to the Input I will add it to the GIX Library. Is there somewhere a backlog for the GIX Library? The I can add it there.

FlavioInacta:
Ok, I'll put it into the GIX library.

FlavioInacta:
Added two PRs to gix libary:
- Autofocus: https://github.com/dfinity/gix-components/pull/622
- Clear button: https://github.com/dfinity/gix-components/pull/623

AntonioVentilii:
@FlavioInacta my doubt is still the same of comment https://github.com/dfinity/oisy-wallet/pull/5901#issuecomment-2814650384


> Not really sure to be honest: why would the two of them be mixed? I see quite a few similarities... UI components are re-usable by definition. Since we risk to duplicate the code, I suggest to at least study if there is re-usablity between the two components.

Why not adapting the existing code?

src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -0,0 +1,36 @@
...
+	}: InputProps & GixInputProps = $props();
+
+	const reset = () => {
+		value = undefined;
+	};
+</script>
+
+<GixInput {...props} bind:value>
+	<svelte:fragment slot="inner-end">
+		{#if value && showResetButton}
```
AntonioVentilii, line 26: `nonNullish(value)` or maybe even `notEmptyString(value)` ?

src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -0,0 +1,36 @@
+<script lang="ts">
+	import { Input as GixInput, IconClose } from '@dfinity/gix-components';
+	import { type ComponentProps } from 'svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+
+	type GixInputProps = ComponentProps<GixInput>;
+	type InputProps = {
+		showResetButton?: boolean;
+		resetButtonAriaLabel?: string;
```
AntonioVentilii, line 9: is `resetButtonAriaLabel` needed in another PR?

src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -0,0 +1,36 @@
+<script lang="ts">
+	import { Input as GixInput, IconClose } from '@dfinity/gix-components';
+	import { type ComponentProps } from 'svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+
+	type GixInputProps = ComponentProps<GixInput>;
+	type InputProps = {
+		showResetButton?: boolean;
```
AntonioVentilii, line 8: if we are creating this specific input component with a clear button, this prop should not exist: it should be `true` by default everytime, no?

otherwise we could use simply the normal `Input`, no difference

src/frontend/src/tests/lib/components/ui/Input.spec.ts
```diff
@@ -0,0 +1,93 @@
+import Input from '$lib/components/ui/Input.svelte';
+import { fireEvent, render } from '@testing-library/svelte';
+
+describe('Input', () => {
+	const defaultProps = {
+		value: undefined,
+		placeholder: 'Test placeholder',
+		inputType: 'text' as const, // Explicitly type as 'text'
+		name: 'test-input', // Add required name property
```
AntonioVentilii, line 9: why these 2 comments?

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: No, but if we want to reuse this component in `InputSearch`, this should be overwritten.

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: The idea was that it by default behaves like the Input from gix components and if an additional feature (like reset button) is needed, the functionality can be opted in.

But yes, at the moment it makes no sense to use this component when you don't use the reset button.

src/frontend/src/tests/lib/components/ui/Input.spec.ts
FlavioInacta: Removed

src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -0,0 +1,37 @@
...
+	}: InputProps & GixInputProps = $props();
+
+	const reset = () => {
+		value = undefined;
+	};
+</script>
+
+<GixInput {...props} bind:value>
+	<svelte:fragment slot="inner-end">
+		{#if (typeof value === 'string' ? notEmptyString(value) : nonNullish(value)) && showResetButton}
```
AntonioVentilii, line 27: for scalability, WDYT of 

`#if nonNullish(value) && notEmptyString(value.toString()) && showResetButton}`

src/frontend/src/lib/components/ui/Input.svelte
AntonioVentilii: ah, will it have a different aria label based on the scope? for what is see the default would be "Reset input value"

Then maybe, better to remove the default? It is either for each specific use-case or for a generic one, I think. WDYT?

src/frontend/src/lib/components/ui/Input.svelte
AntonioVentilii: So, you are removing it for now? We can always add it back when needed?

src/frontend/src/tests/lib/components/ui/Input.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+		const resetButton = getByRole('button');
+
+		expect(resetButton).toBeInTheDocument();
+	});
+
+	it('does not show reset button when showResetButton is true but value is empty', () => {
+		const props = {
+			...defaultProps,
+			showResetButton: true,
+			value: ''
```
AntonioVentilii, line 49: may you please copy the same test but with `undefined`?

src/frontend/src/tests/lib/components/ui/Input.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+		await fireEvent.input(input, { target: { value: 'test value' } });
+
+		expect(input).toHaveValue('test value');
+	});
+
+	it('shows reset button when showResetButton is true and value exists', () => {
+		const props = {
+			...defaultProps,
+			showResetButton: true,
+			value: 'test value'
```
AntonioVentilii, line 34: may you please copy the same test but with a number?

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: I think it's good like this. Having a sane default value and the possibility to overwrite it with a custom text if needed.

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: I would leave it like this. If I should remove the option, I would rename the component to `InputWithReset`.

I want to use this component in `TextInput`. When it is disabled by default, I can just replace the import and expose the `showResetButton`.

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: Ok for me too. I'll change it.

src/frontend/src/tests/lib/components/ui/Input.spec.ts
FlavioInacta: ok will do

src/frontend/src/tests/lib/components/ui/Input.spec.ts
FlavioInacta: ok, will do

src/frontend/src/lib/components/ui/Input.svelte
AntonioVentilii: We should avoid putting logic that we don't necessarily use: if we are thinking of using it with custom aria label, we should remove the default. Otherwise we use the same arialabel for all use cases.

If in the future we need it to have both, we can always add it.

src/frontend/src/lib/components/ui/Input.svelte
AntonioVentilii: not really sure I understand: this component is specifically created to be a wrapper around GIX `Input` to have a reset button. Why should there be a use-case where we use it without reset button instead of the base GIX `Input`? In that case, for sure it is better to use GIX `Input` and avoid providing unnecessary props to simplify the code, no?

FlavioInacta: I try to explain. I want to add a text input with a reset button. My idea was to just be able to do this: `<InputText showResetButton={true} ....>`.

I could add this to the `InputText` component, but I thought because I will need it on an address input too sorthly, I will make it reusable.

My plan:
- Add a common Input component with the feature
- Use that component in `InputText`
- Use `InputText` with `showResetButton={true}`, but since I don't need a specific aria text I want to add a sane default.
- Reuse that functionality in `InputSearch` do avoid duplicated code (there I need a custom aria text)

AntonioVentilii: i am still not convinced? why do you want to do this `<InputText showResetButton={true} ....>` when you can simply do `<InputText ....>` where `InputText` is this one you created

src/frontend/src/lib/components/ui/Input.svelte
AntonioVentilii, line 27: does it work simplyfing?

```suggestion
		{#if nonNullish(value) && notEmptyString(value.toString()) && showResetButton}
```

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: No, `<InputText ...>` is not the one I create. If I might create a larger PR, I think you would understand.

FlavioInacta: I added a PR to illustrate my point/goal. Allow using a common implementation for autofocus and reset button for `InputText` and `InputSearch` and potentially for all inputs. 

https://github.com/dfinity/oisy-wallet/pull/6219/files

AntonioVentilii: let's proceed, but if the usage of this component is always with `showResetButton`, please remove the prop, make it default to true, and rename the component. We can check after the next PR (that I imagine is the usage of it)

src/frontend/src/lib/components/ui/Input.svelte
AntonioVentilii: Any update here?

src/frontend/src/tests/lib/components/ui/Input.spec.ts
AntonioVentilii: any update here?

src/frontend/src/lib/components/ui/Input.svelte
AntonioVentilii: any update here?

FlavioInacta: Changed condition

src/frontend/src/tests/lib/components/ui/Input.spec.ts
FlavioInacta: Added test for undefined.

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: Added this condition

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: I leave it as I will need it (see mentioned PR). https://github.com/dfinity/oisy-wallet/pull/6219

AntonioVentilii (approved):

PR #5902, opened 2025-04-16
feat(shared): List of token standards

bitdivine:
# Motivation
We would like to be able to tell which tokens an address can be used with.  This can be determined, at one level, by the token standard, which specifies e.g. the address format and can therefore be used to determine whether a given address is at least theoretically valid for a given token.

# Changes
- Ad a list of token standards.

# Tests
None - the list is not used yet.

### DIFF
--- a/src/shared/src/types/token_standard.rs
+++ b/src/shared/src/types/token_standard.rs
@@ -1,0 +1,41 @@
+//! Token standards
+
+pub enum TokenStandard {
+    /// Standard for the Internet Computer
+    ///
+    /// # Used for:
+    /// - ICP (native ICP token)
+    /// - Cycles (Native stablecoin for execution fees)
+    /// - SNS tokens (Standardized DAOs on the Internet Computer)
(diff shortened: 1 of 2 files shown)


src/shared/src/types/token_standard.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// # Used for:
+    /// - SOL (native SOL token)
+    /// - Most tokens on Solana.
+    ///
+    /// - [Solana Token Guide](https://solana.com/docs/tokens)
+    /// - [SPL token documentation](https://spl.solana.com/token)
+    ///
+    /// Note: It seems that Solana tokens are defined by the library rather than by a
+    /// specification, but I may be wrong.
+    Spl,
```
AntonioVentilii, line 40: is it possible to make it SPL? Since it is an acronym for `Solana Program Library`

AntonioVentilii (approved):

PR #5906, opened 2025-04-16
feat(frontend): implements micro transactions handling functions

BonomoAlessandro:
# Motivation

We want to be able to filter out so called `microtransactions` from the transactions list of the activity page. It is important that only `receive` transactions smaller than 0.01 USD are filtered out.

# Changes

- provides functionality to filter micro transactions

### DIFF
--- a/src/frontend/src/lib/components/convert/ConvertContexts.svelte
+++ b/src/frontend/src/lib/components/convert/ConvertContexts.svelte
@@ -1,0 +1,37 @@
+<script lang="ts">
+	import { setContext, type Snippet } from 'svelte';
+	import {
+		CONVERT_CONTEXT_KEY,
+		type ConvertContext,
+		initConvertContext
+	} from '$lib/stores/convert.store';
+	import {
+		initTokenActionValidationErrorsContext,
+		TOKEN_ACTION_VALIDATION_ERRORS_CONTEXT_KEY,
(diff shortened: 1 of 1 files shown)


daviddecentage (approved):
LGTM

peterpeterparker (changes requested):
If you don't mind, I'll go ahead and block this since it has been approved to prevent it being merged while we're currently discussing them on Slack. I foresee issues with the concept.

BonomoAlessandro:
As discussed in this [thread](https://dfinity.slack.com/archives/C072H2FMW1F/p1744872344852309), we are aware of potential issues with this utility, such as using it incorrectly in combination with an infinite loader. When used with an infinite loader, we need to ensure that the transactions are filtered during the loading process, not right before displaying them. Filtering the transactions right before displaying them, could lead to unexpected behavior while scrolling through the transactions.

peterpeterparker (commented):
(this was a test, a comment does not revert my request for change)

peterpeterparker (approved):
I'm approving this PR to revert my request for changes. Fundamentally, I believe this PR shouldn't be merged given that it might lead to issues when used with the infinite scroller which is the idea behind the overall feature, but this was discussed with the team, who wish to proceed anyway which is fine with me.

PR #5909, opened 2025-04-16
feat(frontend): include networks Base and BSC in user networks mapping

AntonioVentilii:
# Motivation

We need to include networks Base and BNB Smart Chain (BSC) in the mapping of user networks.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5911, opened 2025-04-16
refactor(frontend): `userNetworks` store will cover all new variants

AntonioVentilii:
# Motivation

We would like to have a fail-proof mapping of user networks in case new variants appear in the backend. To guarantee it, type-safe, the only way is to consume each variant so that the last value shall be `never` type according to the compiler.

# Changes

- In `userNetworks` derived store substitute the ternary conditions with a series of `if` statements.
- Add a final check for the `never` type.
- Add missing variants (results of tests, see below).

# Tests

The compiler raised an error already because there are new variants for networks Base and BNB Smart Chain (BSC), that were not being mapped. So we included those variants too.


DenysKarmazynDFINITY (approved):
Nice, more readable now too 👍

PR #5912, opened 2025-04-17
feat(frontend): util to define the enabled networks for a specific chain

AntonioVentilii:
# Motivation

In this case, we create an util that defines the enabled networks for a chain, and we use it in the derived stores.

# Changes

- Create util that replicate existing code of `enabledBitcoinNetworks`, `enabledEthereumNetworks` and `enabledSolanaNetworks`.

# Tests

Added new tests.


DenysKarmazynDFINITY (approved):
Same here - 👍

PR #5914, opened 2025-04-17
feat(backend): Validate account ids

bitdivine:
# Motivation
We wish to be able to determine account ID formats

# Changes
- Implement validators for each account ID format
- Parse strings to the first compatible account ID format.

# Tests
Unit tests are included.

### DIFF
--- a/src/shared/src/types/account/conversion.rs
+++ b/src/shared/src/types/account/conversion.rs
@@ -1,0 +1,48 @@
+//! Conversion functions for account identifiers.
+use std::{str::FromStr, string::ParseError};
+
+use super::{BtcAddress, EthAddress, Icrcv2AccountId, SolPrincipal, TokenAccountId};
+
+impl FromStr for TokenAccountId {
+    type Err = ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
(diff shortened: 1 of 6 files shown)


src/shared/src/types/account/conversion.rs
```diff
@@ -0,0 +1,255 @@
...
+
+impl FromStr for TokenAccountId {
+    type Err = ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        SolPrincipal::from_str(s)
+            .map(TokenAccountId::Sol)
+            .or_else(|_| BtcAddress::from_str(s).map(TokenAccountId::Btc))
+            .or_else(|_| EthAddress::from_str(s).map(TokenAccountId::Eth))
+            .or_else(|_| Icrcv2AccountId::from_str(s).map(TokenAccountId::Icrcv2))
```
AntonioVentilii, line 33: `SolPrincipal` missing here?

AntonioVentilii (approved):
Big job! LGTM tks! One observation

src/shared/src/types/account/conversion.rs
AntonioVentilii: My bad, I didn't see it above

PR #5916, opened 2025-04-17
refactor(frontend): Add token group types

daviddecentage:
# Motivation

For the upcoming fix for token groups in the token list (#5795) we decided to introduce groups for Tokens, to define how elements are grouped (in the future) an what informations to show in the token group card. 

# Changes

Added nessesary types for the new groupData prop on tokens and icCkInterface


src/frontend/src/lib/types/token.ts
```diff
@@ -34,7 +37,7 @@ export interface TokenLinkedData {
 
 export type TokenWithLinkedData = Token & TokenLinkedData;
 
-export type NonRequiredProps = TokenAppearance & TokenBuyable;
+export type NonRequiredProps = TokenAppearance & TokenBuyable & TokenGroup;
```
AntonioVentilii, line 40: I think you need to do the same for `RequiredErc20Token`

src/frontend/src/lib/types/token.ts
```diff
@@ -26,6 +27,8 @@ export type TokenAppearance = z.infer<typeof TokenAppearanceSchema>;
 
 export type TokenBuyable = z.infer<typeof TokenBuyableSchema>;
 
+export type TokenGroup = z.infer<typeof TokenGroupSchema>;
```
AntonioVentilii, line 30: there is a module for token-group types, shall it be moved there?

src/frontend/src/lib/schema/token.schema.ts
```diff
@@ -48,6 +48,16 @@ export const TokenBuyableSchema = z.object({
...
+export const TokenGroupSchema = z.object({
+	symbol: z.string(),
+	name: z.string(),
+	icon: z.string().optional()
+});
+
+export const TokenGroupPropSchema = z.object({
+	groupData: TokenGroupSchema.optional()
+});
+
```
AntonioVentilii, line 60: since there is a module for token-group types, it makes sense to add a module for token-group schema?

src/frontend/src/lib/types/token.ts
AntonioVentilii: as you said, there was no need to change it

AntonioVentilii (approved):
LGTM tks!

Maybe check my comments about modules, before merging

PR #5917, opened 2025-04-17
feat(frontend): add ETH address check in `LoaderEthBalances` component

AntonioVentilii:
# Motivation

To avoid risk of loading balances without the ETH address loaded (any possible race condition), we create a first check for nullish values.

It is ok, since the loader is re-triggered with changes in the ETH address store. So, as soon as it is not nullish, it will run.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5919, opened 2025-04-17
feat(backend): wire backend pow entrypoint

DecentAgeCoder:
## Motivation

This Pull Request ensures that the allow_signing function is invoked directly—without requiring the user to solve the challenge—when a new user profile is created. This guarantees that the user is granted allowed cycles during their first login.

## Changes

- Introduced an asynchronous task to handle the allow_signing operation after user profile creation in order to streamline the process. The function is invoked in a "fire-and-forget" manner. T
- Adjusted the conditional logic to maintain backward compatibility when the POW_ENABLED feature flag is disabled. When PoW is enabled, the user must always provide a valid nonce.

## Tests

No dedicated test changes have been introduced for the described modifications. The impact on existing functionality should be covered by the current test suite.


src/backend/src/lib.rs
```diff
@@ -648,6 +648,23 @@ pub fn create_user_profile() -> UserProfile {
...
+            // before any signer-related calls (e.g., get_eth_address). Spawn the async
+            // task separately after returning UserProfile synchronously
+            if let Err(e) = signer::allow_signing(None).await {
+                ic_cdk::println!(
+                    "Error enabling signing for user {}: {:?}",
+                    stored_principal.0,
+                    e
+                );
+            }
+        });
```
bitdivine, line 667: This should almost certainly be moved out of the `mutate_state()` call.

Locks should almost never he held for inter canister calls.

bitdivine (approved):
LGTM.  Thanks!

PR #5923, opened 2025-04-17
fix(frontend): use `expect().includes()` in tests

AntonioVentilii:
# Motivation

We are going to add the vitest linter (PR https://github.com/dfinity/oisy-wallet/pull/5797), but it raises quite a few issues. So, in preparation, for now we apply the rule [vitest/valid-expect](https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/valid-expect.md) that was raising issues with expect statements not having a corresponding matcher call.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5924, opened 2025-04-17
refactor(frontend): skip tests instead of commenting them

AntonioVentilii:
# Motivation

We are going to add the vitest linter (PR https://github.com/dfinity/oisy-wallet/pull/5797), but it raises quite a few issues. So, in preparation, for now we apply the rule [vitest/no-commented-out-tests](https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/no-commented-out-tests.md) that suggests to use the `.skip` clause instead of commenting tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5925, opened 2025-04-17
refactor(frontend): sort hooks in tests

AntonioVentilii:
# Motivation

We are going to add the vitest linter (PR https://github.com/dfinity/oisy-wallet/pull/5797), but it raises quite a few issues. So, in preparation, for now we apply the rule [vitest/prefer-hooks-in-order](https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/prefer-hooks-in-order.md) that enforces a specific order for hooks:

```ts
  // consistent order of hooks
  ['beforeAll', 'beforeEach', 'afterEach', 'afterAll']
```


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5926, opened 2025-04-17
feat(frontend): ConvertContexts component

DenysKarmazynDFINITY:
# Motivation

The following 2 convert-related contexts will be re-used multiple times, therefore it makes sense to create a separate component for that (similar as with the send context).


AntonioVentilii (approved):
LGTM tks

PR #5932, opened 2025-04-17
feat(devops): Add a build report to the frontend docker build

bitdivine:
# Motivation
Some parts of the frontend reproducible builds are not yet quite 100% reproducible.  An example is the SHA calculation of the .env file, which can change based on whether there is an end-of line when copying and pasting.

# Changes
- Add a `docker-build.frontend` script to mirror the backend.
- In the docker build, calculate the .env SHA hash, ignoring whitespace.
  - Note: This is done with perl.  Alternative considered: Run shfmt.  This is probably nicer but needs shfmt to be added to the docker build.  Easy to do but is opted for a more lightweight solution, with tools that are already installed.
- Print the build stats when the docker build is complete, as with backend builds.
  - Note: It would make sense to include commit, tags etc in the build report.  These can be added in future commits if they are valuable.
  - Note: The build report can also be run outside Docker, just as for backend builds.

# Tests
- I have run `./scripts/docker-build.frontend` locally.
- CI runs docker build.


AntonioVentilii (approved):
Thank you! Long needed!

PR #5933, opened 2025-04-17
refactor(frontend): reuse decodeErc20AbiDataValue for ckERC20 txs

DenysKarmazynDFINITY:
# Motivation

Instead of having 2 different functions, it makes sense to re-use a single implementation.


AntonioVentilii (approved):
LGTM tks!

since we are here, maybe some small tests?

DenysKarmazynDFINITY:
@AntonioVentilii yes, good idea - added 2 basic tests

PR #5918, opened 2025-04-17
feat(frontend): Add token groups

daviddecentage:
# Motivation
For the upcoming fix for token groups in the token list (https://github.com/dfinity/oisy-wallet/pull/5795) we decided to introduce groups for Tokens, to define how elements are grouped (in the future) and what informations to show in the token group card.

# Changes
Added the token groups and assigned them to the token definitions

PR #5939, opened 2025-04-17
feat(frontend): add EURC token group data

AntonioVentilii:
# Motivation

Splitting PR https://github.com/dfinity/oisy-wallet/pull/5918.

We are going to use `groupData` to group similar ("twin") tokens. Currently we are using different props depending on the type of token (for example `twinToken` for ICRC, or `twinTokenSymbol` for ERC20). With groupData, we aim to have a single center of aggregation for "twin" tokens.

In this PR we create the group for token EURC.


BonomoAlessandro (approved):
LGTM

PR #5943, opened 2025-04-17
feat(frontend): derived store for EVM enabled tokens

AntonioVentilii:
# Motivation

Creating a central list for all EVM enabled tokens using a derived store (similar to the other chains).


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5950, opened 2025-04-17
feat(frontend): util to define the supported tokens for a specific chain

AntonioVentilii:
# Motivation

In this case, we create an util that defines the supported tokens for a chain, and we use it in the env constants.

# Changes

- Create util that replicate existing code of `SUPPORTED_BASE_NETWORKS`, `SUPPORTED_BSC_NETWORKS`, `SUPPORTED_BITCOIN_NETWORKS`, `SUPPORTED_ETHEREUM_NETWORKS` and `SUPPORTED_SOLANA_NETWORKS`.

# Tests

Added new tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5951, opened 2025-04-17
chore(frontend): remove duplicated tests

AntonioVentilii:
# Motivation

Some tests were duplicated in `token-group.utils`, `btc-pending-sent-transactions-status.derived` and `reward.services` test suite.


src/frontend/src/tests/lib/utils/token-group.utils.spec.ts
```diff
@@ -557,238 +557,6 @@ describe('token-group.utils', () => {
...
-		it('should add token to existing group with more than one token already', () => {
-			const tokenGroup: TokenUiGroup = {
-				id: token.id,
-				nativeToken: token,
-				tokens: [twinToken, anotherToken],
-				balance: bn3Bi,
-				usdBalance: 300
-			};
-
-			expect(groupMainToken({ token, tokenGroup })).toEqual({
```
DenysKarmazynDFINITY, line 706: I am not sure I fully understand what was exactly duplicated here - if I check e.g. this test case, I can only find a single appearance, there are no other places where `groupMainToken` is tested. Am I misunderstanding/missing something?

DenysKarmazynDFINITY: Ah, see it now, WebStorm duplicate detection helped :) Thanks 👍

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5952, opened 2025-04-17
chore(frontend): de-duplicate test string titles

AntonioVentilii:
# Motivation

We are going to add the vitest linter (PR https://github.com/dfinity/oisy-wallet/pull/5797), but it raises quite a few issues. So, in preparation, for now we apply the rule [vitest/no-identical-title](https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/no-identical-title.md) that warns about similar titles in tests strings.


DenysKarmazynDFINITY (approved):
👍

PR #5953, opened 2025-04-17
feat(frontend): add ERC20 transactions methods to `EtherscanProvider`

AntonioVentilii:
# Motivation

The classes `EtherscanProvider` and `EtherscanRest` can actually be merged, since the latter simply query the Etherscan API in the same way the first does, just without using `ethers` library.

So, we can replicate the `EtherscanRest` method to fetch ERC20 transactions into `EtherscanProvider`, adapting to the logic of it.

# Changes

- Rename response interface `EtherscanRestTransaction` to `EtherscanProviderTokenTransferTransaction` to be more coherent (and sort the props according to documentation).
- Add method `erc20Transactions` to class `EtherscanProvider` that is a copy (almost _verbatim_) of the method transactions of class `EtherscanRest`.

# Tests

The tests for `EtherscanProvider` were adapted and more were included.


src/frontend/src/eth/providers/etherscan.providers.ts
```diff
@@ -137,6 +139,58 @@ export class EtherscanProvider {
...
+	}: {
+		address: EthAddress;
+		contract: Erc20Token;
+	}): Promise<Transaction[]> => {
+		const params = {
+			action: 'tokentx',
+			contractAddress,
+			address,
+			startblock: 0,
+			endblock: 99999999,
```
BonomoAlessandro, line 156: I guess this is defined as `99999999` to load all transactions. If so, is it needed to define it or would it also work if we just let it away?

BonomoAlessandro (approved):
LGTM

PR #5954, opened 2025-04-17
feat(frontend): check for EVM enabled networks for the address loaders

AntonioVentilii:
# Motivation

We should include the EVM networks in the checks to load/verify the ETH address, since it is the same address for all of them. So, we create derived stores that are a mirror of the Ethereum ones and, apart form the ck-conversions, use them in the same exact code.

# Changes

- Create lists of EVM supported networks (mainnet and testnet).
- Derived store `networkEvmMainnetEnabled` that verifies if any EVM mainnet networks is enabled.
- Derived store `networkEvmTestnetEnabled` that verifies if any EVM testnet networks is enabled.
- Use the new derived stores in the same place as the respective Ethereum stores (apart form ck-conversions).

# Tests

Current tests are sufficient.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5962, opened 2025-04-18
fix(build): Fix code scanning alert no. 6: avoid possible race condition when creating ckERC20 tokens.

AntonioVentilii:
# Motivation

Fixes https://github.com/dfinity/oisy-wallet/security/code-scanning/6

In the script to build ckERC20 tokens, there may be a race condition (CodeQL [js/file-system-race](https://github.com/github/codeql/blob/626c752a0b8cfd98bcf1f4b622d73bbe6f078a4c/javascript/ql/src/Security/CWE-367/FileSystemRace.ql)) when creating the logo file:

```ts
...
if (existsSync(file)) {
  return;
}
...
writeFileSync(file, svgContent, 'utf-8');
// The file may have changed since it was checked
...
```

So, to avoid it, we do similar to other scripts and open the file before writing it.

### DIFF
--- a/scripts/build.tokens.ckerc20.ts
+++ b/scripts/build.tokens.ckerc20.ts
@@ -15,7 +15,7 @@ import {
 } from '@dfinity/cketh';
 import { Principal } from '@dfinity/principal';
 import { fromNullable, isNullish, jsonReplacer, nonNullish } from '@dfinity/utils';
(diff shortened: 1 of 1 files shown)


DenysKarmazynDFINITY (approved):
👍

PR #4165, opened 2025-01-10
fix(build): Fix code scanning alert no. 1: Bad HTML filtering regexp

AntonioVentilii:
Fixes [https://github.com/dfinity/oisy-wallet/security/code-scanning/1](https://github.com/dfinity/oisy-wallet/security/code-scanning/1)

To fix the problem, we need to modify the regular expression to be case-insensitive. This can be achieved by adding the `i` flag to the regular expression. This change ensures that the regular expression will match `<script>`, `<SCRIPT>`, `<Script>`, and other case variations.

The best way to fix the problem without changing existing functionality is to update the regular expression on line 97 to include the `i` flag. This change will ensure that the regular expression matches script tags regardless of their case.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

PR #5966, opened 2025-04-18
fix(build): Fix code scanning alert no. 2: Bad HTML filtering regexp

AntonioVentilii:
# Motivation

Fixes https://github.com/dfinity/oisy-wallet/security/code-scanning/2

To fix the problem, we need to modify the regular expression to be case-insensitive. This can be achieved by adding the `i` flag to the regular expression. This change ensures that the regular expression will match `<script>`, `<SCRIPT>`, `<Script>`, and other case variations.

The best way to fix the problem without changing existing functionality is to update the regular expression on line 97 to include the `i` flag. This change will ensure that the regular expression matches script tags regardless of their case.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

# NOTE

Similar to PR https://github.com/dfinity/oisy-wallet/pull/4165


DenysKarmazynDFINITY (approved):

AntonioVentilii:
Fixes [SECFIND-1823]

PR #5969, opened 2025-04-18
feat(frontend): util to check if a network ID is EVM network

AntonioVentilii:
# Motivation

Similar to the rest, we create a small util to check if a certain network ID pertains to a EVM networks.


BonomoAlessandro (approved):
LGTM

PR #5974, opened 2025-04-19
refactor(frontend): ZERO_BI --> ZERO

AntonioVentilii:
# Motivation

The constant `ZERO_BI` was called like that because we had `ZERO` before, that was the `BigNumber` equivalent. Now that we deprecated `BigNumber`, we can rename `ZERO_BI` to `ZERO`, to make it more understandable.


AntonioVentilii:
Superseded by:

- https://github.com/dfinity/oisy-wallet/pull/6211

PR #5993, opened 2025-04-22
feat(rewards): Add the GOLD rewards API

bitdivine:
# Motivation
We would like to have multiple QR code promotional campaigns.

The API for this has been discussed and agreed in principle.

# Changes
- Update the rewards canister to one that has the agreed API:
  - Defined but without implementation
  - Rearranged a little to make it backwards compatible

# Tests
This API is deployed on staging.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -448,8 +448,8 @@
 		},
 		"rewards": {
 			"type": "custom",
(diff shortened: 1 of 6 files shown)


bitdivine:
These changes have been taken over by @BonomoAlessandro .

PR #5995, opened 2025-04-22
feat(frontend): util to define the supported tokens for a specific chain

AntonioVentilii:
# Motivation

In this case, we create an util that defines the supported tokens for a chain, and we use it in the env constants.

# Changes

- Create util that replicate existing code of `SUPPORTED_ETHEREUM_TOKENS `, `SUPPORTED_BITCOIN_TOKENS `, and `SUPPORTED_SOLANA_TOKENS `.

# Tests

Added new tests.

### DIFF
--- a/src/frontend/src/icp/components/receive/IcReceiveCkEthereum.svelte
+++ b/src/frontend/src/icp/components/receive/IcReceiveCkEthereum.svelte
@@ -1,36 +1,32 @@
 <script lang="ts">
-	import { getContext, setContext } from 'svelte';
+	import { nonNullish } from '@dfinity/utils';
+	import { getContext } from 'svelte';
+	import FeeStoreContext from '$eth/components/fee/FeeStoreContext.svelte';
 	import { erc20UserTokens } from '$eth/derived/erc20.derived';
+	import { ethereumToken } from '$eth/derived/token.derived';
 	import IcReceiveCkEthereumModal from '$icp/components/receive/IcReceiveCkEthereumModal.svelte';
 	import {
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/constants/tokens.constants.ts
```diff
@@ -13,24 +13,26 @@ import {
...
 export const [DEFAULT_ETHEREUM_TOKEN] = SUPPORTED_ETHEREUM_TOKENS;
 
 // The following tokens are used as fallback for any Bitcoin token defined in the token store.
 // That means that the order of the tokens in the array is important, to have a correct fallback chain.
-export const SUPPORTED_BITCOIN_TOKENS: [...Token[], Token] = [
-	...(BTC_MAINNET_ENABLED ? [BTC_MAINNET_TOKEN] : []),
-	BTC_TESTNET_TOKEN,
-	BTC_REGTEST_TOKEN
-];
+const SUPPORTED_BITCOIN_TOKENS: Token[] = defineSupportedTokens({
```
DenysKarmazynDFINITY, line 22: Probably worth aligning between all networks the way we defined supported tokens. Though, not in this PR of course and not urgent.

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/lib/constants/tokens.constants.ts
AntonioVentilii: Right! Done in:

- https://github.com/dfinity/oisy-wallet/pull/5999

PR #5996, opened 2025-04-22
feat(frontend): supports type vip and gold on qr code modal

BonomoAlessandro:
# Motivation

The QR invitation code modal should support both VIP and Gold codes.

# Changes

- adds supported type

# Tests

**VIP code:**
image

**gold code:**
image


src/frontend/src/lib/components/icons/IconBinance.svelte
BonomoAlessandro, line 1: The IconBinance icon is not used in this PR but will be used in the next one. It is included here to show the need for two versions of the Binance Icon. One for the menu that is black and supports dark mode and one for the modal that remains yellow regardless of the theme.

daviddecentage (approved):
LGTM

PR #5997, opened 2025-04-22
feat(frontend): Add component to handle modal with multiple pages

FlavioInacta:
# Motivation

The address book uses several modals on top of each other. This component helps to open a stack of modals and to preserve the state of the previous modals.

See `TestModalComponent.svelte` to see an example of a modal with title and toolbar snippets.

Out of scope: Animations (will be added later)

Why not using `WizardModal`?
- The address book uses at least 6 different modals with independent title and button behaviors. 
- The address book is not a wizard and the modals can open in a variety of orders.
- The contact and address modals will be used independently from the address book and should be reusable.
- For user experience the state of the address book must be preserved when opening a contact. With the wizard this is not the case. (For example, search string, scroll position)

Advantages with this component:
- All related logic for a modal s in one component (buttons behavior, title)
- You can open any svelte component as a modal:  `open({ component: MyComponent})`or `open({component: MyComponent, props: {...}})`
- States of previously opened modals are preserved
- No need of adding components to a common modal component like in `Modals.svelte`
- No need to add methods for each available modal like in `modal.store.ts`
- Modals like "Add contact" can be opened inside an already opened modal or from the main app directly (some more code will be needed for that), and the success/error behavior can be configured independently.

# Changes

Adds a component to handle multiple modals.

# Tests

Added unit tests.

# Demo

[vokoscreenNG-2025-04-22_15-37-11.webm](https://github.com/user-attachments/assets/08094cde-ee90-4bf4-9cad-a7a1d27d915d)


AntonioVentilii (commented):
@FlavioInacta stumbling upon this draft:

It is a topic to discuss because the idea with OISY is to have (if possible) only one modal open at the time. That is why we use `WizardModal` extensively.

Sure, in some recent feature we accepted to have one more modal open (even if we should really avoid it). But in general, the entire OISY is built on the assumption that there will not be more than one modal open at the same time (see `modalStore` usages) to avoid issues and race conditions, especially in the send/convert flow.

So maybe it is better to revisit the solution to check why we need multiple modals open at the same time.

Addressing your points:

- Why do you need 6 different modals open at the same time? If they are independent from each other, it doesn't make sense that they are open at the same time. If they treat the same flow/usage, we have a few examples of `WizardModal` where the buttons do different things and they have different titles. Or even just `Modal`.
- We have a few examples of the `WizardModal` being opened at different steps.
- All components are reusable, including the modals created with `WizardModal`.
- We preserve the state of a few modals already. But in general that is an issue that is independent from `WizardModal`. You can always use store and context.

FlavioInacta:
@AntonioVentilii thank you for your feedback. I don't need to open 6 Modals at the same time, but I have to manage flows between them in a reliably and testable way.

One of the complex use case that I am facing is this:
- Open "Address Book" (first modal)
- Filter contacts with the search field
- Scroll to the contact you want to see
- Click on that contact -> Opens a new modal with contact information
- Close the modal (since it is not the desird contact)
- You should come back to the "Address Book" modal at the scrolled position with your search term
- Click on the next Contact -> Opens a new modal with contact information
- Click on pen to edit contact name -> Opens new modal where you can edit contact
(Click cancel you should come back to contact information)
- You change the contact name (The title of the current modal should change)
- Click save you come back to the "Address Book" at the position you were before

The same thing should work from anywhere in the application (if you see a contact you should be able to see contact detais). And there are more use cases to add delete addresses, jump directly to an address from within the app.

See figma: https://www.figma.com/design/duPCw1leqer7ES0sBb6Uua/7.-OISY-Redesign?node-id=10366-62640&m=dev

I really don't know how I should solve this with a "Wizard". My bet is that the component will get really messy and untestable. And also, I would have to find a solution for scroll positions and internal states of the components to rerender exactly what the user saw before without flickering.

AntonioVentilii:
@FlavioInacta the scenario you described is very similar to what we do with Manage tokens. I am pretty sure it can be all done in `WizardModal`. Have you tried using it, even as a Proof of Concept? They all seems steps in a modal to me, and you can persist the state of each step

PR #6002, opened 2025-04-22
refactor(frontend): use test util for common tests in ic-wallet worker

AntonioVentilii:
# Motivation

We are going to add the vitest linter (PR https://github.com/dfinity/oisy-wallet/pull/5797), but it raises quite a few warnings. So, in preparation, for now we apply the rule [vitest/no-duplicate-hooks](https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/no-duplicate-hooks.md) on the ic-wallet-balance-and-transactions worker tests.

We put the common tests into a single callable function.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6018, opened 2025-04-22
docs: add new setup documents for local development for windows.

SYInacta:
# Motivation

Add documents to help out local development to start fast on windows.

# Changes

Added Windows.md, Linux.md and MacOs.md with steps and rules to setup wallet locally.


DecentAgeCoder:
I was only able to review the linux part. All will have to be reviewed an approved individually. I suggest to split the PR into a linux, macos and windows specific PR.

AntonioVentilii (commented):
same comment as https://github.com/dfinity/oisy-wallet/pull/6049#pullrequestreview-2786399126

AntonioVentilii (commented):
To be honest, I still have to see a DFINITY project that helps devs to port linux-based instructions to windows. Usually, if there are specific packages or flows yes, but not basically translating the entire deployment and dev-ops to windows, that is not done usually


But even so, there are instructions that we do not need to include like the installation of basic tools like node

PR #6021, opened 2025-04-22
feat(frontend): use Etherscan API v2

AntonioVentilii:
# Motivation

We are adding new EVM networks (for example Base and BNB Smart Chain), so we can re-use some providers that works with the difference of passing different Chain IDs. For example Etherscan API provides a single entry point for all supported EVM networks with [the new API v2](https://docs.etherscan.io/etherscan-v2).

This allow us to have less API keys, and re-use part of the code.

# Changes

- Create new Etherscan Plugin with the v2 base URL (see [docs](https://docs.ethers.org/v6/api/providers/thirdparty/#EtherscanPlugin)).
- Change map of Ethereum providers to create instances of Network.
- Attach custom plugin to the network map.
- Add param chainId to each request of class `EtherscanProvider`.

# Tests

Added unittests, but tried in env container and there were no changes.


DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/eth/providers/etherscan.providers.ts
```diff
@@ -88,6 +89,7 @@ export class EtherscanProvider {
 		endBlock
 	}: TransactionsParams): Promise<Transaction[]> {
 		const params = {
+			chainId: this.chainId,
```
DenysKarmazynDFINITY, line 92: Just one question - can we now use this `chainId` when mapping transactions below in this function?

AntonioVentilii: good catch! yes, we can absolutely! done

PR #6022, opened 2025-04-22
fix(frontend): Sign adjust space

daviddecentage:
# Motivation

The addition of the help link at the end of the OISY sign page lead to some Apps that have integrated connect to OISY to be displayed in a Popup with scroll bar.

The page is designed to not have a scrollbar, and since we cannot ask all Apps and Pages to adjust, we remove some slight spacings to make it fit same as before the addition of the help text.

# Changes

Tweaked spacings

# Tests

Before (without new help text):
image

After:
image


AntonioVentilii (approved):
Everything on the left is still a bit strange, but for this PR the rest LGTM! Tks

BonomoAlessandro (approved):
LGTM

PR #6026, opened 2025-04-22
refactor(frontend): Group data in token cards

daviddecentage:
# Motivation

We want to display the token icon, name and symbol in the token group cards. If a CK token is first, currently it displayed the special CK icon for the whole group. We change this so that for example USDC icon/name/symbol shows as the group data, not the data from the CK token which is confusing.

# Changes

- Added `asGroup` prop for `TokenCard`
- Display group icon, name and symbol if `asGroup` and available
- Added token group of first token to `CardData`

# Tests

Before:
image

After:
image


src/frontend/src/lib/utils/token-card.utils.ts
```diff
@@ -21,5 +21,6 @@ export const mapHeaderData = ({
 	oisySymbol: { oisySymbol: name },
 	balance,
 	usdBalance,
-	tokenCount: tokens.length
+	tokenCount: tokens.length,
+	groupData: tokens[0]?.groupData
```
AntonioVentilii, line 25: BTW, maybe in another PR (add a TODO), this can be directly the prop of `TokenUiGroup` type. We have `nativeToken` prop in it, but it can be deprecated in favor of `groupData`, right? And we deconstruct `groupData` instead of `nativeToken`

src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -33,7 +35,9 @@
 	>
 		<span class="flex" slot="logo" class:mr-2={!asNetwork}>
 			<TokenLogo
-				{data}
+				data={asGroup && nonNullish(data?.groupData?.icon)
+					? { ...data, icon: data.groupData.icon }
+					: data}
```
AntonioVentilii, line 40: The `data` structure shall not change. Meaning, we should still use `data` here, but the information inside `data` are fetched based on `groupData`.
So that this component would not change, but only the parsing of the pseudo-tokens that we have in the list. It is done by function `groupTokensByTwin`

BonomoAlessandro (approved):
LGTM

src/frontend/src/lib/components/tokens/TokenCard.svelte
daviddecentage: Not sure there. Wouldnt `mapHeaderData` be a better place until we move it to `TokenUiGroup`? Check the changes I made.

AntonioVentilii (approved):
As discussed, we tackle the TODO in another PR (switch from `nativeToken` to `groupData`)

So, approved

PR #6027, opened 2025-04-22
feat(frontend): add EVM networks in Infura provider

AntonioVentilii:
# Motivation

We add Base and BNB Smart Chain (BSC) among the Infura networks.


src/frontend/src/tests/eth/providers/infura.providers.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+	const networks: EthereumNetwork[] = [
+		ETHEREUM_NETWORK,
+		SEPOLIA_NETWORK,
+		BASE_NETWORK,
+		BASE_SEPOLIA_NETWORK,
+		BSC_NETWORK,
+		BSC_TESTNET_NETWORK
+	];
+
+	const infuraNames: Networkish[] = [
```
DenysKarmazynDFINITY, line 39: how about reusing `providersMap` from `infura.providers.ts` here in tests, instead of creating `networks` and `infuraNames`?

AntonioVentilii: i didn't want to export that to be honest, it feels "forced",  if it makes sense. Meaning, that I am using an object to verify that a function that uses the same object works...

DenysKarmazynDFINITY: Ok, got it, makes sense!

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6031, opened 2025-04-22
refactor(build): extract common function to save token logo

AntonioVentilii:
# Motivation

To avoid repetition of code, we extract a common function to save the logo of ICRC and ckERC20 tokens in the build scripts.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6034, opened 2025-04-22
feat(frontend): enable actions and modals for EVM networks

AntonioVentilii:
# Motivation

Where needed we enable for the EVM networks the same modals and the same actions that Ethereum networks use.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6035, opened 2025-04-22
feat(frontend): add EVM networks to the `explorerUrl` store

AntonioVentilii:
# Motivation

We are using the same components for Ethereum networks and EVM networks. So, we adapt store `explorerUrl` to consider the EVM networks too.


src/frontend/src/evm/derived/network.derived.ts
```diff
@@ -0,0 +1,9 @@
+import type { EthereumNetwork } from '$eth/types/network';
+import { enabledEvmNetworks } from '$evm/derived/networks.derived';
+import { networkId } from '$lib/derived/network.derived';
+import { derived, type Readable } from 'svelte/store';
+
+export const selectedEvmNetwork: Readable<EthereumNetwork | undefined> = derived(
```
DenysKarmazynDFINITY, line 6: I think this one should already be in main, right? In any case, rebasing will do the job

PR #6041, opened 2025-04-22
feat(frontend): map EVM tokens to the same listener of Ethereum

AntonioVentilii:
# Motivation

We use the same Ethereum listener component `EthListener` for all the EVM tokens.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6047, opened 2025-04-23
refactor(frontend): uses enum instead of plain string

BonomoAlessandro:
# Motivation

Instead of using plain strings we want to use an enum to make sure, that only supported types are used.

# Changes

- uses enum instead of plain string


src/frontend/src/lib/enums/qr-code-types.ts
```diff
@@ -2,3 +2,6 @@ export enum QrCodeType {
 	VIP = 'vip',
 	GOLD = 'gold'
 }
+
+export const asQrCodeType = (key: string) =>
```
AntonioVentilii, line 6: maybe this can be moved to utils module? i have no strong opinion, since I am not sure where it should go eheh! so non-blocker

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/enums/qr-code-types.ts
BonomoAlessandro: I think it is better to leave it there, as it is closely related to the enum itself.

PR #6050, opened 2025-04-23
feat(frontend): supports new type while using vip reward state modal

BonomoAlessandro:
# Motivation

When opening the vip reward state modal the campaign ID of the claimed code should be passed to the modal. This allows it to display more accurate information to the user.

# Changes

- supports `Gold` type for vip reward state modal


daviddecentage (approved):
LGTM

PR #6052, opened 2025-04-23
feat(frontend): ignore failed metadata fetched during loading of ICRC custom tokens

AntonioVentilii:
# Motivation

There are some cases where the ledger canister fails completely (most recently we had the episode with [vUSD](https://github.com/dfinity/oisy-wallet/pull/6043)). In those cases, we prefer to ignore the failing ledger canisters and proceed with the working ones during the loading of metadata of ICRC custom tokens.

Otherwise the current implementation resets them all, and does not show anything.

# Changes

- Await all the promises to load metadata of ICRC custom tokens, successful and failing.
- Filter out the failing ones.
- Reset the ICRC custom token store for the failed.

# Tests

Adapted and added tests.

### DIFF
--- a/src/frontend/src/icp/services/icrc.services.ts
+++ b/src/frontend/src/icp/services/icrc.services.ts
@@ -205,11 +205,13 @@ const loadCustomIcrcTokensData = async ({
 			return acc;
 		}
 
-		if (isNullish(result.value)) {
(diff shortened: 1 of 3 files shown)


DenysKarmazynDFINITY:
The changes LGTM, thanks!

PR #6054, opened 2025-04-23
feat(frontend): add specific ID to `TokenGroup` type

AntonioVentilii:
# Motivation

It is useful to have a specific `symbol` ID for each `TokenGroup`, since we will merge them in a list together with the other tokens.


src/frontend/src/lib/types/token-group.ts
```diff
@@ -1,7 +1,15 @@
...
+	type TokenGroupIdSchema,
+	type TokenGroupPropSchema,
+	type TokenGroupSchema
+} from '$lib/schema/token-group.schema';
 import type { TokenFinancialData, TokenId, TokenUi } from '$lib/types/token';
 import type * as z from 'zod';
 
+export type TokenGroupId = z.infer<typeof TokenGroupIdSchema>;
+
+export type TokenGroupData = z.infer<typeof TokenGroupSchema>;
```
DenysKarmazynDFINITY, line 11: Is this gonna be used in the next PRs?

src/frontend/src/lib/types/token-group.ts
```diff
@@ -1,7 +1,15 @@
...
+
 export type TokenGroup = z.infer<typeof TokenGroupPropSchema>;
 
 //todo: separate typing from token id
```
DenysKarmazynDFINITY, line 15: This TODO is a low-hanging fruit now :)

AntonioVentilii: yes, correct, but I will use `TokenGroupId` instead of `GroupId` in another PR for interface `TokenUiGroup`, and remove that TODO

src/frontend/src/lib/types/token-group.ts
AntonioVentilii: Good catch! For next one, removed for now

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6055, opened 2025-04-23
refactor(frontend): move QrCode step to SendModal

DenysKarmazynDFINITY:
# Motivation

Before extracting destination input into a separate wizard step, we need to refactor the Send flow by moving QrCode step up to the SendModal component. It also allows us to clean up the components and to avoid some code repetitions.


AntonioVentilii (approved):
LGTM tks

PR #6060, opened 2025-04-23
refactor(frontend): move SendTokenContext to SendModal

DenysKarmazynDFINITY:
# Motivation

We need to move SendTokenContext to SendModal in order to have access to the context in the future Destination wizard step (will be added to SendModal).


AntonioVentilii (approved):
LGTM tks!

PR #6063, opened 2025-04-23
feat(frontend): use group data for utils to group tokens

AntonioVentilii:
# Motivation

We want to use `groupData` field to group tokens in the list, instead of custom props for each type of token (for example `twinToken` for ICRC).

Unfortunately there is no easy way of doing that unless we completely refactor util `groupTokens`.

The idea is to change its logic to return a group if `groupData` are present, otherwise return the token alone.

# Changes

- In the loop of util `groupTokens`, check if there are `groupData`:
  - if there are, proceed as before, mapping a group variant `{ group: TokenUiGroup }`
  - it there aren't, proceed returning the variant `{ token: TokenUi }`
- Use `groupData` ID when mapping a new `TokenUiGroup`.
- Include `groupData` when mapping a new `TokenUiGroup`.
- Adapt docstrings.
- Remove deprecated util `groupMainToken` (doing it in this PR is simpler, since the types do not match anymore).

# Next Steps

Util `groupSecondaryToken` shall be renamed since it now simply creates a new group or update an existing group.

# Tests

Test adapted. Furthermore, test canister replica works as before:

Screenshot 2025-04-24 at 17 50 23


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6066, opened 2025-04-23
refactor(frontend): `OnramperWidget` to Svelte 5

AntonioVentilii:
# Motivation

Migrating `OnramperWidget` to Svelte v5.


src/frontend/src/lib/components/onramper/OnramperWidget.svelte
```diff
@@ -19,44 +19,43 @@
 	import type { OnramperId, OnramperNetworkId, OnramperNetworkWallet } from '$lib/types/onramper';
 	import { buildOnramperLink, mapOnramperNetworkWallets } from '$lib/utils/onramper.utils';
 
-	let defaultCrypto: OnramperId | undefined;
-	$: defaultCrypto =
+	let defaultCrypto: OnramperId | undefined = $derived(
```
DenysKarmazynDFINITY, line 22: $derived can actually be defined as `const`.

peterpeterparker: They use const in NNS dapp but, I actually suggested Maksims this weekend to use `let` or at least to double check why Svelte uses let in their documentation. Just to be sure that we do not have some day to migrate back const to let.

Aside of this, the type is inteferred, you can avoid it @AntonioVentilii :

```
let defaultCrypto = $derived($token?.buy?.onramperId ??...);
```

If it needs to be specified I think you can use the generic

```
declare function $derived<T>(expression: T): T;
```

They changed it along the way, I've got many places to update in my code base.

DenysKarmazynDFINITY: Good point with the documentation!

AntonioVentilii: Removed implied types!

DenysKarmazynDFINITY (approved):
Oops, sorry, I thought I already approved this PR

PR #6067, opened 2025-04-23
feat(frontend): remove flag `USER_NETWORKS_FEATURE_ENABLED`

AntonioVentilii:
# Motivation

We don't need the flag `USER_NETWORKS_FEATURE_ENABLED` since the feature is already working in the last releases.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6068, opened 2025-04-23
feat(frontend): opens manage tokens modal for gold token if not yet enabled

BonomoAlessandro:
# Motivation

When a user receives a Gold token via an invitation code, they get a notification. As soon as they click the `Take me to the wallet` button, the `GLDT` token should get enabled if it is not already.
Since we cannot guarantee that all data has fully loaded at this point, we are unable to enable the token automatically. For this reason we open the `Manage Tokens` modal with the `GLDT` token preselected. This allows the user to easily enable the token manually.
In the worst case the modal might be open even if the token is already enabled, simply because the data was not fully loaded yet, but this behavior is acceptable.

# Changes

- opens `Manage Tokens` modal if `GLDT` token is not yet enabled

# Tests

Claim reward while `GLDT` token is disabled and wait until all tokens are loaded

https://github.com/user-attachments/assets/b9c25468-0414-4f7b-956e-34117b034b86

Claim reward while `GLDT` token is disabled and do not wait until all tokens are loaded

https://github.com/user-attachments/assets/07d515f1-f496-4d6d-9864-9dd597e4f927

Claim reward while `GLDT` token is already enabled

https://github.com/user-attachments/assets/02cf1a17-e5d1-4059-8d9a-56a7a042418e


src/frontend/src/env/tokens/tokens.icrc.env.ts
```diff
@@ -6,3 +6,5 @@ const additionalIcrcTokensParsed = EnvAdditionalIcrcTokensSchema.safeParse(icrcT
 export const additionalIcrcTokens = additionalIcrcTokensParsed.success
 	? additionalIcrcTokensParsed.data
 	: {};
+
+export const GLDT_SYMBOL = 'GLDT';
```
AntonioVentilii, line 10: let's first add the token, no? even because it is better to use the token ID instead of the symbol

AntonioVentilii (changes requested):

src/frontend/src/lib/components/tokens/ManageTokensButton.svelte
```diff
@@ -13,7 +13,7 @@
 <button
 	class="tertiary"
 	data-tid={MANAGE_TOKENS_MODAL_BUTTON}
-	on:click={modalStore.openManageTokens}
+	on:click={() => modalStore.openManageTokens()}
```
AntonioVentilii, line 16: doubt: why do you need this change?

src/frontend/src/lib/stores/modal.store.ts
```diff
@@ -137,7 +138,7 @@ const initModalStore = <T>(): ModalStore<T> => {
 		openIcTransaction: setTypeWithData('ic-transaction'),
 		openBtcTransaction: setTypeWithData('btc-transaction'),
 		openSolTransaction: setTypeWithData('sol-transaction'),
-		openManageTokens: setType('manage-tokens'),
+		openManageTokens: <(data?: ManageTokensData) => void>setTypeWithData('manage-tokens'),
```
AntonioVentilii, line 141: as info: I am merging PR https://github.com/dfinity/oisy-wallet/pull/6088. Depending on which one merges first, you maybe need to change this part

src/frontend/src/tests/utils/map-icrc-data.ts
```diff
@@ -0,0 +1,22 @@
...
+	Object.entries(icrcTokens).reduce(
+		(acc, [key, value]) => ({
+			...acc,
+			...(LOCAL &&
+				nonNullish(value) && {
+					[key]: value
+				})
+		}),
+		{}
+	);
```
AntonioVentilii, line 22: for this module rename it with `.test-utils.ts` at the end?

src/frontend/src/tests/lib/components/qr/VipRewardStateModal.spec.ts
```diff
@@ -81,4 +90,156 @@ describe('VipRewardStateModal', () => {
 			expect(getByText(get(i18n).vip.reward.text.open_wallet)).toBeInTheDocument();
 		});
 	});
+
+	describe('Handle token state', () => {
+		const additionalIcrcData = mapLocalIcrcData(additionalIcrcTokens);
+		const goldToken = nonNullish(additionalIcrcData?.GLDT)
```
AntonioVentilii, line 96: doubt: have you tried to mock `LOCAL` here? so that it could return the original `additionalIcrcData` and you don't need to calculate them?

src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -84,7 +85,15 @@
 		{/if}
 
 		{#if $modalManageTokens}
-			<ManageTokensModal />
+			<ManageTokensModal initialSearch={$modalManageTokensData?.initialSearch}>
+				{#snippet infoElement()}
+					{#if nonNullish($modalManageTokensData) && nonNullish($modalManageTokensData.message)}
+						<MessageBox level="info">
+							{$modalManageTokensData.message}
```
AntonioVentilii, line 92: shall it maybe be deconstructed in `<script>` block? so here you can just provide `initialSearch` and `message`

src/frontend/src/lib/components/qr/VipRewardStateModal.svelte
```diff
@@ -57,7 +74,15 @@
...
 			type="button"
 			fullWidth
-			on:click={modalStore.close}
+			on:click={() => {
+				codeType === QrCodeType.GOLD && isNullish(goldToken) && nonNullish(goldTokenSymbol)
+					? modalStore.openManageTokens({
+							initialSearch: goldTokenSymbol,
+							message: replaceOisyPlaceholders($i18n.tokens.manage.text.default_message)
+						})
+					: modalStore.close();
```
AntonioVentilii, line 83: just for readability, may you please extract this function to the `<script>` block?

AntonioVentilii (approved):
LGTM tks! A few observations

src/frontend/src/lib/components/tokens/ManageTokensButton.svelte
BonomoAlessandro: Lint tells me that it is needed since we can call `openManageTokens` with an optional parameter.

PR #6073, opened 2025-04-23
feat(frontend): create prop `deprecated` for deprecated SNS tokens

AntonioVentilii:
# Motivation

Instead of relying on a single list of deprecated tokens on both build-side and runtime, it is useful to build the SNS tokens with a `deprecated` flag.

Suggestion by @peterpeterparker 

# Changes

- Add `deprecated` field in type `EnvSnsToken`.
- Set the flag in building script.
- Use the flag instead of the utils to check if a SNS token is deprecated.

# Tests

Adapted tests and ran the script on this same PR successfully (`deprecated` flag appeared for deprecated token CTS.


src/frontend/src/icp/services/icrc-custom-tokens.services.ts
```diff
@@ -78,5 +78,6 @@ const mapIcrcCustomToken = ({
 	fee,
 	alternativeName,
 	explorerUrl: `${SNS_EXPLORER_URL}/${rootCanisterId}`,
-	...(isNotDeprecatedSns({ ledgerCanisterId }) && { icon: `/icons/sns/${ledgerCanisterId}.png` })
+	deprecated,
+	...(!deprecated && { icon: `/icons/sns/${ledgerCanisterId}.png` })
```
peterpeterparker, line 82: `deprecated === false` maybe?

src/frontend/src/icp/schema/ic-token.schema.ts
```diff
@@ -1,3 +1,4 @@
+import { EnvSnsTokenSchema } from '$env/schema/env-sns-token.schema';
```
peterpeterparker, line 1: Does it maybe introduce a recursive imports?

AntonioVentilii: checking all the imports, it shouldn't, but just to be sure, i moved it in a separate module

peterpeterparker (approved):
LGTM, thx

PR #6076, opened 2025-04-23
refactor(frontend): rename images using kebab-case

AntonioVentilii:
# Motivation

For consistency we rename the images using the kebab-case, like the rest.

Furthermore, moved ICP image back in the `icp` folder.


src/frontend/src/env/tokens/tokens.btc.env.ts
```diff
@@ -5,8 +5,8 @@ import {
 	BTC_TESTNET_NETWORK
 } from '$env/networks/networks.btc.env';
 import { BTC_TOKEN_GROUP } from '$env/tokens/groups/groups.btc.env';
+import bitcoinTestnet from '$icp/assets/bitcoin-testnet.svg';
```
DenysKarmazynDFINITY, line 8: Since you're already on it, shall we move BTC-related icons to `$btc/assets`?

AntonioVentilii: makes sense! done

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6086, opened 2025-04-24
chore(frontend): remove deprecated utils for SNS tokens

AntonioVentilii:
# Motivation

Removing deprecated utils to check if a SNS token is deprecated.


src/frontend/src/icp/utils/icrc.utils.ts
```diff
@@ -1,5 +1,4 @@
 import { ICP_NETWORK } from '$env/networks/networks.icp.env';
-import { DEPRECATED_SNES } from '$env/tokens/tokens.sns.deprecated.env';
```
DenysKarmazynDFINITY, line 2: Can this object also be deleted? It seems it was the only usage

DenysKarmazynDFINITY (approved):
Small comment, approving already

src/frontend/src/icp/utils/icrc.utils.ts
AntonioVentilii: it is still used in the script that builds the SNS tokens, to create the `deprecated` flag

DenysKarmazynDFINITY: got it, thanks - I guess I only checked the src folder

PR #6087, opened 2025-04-24
fix(devops): Run bash in strict mode for breaking change check

bitdivine:
# Motivation
PRs with breaking changes should fail if not marked as breaking.

But this is not working.

Presumably because the shell is running in a permissive mode that ignores errors.

# Changes
- Force bash into pedantic mode for that step.
- Improve visibility of the variables.

# Tests
- If I force the compatibility check to fail, the step also fails.


AntonioVentilii (approved):
Thank you! LGTM

PR #6088, opened 2025-04-24
feat(frontend): add ID to modal store that have data input

AntonioVentilii:
# Motivation

It is useful to have a defined ID for each modal. For now we start with the modals that were requiring data as input too.

# Changes

- Make method `setTypeWithData` of `modalStore` accept an ID too.
- Adapt usage of `modalStore` throughout the code for the modals that were changed: create modal ID and pass it.

# Tests

Adapted tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6090, opened 2025-04-24
refactor(frontend): IC transactions utils

daviddecentage:
# Motivation

We need to display all pending transactions on the activity page in an upcoming PR. In preparation for it, we adjust the responsible derived stores by using utils function where we can pass the token additionally, instead of going for the token in store as in the derived stores (as we dont have a token in store on the activity page).

# Changes

- Moved all derived store logic to corresponding utils functions
- Added tests for the new utils functions

# Tests

Theres unit tests for the newly added functions, in turn also covering the derived stores since the whole logic is now in the utils.


AntonioVentilii (approved):
LGTM tks!

PR #6095, opened 2025-04-24
fix(frontend): Display pending CK transactions on activity page

daviddecentage:
# Motivation

As a continuation of #6090, we now make use of the new utils function to correctly display all pending CK transactions on the activity page as well.

# Changes

- Pass all required stores to the `mapAllTransactionsUi` function that gathers the required `TransactionsUI` for rendering
- Replaced todo with implementation by using the new utils functions
- Updated tests for adjusted `mapAllTransactionsUi` function

# Tests

All newly used functions should be covered by tests.


AntonioVentilii (approved):
LGTM tks!

PR #6099, opened 2025-04-24
feat(frontend): make TokenInput selectable in Send flows

DenysKarmazynDFINITY:
# Motivation

The next step of updating the Send flow is to make TokenInput selectable - it should lead users to Tokens list wizard step.

Basically, the idea of this PR is to create a new event in `SendModal` and pass it down to all wizards (BTC, SOL, ETH and IC). We also need to remove the `isSelectable={false}` prop to make TokenInput clickable.

Screenshot 2025-04-24 at 18 15 05


AntonioVentilii (approved):
LGTM tks!

PR #6101, opened 2025-04-25
chore(frontend): add type to token groups

AntonioVentilii:
# Motivation

Better to explicitly define the type of the token groups using interface `TokenGroup`.


BonomoAlessandro (approved):
LGTM

PR #6102, opened 2025-04-25
fix(frontend): Disable send flow only for the correct address not loaded

AntonioVentilii:
# Motivation

Currently the Send flow checks if the ETH address is loaded before proceeding to the next step.

This was working before because we would load all addresses at startup, so it was a good indication of the fact that OISY was still loading.

However, since the users can enable and disable networks now, we don't load all network address, only the enabled ones.

So, we need to change the check in the Send flow to consider the specific address for the selected token.

# Changes

- Create derived stores to check if the network address are loaded (mirroring derived store `ethAddressNotLoaded`).
- Change function isDisabled in component SendModal to check the specific address for the selected token. 

# Tests

Tested in local replica:

### Before

https://github.com/user-attachments/assets/9d8b72d4-e243-4d47-88b4-7bdd3ab0b2d3

### After

https://github.com/user-attachments/assets/0b24264e-24bf-4faf-a1e3-babdecfc3163


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6104, opened 2025-04-25
feat(frontend): create util to check supported native EVM tokens

AntonioVentilii:
# Motivation

We will need to use the EVM native tokens for all the same flows of the base Ethereum network. So, for now, we first create an util to check if a token is a native EVM token (similar to util `isSupportedEthTokenId`).

### DIFF
--- a/src/frontend/src/env/tokens/tokens-evm/tokens-base/tokens.eth.env.ts
+++ b/src/frontend/src/env/tokens/tokens-evm/tokens-base/tokens.eth.env.ts
@@ -1,9 +1,15 @@
 import {
+	BASE_MAINNET_ENABLED,
 	BASE_NETWORK,
 	BASE_SEPOLIA_NETWORK
 } from '$env/networks/networks-evm/networks.evm.base.env';
+import {
+	BNB_MAINNET_TOKEN,
+	BNB_TESTNET_TOKEN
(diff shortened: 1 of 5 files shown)


src/frontend/src/env/tokens/tokens-evm/tokens-base/tokens.eth.env.ts
```diff
@@ -40,3 +46,9 @@ export const BASE_SEPOLIA_ETH_TOKEN: RequiredToken = {
 	decimals: BASE_ETH_DECIMALS,
 	icon: eth
 };
+
+export const SUPPORTED_BASE_TOKENS: RequiredToken[] = defineSupportedTokens({
+	mainnetFlag: BASE_MAINNET_ENABLED,
+	mainnetTokens: [BNB_MAINNET_TOKEN],
```
DenysKarmazynDFINITY, line 52: Shouldn't it be BASE_ETH_TOKEN and BASE_SEPOLIA_ETH_TOKEN?

AntonioVentilii: copy-paste... good catch! you are right!

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6105, opened 2025-04-25
feat(frontend): check if a token is an EVM native token

AntonioVentilii:
# Motivation

In all the same situations where the code checks if a token is an Ethereum native token, we need to add the check for EVM native tokens.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6106, opened 2025-04-25
feat(frontend): deprecated derived store `selectedEthereumNetworkWithFallback`

AntonioVentilii:
# Motivation

We are now enabling and disabling networks, plus we are adding more EVM networks. That means that in some situations we should be aware if the correct network is being used.

To this purpose, we shall start removing the fallbacks from a few Ethereum derived stores.

We start from `selectedEthereumNetworkWithFallback`.

# Changes

- Substitute the usage of `selectedEthereumNetworkWithFallback` with `selectedEthereumNetwork`.
- Where necessary create non-nullish assertions, since it is expected that the derived store has a value when it is called in the correct context.
- Remove deprecated store `selectedEthereumNetworkWithFallback` and its tests.

# Tests

Local replica presented no changes when sending Sepolia tokens.

### DIFF
--- a/src/frontend/src/eth/components/convert/EthConvertTokenWizard.svelte
+++ b/src/frontend/src/eth/components/convert/EthConvertTokenWizard.svelte
@@ -1,13 +1,13 @@
 <script lang="ts">
 	import type { WizardStep } from '@dfinity/gix-components';
-	import { isNullish } from '@dfinity/utils';
+	import { assertNonNullish, isNullish, nonNullish } from '@dfinity/utils';
 	import { createEventDispatcher, getContext } from 'svelte';
 	import { ICP_NETWORK } from '$env/networks/networks.icp.env';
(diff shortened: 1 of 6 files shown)


src/frontend/src/eth/components/convert/EthConvertTokenWizard.svelte
```diff
@@ -114,6 +114,9 @@
 
 		dispatch('icNext');
 
+		// This is a simple type check, since it should not happen since the user arrived here from a selected Ethereum network
+		assertNonNullish($selectedEthereumNetwork);
```
DenysKarmazynDFINITY, line 118: Even though this check is mostly for TS, shall we move it before `dispatch('icNext');`? Just in case, so we stay on the same (Review) step if it fails for some reason

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6107, opened 2025-04-25
feat(frontend): default mainnet user networks to enabled when there are no settings

AntonioVentilii:
# Motivation

If users have already some networks enabled/disabled, and we decide to add new networks, by default they will not see them.

So, we want to change this behavior, showing all the networks by default that have no settings saved.

# Changes

- Use the list of default networks (mainnet enabled and testnets disabled) as base for the map of user networks.

# Tests

For new users, nothing has changed. Existing users that already have some enabled/disabled network have the correct change. For example, for user below:

Screenshot 2025-04-25 at 11 31 10

Once we add new networks:

### Before

Screenshot 2025-04-25 at 11 33 16

### After

Screenshot 2025-04-25 at 11 31 48


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6108, opened 2025-04-25
refactor(frontend): extract sub-function that auto load token

AntonioVentilii:
# Motivation

From service `autoLoadToken` we extract a sub-function called `autoLoadToken` that will take as input the token directly, instead of searching it based on the input parameters.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6109, opened 2025-04-25
feat(frontend): auto-enable token GLDT when receiving a reward

AntonioVentilii:
# Motivation

When an user receives the Gold rewards, it will receive GLDT tokens. However, we would like to auto-enable it, in case it is disabled in the token list. We were showing the Manage tokens modal, but we can actually automatically enable it on closing the Rewards modal.

# Changes

- Create sub-function in component `VipRewardStateModal` that enables the GLDT token using service `autoLoadSingleToken`.
- Trigger that function on closing (either button or clicking outside), but only for Gold rewards and if the token is disabled.

# Tests

Deployed to test canister:

https://github.com/user-attachments/assets/44d14bf2-d559-423a-a28a-28d5e9e5a4da


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6114, opened 2025-04-25
feat(build): temporary replace correct BSC API url for package `ethers`

AntonioVentilii:
# Motivation

There is a known issue with `ethers` package: the BNB Smart Chain (BSC) API URLs are not updated (see issue https://github.com/ethers-io/ethers.js/issues/4951). However we have the BSC network among our list of networks, so we need to temporarily patch it to the correct URL.

To do that we create a custom plugin and apply it (credits to @peterpeterparker ).

# Changes

- Create custom plugin to replace the URLs.
- Adapt `vite` configs to replace the wrong BSC API URLs.

# Tests

Tested in test canister and it works as expected:

Screenshot 2025-04-25 at 15 50 47

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,7 +1,7 @@
 import inject from '@rollup/plugin-inject';
 import { sveltekit } from '@sveltejs/kit/vite';
 import { basename, dirname, resolve } from 'node:path';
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
LGTM, thx

Note: We discussed the context and options offline.

PR #6117, opened 2025-04-25
feat(frontend): include EVM tokens among the checks for initialized token

AntonioVentilii:
# Motivation

Continuing to integrate EVM networks in the same flows as Ethereum networks, we include the EVM networks in the derived store that checks if an Ethereum token is initialized.


DenysKarmazynDFINITY (approved):
Like how the store looks now 🙂  thanks

PR #6121, opened 2025-04-25
refactor(frontend): generalize type of ID for certified stores

AntonioVentilii:
# Motivation

We would like to re-use the certified stores (`initCertifiedStore` and `initCertifiedSetterStore`) for a more generic `record`. So, instead of having only token IDs as keys, we can have any type of `symbols` we specify.

# Changes

- Change `tokenId` input to generic `id` in both store.
- Define generic type for `id` extending `symbol`, defaulting to `TokenId`.
- Adapt the rest of the code.

# Tests

Current tests are sufficients.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6126, opened 2025-04-25
refactor(frontend): include chain ID check in derived store `erc20UserTokensToggleable`

AntonioVentilii:
# Motivation

Since we have now more ERC20 over different networks, like EVM networks (different chain IDs), it makes sense to include the chain ID in the checks.

# Changes

- Change derived store `erc20DefaultTokensAddresses` to check the chain IDs between user tokens and default tokens.
- Remove deprecated store `erc20DefaultTokensAddresses`.

# Tests

Current tests are sufficient.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6127, opened 2025-04-25
feat(frontend): add EVM networks to derived stores for ERC20 tokens

AntonioVentilii:
# Motivation

The ERC20 derived store (both default and user tokens) now includes ERC20 from other networks. So, we must include the EVM networks among the checks when filtering.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6130, opened 2025-04-25
fix(frontend): wrong generic type for PowProtectionScheduler

AntonioVentilii:
# Motivation

Scheduler `PowProtectionScheduler` should not have a generic type.


DenysKarmazynDFINITY (approved):
thanks!

PR #6144, opened 2025-04-28
feature(frontend): make parseBoolEnvVar fail safe

DecentAgeCoder:
# Motivation

The parseBoolEnvVar musn't fail if a environment variable is empty or not defined in the github secret store.

# Changes

- Do not parse anything if the variable is not defined or empty. Return false insteaad.


src/frontend/src/lib/utils/env.utils.ts
```diff
@@ -5,7 +5,9 @@ import type { OptionString } from '$lib/types/string';
 // To avoid possible issues, we decided to give the parameters as array of objects and not as single object for this specific util.
 // eslint-disable-next-line local-rules/prefer-object-params
 export const parseBoolEnvVar = (value: OptionString, check = true): boolean =>
-	JSON.parse(value?.toLowerCase() ?? 'false') === check;
+	value === undefined || value === null
+		? false
```
AntonioVentilii, line 9: isn't it the same as `value?.toLowerCase() ?? 'false'` ? i think the tests covers that case, and in fact they are failing with this change...

AntonioVentilii (changes requested):

src/frontend/src/lib/utils/env.utils.ts
DecentAgeCoder: "Sorry for my delayed response, I completely missed this conversation.

I suggest enhancing the function by adding the ability to pass a default value. The case of an empty string should then be cleanly handled: if the value equals an empty string and no default value is specified, we should immediately throw an error. `parseBoolEnvVar`
This approach would give the caller of the function responsibility for determining whether the environment variable is required or not.
Furthermore, I was wondering why we support the inversion of the parsed string through the parameter `check`?
This parameter is currently only passed by the tests itself.

AntonioVentilii: I would prefer to leave it as it is, since the current implementation allowed us to find out that there was an issue elsewhere (see PR https://github.com/dfinity/oisy-wallet/pull/6216).

In fact it is useful that it raises an error when the variable is not set, so that we know that we should mandatory set it.

If the idea is to implement a non-env variable (like the ones that we use for mainnet networks, BTC_MAINNET_ENABLED for example) I suggest to use `parseEnabledMainnetBoolEnvVar`

peterpeterparker: > if no default value has been provided, throw an error

The utilities do not raise any errors by design because environment variables are located at the root level and are therefore evaluated at boot time. If an error is thrown, the app crashes on load with a completely broken UI, which notably makes debugging difficult since the first issue stops anything else. The idea was also that developping locally do not necessarely needs to have all environment variables, hence the default value.

AntonioVentilii: @DecentAgeCoder sorry for the late answer.

I actually agree with you but I think we should in fact raise the issue. So that we know that there is something wrong with the set-up of the environment. That is why we have the test canisters.

Imagine that we do not raise an error. So we think that we set the variable correctly, when in reality we ddidn't. How would we know?

If you want to go down this road (I completely agree), I would suggest to raise an error directly in the deployment CI workflow, that shows when we have a missing env variable and fails the deployment.

WDYT?

PR #6145, opened 2025-04-28
fix(frontend): Show positive amount pending CK conversion tx

daviddecentage:
# Motivation

On the CK token pages, pending incoming transactions (from conversion) show a negative amount on the right side of the transaction. The value should be positive since its an incoming transaction.

# Changes

Marked the transaction as incoming to solve this issue

# Tests

On ckEth token page:
image

To ckETH on activity:
image

On USDC token page:
image

On ckUSDC token page:
image

To ckUSDC on activity:
image

On ckBTC token page:
image

On BTC token page: 
image

To ckBTC on activity:
image


AntonioVentilii (approved):
LGTM tks!

PR #6154, opened 2025-04-28
chore(build): bump ethers to v6.13.7

AntonioVentilii:
# Motivation

We want to bump ethers to v6.13.7 because it fixes issue https://github.com/ethers-io/ethers.js/issues/4951

Meaning that we can remove our workaround for it too.


peterpeterparker (approved):
LGTM, thx

PR #6157, opened 2025-04-28
feat(frontend): add Infura Gas API

AntonioVentilii:
# Motivation

According to Infura documentation, it is suggested to use the [Gas API](https://docs.metamask.io/services/reference/gas-api) as fallback for suggested gas data for all chains. Since for BSC we are having failing data, we will include the [Infura Gas API](https://docs.metamask.io/services/reference/gas-api).

# Changes

- Create rest class for Infura Gas API (very similar to `EtherscanRest`).
- Create method getSuggestedFeeData in class `InfuraGasRest` to fetch the suggested max fee data.
- Add the URL in the Content Security Policy.

# To-Do

In a follow-up PR we will use the endpoint `getSuggestedFeeData`.

# Tests

Added tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6159, opened 2025-04-28
fix(frontend): Modal tokenlist filter name symbol only

daviddecentage:
# Motivation

We want the modal tokenlist filter to only consider the symbol and name, not the network as currently.

# Changes

Removed condition to filter by network name or description.

# Tests

Before:
image

Now networks matching the search query dont show up anymore:
image


src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
```diff
@@ -478,7 +478,7 @@ describe('tokens.utils', () => {
 		it('should filter correctly by network', () => {
```
AntonioVentilii, line 478: i think you can completely remove this specific tests or change the script to "it should not filter by network"

AntonioVentilii (approved):
LGTM tks! small observation

src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
daviddecentage: Good point! Updated

PR #6165, opened 2025-04-28
feat(frontend): add checks for chain ID in `EthAddTokenReview`

AntonioVentilii:
# Motivation

Since we now have multiple EVM networks in the ERC20 user store, we need to add the check for chain ID in the flow to add custom tokens.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6167, opened 2025-04-28
feat(frontend): check for network in derived store `tokenInitialized`

AntonioVentilii:
# Motivation

We should check for network in the derived store `tokenInitialized` since now we can have tokens with the same name but on different EVM networks.


BonomoAlessandro (approved):
LGTM

PR #6178, opened 2025-04-29
refactor(frontend): upgrades VipQrCodeModal to svelte 5

BonomoAlessandro:
# Motivation

# Changes

- Upgrades `VipQrCodeModal`

### DIFF
--- a/src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
+++ b/src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
@@ -36,7 +36,7 @@
 	const maxRetriesToGetRewardCode = 3;
 	let retriesToGetRewardCode = $state(0);
 
-	let code: string = $state();
+	let code: string | undefined = $state();
 	const generateCode = async () => {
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
```diff
@@ -25,14 +25,18 @@
 	import { modalStore } from '$lib/stores/modal.store';
 	import { replacePlaceholders } from '$lib/utils/i18n.utils';
 
-	export let codeType: QrCodeType = QrCodeType.VIP;
+	interface Props {
+		codeType: QrCodeType;
```
AntonioVentilii, line 29: optional?

```suggestion
		codeType?: QrCodeType;
```

AntonioVentilii (approved):
LGTM tks! Small observation

src/frontend/src/lib/components/qr/VipQrCodeModal.svelte
BonomoAlessandro: You are right

PR #6186, opened 2025-04-29
fix(frontend): Tokenlist filter after navigate

daviddecentage:
# Motivation

If the token list filter on the homepage has a value in it, theres two scenarios where it should stay open after navigating away:
- When coming from the token page (after clicking on a token)
- When switching networks

Currently, after navigating to activity or settings page, the filter stayed open.

# Changes

Changed `$state` to `$derived` for the filter value. Using state was wrong since the value is a store and therefor the local variable must be derived.

# Tests

https://github.com/user-attachments/assets/562a7ec7-7310-4f36-9b12-fc6de4ce5244


AntonioVentilii (approved):
LGTM tks!

PR #6193, opened 2025-04-29
fix(frontend): issue with loading eth txs in batches

DenysKarmazynDFINITY:
# Motivation

This PR fixes 2 related issues:
1. Batch mechanism for fetching ETH txs had an error - instead of creating an array with promises to execute them later by the `batch` function, it was invoking the async function immediately, hence all requests to Etherscan API were made at the same time. 
2. After fixing the error above, I realized `ETHERSCAN_MAX_CALLS_PER_SECOND` was not set correctly - to fetch tx data for a token, we make 2 requests to Etherscan API, therefore that constant should be set as `rate-limit-number / 2`. Our rate limit in BETA and PROD is 10, staging - 5.

# Changes

Fixed the functions and the const.

# Tests

Adjusted.


AntonioVentilii (approved):
Very good catch! Thank you for the fix!

PR #6194, opened 2025-04-29
fix(frontend): Safari toggle size mobile

daviddecentage:
# Motivation

On mobile Safari the toggles are still very small. Seems like safari doesnt recognize or handles CSS `zoom` property correctly.

Most sites recommend using `scale` and `transform-origin`, but since it doesnt behave exactly like zoom, we need to add horizontal margins to position the toggle correctly.

# Changes

Switch to `scale` and `transform-origin`
Added `margin-inline`

# Tests

Safari Mobile:
image
image
image

Medium Desktop:
image

Large Desktop:
image


AntonioVentilii (approved):
LGTM tks!

PR #6203, opened 2025-04-29
fix(frontend): Prevent autofill in tokenlist search safari mobile

daviddecentage:
# Motivation

On Safari mobile the `autocomplete="off"` gets ignored and still shows the `fill out contacts`.

Some solutions have been tested:
- using type="search" for the input
- use `<form autocomplete="off">`  form wrapper
- use form wrapper and add hidden input same with the autocomplete property

However, what has worked in the end was to either add "search" to the inputs `name` attribute, or remove the word "name" from the `placeholder`, as the latter seems to be the reason that the browser thinks its a "name" field.

# Changes

- Added search to name attribute of input
- Changed placeholder text (and made input larger to accommodate slightly longer text)
- Added comment explaining the issue and solution

# Tests

image


AntonioVentilii (commented):
Wow, learned a new thing! Thank you!

BonomoAlessandro (approved):
LGTM

AntonioVentilii (approved):

PR #6207, opened 2025-04-29
feat(frontend): use Coingecko platform ID to map prices of all ERC20

AntonioVentilii:
# Motivation

We want to track the price for all ERC20 tokens, including the ones for EVM networks. To do so, we need to use the Coingecko platform ID for each respective network, when fetching the prices.

# Changes

- Move derived store `enabledErc20TokensAddresses` (since it is used only for that scope, preferable to have all concentrated there).
- Add additional field `coingeckoId` to all the derived stores used when creating the list of ERC20 addresses to fetch.
- Create new types for the interface that is now provided by the derived stores and by the ERC20 price request params.
- Adapt service `syncExchange` to reduce the address-with-network object to a list of network-and-addresses

# Tests

Deployed:

Screenshot 2025-04-29 at 16 37 09


BonomoAlessandro (approved):
LGTM

PR #6212, opened 2025-04-29
chore(frontend): remove deprecated field `twinToken` for SPL tokens

AntonioVentilii:
# Motivation

Since we are using group data now for grouping tokens, we don't need the prop twinToken anymore for SPL tokens. Furthermore, we remove the logic that uses it in the exchange store, since we are using directly Coingecko for the prices.


BonomoAlessandro (approved):
LGTM

PR #6213, opened 2025-04-29
refactor(frontend): use directly prop `groupData` of type `TokenUiGroup`

AntonioVentilii:
# Motivation

Using prop `groupData` directly to map header data.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6214, opened 2025-04-29
refactor(frontend): set explicit type for `data` in modal store

AntonioVentilii:
# Motivation

To be more type-safe, we should set explicitly the type of `data` field in the modal store.


BonomoAlessandro (approved):
LGTM

PR #6216, opened 2025-04-29
fix(devops): missing value for POW flag

AntonioVentilii:
# Motivation

Small mistype in the CI that deploys when fetching the flag for POW feature.

### Before

Screenshot 2025-04-30 at 00 29 29

### After

Screenshot 2025-04-30 at 00 32 18


BonomoAlessandro (approved):
LGTM

PR #6222, opened 2025-04-30
refactor(frontend): upgrades ButtonControl to svelte 5

BonomoAlessandro:
# Motivation

# Changes

- Upgrades `ButtonControl`


src/frontend/src/lib/components/ui/ButtonControl.svelte
```diff
@@ -1,12 +1,20 @@
 <script lang="ts">
-	export let ariaLabel: string;
-	export let styleClass: string | undefined = undefined;
+	import type { Snippet } from 'svelte';
+
+	interface Props {
+		children: Snippet;
+		onclick: () => void;
+		ariaLabel: string;
+		styleClass?: string | undefined;
```
AntonioVentilii, line 8: is `undefined` needed here?

AntonioVentilii (approved):
LGTM tks! One observation

src/frontend/src/lib/components/ui/ButtonControl.svelte
BonomoAlessandro: I don't think it is but most of the time we define and use it exactly like this. On some occasions we also define it as string and use a plain string as the default value.
I have no opintion on which of the two options is the better. @AntonioVentilii what do you think?

BonomoAlessandro: No, you are right

PR #6228, opened 2025-04-30
fix(frontend): issue with SendWizard

DenysKarmazynDFINITY:
# Motivation

The PR fixes issue with the SendWizard - we should fallback to DEFAULT_ETHEREUM_NETWORK if $selectedEthereumNetwork is not available when sending ETH.


AntonioVentilii (approved):
tks for the fix!

PR #6237, opened 2025-04-30
chore(rust)!: Remove migration code

bitdivine:
BREAKING CHANGE: removes migration backend canister API methods

# Motivation
The user migration code has not been used for a year.

# Changes
- Delete the canister migration code
- Where the migration code used to lock access to methods during migration, substitute the lock with "caller is not anonymous".

# Tests
See CI


.github/workflows/backend-tests.yml
```diff
@@ -124,13 +124,13 @@ jobs:
           fi
```
bitdivine, line 124: The changes to this file are in a dedicated PR: https://github.com/dfinity/oisy-wallet/pull/6245

PR #6242, opened 2025-04-30
feat(frontend): Make avatar responsive

FlavioInacta:
# Motivation

I need the avatar to have different sizes on mobile and desktop screens. To do that I compute the size of the avatar now based on the font size. Then you can just have different font-sizes for different screend. Eg. `<Avatar styleClass="font-[20px] md:font-[32px]"...>`

# Changes

- Made avatar responsive
- Moved avatar to contact folder since it is rendering contact data

# Screenshots

Stylewise nothing changed:

Screenshot_2025-04-30_13-30-01

Screenshot_2025-04-30_13-30-17


src/frontend/src/lib/components/contact/Avatar.svelte
```diff
@@ -0,0 +1,44 @@
...
+	);
+</script>
+
+{#if !initials}
+	<div class={`${commonClasses} text-brand-primary ${styleClass}`}>
+		<IconAvatar size={`${size}`}></IconAvatar>
+	</div>
+{:else}
+	<span
+		class={`${commonClasses} inline-block inline-flex items-center justify-center bg-[lightgray] font-bold text-white ${styleClass}`}
```
AntonioVentilii, line 41: do we have this `lightgray` color in our pallet? if not, may you please add it there for better manageability but also for the themes


and instead of `text-white`, shall we use the `text-foreground-*` classes? for the same reasons

src/frontend/src/lib/components/contact/Avatar.svelte
```diff
@@ -0,0 +1,44 @@
+<script lang="ts">
+	import { nonNullish } from '@dfinity/utils';
+	export type AvatarVariants = 'xl' | 'lg' | 'md' | 'sm' | 'xs';
```
AntonioVentilii, line 3: why is this type here and exported? is it used only in this component?

src/frontend/src/lib/components/contact/Avatar.svelte
FlavioInacta: I created a PR to add the necessary colors: https://github.com/dfinity/oisy-wallet/pull/6343

src/frontend/src/lib/components/contact/Avatar.svelte
FlavioInacta: Done

src/frontend/src/lib/components/contact/Avatar.svelte
AntonioVentilii: I suggest to add the single additional color in this PR, so that we can monitor when are we using them

AntonioVentilii: So, `lightgrey` is just a placeholder, understood

AntonioVentilii (approved):
LGTM tks!

PR #6245, opened 2025-04-30
fix(devops): Breaking change

bitdivine:
# Motivation
The breaking change action still isn't quite right.

# Changes
- The description is now in a variable called `DESCRIPTION`
- The name of one of the variables should be `steps.XXX` not `needs.XXX`

# Tests
- Here is an example of a breaking change PR that flags the breaking change correctly: https://github.com/dfinity/oisy-wallet/pull/6237


AntonioVentilii (approved):
Thank you!

PR #6247, opened 2025-04-30
fix(frontend): Modal tokenlist option to disable sort by balance

daviddecentage:
# Motivation

In the modal token list, if we dont display the balance we should be able disable the default sort by balance (in manage tokens for example). For these cases we can pass the already sorted list to influence the sorting.

# Changes

- Added option to disable sorting by balance
- Adjusted manage token modal to disable the default sorting and pass the already sorted list

# Tests

Now disabled on top:
image

Then all enabled:
image

Then disabled:
image


AntonioVentilii (approved):
LGTM tk!

PR #6251, opened 2025-04-30
fix(frontend): do not use lower case for contract addresses in Coingecko requests

AntonioVentilii:
# Motivation

Coingecko requests are actually case-sensitive when treating contract addresses, so we need to adjust the functions.

### Before

Screenshot 2025-04-30 at 16 24 25

### After

Screenshot 2025-04-30 at 16 23 32

### DIFF
--- a/src/frontend/src/lib/services/exchange.services.ts
+++ b/src/frontend/src/lib/services/exchange.services.ts
@@ -21,7 +21,7 @@ const fetchIcrcPricesFromCoingecko = (
 	simpleTokenPrice({
 		id: 'internet-computer',
 		vs_currencies: 'usd',
-		contract_addresses: ledgerCanisterIds.map((id) => id.toLowerCase()),
+		contract_addresses: ledgerCanisterIds,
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/services/exchange.services.ts
```diff
@@ -103,7 +103,7 @@ export const exchangeRateSPLToUsd = (
 	simpleTokenPrice({
 		id: 'solana',
 		vs_currencies: 'usd',
-		contract_addresses: tokenAddresses.map((tokenAddresses) => tokenAddresses.toLowerCase()),
+		contract_addresses: tokenAddresses.map((tokenAddresses) => tokenAddresses),
```
DenysKarmazynDFINITY, line 106: `contract_addresses: tokenAddresses` should be enough :)

AntonioVentilii: 🤦‍♂️

DenysKarmazynDFINITY (approved):

PR #6261, opened 2025-04-30
fix(build): Fix code scanning alert no. 4: Potential file system race condition

AntonioVentilii:
# Motivation

Fixes https://github.com/dfinity/oisy-wallet/security/code-scanning/4

In the script to update `did` types, there may be a race condition (CodeQL [js/file-system-race](https://github.com/github/codeql/blob/626c752a0b8cfd98bcf1f4b622d73bbe6f078a4c/javascript/ql/src/Security/CWE-367/FileSystemRace.ql)) when writing the file:

```ts
...
if (!existsSync(factoryPath)) {
  resolve();
  return;
}
...
await writeFile(factoryPath, cleanInit, 'utf8');
// The file may have changed since it was checked
...
```

So, to avoid it, we first create a temporary file with the content, then we overwrite it to the original destination file.


bitdivine (approved):
Approved assuming that there won't be multiple updates happening at the same time.  If there will be multiple updates, the tempfilename needs to be random.

AntonioVentilii:
@bitdivine good suggestion! changed!

PR #6268, opened 2025-04-30
fix(devops): tag and release CI needs credentials

AntonioVentilii:
# Motivation

The CI that creates a tag and release notes fails because it cannot find a proper user to create the tag.

See https://github.com/dfinity/oisy-wallet/actions/runs/14735249063/job/41359600671

# Changes

- Wrap the tag-creation step with silent credentials (similar to what we do when we add a commit in Github Actions).

# Tests

I simulated a new version and manually dispatched the action successfully: https://github.com/dfinity/oisy-wallet/actions/runs/14766198286


DenysKarmazynDFINITY (approved):
Thanks! That also explains why we had that unreleased v1.4.3 version (removed it and replaced with the "real" release candidate) 😄

PR #6274, opened 2025-05-01
refactor(frontend): simplify store `initReceiveTokenStore`

AntonioVentilii:
# Motivation

Small redundancy in `initReceiveTokenStore`, where we don't need to redefine the `set` function.


DenysKarmazynDFINITY (approved):
good catch!

PR #6279, opened 2025-05-01
test(frontend): remove E2E folder from coverage analysis

AntonioVentilii:
# Motivation

For now, it doesn't really make sense to include the E2E folder in the analysis for `vitest` coverage: that folder contains ONLY code for the E2E tests, so it will not be tested per-se.


DenysKarmazynDFINITY (approved):
Agree 👍

PR #6280, opened 2025-05-01
refactor(frontend): simplify list of networks for Etherscan providers

AntonioVentilii:
# Motivation

For scalability, we will use the list of supported networks (that typically includes all the allowed networks) for the providers map of Etherscan providers.


DenysKarmazynDFINITY (approved):
It changes the logic a little bit - `SUPPORTED_*` consts rely on the mainnet feature flags, but I think it should still work as expected since we anyway don't need `providers` to know about disabled networks.

PR #6281, opened 2025-05-01
refactor(frontend): include Infura network name in the `EthereumNetwork` type

AntonioVentilii:
# Motivation

For better manageability, it is better that when we add a new Ethereum/EVM network we already include all the providers data. So that any consumer is already set for a new network. To do that, the first step is to include the name of each network for Infura provider.

# Changes

- Add prop `providers` to `EthereumNetwork` type.
- Set the correct Infura name for all Ethereum/EVM networks.
- Adapt all Infura providers map to use directly the list of network with their new prop.

# Tests

Current tests are sufficient.


BonomoAlessandro (approved):
LGTM, thx

PR #6284, opened 2025-05-02
feat(frontend): Add address input

FlavioInacta:
# Motivation

To only add valid addresses the address form will need a input that can recognize addresses.

# Changes

- Adds paste functionality to the general input field
- Adds address recognition functionality
- Adds the address input field

# Demo

[vokoscreenNG-2025-05-05_09-04-55.webm](https://github.com/user-attachments/assets/b38c0b85-c1a4-4ed9-9ad1-3f95c7b0bcff)


src/frontend/src/lib/utils/address.utils.ts
```diff
@@ -23,3 +31,41 @@ export const isBtcAddress = (address: BtcAddress | undefined): boolean => {
 
 export const invalidBtcAddress = (address: BtcAddress | undefined): boolean =>
 	!isBtcAddress(address);
+
+const ADDRESS_VALIDATORS: { [key in AddressType]: (address: string) => boolean } = {
+	ICP: (address) => isIcpAccountIdentifier(address) || isIcrcAddress(address),
```
FlavioInacta, line 36: Here I am really unsure if this is correct. I was looking at ic-send.utils.ts and there is both checked (`icrcAddress` and `icpAddress`)

src/frontend/src/lib/utils/address.utils.ts
```diff
@@ -23,3 +31,41 @@ export const isBtcAddress = (address: BtcAddress | undefined): boolean => {
...
+		.map(([addressType]) => addressType as AddressType);
+
+	if (detectedTypes.length > 1) {
+		throw new Error(`Detected more than one network type: ${JSON.stringify(detectedTypes)}`);
+	}
+
+	return detectedTypes[0];
+};
+
+const ADDRESS_TYPE_TO_NETWORKS: { [key in AddressType]: Network[] } = {
```
FlavioInacta, line 63: Is this correct? Or how can I find out the valid networks?

AntonioVentilii (commented):
may you please split this PR?

AntonioVentilii (commented):
I see that we are trying to infer the network according to the input address. However there is an issue especially with networks like bitcoin and solana:

- Bitcoin: testnets have their own address, so we need to recognize if the network is mainnet or testnet. However, in this case, the util `isBtcAddress` does not help, since it just checks that the address respects the address standard. I think you can modify it to accept the network as input, not really sure, need to double check.
- Solana: same as Bitcoin, and in addition, we have ATA address, that are specific for each token, so we need to check if it is ok to accept the ATA address, and if yes, to save it for a specific token

In general, we should be careful if we decide that the task of inferring the network is on OISY side, since it will be OISY's responsibility if we imply the wrong network. While, if we let the user chose the network, we can help verify that it is correct, but at least we have a starting point

src/frontend/src/lib/utils/address.utils.ts
AntonioVentilii: as map goes yes, I would say correct. Just missing EVM networks

```ts
ICP: [ICP_NETWORK],
BTC: SUPPORTED_BITCOIN_NETWORKS,
ETH: [...SUPPORTED_ETHEREUM_NETWORKS, ...SUPPORTED_EVM_NETWORKS],
SOL: SUPPORTED_SOLANA_NETWORKS
```

however this is still in discussion, right? Since we want to use the bindings types coming from the backend (see `TokenAccountId`)

```rust
pub enum TokenAccountId {
    Icrcv2(Icrcv2AccountId),
    Sol(SolPrincipal),
    Btc(BtcAddress),
    Eth(EthAddress),
}
```

FlavioInacta: Cool, thanks!

FlavioInacta:
Just in case it's not clear. Ready for review again.

daviddecentage (approved):
LGTM, thx

StefanBerger-DFINITY:
Just a few notes form my side:

- For Bitcoin, we can just show "Bitcoin" below the field. If a user send test BTC to a mainnet account, it will fail. And vice versa. Same if he pasts the address manually.

- The Solana ATA addresses are currently an issue. Esp. since we only show the ATA address in the transaction info. So this address will only work with one token. But we have the same problem now without the address book. When you copy an address form a previous transaction, you can't use it further. We can soon make a test and only allow direct account addresses.

So the address book can be somehow 'dumb'. It's the send flow that needs to detect wrong or addresses. Users can always fix incorrect addresses, like you do with your phonebook. (But agree that we have room for improvement...)

PR #6285, opened 2025-05-02
feat(frontend): Collapsible bottom sheet component

daviddecentage:
# Motivation

We want a bottom sheet component that can display some info on mobile that usually is placed in collapsibles.

This component serves as a normal collapsible on desktop, but opens as a bottom sheet on small screens to properly display larger amounts of data.

# Changes

- Added component

# Tests

Demo:

https://github.com/user-attachments/assets/736c0050-21bd-4cf2-a791-cfc434434957

Demo of how the height adjusts to content:
image
image

Without passing `bottomSheetFooter` snippet:
image

Whats rendered on desktop:
image


peterpeterparker:
May I ask why you're implementing a custom component instead of using / improving the existing one?

https://gix.design/components/bottom-sheet

daviddecentage:
> May I ask why you're implementing a custom component instead of using / improving the existing one?
> 
> https://gix.design/components/bottom-sheet

This is just a draft of the full component including everything. The naming is probably going to be changed too since this is not only a BottomSheet. The idea was to then add the missing pieces to gix when this component has everything that we need.

peterpeterparker:
Nice!

PR #6290, opened 2025-05-02
feat(frontend): implements isEndedCampaign function

BonomoAlessandro:
# Motivation

We want to be able to find out if a campaign has ended exactly like we already check if a campaign is upcoming or ongoing.

# Changes

- implements `isEndedCampaign`


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6294, opened 2025-05-02
feat(frontend): supports ended date on RewardDateBadge

BonomoAlessandro:
# Motivation

The `RewardDateBadge` should not only support active dates but also ended dates.

# Changes

- supports ended dates

# Tests
image

image


src/frontend/src/lib/components/rewards/RewardDateBadge.svelte
```diff
@@ -1,13 +1,21 @@
 <script lang="ts">
 	import DateBadge from '$lib/components/ui/DateBadge.svelte';
 	import { i18n } from '$lib/stores/i18n.store';
+	import { isEndedCampaign } from '$lib/utils/rewards.utils';
 
 	interface Props {
 		date: Date;
 		testId?: string | undefined;
```
AntonioVentilii, line 8: since we are here, can undefined be removed from here since it is optional"

`testId?: string;`

AntonioVentilii (approved):
LGTM tks! Small request pls

src/frontend/src/lib/components/rewards/RewardDateBadge.svelte
BonomoAlessandro: Good point

PR #6299, opened 2025-05-03
refactor(frontend): apply lint rule `no-import-type-side-effects`

AntonioVentilii:
# Motivation

When bumping the lint library to the next version in PR https://github.com/dfinity/oisy-wallet/pull/6267, the CI actions apply the auto-fix with `eslint . --fix`; however, according to OISY Bot Policy, a bot's PR is not allowed to change some of those files (for example components, services, utils).

So, we run the auto-fix for rule [no-import-type-side-effects](https://typescript-eslint.io/rules/no-import-type-side-effects/) in a separate PR.


peterpeterparker (approved):
Nice, thx!

PR #6300, opened 2025-05-03
refactor(e2e): do not deconstruct params when creating page

AntonioVentilii:
# Motivation

It is more scalable to avoid deconstructing the parameters that are passed on construction of a certain `page` object. So that we guarantee that any new one is passed too to their parent.

That means that some tests will receive the `isMobile` prop too, leading to a "shift" of screenshots: we take 2 more screenshots for mobile devices, expanding the entire viewport.


BonomoAlessandro (approved):
LGTM

PR #6308, opened 2025-05-05
feat(frontend): Adds reset button to name input of address form

FlavioInacta:
# Motivation

For easily resetting the name field, the user should see the reset button of the common Input component.


src/frontend/src/lib/components/ui/InputText.svelte
```diff
@@ -14,6 +15,7 @@
 	{required}
 	bind:value
 	{placeholder}
+	{showResetButton}
```
AntonioVentilii, line 18: just for my understanding: will there be other instances in which we are going to use the custom `Input` component?

FlavioInacta: Yes, in the address field

AntonioVentilii (approved):
LGTM tks!

PR #6319, opened 2025-05-05
refactor(frontend): removes rewards enabled feature flag

BonomoAlessandro:
# Motivation

Since the feature `Rewards` is fully used we don't need the feature flag anymore.

# Changes

- removes feature flag


AntonioVentilii (approved):
Thank you!!!

P.S. as discussed, remove it from the the other instances too (search for `VITE_AIRDROPS_ENABLED` in the code)

PR #6322, opened 2025-05-05
feat(frontend): extract component to load at defined intervals

AntonioVentilii:
# Motivation

The part of component `LoaderEthTransactions` that loads transactions at specific intervals is actually useful for more components (see PRs #6320 and #6321). So we first extract it to a separate new component `IntervalLoader`.

# Changes

- Create new component `IntervalLoader` that has the logic of a timer at specific intervals.
- Use new component in `LoaderEthTransactions`.

# Tests

Create tests.


DenysKarmazynDFINITY (approved):
Nice one 👍

PR #6328, opened 2025-05-05
chore(rewards): Bump rewards canister

bitdivine:
# Motivation
The original airdrop logic was never used.  Maintaining compatibility with the unused code has become a burden.

# Changes
- Delete the API endpoints for the original airdrop logic.

# Tests
- Existing tests should be using only the new airdrop logic, so should be unaffected.
- Some configuration may be deleted; let us see.


AntonioVentilii (approved):
Thank you!

PR #6333, opened 2025-05-05
feat(frontend): added icp swap factory script and updated dfx.json

roman-nazaruk:
# Motivation

This PR includes all the necessary steps for the integration of the icp_swap_factory canister.

# Changes

- Script for building icp_swap_factory by using the respective WASM and candid files available in the ICP_SWAP repo.
- Script for deploying icp_swap_factory locally.
- Script for generating, updating and deleting declaration files (src/declarations/icp_swap_factory/*).
- Dfx config update with the icp_swap_factory canister.


scripts/build.icp_swap_factory.sh
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+print_help() {
+  cat <<-EOF
```
bitdivine, line 5: Note: The `-` in `<<-EOF` means that you can indent the heredoc with tabs, if you wish.  No need, I just note that the option for indenting is there but not used.

Manual: https://www.gnu.org/software/bash/manual/bash.html#Here-Documents

> If the redirection operator is ‘<<-’, then all leading tab characters are stripped from input lines and the line containing delimiter. This allows here-documents within shell scripts to be indented in a natural fashion.

bitdivine (approved):
LGTM.  Thanks!

scripts/build.icp_swap_factory.sh
roman-nazaruk: Yeah, I reused it from the KongSwap script to keep things more consistent.

PR #6334, opened 2025-05-05
feat(backend): providing implementation for get_allowed_cycles endpoint

DecentAgeCoder:
# Motivation

This PR provides the implementation for the `get_allowed_cycles` endpoint

# Changes
- Created implementation for get_allowed_cycles in signer.rs that calls the `icrc_2_allowance` on the Cycles Ledger service.
- Added  function call get_allowed_cycles in signer.rs to retrieve the cycles allowance for users.
- Handled possible errors like `FailedToContactCyclesLedger` and `others` within the new functionality.

# Tests

- Added unit tests in `tests/it/signer.rs` to validate `get_allowed_cycles`, covering both success and error cases. Includes scenarios for custom error messages and unknown errors.


bitdivine (approved):
As mentioned, this is more complicated, slower, more expensive and more error prone than making a direct query call.

If you need to ship e.g. today, I'd recommend merging this & making a ticket to change later.  If you have more time I'd strongly recommend making a direct query call now, instead of this.

PR #6337, opened 2025-05-05
feat(scripts): run format and lint only for specific folders

AntonioVentilii:
# Motivation

When the code runs `npm run format` and `npm run lint`, in a few occasions, it should be able to run for specific folders or files, to spare time.

For example, if we run the scripts to update the tokens, it should format only the specific token JSON file. Or if we update the bindings, it should format only the `src/declarations` folder.

Credits to https://stackoverflow.com/a/60458728


roman-nazaruk (approved):
LGTM, thnx

PR #6339, opened 2025-05-05
refactor(frontend): Manage token modal new design

daviddecentage:
# Motivation

When refactoring the manage token modals, we didnt update the individual token cards, which should display the same information as on the homepage token list.

# Changes

Updated title, subtitle and description of token card to match the homepage token list, displaying Network, Name and symbol instead of just the name.

# Tests

Before:
image

After:
image


BonomoAlessandro (approved):
LGTM

PR #6346, opened 2025-05-06
feat(frontend): added feature flag for icp_swap

roman-nazaruk:
# Motivation

We aim to integrate with ICP Swap. The goal of this PR is to introduce a FEATURE_FLAG that allows us to enable or disable ICP_SWAP logic across different environments for testing and flexibility.

# Changes

- Created an .env file and used parsedBoolEnv utility to handle environment flags.
- Updated the deployment script to support the ICP_SWAP feature flag.


AntonioVentilii (changes requested):
May you please add it in `tag-release.yml` file too?

roman-nazaruk:
> tag-release.yml



> May you please add it in `tag-release.yml` file too?

Added

src/frontend/src/env/icp_swap.env.ts
AntonioVentilii, line 1: may you please rename it to `icp-swap.env.ts` for consistency?

AntonioVentilii (approved):
LGTM tks! One last observation

src/frontend/src/env/icp_swap.env.ts
roman-nazaruk: Updated. Thanks. Good catch

PR #6348, opened 2025-05-06
feat(frontend): Update rewards campaign end date

StefanBerger-DFINITY:
# Motivation

Current end date shows May 7.
We need a bit more time to have the next episode ready.

# Changes

Change May 7 to May 21


AntonioVentilii (approved):
tks!

PR #6349, opened 2025-05-06
fix(rust): Address cargo audit issues

bitdivine:
# Motivation
We have two rust dependencies that have known security issues.  While they don't appear to affect us, it is best to update them.

# Changes
- Update `cargo-audit` to the latest version.
- Resolve issues

# Tests
Existing CI should verify that functionality is unaffected.


AntonioVentilii (approved):
thank you for the fix!

PR #6350, opened 2025-05-06
refactor(frontend): rename IDB addresses functions

AntonioVentilii:
# Motivation

We are going to introduce IDB stores for transactions too. So to keep it more organized, we rename first the current IDB modules to `*-addresses` since they refer just to addresses functions.


roman-nazaruk (approved):
LGTM

PR #6352, opened 2025-05-06
chore(rewards): Update rewards canister version

bitdivine:
# Motivation
- A new version of the rewards canister is available.

# Changes
- Update the rewards canister API to version 0.5.1
  - `campaign_name` is now deprecated.  `campaign_id` is the recommended alternative. It has the same value as `campaign_name` but is not optional.  `campaign_name` will eventually be deleted.
  - A new rewards canister API method is provided: `eligible()`.  The method is not implemented in the rewwards canister but the API signature is provided.

# Tests


AntonioVentilii (approved):
LGTM tks!

PR #6360, opened 2025-05-06
feat(frontend): displays network name on network level

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/components/receive/ReceiveModal.svelte
+++ b/src/frontend/src/lib/components/receive/ReceiveModal.svelte
@@ -15,10 +15,13 @@
 
 	export let network: Network;
 	export let copyAriaLabel: string;
+
+	let title: string | undefined;
+	$: title = isNullish($token) ? network.name : addressToken?.symbol;
 </script>
 
 <Modal on:nnsClose={modalStore.close}>
-	<ReceiveTitle slot="title" {addressToken} />
+	<ReceiveTitle slot="title" {title} />
 
 	<ContentWithToolbar>
 		<ReceiveAddressQRCodeContent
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/receive/ReceiveModal.svelte
```diff
@@ -15,10 +17,13 @@
 
 	export let network: Network;
 	export let copyAriaLabel: string;
+
+	let title: string | undefined;
+	$: title = isNullish($token) ? network.name : addressToken?.symbol;
```
AntonioVentilii, line 22: i would prefer to pass the `token` prop (and get the network from there too) and not use the `tokenStore` if possible: the idea is to deprecate this store somehow, so better to reduce its usage instead of increasing

AntonioVentilii (changes requested):

src/frontend/src/lib/components/receive/ReceiveModal.svelte
BonomoAlessandro: I need a way to find out whether we are on the network or token level. What i can do is to replace the `$token` with `$pageToken`. This store is not depricated.

AntonioVentilii: ah! understood now! thank you for the explanation

src/frontend/src/lib/components/receive/ReceiveModal.svelte
```diff
@@ -1,11 +1,14 @@
 <script lang="ts">
 	import { Modal } from '@dfinity/gix-components';
+	import { isNullish } from '@dfinity/utils';
 	import type { Snippet } from 'svelte';
+	import { derived } from 'svelte/store';
```
AntonioVentilii, line 5:
```suggestion
```

AntonioVentilii (approved):
LGTM tks!

PR #6361, opened 2025-05-06
feat(frontend): implement SendDestination component

DenysKarmazynDFINITY:
# Motivation

We need to implement a new component to display destination address in the Send forms.

Screenshot 2025-05-06 at 13 21 36


src/frontend/src/lib/components/send/SendDestination.svelte
```diff
@@ -0,0 +1,37 @@
...
+	import IconPenLine from '$lib/components/icons/IconPenLine.svelte';
+	import NetworkLogo from '$lib/components/networks/NetworkLogo.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { SEND_CONTEXT_KEY, type SendContext } from '$lib/stores/send.store';
+
+	interface Props {
+		destination: string;
+		invalidDestination?: boolean;
+	}
+	let { destination, invalidDestination = false }: Props = $props();
```
AntonioVentilii, line 12: will the validation happen in this component? if yes, may you please had a TODO for posterity?

src/frontend/src/lib/components/send/SendDestination.svelte
```diff
@@ -0,0 +1,37 @@
...
+	<div
+		class="flex items-center rounded-lg border bg-secondary px-2 py-3"
+		class:border-tertiary={!invalidDestination}
+		class:border-error-solid={invalidDestination}
+	>
+		<NetworkLogo network={$sendToken.network} />
+
+		<div class="w-full truncate pl-2 pr-4 text-sm sm:text-base">{destination}</div>
+
+		<button class="text-brand-primary" onclick={back} aria-label={$i18n.core.text.back}>
```
AntonioVentilii, line 33: i am not sure i understand the action of this button: will it make the component deploy a "go-back" action?

src/frontend/src/lib/components/send/SendDestination.svelte
DenysKarmazynDFINITY: This is just a UI component, no validation is planned here - the respective prop will always be sent down from parent

src/frontend/src/lib/components/send/SendDestination.svelte
DenysKarmazynDFINITY: Yes, if user clicks this button, wizard goes back to the previous step (Destination step). But I see your point, let's make it a bit more descriptive.

PR #6364, opened 2025-05-06
refactor(devops): rename tests status checks

AntonioVentilii:
# Motivation

If a status check have the same name of another, it is considered passed if even only one of them passes. So, to avoid false positive, we change the status check name for all the required CI workflows to be unique.


bitdivine (approved):
Thanks!

PR #6366, opened 2025-05-06
feat(frontend): Add paste button to general input

FlavioInacta:
# Motivation

To make it easier for the user to paste the clipboard into a input field, this PR adds a button to the inner end of the input component.

# Changes

- Adds a paste button
- Fixes padding issue with variable inner end content.

# Tests

See paste button in action: https://github.com/dfinity/oisy-wallet/pull/6284


src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -1,37 +1,84 @@
...
 	};
+
+	const handlePaste = () => {
+		try {
+			navigator.clipboard.readText().then((text) => {
+				value = text;
+			});
+		} catch (e) {
+			console.error('Failed to paste clipboard', e);
+		}
```
AntonioVentilii, line 40: - let's put the function in the `share.utils` module (you can see there are some examples of similar functions there)
- do you need the error handler? what is the case when this fails? for sure I would handle the case if the text is nullish for example:
```ts
const clipboardText = await navigator.clipboard.readText()
if (isNullish(clipboardText)) {return}
value = clipboardText;
```
so we even avoid the usage of `.then()`

src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -1,37 +1,84 @@
...
-				<IconClose />
-			</button>
-		{/if}
-	</svelte:fragment>
-</GixInput>
+{#snippet resetButton()}
+	<div transition:fade>
+		<ButtonIcon
+			styleClass="border-0 px-0 py-2 text-tertiary"
+			colorStyle="tertiary-alt"
```
AntonioVentilii, line 48: is `text-tertiary` necessary if we define the color style as `tertiary-alt`; why a different pallet? if it is by design, may you please double-check that this does not cause inconsistency in the other buttons? we may have different pallets of color for the same buttons

src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -1,37 +1,84 @@
...
+		>
+			<IconClose slot="icon" size="24" />
+		</ButtonIcon>
+	</div>
+	{#if nonNullish(innerEnd) || showPasteButton}
+		<div class="border-r-1 border-black/20" transition:fade></div>
+	{/if}
+{/snippet}
+
+{#snippet pasteButton()}
```
AntonioVentilii, line 60: instead of creating a snippet, let's create a `Paste` component, separated from this one, and use it here (similar to `Copy` component). Better for reusability and manageability

src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -1,37 +1,84 @@
...
-				class="text-tertiary"
-				on:click={reset}
-				aria-label={resetButtonAriaLabel ?? $i18n.convert.text.input_reset_button}
-			>
-				<IconClose />
-			</button>
-		{/if}
-	</svelte:fragment>
-</GixInput>
+{#snippet resetButton()}
```
AntonioVentilii, line 44: same

src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -1,37 +1,84 @@
...
+	<GixInput {...props} bind:value on:nnsInput>
+		<svelte:fragment slot="inner-end">
+			<div bind:clientWidth={endWidth} class="flex pl-2">
+				{#if nonNullish(value) && notEmptyString(value.toString()) && showResetButton}
+					{@render resetButton()}
+				{/if}
+				{#if showPasteButton}
+					{@render pasteButton()}
+				{/if}
+				{@render innerEnd?.()}
```
AntonioVentilii, line 80: will this be used in another PR?

src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -1,37 +1,84 @@
...
+	{/if}
+{/snippet}
+
+{#snippet pasteButton()}
+	<Button
+		type="button"
+		styleClass="text-sm px-1 py-2 mx-1"
+		colorStyle="tertiary-alt"
+		ariaLabel={$i18n.core.text.paste}
+		on:click={handlePaste}>{$i18n.core.text.paste}</Button
```
AntonioVentilii, line 66:
```suggestion
		onclick={handlePaste}>{$i18n.core.text.paste}</Button
```

src/frontend/src/lib/components/ui/Input.svelte
```diff
@@ -1,37 +1,84 @@
...
-			</button>
-		{/if}
-	</svelte:fragment>
-</GixInput>
+{#snippet resetButton()}
+	<div transition:fade>
+		<ButtonIcon
+			styleClass="border-0 px-0 py-2 text-tertiary"
+			colorStyle="tertiary-alt"
+			on:click={handleReset}
```
AntonioVentilii, line 49:
```suggestion
			onclick={handleReset}
```

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: Done. Please check if it's how you wanted it.

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: You were right. The problem was, that it was rendered as a link which overwrote the color. I could now use the muted style.

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: Done. Is it ok how I did it?

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: Yes

src/frontend/src/lib/components/ui/Input.svelte
FlavioInacta: Done, actually the `Button` component still uses `on:click`, but I used `onclick` for the new buttons.

PR #6368, opened 2025-05-06
feat(frontend): implement getKnownDestinations util

DenysKarmazynDFINITY:
# Motivation

In order to get all the needed info about known destinations from transactions list, we need to implement a new util.

### DIFF
--- a/src/frontend/src/lib/utils/transactions.utils.ts
+++ b/src/frontend/src/lib/utils/transactions.utils.ts
@@ -302,7 +302,7 @@ export const getKnownDestinations = (transactions: AnyTransactionUi[]): KnownDes
 											? Math.max(Number(acc[address].timestamp), Number(timestamp))
 											: nonNullish(timestamp)
 												? Number(timestamp)
-												: undefined
+												: acc[address].timestamp
 								}
 							}),
 							{}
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/utils/transactions.utils.ts
```diff
@@ -285,3 +285,29 @@ export const areTransactionsStoresLoading = (
...
+							(innerAcc, address) => ({
+								...innerAcc,
+								[address]: {
+									amounts: [...(nonNullish(acc[address]) ? acc[address].amounts : []), value],
+									timestamp:
+										nonNullish(acc[address]?.timestamp) && nonNullish(timestamp)
+											? Math.max(Number(acc[address].timestamp), Number(timestamp))
+											: nonNullish(timestamp)
+												? Number(timestamp)
+												: undefined
```
AntonioVentilii, line 305: what if `timestamp` is nullish and `acc[address]?.timestamp` is not? for now it defaults to undefined, correct?

DenysKarmazynDFINITY: good catch, now it defaults to acc[address].timestamp which is either number or undefined

AntonioVentilii (approved):
LGTM tks!

PR #6371, opened 2025-05-06
refactor(backend)!: create result type for `add_user_credential` method

AntonioVentilii:
# Motivation

In an effort to make the did file more readable, we create a specific type for the results of method `add_user_credential`.

Suggestion by @DecentAgeCoder  
Credits to @bitdivine 

BREAKING CHANGE: the interface will be changed


src/backend/src/lib.rs
```diff
@@ -498,16 +500,17 @@ pub async fn btc_get_pending_transactions(
...
     let stored_principal = StoredPrincipal(user_principal);
     let current_time_ns = u128::from(time());
 
-    let (vc_flow_signers, root_pk_raw, credential_type, derivation_origin) =
+    let Some((vc_flow_signers, root_pk_raw, credential_type, derivation_origin)) =
         read_config(|config| find_credential_config(&request, config))
-            .ok_or(AddUserCredentialError::ConfigurationError)?;
+    else {
+        return AddUserCredentialResult::Err(AddUserCredentialError::ConfigurationError);
+    };
```
AntonioVentilii, line 513: I am not really convinced about this change: is there a prettier way of doing it? The lint issue I had is that I cannot use `?` now that I have `AddUserCredentialResult` as output

bitdivine: Personally I make an inner implementation that returns a normal result, then convert when the result is returned.  This way ? can be used in the inner implementation.  This has another advantage: The file with all the entry points can become massive, and this way I can break it down into small, focused files.  It also helps with testing.

See e.g.: https://github.com/dfinity/oisy-rewards-canister/blob/airdrop/src/oisy_rewards_canister/src/main.rs#L91

However splitting out all the function bodies into other files is a bigger change and I wouldn't include it in this PR.

What you have is as clean as we can get while the functions are implemented at the entry points.

src/declarations/backend/backend.factory.did.js
```diff
@@ -489,24 +486,24 @@ export const idlFactory = ({ IDL }) => {
 		current_user_version: IDL.Opt(IDL.Nat64)
 	});
 	return IDL.Service({
-		add_user_credential: IDL.Func([AddUserCredentialRequest], [Result], []),
-		add_user_hidden_dapp_id: IDL.Func([AddHiddenDappIdRequest], [Result_1], []),
-		allow_signing: IDL.Func([IDL.Opt(AllowSigningRequest)], [Result_2], []),
-		btc_add_pending_transaction: IDL.Func([BtcAddPendingTransactionRequest], [Result_3], []),
-		btc_get_pending_transactions: IDL.Func([BtcGetPendingTransactionsRequest], [Result_4], []),
-		btc_select_user_utxos_fee: IDL.Func([SelectedUtxosFeeRequest], [Result_5], []),
+		add_user_credential: IDL.Func([AddUserCredentialRequest], [AddUserCredentialResult], []),
```
bitdivine, line 489: I guess it makes sense to start with Result_10, or whatever is the highest number, and work down to minimize the diffs. :wink:

bitdivine (approved):
Looks good.  Thanks!

src/backend/backend.did
```diff
@@ -78,6 +78,7 @@ type AddUserCredentialRequest = record {
 	current_user_version : opt nat64;
 	credential_spec : CredentialSpec
 };
+type AddUserCredentialResult = variant { Ok; Err : AddUserCredentialError };
```
bitdivine, line 81: Reading through the PR, I am surprised that this is classified as a breaking change in candid.  If I get a moment I'll have a poke.  I'd need to see the exact change report.

src/declarations/backend/backend.factory.did.js
AntonioVentilii: completely right, i didn't think of that! for the next PR i will do result_9

PR #6372, opened 2025-05-06
fix(frontend): Placeholder logo opacity

daviddecentage:
# Motivation

The opacity is still off for the fallback icon.

# Changes

Fixed the condition for `loaded` and renamed the variable `isReady` for clarity

# Tests

Before:
image

After:
image


src/frontend/src/lib/components/ui/Logo.svelte
```diff
@@ -19,7 +19,7 @@
 	let sizePx = $state(logoSizes[size]);
 
 	let loadingError: boolean | undefined = $state();
-	let loaded = $derived(nonNullish(src) && nonNullish(loadingError) && !loadingError);
+	let loaded = $derived((nonNullish(src) && nonNullish(loadingError)) || isNullish(src));
```
DenysKarmazynDFINITY, line 22: Shouldn't the condition be `((nonNullish(src) && nonNullish(loadingError) && !loadingError) || isNullish(src))`? Otherwise, the flag `loaded` will be true even if `loadingError` is true which seems a bit strange.

daviddecentage: If the condition is as you say, and the image fails to load the placeholder will be displayed with the 10% opacity as well.

It does make sense, because even if there is an error while loading it, the loading is complete and loader should be true.

DenysKarmazynDFINITY: Thanks! 👍

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6373, opened 2025-05-06
feat(frontend): implement KnowDestination component

DenysKarmazynDFINITY:
# Motivation

To display information about a known destination, we need to implement a new component.

Screenshot 2025-05-06 at 18 21 32

### DIFF
--- a/src/frontend/src/lib/components/send/KnownDestination.svelte
+++ b/src/frontend/src/lib/components/send/KnownDestination.svelte
@@ -1,0 +1,64 @@
+<script lang="ts">
+	import { nonNullish } from '@dfinity/utils';
+	import { normalizeTimestampToSeconds } from '$icp/utils/date.utils';
+	import IconConvertTo from '$lib/components/icons/IconConvertTo.svelte';
+	import Amount from '$lib/components/ui/Amount.svelte';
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/components/send/KnownDestination.svelte
```diff
@@ -0,0 +1,64 @@
...
+		destination: Address;
+		amounts: bigint[];
+		timestamp?: number;
+	}
+	let { token, destination, amounts, timestamp }: Props = $props();
+
+	// we only display the first 3 amounts, and the rest is displayed as "+N more"
+	let amountsToDisplay = $state(amounts.slice(0, MAX_DISPLAYED_KNOWN_DESTINATION_AMOUNTS));
+
+	let restAmountsNumber = $state(amounts.length - amountsToDisplay.length);
```
AntonioVentilii, line 29: shouldn't these be `$derived` since amounts is a prop?

DenysKarmazynDFINITY: Yes, better to use $derived. Updated!

AntonioVentilii (approved):
LGTM tks!

PR #6382, opened 2025-05-06
refactor(frontend): correct PascalCase rename components with QR in the name

AntonioVentilii:
# Motivation

The correct PascalCase syntax for components with QR in the name is `*Qr*` and not `*QR*`, even if it may feel incorrect.


DenysKarmazynDFINITY (approved):
Thanks :)

PR #6383, opened 2025-05-07
chore(rewards): Update rc version

bitdivine:
# Motivation
The number of transactions also has a time limit, as the rewards canister cannot store transactions forever.  So there is a new rewards canister API.

# Changes
- Make the usage award criteria dictionaries, so that adding fields is less disruptive.
- Add duration to the transactions requirement.

# Tests


BonomoAlessandro (approved):
LGTM

PR #6384, opened 2025-05-07
feat(frontend): fixes icp receive modal

BonomoAlessandro:
# Motivation

When opening the `IcReceive` modal, the `loadTokenAndRun` function was called. `loadTokenAndRun` sets the token into the tokenstore, which is unnecessary because we are working with a local context. Additionaly this led to unexpected behavior in other parts of the application because the tokenstore was never reset.

# Changes

- Don't set token into tokenstore anymore

# Tests

**before:**

https://github.com/user-attachments/assets/07d1f834-a834-4862-9153-f3db8c5dcc31

**after:**

https://github.com/user-attachments/assets/3ca4ec5b-ea5d-4264-b3da-4e99eefa7815


src/frontend/src/icp/components/receive/IcReceive.svelte
```diff
@@ -33,7 +32,7 @@
 	$: icrc = token.standard === 'icrc';
 
 	const open: LoadTokenAndOpenModal = async (callback: () => Promise<void>) => {
```
DenysKarmazynDFINITY, line 34: Can you please refactor the component according to the changes? E.g. the type here can now be changed since no more loading happens. Same with the `close` function - type and the comment about resetting the token seem to be outdated. Additionally, please check if we can also remove the `dispatch('nnsClose')` from `close` - the parent does not listen for any events.

BonomoAlessandro: You are right. The `dispatch('nnsClose')` could also be removed.

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6385, opened 2025-05-07
feat(frontend): map Token ID for ICRC tokens

AntonioVentilii:
# Motivation

To standardize the tokens store, we need to first make it so the object that they treat are all compatible with interface `Token`. 

So, we first map the ICRC tokens to have the Token ID, instead of adding it later.


DenysKarmazynDFINITY (approved):
LGTM, thanks! I'm wondering why it was originally implemented to not have id.

PR #6386, opened 2025-05-07
feat(frontend): utils to check IC token standards

AntonioVentilii:
# Motivation

For better manageability, we create utils functions to check the standards for `icp` and `icrc`, and combine them for IC tokens in general.

# Changes

- Create util for standard `icp`.
- Create util for standard `icrc`.
- Merge the utils into a single one to check if the token is an IC token.
- Use all three above in the code.

# Tests

New tests created.


DenysKarmazynDFINITY (approved):
Good one! Thanks.

PR #6388, opened 2025-05-07
feat(frontend): implement KnownDestinations component

DenysKarmazynDFINITY:
# Motivation

To display recent destinations, we need to add a new component.

Screenshot 2025-05-07 at 10 23 11

### DIFF
--- a/src/frontend/src/lib/components/send/KnownDestinations.svelte
+++ b/src/frontend/src/lib/components/send/KnownDestinations.svelte
@@ -8,19 +8,19 @@
 	import type { KnownDestinations } from '$lib/types/transactions';
 
 	interface Props {
-		data?: KnownDestinations;
+		knownDestinations?: KnownDestinations;
 		destination: string;
 	}
-	let { data, destination = $bindable() }: Props = $props();
+	let { knownDestinations, destination = $bindable() }: Props = $props();
 
 	const dispatch = createEventDispatcher();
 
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/send/KnownDestinations.svelte
```diff
@@ -0,0 +1,45 @@
...
+	import { nonNullish } from '@dfinity/utils';
+	import { createEventDispatcher, getContext } from 'svelte';
+	import { fade } from 'svelte/transition';
+	import KnownDestination from '$lib/components/send/KnownDestination.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { SEND_CONTEXT_KEY, type SendContext } from '$lib/stores/send.store';
+	import type { KnownDestinations } from '$lib/types/transactions';
+
+	interface Props {
+		data?: KnownDestinations;
```
AntonioVentilii, line 11: would it be possible to rename it as something more explicit? `knownDestinations` or something like that?

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/components/send/KnownDestinations.svelte
DenysKarmazynDFINITY: Absolutely! Updated.

PR #6390, opened 2025-05-07
refactor(frontend): Activity list register/recognize CKeth and CKbtc listeners

daviddecentage:
# Motivation

In order to always make pending txs visible in the activity list, we register the pending tx listeners in the MultipleListeners on the activity page.

# Changes

- Added condition to handle ICP CK tokens in the `mapListeners` function which is used in `MultipleListeners`
- Updated types to support the new listeners

# Tests

Added tests to existing utils tests

Correctly displays ckBTC and ckETH pending transactions:

image
image


AntonioVentilii (approved):
Thnk you!

PR #6392, opened 2025-05-07
test(storybook): Adds storybook

FlavioInacta:
# Motivation

Integrates storybook. 

Run with `npm run storybook`.

This is a proposal PR. Probably open questions:

- Where to put stories?
- Where to put sotrybook decorators?
- Which components do we want in storybook?
- Do we want to mock services or directly interact with existing backend? (I suggest mocks for different scenarios)
- Do we want do deploy storybook somewhere (that designers or POs can test design)?
- At the moment you cannot easily add stories with slots, for this to work stories would have to be written in svelte, but this did not really work when I tried it. Do we need that? You could write wrapper components to add slots to your components and add these to you stories.
- ...

TODO:
- Ensure storybook code does not end up in production code (eslint?)

# Demo

[storybook.webm](https://github.com/user-attachments/assets/57a09254-1ee0-46b3-9f68-0aaaaae1df50)


src/frontend/src/tests/lib/components/address-book/AddContactStep.spec.ts
```diff
@@ -29,23 +29,23 @@ describe('AddContactStep', () => {
 	});
 
 	it('should disable save button when form is invalid', () => {
-		const addContact = vi.fn();
-		const close = vi.fn();
+		const onAddContact = vi.fn();
```
FlavioInacta, line 32: When the events start with `on...` then storybook automatically adds callbacks and shows when they are called.

AntonioVentilii:
Storybook proposal from May 2025 with open product questions that were never resolved. Closing — please reopen with a concrete proposal if we decide to adopt Storybook.

PR #6395, opened 2025-05-07
fix(frontend): use styleClass in LogoButton

DenysKarmazynDFINITY:
# Motivation

Small fix for the LogoButton component - `styleClass` was added but never used.


AntonioVentilii (approved):
Thank you for the fix!

PR #6396, opened 2025-05-07
fix(frontend): onclick event in Input

DenysKarmazynDFINITY:
# Motivation

While building the frontend, noticed there is a linter warning in the `Input` component.


AntonioVentilii (approved):
Thank you for the fix!

PR #6401, opened 2025-05-07
feat(frontend): pass network ID to transactions store

AntonioVentilii:
# Motivation

# Changes

# Tests


AntonioVentilii:
no need anymore for this PR: the idea was to have the caching in the store per se. However, now we do it in the workers/services directly

PR #6402, opened 2025-05-07
feat(frontend): added new icp swap utils methods

roman-nazaruk:
# Motivation

Added new icp swap utils methods to calculate slippage and map swap responses to common interface

# Changes

Added new icp swap utils methods: 
- mapIcpSwapResult
- mapKongSwapResult
- calculateSlippage

# Tests

Covered new logic with tests

### DIFF
--- a/src/frontend/src/lib/types/swap.ts
+++ b/src/frontend/src/lib/types/swap.ts
@@ -4,7 +4,26 @@ export type SwapSelectTokenType = 'source' | 'destination';
 
 export type DisplayUnit = 'token' | 'usd';
 
+export type SwapProvider = 'kongSwap' | 'icpSwap';
+
 export interface ProviderFee {
 	fee: bigint;
 	token: Token;
+}
+
+export interface ICPSwapResult {
+	receiveAmount: bigint;
+}
+
+export type Slippage = string | number;
+
+export interface SwapMappedResult {
+	provider: SwapProvider;
+	receiveAmount: bigint;
+	slippage?: number;
+	route?: string[];
+	liquidityFees?: ProviderFee[];
+	receiveOutMinimum?: bigint;
+	networkFee?: ProviderFee;
+	swapDetails: unknown;
 }
--- a/src/frontend/src/lib/utils/swap.utils.ts
+++ b/src/frontend/src/lib/utils/swap.utils.ts
@@ -1,7 +1,15 @@
-import type { SwapAmountsTxReply } from '$declarations/kong_backend/kong_backend.did';
+import type {
+	SwapAmountsReply,
+	SwapAmountsTxReply
+} from '$declarations/kong_backend/kong_backend.did';
 import { isIcToken } from '$icp/validation/ic-token.validation';
 import { ZERO } from '$lib/constants/app.constants';
-import type { ProviderFee } from '$lib/types/swap';
+import {
+	ICP_SWAP_PROVIDER,
+	KONG_SWAP_PROVIDER,
+	SWAP_DEFAULT_SLIPPAGE_VALUE
+} from '$lib/constants/swap.constants';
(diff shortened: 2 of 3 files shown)


src/frontend/src/lib/utils/swap.utils.ts
```diff
@@ -51,3 +59,62 @@ export const getNetworkFee = ({
...
+ */
+export const calculateSlippage = ({
+	quoteAmount,
+	slippagePercentage
+}: {
+	quoteAmount: bigint;
+	slippagePercentage: number;
+}): bigint => {
+	const slippageFactor = BigInt(10000 - Math.floor(slippagePercentage * 100));
+	return (quoteAmount * slippageFactor) / 10000n;
```
AntonioVentilii, line 119: this will round the number since it is bigint (so for example `quoteAmount * slippageFactor = 12345n` will become `1n`), is that ok?

src/frontend/src/lib/types/swap.ts
```diff
@@ -4,7 +4,26 @@ export type SwapSelectTokenType = 'source' | 'destination';
 
 export type DisplayUnit = 'token' | 'usd';
 
+export type SwapProvider = 'kongSwap' | 'icpSwap';
```
AntonioVentilii, line 7: now that i see the usage, i think it would be better to use enums for `ICP_SWAP_PROVIDER` and `KONG_SWAP_PROVIDER` instead of constants. It provides better type-maneagability

or do you see an issue with it?

src/frontend/src/lib/types/swap.ts
```diff
@@ -4,7 +4,26 @@ export type SwapSelectTokenType = 'source' | 'destination';
...
+export interface SwapMappedResult {
+	provider: SwapProvider;
+	receiveAmount: bigint;
+	slippage?: number;
+	route?: string[];
+	liquidityFees?: ProviderFee[];
+	receiveOutMinimum?: bigint;
+	networkFee?: ProviderFee;
+	swapDetails: unknown;
+}
```
AntonioVentilii, line 29: for this type I would go for an exclusive type when you use the enums, for example

```ts
export type  SwapMappedResult  = {
	provider: SwapProvider.ICP_SWAP;
	...
} | {
	provider: SwapProvider.KONG_SWAP;
	...
} 
```

since you said that there are properties that are specific for each one, and that they should not be `undefined`

this will help you with reducing the checks, since the type will be defined directly by the field `provider`

roman-nazaruk: Fixed

src/frontend/src/lib/types/swap.ts
roman-nazaruk: Done

src/frontend/src/lib/utils/swap.utils.ts
roman-nazaruk: Agreed. Fixed

src/frontend/src/lib/utils/swap.utils.ts
```diff
@@ -51,3 +59,62 @@ export const getNetworkFee = ({
...
+ * This is typically used when performing swaps to ensure that
+ * the received amount doesn't fall below the expected slippage tolerance.
+ *
+ * @param params - Parameters to calculate slippage
+ * @param params.quoteAmount - The quoted amount without slippage (in smallest token units, e.g., wei)
+ * @param params.slippagePercentage - The allowed slippage percentage (e.g., 0.5 for 0.5%)
+ * @returns The minimum amount after slippage, as bigint
+ *
+ */
+export const calculateSlippage = ({
```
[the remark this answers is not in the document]
roman-nazaruk: It would be exported in the future, so I added tests for this utility method

AntonioVentilii (approved):
LGTM tks!

PR #6405, opened 2025-05-07
feat(frontend): added icpSwapAmount to obtain receiveAmount

roman-nazaruk:
# Motivation

Added icp swap amount service to obtain receiveAmount.

# Changes

Added new icpSwapAmount service.


src/frontend/src/lib/services/icp-swap.services.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { ICPSwapAmountReply, ICPSwapQuoteParams } from '$lib/types/api';
+import { isNullish } from '@dfinity/utils';
+import { get } from 'svelte/store';
+
+export const icpSwapAmounts = async ({
+	identity,
+	sourceToken,
+	destinationToken,
+	sourceAmount,
+	fee = 3000n // The only supported pool fee on ICPSwap at the moment (0.3%)
```
DenysKarmazynDFINITY, line 13: Could you please create a constant for this value? It can go to `swap.constants`

src/frontend/src/lib/services/icp-swap.services.ts
```diff
@@ -0,0 +1,35 @@
...
+}: ICPSwapQuoteParams): Promise<ICPSwapAmountReply> => {
+	const pool = await getPool({
+		identity,
+		token0: { address: sourceToken.ledgerCanisterId, standard: sourceToken.standard },
+		token1: { address: destinationToken.ledgerCanisterId, standard: destinationToken.standard },
+		fee
+	});
+
+	if (isNullish(pool)) {
+		throw new Error(get(i18n).swap.error.pool_not_found);
```
DenysKarmazynDFINITY, line 23: Question, mostly because I miss the context where this service will be used - do we need to throw an error here?  How about returning `undefined` or `null` and just displaying a message inside the swap modals based on the returned value? Asking because I remember for some other swap errors we didn't want to display errors in a toast.

roman-nazaruk: I just realized that getPool always returns a Promise<PoolData>, so there's no need to check it for null or undefined.

src/frontend/src/lib/services/icp-swap.services.ts
roman-nazaruk: Done

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6408, opened 2025-05-07
fix(frontend): missing chain ID in class `EtherscanProvider` for ERC20

AntonioVentilii:
# Motivation

There is a missing parameter chainId in method `erc20Transactions` of class `EtherscanProvider`. It was not raised since the method is still unused.


roman-nazaruk (approved):
LGTM

PR #6409, opened 2025-05-07
feat(frontend): replace `EtherscanRest` with `EtherscanProvider`

AntonioVentilii:
# Motivation

The method `erc20Transactions` of class `EtherscanProvider` is a replica of method `transactions` of class `EtherscanRest`.

So we can replace `EtherscanRest` with `EtherscanProvider` and deprecate everything that is related to Etherscan rest API.


DenysKarmazynDFINITY (approved):
Cool, thanks!

PR #6411, opened 2025-05-07
refactor(frontend): build Etherscan plugin inside the provider map

AntonioVentilii:
# Motivation

Just to avoid any possible interconnections, we create a new plugin for each network provider in Etherscan.


DenysKarmazynDFINITY (approved):
👍

PR #6413, opened 2025-05-07
refactor(frontend): refactored icp swap logic

roman-nazaruk:
# Motivation

Refactored the ICP Swap logic for consistency with other files and updated the test to properly mock the canister method.

# Changes

- Refactored ICP Swap logic
- Aligned structure with other modules
- Updated test to mock canister method correctly


AntonioVentilii (approved):
Thank you!

PR #6415, opened 2025-05-08
fix(e2e): transactions page waits for placeholder

AntonioVentilii:
# Motivation

The transactions page of the E2E tests is flaky since it does not wait for the transaction page to load.

To solve it, we make it wait for the transactions placeholder.

# Note

For Bitcoin we do not load the transaction during the E2E tests, so we can ignore this case.

# Changes

- Generalize the test ID for the transactions placeholder, since we are going to test it for all networks.
- Add prop to showTransactions method of E2E tests that forces to wait or not (true by default) for the placeholder.
- Use the new prop in the tests.
- Set Bitcoin network to ignore this rule.

# Tests

Tested on this PR successfully.


BonomoAlessandro (approved):
LGTM

PR #6416, opened 2025-05-08
feat(frontend): Make contact color computation reusable

FlavioInacta:
# Motivation

Move the color selection in a utils file in order to reuse the functionality for the contact header later.

# Changes

- Move computation into a separate file
- Enhancement: Make the hash computation functional

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/types/utils.ts
+++ b/src/frontend/src/lib/types/utils.ts
@@ -15,3 +15,5 @@ export type AtLeastOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<keyo
 	{ [K in Keys]-?: Required<Pick<T, K>> & Partial<Omit<T, K>> }[Keys];
 
 export type PartialSpecific<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
+
+export type NonEmptyArray<T> = [T, ...T[]];
--- a/src/frontend/src/lib/utils/contact.utils.ts
+++ b/src/frontend/src/lib/utils/contact.utils.ts
@@ -1,0 +1,25 @@
+import { isEmptyString } from '@dfinity/utils';
+
+export const selectColorForName = <T>({
+	colors,
+	name
+}: {
+	colors: T[];
+	name?: string;
+}): T | undefined => {
+	if (colors.length === 0) {
+		throw new Error('Colors array cannot be empty');
+	}
+
+	const trimmedName = name?.trim?.();
+	if (isEmptyString(trimmedName)) {
+		return undefined;
+	}
+
+	const hash = [...trimmedName].reduce<number>(
+		(acc, char) => (acc + char.charCodeAt(0)) % colors.length,
+		0
+	);
+
+	return colors[hash];
+};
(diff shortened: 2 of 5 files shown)


src/frontend/src/lib/utils/contact.utils.ts
```diff
@@ -0,0 +1,25 @@
+import { isEmptyString } from '@dfinity/utils';
+
+export const selectColorForName = <T>({
+	colors,
+	name
+}: {
+	colors: T[];
```
AntonioVentilii, line 7: to avoid that the length is zero we can even use the type here 

`colors: [T, ...T[]]`

src/frontend/src/lib/utils/contact.utils.ts
```diff
@@ -0,0 +1,25 @@
...
+	name?: string;
+}): T | undefined => {
+	if (colors.length === 0) {
+		throw new Error('Colors array cannot be empty');
```
AntonioVentilii, line 11: let's return undefined, better to avoid raising errors in utils otherwise you need to localize this

src/frontend/src/lib/utils/contact.utils.ts
```diff
@@ -0,0 +1,25 @@
+import { isEmptyString } from '@dfinity/utils';
+
+export const selectColorForName = <T>({
+	colors,
+	name
+}: {
+	colors: T[];
+	name?: string;
```
AntonioVentilii, line 8: shall this be undefined? i had imagined that it is always given, since the function is in a sense for this

AntonioVentilii (approved):
LGTM tks!

A few observations

src/frontend/src/lib/utils/contact.utils.ts
FlavioInacta: A nice trick. Didn't know that one, will use. :+1:

src/frontend/src/lib/utils/contact.utils.ts
FlavioInacta: Yes, but I want to allow undefined.

I'll change it to: `name: string | undefined;`

src/frontend/src/lib/utils/contact.utils.ts
FlavioInacta: Not needed anymore due to type `[T,...T[]]`

src/frontend/src/lib/utils/contact.utils.ts
AntonioVentilii: ah then, no issue, you can leave it as it is if it serves the purpose

src/frontend/src/lib/types/utils.ts
```diff
@@ -15,3 +15,5 @@ export type AtLeastOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<keyo
 	{ [K in Keys]-?: Required<Pick<T, K>> & Partial<Omit<T, K>> }[Keys];
 
 export type PartialSpecific<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
+
+export type NonEmptyArray<T> = [T, ...T[]];
```
AntonioVentilii, line 19: nice one! i will re-use it a lot eheh

PR #6418, opened 2025-05-08
feat(frontend): SendDestinationWizardStep component

DenysKarmazynDFINITY:
# Motivation

We need to have a component for the Send Destination step. Base on `sendToken`, it decides which components should be rendered.

Screenshot 2025-05-08 at 10 22 14

### DIFF
--- a/src/frontend/src/lib/components/send/SendDestinationWizardStep.svelte
+++ b/src/frontend/src/lib/components/send/SendDestinationWizardStep.svelte
@@ -14,12 +14,7 @@
 	import ButtonCancel from '$lib/components/ui/ButtonCancel.svelte';
 	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
 	import ContentWithToolbar from '$lib/components/ui/ContentWithToolbar.svelte';
-	import {
-		SEND_DESTINATION_WIZARD_STEP_BTC,
-		SEND_DESTINATION_WIZARD_STEP_ETH,
-		SEND_DESTINATION_WIZARD_STEP_IC,
-		SEND_DESTINATION_WIZARD_STEP_SOL
-	} from '$lib/constants/test-ids.constants';
+	import { SEND_DESTINATION_WIZARD_STEP } from '$lib/constants/test-ids.constants';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { SEND_CONTEXT_KEY, type SendContext } from '$lib/stores/send.store';
 	import { isNullishOrEmpty } from '$lib/utils/input.utils';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/send/SendDestinationWizardStep.svelte
```diff
@@ -0,0 +1,129 @@
...
+	const next = () => dispatch('icNext');
+	const close = () => dispatch('icClose');
+
+	let invalidDestination = $state(false);
+
+	let disabled = $derived(invalidDestination || isNullishOrEmpty(destination));
+</script>
+
+<ContentWithToolbar>
+	{#if isNetworkIdEthereum($sendToken?.network.id) || isNetworkIdEvm($sendToken.network.id)}
```
AntonioVentilii, line 56: i think you can use `sendTokenNetworkId` from he context?

src/frontend/src/lib/constants/test-ids.constants.ts
```diff
@@ -183,3 +183,8 @@ export const ADDRESS_BOOK_CONTACT_FORM = 'address-book-contact-form';
 export const ADDRESS_BOOK_CONTACT_NAME_INPUT = 'address-book-contact-name-input';
 export const ADDRESS_BOOK_SAVE_BUTTON = 'address-book-save-button';
 export const ADDRESS_BOOK_CANCEL_BUTTON = 'address-book-cancel-button';
+
+export const SEND_DESTINATION_WIZARD_STEP_ETH = 'send-destination-wizard-step-eth';
+export const SEND_DESTINATION_WIZARD_STEP_BTC = 'send-destination-wizard-step-btc';
+export const SEND_DESTINATION_WIZARD_STEP_SOL = 'send-destination-wizard-step-sol';
+export const SEND_DESTINATION_WIZARD_STEP_IC = 'send-destination-wizard-step-ic';
```
AntonioVentilii, line 190: shall these test IDs be prefix? so that we can expand to other networks if necessary. Does it make sense?

`${SEND_DESTINATION_WIZARD_STEP}-{network.id.description}`

src/frontend/src/lib/components/send/SendDestinationWizardStep.svelte
DenysKarmazynDFINITY: Absolutely! Updated.

src/frontend/src/lib/constants/test-ids.constants.ts
DenysKarmazynDFINITY: Great idea, done

AntonioVentilii (approved):
LGTM tks!

PR #6420, opened 2025-05-08
fix(frontend): Pending BTC -> ckBTC conversion label

daviddecentage:
# Motivation

The label for pending btc->ckBtc txs was wrong: "Receiving BTC" when it should say "Converting BTC to ckBTC" like for the other pending ck conversion txs.

# Changes

Changed translation label thats being used for these transactions and pass the CK token to the `replacePlaceholder`

# Tests

image


AntonioVentilii (approved):
Thanks for the fix!

PR #6422, opened 2025-05-08
refactor(frontend): remove unused return from `loadNextSolTransactions`

AntonioVentilii:
# Motivation

Service `loadNextSolTransactions` does not need to return a value.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6423, opened 2025-05-08
feat(bitcoin-types): Input validation

bitdivine:
# Motivation
Bitcoin types accepted by the backend can become large.

# Changes
- Verify that vectors and strings in bitcoin types have limited length.

# Tests
Some tests are included, more test vectors are to follow.

### DIFF
--- a/src/shared/src/types/bitcoin.rs
+++ b/src/shared/src/types/bitcoin.rs
@@ -4,6 +4,24 @@ use candid::CandidType;
 use ic_cdk::api::management_canister::bitcoin::{BitcoinNetwork, Utxo};
 use serde::Deserialize;
 
+/// The maximum length of a bitcoin address, expressed as a string.
+/// - The longest current format seems to be `Bech32` and `Bech32m` which are up to 62 characters
+///   long.
+/// - Some obsolete formats seem to be at most 160 characters long.
+pub const MAX_ADDRESS_LEN: usize = 62;
+
+/// The maximum length of a single `txid`:
+///
+/// - 32 bytes (represented as 64 hexadecimal characters) <https://learnmeabitcoin.com/technical/transaction/input/txid/>
+pub const MAX_TXID_LEN: usize = 32;
+/// The maximum number of UTXOs in the pending transaction:
+///
+/// - Theoretically, it seems that the maximum can be over 1000;
+/// - practically the limit seems to be about 200;
+/// - Typical transactions apparently take 1-3 UTXOs;
+/// - Consolidation transactions typically take many more, however that doesn't apply to this API.
+pub const MAX_UTXOS_LEN: usize = 128;
+
 #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
 pub struct SelectedUtxosFeeRequest {
     pub amount_satoshis: u64,
@@ -16,9 +34,6 @@ pub struct SelectedUtxosFeeRequest {
 pub struct SelectedUtxosFeeResponse {
     pub utxos: Vec<Utxo>,
     pub fee_satoshis: u64,
-}
-impl SelectedUtxosFeeResponse {
-    pub const MAX_UTXOS_LEN: usize = 32;
 }
 
 #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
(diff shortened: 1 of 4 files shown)


src/shared/src/types/bitcoin.rs
```diff
@@ -1,7 +1,27 @@
...
+/// The maximum length of a bitcoin address, expressed as a string.
+/// - The longest current format seems to be `Bech32` and `Bech32m` which are up to 62 characters
+///   long.
+/// - Some obsolete formats seem to be at most 160 characters long.
+pub const MAX_ADDRESS_LEN: usize = 62;
+
+/// The maximum length of a single `txid`:
+///
+/// - 32 bytes (represented as 64 hexadecimal characters) <https://learnmeabitcoin.com/technical/transaction/input/txid/>
+pub const MAX_TXID_LEN: usize = 32;
```
DenysKarmazynDFINITY, line 16: All the values seem to be correct! 

Minor observation - do you think it makes sense to rename this constant to `MAX_TXID_BYTES` or `MAX_TXID_BYTES_LEN`? The current naming IMO refers more to a hex string length (64) rather than bytes (32).

bitdivine: Good idea!

src/backend/tests/it/bitcoin.rs
```diff
@@ -72,7 +72,10 @@ fn test_select_user_utxos_fee_pending_transaction_error() {
         add_request,
     );
 
-    assert!(add_response.is_ok());
+    assert!(
+        add_response.is_ok(),
+        "btc_add_pending_transaction failed: {add_response:?}"
+    );
```
bitdivine, line 78: This failed and the error message was not helpful, hence the increased verbosity.

The test failed because the toy transaction ID is empty and the validator checked for exactly 32 bytes.

I have relaxed the validation to allow no more than 32 bytes, to limit how many changes are needed in this one PR, however it would be good to make the validation a bit stricter, eventually.  The backend is just a dumb store and the only thing it really needs to care about is being overwhelmed with data.

src/shared/src/types/bitcoin.rs
```diff
@@ -1,7 +1,27 @@
...
+///
+/// - 32 bytes (represented as 64 hexadecimal characters) <https://learnmeabitcoin.com/technical/transaction/input/txid/>
+pub const MAX_TXID_BYTES: usize = 32;
+/// The maximum number of UTXOs in the pending transaction:
+///
+/// - Theoretically, it seems that the maximum can be over 1000;
+/// - practically the limit seems to be about 200;
+/// - Typical transactions apparently take 1-3 UTXOs;
+/// - Consolidation transactions typically take many more, however that doesn't apply to this API.
+pub const MAX_UTXOS_LEN: usize = 128;
```
bitdivine, line 23: The above are the limits we will enforce for incoming Bitcoin types.

src/shared/src/types/bitcoin.rs
```diff
@@ -10,6 +30,7 @@ pub struct SelectedUtxosFeeRequest {
 }
 
 #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
+#[serde(remote = "Self")]
```
bitdivine, line 33: This allows us to have a custom serializer that calls the standard serilaizer, then the validation code.

src/shared/src/validate.rs
```diff
@@ -42,14 +42,15 @@ pub(crate) use validate_on_deserialize;
 /// - Run `test_validate_on_deserialize!($type)` to test the validation.
 #[cfg(test)]
 macro_rules! test_validate_on_deserialize {
```
bitdivine, line 44: I have tweaked this macro to take the test vectors from an argument instead of a function, as otherwise I would need a separate module for every test.

src/shared/src/validate.rs
```diff
@@ -72,7 +72,18 @@ macro_rules! test_validate_on_deserialize {
...
             }
         }
+        }
     };
 }
 #[cfg(test)]
 pub(crate) use test_validate_on_deserialize;
+
+#[cfg(test)]
+pub struct TestVector<T>
```
bitdivine, line 82: Another tweak: Use a standard test vector type rather than defining a dedicated test vector type for every test.

src/shared/src/types/bitcoin.rs
bitdivine: done

DenysKarmazynDFINITY (approved):
LGTM, thanks!

PR #6424, opened 2025-05-08
feat(devops): Add Rust unit test coverage tool

bitdivine:
# Motivation
Most of our backend tests are integration tests running on pocket-ic, which cannot be measured with tools such as tarpaulin.  However we do have unit tests and would like more.

# Changes
- Add a rust unit test coverage tool.

# Tests
FAIL: tarpaulin fails to compile the code.  Next step: Find out why.

Possible sources of inspiration:
- The rewards canister uses tarpaulin.  Perhaps adjusting the code structure to match, with entry points in main.rs rather than lib.rs, would help?


AntonioVentilii:
Experiment that didn't get past the initial tarpaulin compile failure; no progress since. Closing — reopen as a fresh PR if we want to revisit Rust unit-test coverage tooling.

PR #6430, opened 2025-05-08
fix(frontend): Collapsible bottom sheet breakpoint trigger

daviddecentage:
# Motivation

The bottom sheet is triggered too early for too big screens. The idea was to have the bottom sheet on mobile screens where we have not so much space. Also, if placed in a modal, it will now use the same breakpoint the modal is using to go to full width, which makes the two cases now consistent.

# Changes

Adjusted breakpoints to only show bottom sheet for small screens

# Tests

Before open triggered:
image
After open triggered:
image


roman-nazaruk (approved):
LGTM

PR #6432, opened 2025-05-08
feat(backend): Validate user profiles

bitdivine:
# Motivation
We would like to validate data submitted to the user profile store.

# Changes
- Add basic validation, checking that no fields are unreasonably large.

# Tests
Unit tests are included.


src/shared/src/types/user_profile.rs
```diff
@@ -8,7 +8,13 @@ use serde::Serialize;
 use super::{verifiable_credential::CredentialType, Timestamp};
```
bitdivine, line 8: @AntonioVentilii Would you be so kind as to check the length bounds proposed below?

src/shared/src/types/user_profile.rs
```diff
@@ -35,12 +46,22 @@ pub struct StoredUserProfile {
...
     pub current_user_version: Option<Version>,
 }
+impl AddUserCredentialRequest {
+    /// The maximum supported length for a credential JWT.
+    pub const MAX_CREDENTIAL_JWT_LENGTH: usize = 8 << 10;
+    /// The maximum number of `CredentialSpec.arguments`.
+    pub const MAX_CREDENTIAL_SPEC_ARGUMENTS: usize = 10;
+    // 8K
+    /// The maximum supported length for a credential type.
+    pub const MAX_CREDENTIAL_TYPE_LENGTH: usize = 16;
```
AntonioVentilii, line 63: maybe we should increase this one. If i read correctly the code, for example, we are using the string `ProofOfUniqueness` that has 17 characters already

src/shared/src/types/user_profile.rs
```diff
@@ -35,12 +46,22 @@ pub struct StoredUserProfile {
...
     pub credential_jwt: String,
     pub credential_spec: CredentialSpec,
     pub issuer_canister_id: Principal,
     pub current_user_version: Option<Version>,
 }
+impl AddUserCredentialRequest {
+    /// The maximum supported length for a credential JWT.
+    pub const MAX_CREDENTIAL_JWT_LENGTH: usize = 8 << 10;
+    /// The maximum number of `CredentialSpec.arguments`.
+    pub const MAX_CREDENTIAL_SPEC_ARGUMENTS: usize = 10;
```
AntonioVentilii, line 60: difficult to predict this one for new credentials type, but I want to tink that yes, 10 arguments are enough for any type of object

src/shared/src/types/user_profile.rs
```diff
@@ -35,12 +46,22 @@ pub struct StoredUserProfile {
...
+#[serde(remote = "Self")]
 pub struct AddUserCredentialRequest {
     pub credential_jwt: String,
     pub credential_spec: CredentialSpec,
     pub issuer_canister_id: Principal,
     pub current_user_version: Option<Version>,
 }
+impl AddUserCredentialRequest {
+    /// The maximum supported length for a credential JWT.
+    pub const MAX_CREDENTIAL_JWT_LENGTH: usize = 8 << 10;
```
AntonioVentilii, line 58: correct me if i am wrong, but this is 8192 bytes (e.g. characters), right?

If so, in the tests [https://github.com/dfinity/oisy-wallet/blob/2a61d97691c26118380803c2cf513d305a783c[…]src/tests/lib/services/request-pouh-credential.services.spec.ts](https://github.com/dfinity/oisy-wallet/blob/2a61d97691c26118380803c2cf513d305a783cdd/src/frontend/src/tests/lib/services/request-pouh-credential.services.spec.ts#L17) we are using something like 8.7 mil characters; not really sure if that is always the case, but maybe better to increase this one too?

src/shared/src/types/user_profile.rs
bitdivine: I missed ArgumentValue string types.  Any preference for how long those may be?

AntonioVentilii: good question... current OISY implementation does not even pass arguments (POUH credential type does not require it), so I have no practical usage for it

To be on safe side, we could give `2048` bytes?

src/shared/src/types/user_profile.rs
bitdivine: Isn't a JWT passed as an HTTP header, and so bounded by the maximum length of an HTTP header?

bitdivine: This is based on reading the spec: https://datatracker.ietf.org/doc/html/rfc7519

But HTTP headers and query strings are just suggested uses for JWTs. Perhaps they could be used in other contexts that allow longer strings?

bitdivine: I guess another question is:  Do we store these strings?  If we just process them, the size doesn't mater much.

AntonioVentilii: I had a look around, and it seems we do not store it, but just use it for validate, for example in method `add_user_credential`

AntonioVentilii: I cannot find any indication that we store `credential_jwt`

bitdivine: Thank you for checking.  I have deleted the bound; I think there is probably not much sense in having a bound that we don't need.

AntonioVentilii (approved):
LGTM! 

Not sure if you are waiting for other validations/reviews, but on my side approved, tks!

PR #6435, opened 2025-05-08
feat(frontend): update SendModal component

DenysKarmazynDFINITY:
# Motivation

The last step is to add SendDestinationWizard step to the Modal and update all related events. Additionally, we fetch ETH transactions in case selected token standard is ETH in order to generate known destinations derived.


AntonioVentilii (approved):
LGTM tks!

In another PR, shall we add tests for this component?

PR #6440, opened 2025-05-08
feat(frontend): extract checks from `IcTransactionsScroll` component

AntonioVentilii:
# Motivation

In the coming PR https://github.com/dfinity/oisy-wallet/pull/6438, we want to re-use the logic of `IcTransactionsScroll` to loads the next transactions (that are missing) in the Activity page.

To avoid repetition of code, we extract the checks done in the component to the service `loadNextIcTransactions`.

# Changes

- Rename service loadNextIcTransactions to loadNextIcTransactionsRequest.
- Create new function service loadNextIcTransactions that is a wrapper of the old one plus the checks from `IcTransactionsScroll` component.
- Remove the checks from `IcTransactionsScroll`.
- NOTE: we leave the nullish check for lastId in the component, because we will not need them in the Activity page.

# Tests

Created new tests both for the component and the service. To mock the `InfiniteScroll` component to dispatch the intersection event, I re-used the code developed in [GIX Component tests](https://github.com/dfinity/gix-components/blob/310688d90d3b3692d1f08e2f6323da0a73635302/src/tests/lib/components/InfiniteScroll.spec.ts#L11).

### DIFF
--- a/src/frontend/src/icp/services/ic-transactions.services.ts
+++ b/src/frontend/src/icp/services/ic-transactions.services.ts
@@ -2,18 +2,19 @@ import { getTransactions as getTransactionsIcp } from '$icp/api/icp-index.api';
 import { getTransactions as getTransactionsIcrc } from '$icp/api/icrc-index-ng.api';
 import { icTransactionsStore } from '$icp/stores/ic-transactions.store';
 import type { IcCanistersStrict, IcToken } from '$icp/types/ic-token';
-import type { IcTransaction } from '$icp/types/ic-transaction';
+import type { IcTransaction, IcTransactionUi } from '$icp/types/ic-transaction';
 import { mapIcTransaction } from '$icp/utils/ic-transactions.utils';
 import { mapTransactionIcpToSelf } from '$icp/utils/icp-transactions.utils';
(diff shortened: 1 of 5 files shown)


src/frontend/src/icp/services/ic-transactions.services.ts
```diff
@@ -121,3 +122,43 @@ export const onTransactionsCleanUp = (data: { tokenId: TokenId; transactionIds:
...
+	lastId: IcTransactionUi['id'] | undefined;
+	owner: Principal;
+	identity: OptionIdentity;
+	maxResults?: bigint;
+	token: Token;
+	signalEnd: () => void;
+}): Promise<void> => {
+	try {
+		if (nonNullish(lastId)) {
+			BigInt(lastId.replace('-self', ''));
```
DenysKarmazynDFINITY, line 140: How about we assign this value to a var and re-use it below in the function? Just so we don't this operation twice.

DenysKarmazynDFINITY (approved):
LGTM - just a small suggestion

src/frontend/src/icp/services/ic-transactions.services.ts
AntonioVentilii: Done!

PR #6441, opened 2025-05-08
refactor(backend)!: create result type for `top_up_cycles_ledger` method

AntonioVentilii:
# Motivation

In an effort to make the did file more readable, we create a specific type for the results of method `top_up_cycles_ledger`.

BREAKING CHANGE: the interface will be changed.


bitdivine (approved):
LGTM.  Thanks!

PR #6444, opened 2025-05-09
feat(e2e): run MacOS E2E tests only once per day

AntonioVentilii:
# Motivation

We add a schedule to the E2E CI workflow to run once a day. That is the instance in which we will run the MacOS tests too (they are 10x more expensive than the linux ones).

Furthermore, we leave the trigger for workflow dispatch and for the label.


DenysKarmazynDFINITY (approved):
👍

PR #6451, opened 2025-05-09
test(frontend): test suite for ICP-ETH token utils

AntonioVentilii:
# Motivation

Created some tests for the ICP-ETH token utils.


DenysKarmazynDFINITY (approved):
Thanks!

PR #6453, opened 2025-05-09
feat(e2e): adapt send-and-receive test to new send flow

AntonioVentilii:
# Motivation

Changes to adapt the E2E tests to the new send flow: now there is a first step that is the destination, and the next one is the amount.


DenysKarmazynDFINITY (approved):
Thank you! Approving so you can merge it whenever the pipeline allows you to do so 😄

PR #6457, opened 2025-05-09
feat(frontend): update appearance of ETH fee in the Send form

DenysKarmazynDFINITY:
# Motivation

We need to update the way we display ETH fee in the Send form.

Screenshot 2025-05-09 at 11 56 06


src/frontend/src/tests/eth/components/send/EthSendForm.spec.ts
```diff
@@ -51,9 +50,8 @@ describe('EthSendForm', () => {
 
 		expect(getByTestId(SEND_DESTINATION_SECTION)).toBeInTheDocument();
 
-		const maxFeeEth: HTMLDivElement | null = container.querySelector(maxFeeEthSelector);
-
-		expect(maxFeeEth).not.toBeNull();
```
AntonioVentilii, line 56: shouldn't this test still work?

DenysKarmazynDFINITY: it does work, I just updated the way we check it - the new component doesn't have the old id we used to query this element

PR #6461, opened 2025-05-09
feat(frontend): default standard in ICRC util `mapIcrcToken`

AntonioVentilii:
# Motivation

To be more correct, the util `mapIcrcToken` shall use the `standard` provided by the existing list, instead of fixing it to `icrc`.

# Changes

- Change the default condition for `standard` in the mapper.

# Tests

Since we are here, I added tests for all ICRC utils.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6463, opened 2025-05-09
feat(frontend): update appearance of IC fees in the Send form

DenysKarmazynDFINITY:
# Motivation

We need to update the way we display IC fees in the Send form.

Screenshot 2025-05-09 at 15 14 43


AntonioVentilii (approved):
LGTM tks!

PR #6465, opened 2025-05-09
feat(devops): Reduce coverage thresholds by 0.50%

AntonioVentilii:
# Motivation

The checks for tests coverage are a bit flaky, since the calculation can change for a few basis points: we round the thresholds to the lowest integers; so, if we are close to it, the checks keep failing.

Example: threshold is set at 84%, real coverage is at 84.01%, but the flaky calculation returns sometimes 83.98%.

To avoid this, we change the workflow that changes the thresholds to reduce them by 0.50% instead of flooring them. That should be enough margin for flakiness in the calculation.

### DIFF
--- a/.github/workflows/frontend-update-coverage-thresholds.yml
+++ b/.github/workflows/frontend-update-coverage-thresholds.yml
@@ -38,9 +38,18 @@ jobs:
       - name: Test
         run: npm run test:coverage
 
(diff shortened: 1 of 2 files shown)


DenysKarmazynDFINITY (approved):
Thank you!!!

PR #6468, opened 2025-05-09
feat(frontend): worker service for DIP20 tokens

AntonioVentilii:
# Motivation

To integrate DIP-20 tokens, like XTC, we need a worker to synchronize the wallet. So we basically replicate the one that exists for ICP.


src/frontend/src/icp/services/worker.dip20-wallet.services.ts
```diff
@@ -0,0 +1,83 @@
...
+				syncWallet({
+					tokenId,
+					data: data.data as PostMessageDataResponseWallet
+				});
+				return;
+			case 'syncDip20WalletError':
+				onLoadTransactionsError({
+					tokenId,
+					error: (data.data as PostMessageDataResponseError).error,
+					silent: true
```
DenysKarmazynDFINITY, line 45: Just to make sure - it is wanted to always hide the toast?

DenysKarmazynDFINITY (approved):
Small q, otherwise LGTM!

PR #6471, opened 2025-05-09
feat(frontend): API methods for XTC ledger

AntonioVentilii:
# Motivation

We create the API methods for the XTC ledger using class `XtcLedgerCanister`. This is a replica of what we do with other canisters.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6472, opened 2025-05-09
feat(frontend): add DIP-20 response messages to IC wallet schedulers

AntonioVentilii:
# Motivation

Preparing to create a specific worker for DIP-20 tokens, we first set the appropriate response messages in the classes of IC wallet schedulers. The new worker will extend such classes too, similar to what ICP and ICRC tokens do.


src/frontend/src/icp/types/api.ts
```diff
@@ -0,0 +1,6 @@
+import type { Event } from '$declarations/xtc_ledger/xtc_ledger.did';
+
+export interface TransactionWithId {
```
DenysKarmazynDFINITY, line 3: Shall we make the naming here a bit more specific? E.g. `Dip20TransactionWithId`. Since `Event` of `xtc_ledger.did` is used, I doubt it will be reused for something else.

AntonioVentilii: Makes sense! changed

PR #6474, opened 2025-05-09
fix(frontend): service `loadNextSolTransactions` assigns to correct token ID

AntonioVentilii:
# Motivation

Service `loadNextSolTransactions` was not correctly assigning the next transactions to the respective token. It was mapping all of them to the network-specific native token.

# Changes

- Add token as input to `loadNextSolTransactions`.
- Use it instead of mapping to native Solana tokens.

# Tests

Added tests to verify that it works correctly (same tests would have failed before).

### DIFF
--- a/src/frontend/src/sol/services/sol-transactions.services.ts
+++ b/src/frontend/src/sol/services/sol-transactions.services.ts
@@ -1,10 +1,4 @@
 import { WSOL_TOKEN } from '$env/tokens/tokens-spl/tokens.wsol.env';
-import {
-	SOLANA_DEVNET_TOKEN_ID,
-	SOLANA_LOCAL_TOKEN_ID,
-	SOLANA_TESTNET_TOKEN_ID,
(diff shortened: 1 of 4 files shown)


src/frontend/src/sol/services/sol-transactions.services.ts
```diff
@@ -210,22 +204,15 @@ export const loadNextSolTransactions = async ({
 
 	if (transactions.length === 0) {
 		signalEnd();
+		return;
```
DenysKarmazynDFINITY, line 207: It seems we don't need it

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/sol/services/sol-transactions.services.ts
AntonioVentilii: Right! removed!

PR #6475, opened 2025-05-09
test(frontend): tests for service `loadNextIcTransactions`

AntonioVentilii:
# Motivation

Adding tests for service `loadNextIcTransactions`. Note: correct ID for an IC transaction is an integer in string form.


DenysKarmazynDFINITY (approved):
Thanks!

PR #6476, opened 2025-05-09
fix(build): avoid non-SVG logos when building additional ICRC

AntonioVentilii:
# Motivation

The script to build additional tokens was saving the logos as SVG. However, some of them are in PNG and JPEG formats.

As first step, we introduce a warning to avoid saving wrong formats, since we prefer to have the logos as SVG files.

After this, in another PR, we will try and save them with whatever format they come in, or convert them to SVG automatically.

# Changes

- Include two checks in script util `saveLogo`: one for the correct type definition and one for the correct format of the image.
- Adapt both scripts `build:tokens-icrc` and `build:tokens-ckerc20`.
- Unrelated: unnecessary whitespace at the end of script `build:tokens-ckerc20`.

# Tests

Run the script on this PR successfully and as expected.


DenysKarmazynDFINITY (approved):
Nice, thanks

PR #6480, opened 2025-05-09
test(frontend): additional test for `IcTransactionsScroll`

AntonioVentilii:
# Motivation

For component `IcTransactionsScroll`, we should test if the transactions are not loaded anymore once they are finished.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6481, opened 2025-05-09
test(frontend): flaky timestamp for test of `icKnownDestinations`

AntonioVentilii:
# Motivation

When testing derived store `icKnownDestinations`, the timestamp created for the mock transactions can be higher than the previous one. So, to avoid flakiness (see [failed CI workflow](https://github.com/dfinity/oisy-wallet/actions/runs/14937269291/job/41967527170?pr=6476)), we need to confirm that the store takes the most recent timestamp.


DenysKarmazynDFINITY (approved):
Nice catch, thanks for the fix!

PR #6482, opened 2025-05-10
fix(rust)!: Delete unused list user endpoints

bitdivine:
BREAKING CHANGE: The list users APIs have been deleted.

# Motivation
The list user endpoints are not needed.

# Changes
- Delete the list user endpoints.

# Tests
Existing tests should verify that retained functionality still works.


AntonioVentilii (approved):
Thank you!

PR #6483, opened 2025-05-10
feat(backend): Check UserToken field lengths

bitdivine:
# Motivation
We would like stricter validation of backend input types.

# Changes
- Check `UserToken` field lengths.

# Tests
### Types with fields that are too long or wrong fail to parse
- Unit tests are included that check for this.


src/backend/src/assertions.rs
```diff
@@ -1,16 +1,4 @@
-use shared::types::{token::UserToken, MAX_SYMBOL_LENGTH};
-
-pub fn assert_token_symbol_length(token: &UserToken) -> Result<(), String> {
```
bitdivine, line 3: This is now checked when parsing.

src/backend/src/lib.rs
```diff
@@ -267,7 +267,6 @@ fn parse_eth_address(address: &str) -> [u8; 20] {
 #[update(guard = "caller_is_not_anonymous")]
 #[allow(clippy::needless_pass_by_value)]
 pub fn set_user_token(token: UserToken) {
-    assert_token_symbol_length(&token).unwrap_or_else(|e| ic_cdk::trap(&e));
     assert_token_enabled_is_some(&token).unwrap_or_else(|e| ic_cdk::trap(&e));
```
bitdivine, line 270: Should this be moved into the validator as well?  Are there existing, stored, user tokens where this is not set?

src/shared/src/impls.rs
```diff
@@ -394,8 +394,12 @@ impl Validate for SplToken {
     fn validate(&self) -> Result<(), candid::Error> {
         use crate::types::MAX_SYMBOL_LENGTH;
         if let Some(symbol) = &self.symbol {
-            if symbol.len() > MAX_SYMBOL_LENGTH {
-                return Err(candid::Error::msg("Symbol too long"));
+            if symbol.chars().count() > MAX_SYMBOL_LENGTH {
```
bitdivine, line 397: Note: `.len()` counts bytes in a  string, so the "ten snowmen" example in one of our tests has .len() 30.

src/backend/src/lib.rs
AntonioVentilii: it should not happen, and it is not an issue for the FE if it returned `null`, BUT it is a good double-check, to avoid saving "pointless" information

src/backend/src/lib.rs
```diff
@@ -287,8 +286,6 @@ pub fn set_many_user_tokens(tokens: Vec<UserToken>) {
 
     mutate_state(|s| {
         for token in tokens {
-            assert_token_symbol_length(&token).unwrap_or_else(|e| ic_cdk::trap(&e));
-            assert_token_enabled_is_some(&token).unwrap_or_else(|e| ic_cdk::trap(&e));
```
AntonioVentilii, line 291: should this `assert_token_enabled_is_some` stay? for the same reason of https://github.com/dfinity/oisy-wallet/pull/6483/files#r2082849629 and we can tackle it later?

AntonioVentilii, line 291: should this `assert_token_enabled_is_some` stay? for the same reason of https://github.com/dfinity/oisy-wallet/pull/6483/files#r2082849629 and we can tackle it later?

AntonioVentilii (approved):
LGTM tks!

A small observation

src/backend/src/lib.rs
bitdivine: Good catch!  Reinstated.

PR #6484, opened 2025-05-10
chore(devops): Bump dfx

bitdivine:
# Motivation
A newer version of dfx is available, with a more recent pocket-ic.

# Changes
- Update `dfx`
- Update the Rust `pocket-ic` client to match the `pocket-ic` server version bundled with `dfx`.

# Tests
This will need extensive testing in CI, testnets and with Orbit.


AntonioVentilii (approved):
Amazing work! Thank you!

PR #6485, opened 2025-05-10
feat(devops): remove script folder from vitest coverage analysis

AntonioVentilii:
# Motivation

We don't have unittests for the scripts: run them in the workflows, or manually. Plus some of them are tested in the workflows too.

So, I think it makes sense to remove them from vitest coverage analysis.


DenysKarmazynDFINITY (approved):
Makes sense!

PR #6487, opened 2025-05-10
test(frontend): add simple tests for message switch in simple workers

AntonioVentilii:
# Motivation

We create simple tests to verify the correct scheduler calls for the workers that switch the messages that are posted.


DenysKarmazynDFINITY (approved):
Great job! Thanks!

PR #6492, opened 2025-05-10
chore(frontend): remove unnecessary check in util `isIcrcTokenToggleDisabled`

AntonioVentilii:
# Motivation

Since the input of util `isIcrcTokenToggleDisabled` is always a defined token, we don't need to check for nullish values.


DenysKarmazynDFINITY (approved):
👍

PR #6496, opened 2025-05-11
refactor(frontend): include Alchemy network name and URL in the `EthereumNetwork` type

AntonioVentilii:
# Motivation

Similar to what we did for Infura in PR https://github.com/dfinity/oisy-wallet/pull/6281, for better manageability, it is better that when we add a new Ethereum/EVM network we already include all the providers data. So that any consumer is already set for a new network. To do that, we include the name of each network for Alchemy provider, and their respective RPC URL.

# Changes

- Add prop `alchemy` and `alchemyJsonRpcUrl` to `EthereumNetwork` type.
- Set the correct data for all Ethereum/EVM networks.
- Adapt all Alchemy providers map to use directly the list of network with their new prop.

# Tests

Adapted current tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6501, opened 2025-05-11
feat(e2e): shard E2E tests

AntonioVentilii:
# Motivation

To try and improve the E2E tests duration, we can shard them in 10 parts, using playwright [documentation](https://playwright.dev/docs/test-sharding#github-actions-example).

# IMPORTANT Note
There is a counter-argument to this approach: the total sum of the duration of the CI will increase, since we run the preparation jobs for each parallel test file (it takes 5 minutes more or lesse).
However, in a future PR, we will try to cache such preparation data and reduce the total amount of time.

# Changes

- Add sharding parameters to the E2E test workflow.
- Adapt concurrency to avoid race conditions.
- Remove unnecessary steps.

# Tests

The workflows ran on this PR successfully. I even tested to remove all the snapshots and they were recreated.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6514, opened 2025-05-12
feat(frontend): adjusted modal value and collapsible bottom sheet

roman-nazaruk:
# Motivation

We need to integrate a new UI component and align existing components with the design specifications for consistency

# Changes

- Updated styles for `ModalValue` and `CollapsibleBottomSheet` components


src/frontend/src/lib/components/ui/CollapsibleBottomSheet.svelte
```diff
@@ -3,25 +3,41 @@
...
 
 	let expanded = $state(false);
 </script>
 
 <Responsive down="sm">
+	<div class="flex w-full items-center justify-between">
+		{@render contentHeader({ isInBottomSheet: false })}
+		<ButtonIcon
+			on:click={() => (expanded = true)}
+			ariaLabel="expand"
```
AntonioVentilii, line 30: this should be localized in i18n, and maybe a bit more verbose

PR #6517, opened 2025-05-12
feat(frontend): removes the rewards navigation menu item tag

BonomoAlessandro:
# Motivation

The `New` tag for the `rewards` navigation menu item is not needed anymore.

# Changes

- removes tag

# Tests

**before:**

**after:**
image


AntonioVentilii (commented):
May you please remove it from `NavigationItem` too?

AntonioVentilii (approved):
Good point! I thought we would remove it forever, eheh!

LGTM tks!

PR #6531, opened 2025-05-12
feat(frontend): worker for IC DIP-20 tokens

AntonioVentilii:
# Motivation

We need to create a wallet worker for DIP-20 tokens. Thankfully, the current implementation of class `IcWalletBalanceAndTransactionsScheduler` helps us, replicating for `dip20` what we do for standard `icp` and `icrc`.

# Note

It was decided that we can ignore the transactions implementation for now, but we will work on it in another PR. That means that everything that is related to transactions will temporarily return an empty array, to avoid any possible error handler.

# Changes

- Create DIP-20 Wallet worker, replicating what we do for ICP and ICRC.
- Add new events to list of post messages.
- Create "empty" mapper functions for transactions, since we will always return empty transactions (NOT `null`).
- Include new worker in the generic worker list.

# Tests

Added more tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6533, opened 2025-05-12
feat(frontend): introduce DIP-20 standard

AntonioVentilii:
# Motivation

We want to start supporting [IC DIP-20 tokens](https://github.com/Psychedelic/DIP20), specifically [XTC](https://github.com/Psychedelic/dank/tree/main/xtc). In this PR, we introduce the new standard.

# Changes

- Add `dip20` among the token standard.
- Create util to check the correct new standard.
- Use the util in the receive modal.

# Tests

Add tests.

### DIFF
--- a/src/frontend/src/icp/components/receive/IcReceive.svelte
+++ b/src/frontend/src/icp/components/receive/IcReceive.svelte
@@ -15,7 +15,7 @@
 		isTokenCkErc20Ledger,
 		isTokenCkEthLedger
 	} from '$icp/utils/ic-send.utils';
-	import { isTokenIcrc } from '$icp/utils/icrc.utils';
+	import { isTokenDip20, isTokenIcrc } from '$icp/utils/icrc.utils';
 	import { modalStore } from '$lib/stores/modal.store';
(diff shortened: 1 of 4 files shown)


src/frontend/src/icp/components/receive/IcReceive.svelte
```diff
@@ -52,7 +55,7 @@
 	<IcReceiveCkEthereum />
 {:else if ckBTC}
 	<IcReceiveCkBTC />
-{:else if icrc}
+{:else if icrc || dip20}
```
DenysKarmazynDFINITY, line 58: I assume all other similar places (like send) will be adjusted in the follow-up PRs, right?

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/icp/components/receive/IcReceive.svelte
AntonioVentilii: correct! i will do it by "usage", this was just an example to use it properly

PR #6546, opened 2025-05-12
feat(frontend): added new swapProviderListItem component

roman-nazaruk:
# Motivation
Created a new component to clearly for Swap Provider list item

# Changes
- Added a new component to clearly for Swap Provider list item

# Tests


src/frontend/src/lib/components/swap/SwapProviderListItem.svelte
```diff
@@ -0,0 +1,78 @@
...
+				schema: UrlSchema,
+				value: dapp.website
+			});
+			if (nonNullish(validated)) {
+				const url = new URL(validated);
+				displayURL = url.hostname.startsWith('www.') ? url.hostname.slice(4) : url.hostname;
+			}
+		} catch {
+			displayURL = null;
+		}
```
AntonioVentilii, line 45: instead of try-catch, you can use directly the safe parse, right? since you don't want errors raised

```suggestion
	const parsed = UrlSchema.safeParse(dapp.website);
	
	if (parsed.success) {
	        const url = new URL(parsed.data);
		displayURL = url.hostname.startsWith('www.') ? url.hostname.slice(4) : url.hostname;
		return
	}
        
        displayURL = null
```

src/frontend/src/lib/components/swap/SwapProviderListItem.svelte
```diff
@@ -0,0 +1,78 @@
...
+</script>
+
+// TODO: Migrate to Svelte 5, remove legacy slot usage and use render composition instead
+{#if nonNullish(dapp)}
+	<LogoButton on:click dividers>
+		<span slot="title">{dapp.name}</span>
+
+		<span slot="description">
+			{#if nonNullish(displayURL)}
+				<span class="text-sm text-tertiary">{displayURL}</span>
```
DenysKarmazynDFINITY, line 56: I think this line can be simplified to just `{displayUrl}` - both `text-sm` and `text-tertiary` are already applied to the description slot inside `LogoButton`.

src/frontend/src/lib/components/swap/SwapProviderListItem.svelte
```diff
@@ -0,0 +1,78 @@
...
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+	import { safeParse } from '$lib/validation/utils.validation';
+
+	interface Props {
+		amount: bigint;
+		token: IcTokenToggleable;
+		logoSize?: LogoSize;
+		usdBalance: OptionAmount;
+		dapp?: OisyDappDescription;
+		isBest: boolean;
```
DenysKarmazynDFINITY, line 22: Can we make it a bit more specific? E.g. `isBestRate`

src/frontend/src/lib/components/swap/SwapProviderListItem.svelte
```diff
@@ -0,0 +1,78 @@
...
+	import { i18n } from '$lib/stores/i18n.store';
+	import type { LogoSize } from '$lib/types/components';
+	import type { OisyDappDescription } from '$lib/types/dapp-description';
+	import type { OptionAmount } from '$lib/types/send';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+	import { safeParse } from '$lib/validation/utils.validation';
+
+	interface Props {
+		amount: bigint;
+		token: IcTokenToggleable;
```
DenysKarmazynDFINITY, line 18: Just for my understanding - is there a reason why you're using `IcTokenToggleable` over `IcToken`?

src/frontend/src/lib/components/swap/SwapProviderListItem.svelte
```diff
@@ -0,0 +1,78 @@
...
+				const url = new URL(validated);
+				displayURL = url.hostname.startsWith('www.') ? url.hostname.slice(4) : url.hostname;
+			}
+		} catch {
+			displayURL = null;
+		}
+	});
+</script>
+
+// TODO: Migrate to Svelte 5, remove legacy slot usage and use render composition instead
```
DenysKarmazynDFINITY, line 49: Does this really work? I mean leaving a `//` comment outside of `script`? In general, I don't mind having such a TODO, but it's not absolutely necessary since when the `LogoButton` gets migrated to Svelte5, linter won't allow not to update all the consumers, therefore we don't have a change to miss this one.

src/frontend/src/tests/lib/components/swap/SwapProviderListItem.spec.ts
```diff
@@ -0,0 +1,103 @@
+import type { IcTokenToggleable } from '$icp/types/ic-token-toggleable';
+import SwapProviderListItem from '$lib/components/swap/SwapProviderListItem.svelte';
+import { render } from '@testing-library/svelte';
+import { tick } from 'svelte';
+
+describe('SwapProviderListItem', () => {
+	const tokenMock = {
```
DenysKarmazynDFINITY, line 7: There is a pre-defined `mockValidIcToken` available in the code, in case you wanna use it

DenysKarmazynDFINITY (commented):
Could you please also attach a screenshot to the PR description just to get a feeling of how the component looks like?

src/frontend/src/lib/components/swap/SwapProviderListItem.svelte
roman-nazaruk: Agreed

src/frontend/src/lib/components/swap/SwapProviderListItem.svelte
roman-nazaruk: I use `IcTokenToggleable `here because this property is destinationToken from SwapStore, where IcTokenToggleable is the defined type for it.

src/frontend/src/lib/components/swap/SwapProviderListItem.svelte
roman-nazaruk: I just left this comment as to do, to know that we should migrate this code to Svelte 5

src/frontend/src/tests/lib/components/swap/SwapProviderListItem.spec.ts
roman-nazaruk: Thank you. Will reuse `mockValidIcToken `

src/frontend/src/lib/components/swap/SwapProviderListItem.svelte
roman-nazaruk: Nice catch

src/frontend/src/lib/components/swap/SwapProviderListItem.svelte
roman-nazaruk: Nica catch. Thank you

AntonioVentilii (approved):
LGTM, but please wait for @DenysKarmazynDFINITY review

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6567, opened 2025-05-13
refactor(frontend): use store `networkBsc` for fallback token of BSC

AntonioVentilii:
# Motivation

To be more precise, we need to query the BSC network derived store, to define the BSC default fallback token.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6578, opened 2025-05-13
fix(frontend): Don't allow to add empty contact names

FlavioInacta:
# Motivation

It should not be allowed to add empty contact names.

# Changes

- Make form invalid if name contains only empty characters
- Don't render name in title if it contains only empty characters


AntonioVentilii (approved):
Good catch!

FlavioInacta:
Moin found it. ;)

PR #6580, opened 2025-05-13
fix(devops): Fix dfx settings for staging

bitdivine:
# Motivation
The updated dfx does not deploy to staging.  There is one env var that is now required to use a plaintext key, and one of the flags we use is now deprecated.

# Changes
- Update the dfx configuration.

# Tests
Deployment to fe5: https://github.com/dfinity/oisy-wallet/actions/runs/14998375156


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6586, opened 2025-05-13
feat(frontend): update KnownDestinations stores

DenysKarmazynDFINITY:
# Motivation

We need to update the KnownDestinations stores according to the new specs. ]

Note: there are a few differences in txs stores that do not allow us to easily re-use the code inside `knownDestinations` derived.

### DIFF
--- a/src/frontend/src/btc/derived/btc-transactions.derived.ts
+++ b/src/frontend/src/btc/derived/btc-transactions.derived.ts
@@ -1,6 +1,9 @@
 import { btcTransactionsStore, type BtcTransactionsData } from '$btc/stores/btc-transactions.store';
 import { sortBtcTransactions } from '$btc/utils/btc-transactions.utils';
 import { tokenWithFallback } from '$lib/derived/token.derived';
+import { tokens } from '$lib/derived/tokens.derived';
+import type { TokenId } from '$lib/types/token';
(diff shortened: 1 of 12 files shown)


src/frontend/src/btc/derived/btc-transactions.derived.ts
```diff
@@ -25,6 +28,22 @@ export const btcTransactionsNotInitialized: Readable<boolean> = derived(
...
+
+			if (nonNullish(token)) {
+				($btcTransactionsStore?.[tokenId as TokenId] ?? []).forEach(({ data }) => {
+					mappedTransactions.push({
+						...data,
+						token
+					});
+				});
+			}
+		});
```
AntonioVentilii, line 45: maybe this can be redone functional? not for this PR

The similar ones too, maybe we can create a small util

AntonioVentilii (approved):
LGTM tks! One observation but non-blocker

src/frontend/src/btc/derived/btc-transactions.derived.ts
DenysKarmazynDFINITY: Yes, that was my initial plan. I think the first step would be to create a derived with EthTransactionsUi, and then create a small util for all get-known-destinations tores.

PR #6591, opened 2025-05-14
feat(frontend): custom background for VEUR token

DenysKarmazynDFINITY:
# Motivation

We need to display a custom Hero background for the VEUR token.


src/frontend/src/icp-eth/utils/token.utils.ts
```diff
@@ -4,3 +4,5 @@ import type { Token } from '$lib/types/token';
 export const isGLDTToken = (token: Token): boolean => isTokenIcrc(token) && token.symbol === 'GLDT';
 
 export const isVCHFToken = (token: Token): boolean => isTokenIcrc(token) && token.symbol === 'VCHF';
+
+export const isVEURToken = (token: Token): boolean => isTokenIcrc(token) && token.symbol === 'VEUR';
```
AntonioVentilii, line 8: i created some very small tests for this types of utils, just to keep the coverage up, may you please add the same before merging?

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/icp-eth/utils/token.utils.ts
DenysKarmazynDFINITY: sure, good point!

PR #6600, opened 2025-05-14
feat(frontend): removes defined sprinkles carousel card

BonomoAlessandro:
# Motivation

The `Season 3 Episode 1 sprinkles` are no longer new, so we don't want to show them in the carousel anymore. They will be added back once `Season 4` is released.

# Changes

- removes `sprinkles` carousel card.

### DIFF
--- a/src/frontend/src/env/reward-campaigns.env.ts
+++ b/src/frontend/src/env/reward-campaigns.env.ts
@@ -6,4 +6,4 @@ import * as z from 'zod';
 const parseResult = z.array(RewardEventsSchema).safeParse(rewardCampaignsJson);
 export const rewardCampaigns: RewardDescription[] = parseResult.success ? parseResult.data : [];
 
-export const FEATURED_REWARD_CAROUSEL_SLIDE_ID = 'OISY Airdrop #1';
+export const FEATURED_REWARD_CAROUSEL_SLIDE_ID = undefined;


AntonioVentilii (commented):
is this PR required? are you not updating it in https://github.com/dfinity/oisy-wallet/pull/6603 ?

BonomoAlessandro:
@AntonioVentilii Yes this one is needed. In this PR we want to remove the carousel card for the `sprinkles` and in the other PR we just change the reward id.

AntonioVentilii:
@BonomoAlessandro my point is: these changes override each other. Are they going to be in different releases? If not, it does really make sense to have this one.

I understand that you want to split the scopes. I will approve this one, but have a look if really needed (if the other one merges first, this one is not needed for example)

AntonioVentilii (approved):
@BonomoAlessandro my point is: these changes override each other. Are they going to be in different releases? If not, it does really make sense to have this one.

I understand that you want to split the scopes. I will approve this one, but have a look if really needed (if the other one merges first, this one is not needed for example)

BonomoAlessandro:
@AntonioVentilii I am aware of this conflict. My plan was to resolve the conflict correctly so that this changes does not get overridden.

AntonioVentilii (changes requested):
Let's close this PR, the other one is being merged already, so it does not really make sense to have this one. It will become an empty PR if the other one is merged first...

BonomoAlessandro:
@AntonioVentilii Those 2 PR do not the same thing. This one removes the sprinkles card from the carousel. In the orther PR i removed the line of code that conflicts with this one. Both PRs are needed.

AntonioVentilii (approved):
So, this variable will not change anymore to the new campaign in the other PR (see https://github.com/dfinity/oisy-wallet/pull/6603#discussion_r2088653225)

That means it is good

PR #6605, opened 2025-05-14
fix(frontend): updated swap context types

roman-nazaruk:
# Motivation
Updated swap context types

# Changes 

Imported real Context Types for icTokenFeeStore and swapAmountsStore


AntonioVentilii (approved):
Thank you for the fix!

PR #6607, opened 2025-05-14
fix(frontend): updated swap context type in SwapModal

roman-nazaruk:
# Motivation

Updated swap context types 

# Changes

Imported real context type swapAmountsStore


AntonioVentilii (approved):

AntonioVentilii (approved):
Thank you for the fix

PR #6609, opened 2025-05-14
test(frontend): flaky timestamp for test of `btcKnownDestinations` and `solKnownDestinations`

AntonioVentilii:
# Motivation

Similar to what we did in PR https://github.com/dfinity/oisy-wallet/pull/6481, when testing derived store `btcKnownDestinations` and `solKnownDestinations`, the timestamp created for the mock transactions can be higher than the previous one. So, to avoid flakiness (see [failed CI workflow](https://github.com/dfinity/oisy-wallet/actions/runs/15020207936/job/42207367087?pr=6606)), we need to confirm that the store takes the most recent timestamp.


DenysKarmazynDFINITY (approved):
Thanks!

PR #6612, opened 2025-05-14
feat(backend)!: introducing new endpoint for contacts structure

SYInacta:
# Motivation

This PR introduces a new "Contacts" feature to the wallet backend. The motivation is to provide users with an address book functionality, allowing them to securely store and manage addresses (Token Account IDs) of frequently used contacts or services directly within their wallet profile. This enhances usability by making it easier to send funds to known recipients without needing to re-enter addresses repeatedly.

# Changes

This PR includes the following changes:

1.  **Added Contact Data Structures:** Defined new Candid types (`Contact`, `ContactAddressData`, `ContactSettings`, `AddContactRequest`, `RemoveContactRequest`, `UpdateContactRequest`, `AddAddressRequest`, `UpdateAddressRequest`, `ContactError`) in `shared/src/types/contact.rs` to represent contact information and the associated request/response structures.
2.  **Introduced Stable Storage for Contacts:** Added a new `StableBTreeMap`, `ContactMap`, to `src/backend/src/types.rs` to persistently store user contacts keyed by their principal.
3.  **Exposed Contact API Endpoints:** Added new update and query methods to `src/backend/src/lib.rs` that expose the contact management functionality via the canister's public interface. These methods delegate calls to the corresponding functions in the `contacts` module and include necessary guards (`caller_is_not_anonymous`). The new methods are:
    *   `create_contact`
    *   `update_existing_contact`
    *   `delete_contact`
    *   `get_contact_by_id`
    *   `list_contacts`
    *   `get_contact_settings`
    *   `search_contacts_by_query`
    *   `filter_contacts_by_token_type`
    *   `add_address_to_contact`
    *   `update_contact_address`
5.  **Updated Candid Interface:** Generated the updated Candid definition (`backend.did`) reflect the new contact management methods and types.


src/backend/src/lib.rs
```diff
@@ -834,4 +842,253 @@ pub fn get_snapshot() -> Option<UserSnapshot> {
...
 }
 
+// ===== Contacts CRUD API =====
+
+/// Creates a new contact for the user
+///
+/// # Errors
+/// Errors are enumerated by: `ContactError`
+#[update(guard = "caller_is_not_anonymous")]
+pub fn create_contact(request: AddContactRequest) -> Result<(), ContactError> {
```
AntonioVentilii, line 852: may you please define a type for the return? similar to `AddUserCredentialResult`

AntonioVentilii (changes requested):
Before opening the endpoint, it is necessary to create checks for the length of the inputs, as suggested in https://github.com/dfinity/oisy-wallet/pull/5815#discussion_r2088551305

furthermore, may you please add tests here too?

src/backend/src/lib.rs
SYInacta: Will add in other PR, now it is just the structure.

src/shared/src/types/contact.rs
```diff
@@ -60,12 +59,8 @@ pub struct UpdateAddressRequest {
 
 #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
 pub enum ContactError {
-    UserNotFound,
     ContactNotFound,
     ContactIdAlreadyExists,
     ContactNameAlreadyExists,
```
FlavioInacta, line 64: Remove

src/backend/backend.did
```diff
@@ -407,27 +431,32 @@ service : (Arg) -> {
...
 	set_many_user_tokens : (vec UserToken) -> ();
 	set_snapshot : (UserSnapshot) -> ();
-	set_user_show_testnets : (SetShowTestnetsRequest) -> (Result_8);
+	set_user_show_testnets : (SetShowTestnetsRequest) -> (Result_10);
 	set_user_token : (UserToken) -> ();
 	stats : () -> (Stats) query;
-	top_up_cycles_ledger : (opt TopUpCyclesLedgerRequest) -> (Result_9);
-	update_user_network_settings : (SaveNetworksSettingsRequest) -> (Result_8)
+	top_up_cycles_ledger : (opt TopUpCyclesLedgerRequest) -> (Result_11);
+	update_existing_contact : (AddContactRequest) -> (Result_5);
```
FlavioInacta, line 460: Rename to `update_contact`.

src/shared/src/types/contact.rs
```diff
@@ -60,12 +59,8 @@ pub struct UpdateAddressRequest {
 
 #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
 pub enum ContactError {
-    UserNotFound,
     ContactNotFound,
     ContactIdAlreadyExists,
```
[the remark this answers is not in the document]
SYInacta: Removed

src/shared/src/types/contact.rs
SYInacta: Removed

src/backend/backend.did
SYInacta: Renamed

src/backend/backend.did
```diff
@@ -407,27 +431,32 @@ service : (Arg) -> {
 	);
 	btc_select_user_utxos_fee : (SelectedUtxosFeeRequest) -> (Result_4);
 	config : () -> (Config) query;
-	create_pow_challenge : () -> (Result_5);
+	create_contact : (AddContactRequest) -> (Result_5);
```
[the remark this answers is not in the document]
SYInacta: renamed,
autogenerated did.

DecentAgeCoder (approved):
LGTM

src/backend/src/lib.rs
```diff
@@ -834,4 +842,111 @@ pub fn get_snapshot() -> Option<UserSnapshot> {
...
 }
 
+/// Creates a new contact for the caller.
+///
+/// # Errors
+/// Errors are enumerated by: `ContactError`.
+#[update(guard = "caller_is_not_anonymous")]
+pub fn create_contact(_request: AddContactRequest) -> Result<(), ContactError> {
+    // Placeholder implementation
+    Err(ContactError::ContactIdAlreadyExists)
```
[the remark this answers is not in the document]
SYInacta: Fixed.

src/backend/src/lib.rs
```diff
@@ -834,4 +842,111 @@ pub fn get_snapshot() -> Option<UserSnapshot> {
...
+    // Placeholder implementation
+    Vec::new()
+}
+
+/// Adds an address to an existing contact.
+///
+/// # Errors
+/// Errors are enumerated by: `ContactError`.
+#[update(guard = "caller_is_not_anonymous")]
+pub fn add_address_to_contact(_request: AddAddressRequest) -> Result<(), ContactError> {
```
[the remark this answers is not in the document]
SYInacta: Removed.

src/backend/src/lib.rs
```diff
@@ -833,4 +837,91 @@ pub fn get_snapshot() -> Option<UserSnapshot> {
...
 
+/// Creates a new contact for the caller.
+///
+/// # Errors
+/// Errors are enumerated by: `ContactError`.
+///
+/// # Returns
+/// The ID of the created contact on success.
+#[update(guard = "caller_is_not_anonymous")]
+pub fn create_contact(request: CreateContactRequest) -> Result<Contact, ContactError> {
```
AntonioVentilii, line 848: may you please create returns type like `AddUserCredentialResult` for all these new endpoints?

AntonioVentilii (changes requested):

src/backend/src/lib.rs
AntonioVentilii: If it helps and the PR becomes too big, you can add each method in a different PR (how it should be BTW)

AntonioVentilii: and tests?

src/backend/src/lib.rs
SYInacta: The endpoints now are place holders.
The tests and return types will be added in our next PR for every single endpoint logic with tests and return type.

AntonioVentilii: This is prone to attacks, since it can be queried infinitely with no logical blocker, and consume all the cycles.

If you need it just for the type interface on the Frontend, add a guard that block all requests to this endpoints, maybe `caller_is_allowed`.

Another suggestion would be to mark them as `query` but maybe it will have the wrong interface that you need, certified or uncertified? But I imagine you will need certified calls

AntonioVentilii: RE return types: you already have the structure for response and errors I see. You just need to change it to a structure similar to `AddUserCredentialResult`

src/declarations/backend/backend.did
```diff
@@ -407,15 +428,19 @@ service : (Arg) -> {
 	);
 	btc_select_user_utxos_fee : (SelectedUtxosFeeRequest) -> (Result_4);
 	config : () -> (Config) query;
-	create_pow_challenge : () -> (Result_5);
+	create_contact : (CreateContactRequest) -> (Result_5);
+	create_pow_challenge : () -> (Result_6);
 	create_user_profile : () -> (UserProfile);
+	delete_contact : (nat64) -> (Result_7);
```
FlavioInacta, line 434: Why do you return a nat64 here?

src/declarations/backend/backend.did
```diff
@@ -174,6 +185,7 @@ type CreateChallengeResponse = record {
 	start_timestamp_ms : nat64;
 	expiry_timestamp_ms : nat64
 };
+type CreateContactRequest = record { name : text };
```
FlavioInacta, line 188: Why can we only provide a name for the new contact? Its a bit strange but would work in our usecase.

FlavioInacta (changes requested):
Yes please, name the return types properly and the PR has failing tests. Please fix those.

src/backend/src/lib.rs
SYInacta: Understanded your point, added the guard.

src/declarations/backend/backend.did
SYInacta: this is autogenerated.

src/declarations/backend/backend.did
FlavioInacta: Your answer makes no sence to me. You added the return type here. So it just generates the type that you provide. Why is there only a name property is the question.

https://github.com/dfinity/oisy-wallet/pull/6612/files#diff-1bf1712159e48316cb0f391cb4cedd766e5961e1f1dbfaf3501a83ce962bbeb0R29

src/declarations/backend/backend.did
FlavioInacta: Yes I know. But you defined u64 as return value for the method:
https://github.com/dfinity/oisy-wallet/pull/6612/files#diff-90adc9a158c75e05b279f456c7b387cbe4b50b39f48c7a47e2a040522a50c510R882

Why?

src/backend/src/lib.rs
SYInacta: Response types and tests will be added with the full logic.

FlavioInacta (commented):
I leave it to Antonio how the return types must look like in this stage. I'm happy if I finally get the `TokenAccountId` merged somehow.

src/backend/src/types.rs
```diff
@@ -22,5 +23,5 @@ pub struct Candid<T>(pub T)
 where
     T: CandidType + for<'de> Deserialize<'de>;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, CandidType, SerdeDeserialize)]
```
AntonioVentilii, line 26: why is this being changed? does it create backwards incompatibility? is it ok, since it is used throughout the backend?

AntonioVentilii (changes requested):
To be honest, this PR is becoming a bit messy: I see that some types and validations are being removed.

Not so sure that the approach "first empty methods, then implementation" is working.

If you need types for the frontend, why not just mock them in the frontend?

I mean, unless you are 100% that these are the final types that you will need. We should really try and avoid changing the backend too frequently.

src/declarations/backend/backend.did
SYInacta: pub id: u64 definition in the types., why it generates nat64 will have to check.

src/declarations/backend/backend.did
SYInacta: I can't answer for why auto generation does that. Will investigate if we have the time.

src/backend/src/types.rs
SYInacta: Fixed.

src/backend/backend.did
```diff
@@ -427,11 +452,12 @@ service : (Arg) -> {
 	set_many_custom_tokens : (vec CustomToken) -> ();
 	set_many_user_tokens : (vec UserToken) -> ();
 	set_snapshot : (UserSnapshot) -> ();
-	set_user_show_testnets : (SetShowTestnetsRequest) -> (Result_8);
+	set_user_show_testnets : (SetShowTestnetsRequest) -> (Result_10);
```
bitdivine, line 455: You asked about result types.  Here is an example of a spurious diff caused by Candid  not having a better name than Result_N.

src/backend/src/lib.rs
```diff
@@ -833,4 +837,95 @@ pub fn get_snapshot() -> Option<UserSnapshot> {
...
+/// The created contact on success.
+///
+/// # Test
+/// This endpoint is currently a placeholder and will be fully implemented in a future PR.
+#[update(guard = "caller_is_allowed")]
+#[must_use]
+pub fn create_contact(request: CreateContactRequest) -> CreateContactResult {
+    // TODO replace mock data with contact service that returns Contact
+    let contact = Contact {
+        id: time(),
```
bitdivine, line 855: Note that the precision of this time is about 1 second.  More digits are provided, but the clock updates only once per block.  So the precision is only as good as the block frequency.

src/shared/src/types/contact.rs
```diff
@@ -5,10 +5,10 @@ use super::account::TokenAccountId;
 #[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
 #[serde(remote = "Self")]
 pub struct Contact {
-    pub id: String,
+    pub id: u64,
     pub name: String,
     pub addresses: Vec<ContactAddressData>,
-    pub avatar: Option<String>,
+    pub update_timestamp: u64,
```
bitdivine, line 11: Please mention the timestamp units in the Rustdoc, and/or use a suffix to indicate the timestamp precision.  E.g. `update_timestamp_ns`.

src/backend/src/lib.rs
```diff
@@ -833,4 +837,95 @@ pub fn get_snapshot() -> Option<UserSnapshot> {
...
+
+    CreateContactResult::Ok(contact)
+}
+
+/// Updates an existing contact for the caller.
+///
+/// # Errors
+/// Errors are enumerated by: `ContactError`.
+#[update(guard = "caller_is_allowed")]
+pub fn update_contact(request: UpdateContactRequest) -> Result<Contact, ContactError> {
```
bitdivine, line 869: You will need to change the Result type for a custom type, in order to clean up the candid type names.

src/backend/src/lib.rs
```diff
@@ -833,4 +837,95 @@ pub fn get_snapshot() -> Option<UserSnapshot> {
...
+
+    Ok(contact)
+}
+
+/// Deletes a contact for the caller.
+///
+/// # Errors
+/// Errors are enumerated by: `ContactError`.
+#[update(guard = "caller_is_allowed")]
+pub fn delete_contact(contact_id: u64) -> Result<u64, ContactError> {
```
bitdivine, line 886: Same here.

src/backend/src/lib.rs
```diff
@@ -833,4 +837,95 @@ pub fn get_snapshot() -> Option<UserSnapshot> {
...
+    // TODO intergrate delete contact service
+    Ok(contact_id)
+}
+
+/// Gets a contact by ID for the caller.
+///
+/// # Errors
+/// Errors are enumerated by: `ContactError`.
+#[query(guard = "caller_is_not_anonymous")]
+pub fn get_contact(contact_id: u64) -> Result<Contact, ContactError> {
```
bitdivine, line 896: Same here.

bitdivine (approved):
A few more interface types are still needed but otherwise this looks good to me.  Thanks!

src/declarations/backend/backend.did
FlavioInacta: The question is why you return a number. Not what kind of number.

src/backend/src/lib.rs
SYInacta: will somthing like that work ok?
 thread_local! {
        static COUNTER: RefCell<u64> = RefCell::new(0);
    }
    
    let unique_id = COUNTER.with(|counter| {
        let current = *counter.borrow();
        *counter.borrow_mut() = current.wrapping_add(1);
        (time() << 16) | (current & 0xFFFF) // Combine timestamp with counter
    });

AntonioVentilii (approved):
Unblocking with the strong TODO of creating the result types, sooner than later.

I was in the process of adjusting the other methods too (PR https://github.com/dfinity/oisy-wallet/pull/6371 and https://github.com/dfinity/oisy-wallet/pull/6441), and I was going "bottom-up" to have less disruptions

bitdivine:
Just discussed with Marco and David.  I'll do the type changes for here.  I feel it would be worth getting all the results adapted so that there are no two ways, just one clean way!  I'll give myself a time budget to help.

DecentAgeCoder:
@bitdivine I'm very grateful for your support and for ensuring that all result structures and Candid files remain consistent.

src/backend/src/lib.rs
DecentAgeCoder: Yes we are aware of this issue. The time() call is only used to create mock data. We will correctly address the id issue with the PR containing the implementation.

src/shared/src/types/contact.rs
DecentAgeCoder: I have renamed all update_timestamp to update_timestamp_ns

src/backend/src/lib.rs
SYInacta: Adding in next PR.

src/backend/src/lib.rs
SYInacta: Adding in next PR.

src/backend/src/lib.rs
SYInacta: Adding in next PR.

src/shared/src/types/contact.rs
```diff
@@ -5,10 +5,10 @@ use super::account::TokenAccountId;
 #[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
 #[serde(remote = "Self")]
 pub struct Contact {
-    pub id: String,
+    pub id: u64,
     pub name: String,
     pub addresses: Vec<ContactAddressData>,
-    pub avatar: Option<String>,
```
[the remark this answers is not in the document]
SYInacta: no contacts have been saved.

src/shared/src/types/contact.rs
SYInacta: ok.

src/backend/backend.did
SYInacta: Understood.

PR #6613, opened 2025-05-14
feat(frontend): added new SwapProviderListModal component

roman-nazaruk:
# Motivation

Added new component to see list of available swaps to select the best one

# Changes

Created a new component SwapProviderListModal

# Tests


src/frontend/src/tests/lib/components/swap/SwapProviderListModal.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+		expect(getByText('Swap provider')).toBeInTheDocument();
+		expect(getByText('You receive')).toBeInTheDocument();
+	});
+
+	it('renders all provider rows with USD values', () => {
+		const { getAllByText } = render(SwapProviderListModal, {
+			context: mockContext
+		});
+
+		expect(getAllByText(/\$\d+\.\d{2}/)).toHaveLength(2);
```
AntonioVentilii, line 43: may you please add some test case about the inputs of `getUsdBalance` and the outputs of the context store? the logic is inside the component, so it should be tested

For example, first case: does the amount value recalculates if $`destinationTokenExchangeRate` changes?

src/frontend/src/lib/components/swap/SwapProviderListModal.svelte
```diff
@@ -0,0 +1,71 @@
...
+		formatUSD({
+			value:
+				nonNullish(amount) && nonNullish(token) && nonNullish($destinationTokenExchangeRate)
+					? formatTokenBigintToNumber({
+							value: amount,
+							unitName: token.decimals,
+							displayDecimals: token.decimals
+						}) * $destinationTokenExchangeRate
+					: 0
+		});
```
AntonioVentilii, line 43: this is not reactive on `$destinationTokenExchangeRate`, is that ok?

AntonioVentilii (approved):
LGTM tks!

PR #6614, opened 2025-05-14
feat(frontend): added sorting for swap results

roman-nazaruk:
# Motivation

Added logic to filter and sort swap results to improve data integrity and user experience.
Only swap results with a receiveAmount greater than 0 are returned, to avoid unnecessary UI-level checks and prevent displaying invalid swap options.
Sorting ensures that the swap option with the best rate is displayed first.

# Changes

- Added filtering logic to exclude swap results where receiveAmount <= 0.

- Added sorting logic to order swap results by receiveAmount in descending order.

- This logic helps simplify UI rendering and prevents invalid swap options from being shown.

# Tests 

Added unit tests to:

- Ensure swap results with zero receiveAmount are filtered out.

- Ensure results are sorted in descending order by receiveAmount.


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -163,6 +163,8 @@ export const fetchSwapAmounts = async ({
 			acc.push(mapped);
 		}
 
-		return acc;
+		return acc
+			.filter((swap) => Number(swap.receiveAmount) > 0)
```
AntonioVentilii, line 167: this is something new too, is it intended? Asking because it wasn't in the description

AntonioVentilii (commented):
Some tests?

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: I will update this logic and will update description

roman-nazaruk: Updated logic and description, also updated PR info.

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -147,22 +147,29 @@ export const fetchSwapAmounts = async ({
...
-		}
-
-		return acc;
-	}, []);
+		},
+		[]
+	);
+
+	return mappedProvidersResults
+		.filter((swap) => Number(swap.receiveAmount) > 0)
```
AntonioVentilii, line 173: this `filter` can be introduced directly inside the `reduce` as if condition

PR #6615, opened 2025-05-14
refactor(frontend): updated swap factory canister constant name

roman-nazaruk:
# Motivation

Updated swap factory canister constant name


AntonioVentilii (approved):
LGTM tks!

PR #6616, opened 2025-05-14
feat(backend): Added and enforced validation for address book structures

DecentAgeCoder:
# Motivation

Enhanced and enforce data validation for contact-related structures to improve reliability and prevent potential DDoS attacks and excessive data store growth.

# Changes

- Added several constants to set limits on string length and collection sizes for contact data.
- Introdduced helper validation functions for checking string length and collection sizes.
- Implemented the Validate trait for all address book related structures
- Added serde attributes (#[serde(remote = "Self")]) to multiple structs in contact.rs for improved deserialization.
- Applied `validate_on_deserialize`! macros to enforce validation during deserialization for new and existing contact-related types.


src/shared/src/impls.rs
```diff
@@ -24,6 +28,39 @@ use crate::{
...
 };
 
+// Constants for validation limits
+const MAX_ID_LENGTH: usize = 100;
+const MAX_NAME_LENGTH: usize = 100;
+const MAX_AVATAR_LENGTH: usize = 100;
+const MAX_ADDRESSES: usize = 40;
+const MAX_LABEL_LENGTH: usize = 50;
+
+// TODO what is a reasanoble value here?
```
bitdivine, line 38:
```suggestion
// TODO: what is a reasonable value here?
```

src/shared/src/impls.rs
```diff
@@ -24,6 +28,39 @@ use crate::{
...
+const MAX_CONTACTS: usize = 100;
+
+// Helper functions for validation
+fn validate_string_length(value: &str, max_length: usize, field_name: &str) -> Result<(), Error> {
+    if value.len() > max_length {
```
bitdivine, line 43: `value.len()` counts bytes, so e.g. a string with just a unicode snowman has one character but .len() returns 4.  If you would like to count characters, you need `.chars().count()`.  Or you could update the error message below, but I suspect that communicating a character limit is easier that communicating a byte limit.  You may well get users saying "but my string _is_ less than 10 characters long".  Because users don't see their strings as unicode.

bitdivine (approved):
LGTM.  Thanks!

src/shared/src/impls.rs
DecentAgeCoder: Thankns for the input. I was not aware that the length on a String actually returns a the bytes and not the number of "characters". I will adapt the code to use .chars().count() as you suggest.

src/shared/src/impls.rs
```diff
@@ -24,6 +28,39 @@ use crate::{
     validate::{validate_on_deserialize, Validate},
 };
 
+// Constants for validation limits
+const MAX_ID_LENGTH: usize = 100;
```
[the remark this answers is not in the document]
DecentAgeCoder: I have prefixed the validation rule constants with "CONTACT". I recommend modularizing the implementations in the near future, which would eliminate the need for namespace-style prefixes.

src/shared/src/impls.rs
DecentAgeCoder: The TODO has been removed

PR #6618, opened 2025-05-14
feat(frontend): TokenAccountId and Address mappings and schemas

FlavioInacta:
# Motivation

Addresses in the address book will be stored as `TokenAccountId`s. Since the user provides the addresses as strings we need code that does the mapping between `string`s and `TokenAccountId`s. This PR adds parsing and methods to get the string representation back from `TokenAccountId`s.

# Changes

- Add some utils to work with discriminative unions
- Adds parsing/serialization for all supported address types
- Adds parsing/serialization for all `TokenAccountId`


AntonioVentilii (commented):
May you please split this PR?

FlavioInacta:
Will split up

PR #6619, opened 2025-05-14
fix(devops): Don't re-zip the frontend Wasm

bitdivine:
# Motivation
`dfx deploy frontend` currently, curiously, deploys a different Wasm from the one provided upstream and bundled in dfx.  This is evidenced by the different Wasm hash shown by `dfx canister info frontend`.

A vanilla `dfx deploy frontend` in a brand new toy project _does_ deploy the expected Wasm.

It turns out that  removing the `"gzip": true,` in `dfx.json` for the front-end canister deploys a Wasm with the expected hash.  Presumably that field causes the Wasm to be re-compressed in a slightly different way.

# Changes

- Remove the offending configuration from `dfx.json`.
- Set `--no-asset-upgrade` in the canister deployment workflow to minimize disruption.  This way the workflow can run without either having to be controller of all the test canisters, or all the test & staging canisters having to be upgraded at the moment that this PR is merged.

# Tests
I deployed the front-end to test_fe_5 and the front-end Wasm hash has the expected value:

```
$ curl -sSL "https://github.com/dfinity/sdk/raw/refs/tags/$(jq -r .dfx dfx.json)/src/distributed/assetstorage.wasm.gz" | sha256sum
865eb25df5a6d857147e078bb33c727797957247f7af2635846d65c5397b36a6  -
$ dfx canister info  frontend --network test_fe_5
[snip]
Module hash: 0x865eb25df5a6d857147e078bb33c727797957247f7af2635846d65c5397b36a6
```


AntonioVentilii (approved):
Thank you for this!

PR #6621, opened 2025-05-14
refactor(frontend): simplify transactions in component `AllTransactionsLoader`

AntonioVentilii:
# Motivation

It doesn't really make sense that we fix the values of the transactions stores in the function to load of component `AllTransactionsLoader`. In fact, the fear that one token would make the other loop is unfounded since we use ONLY the transactions of each token for the analysis of itself.

Each token will be agnostic to the other tokens transactions inside that query.

It is cleaner and more consistent to put it directly where they are queried.


DenysKarmazynDFINITY (approved):
👍

PR #6624, opened 2025-05-14
refactor(frontend): move `result_types` to shared folder

AntonioVentilii:
# Motivation

We want to use the result types in the tests too. So we need to move them to the shared folder.

### DIFF
--- a/src/backend/src/lib.rs
+++ b/src/backend/src/lib.rs
@@ -38,6 +38,7 @@ use shared::{
             AllowSigningStatus, ChallengeCompletion, CreateChallengeError, CreateChallengeResponse,
             CYCLES_PER_DIFFICULTY, POW_ENABLED,
         },
+        result_types::AddUserCredentialResult,
         signer::{
             topup::{TopUpCyclesLedgerRequest, TopUpCyclesLedgerResult},
             AllowSigningRequest, AllowSigningResponse, GetAllowedCyclesError,
@@ -52,7 +53,6 @@ use shared::{
         Stats, Timestamp,
     },
 };
-use shared::types::result_types::AddUserCredentialResult;
 use signer::{btc_principal_to_p2wpkh_address, AllowSigningError};
 use types::{
     Candid, ConfigCell, CustomTokenMap, StoredPrincipal, UserProfileMap, UserProfileUpdatedMap,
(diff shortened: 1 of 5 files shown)


src/shared/src/types/result_types.rs
```diff
@@ -9,6 +10,26 @@ pub enum AddUserCredentialResult {
     /// The user's credential was not added due to an error.
     Err(AddUserCredentialError),
 }
+impl AddUserCredentialResult {
```
AntonioVentilii, line 13: not sure if this should go here or in `shared/src/impls.rs`

bitdivine: IMO the important thing is to put externally visible types like this in a separate crate, which this does, so it looks good to me.

src/backend/src/user_profile.rs
```diff
@@ -46,7 +47,7 @@ pub fn add_credential(
     credential_type: &CredentialType,
     issuer: String,
     user_profile_model: &mut UserProfileModel,
-) -> Result<(), AddUserCredentialError> {
```
bitdivine, line 49: I would recommend using Result in the internal implementation, and using the Result type only at the entry point.  This way you can use normal Rust constructs, such as `?`, in the implementation of the API.

src/backend/src/lib.rs
```diff
@@ -528,9 +527,8 @@ pub fn add_user_credential(request: AddUserCredentialRequest) -> AddUserCredenti
                 &mut user_profile_model,
             )
```
bitdivine, line 528:
```suggestion
            ).into()
```
I would recommend using normal rust Results in the inner implementation.  These custom Result types exist only because of an annoying aspect of Candid bindings; they don't benefit the rest of the code.  And at some point Candid might make this easier, at which point we can throw the Result types away.  So let's make it easy to remove them.

bitdivine (approved):

src/backend/src/user_profile.rs
AntonioVentilii: reverted!

src/backend/src/lib.rs
AntonioVentilii: reverted!

PR #6629, opened 2025-05-15
fix(frontend): update ConvertAmountExchange skeleton logic

DenysKarmazynDFINITY:
# Motivation

We need to hide skeleton in ConvertAmountExchange if exchange rate is not available for a token.

# Changes

1. `ConvertAmountExchange` update - the exchange skeleton will be shown only while `value` is not available.
2. Replaced hardcoded testIds in the related components.

# Tests

Added new tests.


AntonioVentilii (approved):
LGTM tks!

PR #6631, opened 2025-05-15
refactor(devops): Don't prompt on re-download

bitdivine:
# Motivation
The prompt on whether to re-download assets is annoying and time-consuming.

# Changes
- Re-download only if the URL has changed.

# Tests
- This is used in many of the build commands execrised in CI.
- I have used this locally and can confirm that re-running `npm run generate` no longer requires prompts.


AntonioVentilii (approved):
Thank you!

PR #6633, opened 2025-05-15
fix(frontend): avoid looping of restarting IC Wallet Workers

AntonioVentilii:
# Motivation

The IC wallet worker restarts itself if there is an error with the index canister, to restart it with ledger balance only. However, this should happen only once. Otherwise it will keep going in loops to restart itself.

# Changes

- Restart the worker `initIcrcWalletWorker` only if there was an index canister before (it does not make sense to restart it for tokens that already have only the ledger).
- Add a check to verify if it was already restarted once.

# Tests

Created a few tests. Plus, mocking a ledger canister error, we can check that it would have gone in infinite loops before this PR.

### Before

Screenshot 2025-05-15 at 16 15 51

### After

Screenshot 2025-05-15 at 16 15 16


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6635, opened 2025-05-15
build(frontend): Bump gix-components to next

AntonioVentilii:
# Motivation

We want to add some breaking changes done to `InfiniteScroll` of GIX components (PR https://github.com/dfinity/gix-components/pull/629). So we bump the package to `next` version.

For its correct usage, we need to adapt the components that use `InfiniteScroll`.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -512,8 +512,8 @@
 		},
 		"rewards": {
 			"type": "custom",
-			"candid": "https://github.com/dfinity/oisy-wallet/releases/download/rc0.5.2-api.1/rewards.did",
-			"wasm": "https://github.com/dfinity/oisy-wallet/releases/download/rc0.5.2-api.1/rewards.wasm.gz",
+			"candid": "https://github.com/dfinity/oisy-wallet/releases/download/rc0.5.3-api.1/rewards.did",
+			"wasm": "https://github.com/dfinity/oisy-wallet/releases/download/rc0.5.3-api.1/rewards.wasm.gz",
 			"remote": {
 				"id": {
 					"ic": "nynz6-haaaa-aaaan-qzqda-cai",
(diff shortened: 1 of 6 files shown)


src/frontend/src/sol/components/transactions/SolTransactionsScroll.svelte
```diff
@@ -1,14 +1,20 @@
...
+	import type { Snippet } from 'svelte';
 	import type { Token } from '$lib/types/token';
 	import { last } from '$lib/utils/array.utils';
 	import { solTransactions } from '$sol/derived/sol-transactions.derived';
 	import { loadNextSolTransactions } from '$sol/services/sol-transactions.services';
 
-	export let token: Token;
+	interface Props {
+		token: Token;
+		children?: Snippet;
```
peterpeterparker, line 12: children is mandatory here no?

peterpeterparker: But maybe that's something you want to fix in another PR since IcTransactionsScroll also takes option instead of mandatory

peterpeterparker (approved):
LGTM, thx

A comment that can be resolved in separate PR

src/frontend/src/sol/components/transactions/SolTransactionsScroll.svelte
AntonioVentilii: no, you are right, it should be mandatory. I put it mandatory on purpose in GIX components, because the idea is that it should enforce the usage with children. Changed on both in this PR.

Or do you prefer in a separate one?

peterpeterparker: All good, thx

peterpeterparker (approved):

PR #6640, opened 2025-05-16
feat(frontend): SendFeeInfo component

DenysKarmazynDFINITY:
# Motivation

In order to display an info banner when a fee symbol does not match with a send token symbol, we need to have a new reusable component.

Screenshot 2025-05-16 at 11 49 16


AntonioVentilii (approved):
LGTM tks!

PR #6643, opened 2025-05-16
refactor(frontend): apply lint rule `require-event-prefix`

AntonioVentilii:
# Motivation

Before bumping the lint library in PR https://github.com/dfinity/oisy-wallet/pull/6637, we need to manually solve lint rule `require-event-prefix`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6645, opened 2025-05-16
feat(frontend): added icp swap service method

roman-nazaruk:
# Motivation

Implemented a new service method fetchIcpSwap to handle token swaps, including transfer, deposit, swap, and withdrawal steps.

# Changes

- Created fetchIcpSwap to encapsulate the ICP Swap logic.

- Integrated i18n error messages.

- Handled different token standards (ICRC1 and ICRC2) with branching logic.

- Added fallback and compensation logic for failed swap/withdraw steps.

# Tests

Covered new logic with the tests.

### DIFF
--- a/src/frontend/src/lib/services/swap.services.ts
+++ b/src/frontend/src/lib/services/swap.services.ts
@@ -2,12 +2,14 @@ import type { SwapAmountsReply } from '$declarations/kong_backend/kong_backend.d
 import { approve } from '$icp/api/icrc-ledger.api';
 import { sendIcp, sendIcrc } from '$icp/services/ic-send.services';
 import { loadCustomTokens } from '$icp/services/icrc.services';
-import type { IcTokenToggleable } from '$icp/types/ic-token-toggleable';
 import { nowInBigIntNanoSeconds } from '$icp/utils/date.utils';
 import { isTokenIcrc } from '$icp/utils/icrc.utils';
 import { setCustomToken } from '$lib/api/backend.api';
+import { getPoolCanister } from '$lib/api/icp-swap-factory.api';
+import { deposit, depositFrom, swap as swapIcp, withdraw } from '$lib/api/icp-swap-pool.api';
 import { kongSwap, kongTokens } from '$lib/api/kong_backend.api';
 import { KONG_BACKEND_CANISTER_ID, NANO_SECONDS_IN_MINUTE } from '$lib/constants/app.constants';
+import { ICP_SWAP_POOL_FEE } from '$lib/constants/swap.constants';
 import { ProgressStepsSwap } from '$lib/enums/progress-steps';
 import { swapProviders } from '$lib/providers/swap.providers';
 import { i18n } from '$lib/stores/i18n.store';
@@ -15,23 +17,23 @@ import {
 	kongSwapTokensStore,
 	type KongSwapTokensStoreData
 } from '$lib/stores/kong-swap-tokens.store';
-import type { OptionIdentity } from '$lib/types/identity';
-import type { Amount } from '$lib/types/send';
 import {
 	SwapProvider,
 	type FetchSwapAmountsParams,
 	type ICPSwapResult,
-	type SwapMappedResult
+	type SwapMappedResult,
+	type SwapParams
 } from '$lib/types/swap';
 import { toCustomToken } from '$lib/utils/custom-token.utils';
 import { parseToken } from '$lib/utils/parse.utils';
+import { calculateSlippage } from '$lib/utils/swap.utils';
 import { waitAndTriggerWallet } from '$lib/utils/wallet.utils';
 import type { Identity } from '@dfinity/agent';
 import { Principal } from '@dfinity/principal';
-import { nonNullish } from '@dfinity/utils';
+import { isNullish, nonNullish } from '@dfinity/utils';
 import { get } from 'svelte/store';
 
-export const swap = async ({
+export const fetchKongSwap = async ({
 	identity,
 	progress,
 	sourceToken,
@@ -41,17 +43,7 @@ export const swap = async ({
 	slippageValue,
 	sourceTokenFee,
 	isSourceTokenIcrc2
-}: {
-	identity: OptionIdentity;
-	progress: (step: ProgressStepsSwap) => void;
-	sourceToken: IcTokenToggleable;
-	destinationToken: IcTokenToggleable;
-	swapAmount: Amount;
-	receiveAmount: bigint;
-	slippageValue: Amount;
-	sourceTokenFee: bigint;
-	isSourceTokenIcrc2: boolean;
-}) => {
+}: SwapParams) => {
 	progress(ProgressStepsSwap.SWAP);
 
 	const parsedSwapAmount = parseToken({
@@ -174,4 +166,128 @@ export const fetchSwapAmounts = async ({
 	);
 
 	return mappedProvidersResults.sort((a, b) => Number(b.receiveAmount) - Number(a.receiveAmount));
+};
+
+export const fetchIcpSwap = async ({
+	identity,
+	progress,
+	sourceToken,
+	destinationToken,
+	swapAmount,
+	receiveAmount,
+	slippageValue,
+	sourceTokenFee,
+	isSourceTokenIcrc2
+}: SwapParams): Promise<void> => {
+	progress(ProgressStepsSwap.SWAP);
+
+	const parsedSwapAmount = parseToken({
+		value: `${swapAmount}`,
+		unitName: sourceToken.decimals
+	});
+
+	const pool = await getPoolCanister({
+		identity,
+		token0: { address: sourceToken.ledgerCanisterId, standard: sourceToken.standard },
+		token1: { address: destinationToken.ledgerCanisterId, standard: destinationToken.standard },
+		nullishIdentityErrorMessage: get(i18n).auth.error.no_internet_identity,
+		fee: ICP_SWAP_POOL_FEE
+	});
+
+	if (isNullish(pool)) {
+		throw new Error(get(i18n).swap.error.pool_not_found);
+	}
+
+	const slippageMinimum = calculateSlippage({
+		quoteAmount: receiveAmount,
+		slippagePercentage: Number(slippageValue)
+	});
+
+	const transferParams = {
+		identity,
+		token: sourceToken,
+		amount: parsedSwapAmount,
+		to: pool.canisterId.toString(),
+		ledgerCanisterId: sourceToken.ledgerCanisterId
+	};
+
+	try {
+		if (!isSourceTokenIcrc2) {
+			await sendIcrc(transferParams);
+			await deposit({
+				identity,
+				canisterId: pool.canisterId.toString(),
+				token: sourceToken.ledgerCanisterId,
+				amount: parsedSwapAmount,
+				fee: sourceTokenFee
+			});
+		} else {
+			await approve({
+				identity,
+				ledgerCanisterId: sourceToken.ledgerCanisterId,
+				amount: parsedSwapAmount + sourceTokenFee * 2n,
+				expiresAt: nowInBigIntNanoSeconds() + 5n * NANO_SECONDS_IN_MINUTE,
+				spender: { owner: pool.canisterId }
+			});
+
+			await depositFrom({
+				identity,
+				canisterId: pool.canisterId.toString(),
+				token: sourceToken.ledgerCanisterId,
+				amount: parsedSwapAmount,
+				fee: sourceTokenFee
+			});
+		}
+	} catch (_: unknown) {
+		throw new Error(get(i18n).swap.error.deposit_error);
+	}
+
+	try {
+		await swapIcp({
+			identity,
+			canisterId: pool.canisterId.toString(),
+			amountIn: parsedSwapAmount.toString(),
+			zeroForOne: pool.token0.address === sourceToken.ledgerCanisterId,
+			amountOutMinimum: slippageMinimum.toString()
+		});
+	} catch (_: unknown) {
+		try {
+			await withdraw({
+				identity,
+				canisterId: pool.canisterId.toString(),
+				token: sourceToken.ledgerCanisterId,
+				amount: parsedSwapAmount,
+				fee: sourceTokenFee
+			});
+		} catch (_: unknown) {
+			throw new Error(get(i18n).swap.error.withdraw_failed);
+		}
+
+		throw new Error(get(i18n).swap.error.swap_failed_withdraw_success);
+	}
+
+	try {
+		await withdraw({
+			identity,
+			canisterId: pool.canisterId.toString(),
+			token: destinationToken.ledgerCanisterId,
+			amount: receiveAmount,
+			fee: destinationToken.fee
+		});
+	} catch (_: unknown) {
+		throw new Error(get(i18n).swap.error.withdraw_failed);
+	}
+
+	progress(ProgressStepsSwap.UPDATE_UI);
+
+	if (!destinationToken.enabled) {
+		await setCustomToken({
+			token: toCustomToken({ ...destinationToken, enabled: true, networkKey: 'Icrc' }),
+			identity,
+			nullishIdentityErrorMessage: get(i18n).auth.error.no_internet_identity
+		});
+		await loadCustomTokens({ identity });
+	}
+
+	await waitAndTriggerWallet();
 };
--- a/src/frontend/src/tests/lib/services/icp-swap.services.spec.ts
+++ b/src/frontend/src/tests/lib/services/icp-swap.services.spec.ts
@@ -1,9 +1,22 @@
 import type { IcToken } from '$icp/types/ic-token';
+import type { IcTokenToggleable } from '$icp/types/ic-token-toggleable';
 import * as factoryApi from '$lib/api/icp-swap-factory.api';
+import { getPoolCanister } from '$lib/api/icp-swap-factory.api';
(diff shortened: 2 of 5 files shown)


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +167,127 @@ export const fetchSwapAmounts = async ({
...
+	destinationToken,
+	swapAmount,
+	receiveAmount,
+	slippageValue,
+	sourceTokenFee,
+	isSourceTokenIcrc2
+}: SwapParams): Promise<void> => {
+	progress(ProgressStepsSwap.SWAP);
+
+	const parsedSwapAmount = parseToken({
```
AntonioVentilii, line 184: some initial checks for the correct inputs? for example, amounts non null nor ZERO, Tokens with same standard (if that is a thing), not same ledger canister

and a check for nullish identity too, or is it checked in the caller. If it is, then better to specify the type as non optional

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +167,127 @@ export const fetchSwapAmounts = async ({
...
+				canisterId: pool.canisterId.toString(),
+				token: sourceToken.ledgerCanisterId,
+				amount: parsedSwapAmount,
+				fee: sourceTokenFee
+			});
+		} catch (_: unknown) {
+			throw new Error(get(i18n).swap.error.withdraw_failed);
+		}
+
+		throw new Error(get(i18n).swap.error.swap_failed_withdraw_success);
```
AntonioVentilii, line 266: not really sure I understand this part: how can you be sure that the error requires the `withdraw` call? what if the error is given by something else, like failure in connection? or any other issue. It feels strange to have nested try-catch

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +167,127 @@ export const fetchSwapAmounts = async ({
...
+	}
+
+	try {
+		await withdraw({
```
AntonioVentilii, line 270: are you not calling this twice? is that on purpose?

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +167,127 @@ export const fetchSwapAmounts = async ({
...
+
+	progress(ProgressStepsSwap.UPDATE_UI);
+
+	if (!destinationToken.enabled) {
+		await setCustomToken({
+			token: toCustomToken({ ...destinationToken, enabled: true, networkKey: 'Icrc' }),
+			identity,
+			nullishIdentityErrorMessage: get(i18n).auth.error.no_internet_identity
+		});
+		await loadCustomTokens({ identity });
```
AntonioVentilii, line 289: I think that service `autoLoadSingleToken` does this entire snippet

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +167,127 @@ export const fetchSwapAmounts = async ({
...
+	}
+
+	await waitAndTriggerWallet();
```
AntonioVentilii, line 292: shouldn't there be a last step, like `ProgressStepsSwap.DONE`?

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +167,127 @@ export const fetchSwapAmounts = async ({
...
+	try {
+		await withdraw({
+			identity,
+			canisterId: pool.canisterId.toString(),
+			token: destinationToken.ledgerCanisterId,
+			amount: receiveAmount,
+			fee: destinationToken.fee
+		});
+	} catch (_: unknown) {
+		throw new Error(get(i18n).swap.error.withdraw_failed);
```
AntonioVentilii, line 278: no need to throw this error: the function `withdraw` will already raise an error. If you want to change the feedback, change it directly there

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +167,127 @@ export const fetchSwapAmounts = async ({
...
+		try {
+			await withdraw({
+				identity,
+				canisterId: pool.canisterId.toString(),
+				token: sourceToken.ledgerCanisterId,
+				amount: parsedSwapAmount,
+				fee: sourceTokenFee
+			});
+		} catch (_: unknown) {
+			throw new Error(get(i18n).swap.error.withdraw_failed);
```
AntonioVentilii, line 263: same

src/frontend/src/tests/lib/services/icp-swap.services.spec.ts
```diff
@@ -71,3 +84,112 @@ describe('icpSwapAmounts', () => {
...
+	it('Swap failed. Withdraw Success', async () => {
+		vi.mocked(getPoolCanister).mockResolvedValue(mockPool);
+
+		vi.mocked(sendIcrc).mockResolvedValue(1n);
+		vi.mocked(deposit).mockResolvedValue(1n);
+		vi.mocked(swapIcp).mockRejectedValue(new Error('swap fail'));
+		vi.mocked(withdraw).mockResolvedValue(1n);
+
+		await expect(fetchIcpSwap({ ...swapArgs, isSourceTokenIcrc2: false })).rejects.toThrow(
+			'Swap failed. Your tokens were refunded successfully. You can try again or increase the slippage tolerance.'
```
AntonioVentilii, line 179: use the mock `en` that is basically the mock i18n file

src/frontend/src/tests/lib/services/icp-swap.services.spec.ts
```diff
@@ -71,3 +84,112 @@ describe('icpSwapAmounts', () => {
...
+	});
+
+	it('Swap failed. Deposit failed', async () => {
+		vi.mocked(getPoolCanister).mockResolvedValue(mockPool);
+
+		vi.mocked(sendIcrc).mockResolvedValue(1n);
+		vi.mocked(deposit).mockRejectedValue(new Error('fail'));
+
+		await expect(fetchIcpSwap({ ...swapArgs, isSourceTokenIcrc2: false })).rejects.toThrow(
+			'Swap failed. We couldn’t deposit your tokens into the pool. Please try again.'
```
AntonioVentilii, line 166: same

src/frontend/src/tests/lib/services/icp-swap.services.spec.ts
```diff
@@ -71,3 +84,112 @@ describe('icpSwapAmounts', () => {
...
+	it('Swap and withdraw both failed', async () => {
+		vi.mocked(getPoolCanister).mockResolvedValue(mockPool);
+
+		vi.mocked(sendIcrc).mockResolvedValue(1n);
+		vi.mocked(deposit).mockResolvedValue(1n);
+		vi.mocked(swapIcp).mockRejectedValue(new Error('swap fail'));
+		vi.mocked(withdraw).mockRejectedValue(new Error('withdraw fail'));
+
+		await expect(fetchIcpSwap({ ...swapArgs, isSourceTokenIcrc2: false })).rejects.toThrow(
+			'Swap failed and withdraw also failed. Please check your unused balance and try manual withdrawal: $url%`'
```
AntonioVentilii, line 192: same, but with the replacers

src/frontend/src/tests/lib/services/icp-swap.services.spec.ts
```diff
@@ -71,3 +84,112 @@ describe('icpSwapAmounts', () => {
...
+		expect(depositFrom).toHaveBeenCalled();
+		expect(swapIcp).toHaveBeenCalled();
+		expect(withdraw).toHaveBeenCalled();
+	});
+
+	it('Swap failed. Pool not found', async () => {
+		vi.mocked(getPoolCanister).mockRejectedValue(new Error('pool_not_found'));
+
+		await expect(fetchIcpSwap({ ...swapArgs, isSourceTokenIcrc2: false })).rejects.toThrow(
+			'pool_not_found'
```
AntonioVentilii, line 155: same

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +167,127 @@ export const fetchSwapAmounts = async ({
...
+			await depositFrom({
+				identity,
+				canisterId: pool.canisterId.toString(),
+				token: sourceToken.ledgerCanisterId,
+				amount: parsedSwapAmount,
+				fee: sourceTokenFee
+			});
+		}
+	} catch (_: unknown) {
+		throw new Error(get(i18n).swap.error.deposit_error);
```
AntonioVentilii, line 242: no need to raise this error i believe, each of those functions will raise their specific error, and it is important to receive the specific feedback from them, to see what went wrong

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +167,127 @@ export const fetchSwapAmounts = async ({
...
+				canisterId: pool.canisterId.toString(),
+				token: sourceToken.ledgerCanisterId,
+				amount: parsedSwapAmount,
+				fee: sourceTokenFee
+			});
+		} else {
+			await approve({
+				identity,
+				ledgerCanisterId: sourceToken.ledgerCanisterId,
+				amount: parsedSwapAmount + sourceTokenFee * 2n,
```
AntonioVentilii, line 228: maybe add an explanation here for posterity: why double?

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +167,127 @@ export const fetchSwapAmounts = async ({
...
+				expiresAt: nowInBigIntNanoSeconds() + 5n * NANO_SECONDS_IN_MINUTE,
```
AntonioVentilii, line 229: same, why 5 minutes?

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +167,127 @@ export const fetchSwapAmounts = async ({
...
+	const slippageMinimum = calculateSlippage({
+		quoteAmount: receiveAmount,
+		slippagePercentage: Number(slippageValue)
+	});
+
+	const transferParams = {
+		identity,
+		token: sourceToken,
+		amount: parsedSwapAmount,
+		to: pool.canisterId.toString(),
```
AntonioVentilii, line 210: just to be double sure, is this guaranteed to return a correct canister type? is it worth adding a zod parser using `CanisterIdTextSchema` and raise an error in case? I imagine so, since this is 3rd-party-dependent.

And i see that you use it in the other functions too, so it can be made a separate variable

AntonioVentilii (commented):
Let's add some explanation for each step: the reason why something is done independently and something is handled, and something else triggered only in specific cases.

Plus, possibly, if it exists, some documentation link?

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: No, it is called in the component after successfull swap

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Yes, as discussed, we perform two types of withdrawals depending on the outcome of the swap: if the swap succeeds, we withdraw the destination token; if it fails, we withdraw the source token as a refund.

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Yes, as discussed, we perform depending on the outcome of the swap

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Added

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Added comment

src/frontend/src/lib/services/swap.services.ts
AntonioVentilii: yes, but the service should be independent from the component, since it can be called in different places. So the checks should be here too (or only here, and not in the component)

src/frontend/src/tests/lib/services/icp-swap.services.spec.ts
roman-nazaruk: Fixed

src/frontend/src/tests/lib/services/icp-swap.services.spec.ts
roman-nazaruk: Fixed

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Added comment

src/frontend/src/lib/services/swap.services.ts
AntonioVentilii: all good then! but create a variable maybe? so you don't call `toString()` several times

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +169,144 @@ export const fetchSwapAmounts = async ({
...
+				amount: parsedSwapAmount + sourceTokenFee * 2n,
+				// Sets approve expiration to 5 minutes ahead to allow enough time for the full swap flow
+				expiresAt: nowInBigIntNanoSeconds() + 5n * NANO_SECONDS_IN_MINUTE,
+				spender: { owner: pool.canisterId }
+			});
+
+			await depositFrom({
+				identity,
+				canisterId: pool.canisterId.toString(),
+				token: sourceToken.ledgerCanisterId,
```
AntonioVentilii, line 244:
```suggestion
				token: sourceLedgerCanisterId,
```

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +169,144 @@ export const fetchSwapAmounts = async ({
...
+				amount: parsedSwapAmount,
+				fee: sourceTokenFee
+			});
+		} else {
+			await approve({
+				identity,
+				ledgerCanisterId: sourceToken.ledgerCanisterId,
```
AntonioVentilii, line 233:
```suggestion
				ledgerCanisterId: sourceLedgerCanisterId,
```

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +169,144 @@ export const fetchSwapAmounts = async ({
...
+		ledgerCanisterId: sourceToken.ledgerCanisterId
+	};
+
+	try {
+		if (!isSourceTokenIcrc2) {
+			await sendIcrc(transferParams);
+			await deposit({
+				identity,
+				canisterId: pool.canisterId.toString(),
+				token: sourceToken.ledgerCanisterId,
```
AntonioVentilii, line 226:
```suggestion
				token: sourceLedgerCanisterId,
```

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +169,144 @@ export const fetchSwapAmounts = async ({
...
+		quoteAmount: receiveAmount,
+		slippagePercentage: Number(slippageValue)
+	});
+
+	const transferParams = {
+		identity,
+		token: sourceToken,
+		amount: parsedSwapAmount,
+		to: pool.canisterId.toString(),
+		ledgerCanisterId: sourceToken.ledgerCanisterId
```
AntonioVentilii, line 217:
```suggestion
		ledgerCanisterId: sourceLedgerCanisterId
```

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +169,144 @@ export const fetchSwapAmounts = async ({
...
+		});
+	} catch (err: unknown) {
+		console.error(err);
+
+		try {
+			// If the swap fails, attempt to refund the user's original tokens
+			await withdraw({
+				identity,
+				canisterId: pool.canisterId.toString(),
+				token: sourceToken.ledgerCanisterId,
```
AntonioVentilii, line 271:
```suggestion
				token: sourceLedgerCanisterId,
```

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +169,144 @@ export const fetchSwapAmounts = async ({
...
+		throw new Error(get(i18n).swap.error.deposit_error);
+	}
+
+	try {
+		// Perform the actual token swap after a successful deposit
+		await swapIcp({
+			identity,
+			canisterId: pool.canisterId.toString(),
+			amountIn: parsedSwapAmount.toString(),
+			zeroForOne: pool.token0.address === sourceToken.ledgerCanisterId,
```
AntonioVentilii, line 260:
```suggestion
			zeroForOne: pool.token0.address === sourceLedgerCanisterId,
```

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -175,3 +169,144 @@ export const fetchSwapAmounts = async ({
...
+		// Inform the user that the swap failed, but refund was successful
+		throw new Error(get(i18n).swap.error.swap_failed_withdraw_success);
+	}
+
+	try {
+		// Withdraw the swapped destination tokens from the pool
+		await withdraw({
+			identity,
+			canisterId: pool.canisterId.toString(),
+			token: destinationToken.ledgerCanisterId,
```
AntonioVentilii, line 289:
```suggestion
			token: destinationLedgerCanisterId,
```

src/frontend/src/lib/services/swap.services.ts
AntonioVentilii: but we can cross that bridge when the time come

AntonioVentilii (approved):
LGTM tks!

PR #6649, opened 2025-05-16
feat(frontend): define default token for Polygon networks

AntonioVentilii:
# Motivation

As for the other networks, we need to define the default token of Polygon networks.

# Changes

- Create util to check id a network ID is Polygon.
- Create derived store to verify the current network if it's Polygon.
- Map the current network to POL default token if it's Polygon network.

# Tests

More test and current tests will be sufficient.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6650, opened 2025-05-16
feat(frontend): create derived store for enabled Polygon tokens

AntonioVentilii:
# Motivation

Similar to the other networks, we create a derived store with all the enabled Polygon tokens.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6653, opened 2025-05-16
feat(frontend): improve ConvertAmountExchange component

DenysKarmazynDFINITY:
# Motivation

We need to improve the ConvertAmountExchange component so it can handle the following cases:
1. If `usdValue` is 0, we display it as `$0.00`.
2. If `usdValue` is less than 0.01, we display it as `< $0.00`.
3. If `usdValue` is greater than 0.01, we display it as `~$0.03`.


AntonioVentilii (approved):
LGTM Tks!

PR #6656, opened 2025-05-16
feat(frontend): use provided destination as address picker filter

DenysKarmazynDFINITY:
# Motivation

We want to filter known destinations when user starts typing an address.


AntonioVentilii (approved):
LGTM tks

PR #6657, opened 2025-05-16
feat(backend)!: Create Contact API change

SYInacta:
# Motivation

This PR introduces a new "Contacts" feature to the wallet backend. The motivation is to provide users with an address book functionality, allowing them to securely store and manage addresses (Token Account IDs) of frequently used contacts or services directly within their wallet profile. This enhances usability by making it easier to send funds to known recipients without needing to re-enter addresses repeatedly.

BREAKING CHANGE: The backend API is now incompatible with production.'

# Changes

This PR includes the following changes:

1.  **Updated Contact Data Structures:** Defined new Candid types (`Contact`, `ContactAddressData`, `ContactSettings`, `CreateContactRequest`, `RemoveContactRequest`, `UpdateContactRequest`,  `ContactError`) in `shared/src/types/contact.rs` to represent contact information and the associated request/response structures.

2.  **Exposed Contact API Endpoints:** Added new update and query methods to `src/backend/src/lib.rs` that expose the contact management functionality via the canister's public interface. 
    *   `create_contact`

5.  **Updated Candid Interface:** Generated the updated Candid definition reflect the new contact management methods and types.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,7 @@
 				"zod": "^3.23.8"
 			},
 			"devDependencies": {
-				"@dfinity/eslint-config-oisy-wallet": "^0.1.10",
+				"@dfinity/eslint-config-oisy-wallet": "^0.1.12",
 				"@dfinity/ic-management": "^6.1.1",
 				"@dfinity/identity-secp256k1": "^2.4.0",
 				"@dfinity/internet-identity-playwright": "^0.0.4",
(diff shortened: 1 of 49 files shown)


peterpeterparker (changes requested):
I randomly noticed this PR, which seems to contain zillion of unrelated changes in regards to its description. Thought I'd block it before it's too late - especially with the weekend coming up.

peterpeterparker (approved):
Ah, my bad - you want to merge main into your branch. Is that a new pattern?

Approving to unblock you then. Sorry for the misunderstanding - I haven’t seen this approach used at the foundation before. I’d suggest not using a PR for this purpose; it can be confusing, as you’ve probably noticed.

peterpeterparker (changes requested):
Actually, I’m quite confused about the goal of this PR given its description. What’s the real objective here?

PR #6658, opened 2025-05-16
feat(frontend): add a flag to check if `AllTransactionsLoader` is destroyed

AntonioVentilii:
# Motivation

To add another check to `AllTransactionsLoader`, we can make it stop loading transactions as soon as the component is destroyed.


src/frontend/src/lib/components/transactions/AllTransactionsLoader.svelte
```diff
@@ -25,6 +25,12 @@
 
 	let disableLoader: Record<TokenId, boolean> = $state({});
 
+	let destroyed = $state(false);
+
+	onDestroy(() => {
+		destroyed = true;
```
DenysKarmazynDFINITY, line 31: Interesting, so without it the $effect with `loadMissingTransactions` keeps getting updated even after the component is destroyed? Or is it just to prevent it being called as early as possible, meaning that with the `onDestroy` approach the transactions fetching will be cancelled sooner rather than with waiting for the "normal" destroy to happen?

AntonioVentilii: exactly the second! It is just an additional layer of "sparing cycles and improving performance"

It is not much, but anything helps, right?

DenysKarmazynDFINITY: Yes! Thanks for the explanation.

DenysKarmazynDFINITY (approved):

PR #6660, opened 2025-05-16
feat(frontend): add ERC20 tokens for Polygon

AntonioVentilii:
# Motivation

The tokens for network Polygon have the same standard of ERC20 of the Ethereum network. So we can easily include them like the others. In this PR we introduce USDC and USDT on Polygon.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6661, opened 2025-05-16
refactor(frontend): generalize loading balance error message for EVMs

AntonioVentilii:
# Motivation

Since we now have multiple native tokens, it is better to make the error message of the loading balance more verbose: we add the symbol and the network.


src/frontend/src/eth/services/eth-balance.services.ts
```diff
@@ -55,7 +55,12 @@ const loadEthBalance = async ({
 		balancesStore.reset(tokenId);
 
 		toastsError({
-			msg: { text: loading_balance },
+			msg: {
+				text: replacePlaceholders(loading_balance, {
+					$symbol: tokenId.description ?? 'ETH',
```
DenysKarmazynDFINITY, line 60: Shall we use some of the consts here? ETHEREUM_NETWORK_SYMBOL for instance

DenysKarmazynDFINITY (approved):
Small q, LGTM otherwise

PR #6662, opened 2025-05-16
feat(frontend): add exchange rate for native token POL

AntonioVentilii:
# Motivation

Similar to what we did for BNB, we provide a custom price for token POL that is native to Polygon, so it does not have a contract to fetch the price, nor it is a bridge of ETH token.

Screenshot 2025-05-16 at 21 49 26


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6663, opened 2025-05-18
refactor(frontend): collect all EVM ERC20 tokens in a single list

AntonioVentilii:
# Motivation

All the current ERC20 tokens for EVM networks (and their wrapped standard like BEP20) are treated the same way. So, to improve manageability, we can collect them in a single list that can be updated accordingly.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6679, opened 2025-05-19
fix(frontend): fixes token hiding process

BonomoAlessandro:
# Motivation

When trying to hide a token, a browser warning appears that must be accepted. This should not be necessary to hide a token.

# Changes

- fixes `hiding` process

# Tests

**before:**

https://github.com/user-attachments/assets/f4315843-b31e-42a1-96d8-6f999fae60a1

**after:**

https://github.com/user-attachments/assets/1cf816c0-c80a-4c96-811c-ea484348ee9e


src/frontend/src/lib/components/tokens/HideTokenModal.svelte
```diff
@@ -53,9 +54,6 @@
 
 			hideProgressStep = ProgressStepsHideToken.UPDATE_UI;
 
-			// We must navigate first otherwise we might land on the default token Ethereum selected while being on network ICP.
```
AntonioVentilii, line 56: just to double check, this sentence is not correct anymore?

AntonioVentilii (approved):
LGTM tks!

PR #6681, opened 2025-05-19
feat(frontend): sort known destinations by timestamp

DenysKarmazynDFINITY:
# Motivation

We need to sort all known destinations by timestamp. To do that, we add `address` to the `knownDestinations` stores and sort the values before filtering them in the UI component.

### DIFF
--- a/src/frontend/src/tests/btc/derived/btc-transactions.derived.spec.ts
+++ b/src/frontend/src/tests/btc/derived/btc-transactions.derived.spec.ts
@@ -115,7 +115,8 @@ describe('btc-transactions.derived', () => {
 						value: data.value,
 						token: BTC_MAINNET_TOKEN
 					})),
-					timestamp: maxTimestamp
+					timestamp: maxTimestamp,
+					address: transactions[0].data.to?.[0]
 				}
 			});
 		});
--- a/src/frontend/src/tests/icp/derived/ic-transactions.derived.spec.ts
+++ b/src/frontend/src/tests/icp/derived/ic-transactions.derived.spec.ts
@@ -176,7 +176,8 @@ describe('ic-transactions.derived', () => {
 			expect(get(icKnownDestinations)).toEqual({
 				[transactions[0].data.to as string]: {
 					amounts: transactions.map(({ data }) => ({ value: data.value, token: ICP_TOKEN })),
-					timestamp: maxTimestamp
+					timestamp: maxTimestamp,
+					address: transactions[0].data.to
 				}
 			});
 		});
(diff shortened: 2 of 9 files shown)


src/frontend/src/tests/sol/derived/sol-transactions.derived.spec.ts
```diff
@@ -141,7 +141,8 @@ describe('sol-transactions.derived', () => {
 			expect(get(solKnownDestinations)).toEqual({
 				[transactions[0].data.to as string]: {
 					amounts: transactions.map(({ data }) => ({ value: data.value, token: SOLANA_TOKEN })),
-					timestamp: maxTimestamp
+					timestamp: maxTimestamp,
+					address: transactions[0].data.to
```
AntonioVentilii, line 145: theoretically it should be the same address of maxTimestamp, right?

src/frontend/src/tests/icp/derived/ic-transactions.derived.spec.ts
```diff
@@ -176,7 +176,8 @@ describe('ic-transactions.derived', () => {
 			expect(get(icKnownDestinations)).toEqual({
 				[transactions[0].data.to as string]: {
 					amounts: transactions.map(({ data }) => ({ value: data.value, token: ICP_TOKEN })),
-					timestamp: maxTimestamp
+					timestamp: maxTimestamp,
+					address: transactions[0].data.to
```
AntonioVentilii, line 180: theoretically it should be the same address of maxTimestamp, right?same?

src/frontend/src/tests/btc/derived/btc-transactions.derived.spec.ts
```diff
@@ -115,7 +115,8 @@ describe('btc-transactions.derived', () => {
 						value: data.value,
 						token: BTC_MAINNET_TOKEN
 					})),
-					timestamp: maxTimestamp
+					timestamp: maxTimestamp,
+					address: transactions[0].data.to?.[0]
```
AntonioVentilii, line 119: same?

AntonioVentilii (approved):
LGTM tks!

A doubt about tests

src/frontend/src/tests/sol/derived/sol-transactions.derived.spec.ts
DenysKarmazynDFINITY: There is no particular logic around the `address` field - the store is grouped by unique destinations, so that field will always match the key. Therefore, we can simply take the `to` value here.

PR #6682, opened 2025-05-19
refactor(frontend): upgrades InputCurrency to svelte 5

BonomoAlessandro:
# Motivation

# Changes

- Upgrades `InputCurrency`


src/frontend/src/lib/components/ui/InputCurrency.svelte
```diff
@@ -3,16 +3,29 @@
...
+		disabled,
+		name,
+		placeholder,
+		required = true,
+		decimals,
+		testId,
+		autofocus = false
+	}: Props = $props();
+
+	let inputElement: HTMLInputElement | undefined = $state();
```
AntonioVentilii, line 28:
```suggestion
	let inputElement = $state<HTMLInputElement | undefined>();
```

AntonioVentilii (approved):
LGTM tks!

AntonioVentilii (changes requested):
Sorry, too hasty: there is a slot that should be integrated in the props

BonomoAlessandro:
@AntonioVentilii I am aware that there is still a `slot`. My idea was to first upgrade the components and then in a second step, when also the components around are upgraded, also update the `slot` definitions.

PR #6683, opened 2025-05-19
feat(frontend): clear failedSwapError on modal back step

roman-nazaruk:
# Motivation

We want to clear the ErrorMessage when clicking the back button in the Swap Modal to ensure that the error does not persist and appear again in the Review state.

# Changes

Added `set(undefined)` to `failedSwapError `on back button click to reset the error state and prevent it from displaying again in the Review state of the Swap Modal.


AntonioVentilii (approved):
LGTM tks! One obs:

PR #6686, opened 2025-05-19
feat(frontend): autofocus on desktop for SendInputDestination

DenysKarmazynDFINITY:
# Motivation

We need to set focus on the input on desktop when user opens the destination wizard step.


src/frontend/src/lib/components/send/SendInputDestination.svelte
```diff
@@ -35,6 +36,7 @@
 		bind:value={destination}
 		placeholder={inputPlaceholder}
 		testId={DESTINATION_INPUT}
+		autofocus={isDesktop()}
```
AntonioVentilii, line 39: is this reactive? meaning, in a responsive mode, does it work?

AntonioVentilii (approved):

PR #6691, opened 2025-05-19
feat(frontend): replace manual slippage formatting with formatReceiveOutMinimum

roman-nazaruk:
# Motivation

The logic for calculating and formatting the minimum receive amount with slippage was previously handled inline within components. This led to code duplication and reduced clarity. The goal of this change is to centralize that logic using a utility function (`formatReceiveOutMinimum`) to improve readability, reusability, and testability.

# Changes

- Replaced inline slippage calculation and formatting in ICP swap view with `formatReceiveOutMinimum` utility
- Reduced duplication and improved separation of concerns between logic and UI

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapDetailsIcp.svelte
+++ b/src/frontend/src/lib/components/swap/SwapDetailsIcp.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { isNullish, nonNullish } from '@dfinity/utils';
+	import { nonNullish } from '@dfinity/utils';
 	import { getContext } from 'svelte';
 	import ModalValue from '$lib/components/ui/ModalValue.svelte';
 	import { i18n } from '$lib/stores/i18n.store';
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/swap/SwapDetailsIcp.svelte
```diff
@@ -4,28 +4,36 @@
...
+	import type { OptionAmount } from '$lib/types/send';
 	import type { SwapMappedResult, SwapProvider } from '$lib/types/swap';
-	import { formatToken } from '$lib/utils/format.utils';
+	import { formatReceiveOutMinimum } from '$lib/utils/swap.utils';
 
-	const { provider } = $props<{
+	interface Props {
 		provider: Extract<SwapMappedResult, { provider: SwapProvider.ICP_SWAP }>;
-	}>();
+		slippageValue: OptionAmount;
```
AntonioVentilii, line 13: since it can be undefined, even optional? or better not? it doesn't make sense, right?

in case yes, change it for the other components too pls

```suggestion
		slippageValue?: OptionAmount;
```

AntonioVentilii (approved):
LGTM tks! I have a doubt:

src/frontend/src/lib/components/swap/SwapDetailsIcp.svelte
roman-nazaruk: it doesn't make any sense as OptionAmount is
```
export type Amount = string | number;
export type OptionAmount = Amount | undefined;
```

PR #6693, opened 2025-05-19
feat(frontend): Contacts API

daviddecentage:
# Motivation

Adding the TS API for the contacts services.

# Changes

- Added params TS types for API methods
- Added backend.canister methods
- Added backend.api methods


BonomoAlessandro (commented):
please provide some tests

roman-nazaruk:
Could you please split this into two separate PRs — one for the canister logic with its tests, and another for the API? Thanks.

src/frontend/src/lib/api/backend.api.ts
```diff
@@ -192,6 +199,45 @@ export const updateUserNetworkSettings = async ({
 	return updateUserNetworkSettings(params);
 };
 
+export const getContact = async ({
+	contactId,
+	identity
+}: CanisterApiFunctionParams<GetContactParams>): Promise<GetContactResult> => {
```
BonomoAlessandro, line 205: This in completely fine but if you don't want to create a new type for each API function you can implement it like this:

`export const getContact = async ({
	contactId,
	identity
}: CanisterApiFunctionParams<{contactId: bigint}>): Promise<GetContactResult> => {`

BonomoAlessandro (approved):
LGTM

PR #6697, opened 2025-05-19
feat(frontend): updated images for banner loader

artkorotkikh-dfinity:
# Motivation

We'd like to show Polygon network logo when the app is loading

# Changes

Replaced two images for banner asset 

# Tests

image

image


AntonioVentilii (approved):
Thank you!!!

PR #6698, opened 2025-05-19
feat(frontend): updated main image with Polygon asset

artkorotkikh-dfinity:
# Motivation

We'd like to show Polygon token and network on the main image when a user is not signed in 

# Changes

Updated 2 images (light / dark mode) 

# Tests

image

image


AntonioVentilii (approved):
Thank you!!

PR #6699, opened 2025-05-20
refactor(frontend): rename Polygon network

AntonioVentilii:
# Motivation

It is most common to call the Polygon network without the `PoS` suffix.


BonomoAlessandro (approved):
LGTM

PR #6702, opened 2025-05-20
fix(frontend): Correct mapping for Polygon ERC20 exchange rate

AntonioVentilii:
# Motivation

To correctly map the ERC20 prices for Polygon, there is a check missing in the exchange worker.

Screenshot 2025-05-20 at 08 51 27


BonomoAlessandro (approved):
LGTM

PR #6710, opened 2025-05-20
refactor(frontend): make networkId optional in isInvalidDestinationIc

DenysKarmazynDFINITY:
# Motivation

We want to clean up the Send flow from the convert-related things. To do that, we need to make networkId optional in isInvalidDestinationIc. 

Also, added missing tests.


AntonioVentilii (approved):
LGTM tks! Obs:

PR #6711, opened 2025-05-20
refactor(frontend): make targetNetwork optional in sendIc

DenysKarmazynDFINITY:
# Motivation

We want to clean up the Send flow from the convert-related things. To do that, we need to make targetNetwork optional in sendIc.

### DIFF
--- a/src/frontend/src/icp/services/ic-send.services.ts
+++ b/src/frontend/src/icp/services/ic-send.services.ts
@@ -35,7 +35,7 @@ export const sendIc = async ({
 	...rest
 }: IcTransferParams & {
 	token: IcToken;
-	targetNetworkId: NetworkId | undefined;
+	targetNetworkId?: NetworkId;
 	sendCompleted: () => void;
 }): Promise<void> => {
 	await send({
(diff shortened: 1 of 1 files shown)


src/frontend/src/icp/services/ic-send.services.ts
```diff
@@ -35,7 +35,7 @@ export const sendIc = async ({
 	...rest
 }: IcTransferParams & {
 	token: IcToken;
-	targetNetworkId: NetworkId | undefined;
+	targetNetworkId?: NetworkId | undefined;
```
AntonioVentilii, line 38: shall it be `targetNetworkId?: NetworkId;`?

src/frontend/src/icp/services/ic-send.services.ts
```diff
@@ -56,7 +56,7 @@ const send = async ({
...
```
AntonioVentilii, line 59: same

AntonioVentilii (approved):
LGTM tks! Obs:

src/frontend/src/icp/services/ic-send.services.ts
DenysKarmazynDFINITY: yes, thanks! updated

PR #6716, opened 2025-05-20
feat(frontend): Add IcpAccountSchema

FlavioInacta:
# Motivation

In order to be able to parse ICP adresses, this PR adds a `IcpAddressSchema`

# Changes

- Adds `IcpAddressSchema`


src/frontend/src/icp/utils/icp-account.utils.ts
```diff
@@ -1,5 +1,56 @@
...
 
 export const getAccountIdentifier = (principal: Principal): AccountIdentifier =>
 	AccountIdentifier.fromPrincipal({ principal, subAccount: undefined });
+
+/**
+ * Parses a string into an Icrcv2AccountId object
+ * @param address The address string to parse
+ * @returns The parsed Icrcv2AccountId or undefined if parsing fails
+ */
+export const parseIcpAccountId = (address: string): Icrcv2AccountId | undefined => {
```
AntonioVentilii, line 15: let's call it something else, since it refers to a Principal structure for the backend, not necessarily to an ICP account that we use in the frontend (see function `isIcpAccountIdentifier`)

src/frontend/src/lib/schema/address.schema.ts
```diff
@@ -26,3 +27,25 @@ export const BtcAddressObjectSchema = BtcAddressSchema.transform<BtcAddress>((da
 
 	return z.NEVER;
 }).transform<BtcAddress>((v) => v);
+
+export const Icrcv2AccountIdSchema = AddressSchema.refine((v) => !!parseIcpAccountId(v));
```
AntonioVentilii, line 31: same

src/frontend/src/icp/utils/icp-account.utils.ts
FlavioInacta: Ok, how?

src/frontend/src/icp/utils/icp-account.utils.ts
```diff
@@ -1,5 +1,56 @@
...
+			Account: new TextEncoder().encode(address)
+		};
+	}
+
+	try {
+		const decoded = decodeIcrcAccount(address);
+		return {
+			WithPrincipal: {
+				owner: decoded.owner,
+				subaccount: decoded.subaccount ? [decoded.subaccount] : []
```
AntonioVentilii, line 27:
```suggestion
				subaccount: toNullable(decoded.subaccount)
```

src/frontend/src/icp/utils/icp-account.utils.ts
```diff
@@ -1,5 +1,56 @@
...
+};
+
+/**
+ * Extracts the address string from an Icrcv2AccountId object
+ * @param accountId The Icrcv2AccountId object
+ * @returns The address string or undefined if extraction fails
+ */
+export const getIcpAccountIdString = (accountId: Icrcv2AccountId): string => {
+	if ('Account' in accountId) {
+		return new TextDecoder().decode(new Uint8Array(accountId.Account));
```
AntonioVentilii, line 42: are you sure this is the correct decoding? maybe here you can use `Principal.fromUint8Array` and then `toString`

src/frontend/src/icp/utils/icp-account.utils.ts
```diff
@@ -1,5 +1,56 @@
...
+	}
+
+	if ('WithPrincipal' in accountId) {
+		const { owner, subaccount } = accountId.WithPrincipal;
+		const subaccountArray = subaccount.length > 0 ? subaccount[0] : undefined;
+
+		return encodeIcrcAccount({
+			owner,
+			subaccount: subaccountArray
+		});
```
AntonioVentilii, line 52:
```suggestion
		return encodeIcrcAccount({
			owner,
			subaccount: fromNullable(subaccount)
		});
```

src/frontend/src/icp/utils/icp-account.utils.ts
```diff
@@ -1,5 +1,56 @@
...
+
+/**
+ * Parses a string into an Icrcv2AccountId object
+ * @param address The address string to parse
+ * @returns The parsed Icrcv2AccountId or undefined if parsing fails
+ */
+export const parseIcpAccountId = (address: string): Icrcv2AccountId | undefined => {
+	if (isIcpAccountIdentifier(address)) {
+		return {
+			Account: new TextEncoder().encode(address)
```
AntonioVentilii, line 18: is this decoding necessary?

src/frontend/src/icp/utils/icp-account.utils.ts
bitdivine: Looking at the Rust code, Icrcv2AccountId can be either just a subaccountID (as used natively in ledgers) or principal and optional subaccount ID:
```
pub enum Icrcv2AccountId {
    /// An account identifier that declares the owner's principal.
    ///
    /// Note: Given a transaction, it is NOT typically possible to determine the owner
    /// principal & subaccount ID.  A such, this type is primarily used for internal management
    /// of accounts within a wallet.  It is not normally available for transactions with third
    /// parties.
    WithPrincipal {
        owner: Principal,
        subaccount: Option<IcrcSubaccountId>,
    },
    /// This is a redacted identifier typically available from transaction records.
    Account(IcrcSubaccountId),
}
```
The string can presumably be any of:
- A string subaccountid (64 hex chars)
- A principal (variable length, maximum about 29 - I need to check)
- A principal + subaccount ID (long - but I always struggle to find the link to where the combined form is specified.)

So, I guess you need to decide which string formats you want to support (all?), check the string length and convert accordingly.

bitdivine: @AntonioVentilii Do you remember where the combined principal+subaccount string format is specified?

FlavioInacta: I think the "WithPrincipal" should be correct. For the Account I saw how they decoded/encoded it here: https://github.com/dfinity/ic-js/blob/main/packages/ledger-icp/src/utils/account_identifier.utils.ts

And I tested it with the update contact backend endpoint and there are no validation errors anymore. Should be fine now.

src/frontend/src/icp/utils/icp-account.utils.ts
FlavioInacta: No, it is not. Found the solution here: https://github.com/dfinity/ic-js/blob/main/packages/ledger-icp/src/utils/account_identifier.utils.ts

src/frontend/src/icp/utils/icp-account.utils.ts
FlavioInacta: Done, thx.

src/frontend/src/lib/schema/address.schema.ts
```diff
@@ -26,3 +27,25 @@ export const BtcAddressObjectSchema = BtcAddressSchema.transform<BtcAddress>((da
...
+
+export const Icrcv2AccountIdObjectSchema = Icrcv2AccountIdSchema.transform<Icrcv2AccountId>(
+	// eslint-disable-next-line local-rules/prefer-object-params
+	(data, context): Icrcv2AccountId => {
+		const accountId = parseIcpAccountId(data);
+
+		if (accountId) {
```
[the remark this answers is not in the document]
FlavioInacta: Done

src/frontend/src/icp/utils/icp-account.utils.ts
FlavioInacta: Done, thx.

src/frontend/src/icp/utils/icp-account.utils.ts
FlavioInacta: Corrected it to: Buffer.from(accountId.Account).toString('hex');

src/frontend/src/icp/utils/icp-account.utils.ts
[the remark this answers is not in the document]
AntonioVentilii: @FlavioInacta did you get a chance to see this one? We really should use the types that are already used through the code. Would that be possible in this case?

AntonioVentilii (approved):
I am going to approve, but if these functions are going to be used to parse data from the backend to the frontend (as I imagine they will), you need to return a `Principal` type (like we do in the rest of the frontend)

So you will have to change all the return types here and the parsing too, before any other PR.

However, if these functions are used to parse a `Principal` from the frontend to the type `Icrcv2AccountId` of the backend, then all good

FlavioInacta:
Yes, they are used to parse string to `Icrcv2AccountId` in the FE. All good then. :)

PR #6719, opened 2025-05-20
feat(frontend): Adapting snapshots to new structure of Rewards canister

AntonioVentilii:
# Motivation

The Rewards canister now accepts any kind of token/network. So, we adapt the service to the new interface.

To do it in a cleaner way, we first create parsing/mapping functions that are a merged version of the existing ones.

# Changes

- Create transaction mapper for new RC type `Transaction_Any`. It is done merging the existing ones for `Transaction_Icrc` and `Transaction_Spl`.
- Create last transactions selector, using the same logic for the existing functions, for all networks.
- Create snapshot mapper for new RC type `AccountSnapshot_Any`. It is done merging the existing ones for `AccountSnapshot_Icrc` and `AccountSnapshot_Spl`.

# Note

I will remove the old functions in another PR, so that it facilitate the review of this PR.

# Tests

Adapted tests.


src/frontend/src/lib/services/user-snapshot.services.ts
```diff
@@ -151,6 +193,130 @@ const toBaseSnapshot = ({
 	network: {}
 });
 
+const getLastTransactionsByToken = ({
```
AntonioVentilii, line 196: this function is a copy of the existing ones

src/frontend/src/lib/services/user-snapshot.services.ts
```diff
@@ -93,6 +101,40 @@ const toBaseTransaction = ({
...
+
+	if (isNullish(timestamp)) {
+		return;
+	}
+
+	if (isNullish(from) || isNullish(to)) {
+		return;
+	}
+
+	const counterparty: AccountId_Any = account === from ? (Array.isArray(to) ? to[0] : to) : from;
```
DenysKarmazynDFINITY, line 125: If one day we implement sending BTC to multiple addresses, we'll need to revisit this logic. But for now it's completely accurate 🙂

src/frontend/src/lib/services/user-snapshot.services.ts
```diff
@@ -151,6 +193,130 @@ const toBaseSnapshot = ({
...
+	const ckEthMinterInfoAddressesMainnet = toCkMinterInfoAddresses(
+		get(ckEthMinterInfoStore)?.[ETHEREUM_TOKEN_ID]
+	);
+	// If there are no minter info addresses, we cannot map the transactions. We return undefined to skip this token.
+	// Since we are sending snapshots at several intervals and refreshes, it is not necessary to raise an error.
+	if (isEthOrEvm && isNullish(ckEthMinterInfoAddressesMainnet)) {
+		return [];
+	}
+	const ckEthMinterInfoAddressesSepolia = toCkMinterInfoAddresses(
+		get(ckEthMinterInfoStore)?.[SEPOLIA_TOKEN_ID]
```
DenysKarmazynDFINITY, line 219: Should we add the `isNetworkIdSepolia` check to return early only when a required for the token minter is not available?

DenysKarmazynDFINITY (approved):
LGTM, thanks!

bitdivine (approved):
This looks good to me but would like to test on e.g. staging to make sure it sends the expected data to the rewards canister.

PR #6722, opened 2025-05-20
feat(frontend): enables to set Tags to ButtonMenu

BonomoAlessandro:
# Motivation

We want to be able to set `Tags` to the Menu items.

# Changes

- supports `Tags`

# Tests

image


daviddecentage (changes requested):
Since you upgrade the other components to svelte 5, can you do this one as well?

BonomoAlessandro:
@daviddecentage The upgrade of this component will come with a separate PR

daviddecentage (approved):
LGTM, thx

PR #6723, opened 2025-05-20
refactor(frontend): upgrades Tag to svelte 5

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/components/ui/Tag.svelte
+++ b/src/frontend/src/lib/components/ui/Tag.svelte
@@ -1,7 +1,15 @@
 <script lang="ts">
 	import type { TagVariant } from '$lib/types/style';
+	import type {Snippet} from "svelte";
 
-	export let ariaHidden = false;
+	interface Props {
+		children?: Snippet;
+		ariaHidden?: boolean;
+		variant?: TagVariant;
+		size?: 'xs' | 'sm' | 'md';
+	}
+
+	let {children, ariaHidden = false, variant = 'default', size = 'xs'}: Props = $props();
 
 	const variantClassNames = {
 		default: 'bg-primary text-primary border border-secondary',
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/ui/Tag.svelte
```diff
@@ -30,5 +35,5 @@
 	class:py-0.5={size === 'sm' || size === 'md'}
 	class:md:py-1={size === 'sm' || size === 'md'}
 >
-	<slot />
+	{@render children?.()}
```
daviddecentage, line 38: Just checked, usually we do
```suggestion
	{#if nonNullish(children)}
		{@render children()}
	{/if}
```

src/frontend/src/lib/components/ui/Tag.svelte
```diff
@@ -1,7 +1,15 @@
 <script lang="ts">
+	import type { Snippet } from 'svelte';
 	import type { TagVariant } from '$lib/types/style';
 
-	export let ariaHidden = false;
+	interface Props {
+		children?: Snippet;
```
AntonioVentilii, line 6: maybe `children` is not optional?

AntonioVentilii (approved):
LGTM tks! One doubt:

src/frontend/src/lib/components/ui/Tag.svelte
BonomoAlessandro: @daviddecentage TBH i think my implementation is cleaner. Also its the implementation that is recomended in the Migration Guide

daviddecentage: Ok then

daviddecentage (approved):
LGTM

PR #6726, opened 2025-05-20
feat(frontend): Contact store

daviddecentage:
# Motivation

We need a store for contact data and integrate it in the services to work with contacts in the address book.

# Changes

- Added contact store
- Use contact store in contact service
- Added loader to initialize contacts store
- Added mappng methods to utils

# Tests

Added tests for utils functions


src/frontend/src/lib/stores/contacts.store.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const initContactsStore = (): ContactsStore => {
+	const { set, update, subscribe } = writable<ContactsStoreData>(undefined);
+
+	const reset = () => {
+		set(undefined);
+	};
+
+	const addContact = (contact: ContactUi) => {
+		update((contacts) => [...contacts!, contact]);
```
FlavioInacta, line 21: Ensure that contacts are not `undefined`. Remove !

src/frontend/src/lib/stores/contacts.store.ts
```diff
@@ -0,0 +1,43 @@
...
+	};
+
+	const updateContact = (contact: ContactUi) => {
+		update((contacts) => contacts!.map((c) => (c.id === contact.id ? contact : c)));
```
FlavioInacta, line 25: Remove !

src/frontend/src/lib/stores/contacts.store.ts
```diff
@@ -0,0 +1,43 @@
...
+	};
+
+	const removeContact = (id: ContactUi['id']) => {
+		update((contacts) => contacts!.filter((c) => c.id !== id));
```
FlavioInacta, line 29: Remove !

daviddecentage: Updated

src/frontend/src/lib/stores/contacts.store.ts
daviddecentage: Updated

src/frontend/src/lib/services/manage-contacts.service.ts
```diff
@@ -0,0 +1,76 @@
...
+	throw result.Err;
+};
+
+export const deleteContact = async ({
+	id,
+	identity
+}: {
+	id: ContactUi['id'];
+	identity: Identity;
+}): Promise<bigint> => {
```
FlavioInacta, line 67: I don't know what this BigInt is. Do you know?
Otherwise remove.

src/frontend/src/lib/services/manage-contacts.service.ts
```diff
@@ -0,0 +1,76 @@
...
+	const result = await deleteContactApi({ contactId: BigInt(id), identity });
```
FlavioInacta, line 68:
```suggestion
	const result = await deleteContactApi({ contactId: id, identity });
```

src/frontend/src/lib/stores/contacts.store.ts
```diff
@@ -0,0 +1,59 @@
...
+			return [...contacts, contact];
+		});
+	};
+
+	const updateContact = (contact: ContactUi) => {
+		update((contacts) => {
+			if (isNullish(contacts)) {
+				throw new Error('Contacts store is not initialized - cannot update contact');
+			}
+			return contacts.map((c) => (c.id === contact.id ? contact : c));
```
FlavioInacta, line 35: We should probably make a deep clone. Otherwise the contacts can accidently be changed from the outside.

Hm, on the other hand. The objects that go out, could potentially be changed too. Maybe svelte store is doing some magic to do this? idk.

FlavioInacta (approved):
For me basically ok, but I don't know what the requirements for stores in general is and if cloning of incoming and outgoing objects is needed.

src/frontend/src/lib/services/manage-contacts.service.ts
daviddecentage: updated

src/frontend/src/lib/stores/contacts.store.ts
daviddecentage: Very good point indeed. I create a copy of the received object to ensure that not the reference is saved in store

PR #6727, opened 2025-05-20
refactor(frontend): sets default value for hideBalance parameter

BonomoAlessandro:
# Motivation

The optional boolean parameter `hideBalance` should have a default value that is not `undefined.`

# Changes

- sets default value


AntonioVentilii (approved):
Sadly not the default value I wished 😏, but LGTM tks!

PR #6728, opened 2025-05-20
feat(frontend): Remove toast error during IC loading transactions

AntonioVentilii:
# Motivation

We don't want to show anymore the toast error about the loading IC transactions failing. However, we want to keep track of it.

# Changes

- Removing the toast error from service `onLoadTransactionsError`.
- Remove unnecessary `silent` flag: it will always be `true` otherwise.

# Tests

Adapted tests.


roman-nazaruk (approved):
LGTM, thnx

PR #6729, opened 2025-05-20
feat(frontend): Add TokenAccountIdSchema

FlavioInacta:
# Motivation

Adds `TokenAccountIdSchema` to parse addresses into backend objects and convert them back to string representation

# Changes

- Adds union schema for the different address types


daviddecentage (approved):
Small change, otherwise LGTM, thx

PR #6733, opened 2025-05-20
feat(frontend): Remove toast error during ETH loading balances

AntonioVentilii:
# Motivation

We don't want to show anymore the toast error about the loading ETH balances failing. However, we want to keep track of it.

# Changes

- Removing the toast error from services `loadEthBalances` and `loadErc20Balances`.
- Add console warn for the error.

# Tests

Adapted tests.


roman-nazaruk (approved):
LGTM, thnx

PR #6737, opened 2025-05-20
fix(frontend): HOTFIX: Disable reactivity for CK Ethereum pending transactions

AntonioVentilii:
# Motivation

We included the CK Ethereum pending transactions listener in the main page. However, this triggers too many calls. Just as HOTFIX, we disable the reactivity of the component for now. In the meanwhile we investigate.

NOTE: we simplified the reactivity too, since it had redundant elements.


DenysKarmazynDFINITY (approved):
Thanks!

PR #6742, opened 2025-05-21
refactor(frontend): replaces deprecated page from app/store

BonomoAlessandro:
# Motivation

Since `page` from `app/store` is deprecated and we upgraded the `NavigationMenuMainItems` component to svelte 5, we can use the `page` from `app/state` instead.

The usage of the two `page` should be the same: [read here](https://svelte.dev/docs/kit/migrating-to-sveltekit-2#SvelteKit-2.12:-$app-stores-deprecated)

# Changes

- uses page from `app/state`


AntonioVentilii (approved):

AntonioVentilii (approved):
LGTM tks!

PR #6752, opened 2025-05-21
refactor(frontend): improves implementation

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/components/receive/ReceiveAddresses.svelte
+++ b/src/frontend/src/lib/components/receive/ReceiveAddresses.svelte
@@ -275,9 +275,9 @@
 
 <ContentWithToolbar>
 	<div class="flex flex-col gap-2">
-		{#each receiveAddressList as { title, text, condition, on, labelRef, address, network, testId, copyAriaLabel, qrCodeAction } (labelRef)}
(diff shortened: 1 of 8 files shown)


src/frontend/src/lib/components/receive/ReceiveAddresses.svelte
```diff
@@ -275,9 +275,9 @@
 
 <ContentWithToolbar>
 	<div class="flex flex-col gap-2">
-		{#each receiveAddressList as { title, text, condition, on, labelRef, address, network, testId, copyAriaLabel, qrCodeAction } (labelRef)}
+		{#each receiveAddressList as { title: _title, text: _text, condition, on, labelRef, address, network, testId, copyAriaLabel, qrCodeAction } (labelRef)}
```
AntonioVentilii, line 278: just wondering, can you pass it directly, instead of snippet?

for example

```svelte
<ReceiveAddress
...
{title}
...
```

AntonioVentilii (approved):
LGTM tks! One doubt

src/frontend/src/lib/components/receive/ReceiveAddresses.svelte
BonomoAlessandro: I don't think so. This would work when `_title` is of type `Snippet` but it is a string.

PR #6753, opened 2025-05-21
fix(frontend): Add `polling` for Alchemy providers

AntonioVentilii:
# Motivation

As per thread https://github.com/ethers-io/ethers.js/issues/4784#issuecomment-2820177791:

> By default, etherjs uses RPC Filter and turns off polling. Filter relies on `eth_newFilter` / `eth_getFilterChanges`, both of which are state interfaces and need to store `filterId` status on the node. For service providers with distributed architectures such as Infura and QuickNode, each request may hit a different backend node, and these nodes do not share `filter` status. Therefore, they choose to disable `eth_newFilter` and `eth_getFilterChanges` and only support stateless requests (such as `eth_getLogs`) to ensure system stability and scalability.

This causes an extreme amount of console logs. Not errors, since it is expected according to https://github.com/ethers-io/ethers.js/issues/4796.

image-20250520-062407 (1)

So, as [suggested in the same thread](https://github.com/ethers-io/ethers.js/issues/4784#issuecomment-2273652405), we turn on `polling` for Alchemy JSON RPC providers.

### DIFF
--- a/src/frontend/src/eth/providers/alchemy-erc20.providers.ts
+++ b/src/frontend/src/eth/providers/alchemy-erc20.providers.ts
@@ -19,7 +19,10 @@ export class AlchemyErc20Provider {
 
 	// AlchemyProvider of ether.js does not support Sepolia
 	constructor(private readonly providerUrl: string) {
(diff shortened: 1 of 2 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6754, opened 2025-05-21
refactor(frontend): Move 1INCH token to additional ERC20 tokens

AntonioVentilii:
# Motivation

To start reducing the calls for metadata of ERC20 tokens, we can hard-code the remaining tokens, as we do for the ones that have CK twins. We start with 1INCH token.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6755, opened 2025-05-21
refactor(frontend): Specify network type for ERC20 tokens

AntonioVentilii:
# Motivation

It make sense to specify the type of Ethereum network for all ERC20 tokens.


bitdivine (approved):
I don't have all the context but this seems to make sense to me.

PR #6760, opened 2025-05-21
refactor(frontend): Handle responses contact api

daviddecentage:
# Motivation

We need to handle the response parsing on th canister service layer, not in the UI service.

# Changes

Refactored canister service layer and API service layer

# Tests

Tests have been adjusted accordingly


AntonioVentilii (approved):
LGTM tks

PR #6765, opened 2025-05-21
refactor(devops): No repeat downloads

bitdivine:
# Motivation
We download a lot of data with curl.  In some places we try to make this more efficient by not downloading a file if it already exists, however that breaks if the URL changes, giving strange results.  We have a script now that does this properly, so we can roll it out everywhere.

# Changes
- Use  `download-immutable` in most places where we used to curl.

# Tests
Existing tests should verify that existing functionality still works.


AntonioVentilii (approved):
LGTM tks!

PR #6766, opened 2025-05-21
feat(frontend): update ETH fees logic

DenysKarmazynDFINITY:
# Motivation

We want to update the ETH fees appearance logic:
1. To make sure we don't update UI too often, we throttle the `maxGasFee` update by 10s. On practice, It only affects Base tokens where the fees are being changed every ~1s. 
2. We now display skeleton only on initial fee fetch. The idea is to add some indicator (e.g. loader) that provides user a visual feedback when the fee values changes. It will be done separately.


AntonioVentilii (approved):
LGTM tks

PR #6768, opened 2025-05-21
refactor(frontend): remove networkId usage from IC Send review

DenysKarmazynDFINITY:
# Motivation

As the next step of the Send flow clean up, we remove `networkId` from IC send review form and related components.


AntonioVentilii (approved):
LGTM tks!

PR #6770, opened 2025-05-21
feat(frontend): Use static metadata for user Tokens

AntonioVentilii:
# Motivation

To solve one of our TODOs, instead of loading the metadata for all the User Tokens, we first check if we already have them in our static code metadata.


DenysKarmazynDFINITY (approved):
LGTM, thanks

Just for the info: we agreed to add tests separately

PR #6772, opened 2025-05-21
test(backend): Check that the backend returns custom Result types

bitdivine:
# Motivation
The candid generator doesn't do a nice job of encoding Result types.  To address this, we define custom result types that we use in the canister API.

# Changes
- Add a  CI check that verifies that there are no generic Results in the backend .did file.

# Tests
- If I change a response type to Result or Result_1 locally, `./scripts/lint.did.sh` fails.

### DIFF
--- a/.github/workflows/backend-checks.yml
+++ b/.github/workflows/backend-checks.yml
@@ -39,8 +39,10 @@ jobs:
           key: ${{ runner.os }}-cargo-backend-tests-${{ hashFiles('Cargo.lock', 'rust-toolchain.toml') }}
       - name: Install rust
         run: ./scripts/setup rust
-      - name: Lint
+      - name: Lint rust
         run: ./scripts/lint.rust.sh
+      - name: Lint candid
+        run: ./scripts/lint.did.sh
 
   workspace-dependencies:
     runs-on: ubuntu-24.04
--- a/scripts/lint.did.sh
+++ b/scripts/lint.did.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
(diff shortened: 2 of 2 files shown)


.github/workflows/backend-checks.yml
```diff
@@ -39,8 +39,10 @@ jobs:
           key: ${{ runner.os }}-cargo-backend-tests-${{ hashFiles('Cargo.lock', 'rust-toolchain.toml') }}
       - name: Install rust
         run: ./scripts/setup rust
-      - name: Lint
+      - name: Lint rust
         run: ./scripts/lint.rust.sh
+      - name: Lint candid
+        run: ./scripts/lint.did.sh
```
AntonioVentilii, line 45: if i recall correctly, this workflow is not triggered by the `src/declarations/` folder that has some did files. Shall we include it?

AntonioVentilii: problem is that we have more did files inside declarations, for example `kong_backend.did` or `rewards.did`

Or is the scope of the workflow ONLY for the backend?

bitdivine: Only the backend.  Reason: We don't control the kong did file.  We could ask them politely to publish a nicer API or we could (very carefully) rewrite their API for them.  But the backend is something we definitely do control.

You could also make a polite request to the rewards canister maintainer to make a nicer .did file, but I'm pretty sure that doesn't contain Result_[1-9] :wink:

bitdivine: If we did rewrite the kong (and similar) .did file, we'd probablywant to keep the original and run a CI test that the original and the rewritten IDL files are equivalent.

PR #6773, opened 2025-05-21
chore(rewards): Bump rewards canister version

bitdivine:
# Motivation
A newer version of the rewards canister is available that supports amounts larger than u64.

# Changes
- Update the rewards canister version.

# Tests


AntonioVentilii (approved):
Thank you!

PR #6775, opened 2025-05-21
refactor(frontend): remove convert-related code from IcSendAmount

DenysKarmazynDFINITY:
# Motivation

As the next step of the Send flow clean up, we remove convert-related code from IcSendAmount.


AntonioVentilii (approved):
LGTM tks!

PR #6781, opened 2025-05-21
fix(frontend): re-instate reactivity for ERC20 loading pending transactions

AntonioVentilii:
# Motivation

We re-instate the reactivity of component `CkEthereumPendingTransactionsListener`. We turned it off because it was triggering too many calls. However, after cleaning the statements, reducing the number of `metadata` calls, and introducing a debouncing, I verified that we do a normal amount of calls, the expected ones.


DenysKarmazynDFINITY (approved):
Awesome, let's test it on staging 🚀

PR #6783, opened 2025-05-22
feat(frontend): Add utility to get supported networks for token account id

FlavioInacta:
# Motivation

To display the supported networks for a given contact address, this PR adds a utility to get from `TokenAccountIdType` to the supported networks

# Changes

- Adds `getNetworksForTokenAccountIdType`

# Tests

- Add a test to ensure all types returns some networks

### DIFF
--- a/src/frontend/src/lib/constants/token-account-id.constants.ts
+++ b/src/frontend/src/lib/constants/token-account-id.constants.ts
@@ -1,0 +1,14 @@
+import { SUPPORTED_EVM_NETWORKS } from '$env/networks/networks-evm/networks.evm.env';
+import { SUPPORTED_BITCOIN_NETWORKS } from '$env/networks/networks.btc.env';
+import { SUPPORTED_ETHEREUM_NETWORKS } from '$env/networks/networks.eth.env';
+import { ICP_NETWORK } from '$env/networks/networks.icp.env';
+import { SUPPORTED_SOLANA_NETWORKS } from '$env/networks/networks.sol.env';
+import type { Network } from '$lib/types/network';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/constants/token-account-id.constants.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const TOKEN_ACCOUNT_ID_TO_NETWORKS: { [key in TokenAccountIdTypes]: Network[] } = {
+	Icrcv2: [ICP_NETWORK],
+	Btc: SUPPORTED_BITCOIN_NETWORKS,
+	Eth: [...SUPPORTED_ETHEREUM_NETWORKS, ...SUPPORTED_EVM_NETWORKS],
+	Sol: SUPPORTED_SOLANA_NETWORKS
+};
+
+// Get all possible keys from TokenAccountId at compile time
+export const TOKEN_ACCOUNT_ID_TYPES = Object.keys(
```
AntonioVentilii, line 17: if this is used only in the tests, may you please move it there?

AntonioVentilii (approved):
Supposing you prefer to concentrate it in an util, instead of using directly `TOKEN_ACCOUNT_ID_TO_NETWORKS`, LGTM

src/frontend/src/lib/constants/token-account-id.constants.ts
FlavioInacta: I'm not sure if I'll use it. But I thought it would be good to have a complete list. But I can remove it. Yes.

PR #6786, opened 2025-05-22
chore(frontend): Remove unused Goerli ICP contract

AntonioVentilii:
# Motivation

We don't support the Goerli network anymore, so we can remove the leftover contract of ICP on Goerli.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6788, opened 2025-05-22
feat(frontend): Lookup utils address book

daviddecentage:
# Motivation

We need a function to get a singular contact for a given address. If multiple addresses match we should only return the first match.

# Changes

- Added util function `getContactForAddress`
- Adjusted mock contacts generation

# Tests

Added tests for new utils function and adjusted other tests due to changes in the mock generation util function


AntonioVentilii (approved):
LGTM tks!

PR #6789, opened 2025-05-22
feat(frontend): Cache custom tokens

AntonioVentilii:
# Motivation

We want to start caching custom tokens, to increase the login speed. First, we save the cache for each update call to the backend canister.

# Changes

- Create IDB functions to store `CustomTokens` lists to the IDB cache.
- Use the set IDB functions in ICRC services and SPL services.

# Tests

Included more tests and a practical one:

Screenshot 2025-05-22 at 11 48 17


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6793, opened 2025-05-22
fix(frontend): Loading metadata shall be case-insensitive

AntonioVentilii:
# Motivation

When we are loading metadata from existing static tokens, we should compare the addresses without being case-sensitive.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6797, opened 2025-05-22
refactor(frontend): Transaction details

daviddecentage:
# Motivation

# Changes

# Tests


DenysKarmazynDFINITY:
I like how the new modals look like!

Since now we have the whole code in place, seems like we still can move some common between all networks items into a separate component to reduce some code duplications. Not in this PR, of course. 

Also, since we need to add some (at least basic) tests for those modals - how about we split this PR into smaller pieces just so we can add tests directly with their components?

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6812, opened 2025-05-22
refactor(frontend): Address card component

daviddecentage:
# Motivation

# Changes

# Tests


src/frontend/src/lib/components/address/AddressCard.svelte
```diff
@@ -0,0 +1,39 @@
...
+</script>
+
+<div
+	class={`flex rounded-lg border border-disabled bg-secondary px-2 py-3 items-${items}`}
+	class:border-error-solid={hasError}
+>
+	{@render logo?.()}
+
+	<div class="flex w-full flex-col truncate pl-2 pr-4 text-sm sm:text-base">
+		{@render content()}
```
DenysKarmazynDFINITY, line 31: Shall we reduce it to just 1 snipper with content - either `content` or `description`?

src/frontend/src/lib/components/send/SendDestination.svelte
```diff
@@ -22,21 +23,23 @@
...
-			<IconPenLine />
-		</button>
-	</div>
+	<AddressCard hasError={invalidDestination} items="center">
+		{#snippet logo()}
+			<NetworkLogo network={$sendToken.network} />
+		{/snippet}
+
+		{#snippet content()}
+			<div class="w-full truncate pl-2 pr-4 text-sm sm:text-base">{destination}</div>
```
DenysKarmazynDFINITY, line 32: I think we don't need this div anymore - all those classes are set in `AdressCard`

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/lib/components/send/SendDestination.svelte
daviddecentage: updated

PR #6817, opened 2025-05-22
feat(frontend): Divider components

daviddecentage:
# Motivation

Since we have some sort of dividers between mostly texts, and the current ones differ creatly in spacings and sizes, were creating components to streamline the usage of dividers (currently only dot dividers).

# Changes

Added `Divider` and `DividerList` components

# Tests

image

image

Demo of more elements with custom html
image


DenysKarmazynDFINITY (approved):
The PR and its future use cases have been discussed offline - LGTM, thanks :)

src/frontend/src/lib/components/common/DividerList.svelte
```diff
@@ -0,0 +1,23 @@
...
+		[key: `item${number}`]: Snippet;
+		dividerProps?: DividerProps;
+		dividerColor?: string;
+	}
+
+	const { dividerProps, dividerColor = '', ...snippets }: Props = $props();
+</script>
+
+<span class="flex flex-row items-center">
+	{#each Object.entries(snippets) as [key, snippet], index (key)}
```
FlavioInacta, line 15: Hm, probably the order is not guaranteed like that.

src/frontend/src/lib/components/common/Divider.svelte
```diff
@@ -0,0 +1,17 @@
...
+	export interface Props {
+		variant?: 'dot';
+	}
+
+	const { variant = 'dot' }: Props = $props();
+</script>
+
+{#if variant === 'dot'}
+	<svg
+		width="4px"
```
FlavioInacta, line 11: The size should depend on the font size

PR #6823, opened 2025-05-23
feat(devops): Trigger update tokens workflow on files changes

AntonioVentilii:
# Motivation

It makes sense that the workflow that updates the tokens is triggered by changes on the tokens JSON files.

So that any PR that changes it will be updated accordingly.


DenysKarmazynDFINITY (approved):
👍

PR #6825, opened 2025-05-23
refactor(frontend): Correct utils to verify Standard Ethereum tokens

AntonioVentilii:
# Motivation

The utils `isEthereumTokenToggleDisabled` and `isEthereumTokenToggleEnabled` are not exactly doing what the name suggests.

Basically, they should define whether an Ethereum or EVM native token can be toggled or not. For example, we don't want to give the choice to disable the ETH token in specific cases.

To that scope we refactor a bit the utils to understand better their scope.

# Changes

- Move utils to proper module `eth.utils.ts`.
- Rename them to `isStandardEthereumToken` and `isNotStandardEthereumToken`.
- Adapt usage.

# Tests

We created new tests, and since we are here, for all the ETH utils.

### DIFF
--- a/src/frontend/src/eth/utils/eth.utils.ts
+++ b/src/frontend/src/eth/utils/eth.utils.ts
@@ -5,6 +5,12 @@ import {
 import type { OptionToken, TokenId } from '$lib/types/token';
 import { nonNullish } from '@dfinity/utils';
 
+export const isStandardEthereumToken = (token: OptionToken): boolean =>
+	nonNullish(token) && token.category === 'default' && token.standard === 'ethereum';
+
+export const isNotStandardEthereumToken = (token: OptionToken): boolean =>
+	!isStandardEthereumToken(token);
+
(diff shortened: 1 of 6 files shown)


src/frontend/src/eth/utils/eth.utils.ts
```diff
@@ -5,6 +5,12 @@ import {
 import type { OptionToken, TokenId } from '$lib/types/token';
 import { nonNullish } from '@dfinity/utils';
 
+export const isStandardEthereumToken = (token: OptionToken): boolean =>
```
DenysKarmazynDFINITY, line 8: "Standard" works for me too, but did you consider `isDefaultEthereumToken`/`isNotDefaultEthereumToken`?

DenysKarmazynDFINITY (approved):

src/frontend/src/eth/utils/eth.utils.ts
AntonioVentilii: Way better! Changed! Naming is hard...

PR #6826, opened 2025-05-23
feat(backend): add implementation for delete_contact logic

SYInacta:
# Motivation

This PR implements the delete_contact endpoint to allow users to remove contacts from their address book. This completes the basic CRUD operations for contact management, enabling users to fully manage their contacts within the application.

# Changes

* Implemented `delete_contact` functionality in `contacts.rs`, which:

  * Takes a contact ID as input
  * Validates that the contact exists for the current user
  * Removes the contact from the user's stored contacts
  * Updates the storage with the modified contact list
  * Returns the ID of the deleted contact or an appropriate error

# Tests

* Verified that contacts can be successfully deleted when they exist
* Confirmed that attempting to delete non-existent contacts returns the appropriate error
* Validated that when multiple contacts exist, only the specified contact is deleted
* Ensured that all contacts can be deleted one by one, and new contacts can be created afterward
* Added a commented-out test for user isolation that can be enabled when dynamic user registration is implemented

## Note

The implementation follows the same pattern as the existing contact management functionality, ensuring consistency in the codebase.


src/backend/src/lib.rs
```diff
@@ -913,9 +913,7 @@ pub fn update_contact(request: UpdateContactRequest) -> UpdateContactResult {
 #[update(guard = "caller_is_allowed")]
```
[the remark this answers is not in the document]
SYInacta: Merged main.

src/backend/src/contacts.rs
```diff
@@ -183,3 +183,47 @@ fn get_stored_contacts_safely(stored_principal: &StoredPrincipal) -> StoredConta
...
+            s.contact.insert(stored_principal, Candid(stored_contacts));
+
+            Ok(contact_id)
+        } else {
+            // No contacts exist for this user, so the specified contact doesn't exist either
+            // For idempotent behavior, we return Ok with the contact_id
+            Ok(contact_id)
+        }
+    })
+}
```
DecentAgeCoder, line 229: 1. Since the contact storage structure was changed from a Vec to a BTreeMap, entries can now be removed directly using the remove() method.
2. If the contact does not exist, or the contacts store has not been initialized we should return ContactError::ContactNotFound.

Something like the following:
```
pub fn delete_contact(contact_id: u64) -> Result<u64, ContactError> {
    let stored_principal = StoredPrincipal(ic_cdk::caller());
    let current_time = time();

    mutate_state(|s| {
        // Get the user's contacts directly from the state
        let mut stored_contacts = if let Some(stored_contacts) = s.contact.get(&stored_principal) {
            stored_contacts.clone()
        } else {
            // If the user has no contacts, return ContactNotFound
            return Err(ContactError::ContactNotFound);
        };

        // Check if the contact exists
        if !stored_contacts.contacts.contains_key(&contact_id) {
            return Err(ContactError::ContactNotFound);
        }

        // Remove the contact
        stored_contacts.contacts.remove(&contact_id);
        stored_contacts.update_timestamp_ns = current_time;

        // Update the storage
        s.contact.insert(stored_principal, Candid(stored_contacts));

        Ok(contact_id)
    })
}
```

DecentAgeCoder: Ensure that ContactError::ContactNotFound is returned when the data store has not been initialized, such as in the following cases:

1. Calling delete_contact before any call to create_contact.
2. Calling delete_contact with a non-existent contact_id.

DecentAgeCoder (changes requested):
Please provide the requested changes. Thanks!

src/backend/src/contacts.rs
```diff
@@ -191,39 +191,32 @@ fn get_stored_contacts_safely(stored_principal: &StoredPrincipal) -> StoredConta
 ///
 /// # Returns
 /// * `Ok(u64)` - The ID of the deleted contact if found and deleted
-/// * `Ok(u64)` - The ID of the contact if it was already deleted (idempotent operation)
+/// * `Err(ContactError::ContactNotFound)` - If the contact does not exist or the contacts store has not been initialized
```
bitdivine: Idempotent is actually generally preferable, so that API calls express desired state rather than how to get to that state, but it's not critical.  It's fine to merge as is.

bitdivine (approved):

src/backend/src/contacts.rs
DecentAgeCoder: Thanks for the input

PR #6831, opened 2025-05-23
refactor(frontend): Dividers on tokenlist

daviddecentage:
# Motivation

Due to the new divider components we adjsut the current usages where a dot divider should be placed.

# Changes

Adjusted TokenCard and LogoButton

# Tests
image


roman-nazaruk (approved):
LGTM

PR #6844, opened 2025-05-23
feat(frontend): custom symbols for icrc tokens

DenysKarmazynDFINITY:
# Motivation

As requested, we need to overwrite GHOSTNODE token symbol with a custom value.


roman-nazaruk (approved):
LGTM, thnx

PR #6872, opened 2025-05-26
fix(frontend): displays oisy symbol if available

BonomoAlessandro:
# Motivation

Some tokens does have a definded oisy token symbol. In this cases we want to display the oisy symbol instead of the default symbol.

# Changes

- displays oisy symbol if exists

# Tests

**before:**

Menu
image

Receive
image

Details
image

Hide
image

**after:**

Menu
image

Receive
image

Details
image

Hide
image


src/frontend/src/lib/components/tokens/Token.svelte
```diff
@@ -59,7 +60,9 @@
 	{/snippet}
 
 	{#snippet content()}
-		<output>{token.symbol}</output>
+		<output
+			>{`${getTokenDisplaySymbol(token)}${token.oisySymbol?.oisySymbol ? ` (${token.symbol})` : ''}`}</output
```
AntonioVentilii, line 64: isn't this redundant? meaning, are we not going to show it twice?

src/frontend/src/lib/components/tokens/HideTokenReview.svelte
```diff
@@ -25,7 +26,7 @@
 
 		<p class="text-center font-bold">
 			{#if nonNullish($token)}
-				{$token.name}
+				{getTokenDisplaySymbol($token)}
```
AntonioVentilii, line 29: i think we want to still show the name here, no?

src/frontend/src/lib/components/tokens/TokenMenu.svelte
```diff
@@ -48,7 +50,7 @@
 
 	let hideTokenLabel: string;
 	$: hideTokenLabel = replacePlaceholders($i18n.tokens.hide.token, {
-		$token: $token?.name ?? ''
+		$token: nonNullish($token) ? getTokenDisplaySymbol($token) : ''
```
AntonioVentilii, line 53: we still want to show the name here, no? why the symbol?

src/frontend/src/lib/components/tokens/HideTokenReview.svelte
BonomoAlessandro: @AntonioVentilii No, at this point we want to display the symbol instead of the name. I disucessed it with @StefanBerger-DFINITY .

AntonioVentilii (approved):
LGTM tks!

PR #6876, opened 2025-05-26
feat(frontend): Derived contact list

daviddecentage:
# Motivation

We need an easy way to access the list of contacts.

# Changes

Added derived store for contacts store

# Tests

Added tests for derived store


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6877, opened 2025-05-26
fix(frontend): displays oisy token symbol if exists instead of token symbol

BonomoAlessandro:
# Motivation

Some tokens do have a defined oisy token symbol. In this cases we want to display the oisy token symbol instead of the default token symbol in the send flow.

# Changes

- displays correct token symbol during send flow

# Tests

**before:**

form
image

review
image

**after:**

form
image

review
image


daviddecentage (approved):
LGTM, thx

PR #6879, opened 2025-05-26
fix(frontend): Reduce Coingecko calls for ERC20 tokens

AntonioVentilii:
# Motivation

The calls for prices of ERC20 tokens were being duplicated, because there is a filter missing when we build the price parameters.

# Changes

- Add missing filter when we update the list of ERC20 price parameters: we are updating existing data to group contract addresses of similar networks; so we need to remove them first and update it after.
- Add check and return for empty lists of ERC20 and SPL tokens.

# Tests

Created tests, since we are here.


src/frontend/src/lib/workers/exchange.worker.ts
```diff
@@ -96,7 +96,7 @@ const syncExchange = async ({
 		const existing = acc.find(({ coingeckoPlatformId }) => coingeckoPlatformId === coingeckoId);
 
 		return [
-			...acc,
+			...acc.filter(({ coingeckoPlatformId }) => coingeckoPlatformId !== coingeckoId),
```
DenysKarmazynDFINITY, line 99: I think it would be a bit cleaner if we define `acc` as `Record<coingeckoPlatformId, CoingeckoErc20PriceParams>` - we wouldn't need to loop through the array to find `existing` entry, as well as there would be no more need to run `acc.filter` on every iteration.

DenysKarmazynDFINITY: Awesome, thanks for the update! 

Can you please also take a look at the failing tests?

AntonioVentilii: fixed!

PR #6881, opened 2025-05-26
fix(frontend): displays oisy symbol on activtiy

BonomoAlessandro:
# Motivation

Some tokens do have a defined oisy token symbol. In this cases we want to display the oisy token symbol instead of the default token symbol on the activity page.

# Changes

- displays correct token symbol on activity page

# Tests

**before:**
image

**after:**
image


daviddecentage (approved):
LGTM, thx

PR #6882, opened 2025-05-26
fix(frontend): Remove duplicates SPL tokens from Coingecko queries

AntonioVentilii:
# Motivation

Derived store `enabledSplTokenAddresses` was simply mapping all Default and Custom SPL tokens. However, there are duplicates among them (some default could be among the custom, or some address can be non-unique, different owner for example).

So, we map the values to be unique.


DenysKarmazynDFINITY (approved):
Shall we add some tests? Not necessarily in this PR.

Otherwise, nice catch - LGTM :)

AntonioVentilii:
@DenysKarmazynDFINITY nono, you are right! Included here! Have a look please

DenysKarmazynDFINITY (approved):
Awesome, thanks!

PR #6883, opened 2025-05-26
fix(frontend): Remove duplicates ICRC tokens from Coingecko queries

AntonioVentilii:
# Motivation

Derived store `enabledIcrcLedgerCanisterIdsNoCk ` was simply mapping all Default and Custom ICRC tokens. However, there are duplicates among them (some default could be among the custom, or some address can be non-unique).

So, we map the values to be unique.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6884, opened 2025-05-26
fix(frontend): displays token symbol even balance is ZERO

BonomoAlessandro:
# Motivation

On the `token view` the symbol of the token should be displayed even if the token balance is ZERO.

# Changes

- displays token symbol

# Tests

**before:**

**after:**


daviddecentage (approved):
LGTM, thx

PR #6885, opened 2025-05-26
feat(frontend):  Expose InputAddress parse error

FlavioInacta:
# Motivation

To react in the address form to any errors, this PR exposes the error.

# Changes

- Expose the error

# Tests

No tests are added since for testing the `svelte-htm` library would be neede, which we don't have integrated.


daviddecentage (approved):
LGTM, thx

PR #6888, opened 2025-05-26
refactor(frontend): Extract service to load custom tokens

AntonioVentilii:
# Motivation

We want to use the cached tokens for the loading of custom tokens. So, first, we generalize the way that the custom tokens are loaded for each network.

To do that, we create a common service.

# Changes

- Create a common service that loads all custom tokens from the backend: it is a copy of the current services, but providing specific network features, like the filter.
- Use the service in another generic one, where we will add the loading from the cache too.
- Apply usage to SPL and ICRC tokens.

# Tests

Created tests, but no change in local replica.


src/frontend/src/lib/services/custom-tokens.services.ts
```diff
@@ -0,0 +1,53 @@
...
+import type { SetIdbTokensParams } from '$lib/types/idb-tokens';
+import type { OptionIdentity } from '$lib/types/identity';
+import { get } from 'svelte/store';
+
+const loadCustomTokensFromBackend = async ({
+	identity,
+	certified,
+	filterTokens,
+	setIdbTokens
+}: {
```
DenysKarmazynDFINITY, line 13: nitpick: let's create an interface for these params and re-use in both functions

AntonioVentilii: yes, makes sense!

PR #6899, opened 2025-05-26
feat(frontend): Transaction contact card component

daviddecentage:
# Motivation

In the new design for transaction details we need to display a contact card (if one found) or alternatively present a button to add the address to a contact. We also need to display some actions etc.

# Changes

- Added transaction contact card
- Added no contact assosiation icon
- Adjusted AvatarWithBadge to display then

# Tests

image

image

image


src/frontend/src/lib/components/contact/AvatarWithBadge.svelte
```diff
@@ -21,13 +23,18 @@
...
-		{#if nonNullish(badge)}
-			{#if badge.type === 'addressTypeOrCount'}
-				<AddressesBadge addresses={contact.addresses} />
-			{:else if badge.type === 'addressType'}
-				<AddressesBadge addresses={contact.addresses} selectedAddress={badge.address} />
+		{#if nonNullish(contact)}
+			<Avatar name={contact.name} {variant} />
+			{#if nonNullish(badge)}
+				{#if badge.type === 'addressTypeOrCount'}
+					<AddressesBadge addresses={contact.addresses} />
```
DenysKarmazynDFINITY, line 30: Instead of `if`, it seems it can be simplified to smth like:
`<AddressesBadge addresses={contact.addresses} selectedAddress={badge.type === 'addressTypeOrCount' ? badge.address : undefined} />`

src/frontend/src/lib/components/contact/AvatarWithBadge.svelte
```diff
@@ -21,13 +23,18 @@
 
 <div class="flex">
 	<div class="relative">
```
DenysKarmazynDFINITY, line 25: Sorry, it's not related to your PR, but since we are here - can you please try to move `relative` to the parent div? I somehow don't why we need 2 separate divs in this case.

daviddecentage: Updated

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6901, opened 2025-05-27
fix(e2e): Change expected string for received ICP

AntonioVentilii:
# Motivation

In a recent PR, we changed how we display the amounts.


BonomoAlessandro (approved):
LGTM

PR #6902, opened 2025-05-27
feat(dependabot): Enable dependabot for all dependencies of npm

AntonioVentilii:
# Motivation

We want to stay up to date with all the dependencies of the frontend, for example `ethers` or solana kit. In any case, we can analyze the PRs created and decide singularly if we accept the update or not.


BonomoAlessandro (approved):
LGTM

PR #6906, opened 2025-05-27
feat(frontend): Create event for enabled and disabled tokens

AntonioVentilii:
# Motivation

It is useful to know which tokens are included and which ones are disabled by the users. So that we can have static metadata for them (or remove them) and improve performance.


roman-nazaruk (approved):
LGTM

PR #6908, opened 2025-05-27
chore(frontend): Remove deprecated TODO from SOL wallet worker

AntonioVentilii:
# Motivation

We decided to never show the toast errors for the SOL wallet workers, just to log to console. So we can remove the deprecated TODO.


roman-nazaruk (approved):
LGTM

PR #6915, opened 2025-05-27
feat(devops): Shard frontend config tests

AntonioVentilii:
# Motivation

# Changes

- Split the list of frontend networks in a different script (to be re-used).
- Adapt script that tests IC domains to accept an optional domains file to compare: it will be used by the Github workflow.
- Modify the Github workflow to shard the tests:
  - First step creates the matrix of networks, using the new script.
  - Second step builds the frontend for each network in parallel and it saves the results.
  - Third and last step merges the results and uses the test script to compare them.

# Tests

The scripts still work the same and the workflow works in this same PR.

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -78,7 +78,7 @@ export default defineConfig(
 				thresholds: {
 					autoUpdate: true,
 					statements: 72,
-					branches: 86,
+					branches: 85,
 					functions: 75,
 					lines: 72
 				}


.github/workflows/frontend-checks.yml
```diff
@@ -151,10 +151,34 @@ jobs:
...
+          echo "network-matrix=$NETWORKS" >> $GITHUB_OUTPUT
+
+  config-build:
     runs-on: ubuntu-24.04
+    needs: config-preparation
     permissions:
       contents: read
+    strategy:
+      matrix:
+        network: ${{ fromJson(needs.config-preparation.outputs.network-matrix) }}
```
bitdivine, line 181: Nice, so this is the parallelism you mentioned.

bitdivine (approved):

PR #6916, opened 2025-05-27
feat(frontend): implement Tabs component

DenysKarmazynDFINITY:
# Motivation

To implement the update design for the SendDestinationWizardStep, we need to create a new component - `Tabs`.

Screenshot 2025-05-27 at 10 43 24

Screenshot 2025-05-27 at 10 43 18

### DIFF
--- a/src/frontend/src/lib/components/ui/Tabs.svelte
+++ b/src/frontend/src/lib/components/ui/Tabs.svelte
@@ -1,0 +1,34 @@
+<script lang="ts">
+	import type { Snippet } from 'svelte';
+	import type { NonEmptyArray } from '$lib/types/utils';
+
+	interface Props {
+		tabs: NonEmptyArray<{ label: string; id: string }>;
+		activeTab: string;
+		children: Snippet;
+		styleClass?: string;
+	}
+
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/ui/Tabs.svelte
```diff
@@ -0,0 +1,34 @@
...
+			onclick={() => (activeTab = id)}
+			aria-label={label}
+			class="w-full justify-center rounded-none border-0 border-b-2 p-2 text-sm font-semibold transition hover:border-brand-primary hover:text-brand-primary sm:text-base"
+			class:ml-4={index !== 0}
+			class:text-tertiary-inverted={activeTab !== id}
+			class:border-primary={activeTab !== id}
+			class:text-brand-primary={activeTab === id}
+			class:border-brand-primary={activeTab === id}
+		>
+			{label}
```
AntonioVentilii, line 27: shall this maybe be a snippet?

DenysKarmazynDFINITY: I tried the snippet approach first, but since for detecting if a tab is active or not I need to work with a combination of label and id params, it appeared to be much more convenient to pass this data as a prop.

PR #6927, opened 2025-05-27
feat(frontend): cleans up DeleteAddressConfirmBottomSheet

BonomoAlessandro:
# Motivation

The confirm delete address bottom sheet displays a line between the header and the body of the component. This line should be removed.

# Changes

- removes bad looking line

# Tests

**before:**
image

**after:**
image


AntonioVentilii (approved):
Thank you!

PR #6971, opened 2025-05-28
feat(frontend): Remove deprecated cached fallback for SPL custom tokens

AntonioVentilii:
# Motivation

We are now caching the custom tokens in the IDB cache. So we can deprecate the old session cache for SPL custom tokens.

# Changes

- Remove all operations linked to cache key `SPL_CUSTOM_TOKENS_KEY`.

# Tests

Since we are here, new tests for service `saveCustomTokens`.


BonomoAlessandro (approved):
LGTM

PR #6973, opened 2025-05-28
feat(frontend): Add wrapper method for service method calls

FlavioInacta:
# Motivation

In order to reuse common patterns when calling a service method, this PR adds a wrapper method for adding identity, tracking and toast functionality.

# Changes

- Adds wrapper method

# Tests

- Added tests for tracking, identity and toast


src/frontend/src/lib/services/utils.services.ts
```diff
@@ -0,0 +1,58 @@
...
+		} catch (err) {
+			if (nonNullish(trackEventNames.error)) {
+				trackEvent({
+					name: trackEventNames.error,
+					metadata: { error: `${errorDetailToString(err)}` }
+				});
+			}
+
+			if (nonNullish(toastErrorMessage)) {
+				toastsError({ msg: { text: toastErrorMessage }, err });
```
AntonioVentilii, line 53: no need to throw it if we handle it with a toast error

```suggestion
				toastsError({ msg: { text: toastErrorMessage }, err });
				return 
```

src/frontend/src/lib/services/utils.services.ts
```diff
@@ -0,0 +1,58 @@
...
+import type { OptionIdentity } from '$lib/types/identity';
+import { errorDetailToString } from '$lib/utils/error.utils';
+import type { Identity } from '@dfinity/agent';
+import { isNullish, nonNullish } from '@dfinity/utils';
+
+/**
+ * Higher-order function to handle common authentication, tracking, and error handling patterns
+ */
+export const wrapCallWith =
+	<T extends object, R>({
```
AntonioVentilii, line 13: maybe this one should extend `{ identity: OptionIdentity }`, WDYT?

`<T extends { identity: OptionIdentity }, R>`

src/frontend/src/lib/services/utils.services.ts
FlavioInacta: I think we should, maybe the consumer need to know if there was an error or not.

src/frontend/src/lib/services/utils.services.ts
FlavioInacta: Yes, nice. Didn't figure out how the compiler does not complain with the types. That's what Claude suggested. ;)

Fixed.

src/frontend/src/lib/services/utils.services.ts
```diff
@@ -0,0 +1,58 @@
...
+			} as T;
+
+			const result = await methodToCall(paramsWithIdentity);
+
+			if (nonNullish(trackEventNames.success)) {
+				trackEvent({ name: trackEventNames.success });
+			}
+
+			return result;
+		} catch (err) {
```
AntonioVentilii, line 44:
```suggestion
		} catch (err: unknown) {
```

src/frontend/src/lib/services/utils.services.ts
FlavioInacta: Agree, done.

src/frontend/src/lib/services/utils.services.ts
FlavioInacta: Done

PR #6983, opened 2025-05-28
feat(frontend): added new replaceErrorFields util

roman-nazaruk:
# Motivation

We need to log and send certain errors to Plausible, but some fields must be excluded.
This PR introduces a utility that removes specified fields from errors, whether they are strings or plain objects.

# Changes

Created new replaceErrorFields util.

# Tests

Covered  replaceErrorFields util method with the tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2538,9 +2538,10 @@
 			}
 		},
 		"node_modules/@jridgewell/resolve-uri": {
-			"version": "3.1.1",
-			"resolved": "https://registry.npmjs.org/@jridgewell/resolve-uri/-/resolve-uri-3.1.1.tgz",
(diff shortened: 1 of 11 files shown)


src/frontend/src/tests/lib/utils/error.utils.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+		});
+
+		expect(result).toBe(input);
+	});
+
+	it('handles string with keys having no comma', () => {
+		const input = `
+    {
+      "message": "fail"
+      "requestId": "abc123"
```
AntonioVentilii, line 53: can you add a test with "Request ID" ? basically with whitespaces; that is mostly the case for ICRC errors

AntonioVentilii (approved):

PR #6994, opened 2025-05-28
feat(frontend): displays skeletons while loading contacts

BonomoAlessandro:
# Motivation

When the user opens the Address Book and the contacts are not loaded yet, the user should see that the loading process is still going on. As soon as the contacts are loaded, they should be displayed in the Address Book

# Changes

- adds skeletons

# Tests

**before:**
image

**after:**

not yet loaded
image

empty Address Book
image


src/frontend/src/lib/components/address-book/AddressBookModal.svelte
```diff
@@ -240,7 +243,10 @@
 	</svelte:fragment>
 
 	{#if isNullish(contacts)}
-		{$i18n.address_book.text.loading_contacts}
+		<ContentWithToolbar styleClass="mx-2 flex flex-col">
+			<SkeletonCards rows={3} />
+			<ButtonCloseModal slot="toolbar" />
+		</ContentWithToolbar>
```
AntonioVentilii, line 249: shouldn't this go in the `ADDRESS_BOOK` step?

AntonioVentilii (approved):

PR #6996, opened 2025-05-28
style(frontend): Remove address from transaction modal if it is myself

AntonioVentilii:
# Motivation

We decided to remove the information about the user address in the transaction modal. For example, when it is a `receive` transaction, the `To` field will not appear.

### Before

Screenshot 2025-05-28 at 14 46 56

### After

Screenshot 2025-05-28 at 14 39 59


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6998, opened 2025-05-28
fix(frontend): Correctly parse `Principal` from cached tokens

AntonioVentilii:
# Motivation

The `Principal` objects of the Ledger and Index canister IDs of custom tokens are saved in the cached as `Uint8Array` objects. So we need to parse them correctly when we re-fetch them form the cache.

# Changes

Add mapper for Icrc tokens in service `loadNetworkCustomTokens`.

# Tests

Included tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #6999, opened 2025-05-28
feat(frontend): Disable edit contact name form when backend call is in progress

FlavioInacta:
# Motivation

To prevent sending a backend call multiple times, the form should be disabled during backend calls.

# Changes

- Disabled form during backend call

# Tests

- Added test


FlavioInacta:
Is now already implemented by @BonomoAlessandro. Thx

PR #7002, opened 2025-05-28
feat(backend): ensure address book guards are tested and enforced

DecentAgeCoder:
# Motivation

Enhance test coverage by ensuring all address book guards have the expected guards defined.

# Changes
- Implemented tests to verify the correctness of the implemented guards

# Tests

- Added tests to ensure creating a contact requires authentication.
- Added tests to verify that retrieving a single contact requires authentication.-
- Added test to confirm that fetching all contacts requires authentication.
- Added test to ensure updating a contact requires authentication.
- Added test to verify that deleting a contact requires authentication.


bitdivine (approved):
Thanks!

peterpeterparker (commented):
I might be picky, but in addition to testing with an anonymous identity, I would duplicate the test (like `describe.each` in JS) for a valid identity that is not recognized as a user by OISY.

DecentAgeCoder:
@peterpeterparker  This is a good proposal, it helps keep the number of tests manageable. I also believe we should include a positive test case. To avoid introducing an additional crate like rstest, we will use a loop-based approach in Rust to test access for both anonymous users and various principals. We will provide them in a separate PR.

PR #7003, opened 2025-05-28
refactor(frontend): Use modaldata address in edit address book step

daviddecentage:
# Motivation

For the upcoming feature #6980 we need the possibility to use an address from modal data and prefill and disable the EditAddress input.

# Changes

Take modal data address if supplied and prefill/disable address input

# Tests

WIth entrypoint & address defined in store data:
image

Without store data:
image


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7005, opened 2025-05-28
refactor(frontend): Save address link on transaction contact card

daviddecentage:
# Motivation

If an address isnt found in the current contacts, we offer a link to add the address to a contact (or create a new contact) via a direct entrypoint in the address book.

# Changes

- Added save address button to open address book

# Tests

With contact:
image

Without contact:
image

### DIFF
--- a/src/frontend/src/btc/derived/btc-contacts.derived.ts
+++ b/src/frontend/src/btc/derived/btc-contacts.derived.ts
@@ -1,0 +1,8 @@
+import { contacts } from '$lib/derived/contacts.derived';
+import type { NetworkContacts } from '$lib/types/contacts';
+import { getNetworkContacts } from '$lib/utils/contacts.utils';
+import { derived, type Readable } from 'svelte/store';
+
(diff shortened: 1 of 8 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

DenysKarmazynDFINITY (approved):

src/frontend/src/lib/components/transactions/TransactionContactCard.svelte
```diff
@@ -19,38 +22,55 @@
 
 	const { type, to, from, toExplorerUrl, fromExplorerUrl }: Props = $props();
 
+	let address: string | undefined = $derived(
+		type === 'send' && nonNullish(to) ? to : type !== 'send' && nonNullish(from) ? from : undefined
```
peterpeterparker, line 26: The PR is a refactoring but, the logic isn't exactly the same. For example, the address can now be set with `to` even if `from` is nullish which was not the case before. Is that expected @daviddecentage ?

PR #7010, opened 2025-05-28
feat(frontend): Show owner address instead of ATA in Solana transaction modal

AntonioVentilii:
# Motivation

For the Solana transactions, we want to show the owner address instead of the ATA address in the modals, when it concerns SPL tokens.

# Changes

- Calculate owner address (if any) for the `from` and `to` props of the transaction.
- Fallback on the original addresses if nothing found (either it is not an ATA address, or it is closed).
- Render the additional ATA address field if necessary.

# Tests

### Before

Screenshot 2025-05-28 at 15 25 17

### After

Screenshot 2025-05-28 at 15 24 26

### No changes for native tokens

Screenshot 2025-05-28 at 15 27 44


src/frontend/src/sol/components/transactions/SolTransactionModal.svelte
```diff
@@ -29,16 +31,46 @@
...
+
+	const updateToAddress = async () => {
+		to =
+			nonNullish(solanaNetwork) && nonNullish(toAddress)
+				? ((await getAccountOwner({ address: toAddress, network: solanaNetwork })) ?? toAddress)
+				: toAddress;
+	};
+
+	$effect(() => {
+		[fromAddress];
```
DenysKarmazynDFINITY, line 65: Interesting, I assume this line specifies a dependency for the `$effect` block?

DenysKarmazynDFINITY (approved):
Small questions, otherwise LGTM

src/frontend/src/sol/components/transactions/SolTransactionModal.svelte
AntonioVentilii: Indeed, it is the only way i found to trigger the effect... We use it elsewhere IIRC

src/frontend/src/sol/components/transactions/SolTransactionModal.svelte
```diff
@@ -29,16 +31,46 @@
...
+		to: toAddress,
 		type,
 		status
 	} = $derived(transaction);
 
+	let solanaNetwork = $derived(
+		nonNullish(token?.network.id) ? mapNetworkIdToNetwork(token?.network.id) : undefined
+	);
+
+	let from = $state<SolTransactionUi['from']>();
```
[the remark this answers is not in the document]
AntonioVentilii: Good pone! Changed to a `$derived` rune

PR #7011, opened 2025-05-28
fix(frontend): Include suggested changes SaveAddressStep

daviddecentage:
# Motivation

For testing purposes a handler has been added to onClick, but it should be empty as the behavior is broken now.
See: https://github.com/dfinity/oisy-wallet/pull/7000#discussion_r2111968875

# Changes

Revert back to empty onClick handler


roman-nazaruk (approved):
LGTM

PR #7012, opened 2025-05-28
feat(frontend): Load cached custom tokens for SPL and ICRC

AntonioVentilii:
# Motivation

We can now enable the cache-fetching for the ICRC and SPL tokens.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7014, opened 2025-05-28
fix(frontend): moved Toast Error for minter info to plausible

roman-nazaruk:
# Motivation

We don't want to show anymore the toast error about the loading ckETH and ckBTC minter info. However, we want to keep track of it.

# Changes
- Removing the toast error from services `syncCkEthMinterError `and `loadData`.
- Add console warn for the error.

# Tests

Added new tests.


AntonioVentilii (approved):
As discussed, please pass the error too as metadata. Rest LGTM tks

PR #7020, opened 2025-05-29
feat(frontend): implement network contacts stores

DenysKarmazynDFINITY:
# Motivation

We need to implement network contacts stores that are gonna be used in the Address Picker send step.


daviddecentage (approved):
LGTM, thxx

PR #7029, opened 2025-05-29
feat(frontend): change labels and alts from dapp to app

StefanBerger-DFINITY:
# Motivation

Since also show apps in the explorer, we should correctly name them apps and not dapps.
Note: The page header was already set to "Explore Apps"

# Changes

Labels and alts texts for buttons change from dapp to app


DenysKarmazynDFINITY (approved):
Thanks :)

PR #7030, opened 2025-05-29
fix(frontend): pass networkId to BtcSendDestination

DenysKarmazynDFINITY:
# Motivation

We need to pass networkId to BtcSendDestination for the validation to work correctly.


BonomoAlessandro (approved):
LGTM

PR #7034, opened 2025-05-29
feat(frontend): user should be able to create contact when pressing Enter

CvetelinaAngelova:
# Motivation

Fixes the bug: Contact Creation should create the contact when I press Enter when entering a name.

Previously, pressing Enter while typing a contact's name did not trigger the creation of the contact. This PR ensures that the Enter key functions as expected, enhancing user experience during contact creation.

97d452c6-2669-4975-867c-9a12461dfc58

# Changes

Event Handling: Added a global keydown event listener in ContactForm.svelte to detect the Enter key press.
Function Implementation: Implemented the handleKeydown.

# Tests

Manual Testing:
Navigated to the contact creation form.
Entered a name in the input field.
Pressed the Enter key.
Confirmed that the contact was successfully created upon pressing Enter.


BonomoAlessandro (approved):
LGTM

src/frontend/src/lib/components/address-book/ContactForm.svelte
```diff
@@ -7,14 +7,24 @@
...
 	let isValid = $derived(notEmptyString(contact?.name?.trim?.()));
 
+	const handleKeydown = (event: KeyboardEvent): void => {
+		if (event.key === 'Enter') {
+			event.preventDefault();
+			onSubmit();
+		}
+	};
+
 	export { isValid };
```
peterpeterparker, line 23: Is the export expected?

src/frontend/src/lib/components/address-book/ContactForm.svelte
```diff
@@ -7,14 +7,24 @@
...
 </script>
 
+<svelte:window on:keydown={handleKeydown} />
```
peterpeterparker, line 26: Are you sure? Shouldn't rather the event listener be solely scoped to the input text field?

src/frontend/src/lib/components/address-book/ContactForm.svelte
```diff
@@ -7,14 +7,24 @@
...
+
 <form class="w-full">
```
peterpeterparker, line 28: Why is there a form if it does not submit or do anything? Is that expected?

peterpeterparker (commented):
Radomly noticed this PR which is already merged but, few things are suspicious to me so added few questions @CvetelinaAngelova and @BonomoAlessandro .

Also, there is no tests?

src/frontend/src/lib/components/address-book/ContactForm.svelte
BonomoAlessandro: @peterpeterparker you are right, the submit of the form should be triggered by a "Enter" but tbh i don't know if this is the default. I will investigate it.

PR #7039, opened 2025-05-29
fix(frontend): Multiple addresses overflow

daviddecentage:
# Motivation

In the address book lists, too many addresses lead to overflowing content.

# Changes

Switch to transition instead of collapsible

# Tests
Before
image

After:
image

### DIFF
--- a/src/frontend/src/lib/components/contact/AddressesBadge.svelte
+++ b/src/frontend/src/lib/components/contact/AddressesBadge.svelte
@@ -17,7 +17,7 @@
 </script>
 
 {#if addresses.length > 0}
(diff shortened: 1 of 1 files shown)


BonomoAlessandro (approved):
LGTM

src/frontend/src/lib/components/contact/ContactCard.svelte
```diff
@@ -46,10 +48,10 @@
 		{/snippet}
 
 		{#snippet description()}
-			<span class="flex items-center">
+			<span class="block w-full items-center truncate">
```
DenysKarmazynDFINITY, line 51: since you removed flex, items-center does not work anymore

PR #7046, opened 2025-05-29
feat(frontend): destination tab and selected contact vars

DenysKarmazynDFINITY:
# Motivation

Destination tab and selected contact vars should defined in the most parent component so we can access them in different send wizard steps.


roman-nazaruk (approved):
LGTM, thnx

PR #7048, opened 2025-05-29
fix(frontend): add aria-label and role to avatar element

CvetelinaAngelova:
# Motivation

The primary goal is to make the Avatar component more accessible, aligning with best practices for web accessibility.

# Changes

Updated the Avatar component to include role="img" and aria-label attributes.

# Tests

Manual Testing: 
   Verified that screen readers announce the avatar correctly with both name and alt props.
    Ensured that the IconAvatar component renders without issues.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7050, opened 2025-05-29
feat(e2e): Remove Sol Testnet (not Devnet)

AntonioVentilii:
# Motivation

We are going to remove the support for Solana Testnet (not Solana Devnet), since the new [Solana RPC Canister](https://github.com/dfinity/sol-rpc-canister) (that we are going to integrate) will not support it.

So, we start with the E2E tests.


BonomoAlessandro (approved):
LGTM

PR #7053, opened 2025-05-30
feat(frontend): display only valid contacts for ICP/ICRC tokens

DenysKarmazynDFINITY:
# Motivation

We need to separate `icNetworkContacts` for the ICP and ICRC cases in order to display only valid addresses for the selected IC token.


roman-nazaruk (approved):
LGTM, thnx

PR #7054, opened 2025-05-30
fix: update font size and colors

CvetelinaAngelova:
# Motivation

The Address Info section had visual issues impacting readability and design consistency:

The font appeared too light, especially in light mode, making it harder to read 

# Changes

Change the font to the main color and size by design

# Tests
Manual test the view
Screenshot 2025-05-30 at 12 36 19

### DIFF
--- a/src/frontend/src/lib/components/address-book/AddressInfoCard.svelte
+++ b/src/frontend/src/lib/components/address-book/AddressInfoCard.svelte
@@ -15,7 +15,7 @@
 
 <Value ref="address-info" element="div">
 	{#snippet label()}
-		<div class="mb-3 text-sm font-bold text-primary">
+		<div class="mb-3 text-md font-bold text-primary">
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/address-book/AddressInfoCard.svelte
```diff
@@ -15,7 +15,7 @@
 
 <Value ref="address-info" element="div">
 	{#snippet label()}
-		<div class="mb-3 text-sm font-bold text-primary">
+		<div class="text-md mb-3 font-bold text-primary">
```
DenysKarmazynDFINITY, line 18: I don't think `text-md` class exists - if you want to set 16px/1rem, then `text-base` is the class you're looking for. Also, since it is the default text size, you only need to set it if you want to override some value. Assuming that you tested it with `text-md` and it looked good for you, you can just remove it and proceed with the default size.

PR #7056, opened 2025-05-30
feat(frontend): filter network contacts

DenysKarmazynDFINITY:
# Motivation

Similar to what we do with the "recent destinations", we should filter use contacts based on the inputted address. Also, the "unknown destination" message should be only displayed if the address is not available in any of the categories.


src/frontend/src/eth/components/send/EthSendDestination.svelte
```diff
@@ -16,6 +17,7 @@
 	export let destination = '';
 	export let invalidDestination = false;
 	export let knownDestinations: KnownDestinations | undefined = undefined;
+	export let networkContacts: NetworkContacts | undefined = undefined;
```
roman-nazaruk, line 20: Should we migrate it to Svelte 5 now, or if it's planned for a future PR, could you please add a TODO for it?

roman-nazaruk (approved):
LGTM, thanks! Just a few observations.

PR #7058, opened 2025-05-30
feat(frontend): display contact label if available in SendContact

DenysKarmazynDFINITY:
# Motivation

We need to display contact label if available in SendContact.


roman-nazaruk (approved):
LGTM, thnx

PR #7061, opened 2025-05-30
fix(frontend): check for an empty string before showing contacts label

DenysKarmazynDFINITY:
# Motivation

We should also check if label is an empty in SendContact and SendDestination.


src/frontend/src/lib/components/send/SendContact.svelte
```diff
@@ -33,7 +33,7 @@
 	{#snippet title()}
 		{contact.name}
 
-		{#if nonNullish(contactLabel)}
+		{#if notEmptyString(contactLabel)}
```
roman-nazaruk, line 36: Can it be undefined, or is it only an empty string?

PR #7065, opened 2025-05-30
fix(frontend): incorrect address labels in SendContact

DenysKarmazynDFINITY:
# Motivation

There was an issue with the address label extraction in SendContact.


roman-nazaruk (approved):
LGTM, thnx

PR #7068, opened 2025-05-30
fix(frontend): add submit on the Enter key within the Add Address and Edit Address view.

CvetelinaAngelova:
# Motivation

This PR addresses the need to enhance the user experience by allowing form submission via the Enter key within the Add Address  and Edit Address view.
Previously, users could only submit the form by clicking the "Save" button, which was less intuitive and potentially hindered accessibility.

# Changes

Global Keydown Listener: Introduced a <svelte:window> event listener at the top level of the EditAddressStep.svelte component to capture Enter key presses.

Submission Handler: Implemented a handleKeydown function that prevents the default Enter key behavior and triggers the handleSave function when the form is valid.

Validation Check: Ensured that the form's isInvalid state is checked before allowing submission via the Enter key to prevent unintended submissions.

# Tests

Manual Testing:
Verified that pressing Enter while focused on any input field within the address form triggers the save action when the form is valid.
Confirmed that pressing Enter does not trigger the save action when the form is invalid.
Ensured that clicking the "Save" and "Cancel" buttons continues to function as expected.


src/frontend/src/lib/components/address-book/EditAddressStep.svelte
```diff
@@ -71,16 +80,20 @@
 	let isInvalid = $state(false);
 </script>
 
+<svelte:window on:keydown={handleKeydown} />
```
peterpeterparker, line 83: To be frank, I think adding those global listeners is a relatively bad pattern. Why aren't you using proper forms?

CvetelinaAngelova: I understand your concerns regarding the use of global event listeners. While they can introduce complexities, in this specific scenario, they were initially implemented to address a particular browser behavior. When a form contains multiple buttons, pressing the Enter key can inadvertently trigger the first button in the DOM, regardless of its intended function. This default behavior can lead to unintended actions, such as activating a reset button instead of submitting the form.

However, you're correct in pointing out the drawbacks of global listeners. To align with best practices and ensure cleaner code, I've refactored the implementation. Now, form submissions are handled through the onsubmit event, and buttons not intended to submit the form, like the reset button, are explicitly set with type="button". This approach adheres to standard HTML behavior, enhances the user experience, and eliminates the need for global keydown listeners.

PR #7073, opened 2025-06-02
feat(frontend): Cache ERC20 User tokens in IDB

AntonioVentilii:
# Motivation

We are caching custom tokens (ICRC and SPL) in IDB. Now we can use the same flow to cache User tokens (ERC20).

# Changes

- Create IDB functions to save and delete cached ETH tokens.
- Use the functions in the same places that we use it for the Custom Tokens.

# Tests

New tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7081, opened 2025-06-02
feat(frontend): SendReviewDestination component

DenysKarmazynDFINITY:
# Motivation

We need to implement a component for reviewing selected contact or recently used address in the Send Review form. Potentially, we can combine it later with the one from TransactionModal, but due to design specs differences (subject to discuss), for now it's easier to keep them separate.

Screenshot 2025-06-02 at 10 25 10


AntonioVentilii (approved):
LGTM tks!

PR #7084, opened 2025-06-02
fix(frontend): improves logic to display message box

BonomoAlessandro:
# Motivation

During the send flow, if i start entering an address into the address picker, a "not yet used" warning appears for around a second. The validation process should be improves in a way that this only happens if a valid not yet used address was entered.

# Changes

- improves display logic for message box

# Tests

**before:**

https://github.com/user-attachments/assets/c44e00f0-45d8-4bf1-a0ae-199924c08534

**after:**

https://github.com/user-attachments/assets/ebd7793d-176b-42a0-9468-da78689a0f28


BonomoAlessandro:
@DenysKarmazynDFINITY just added a testcase

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7092, opened 2025-06-02
feat(frontend): update SendReview form design

DenysKarmazynDFINITY:
# Motivation

As the next step, we want to use the new SendReviewDestination component in the SendReview form. Also, SendReview is now ready to display `selectedContact`.

Screenshot 2025-06-02 at 11 23 53


AntonioVentilii (approved):
LGTM tks!

PR #7094, opened 2025-06-02
feat(frontend): use selectedContact param in Send Review forms

DenysKarmazynDFINITY:
# Motivation

To start using the new SendReview form design, we should pass `selectedContact` from wizard to the form.


AntonioVentilii (approved):
LGTM tks!

PR #7098, opened 2025-06-02
feat(frontend): clicking an address should copy it

BonomoAlessandro:
# Motivation

When clicking an address on the contacts page the address should be copied to the clipboard.

# Changes

- copies address

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -169,9 +169,9 @@
 			}
 		},
 		"node_modules/@babel/parser": {
-			"version": "7.27.3",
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/contact/AddressListItem.svelte
```diff
@@ -33,8 +34,9 @@
 
 <button
 	onclick={() => onClick?.()}
-	disabled={nonNullish(onClick)}
+	disabled={isNullish(onClick)}
```
AntonioVentilii, line 37: good catch on this

AntonioVentilii (approved):
LGTM tks!

PR #7099, opened 2025-06-02
fix(frontend): use correct stores for validation

DenysKarmazynDFINITY:
# Motivation

To fix the problem with the "unknown address" validation, we should pass the correct stores to the input components.


AntonioVentilii (approved):
Thank you for the fix!

PR #7100, opened 2025-06-02
feat(frontend): Map owner address in Solana transaction services

AntonioVentilii:
# Motivation

We are going to show the owner address instead of the ATA address for Solana transactions. This will happen in multiple places: modal and recent addresses, for example.

So, it is better to map it directly during the fetching service of the transactions. In a follow-up PR, we will use it.

# Changes

- Expand type `SolTransactionUi` to have the `fromOwner` and the `toOwner`.
- Query the owner address both for from and to address. This can happen to be even for non token transactions, so we are forced to do it always.

# Tests

Added tests.

### DIFF
--- a/src/frontend/src/sol/services/sol-transactions.services.ts
+++ b/src/frontend/src/sol/services/sol-transactions.services.ts
@@ -16,7 +16,7 @@ import {
 	isNetworkIdSOLTestnet
 } from '$lib/utils/network.utils';
 import { findOldestTransaction } from '$lib/utils/transactions.utils';
-import { fetchTransactionDetailForSignature } from '$sol/api/solana.api';
+import { fetchTransactionDetailForSignature, getAccountOwner } from '$sol/api/solana.api';
 import { getSolTransactions } from '$sol/services/sol-signatures.services';
 import {
 	solTransactionsStore,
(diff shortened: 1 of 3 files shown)


src/frontend/src/sol/services/sol-transactions.services.ts
```diff
@@ -169,14 +169,25 @@ export const fetchSolTransactionsForSignature = async ({
 				return { parsedTransactions, cumulativeBalances, addressToToken };
 			}
 
+			const fromOwner: SolTransactionUi['from'] | undefined = await getAccountOwner({
```
DenysKarmazynDFINITY: shall we make it `SolTransactionUi['fromOwner'] | undefined`?

src/frontend/src/sol/services/sol-transactions.services.ts
```diff
@@ -169,14 +169,25 @@ export const fetchSolTransactionsForSignature = async ({
...
+				address: from,
+				network
+			});
+
+			const toOwner: SolTransactionUi['to'] | undefined = nonNullish(to)
```
DenysKarmazynDFINITY:
```suggestion
			const toOwner: SolTransactionUi['toOwner'] | undefined = nonNullish(to)
```

DenysKarmazynDFINITY (approved):
Small observation, otherwise LGTM

src/frontend/src/sol/services/sol-transactions.services.ts
AntonioVentilii: good catch! changed

src/frontend/src/sol/services/sol-transactions.services.ts
AntonioVentilii: and changed!

PR #7102, opened 2025-06-02
feat(frontend): Use Solana transaction data for owner address

AntonioVentilii:
# Motivation

After PR #7100 , we can use directly the props fromOwner and toOwner of the Solana transaction, instead of calculating the owner addresses in the component.

### DIFF
--- a/src/frontend/src/sol/components/transactions/SolTransactionModal.svelte
+++ b/src/frontend/src/sol/components/transactions/SolTransactionModal.svelte
@@ -19,9 +19,7 @@
 	} from '$lib/utils/format.utils';
 	import { replacePlaceholders } from '$lib/utils/i18n.utils';
 	import { isNetworkSolana } from '$lib/utils/network.utils';
-	import { getAccountOwner } from '$sol/api/solana.api';
 	import type { SolTransactionUi } from '$sol/types/sol-transaction';
-	import { mapNetworkIdToNetwork } from '$sol/utils/network.utils';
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/sol/components/transactions/SolTransactionModal.svelte
```diff
@@ -32,44 +30,19 @@
...
-	$effect(() => {
-		[fromAddress];
-		updateFromAddress();
-	});
-
-	$effect(() => {
-		[toAddress];
-		updateToAddress();
-	});
+	let from = $derived<SolTransactionUi['from'] | undefined>(fromOwner ?? fromAddress);
```
DenysKarmazynDFINITY: Do we want to also include `SolTransactionUi['fromOwner']`? Since it is one of the possible values

src/frontend/src/sol/components/transactions/SolTransactionModal.svelte
```diff
@@ -32,44 +30,19 @@
...
+	let to = $derived<SolTransactionUi['to'] | undefined>(toOwner ?? toAddress);
```
DenysKarmazynDFINITY: Here too

src/frontend/src/sol/components/transactions/SolTransactionModal.svelte
AntonioVentilii: Right! CHANGED!

PR #7103, opened 2025-06-02
feat(frontend): Use cached tokens for Ethereum and EVM networks

AntonioVentilii:
# Motivation

We can use the cached tokens for the Ethereum and EVM networks like we do for the other networks.

# Changes

- Create function to get the tokens from IDB cache.
- Extract sub-service of `loadUserTokens` that will use the cache ONLY for query calls and if required (defaults to not using the cache).
- Adapt the rest of the code to use the cache ONLY during the initial loader.

# Tests

Added tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7112, opened 2025-06-02
feat(frontend): Match contacts by owner address for Solana transactions

AntonioVentilii:
# Motivation

For Solana transactions, we want to show the owner address in the list (and match it with contacts), instead of the ATA address.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7121, opened 2025-06-03
feat(frontend): disable save button if alias is not changed

CvetelinaAngelova:
# Motivation

The motivation behind this PR is to enhance user experience by ensuring that the Save button is only enabled when alias is  changed during the edit process.

# Changes

Introduced originalLabel and labelChanged Derived States.
Updated Save Button Logic - Modified the disabled property of the Save button to consider both the form's validity (isInvalid) and whether the alias has been altered (labelChanged).

# Tests

Unit test
Verifies that the Save button remains disabled if the alias field is not modified during editing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6317,11 +6317,10 @@
 			}
 		},
 		"node_modules/bufferutil": {
-			"version": "4.0.9",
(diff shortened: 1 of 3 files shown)


daviddecentage (approved):
LGTM, thx

PR #7122, opened 2025-06-03
style(frontend): New address form

daviddecentage:
# Motivation

We adjust the `New Address` form according to how it was designed in Figma.

# Changes

- Changed bg color to `brand-subtle-10`
- Input font to regular
- Paste button in input as link not button

# Tests

image

### DIFF
--- a/src/frontend/src/lib/components/address/AddressForm.svelte
+++ b/src/frontend/src/lib/components/address/AddressForm.svelte
@@ -52,7 +52,6 @@
 	/>
 
 	<label for="label" class="font-bold">{$i18n.address.fields.label}</label>
-
 	<InputText
 		name="label"
 		placeholder={$i18n.address.form.label_placeholder}
--- a/src/frontend/src/lib/components/address/AddressForm.svelte
+++ b/src/frontend/src/lib/components/address/AddressForm.svelte
@@ -53,17 +53,15 @@
 
 	<label for="label" class="font-bold">{$i18n.address.fields.label}</label>
 
-	<span class="font-normal">
-		<InputText
-			name="label"
-			placeholder={$i18n.address.form.label_placeholder}
-			bind:value={address.label}
-			testId={ADDRESS_BOOK_ADDRESS_ALIAS_INPUT}
-			{disabled}
-			showResetButton={!disabled}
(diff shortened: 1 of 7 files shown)


src/frontend/src/lib/components/address/AddressForm.svelte
```diff
@@ -52,15 +52,18 @@
...
-		name="label"
-		placeholder={$i18n.address.form.label_placeholder}
-		bind:value={address.label}
-		testId={ADDRESS_BOOK_ADDRESS_ALIAS_INPUT}
-		{disabled}
-		showResetButton={!disabled}
-		required={false}
-	/>
+
+	<span class="font-normal">
```
roman-nazaruk: Maybe better to pass font as a prop to InputText instead?
font: 'normal' | 'bold' = 'normal';

src/frontend/src/lib/components/address/InputAddress.svelte
```diff
@@ -69,7 +69,9 @@
 <div
 	style={`--input-custom-border-color: ${borderColor}; --input-padding-inner-end: 100px; ${disabled ? '--input-background: var(--color-background-disabled);' : ''}`}
 >
-	<Input {disabled} inputType="text" bind:value innerEnd={qrButton} {...props}></Input>
+	<span class="font-normal">
```
roman-nazaruk: Same here

src/frontend/src/lib/components/address/AddressForm.svelte
daviddecentage: Just realized we can remove a class instead of manually setting font-normal

src/frontend/src/lib/components/address/InputAddress.svelte
daviddecentage: Just realized we can remove a class instead of manually setting font-normal

src/frontend/src/lib/components/address/AddressForm.svelte
roman-nazaruk: Thnx

roman-nazaruk (approved):
LGTM, thnx

PR #7124, opened 2025-06-03
style(frontend): Add address button

daviddecentage:
# Motivation

The add address button is not correctly styled.

# Changes

Adjusted to `secondary-light` and added more vertical padding as defined by Dan

# Tests

image


roman-nazaruk (approved):
LGTM, thnx

PR #7126, opened 2025-06-03
feat(frontend): integrate cycles ledger canister

DecentAgeCoder:
# Motivation

# Changes

# Tests


src/frontend/src/lib/canisters/cycles-ledger.canister.ts
```diff
@@ -0,0 +1,226 @@
...
+
+		return new CyclesLedgerCanister(canisterId, service, certifiedService);
+	}
+
+	icrc1BalanceOf = ({
+		account,
+		certified
+	}: { account: Account } & QueryParams): Promise<bigint> => {
+		const { icrc1_balance_of } = this.caller({ certified });
+		return icrc1_balance_of(account);
```
peterpeterparker: The Cycles ledger supports the ICRC standard. In my opinion, all standardized functions should be accessed through `@dfinity/ledger-icrc` and not added directly to this API. Otherwise, you might end up with code patterns like:

```
if not cycles ledger
  then dfinity/ledger-icrc
else
  use custom API
  map result
```

src/frontend/src/lib/canisters/cycles-ledger.canister.ts
```diff
@@ -0,0 +1,226 @@
...
+		const result = await withdraw_from(args);
+
+		if ('Ok' in result) {
+			return result.Ok;
+		}
+
+		throw mapWithdrawFromError(result.Err);
+	};
+
+	icrc3GetArchives = ({
```
peterpeterparker: If any of those ICRC-3 methods are useful for OISY, they should be implemented in `@dfinity/ledger-icrc`, not declared here.

src/frontend/src/lib/canisters/cycles-ledger.canister.ts
DecentAgeCoder: Thank you for the valuable input. I wasn't aware that DFINITY already provides an API for accessing ICRC-compliant ledger canisters.

At this stage, I only need to interact with the `icrc_2_allowance` endpoint.
Therefore, I will remove `cycles-ledger.canister.ts` and integrate the allowance functionality directly using the `dfinity/ledger-icrc` library int the generic cycle ledger API (`cycles-ledger.api.ts`).

src/frontend/src/lib/canisters/cycles-ledger.canister.ts
DecentAgeCoder: Thanks! See my other comment

DecentAgeCoder: Although icrc3_get_archives is defined in icrc_ledger.d.ts, the @dfinity/ledger-icrc package does not support this endpoint. It could be a candidate for a custom API implementation. However, since this endpoint is not required for my current use case, I will not implement it proactively.

scripts/generate.sh
```diff
@@ -55,6 +55,10 @@ DFX_NETWORK=ic ./scripts/build.icp_swap_pool.sh
 DFX_NETWORK=ic ./scripts/build.icp_swap_factory.sh
 # .. downloads candid for the xtc_ledger
 DFX_NETWORK=ic ./scripts/build.xtc_ledger.sh
+# .. downloads candid for the cycles_ledger
+DFX_NETWORK=ic ./scripts/build.cycles_ledger.sh
+# .. downloads candid for the cycles_depositor
+DFX_NETWORK=ic ./scripts/build.cycles_depositor.sh
```
AntonioVentilii: out of curiosity: why do we need to build them both or why do we need the candid files? aren't we already using them? both on local replica and in the others envs?

P.S. we deploy them with script `deploy.sh`

AntonioVentilii: As discussed offline, ignore my comment. We can always remove the scripts in the future, if we don't need them

DecentAgeCoder: Yes you are right. I have removed the scripts

PR #7129, opened 2025-06-04
refactor(frontend): implements more generalized qr code scanner

BonomoAlessandro:
# Motivation

We are using the `QrCodeReader` for the `Address Book`, `send flow` and the `convert flow`. We do want to have one generalized implementation that can be used on the different places.

# Changes

- adds generalized qr code scanner

# Tests

**add address:**

https://github.com/user-attachments/assets/c995d1b9-9524-44e2-abcc-50ee11a9cf64

**send:**

https://github.com/user-attachments/assets/8c5844a7-ea98-46f5-a37b-22667c9ff8d6

**convert:**

https://github.com/user-attachments/assets/cd854c40-954d-4e1b-bd66-5c5357a64bb6


src/frontend/src/lib/components/qr/QrCodeScanner.svelte
```diff
@@ -50,15 +42,12 @@
...
-	<div class="qr-code-wrapper h-full w-full md:min-h-[300px]" data-tid={ADDRESS_BOOK_QR_CODE_SCAN}>
-		<QRCodeReader on:nnsCancel={onCancel} on:nnsQRCode={onQRCode} />
-	</div>
-
-	<ButtonGroup slot="toolbar">
-		<ButtonCancel onclick={onClose} testId={ADDRESS_BOOK_CANCEL_BUTTON} />
-	</ButtonGroup>
-</ContentWithToolbar>
+<div
+	class="stretch qr-code-wrapper h-full w-full md:min-h-[300px]"
```
AntonioVentilii, line 46: shall these classes (or at least `qr-code-wrapper`) go into the new component? Does it make sense?

AntonioVentilii (approved):
Great! Tks! One obs:

src/frontend/src/lib/components/qr/QrCodeScanner.svelte
BonomoAlessandro: No, it should be in the `QrCodeScanner` component, otherwise you need to define `qr-code-wrapper` etc everywhere you want to use the `QrCodeScanner`.

PR #7131, opened 2025-06-04
fix(e2e): Bring page to front for guaranteed screenshot

AntonioVentilii:
# Motivation

With Playwright, when you are screenshotting a background page, chromium is very very slow, and sometimes hangs indefinitely. Using `page.bringToFront()` resolved the issue.

Source: https://github.com/microsoft/playwright/issues/20434#issuecomment-1477560521


DenysKarmazynDFINITY (approved):
Thanks!

PR #7141, opened 2025-06-04
style(frontend): Transparent button variant

daviddecentage:
# Motivation

According to figma design there is Button variants that we dont have in the codebase yet. Here we add them by allowing buttons to take a `transparent` prop.

# Changes

Added transparent prop and styling for it

# Tests

image

### DIFF
--- a/src/frontend/src/lib/styles/global/button.scss
+++ b/src/frontend/src/lib/styles/global/button.scss
@@ -481,7 +481,7 @@ a.as-button {
 	&.link {
 		@apply p-0;
 		@apply bg-transparent;
-		@apply text-brand-primary;
+		@apply text-brand-primary-alt;
 
 		&.disabled,
 		&[disabled] {
--- a/src/frontend/src/lib/styles/global/button.scss
+++ b/src/frontend/src/lib/styles/global/button.scss
@@ -12,9 +12,9 @@ a.as-button {
 	border-radius: var(--button-border-radius);
 
 	transition:
-					color 0.25s ease-out,
-					background 0.25s ease-out,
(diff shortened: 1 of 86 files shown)


src/frontend/src/lib/styles/global/button.scss
```diff
@@ -39,7 +39,7 @@ a.as-button {
 
 		&.icon {
 			--button-border-radius: var(--border-radius-sm-1_5x);
-			--button-padding: var(--padding);
+			--button-padding: var(--padding-0_75x);
```
DenysKarmazynDFINITY: This will affect all the existing buttons with icons, are we fine with it?

src/frontend/src/lib/styles/global/button.scss
```diff
@@ -95,6 +95,10 @@ a.as-button {
 		@apply text-primary-inverted;
 		@apply bg-brand-primary;
```
DenysKarmazynDFINITY: Let's update all non-transparent styling as shown below to avoid having 2 bg styles for transparent buttons:
```
&:not(.transparent) {
			@apply bg-brand-primary;
		}
&.transparent {
			@apply bg-transparent;
		}
```

src/frontend/src/lib/styles/global/button.scss
daviddecentage: Oh this one belongs in another PR, removing it

src/frontend/src/lib/styles/global/button.scss
daviddecentage: updated

PR #7143, opened 2025-06-04
refactor(frontend): use localized value in SwapProvider

DenysKarmazynDFINITY:
# Motivation

We had an unlocalized value displayed in SwapProvider.


AntonioVentilii (approved):
Good catch

PR #7144, opened 2025-06-04
refactor(frontend): Copy Icon

daviddecentage:
# Motivation

The copy icon differs from other icon buttons. We use the icon button now and remove the color prop since we can now use the colorStyle prop.

# Changes

Adjusted all occurrences and upgraded CopyIcon to svelte 5 as well as reworking the props.

# Tests

image

image

image

image

image

image

### DIFF
--- a/src/frontend/src/lib/components/ui/Copy.svelte
+++ b/src/frontend/src/lib/components/ui/Copy.svelte
@@ -1,9 +1,9 @@
 <script lang="ts">
 	import IconCopy from '$lib/components/icons/IconCopy.svelte';
-	import { i18n } from '$lib/stores/i18n.store';
-	import { copyToClipboard } from '$lib/utils/clipboard.utils';
 	import ButtonIcon from '$lib/components/ui/ButtonIcon.svelte';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/transactions/TransactionAddressActions.svelte
```diff
@@ -22,16 +22,16 @@
...
-	<ExternalLink
-		iconSize="18"
-		href={explorerUrl}
-		ariaLabel={explorerUrlAriaLabel}
-		inline
-		color="blue"
-		testId={explorerUrlTestId}
-	/>
-{/if}
+<span class="float-right flex pl-1.5 align-top">
```
DenysKarmazynDFINITY, line 25: can we replace float-right with items-right?

DenysKarmazynDFINITY (approved):
Small q, otherwise LGTM

src/frontend/src/lib/components/transactions/TransactionAddressActions.svelte
daviddecentage: No, as I want that item to appear right not the contents

PR #7145, opened 2025-06-04
feat(frontend): TokenModal component

DenysKarmazynDFINITY:
# Motivation

We need to create a re-usable TokenModal component. It is based on the existing `Token` component, but now includes the Modal-related usage. Also, instead of `Value`, `ModalValue` is now used for displaying the data.

UPD: the design is now a bit different

Screenshot 2025-06-05 at 11 48 29

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokenModal.svelte
+++ b/src/frontend/src/lib/components/tokens/TokenModal.svelte
@@ -2,6 +2,8 @@
 	import { Modal } from '@dfinity/gix-components';
 	import { nonNullish } from '@dfinity/utils';
 	import type { Snippet } from 'svelte';
+	import { isTokenErc20 } from '$eth/utils/erc20.utils';
+	import { isTokenIcrc, isTokenDip20 } from '$icp/utils/icrc.utils';
 	import NetworkLogo from '$lib/components/networks/NetworkLogo.svelte';
 	import ButtonDone from '$lib/components/ui/ButtonDone.svelte';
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/tokens/TokenModal.svelte
```diff
@@ -0,0 +1,98 @@
...
+							alt={replacePlaceholders($i18n.core.alt.logo, { $name: token.name })}
+							color="white"
+						/>
+					</span>
+				{/snippet}
+			</ModalValue>
+
+			{@render children?.()}
+
+			{#if ['icrc', 'erc20'].includes(token.standard)}
```
AntonioVentilii: two things:

- may you please include `dip20` standard here?
- IIRC, comparing it like `token.standard === 'icrc' || ...`  makes the type-compiler flags bad values in this case. For example:

```typescript
// FLAGGED
token.standard === 'not-a-value-of-standard'

// NOT FLAGGED
['icrc', 'erc20', 'not-a-value-of-standard'].includes(token.standard)
```

AntonioVentilii: P.S. just remembered something: to have it centralized you could use the utils `isTokenIcrc` and `isTokenDip20`, or `isTokenIc`? or even create a new one

DenysKarmazynDFINITY: Good catch and idea! Switched to `isToken...` approach.

AntonioVentilii (approved):
LGTM tks!

AntonioVentilii (approved):

PR #7146, opened 2025-06-04
feat(frontend): validate contact name

CvetelinaAngelova:
# Motivation

We introduced a maximum character limit of 100 for the contact name.

# Changes

Added a derived reactive value isNameTooLong to check if the contact name exceeds 100 characters.
Rendered an animated error message using slide transition when the name exceeds the limit.
Updated the error message to use replacePlaceholders with the correct max length from a constant.

# Tests
Added a new unit test to ensure the error message is displayed when the name exceeds 100 characters.

Manual testing:
Snapshot and visual check for the slide transition animation of the error message.
Screenshot 2025-06-04 at 20 21 55

### DIFF
--- a/src/frontend/src/lib/components/address-book/ContactForm.svelte
+++ b/src/frontend/src/lib/components/address-book/ContactForm.svelte
@@ -1,9 +1,11 @@
 <script lang="ts">
 	import { notEmptyString } from '@dfinity/utils';
+	import { slide } from 'svelte/transition';
 	import InputText from '$lib/components/ui/InputText.svelte';
 	import { ADDRESS_BOOK_CONTACT_NAME_INPUT } from '$lib/constants/test-ids.constants';
 	import { i18n } from '$lib/stores/i18n.store';
 	import type { ContactUi } from '$lib/types/contact';
+	import { SLIDE_DURATION } from '$lib/constants/transition.constants';
 
 	let {
 		contact = $bindable(),
@@ -11,10 +13,12 @@
 		onSubmit = () => {}
 	}: { contact: Partial<ContactUi>; disabled?: boolean; onSubmit?: () => void } = $props();
 
-	let isValid = $derived(notEmptyString(contact?.name?.trim?.()));
+	const trimmedName = $derived(contact?.name?.trim?.() ?? '');
+	const isNameTooLong = $derived(trimmedName.length > 100);
+	const isValid = $derived(notEmptyString(trimmedName) && !isNameTooLong);
 
 	const handleKeydown = (event: KeyboardEvent): void => {
-		if (event.key === 'Enter') {
+		if (event.key === 'Enter' && isValid) {
 			event.preventDefault();
 			onSubmit();
 		}
@@ -37,5 +41,10 @@
 			autofocus={true}
 			{disabled}
 		/>
+		{#if isNameTooLong}
+			<p transition:slide={SLIDE_DURATION} class="pt-2 text-error-primary">
+				{$i18n.contact.error.name_too_long}</p
+			>
+		{/if}
 	</div>
 </form>
--- a/src/frontend/src/lib/components/address-book/ContactForm.svelte
+++ b/src/frontend/src/lib/components/address-book/ContactForm.svelte
@@ -17,10 +17,13 @@
 
 	const trimmedName = $derived((contact.name ?? '').trim());
 	const isNameTooLong = $derived(trimmedName.length > CONTACT_MAX_NAME_LENGTH);
-	const isValid = $derived(notEmptyString(trimmedName) && !isNameTooLong);
+
+	const assertValid = (): boolean => notEmptyString(trimmedName) && !isNameTooLong;
+
+	const isValid = $derived(assertValid());
 
 	const handleKeydown = (event: KeyboardEvent): void => {
-		if (event.key === 'Enter' && isValid) {
+		if (event.key === 'Enter' && assertValid()) {
 			event.preventDefault();
 			onSubmit();
 		}
--- a/src/frontend/src/lib/components/address-book/ContactForm.svelte
+++ b/src/frontend/src/lib/components/address-book/ContactForm.svelte
@@ -15,12 +15,12 @@
 		onSubmit = () => {}
 	}: { contact: Partial<ContactUi>; disabled?: boolean; onSubmit?: () => void } = $props();
 
-	const trimmedName = $derived(contact?.name?.trim?.() ?? '');
-	const isNameTooLong = $derived(trimmedName.length > 100);
+	const trimmedName = $derived((contact.name ?? '').trim());
+	const isNameTooLong = $derived(trimmedName.length > CONTACT_MAX_NAME_LENGTH);
 	const isValid = $derived(notEmptyString(trimmedName) && !isNameTooLong);
 
 	const handleKeydown = (event: KeyboardEvent): void => {
-		if (event.key === 'Enter' && isValid) {
+		if (event.key === 'Enter') {
 			event.preventDefault();
 			onSubmit();
 		}
@@ -31,7 +31,7 @@
 
 <svelte:window on:keydown={handleKeydown} />
 
-<form class="w-full" method="POST">
+<form class="w-full">
 	<div class="rounded-lg bg-brand-subtle-10 p-4 pb-6 pt-4 text-sm md:p-6 md:text-base md:font-bold">
 		{$i18n.contact.fields.name}
 		<InputText
--- a/src/frontend/src/lib/components/address-book/ContactForm.svelte
+++ b/src/frontend/src/lib/components/address-book/ContactForm.svelte
@@ -2,10 +2,12 @@
 	import { notEmptyString } from '@dfinity/utils';
 	import { slide } from 'svelte/transition';
 	import InputText from '$lib/components/ui/InputText.svelte';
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/address-book/ContactForm.svelte
```diff
@@ -1,20 +1,26 @@
...
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
 
 	let {
 		contact = $bindable(),
 		disabled = false,
 		onSubmit = () => {}
 	}: { contact: Partial<ContactUi>; disabled?: boolean; onSubmit?: () => void } = $props();
 
-	let isValid = $derived(notEmptyString(contact?.name?.trim?.()));
+	const trimmedName = $derived(contact?.name?.trim?.() ?? '');
```
peterpeterparker: Why `trim?.()`? Isn't `name` a string and therefore the prototype is always defined?

src/frontend/src/lib/components/address-book/ContactForm.svelte
```diff
@@ -1,20 +1,26 @@
...
+	const isNameTooLong = $derived(trimmedName.length > 100);
```
peterpeterparker: Did you maybe forgot to update the code after adding the constant `CONTACT_MAX_NAME_LENGTH` you declared in this PR?

src/frontend/src/lib/components/address-book/ContactForm.svelte
```diff
@@ -1,20 +1,26 @@
...
+	const isValid = $derived(notEmptyString(trimmedName) && !isNameTooLong);
 
 	const handleKeydown = (event: KeyboardEvent): void => {
-		if (event.key === 'Enter') {
+		if (event.key === 'Enter' && isValid) {
```
peterpeterparker: `handleKeydown` is imperative, isn't there a potential race condition where the flag is not set when the function is evaluated?

src/frontend/src/lib/components/address-book/ContactForm.svelte
```diff
@@ -25,7 +31,7 @@
 
 <svelte:window on:keydown={handleKeydown} />
 
-<form class="w-full">
+<form class="w-full" method="POST">
```
peterpeterparker: Why this change, how is it related to trimming the length of the input?

src/frontend/src/tests/lib/components/address-book/ContactForm.spec.ts
```diff
@@ -40,4 +40,18 @@ describe('ContactForm', () => {
 		// Check that validation failed
 		expect(component.isValid).toBeFalsy();
 	});
+
+	it('should show error when name exceeds max length', async () => {
```
peterpeterparker, line 44: Should we also add a test that assert that the error message is NOT displayed on load and when the length is shorter than 100? Maybe also a test that assert the msg is displayed and then become hidden?

src/frontend/src/lib/i18n/en.json
```diff
@@ -1089,7 +1089,8 @@
 		"error": {
 			"create": "Failed to create the contact",
 			"update": "Failed to update the contact",
-			"delete": "Failed to delete the contact"
+			"delete": "Failed to delete the contact",
+			"name_too_long": "Name must be $maxCharacters characters or fewer."
```
peterpeterparker: nitpick: maybe a sentence that sounds less technical, like "Please keep the name to $maxCharacters characters or fewer."

src/frontend/src/lib/components/address-book/ContactForm.svelte
CvetelinaAngelova: It's not related to trimming the input or validating its length.

src/frontend/src/tests/lib/components/address-book/ContactForm.spec.ts
CvetelinaAngelova: I’ve added 3 new tests to cover exactly that:
One to assert the error message is not shown on load
One to confirm it's not shown for valid input
And one to check that it appears when the input is too long and disappears once corrected

src/frontend/src/lib/components/address-book/ContactForm.svelte
CvetelinaAngelova: contact.name can be undefined due to Partial<ContactUi>, fallback to '' to safely trim

src/frontend/src/lib/i18n/en.json
CvetelinaAngelova: Changed

src/frontend/src/lib/components/address-book/ContactForm.svelte
CvetelinaAngelova: yes, thank you

src/frontend/src/lib/components/address-book/ContactForm.svelte
peterpeterparker: I understand that my point was about `trim?.()`, why so?

```
// incorrect
contact?.name?.trim?.()

// correct
contact?.name?.trim()
```

No?

If you agree, in a separate PR, can you fix the similar occurences that exist in the code base?

src/frontend/src/lib/components/address-book/ContactForm.svelte
peterpeterparker: I'm confused, what did you modified? It seems that now you removed the usage of `isValid` but I don't find checks anymore, is that possible?

src/frontend/src/lib/components/address-book/ContactForm.svelte
peterpeterparker: So, you reverted it, right?

src/frontend/src/tests/lib/components/address-book/ContactForm.spec.ts
peterpeterparker: Nice!

src/frontend/src/lib/components/address-book/ContactForm.svelte
CvetelinaAngelova: Yes, you're right — optional chaining on .trim?.() isn’t needed here.

My current change uses:
const trimmedName = $derived((contact.name ?? '').trim());
This safely handles the case where contact.name could be undefined (due to Partial<ContactUi>), while avoiding unnecessary optional chaining on the method itself.

src/frontend/src/lib/components/address-book/ContactForm.svelte
CvetelinaAngelova: Yes, that’s correct — it’s not needed in ContactForm since the validation has already been handled in EditContactNameStep via form?.isValid in the handleSave function. I removed the local check to avoid redundant logic.

src/frontend/src/lib/components/address-book/ContactForm.svelte
peterpeterparker: You did not answer my question. 

> can you fix the similar occurences that exist in the code base?

Can you take care of this or whom could? (in another PR)

src/frontend/src/lib/components/address-book/ContactForm.svelte
CvetelinaAngelova: Yes, I reverted it. Since it doesn’t correspond to the actual behavior here, I removed the method="POST" to keep the scope focused. While it can be good practice for native form submissions, it’s not relevant to the trimming or validation logic in this PR.

peterpeterparker: > Yes, I reverted it.

👍

src/frontend/src/lib/components/address-book/ContactForm.svelte
peterpeterparker: Right, however, while not stricto sensu linked to this PR, my comment about the potential race condition is still valid no? Isn't it possible that when `isValid` is evaluated by the imperative form submission, the derivation and evaluation of the state hasn't occured?

peterpeterparker:
> Side note not directly related to the PR but to the feature: is 100 characters enforced in the backend?

Yes it is. @CvetelinaAngelova found the implementation there https://github.com/dfinity/oisy-wallet/blob/e5a684144278bf8d6c3634cdad0040be9e22141c/src/shared/src/impls.rs#L28C3-L30C1

src/frontend/src/lib/components/address-book/ContactForm.svelte
CvetelinaAngelova: Yes, I found a similar usage in EditContactNameStep.svelte, and I’ll fix it in a separate PR to keep things clean and focused.

peterpeterparker: Thx!

src/frontend/src/lib/components/address-book/ContactForm.svelte
peterpeterparker: We discussed this online and @CvetelinaAngelova mentionned the following:

>  in Svelte 5 $derived values are computed synchronously and update immediately with their dependencies, so using isValid directly in the handler should be safe.

While I'm not 100% sure, because the forum is manually triggered by a keyboard event and therefore not sure if the submission might not happens before the derivation since we are using keydown, I agree, I think it's safe enough for now.

Maybe we can improve this in a later iteration by also imperatively validating the form on submission.

peterpeterparker (approved):
LGTM, thx

PR #7148, opened 2025-06-05
feat(frontend): Specific util to parse IC errors

AntonioVentilii:
# Motivation

We are going to use the util `replaceErrorFields` several times for IC errors. That means that it is useful to have a specific util with the IC keys to replace, and avoid repetition of code.

# Changes

- Create util `replaceIcErrorFields`.
- Use it in the code

# Tests

Included more tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7151, opened 2025-06-05
fix(frontend): Reset expanded contacts (addressBookStore) on address book modal open

daviddecentage:
# Motivation

Currently the expanded contacts persist throughout the whole session. We want to reset the store on modal open.

# Changes

Added optional callback that can be passed to the open modal function and reset the modal store in the callback for `openAddressBookModal`

# Tests

https://github.com/user-attachments/assets/72e922c3-8fc0-46c3-901e-67286948ed62


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7152, opened 2025-06-05
refactor(frontend): Change input of `ModalTokensListItem` to be `Token`

AntonioVentilii:
# Motivation

For component `ModalTokensListItem` it is ok to use directly `Token` type instead of `CardData`. Its usage will become more explicit.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7156, opened 2025-06-05
feat(frontend): Add property to always show a token in a group

AntonioVentilii:
# Motivation

A few times, we want to show only CK tokens or "native" tokens, for example when a token group has zero balance. For this, instead on relying on a propriety of the group, we can rely on the specific tokens. This gives us more flexibility to add more if we want.

# Changes

- Add new optional prop `alwaysShowInTokenGroup` to `TokenAppearanceSchema`.
- Set it to true for the "native" tokens.
- Adapt component `TokenGroupCard`.

# Tests

No changes in local replica.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7160, opened 2025-06-05
feat(build)!: Open Content Security Policy to all images

AntonioVentilii:
# Motivation

We will onboard NFTs as a new feature.

However, to do so, we need to first open our Content Security Policy (CSP) to load images from any external source (most of the NFTs images are stored on image-hosting web services).

As first step, we open it to any **SECURE** source, after having restricted the parts of the code where we know that we are using this policy: loading the metadata of the tokens (specifically the icon).


src/frontend/src/sol/components/tokens/SolAddTokenReview.svelte
```diff
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { isNullish } from '@dfinity/utils';
+	import { assertNonNullish, isNullish, nonNullish } from '@dfinity/utils';
```
AntonioVentilii:
```suggestion
	import { isNullish } from '@dfinity/utils';
```

PR #7161, opened 2025-06-05
fix(frontend): Correct explorer url for different eth networks

daviddecentage:
# Motivation

When we have the network switcher on `All networks` and look at a Base network transaction on the activity page, the explorer Url is wrong.
This is due to it being constructed from a derived store that depends on a network being selected.

# Changes

- Added a util function to get an explorer Url
- Adjusted the EthTransactionModal to use the new function instead of the derived store value

# Tests

Added tests for the utils function and verified on FE2

### DIFF
--- a/src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
+++ b/src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
@@ -5,6 +5,7 @@
 	import EthTransactionStatus from '$eth/components/transactions/EthTransactionStatus.svelte';
 	import { erc20Tokens } from '$eth/derived/erc20.derived';
 	import type { EthTransactionUi } from '$eth/types/eth-transaction';
+	import { getExplorerUrl } from '$eth/utils/eth.utils';
 	import { mapAddressToName } from '$eth/utils/transactions.utils';
 	import { ckEthMinterInfoStore } from '$icp-eth/stores/cketh.store';
 	import type { OptionCertifiedMinterInfo } from '$icp-eth/types/cketh-minter';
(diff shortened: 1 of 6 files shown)


src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
```diff
@@ -39,13 +39,13 @@
 	let { from, value, timestamp, hash, blockNumber, to, type } = $derived(transaction);
 
 	let explorerUrl: string | undefined = $derived(
-		notEmptyString(hash) ? `${$explorerUrlStore}/tx/${hash}` : undefined
+		notEmptyString(hash) ? `${getExplorerUrl({ token })}/tx/${hash}` : undefined
```
DenysKarmazynDFINITY: let's create a var for `getExplorerUrl({ token })` since we use it 3 time

src/frontend/src/eth/utils/eth.utils.ts
```diff
@@ -21,3 +24,16 @@ export const isSupportedEthToken = (
 
 export const isNotSupportedEthTokenId = (tokenId: TokenId): boolean =>
 	!isSupportedEthTokenId(tokenId);
+
+export const getExplorerUrl = ({
+	token,
+	network
+}: {
+	token?: Token | OptionToken;
```
DenysKarmazynDFINITY: just `OptionToken` should be enough

src/frontend/src/eth/utils/eth.utils.ts
```diff
@@ -21,3 +24,16 @@ export const isSupportedEthToken = (
...
+	network?: Network;
```
DenysKarmazynDFINITY, line 33: It seems we only use the `token` param. Is it planned to use `network` later?

DenysKarmazynDFINITY: discussed

DenysKarmazynDFINITY (approved):

src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
daviddecentage: done

src/frontend/src/eth/utils/eth.utils.ts
daviddecentage: updated

PR #7163, opened 2025-06-05
update

daviddecentage:
# Motivation

On IOS when you click on certain inputs, the page zooms to that input which feels not good for the UX.

# Changes

Changed viewport meta tag props to avoid this issue.

# Tests

Tested on FE2, didnt happen anymore on Stefans iPhone 15 and my 14 Pro Max

### DIFF
--- a/src/frontend/src/app.html
+++ b/src/frontend/src/app.html
@@ -42,7 +42,7 @@
 		<!-- CONTENT_SECURITY_POLICY -->
 
 		<meta charset="utf-8" />
-		<meta content="width=device-width, initial-scale=1, viewport-fit=cover" name="viewport" />
+		<meta content="width=device-width, initial-scale=1, maximum-scale=1.0, user-scalable=no, viewport-fit=cover" name="viewport" />
 
 		<title>{{OISY_NAME}}</title>
 		<link href="https://oisy.com" rel="canonical" />


DenysKarmazynDFINITY:
@daviddecentage I did some research and there appeared to be a solution to this problem: https://defensivecss.dev/tip/input-zoom-safari/. I tried on fe4, and setting font-size to 16px really fixed the issue. I suggest we use it instead of updating the meta to avoid a potential a11y downgrade. 

For testing purposes, I did the fix directly in the `Input` component but you can do it specifically for the AddressBook inputs to not affect the other forms of the app. 

Screenshot 2025-06-05 at 17 32 09

daviddecentage:
@DenysKarmazynDFINITY I adjusted the solution according to your comment and our offline discussion. This solution is certainly less dangerous and we can apply the font-size: 16px hack to any input when we need it.

src/frontend/src/lib/components/address/AddressForm.svelte
```diff
@@ -42,7 +42,7 @@
 	});
 </script>
 
-<div class="w-full">
+<div class="w-full" style="--input-font-size: 16px">
```
DenysKarmazynDFINITY: let's use a var here instead of the value - `var(--text-base)` should do the job

daviddecentage: done

PR #7168, opened 2025-06-05
chore(frontend): Removes explorer url derived store

daviddecentage:
# Motivation

Were removing an unneeded derived store to avoid future misuse of it. Since it depends on a page token we should not have a derived for it as it can be mistaken to be used anywhere, which will not work on pages without `pageToken` context.

# Changes

Removed derived store eplorerUrl
Removed tests for it
Adjusted all occurrences

# Tests

All affected UIs tested on FE2, working.


src/frontend/src/eth/components/core/EthWalletAddress.svelte
```diff
@@ -1,15 +1,22 @@
...
 	import { i18n } from '$lib/stores/i18n.store';
 	import { shortenWithMiddleEllipsis } from '$lib/utils/format.utils';
 
 	let explorerUrl: string | undefined;
 	$: explorerUrl = notEmptyString($ethAddress)
-		? `${$explorerUrlStore}/address/${$ethAddress}`
+		? `${getExplorerUrl({
+				network: nonNullish($selectedEvmNetwork)
+					? $selectedEvmNetwork
+					: ($selectedEthereumNetwork ?? DEFAULT_ETHEREUM_NETWORK)
```
DenysKarmazynDFINITY: I think we don't need `?? DEFAULT_ETHEREUM_NETWORK` since it fallbacks to this const inside `getExplorerUrl`

src/frontend/src/eth/components/tokens/EthTokenMenu.svelte
```diff
@@ -2,22 +2,22 @@
...
 
 	let explorerUrl: string | undefined;
 	$: explorerUrl =
-		$tokenStandard === 'erc20' && nonNullish($token)
-			? `${$explorerUrlStore}/token/${($token as Erc20Token).address}`
+		$tokenStandard === 'erc20' && nonNullish($pageToken)
+			? `${getExplorerUrl({ token: $pageToken })}/token/${($pageToken as Erc20Token).address}`
 			: notEmptyString($ethAddress)
-				? `${$explorerUrlStore}/address/${$ethAddress}`
+				? `${getExplorerUrl({ token: $pageToken })}/address/${$ethAddress}`
```
DenysKarmazynDFINITY, line 20: since you started using `$pageToken` in this part of the `if` statement, you need to move the `nonNullish($pageToken)` one level above:
```
if nonNullish($pageToken)
...all the stuff...
else
undefined
```

src/frontend/src/eth/components/core/EthWalletAddress.svelte
daviddecentage: good point, removing it

src/frontend/src/eth/components/tokens/EthTokenMenu.svelte
daviddecentage: updated it

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7170, opened 2025-06-05
feat(frontend): trim whitespace from Contact Name and Alias fields before submission

CvetelinaAngelova:
# Motivation

Users were able to enter contact names and aliases with unintended leading or trailing whitespace. This introduced inconsistency in stored data and caused avoidable backend validation errors. To improve data integrity and prevent issues during submission, input values are now trimmed before being sent to the backend.

# Changes

Trimmed contact.name before calling onAddContact or onSaveContact
Added a reactive $effect in AddressForm.svelte to trim address.label whenever it changes

# Tests

Unit tests added to verify that:
Names and labels with leading/trailing spaces are trimmed before backend submission
Whitespace-only inputs are treated as empty
 Confirmed that both EditContactNameStep and AddressForm behave as expected through test coverage


daviddecentage (approved):
LGTM, thx

PR #7174, opened 2025-06-06
feat(frontend): Do not match filter for empty strings in util

AntonioVentilii:
# Motivation

The util `isContactMatchingFilter` should not match a contact for empty filter. This logic should be handled by the possible consumer (if the component requires it). So, we want to keep the same logic but to extract it from the util.

# Changes

- Change util `isContactMatchingFilter` to return false for empty strings.
- Adapt components that uses it to still maintain the same behaviour: match the entire list if the filter is an empty string

# Tests

Added some tests to the util, and the rest must stay the same to verify that it worked.


src/frontend/src/lib/components/send/KnownDestinations.svelte
```diff
@@ -35,15 +35,17 @@
...
+		isEmptyString(destination)
+			? sortedKnownDestinations
+			: sortedKnownDestinations.filter(({ address }) =>
+					nonNullish(networkContacts?.[address])
+						? isContactMatchingFilter({
+								address,
+								contact: networkContacts[address],
+								filterValue: destination
+							})
+						: address === destination
```
DenysKarmazynDFINITY: with this change it will recognize the inputted destination only in case it fully matches with the address - shouldn't we keep `address.includes(destination)`?

AntonioVentilii: the idea is indeed that it does not matches partial addresses, it should match the entire one, otherwise it's risky, no? In a follow up, I will change it to match it lower-/upper-case depending on the network.

AntonioVentilii: CC @StefanBerger-DFINITY (just to be sure that it makes sense)

DenysKarmazynDFINITY: IMO, as a user, I would expect to see filtered addresses as I start typing something. Otherwise, it will mostly show "known destinations" only if someone is pasting a full address - I doubt someone will be typing the whole destination manually just to check if it's matching with an existing contact/previous address. It could be that it's even more risky to force users to type the whole address manually.

Also, from pure UX side the process would look a bit off: user opens send modal -> known destinations/contacts are shown -> user starts typing an address -> known destinations/contacts are gone -> user enters a full address -> something again appears on the screen.

AntonioVentilii: @DenysKarmazynDFINITY, I reverted that specific change after having asked @StefanBerger-DFINITY: he agrees with you that that is the expected behaviour

DenysKarmazynDFINITY: thanks for double-checking!

PR #7201, opened 2025-06-10
refactor(frontend): Remove `_count` suffix from track-event names

AntonioVentilii:
# Motivation

To simplify the event analytics, we can remove the suffix `_count`, since it is already implied.


BonomoAlessandro (approved):
LGTM

PR #7207, opened 2025-06-10
feat(frontend): TokenModalDeleteConfirmation component

DenysKarmazynDFINITY:
# Motivation

We need to implement a component for confirming token deletion. 

Screenshot 2025-06-10 at 11 34 20

### DIFF
--- a/src/frontend/src/lib/i18n/en.json
+++ b/src/frontend/src/lib/i18n/en.json
@@ -673,7 +673,7 @@
 			"twin_token": "Twin token",
 			"standard": "Standard",
 			"confirm_deletion_title": "Delete from the list of your tokens?",
-			"confirm_deletion_description": "This will remove the <span class=\"font-bold\">$token</span> from your $oisy_short token list. Your finds will remain untouched — you can re-add the token at any time."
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/i18n/en.json
```diff
@@ -671,7 +671,9 @@
 			"contract_address_copied": "Contract address copied to clipboard.",
 			"token_address_copied": "Token address copied to clipboard.",
 			"twin_token": "Twin token",
-			"standard": "Standard"
+			"standard": "Standard",
+			"confirm_deletion_title": "Delete from the list of your tokens?",
+			"confirm_deletion_description": "This will remove the <span class=\"font-bold\">$token</span> from your $oisy_short token list. Your finds will remain untouched — you can re-add the token at any time."
```
AntonioVentilii: small question: what does "Your finds will remain untouched" mean?

AntonioVentilii (approved):

src/frontend/src/lib/i18n/en.json
DenysKarmazynDFINITY: good catch, it should be "funds" - updating

PR #7210, opened 2025-06-10
test(frontend): Refactor tests for `Info` component to use stores

AntonioVentilii:
# Motivation

We want to include more tests linked to enabled/disabled networks. So, for now, we refactor the tests of `Info` component to use the stores instead of mocking the derived for the ICRC tokens.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7216, opened 2025-06-10
fix(frontend): Title for Edit contact Name step

CvetelinaAngelova:
# Motivation

Svelte components should not export derived state. For example:
<script>
   let title = $derived(...);
   export { title };
</script>
This pattern is non-standard and unsupported. The goal of this PR is to proactively refactor components to avoid such usage and ensure long-term stability with Svelte.

# Changes

Replaced export $derived(...) with $bindable<string>() in $props().
Used $effect to update title reactively based on editingContact.name.
In the parent, bound title as editContactNameTitle and displayed it in the modal's title slot.

# Tests

Verified modal title updates dynamically when editing contact names.
Confirmed no regressions via manual testing.

### DIFF
--- a/src/frontend/src/lib/components/address-book/AddressBookModal.svelte
+++ b/src/frontend/src/lib/components/address-book/AddressBookModal.svelte
@@ -165,6 +165,7 @@
 	let currentStepName = $derived(currentStep?.name as AddressBookSteps | undefined);
 	let previousStepName = $state<AddressBookSteps | undefined>();
 	let editContactNameStep = $state<EditContactNameStep>();
+	let editContactNameTitle = $state<string>($i18n.contact.form.add_new_contact);
 
 	let isDeletingContact = $state<boolean>(false);
 
@@ -301,7 +302,7 @@
 		{:else if currentStepName === AddressBookSteps.DELETE_CONTACT && nonNullish(currentContact)}
 			{replacePlaceholders($i18n.contact.delete.title, { $contact: currentContact.name })}
 		{:else if currentStepName === AddressBookSteps.EDIT_CONTACT_NAME && nonNullish(editContactNameStep)}
-			{editContactNameStep.title}
+			{editContactNameTitle}
 		{:else}
 			{currentStep?.title}
 		{/if}
@@ -397,6 +398,7 @@
 	{:else if currentStep?.name === AddressBookSteps.EDIT_CONTACT_NAME}
 		<EditContactNameStep
 			bind:this={editContactNameStep}
+			bind:title={editContactNameTitle}
 			contact={currentContact}
 			onAddContact={async (contact: Pick<ContactUi, 'name'>) => {
 				const createdContact = await callCreateContact({ name: contact.name });
--- a/src/frontend/src/lib/components/address-book/AddressBookModal.svelte
+++ b/src/frontend/src/lib/components/address-book/AddressBookModal.svelte
@@ -165,7 +165,7 @@
 	let currentStepName = $derived(currentStep?.name as AddressBookSteps | undefined);
 	let previousStepName = $state<AddressBookSteps | undefined>();
 	let editContactNameStep = $state<EditContactNameStep>();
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/address-book/AddressBookModal.svelte
```diff
@@ -165,6 +165,7 @@
 	let currentStepName = $derived(currentStep?.name as AddressBookSteps | undefined);
 	let previousStepName = $state<AddressBookSteps | undefined>();
 	let editContactNameStep = $state<EditContactNameStep>();
+	let editContactNameTitle = $state<string>($i18n.contact.form.add_new_contact);
```
peterpeterparker: sharing if interesting: here you don't necessarely need to express the type, TypeScript is able to infer the `<string>` - i.e. I think simply `$state($i18n...)` should do.

src/frontend/src/lib/components/address-book/EditContactNameStep.svelte
```diff
@@ -20,15 +20,17 @@
 		onClose: () => void;
 		isNewContact: boolean;
 		disabled?: boolean;
+		title?: string;
```
peterpeterparker, line 23: Is the prop optional - in that case the `$bindable<string>` should probably be set as `$bindable<string | undefined>` - or is the prop mandatory - in that case `title: string`?

src/frontend/src/lib/components/address-book/EditContactNameStep.svelte
```diff
@@ -20,15 +20,17 @@
...
 	}
 
 	let {
+		contact = {},
```
peterpeterparker: Out of curiosity, the order was change by an auto formatter?

AntonioVentilii: Curious about this one too: if it exists it would be interesting to include it in the formatter/linter to have more consistency

src/frontend/src/lib/components/address-book/AddressBookModal.svelte
CvetelinaAngelova: Definitely, you're right! I've removed the explicit type so it can rely on TypeScript's inference instead. Thanks for pointing that out!

src/frontend/src/lib/components/address-book/EditContactNameStep.svelte
CvetelinaAngelova: title is optional in the Props interface, so it should be $bindable<string | undefined>(). I've updated it accordingly. Thanks!

src/frontend/src/lib/components/address-book/EditContactNameStep.svelte
CvetelinaAngelova: Good catch — the change in order was unintentional in this case. I usually work with the eslint-plugin-sort-destructure-keys plugin that enforces alphabetical sorting, so I must’ve instinctively reordered it out of habit. Happy to revert it to the original order to avoid unnecessary diff noise.

CvetelinaAngelova: And I didn’t even do it correctly — it should actually be:
  contact = {},
  disabled = false,
  isNewContact,
  onAddContact,
  onClose,
  onSaveContact,
  title = $bindable<string>()

peterpeterparker: Thanks for the feedback. @AntonioVentilii it can be an interesting plugin if we don't use it yet, on the other side, I'm a bit afraid to introduce it as we rely a lot on destructing objects for anything. WDYT?

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/components/address-book/EditContactNameStep.svelte
AntonioVentilii: For sure it will be quite a lot of changes in the code eheh! But it is quite interesting, I was looking forwards to something similar, especially for Svelte 5 props

PR #7218, opened 2025-06-10
fix(frontend): Input zooms contact form IOS

daviddecentage:
# Motivation

The input fields still zoom in on focus of the contact name on IOS. 

# Changes

Applied font size fix to the contact form as well


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7220, opened 2025-06-10
refactor(frontend): remove_user_token params

DenysKarmazynDFINITY:
# Motivation

`remove_user_token` expects `UserTokenId` instead of `UserToken`.

### DIFF
--- a/src/frontend/src/lib/canisters/backend.canister.ts
+++ b/src/frontend/src/lib/canisters/backend.canister.ts
@@ -92,13 +92,10 @@ export class BackendCanister extends Canister<BackendService> {
 		return set_user_token(token);
 	};
 
-	removeUserToken = ({ chain_id, contract_address }: UserTokenId): Promise<void> => {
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/canisters/backend.canister.ts
```diff
@@ -91,10 +92,13 @@ export class BackendCanister extends Canister<BackendService> {
 		return set_user_token(token);
 	};
 
-	removeUserToken = ({ token }: { token: UserToken }): Promise<void> => {
+	removeUserToken = ({ chain_id, contract_address }: UserTokenId): Promise<void> => {
```
AntonioVentilii: can it be?

`removeUserToken = (params: UserTokenId): Promise<void> => {...`

DenysKarmazynDFINITY: yes, thanks! updated

PR #7223, opened 2025-06-10
feat(frontend): TokenModalDeleteConfirmation loading state

DenysKarmazynDFINITY:
# Motivation

We need to disable buttons in TokenModalDeleteConfirmation after user starts the removal actions. Also, added a test id to the submit button for future test cases (coming in the next PRs).


AntonioVentilii (approved):
LGTM tks! 

One thing: is it missing tests? I see you added the test ID

DenysKarmazynDFINITY:
The tests are coming in the following PR - added the test id here to make the next request a bit smaller.

PR #7224, opened 2025-06-10
feat(frontend): Util to get the case-sensitiveness of a network

AntonioVentilii:
# Motivation

It is useful to use a generic function to get the case-sensitiveness of a network, either by type of address or by network ID.

It will be used for the known destinations and the contacts too.

# Changes

- Create map of networks to address type.
- Create util `getCaseSensitiveness` that uses either network Id or address type.
- Use it in the code.

# Tests

Added tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7225, opened 2025-06-10
feat(frontend): Util to get case-safe known destinations

AntonioVentilii:
# Motivation

Some networks are case-sensitive (for example Solana). So, when we lit the known destinations, we should use the network/address property to verify that we respect the case-sensitiveness.

# Changes

- Create generic util to get a value from a record depending on network's case-sensitiveness.
- Create wrapper-utils for known destinations and for network contacts.
- Use utils in component to send tokens.

# Tests

Added tests and a practical example below:

Screenshot 2025-06-11 at 09 04 56
Screenshot 2025-06-11 at 09 05 02


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7237, opened 2025-06-11
feat(frontend): add delete functionality to TokenModal

DenysKarmazynDFINITY:
# Motivation

We can now introduce the token delete functionality to the TokenModal component. For now, it will only work for "user tokens" (erc20), but later will be extended with ICRC/SPL deletions (BE needs to have a new endpoint for that).

Screenshot 2025-06-11 at 11 21 26
Screenshot 2025-06-11 at 11 21 33

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokenModal.svelte
+++ b/src/frontend/src/lib/components/tokens/TokenModal.svelte
@@ -1,111 +1,103 @@
 <script lang="ts">
 	import { Modal } from '@dfinity/gix-components';
-	import { nonNullish } from '@dfinity/utils';
+	import { isNullish, nonNullish } from '@dfinity/utils';
 	import type { Snippet } from 'svelte';
-	import { isTokenErc20 } from '$eth/utils/erc20.utils';
-	import { isTokenIcrc, isTokenDip20 } from '$icp/utils/icrc.utils';
-	import List from '$lib/components/common/List.svelte';
-	import ModalHero from '$lib/components/common/ModalHero.svelte';
-	import ModalListItem from '$lib/components/common/ModalListItem.svelte';
-	import NetworkLogo from '$lib/components/networks/NetworkLogo.svelte';
-	import TokenLogo from '$lib/components/tokens/TokenLogo.svelte';
-	import ButtonDone from '$lib/components/ui/ButtonDone.svelte';
-	import ContentWithToolbar from '$lib/components/ui/ContentWithToolbar.svelte';
-	import Logo from '$lib/components/ui/Logo.svelte';
+	import { erc20UserTokensStore } from '$eth/stores/erc20-user-tokens.store';
+	import { isTokenErc20UserToken } from '$eth/utils/erc20.utils';
+	import { toUserToken } from '$icp-eth/services/user-token.services';
+	import { removeUserToken } from '$lib/api/backend.api';
+	import TokenModalContent from '$lib/components/tokens/TokenModalContent.svelte';
+	import TokenModalDeleteConfirmation from '$lib/components/tokens/TokenModalDeleteConfirmation.svelte';
+	import { authIdentity } from '$lib/derived/auth.derived';
+	import { nullishSignOut } from '$lib/services/auth.services';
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/components/tokens/TokenModal.svelte
```diff
@@ -1,111 +1,103 @@
...
+				chain_id,
+				contract_address,
+				identity: $authIdentity
+			});
+
+			erc20UserTokensStore.reset(token.id);
+
+			await onTokenDeleteSuccess();
+		}
+	};
```
AntonioVentilii, line 86: shouldn't this function go back at any point here? for example, if it fails, or if `isTokenErc20UserToken` is false?

src/frontend/src/lib/components/tokens/TokenModal.svelte
```diff
@@ -1,111 +1,103 @@
...
+
+	const onTokenDelete = async () => {
+		if (isNullish($authIdentity)) {
+			await nullishSignOut();
+			return;
+		}
+
+		if (isNullish(token)) {
+			return;
```
AntonioVentilii, line 57: same question (but i imagine this is an edge case)

src/frontend/src/lib/components/tokens/TokenModal.svelte
```diff
@@ -1,111 +1,103 @@
...
+
+	let showDeleteConfirmation = $state(false);
+	let loading = $state(false);
+
+	const onTokenDeleteSuccess = async () => {
+		loading = false;
+
+		modalStore.close();
+
+		await gotoReplaceRoot();
```
AntonioVentilii, line 37: we go to the root always? even if for example the user is in the main page with network filter?

src/frontend/src/lib/components/tokens/TokenModal.svelte
```diff
@@ -1,111 +1,103 @@
...
+			return;
+		}
+
+		// TODO: update this function to handle ICRC/SPL when BE supports removing custom tokens
+		if (isTokenErc20UserToken(token)) {
+			loading = true;
+
+			const { chain_id, contract_address } = toUserToken(token);
+
+			await removeUserToken({
```
AntonioVentilii: how is an error handled here?

src/frontend/src/lib/components/tokens/TokenModal.svelte
DenysKarmazynDFINITY: Very good points! I added error handling, so now in case of error the modal will go back to showing token data + an error toast will appear.

src/frontend/src/lib/components/tokens/TokenModal.svelte
DenysKarmazynDFINITY: yes, this is mostly for TS

src/frontend/src/lib/components/tokens/TokenModal.svelte
DenysKarmazynDFINITY: It is indeed planned to always redirect to the assets page, especially since using `history.back` might not work like expect it to if user opens a token page directly (w/o navigation through previous pages)

src/frontend/src/lib/components/tokens/TokenModal.svelte
DenysKarmazynDFINITY: added this part

src/frontend/src/lib/components/tokens/TokenModal.svelte
AntonioVentilii: Thank you for checking! And what happens if it is not an ERC20 token? imagine that the user lands on this somehow but with SPL token

DenysKarmazynDFINITY: The delete button will only be shown if a parent passes`isDeletable` prop as true. This will only be the case for EthereumTokenModal (next PR), so other tokens won't be affected.

AntonioVentilii (approved):
LGTM tks!

PR #7238, opened 2025-06-11
feat(frontend): Adds earning feature flag

daviddecentage:
# Motivation

In preparation of the new Earning feature, we add a feature flag that, if not set, will default to true for Local and Staging environments.

# Changes

Added feature flag

### DIFF
--- a/src/frontend/src/env/earning.ts
+++ b/src/frontend/src/env/earning.ts
@@ -1,5 +1,6 @@
 import { LOCAL, STAGING } from '$lib/constants/app.constants';
 
 // Enabled on Staging and Local if not set
+// Todo: remove once the feature has been completed
 export const EARNING_ENABLED =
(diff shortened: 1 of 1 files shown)


BonomoAlessandro (approved):
LGTM

small comment:
I guess this flag will be removed as soon as the feature is completed. So lets add a TODO that this flag needs to be removed. (see actions.env.ts)

src/frontend/src/env/earning.ts
```diff
@@ -0,0 +1,5 @@
+import { LOCAL, STAGING } from '$lib/constants/app.constants';
+
+// Enabled on Staging and Local if not set
```
daviddecentage, line 3:
```suggestion
// Enabled on Staging and Local if not set
// Todo: remove once the feature has been completed
```

PR #7239, opened 2025-06-11
refactor(frontend): upgrades Dropdown to svelte 5

BonomoAlessandro:
# Motivation

# Changes

- upgrades `Dropdown`


src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
```diff
@@ -56,9 +56,11 @@
 
 	<span class="hidden md:block">{$selectedNetwork?.name ?? $i18n.networks.chain_fusion}</span>
 
-	<svelte:fragment slot="title">{$i18n.networks.filter}</svelte:fragment>
+	{#snippet title()}
+		{$i18n.networks.filter}
+	{/snippet}
 
-	<div slot="items">
+	{#snippet items()}
```
AntonioVentilii, line 63: just to confirm: have you checked that removing the div does not affect the layout of the component? Looking at the code, I would say so, but just to be sure

src/frontend/src/lib/components/ui/Dropdown.svelte
```diff
@@ -1,15 +1,31 @@
...
 	import Responsive from '$lib/components/ui/Responsive.svelte';
 
-	export let disabled = false;
-	export let asModalOnMobile = false;
-	export let ariaLabel: string;
-	export let testId: string | undefined = undefined;
+	interface Props {
+		children?: Snippet;
+		title?: Snippet;
+		items?: Snippet;
```
AntonioVentilii: i would imagine this should be non-optional, no?

src/frontend/src/lib/components/ui/Dropdown.svelte
```diff
@@ -1,15 +1,31 @@
...
+	import type { Snippet } from 'svelte';
 	import DropdownButton from '$lib/components/ui/DropdownButton.svelte';
 	import Responsive from '$lib/components/ui/Responsive.svelte';
 
-	export let disabled = false;
-	export let asModalOnMobile = false;
-	export let ariaLabel: string;
-	export let testId: string | undefined = undefined;
+	interface Props {
+		children?: Snippet;
```
AntonioVentilii: same?

AntonioVentilii (approved):
LGTM tks

A few obs

src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
BonomoAlessandro: @AntonioVentilii yess i checked this one

src/frontend/src/lib/components/ui/Dropdown.svelte
BonomoAlessandro: @AntonioVentilii I guess you are right, this Snippet should be non-optional

PR #7241, opened 2025-06-11
refactor(frontend): upgrades ExternalLinkIcon to svelte 5

BonomoAlessandro:
# Motivation

# Changes

- upgrades `ExternalLinkIcon`


src/frontend/src/lib/components/ui/ExternalLinkIcon.svelte
```diff
@@ -1,18 +1,25 @@
 <script lang="ts">
+	import type { Snippet } from 'svelte';
 	import type { ButtonColorStyle } from '$lib/types/style';
 
-	export let href: string;
-	export let ariaLabel: string;
-	export let colorStyle: ButtonColorStyle = 'tertiary';
+	interface Props {
+		children?: Snippet;
+		onclick?: () => void;
```
AntonioVentilii: just for consistency: for non-built-in components, we are using camelCase. For example `onClick`, `onError`.

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/components/ui/ExternalLinkIcon.svelte
BonomoAlessandro: @AntonioVentilii This means you think it is better to do it like this?

<a
	...
	onclick={onClick}
	...

AntonioVentilii: i mean, just for consistency: the other components will all have `onError`, `onBack`, `onNext`, etc. So, we should keep `onClick`

PR #7242, opened 2025-06-11
refactor(frontend): Simplify component `SendContacts`

AntonioVentilii:
# Motivation

We will start selecting the values of a contacts record by case-sensitive keys. That means to use a specific utils. However, before that, we simplify a little the possible instances where it could be used.

For example in component `SendContacts`, we can use directly list `filteredNetworkContacts`, instead of iterating in it.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7243, opened 2025-06-11
feat(frontend): Earning nav utils

daviddecentage:
# Motivation

In preparation of the new Earning pages/features, we add the nav utils for the new routes.

# Changes

Added AppPaths `/earning` and `/earning/rewards`
Added util functions for path matching

# Tests

Added tests for new util functions


BonomoAlessandro (approved):
LGTM

PR #7244, opened 2025-06-11
refactor(frontend): Expand logic of component `KnownDestinations`

AntonioVentilii:
# Motivation

In component `KnownDestinations` we will change the logic of how the addresses are filtered, using case-sensitive logic. So, before that, we expand the logic maintaining the same flow, to be easier to modify after.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7249, opened 2025-06-11
feat(frontend): enable deletion of custom ERC20 tokens

DenysKarmazynDFINITY:
# Motivation

The last step of enabling the deletion of custom ERC20 tokens is to check whether a token is a non-default ERC20 and pass `isDeletable` prop to TokenModal accordingly.


AntonioVentilii (approved):
LGMT tks!

PR #7251, opened 2025-06-11
feat(frontend): Use network ID instead of address type in util `areAddressesEqual`

AntonioVentilii:
# Motivation

For util `areAddressesEqual` it is more versatile if we use the network ID.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7256, opened 2025-06-11
feat(frontend): Use case-sensitive partial comparison in component `KnownDestinations`

AntonioVentilii:
# Motivation

In component `KnownDestinations` we need to partially match the addresses based on network case-sensitiveness.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7262, opened 2025-06-12
feat(frontend): Use network case-sensitiveness in transaction list alias

AntonioVentilii:
# Motivation

We need to match the alias in the transaction list based on network case-sensitiveness.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7265, opened 2025-06-12
feat(frontend): Util to filter an address inside a contact

AntonioVentilii:
# Motivation

To avoid repetition of code we create an util to filter the address inside a contact. 

# Changes

- Util `filterAddressFromContact` will filter among the addresses of a contact, based on network case-sensitiveness (it uses existing code).
- Use the util in the code.

# Tests

Added tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7268, opened 2025-06-12
feat(frontend): Force util `getRecordValueByCaseSensitivity` to have non-nullish network

AntonioVentilii:
# Motivation

It doesn't make sense that the util `getRecordValueByCaseSensitivity` accepts nullish network ID: the scope of it is to match BY network ID.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7274, opened 2025-06-12
refactor(frontend): Contact name form

daviddecentage:
# Motivation

In order for a future feature where we want to have a create contact page in the address book, we clean up the current forms as they have been built differently.

# Changes

- Renamed to input because the form inside the component makes compositions difficult and moved the form to the page where the submit happens and the input is used
- No exported variables in the component anymore
- Cleaned up usage of form
- Removed manually event handler for key presses (submit on enter - now down with form submit)
- Cleaned up isValid vars
- Added a condition to not submit when disabled (it was possible to submit multiple times when hitting enter multiple times)

# Tests

Added a test host component and adjusted the tests accordingly


BonomoAlessandro (approved):
LGTM

PR #7276, opened 2025-06-12
feat(frontend): Contact name an address single step on save address

daviddecentage:
# Motivation

When we save an address (e.g Trasaction modal) we dont want to create a new contact in two steps. We add a new step called create contact that merges both Contact name and address forms so it can be done in a single step.

# Changes

- Added new step
- Added create contact step component

# Tests

Added test for the new step

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "@dfinity/oisy-wallet",
-	"version": "1.5.0",
+	"version": "1.5.1",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "@dfinity/oisy-wallet",
-			"version": "1.5.0",
+			"version": "1.5.1",
 			"license": "Apache-2.0",
 			"dependencies": {
 				"@dfinity/agent": "^2.4.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@dfinity/oisy-wallet",
-	"version": "1.5.0",
+	"version": "1.5.1",
 	"private": true,
 	"license": "Apache-2.0",
 	"repository": {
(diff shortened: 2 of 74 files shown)


peterpeterparker:
@daviddecentage @roman-nazaruk there is a bug in this PR or at least a warning. If you build the app you will notice this output:

```
transforming (226) node_modules/@sveltejs/kit/src/utils/routing.js06:39:54 [vite-plugin-svelte] src/frontend/src/lib/components/address-book/CreateContactStep.svelte:35:14 This reference only captures the initial value of `modalDataAddress`. Did you mean to reference it inside a derived instead?
https://svelte.dev/e/state_referenced_locally
33:   );
34:   let modalDataAddressUi = $state(
35:     (nonNullish(modalDataAddress)
                                    ^
36:       ? mapAddressToContactAddressUi(modalDataAddress)
37:       : {}) as ContactAddressUi
06:39:54 [vite-plugin-svelte] src/frontend/src/lib/components/address-book/CreateContactStep.svelte:36:34 This reference only captures the initial value of `modalDataAddress`. Did you mean to reference it inside a derived instead?
https://svelte.dev/e/state_referenced_locally
34:   let modalDataAddressUi = $state(
35:     (nonNullish(modalDataAddress)
36:       ? mapAddressToContactAddressUi(modalDataAddress)
                                                         ^
37:       : {}) as ContactAddressUi
38:   );
```

The root cause of the issue is the fact that a `derived` is used within a `state`, I might be wrong but I think it is not a valid pattern.

```
let modalData: AddressBookModalParams = $derived($modalStore?.data as AddressBookModalParams);
	let modalDataAddress: string | undefined = $derived(
		modalData?.entrypoint?.type === AddressBookSteps.SAVE_ADDRESS
			? modalData.entrypoint.address
			: undefined
	);
	let modalDataAddressUi = $state(
		(nonNullish(modalDataAddress)
			? mapAddressToContactAddressUi(modalDataAddress)
			: {}) as ContactAddressUi
	);
```

Can you fix the problem?



Capture d’écran 2025-06-13 à 06 45 06

Capture d’écran 2025-06-13 à 06 44 59

PR #7278, opened 2025-06-12
test(frontend): Remove flakiness from `ConvertWizard` tests

AntonioVentilii:
# Motivation

Similar to PR #7269.

The tests for `ConvertWizard` are causing a bit of flakiness since they have a debounce check for non-nullish identity (fee context components), and they sign out in case of nullish identity. Since it is debounced, it happens only if the tests takes longer that the interval, hence the flakiness.

To avoid it, we mock the identity and we mock the sign out function.


DenysKarmazynDFINITY (approved):
Thanks!

PR #7279, opened 2025-06-12
refactor(frontend): Contact name input default value isValid false

daviddecentage:
# Motivation

When the default value of isValid is true, the submit button briefly is enabled before validation triggers and sets it to disabled.

# Changes

isValid default value to false

# Tests

Tests adjusted


BonomoAlessandro (approved):
LGTM

PR #7280, opened 2025-06-12
feat(frontend): Back to transaction modal after saving address

daviddecentage:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/components/transactions/TransactionContactCard.svelte
+++ b/src/frontend/src/lib/components/transactions/TransactionContactCard.svelte
@@ -9,9 +9,17 @@
 	import { contacts } from '$lib/derived/contacts.derived';
 	import { AddressBookSteps } from '$lib/enums/progress-steps';
 	import { i18n } from '$lib/stores/i18n.store';
-	import { modalStore } from '$lib/stores/modal.store';
+	import { modalStore, type OpenTransactionParams } from '$lib/stores/modal.store';
 	import type { ContactUi } from '$lib/types/contact';
 	import { getContactForAddress } from '$lib/utils/contact.utils';
+	import type { IcTransaction, IcTransactionUi } from '$icp/types/ic-transaction';
+	import type { AnyTransactionUi } from '$lib/types/transaction';
+	import type { EthTransactionUi } from '$eth/types/eth-transaction';
+	import type { BtcTransactionUi } from '$btc/types/btc';
+	import type { SolTransactionUi } from '$sol/types/sol-transaction';
+	import { isIcToken } from '$icp/validation/ic-token.validation.js';
+	import { isSupportedEthToken } from '$eth/utils/eth.utils';
+	import { isSolanaToken } from '$sol/utils/token.utils';
 
 	interface Props {
 		type: 'send' | 'receive';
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/transactions/TransactionContactCard.svelte
```diff
@@ -35,6 +43,32 @@
 				})
 			: undefined
 	);
+
+	let modalStoreData = $derived($modalStore?.data as OpenTransactionParams<AnyTransactionUi>);
+
+	const getOnComplete = (modalStoreDataClosure: OpenTransactionParams<AnyTransactionUi>) => () => {
```
DenysKarmazynDFINITY: Since we tend to keep TransactionContactCard token-agnostic, I suggest we pass such a callback (e.g. `onAddressBookModalComplete` or something similar) from the parent components (IcTransactionModal, BtcTransactionModal, and so on).

src/frontend/src/lib/components/transactions/TransactionContactCard.svelte
```diff
@@ -35,6 +43,32 @@
...
+			});
+		} else if (isSolanaToken(modalStoreDataClosure.token)) {
+			modalStore.openSolTransaction({
+				id: Symbol(),
+				data: modalStoreDataClosure as OpenTransactionParams<SolTransactionUi>
+			});
+		} else {
+			modalStore.openBtcTransaction({
+				id: Symbol(),
+				data: modalStoreData as OpenTransactionParams<BtcTransactionUi>
```
DenysKarmazynDFINITY: This will be very likely changed after moving this code to related network components, but I noticed that you are using `modalStoreData` here when all other cases of this function proceed with `modalStoreDataClosure`.

src/frontend/src/lib/components/transactions/TransactionContactCard.svelte
daviddecentage: Genius way of not having to do the if else! I like it, will refactor it

src/frontend/src/lib/components/transactions/TransactionContactCard.svelte
daviddecentage: Thats a mistake indeed! Will fix

src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
```diff
@@ -47,6 +48,13 @@
 	let fromExplorerUrl: string | undefined = $derived(
 		nonNullish(explorerUrl) && nonNullish(from) ? `${explorerUrl}/address/${from}` : undefined
 	);
+
+	const onSaveAddressComplete = (data: OpenTransactionParams<AnyTransactionUi>) => {
```
DenysKarmazynDFINITY, line 52: Can we pass the data directly to avoid using params? something like
```
const onSaveAddressComplete = () => {
		modalStore.openBtcTransaction({
			id: Symbol(),
			data: $modalStore?.data as OpenTransactionParams<BtcTransactionUi>
		});
	};
```

DenysKarmazynDFINITY: And one more point - since we trigger this callback on cancel too, let's make the naming a bit more general

daviddecentage: We probably could, but the key to get this working is the closure that, by passing the modalData to the outer function, makes it a snapshot of what the modal data looked like. This is needed because once we're on the address book modal, the data from the transaction modal is gone.

This would in turn mean we would repeat this closure pattern for each transaction modal, which IMO is worse than having to pass the data as a param (which also is required since the reference to the original object will be broken when we pass it as a param - the key that makes this all work in the first place).

Regarding the naming, I would find it more confusing when the `TransactionContactCard` would have just `onComplete` since we dont know what will be completed. Also, closing the modal can also be considered the action as "completed". I'd leave it like that?

DenysKarmazynDFINITY: Thanks for the explanation - makes sense 👍

DenysKarmazynDFINITY (approved):
LGTM, thanks

P.S.: would be also great to add some basic tests for `modalData.entrypoint.onComplete()`, but since this PR is already quite a big one, we can do it separately.

PR #7284, opened 2025-06-13
feat(frontend): improves digits shown for values

BonomoAlessandro:
# Motivation

Currently, it is possible for a token's displayed value to appear as 0 if you own a very small amount, even if the actual value is not zero. In addition values smaller than 0.00000001 should be displayed as `< 0.00000001`.

# Changes

- improves displayed values

# Tests

**before:**
image

**after:**
image


src/frontend/src/tests/lib/utils/format.utils.spec.ts
```diff
@@ -110,6 +110,38 @@ describe('format.utils', () => {
...
+			expect(formatToken({ value: 120000001000000n, displayDecimals: 12 })).toBe('0.000120000001');
+			expect(formatToken({ value: 12000001000000n, displayDecimals: 12 })).toBe('0.000012000001');
+			expect(formatToken({ value: 1200001000000n, displayDecimals: 12 })).toBe('0.000001200001');
+			expect(formatToken({ value: 120001000000n, displayDecimals: 12 })).toBe('0.000000120001');
+			expect(formatToken({ value: 12001000000n, displayDecimals: 12 })).toBe('0.000000012001');
+			expect(formatToken({ value: 1201000000n, displayDecimals: 12 })).toBe('0.000000001201');
+		});
+
+		it('should format too small value with default displayDecimals', () => {
+			expect(formatToken({ value: 1200000000n })).toBe('< 0.00000001');
```
AntonioVentilii, line 143: can you add a test for `7000000000n` just to be sure? Should it be rendered as `'< 0.00000001'` or `'0.00000001'`?

src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -18,15 +19,33 @@ interface FormatTokenParams {
...
+
+	if (isNullish(displayDecimals) && leadingZeros >= MAX_DEFAULT_DISPLAY_DECIMALS) {
+		return '< 0.00000001';
+	}
+
+	const maxFractionDigits =
+		leadingZeros + 2 > MAX_DEFAULT_DISPLAY_DECIMALS
+			? MAX_DEFAULT_DISPLAY_DECIMALS
+			: leadingZeros + 2;
+	const minFractionDigits = isNullish(displayDecimals) ? DEFAULT_DISPLAY_DECIMALS : displayDecimals;
```
AntonioVentilii: can this work?

```suggestion
	const minFractionDigits = displayDecimals ?? DEFAULT_DISPLAY_DECIMALS;
```

src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -18,15 +19,33 @@ interface FormatTokenParams {
...
+
 	const formatted = (+res).toLocaleString('en-US', {
 		useGrouping: false,
-		maximumFractionDigits: displayDecimals,
-		minimumFractionDigits: trailingZeros ? displayDecimals : undefined
+		maximumFractionDigits: isNullish(displayDecimals)
+			? leadingZeros > 2
+				? maxFractionDigits
+				: DEFAULT_DISPLAY_DECIMALS
+			: displayDecimals,
```
AntonioVentilii: does this work?

```suggestion
		maximumFractionDigits: displayDecimals
			?? (leadingZeros > 2
				? maxFractionDigits
				: DEFAULT_DISPLAY_DECIMALS)
```

src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -18,15 +19,33 @@ interface FormatTokenParams {
...
+	const leadingZeros = match ? match[0].length - 2 : 0;
+
+	if (isNullish(displayDecimals) && leadingZeros >= MAX_DEFAULT_DISPLAY_DECIMALS) {
+		return '< 0.00000001';
+	}
+
+	const maxFractionDigits =
+		leadingZeros + 2 > MAX_DEFAULT_DISPLAY_DECIMALS
+			? MAX_DEFAULT_DISPLAY_DECIMALS
+			: leadingZeros + 2;
```
AntonioVentilii: do I read it correctly?

```suggestion
	const maxFractionDigits =Math.min(
		leadingZeros + 2 , MAX_DEFAULT_DISPLAY_DECIMALS)
```

src/frontend/src/tests/lib/utils/format.utils.spec.ts
BonomoAlessandro: @AntonioVentilii Yes i will add it. I would expect it to return `'< 0.00000001'`.

src/frontend/src/lib/utils/format.utils.ts
BonomoAlessandro: Yes you are right, this one is better

src/frontend/src/lib/utils/format.utils.ts
BonomoAlessandro: Yes this is also better

PR #7285, opened 2025-06-13
feat(backend)!: implement remove_custom_token

DenysKarmazynDFINITY:
# Motivation

BREAKING CHANGE: this PR includes a new BE method, therefore it contains did changes.

For deleting custom tokens (`icrc` and `spl`), we need to implement a new endpoint.


bitdivine (approved):
LGTM, thanks!

It is a bit odd that deletion needs the whole custom token in the argument, not just the key and maybe a version, but this is consistent with the existing code so not worth changing now, in my opinion.

PR #7298, opened 2025-06-13
refactor(frontend): upgrades InputText to svelte 5

BonomoAlessandro:
# Motivation

# Changes

- upgrades `InputText`


src/frontend/src/lib/components/ui/InputText.svelte
```diff
@@ -1,14 +1,27 @@
...
-	export let value = '';
-	export let name: string;
-	export let placeholder: string;
-	export let showResetButton = false;
-	export let required = true;
-	export let testId: string | undefined = undefined;
-	export let autofocus = false;
-	export let disabled: boolean | undefined = undefined;
+	interface Props {
+		value: string | undefined;
```
DenysKarmazynDFINITY: `value` was previously optional, so I suggest we keep it like it was: `value?: string`

src/frontend/src/lib/components/ui/InputText.svelte
```diff
@@ -1,14 +1,27 @@
...
+		placeholder: string;
+		showResetButton?: boolean;
+		required?: boolean;
+		testId?: string;
+		autofocus?: boolean;
+		disabled?: boolean;
+	}
+
+	let {
+		value = $bindable(),
```
DenysKarmazynDFINITY: default value is missing (`''`)

src/frontend/src/lib/components/ui/InputText.svelte
BonomoAlessandro: @DenysKarmazynDFINITY I don't think it makes sense, that value is optional. `InputText` is never used without `value`.

src/frontend/src/lib/components/ui/InputText.svelte
BonomoAlessandro: @DenysKarmazynDFINITY To be able to set the default value as `''` i needed to adjust two other files. In svelte 5 if you set a default value like this you are not allowed to call it with undefined. This is why we make sure that `value` is never undefined.

src/frontend/src/lib/components/address-book/InputContactName.svelte
```diff
@@ -16,7 +16,13 @@
 	}
 	let { contact = $bindable(), disabled = false, isValid = $bindable() }: Props = $props();
 
-	const trimmedName = $derived((contact.name ?? '').trim());
+	let name = $state(contact.name ?? '');
+
+	$effect(() => {
+		contact.name = name;
```
DenysKarmazynDFINITY, line 22: curious if it updates the `contact.name` with the initial value on the first render? if yes, it probably won't hurt to add a check `contact.name !== name`

DenysKarmazynDFINITY (approved):
One small q, approving already though

src/frontend/src/lib/components/address-book/InputContactName.svelte
BonomoAlessandro: It does update `contact.name` on the first render. But i don't think this is bad, because at the beginning `contact.name` is `undefined` and then will be set to an empty string.

PR #7300, opened 2025-06-13
refactor(frontend): upgrades MessageBox to svelte 5

BonomoAlessandro:
# Motivation

# Changes

- upgrades `MessageBox`


src/frontend/src/lib/components/ui/MessageBox.svelte
```diff
@@ -1,22 +1,25 @@
...
+	}
 
-	let closable = false;
-	$: closable = nonNullish(closableKey);
+	let { children, level = 'info', closableKey, styleClass, testId }: Props = $props();
 
-	let visible = true;
-	$: visible = isNullish(closableKey) || !shouldHideInfo(closableKey);
+	const closable = $derived(nonNullish(closableKey));
+	let visible = $state(isNullish(closableKey) || !shouldHideInfo(closableKey));
```
AntonioVentilii, line 22: this should be `$derived`, no?

BonomoAlessandro: @AntonioVentilii no because it's values is also changed in the `close` function

daviddecentage (approved):
LGTM, thx

src/frontend/src/lib/components/ui/MessageBox.svelte
AntonioVentilii: the previous logic was treating as `$derived` (reactive symbol `$:`), so we would be changing the logic in this case, no? `$state` will not make the same component. It is not that is wrong, just wanted to verify that there are no changes in the UI if we use `$state` instead of `$derived`

BonomoAlessandro: @AntonioVentilii No in this case it is not possible to use `$derived`. It would be possible to use `$state` in combination with `$effect` but in this case it is not needed.

AntonioVentilii (approved):

PR #7313, opened 2025-06-16
feat(frontend): uses welcome modal

BonomoAlessandro:
# Motivation

As soon as the sprinkles episode 4 is active we want to display our `Welcome` modal to the users when they register and login in the first time.

# Changes

- displays `Welcome` modal


src/frontend/src/lib/components/guard/RewardGuard.svelte
```diff
@@ -65,6 +68,26 @@
 				});
 			}
 		}
+
+		const season1Episode4Campaign = rewardCampaigns.find(
+			({ id }) => id === SPRINKLES_SEASON_1_EPISODE_4_ID
+		);
+		if (
+			nonNullish(lastTimestamp) &&
+			lastTimestamp === 0n &&
```
AntonioVentilii:
```suggestion
			lastTimestamp === ZERO &&
```

PR #7322, opened 2025-06-16
fix(frontend): Send contact no override if set

daviddecentage:
# Motivation

There is a slight issue in how we always select a contact manually when clicking next on the send flow destination step. If we select a contact manually, we still look up the contact as the select contact function also calls the "next" function.

# Changes

Added condition to only lookup if theres no currently selected contact.

# Tests

Tested manually


DenysKarmazynDFINITY (commented):
Thanks for the fix! 

Could you please also adjust the existing test suite (`should set selectedContact when a contact is selected`) to make sure that `selectedContact` is not being overwritten if it is already available

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7324, opened 2025-06-16
feat(frontend): add privacy mode indicator in activity

CvetelinaAngelova:
# Motivation

Users should be able to see when Privacy Mode is enabled during Activity. This helps improve transparency and UX by visually indicating that sensitive data is being hidden.

# Changes

Added a Privacy Mode indicator to the Activity view
Hooked into existing Privacy Mode state
Styled icon appropriately using Svelte SVG component
Supported dynamic color prop

# Tests
Manually verified that the Privacy Mode icon appears only when enabled
Screenshot 2025-06-16 at 19 02 04
Screenshot 2025-06-16 at 19 02 04

### DIFF
--- a/src/frontend/src/lib/components/hero/Balance.svelte
+++ b/src/frontend/src/lib/components/hero/Balance.svelte
@@ -55,7 +55,7 @@
 			/>
 		{:else}
 			<span class="flex items-center justify-center gap-2">
-				<IconEyeOff />
+				<IconEyeOff color="white" />
 				{$i18n.hero.text.hidden_balance}
 			</span>
 		{/if}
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/hero/Balance.svelte
```diff
@@ -55,7 +55,7 @@
 			/>
 		{:else}
 			<span class="flex items-center justify-center gap-2">
-				<IconEyeOff />
+				<IconEyeOff color="white" />
```
AntonioVentilii: just to confirm: is this adapted to the theme change? for example, in dark mode is still white? do we need it still white?

CvetelinaAngelova: Thanks for pointing that out! I've removed the explicit color prop from the Icon and added a theme-aware class instead, so it now adapts properly to light/dark mode.

CvetelinaAngelova:
@AntonioVentilii 
Thanks for the approval! I've just added some tests as suggested — happy to expand further if needed.

PR #7334, opened 2025-06-17
feat(frontend): update network contacts functionality

DenysKarmazynDFINITY:
# Motivation

In order to start displaying all contacts that have the same address (instead of just the first one), we need to change the scheme of `NetworkContact` - now, the dictionary key is a combination of address + contact id. This update leads to quite a few changes in the components and in the tests.

Note: the new keys are not 100% unique since the same address can be added multiple times for a single contact, but we specifically ignore this corner case.


AntonioVentilii (approved):
LGTM tks!

PR #7335, opened 2025-06-17
feat(frontend): Feature flag multilang

daviddecentage:
# Motivation

We are going to support multiple languages soon. Until the feature is completely ready we add a feature flag.

# Changes

Added feature flag


src/frontend/src/env/i18n.ts
```diff
@@ -0,0 +1,6 @@
+import { LOCAL, STAGING } from '$lib/constants/app.constants';
+
+// Enabled on Staging and Local if not set
+// Todo: remove once the feature has been completed
+export const I18N_ENABLED =
+	JSON.parse(import.meta.env.VITE_I18N_ENABLED ?? (STAGING || LOCAL)) === true;
```
AntonioVentilii: `parseBoolEnvVar`

PR #7336, opened 2025-06-17
feat(frontend): implements WelcomeModal

BonomoAlessandro:
# Motivation

We want to be able to display a `Welcome` modal to new users. The modal should contain some information about the current sprinkles episode.

# Changes

- adds `WelcomeModal`

# Tests

**New:**

image


src/frontend/src/lib/components/welcome/WelcomeModal.svelte
```diff
@@ -0,0 +1,61 @@
...
+				styleClass="h-[351px] w-full object-cover"
+				testId={WELCOME_MODAL_IMAGE_BANNER}
+			/>
+		</div>
+
+		<h3 class="my-3 text-center">{$i18n.welcome.title}</h3>
+		<Html text={$i18n.welcome.description} />
+
+		<div class="flex justify-center pt-4">
+			<div>
```
AntonioVentilii, line 34: is this `div` required?

BonomoAlessandro: @AntonioVentilii yes, this `div` is needed to handle the both buttons as one.

PR #7337, opened 2025-06-17
fix(frontend): Race condition in ICRC wallet worker with index canister

AntonioVentilii:
# Motivation

We have a curious race condition: the wallet worker for a typical ICRC token with Index canister starts normally and raises an error if the index canister raises an error.
In this case, we decide to restart the worker without index canister (only ledger canister), to guarantee that at least the balance is being fetched.

However, the error is raised right before the interval of the first worker is set. That means that the first worker is not stopped before starting the new one, since the timer is `undefined`: we send a message to stop the worker before starting the new one, but this message find the interval still not set.

There are a few solutions for it. The one coded in this PR simply does not `await` the execution of the sync-wallet function (that one that would raise the error): doing so, makes it so with the utmost probability, the interval is already set when the function raise the error.

# Alternative Solution

Another solution would be to modify the stopTimer function so that it runs in loop until it is executed at least once (and max to a certain number of try).

# Tests

Added tests.

### DIFF
--- a/src/frontend/src/lib/schedulers/scheduler.ts
+++ b/src/frontend/src/lib/schedulers/scheduler.ts
@@ -51,7 +51,10 @@ export class SchedulerTimer {
 		const execute = async () => await this.executeJob<T>({ identity, ...rest });
 
 		// We sync the cycles now but also schedule the update after wards
-		await execute();
+		// We don't wait for the execution to finish before scheduling the next one, because, in case of error, there could be a race condition:
+		// the error would trigger another execution that first try and stop the timer, but the timer is still not set, so it would not stop it.
+		// TODO: If this doesn't really fix it, another solution is to loop the function that stops the timer, so that it will run until it is triggered at least once (or until a max number of attempts).
+		execute();
 
 		// Support for features that implement the exact same pattern of a repetitive task but, are currently not scheduled for being refreshed automatically.
 		if (interval === 'disabled') {
--- a/src/frontend/src/tests/lib/schedulers/scheduler.spec.ts
+++ b/src/frontend/src/tests/lib/schedulers/scheduler.spec.ts
@@ -1,0 +1,307 @@
+import { SchedulerTimer } from '$lib/schedulers/scheduler';
+import * as authUtils from '$lib/utils/auth.utils';
+import { loadIdentity } from '$lib/utils/auth.utils';
+import { mockIdentity } from '$tests/mocks/identity.mock';
+import { expect, vi, type MockInstance } from 'vitest';
+
+describe('scheduler', () => {
+	describe('SchedulerTimer', () => {
+		let originalPostmessage: unknown;
+
+		let scheduler: SchedulerTimer;
+
+		const mockInterval = 10000;
+		const mockJob = vi.fn();
+		const mockData = { value: 'mock-data' };
+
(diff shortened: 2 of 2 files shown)


src/frontend/src/lib/schedulers/scheduler.ts
```diff
@@ -51,7 +51,10 @@ export class SchedulerTimer {
 		const execute = async () => await this.executeJob<T>({ identity, ...rest });
 
 		// We sync the cycles now but also schedule the update after wards
-		await execute();
+		// We don't wait for the execution to finish before scheduling the next one, because, in case of error, there could be a race condition:
```
peterpeterparker: Maybe ?

> We don't wait for the execution to finish before scheduling the next one, because, in case of error, there could be a race condition in case the UI submit a re-start of the worker.
> 
> For example, awaiting the current promise might take longer than receiving a new message from the UI requesting to stop the interval—which hasn’t been initialized yet, as it only starts after the promise resolves.
> 
> We observed this behavior empirically while debugging a live test instance deployed on the Internet Computer.

src/frontend/src/tests/btc/workers/btc-wallet.worker.spec.ts
```diff
@@ -205,6 +207,8 @@ describe('btc-wallet.worker', () => {
 				it('should postMessage with status of the worker', async () => {
 					await scheduler.start(startData);
 
+					await vi.advanceTimersByTimeAsync(WALLET_TIMER_INTERVAL_MILLIS - 100);
```
peterpeterparker: Maybe extract a mock with a comment to avoid repeating `WALLET_TIMER_INTERVAL_MILLIS - 100` and to provide some explanation?

src/frontend/src/tests/lib/schedulers/scheduler.spec.ts
```diff
@@ -0,0 +1,307 @@
+import { SchedulerTimer } from '$lib/schedulers/scheduler';
+import * as authUtils from '$lib/utils/auth.utils';
+import { loadIdentity } from '$lib/utils/auth.utils';
+import { mockIdentity } from '$tests/mocks/identity.mock';
+import { expect, vi, type MockInstance } from 'vitest';
+
+describe('scheduler', () => {
```
peterpeterparker, line 7: Nice!

src/frontend/src/lib/schedulers/scheduler.ts
AntonioVentilii: Yes, better! thank you!

src/frontend/src/tests/btc/workers/btc-wallet.worker.spec.ts
AntonioVentilii: Done! Have a look please! Any suggestion on a better explanation?

peterpeterparker: I meant generally speaking for all the tests:

```
// something.mock.ts
export const SOMETHING_MOCK = WALLET_TIMER_INTERVAL_MILLIS - 100;
```

peterpeterparker: But nevermind, looks good.

PR #7343, opened 2025-06-17
feat(frontend): implements new hangover campaign requirement

BonomoAlessandro:
# Motivation

We want to support the new `Hangover` requirement for the campaigns. With this requirement we show for how many days in a row you didn't win a reward.

# Changes

- supports `hangover` requirement

# Tests

**New:**

image


daviddecentage (approved):
LGTM

PR #7344, opened 2025-06-17
fix(frontend): Use convert context in component `SendInputDestination`

AntonioVentilii:
# Motivation

The component `SendInputDestination` is used both in the send flow and in the convert flow. That means that, since we use the send context to define the destination network, we need to add the convert context for the same tasks too.

### DIFF
--- a/src/frontend/src/lib/components/send/SendInputDestination.svelte
+++ b/src/frontend/src/lib/components/send/SendInputDestination.svelte
@@ -37,11 +37,9 @@
 
 	const convertContext = getContext<ConvertContext>(CONVERT_CONTEXT_KEY);
 
-	let sendTokenNetworkId: Readable<NetworkId> | undefined;
-	$: sendTokenNetworkId = nonNullish(sendContext) ? sendContext.sendTokenNetworkId : undefined;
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/send/SendInputDestination.svelte
```diff
@@ -49,7 +52,7 @@
 			getKnownDestination({
 				knownDestinations,
 				address: destination,
-				networkId: $sendTokenNetworkId
+				networkId: $sendTokenNetworkId ?? $destinationToken?.network.id
```
DenysKarmazynDFINITY: I saw there is `networkId` prop, can we maybe use it? 

Also, isn't it complaining "send context is undefined" when you open convert flow? And other way around

AntonioVentilii: the `networkId` is not necessarily the one we need: for example, for conversions we need the destination network Id, while the prop is given by the current network ID (for example the token we are trying to convert from).

For the send flow, it could be undefined apparently, so I am not 100% sure it is always the same as the one of the send context

WDYT?

DenysKarmazynDFINITY: Ok, yes, you're right - the naming is a bit confusing, it should have been something like `destinationNetworkId`.

PR #7351, opened 2025-06-17
refactor(frontend): refactors loadRewardResult function

BonomoAlessandro:
# Motivation

Soon some additional reward types will be added. To have a better designed function we will no longer return booleans like `receivedRewards` or `receivedJackpot` but the type of the received `Reward`.

# Changes

- refactors `loadRewardResult` function


daviddecentage (approved):
LGTM

PR #7355, opened 2025-06-17
feat(frontend): Remove Solana Testnet native token

AntonioVentilii:
# Motivation

We want to start integrating the [Solana RPC canister](https://github.com/dfinity/sol-rpc-canister). However, it still does not support Solana Testnet (not Devnet). After a small analysis, we decided to remove the support on OISY side too.

In this PR, we remove the native token for Solana Testnet.


BonomoAlessandro (approved):
LGTM

PR #7360, opened 2025-06-18
feat(frontend): Track errors when sending snapshots

AntonioVentilii:
# Motivation

We want to be aware when there is an issue in sending the user's snapshots.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7365, opened 2025-06-18
feat(frontend): Remove IC wallet pagination in all transactions list

AntonioVentilii:
# Motivation

For the activity page, it does not really make sense to paginate the loading of the missing transactions. pagination is useful for the scroller, but not when we first load the component.


AntonioVentilii:
This was a stupid idea: it will take too long to load all the transactions... and anyway now we are caching, so it should facilitate the first loading

PR #7371, opened 2025-06-18
feat(frontend): implement deleteIdbIcToken

DenysKarmazynDFINITY:
# Motivation

Similar to user tokens (erc20), we need a function to delete a particular IC token from IDB.

### DIFF
--- a/src/frontend/src/lib/api/idb-tokens.api.ts
+++ b/src/frontend/src/lib/api/idb-tokens.api.ts
@@ -1,5 +1,5 @@
 import { browser } from '$app/environment';
-import type { CustomToken, IcrcToken, UserToken } from '$declarations/backend/backend.did';
+import type { CustomToken, UserToken } from '$declarations/backend/backend.did';
 import { ETHEREUM_NETWORK_SYMBOL } from '$env/networks/networks.eth.env';
 import { ICP_NETWORK_SYMBOL } from '$env/networks/networks.icp.env';
 import { SOLANA_MAINNET_NETWORK_SYMBOL } from '$env/networks/networks.sol.env';
@@ -105,7 +105,7 @@ export const deleteIdbIcToken = async ({
 
 	const {
 		Icrc: { ledger_id: tokenToDeleteLedgerId }
-	} = tokenToDelete as { Icrc: IcrcToken };
+	} = tokenToDelete;
 
 	const currentTokens = await getIdbIcTokens(identity.getPrincipal());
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/tests/lib/api/idb-tokens.api.spec.ts
```diff
@@ -228,4 +235,44 @@ describe('idb-tokens.api', () => {
 			);
 		});
 	});
+
+	describe('deleteIdbIcToken', () => {
+		it('should delete provided IC token', async () => {
```
AntonioVentilii, line 246: can you add a test for nullish `identity` that calls `nullishSignOut`?

src/frontend/src/lib/api/idb-tokens.api.ts
```diff
@@ -87,3 +87,37 @@ export const deleteIdbEthToken = async ({
...
+
+	const { token: tokenToDelete } = token;
+
+	if (!('Icrc' in tokenToDelete)) {
+		return;
+	}
+
+	const {
+		Icrc: { ledger_id: tokenToDeleteLedgerId }
+	} = tokenToDelete as { Icrc: IcrcToken };
```
AntonioVentilii: i thought that the casting was not needed, since you checked above, but if the compiler complains, then all good

src/frontend/src/lib/api/idb-tokens.api.ts
```diff
@@ -87,3 +87,37 @@ export const deleteIdbEthToken = async ({
...
+
+	const currentTokens = await getIdbIcTokens(identity.getPrincipal());
+
+	if (nonNullish(currentTokens)) {
+		await setIdbIcTokens({
+			identity,
+			tokens: currentTokens.filter(({ token: savedToken }) =>
+				'Icrc' in savedToken
+					? Principal.from((savedToken as { Icrc: IcrcToken }).Icrc.ledger_id).toText() !==
```
AntonioVentilii: same here

src/frontend/src/tests/lib/api/idb-tokens.api.spec.ts
DenysKarmazynDFINITY: sure, done!

src/frontend/src/lib/api/idb-tokens.api.ts
DenysKarmazynDFINITY: you're completely right - removed

PR #7372, opened 2025-06-18
refactor(frontend): Make `burnStatus` util type-safe

AntonioVentilii:
# Motivation

When we map a ckBTC transaction, instead of raising a console error for unknown statuses, it is better to make it type-safe with the expected leftover types. In fact, we don't really act upon an unknown variant of the status.


BonomoAlessandro (approved):
LGTM

PR #7375, opened 2025-06-18
feat(frontend): Language selector dropdown

daviddecentage:
# Motivation

Here we add a dropdown to change the selected language.

# Changes

- Added `SUPPORTED_LANGUAGES` which is a based on the enum to extract key and value of the enum
- Added LanguageDropdown component

# Tests

image


src/frontend/src/lib/components/core/LanguageDropdown.svelte
```diff
@@ -0,0 +1,70 @@
...
+	import Dropdown from '$lib/components/ui/Dropdown.svelte';
+	import { SUPPORTED_LANGUAGES } from '$env/i18n';
+	import Button from '$lib/components/ui/Button.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { Languages } from '$lib/types/languages';
+	import ListItem from '$lib/components/common/ListItem.svelte';
+	import List from '$lib/components/common/List.svelte';
+	import IconCheck from '$lib/components/icons/IconCheck.svelte';
+
+	let dropdown: Dropdown | undefined;
```
BonomoAlessandro: shouldn't this be a state?

src/frontend/src/lib/components/core/LanguageDropdown.svelte
```diff
@@ -0,0 +1,70 @@
...
+		bind:this={dropdown}
+		ariaLabel={$i18n.core.alt.switch_language}
+		asModalOnMobile
+		buttonFullWidth
+		buttonBorder
+	>
+		{$i18n.languages[currentLang]}
+
+		{#snippet title()}
+			Select Language
```
BonomoAlessandro: This should be in the i18n i guess

src/frontend/src/lib/components/core/LanguageDropdown.svelte
daviddecentage: updated

src/frontend/src/lib/components/core/LanguageDropdown.svelte
daviddecentage: updated

PR #7378, opened 2025-06-18
feat(frontend): Menu language selector

daviddecentage:
# Motivation

Here we add the language selector component which will be shown in the user menu.

# Changes

Added icon
Added translation
Added menu item component

# Tests

image

image

image


src/frontend/src/lib/components/icons/IconLanguage.svelte
BonomoAlessandro, line 1: This component should be in svelte 5

daviddecentage: done

BonomoAlessandro (approved):
LGTM

PR #7383, opened 2025-06-18
refactor(frontend): User menu

daviddecentage:
# Motivation

We need to make some adjustments to the user menu because we want some things to be accessible from the landing page (logged out state). Additionally, we also prepare it for the addition of the language selector

# Changes

- Added `Open or create` button as on the landing page with licence agreement link
- Hide all menu entries but `Why OISY`, `Documentation` and `Support`
- Moved `onclick={hidePopover}` to the underlying div: We want to close the user menu when clicking any Button, but since in future we will be having the language selector dropdown and the light/dark mode switcher in there, we need to create a separate div adjacent to this one where we dont close the popover

# Tests

Adjusted existing tests and added test for logged out version

Logged out state:
image

Logged in state (unchanged):
image


src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -87,76 +90,89 @@
...
-		</ButtonMenu>
-
-		{#if isGold}
 			<ButtonMenu
-				ariaLabel={$i18n.navigation.alt.binance_qr_code}
-				testId={NAVIGATION_MENU_GOLD_BUTTON}
-				onclick={() => modalStore.openVipQrCode({ id: vipModalId, data: QrCodeType.GOLD })}
+				ariaLabel={$i18n.navigation.alt.address_book}
+				testId={NAVIGATION_MENU_ADDRESS_BOOK_BUTTON}
+				onclick={() => modalStore.openAddressBook({ id: addressModalId })}
```
BonomoAlessandro, line 135: You can call it the same way as everywhere else: `modalStore.openAddressBook(addressModalId)`

src/frontend/src/tests/lib/components/core/Menu.spec.ts
```diff
@@ -70,41 +80,59 @@ describe('Menu', () => {
 		});
 
 	it('renders the privacy mode menu item', async () => {
+		mockAuthSignedIn(true);
```
BonomoAlessandro: Why don't move this into the `beforeEach`?

daviddecentage: Because the last test needs `false` argument

BonomoAlessandro: But this means you can move it into the `beforeEach` and define it only in the function that needs the argument as `false`. Or am i wrong?

PR #7385, opened 2025-06-18
feat(frontend): allKnownIcrcTokensLedgerCanisterIds derived

DenysKarmazynDFINITY:
# Motivation

We need to detect whether an ICRC token is known to OISY to decide whether the "delete" button should be shown or not. The existing stores include customly added by user tokens, therefore we cannot rely on them. Instead, we implement a new derived which is a combination of all default + all env ICRC tokens.


src/frontend/src/icp/derived/icrc.derived.ts
```diff
@@ -13,13 +15,14 @@ import { sortIcTokens } from '$icp/utils/icrc.utils';
...
 import { mapDefaultTokenToToggleable } from '$lib/utils/token.utils';
+import { parseTokenId } from '$lib/validation/token.validation';
 import { nonNullish } from '@dfinity/utils';
 import { derived, type Readable } from 'svelte/store';
 
 /**
  * The list of Icrc default tokens - i.e. the statically configured Icrc tokens of Oisy + their metadata, unique ids etc. fetched at runtime.
  */
-const icrcDefaultTokens: Readable<IcToken[]> = derived(
+export const icrcDefaultTokens: Readable<IcToken[]> = derived(
```
AntonioVentilii: just to confirm, do you need this one exported?

DenysKarmazynDFINITY: good catch - it's just a leftover, removing

PR #7388, opened 2025-06-19
feat(frontend): improves hide token flow

BonomoAlessandro:
# Motivation

When a user applys a `network filter` and then disabled a token, the `network filter` is also removed. Even after hiding a token, the filter should still be active.

# Changes

- fixes hide token flow

# Tests

**before:**

https://github.com/user-attachments/assets/3a0f1bb2-f5cf-4f96-afb7-567389031f96

**after:**

https://github.com/user-attachments/assets/f7d23cec-6782-49bd-ac82-225b6d5393e2


AntonioVentilii (approved):
LGTM tks! AS discussed, before merging, let's call it all `fromRoute`

PR #7391, opened 2025-06-19
feat(frontend): improves delete token flow

BonomoAlessandro:
# Motivation

When a user applys a `network filter` and then deletes a token, the `network filter` is also removed. Even after deleting a token, the filter should still be active.

# Changes

- fixes delete token flow

# Tests

**before:**

https://github.com/user-attachments/assets/f6f15299-f8ad-48aa-828d-ad231e208692

**after:**

https://github.com/user-attachments/assets/d3ff4dde-f88d-42fd-b1d1-19023e7c552c


src/frontend/src/lib/components/tokens/TokenModal.svelte
```diff
@@ -63,9 +65,9 @@
 	const onTokenDeleteSuccess = async (deletedToken: Token) => {
 		loading = false;
 
-		close();
+		nonNullish(fromRoute) ? await back({ pop: nonNullish(fromRoute) }) : await gotoReplaceRoot();
 
-		await gotoReplaceRoot();
+		close();
```
AntonioVentilii: this was inverted, is that right?

BonomoAlessandro: @AntonioVentilii Yess we need to invert this order to have access to the from `fromRoute` value

AntonioVentilii: Let's revisit in the future how to properly guarantee that the modal closes

PR #7393, opened 2025-06-19
feat(frontend): add privacy mode notification

CvetelinaAngelova:
# Motivation

Improve user feedback when toggling Privacy Mode. Previously, enabling or disabling Privacy Mode updated state silently, with no confirmation or indication for the user. This PR introduces a toast notification to clearly indicate the new status of Privacy Mode.

# Changes

Added toasts message handlePrivacyToggle to notify the user:
"Privacy mode enabled" when hiding balances
"Privacy mode disabled" when showing balances again
Ensured localization is used for both toast messages

# Tests

Unit tests for Menu component updated to verify:
The correct toast message is shown when privacy mode is enabled
The correct toast message is shown when privacy mode is disabled
Verified manually in the UI that toggling the Privacy Mode shows the correct toast message
Screenshot 2025-06-19 at 13 29 19


src/frontend/src/lib/components/core/Menu.svelte
```diff
@@ -62,6 +63,23 @@
 
 	const hidePopover = () => (visible = false);
 
+	const handlePrivacyToggle = () => {
+		const nextValue = !$isPrivacyMode;
+
+		privacyModeStore.set({
+			key: 'privacy-mode',
+			value: { enabled: nextValue }
+		});
```
AntonioVentilii, line 74: just noticed that we use this in more places (and probably will use more. Maybe in another PR we could put it in a single service to be re-used (less code duplication)

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/components/core/Menu.svelte
CvetelinaAngelova: Great point — I agree this logic is a good candidate for abstraction.
I’ll create a follow-up PR to move it into a shared service so we can clean things up and avoid duplication.
This one will stay as-is to keep the changes focused.

CvetelinaAngelova: @AntonioVentilii  pr created
https://github.com/dfinity/oisy-wallet/pull/7420

PR #7398, opened 2025-06-19
refactor(frontend): Authenticate button

daviddecentage:
# Motivation

In preparation of the user menu refactoring we refactor the authenticate button.

# Changes

- Added prop to specify if the help link should be rendered too
- Fixed `text-center` prop and use `class:` syntax instead of strings with template literals

# Tests

Centered without help link:
image

Left with help link:
image


BonomoAlessandro (approved):
LGTM

PR #7399, opened 2025-06-19
style(frontend): Adjust HR styling

daviddecentage:
# Motivation

The HR tags look very ugly and not at all as specified in Figma. Here we adjust it to match the design.

# Changes

Fixed color and height of `<HR/>`s

# Tests

On user menu:
image

In Address book:
image


BonomoAlessandro (approved):
LGTM

PR #7403, opened 2025-06-19
refactor(frontend): Resolve text

daviddecentage:
# Motivation

For the upcoming task of making all Dapps and Rewards texts multilanguage, we are going to use the `resolveText` function to make a lookup for the translated texts.

In order to make our lifes easier in the future, we refactor the function to require a string path, not optional as before. The only place this function gets called (IcTransactionLabel) can do the nullish check there.

# Changes

- Refactored resolveText to require path string
- Adjusted usage of this function

# Tests

Functionally no difference


src/frontend/src/icp/components/transactions/IcTransactionLabel.svelte
```diff
@@ -1,20 +1,28 @@
...
+		fallback: string;
+		token: OptionToken;
+	}
 
-	$: fallback = fallback.charAt(0).toUpperCase() + fallback.slice(1);
+	const { label, fallback = '', token }: Props = $props();
 
-	let twinToken: Token | undefined;
-	$: twinToken = (token as OptionIcCkToken)?.twinToken;
+	const fallbackLabel: string = $derived(fallback.charAt(0).toUpperCase() + fallback.slice(1));
```
DenysKarmazynDFINITY: I think we silently agreed to use "let" for $derived/$state - let's update it in this component

DenysKarmazynDFINITY (approved):
small comment, otherwise LGTM

src/frontend/src/icp/components/transactions/IcTransactionLabel.svelte
daviddecentage: updated

PR #7407, opened 2025-06-19
feat(frontend): implement PoW worker services and message handling system 

DecentAgeCoder:
# Motivation

This PR establishes the communication layer between the main thread and PoW workers, enabling real-time progress updates during challenge solving.

# Changes

- Added `pow-protector-listener.services.ts` to handle syncing of Proof of Work (PoW) progress and next allowance data with dedicated functions: `syncPowProgress` and `syncPowNextAllowance`.
- Updated `worker.pow-protection.services.ts `to integrate the new listener functions for processing PoW-related worker messages (`syncPowProgress` and `syncPowNextAllowance`).

# Tests
Introduced new tests in worker.pow-protector.services.spec.ts) to test:
- Worker initialization and behavior (start, stop, trigger).
- Handling messages such as syncPowProgress and syncPowNextAllowance.
- Verifying data syncing to the respective Svelte stores.


src/frontend/src/icp/services/worker.pow-protection.services.ts
```diff
@@ -1,16 +1,54 @@
...
+				| PostMessageDataResponsePowProtectorNextAllowance
+				| PostMessageDataResponseError
+			>
+		>) => {
+			const { msg } = data;
+
+			switch (msg) {
+				case 'syncPowProgress': {
+					// Check if data.data exists and has proper structure
+					if (data.data && 'progress' in data.data) {
```
AntonioVentilii: `nonNullish`

src/frontend/src/icp/services/worker.pow-protection.services.ts
```diff
@@ -1,16 +1,54 @@
...
+						syncPowProgress({
+							data: data.data
+						});
+					}
+					return;
+				}
+				case 'syncPowNextAllowance': {
+					// Check if data.data exists and has proper structure
+					if (data.data && 'nextAllowanceMs' in data.data) {
```
AntonioVentilii: `nonNullish`

src/frontend/src/icp/services/pow-protector-listener.ts
```diff
@@ -0,0 +1,40 @@
...
+	powProtectoreProgressStore,
+	type PowProtectorNextAllowanceData,
+	type PowProtectorProgressData
+} from '$lib/stores/pow-protection.store';
+import type {
+	PostMessageDataResponsePowProtectorNextAllowance,
+	PostMessageDataResponsePowProtectorProgress
+} from '$lib/types/post-message';
+
+export interface PowProtectorWorkerInitResult {
```
AntonioVentilii: i would have left it in the type module, so that we can separate it and re-use it, no?

src/frontend/src/icp/services/worker.pow-protection.services.ts
DecentAgeCoder: I have added a assertNonNullish(data.data) to immediately abort and throw error in case data.data should be undefined/null. Otherwise a nonNullish would have to be added to every case statement.

src/frontend/src/icp/services/pow-protector-listener.ts
DecentAgeCoder: I have added a assertNonNullish(data.data) to immediately abort and throw error in case data.data should be undefined/null. Otherwise a nonNullish would have to be added to every case statement.

src/frontend/src/icp/services/worker.pow-protection.services.ts
DecentAgeCoder: I have added a assertNonNullish(data.data) to immediately abort and throw error in case data.data should be undefined/null. Otherwise a nonNullish would have to be added to every case statement.

src/frontend/src/icp/services/pow-protector-listener.ts
AntonioVentilii: no, i meant, the interface `PowProtectorWorkerInitResult`, should go in the module were it was already, no?

src/frontend/src/icp/services/worker.pow-protection.services.ts
```diff
@@ -1,16 +1,56 @@
...
+			data
+		}: MessageEvent<
+			PostMessage<
+				| PostMessageDataResponsePowProtectorProgress
+				| PostMessageDataResponsePowProtectorNextAllowance
+				| PostMessageDataResponseError
+			>
+		>) => {
+			const { msg } = data;
+			assertNonNullish(data.data);
```
AntonioVentilii: were will the error be handled?

src/frontend/src/icp/services/worker.pow-protection.services.ts
```diff
@@ -1,16 +1,56 @@
...
+
+			switch (msg) {
+				case 'syncPowProgress': {
+					// Check if data.data exists and has proper structure
```
AntonioVentilii: don't think this comment is necessary, right?

DecentAgeCoder: Good catch. I have removed the comment

src/frontend/src/icp/services/pow-protector-listener.ts
DecentAgeCoder: Yes that is correct the type and interface definition should be defined in pow-protector-listener.ts

src/frontend/src/icp/services/worker.pow-protection.services.ts
DecentAgeCoder: I have aligned the handling of the pow messages with the existing implementations

src/frontend/src/tests/icp/workers/pow-protection.worker.spec.ts
AntonioVentilii, line 1: this test module is for the web-worker, but the tests are for the worker that is part of the frontend, no? let's put it in a separate module, similar to what we do for `initIcrcWalletWorker`; plus let's add the tests for `onmessage`, so that we can verify that the stores `powProtectoreProgressStore` and `powProtectoreNextAllowanceStore` are updated

DecentAgeCoder: I have introduced `worker.pow-protector.services.spec.ts` to test the frontend worker including onmessage and store related tests.

AntonioVentilii: what i mean is that you are testing the module `pow-protector-listener.services.ts`, so, just for consistency, the tests specific to this module (like the ones that you added) should be in `pow-protector-listener.services.spec.ts`, no?

PR #7408, opened 2025-06-19
feat(frontend): add cycle management and PoW-aware loader services 

DecentAgeCoder:
# Motivation

This PR ensures that cycle validation happens before expensive PoW operations and that the system gracefully handles both PoW-enabled and traditional modes.

# Changes

- **Cycle Management**: New `hasRequiredCycles` function to check user's cycle allowance against minimum threshold
- **PoW-Aware Loading**: Updated and to conditionally handle PoW features based on feature flags `initLoader``initSignerAllowance`
- **Scheduler Integration**: Enhanced PoW protection scheduler to check cycles before starting challenges
- **Guard Updates**: Modified to respect PoW feature flags when initializing signer allowances `AddressGuard`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -68,7 +68,7 @@
 				"pem-file": "^1.0.1",
 				"postcss": "^8.5.6",
 				"prettier": "^3.5.3",
-				"prettier-plugin-motoko": "^0.11.0",
+				"prettier-plugin-motoko": "^0.11.3",
 				"prettier-plugin-organize-imports": "^4.2.0",
 				"prettier-plugin-svelte": "^3.4.0",
 				"prettier-plugin-tailwindcss": "^0.6.13",
@@ -2552,9 +2552,9 @@
 			}
 		},
 		"node_modules/@jridgewell/gen-mapping": {
-			"version": "0.3.12",
-			"resolved": "https://registry.npmjs.org/@jridgewell/gen-mapping/-/gen-mapping-0.3.12.tgz",
-			"integrity": "sha512-OuLGC46TjB5BbN1dH8JULVVZY4WTdkF7tV9Ys6wLL1rubZnCMstOhNHueU5bLCrnRuDhKPDM4g6sw4Bel5Gzqg==",
+			"version": "0.3.13",
+			"resolved": "https://registry.npmjs.org/@jridgewell/gen-mapping/-/gen-mapping-0.3.13.tgz",
+			"integrity": "sha512-2kkt/7niJ6MgEPxF0bYdQ6etZaA+fQvDcLKckhy1yIQOzaoKjBBjSj63/aLVjYE3qhRt5dvM+uUyfCg6UKCBbA==",
 			"license": "MIT",
 			"dependencies": {
 				"@jridgewell/sourcemap-codec": "^1.5.0",
@@ -2571,15 +2571,16 @@
 			}
 		},
 		"node_modules/@jridgewell/sourcemap-codec": {
-			"version": "1.5.4",
-			"resolved": "https://registry.npmjs.org/@jridgewell/sourcemap-codec/-/sourcemap-codec-1.5.4.tgz",
-			"integrity": "sha512-VT2+G1VQs/9oz078bLrYbecdZKs912zQlkelYpuf+SXF+QvZDYJlbx/LSx+meSAwdDFnF8FVXW92AVjjkVmgFw==",
+			"version": "1.5.5",
+			"resolved": "https://registry.npmjs.org/@jridgewell/sourcemap-codec/-/sourcemap-codec-1.5.5.tgz",
+			"integrity": "sha512-cYQ9310grqxueWbl+WuIUIaiUaDcj7WOq5fVhEljNVgRfOUhY9fy2zTvfoqWsnebh8Sl70VScFbICvJnLKB0Og==",
 			"license": "MIT"
 		},
 		"node_modules/@jridgewell/trace-mapping": {
-			"version": "0.3.29",
-			"resolved": "https://registry.npmjs.org/@jridgewell/trace-mapping/-/trace-mapping-0.3.29.tgz",
-			"integrity": "sha512-uw6guiW/gcAGPDhLmd77/6lW8QLeiV5RUTsAX46Db6oLhGaVj4lhnPwb184s1bkc8kdVg/+h988dro8GRDpmYQ==",
+			"version": "0.3.30",
+			"resolved": "https://registry.npmjs.org/@jridgewell/trace-mapping/-/trace-mapping-0.3.30.tgz",
+			"integrity": "sha512-GQ7Nw5G2lTu/BtHTKfXhKHok2WGetd4XYcVKGx00SjAk8GMwgJM3zr6zORiPGuOE+/vkc90KtTosSSvaCjKb2Q==",
+			"license": "MIT",
 			"dependencies": {
 				"@jridgewell/resolve-uri": "^3.1.0",
 				"@jridgewell/sourcemap-codec": "^1.4.14"
@@ -4789,9 +4790,9 @@
 			}
 		},
 		"node_modules/@testing-library/dom": {
-			"version": "10.4.0",
-			"resolved": "https://registry.npmjs.org/@testing-library/dom/-/dom-10.4.0.tgz",
-			"integrity": "sha512-pemlzrSESWbdAloYml3bAJMEfNh1Z7EduzqPKprCH5S341frlpYnUEW0H72dLxa6IsYr+mPno20GiSm+h9dEdQ==",
+			"version": "10.4.1",
+			"resolved": "https://registry.npmjs.org/@testing-library/dom/-/dom-10.4.1.tgz",
+			"integrity": "sha512-o4PXJQidqJl82ckFaXUeoAW+XysPLauYI43Abki5hABd853iMhitooc6znOnczgbTYmEP6U6/y1ZyKAIsvMKGg==",
 			"dev": true,
 			"license": "MIT",
 			"dependencies": {
@@ -4799,9 +4800,9 @@
 				"@babel/runtime": "^7.12.5",
 				"@types/aria-query": "^5.0.1",
 				"aria-query": "5.3.0",
-				"chalk": "^4.1.0",
 				"dom-accessibility-api": "^0.5.9",
 				"lz-string": "^1.5.0",
+				"picocolors": "1.1.1",
 				"pretty-format": "^27.0.2"
 			},
 			"engines": {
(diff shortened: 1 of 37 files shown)


src/frontend/src/icp/services/pow-protector-listener.ts
```diff
@@ -0,0 +1,32 @@
...
+	powProtectoreProgressStore,
+	type PowProtectorNextAllowanceData,
+	type PowProtectorProgressData
+} from '$lib/stores/pow-protection.store';
+import type {
+	PostMessageDataResponsePowProtectorNextAllowance,
+	PostMessageDataResponsePowProtectorProgress
+} from '$lib/types/post-message';
+
+export const syncPowProgress = ({
```
AntonioVentilii: these functions already exist, no?

src/frontend/src/lib/services/loader.services.ts
```diff
@@ -20,6 +22,29 @@ import type { ResultSuccess } from '$lib/types/utils';
...
+ * This asynchronous function verifies whether the user has sufficient cycles to proceed with further operations.
+ * It retrieves the user's identity and calculates the number of allowed cycles. If the number of allowed cycles
+ * meets or exceeds the defined threshold (`POW_MIN_CYCLES_THRESHOLD`), the function returns `true`. Otherwise,
+ * it performs necessary error handling and signs the user out in the event of insufficient cycles or any other
+ * encountered error.
+ *
+ * @returns {Promise<boolean>} A promise resolving to `true` if the required cycles are met or exceeded,
+ * otherwise `false` if insufficient cycles are detected or an error occurs during processing.
+ */
+export const handleInsufficientCycles = async (): Promise<boolean> => {
```
AntonioVentilii, line 37: where is this function used? it would be better to pass the identity store, instead of getting the value imperatively

src/frontend/src/icp/services/pow-protector.services.ts
```diff
@@ -1,4 +1,36 @@
...
+import type { OptionIdentity } from '$lib/types/identity';
+import { Principal } from '@dfinity/principal';
+import { assertNonNullish, hashText } from '@dfinity/utils';
+
+export const hasRequiredCycles = async ({
+	identity
+}: {
+	identity: OptionIdentity;
+}): Promise<boolean> => {
+	assertNonNullish(identity);
```
AntonioVentilii: should be a check for signing out here, no? I think we do this usually:

```ts
if (isNullish(identity)) {
  await nullishSignOut();
  return;
}
```
or

```ts
if (isNullish(identity)) {
  return;
}
```

src/frontend/src/icp/services/pow-protector.services.ts
```diff
@@ -1,4 +1,36 @@
...
+	assertNonNullish(SIGNER_CANISTER_ID);
```
AntonioVentilii: this check should not happen here, we usually don't check for nullish value of canister ID; or in general the error should be handled

src/frontend/src/icp/services/pow-protector.services.ts
DecentAgeCoder: There is already a function that exactly does that: `handleInsufficientCycles()`
Since 'hasRequiredCycles()' is directly called by the PowProtectionScheduler we musn't trigger a signout in the worker context (will also not compile since it would introduce a dependency to the store).

DecentAgeCoder: I have removed the the non nullish check for the SIGNER_CANISTER_ID

src/frontend/src/icp/services/pow-protector-listener.ts
DecentAgeCoder: Yes the code is duplicated. I have removed src/frontend/src/icp/services/pow-protector-listener.ts

src/frontend/src/lib/services/loader.services.ts
DecentAgeCoder: This function is called by the PowProtector.svelte in two different places. I have created the function to avoid code duplication. 
The initSignerAllowance(..) also does not pass the store and "gets the the value imperatively".

src/frontend/src/icp/services/pow-protector.services.ts
DecentAgeCoder: I have removed assertNonNullish(SIGNER_CANISTER_ID) and made identity mandatory (replaced type OptionIdentity with Identity)

src/frontend/src/icp/services/pow-protector.services.ts
```diff
@@ -1,4 +1,35 @@
...
+import type { OptionIdentity } from '$lib/types/identity';
+import { Principal } from '@dfinity/principal';
+import { assertNonNullish, hashText } from '@dfinity/utils';
+
+export const hasRequiredCycles = async ({
+	identity
+}: {
+	identity: OptionIdentity;
+}): Promise<boolean> => {
+	assertNonNullish(identity);
```
AntonioVentilii: how is this error handled? if you are sure that the `identity` passed is non-nullish, maybe better to change the parameter type to `identity: Identity`

src/frontend/src/lib/components/guard/AddressGuard.svelte
```diff
@@ -18,13 +19,14 @@
 	let signerAllowanceLoaded = false;
 
 	const loadSignerAllowanceAndValidateAddresses = async () => {
-		const { success: initSignerAllowanceSuccess } = await initSignerAllowance();
+		if (!POW_FEATURE_ENABLED) {
```
AntonioVentilii: this is quite a big change, i would prefer that it goes in another PR, since it is not related here, I imagine

src/frontend/src/lib/services/loader.services.ts
```diff
@@ -110,12 +136,13 @@ export const initLoader = async ({
 
 	// We are loading the addresses from the backend. Consequently, we aim to animate this operation and offer the user an explanation of what is happening. To achieve this, we will present this information within a modal.
 	setProgressModal(true);
+	if (!POW_FEATURE_ENABLED) {
```
AntonioVentilii: same

src/frontend/src/icp/services/pow-protector.services.ts
```diff
@@ -1,4 +1,35 @@
...
+import { IC_CYCLES_LEDGER_CANISTER_ID } from '$env/networks/networks.icrc.env';
+import { allowance } from '$icp/api/icrc-ledger.api';
+import { getIcrcSubaccount } from '$icp/utils/icrc-account.utils';
+import { BACKEND_CANISTER_PRINCIPAL, SIGNER_CANISTER_ID } from '$lib/constants/app.constants';
+import { POW_MIN_CYCLES_THRESHOLD } from '$lib/constants/pow.constants';
+import type { OptionIdentity } from '$lib/types/identity';
+import { Principal } from '@dfinity/principal';
+import { assertNonNullish, hashText } from '@dfinity/utils';
+
+export const hasRequiredCycles = async ({
```
AntonioVentilii: let's add tests for this service too

DecentAgeCoder: I have added tests in pow-protector.services.spec.ts for testing solvePowChallenge

src/frontend/src/icp/services/pow-protector.services.ts
DecentAgeCoder: I have changed the type identity to type Identity

src/frontend/src/lib/components/guard/AddressGuard.svelte
DecentAgeCoder: I have reverted the change and moved it to a follow-up PR

src/frontend/src/lib/services/loader.services.ts
DecentAgeCoder: I have reverted this change and moved it to a follow-up PR

src/frontend/src/lib/services/loader.services.ts
```diff
@@ -40,14 +65,15 @@ export const initSignerAllowance = async (): Promise<ResultSuccess> => {
 	try {
 		const { identity } = get(authStore);
 
-		await allowSigning({ identity });
+		await allowSigning({
+			identity
+		});
```
AntonioVentilii:
```suggestion
		await allowSigning({	identity});
```

src/frontend/src/lib/services/loader.services.ts
AntonioVentilii: but, is it used in this PR? I cannot find any reference

src/frontend/src/icp/services/pow-protector.services.ts
```diff
@@ -1,5 +1,30 @@
...
+import { POW_MIN_CYCLES_THRESHOLD } from '$lib/constants/pow.constants';
+import type { Identity } from '@dfinity/agent';
+import { Principal } from '@dfinity/principal';
 import { hashText } from '@dfinity/utils';
 
+export const hasRequiredCycles = async ({ identity }: { identity: Identity }): Promise<boolean> => {
+	const allowanceResult = await allowance({
+		identity,
+		certified: false,
+		ledgerCanisterId: IC_CYCLES_LEDGER_CANISTER_ID,
```
AntonioVentilii: will this work in staging/beta/other envs? The variable is another I imagine (`STAGING_CYCLES_LEDGER_CANISTER_ID`, `BETA_...`)

DecentAgeCoder: Introduced environment dependent variable CYCLES_LEDGER_CANISTER_ID

src/frontend/src/lib/services/loader.services.ts
DecentAgeCoder: The function can be found in PowProtector.svelte in PR: 
https://github.com/dfinity/oisy-wallet/pull/7409

src/frontend/src/icp/schedulers/pow-protection.scheduler.ts
```diff
@@ -44,6 +44,11 @@ export class PowProtectionScheduler implements Scheduler<PostMessageDataRequest>
 	 * @throws Errors if any step with no specific error handling in the sequence fails.
 	 */
 	private requestSignerCycles = async ({ identity }: SchedulerJobData<PostMessageDataRequest>) => {
+		if (await hasRequiredCycles({ identity })) {
+			// we can skip the PoW process if the user has not enough cycles
```
AntonioVentilii: isn't it the opposite?

```suggestion
			// We can skip the PoW process if the user has enough cycles
```

src/frontend/src/icp/services/pow-protector.services.ts
DecentAgeCoder: I have commited and integrated the generic variable CYCLES_LEDGER_CANISTER_ID

AntonioVentilii (approved):
LGTM tks

PR #7410, opened 2025-06-19
feat(frontend): Add code from library `ic-pub-key`

AntonioVentilii:
# Motivation

We will start to derive the addresses in the frontend using the library [ic-pub-key](https://github.com/dfinity/ic-pub-key). However the library is not public yet (it is under publication process).

So, in the interest of time, we simply copy the code from it and paste it here, to use it and derive the addresses.

NOTE: there is a feature flag that enable/disable this new feature in case something goes awry.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "@dfinity/oisy-wallet",
-	"version": "1.5.2",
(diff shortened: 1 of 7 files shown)


bitdivine (approved):
LGTM.  Thanks!

PR #7417, opened 2025-06-20
refactor(frontend): Extract function to get the BTC address

AntonioVentilii:
# Motivation

Since we are going to derive the BTC address in the frontend directly, we need to first refactor the services to handle the feature.


src/frontend/src/btc/services/btc-address.services.ts
```diff
@@ -52,6 +53,24 @@ const bitcoinMapper: Record<
...
 };
 
+const getBtcAddress = async ({
+	identity,
+	network
+}: {
+	identity: OptionIdentity;
+	network: BitcoinNetwork;
+}): Promise<BtcAddress> => {
+	if (FRONTEND_DERIVATION_ENABLED) {
```
roman-nazaruk, line 63: I'm not sure, but I think it's better to leave a comment suggesting to add this logic only if FRONTEND_DERIVATION_ENABLED. Right now, it doesn't make much sense — we're just using the import, which is currently useless in this file.

AntonioVentilii: ah yes yes! it is actually coming in the next PR (since now this change is like it does not exist, it just continues)

PR #7423, opened 2025-06-20
feat(backend): add image to contacts types

SYInacta:
# Motivation

# Changes

# Tests


AntonioVentilii (changes requested):
Something happened with the CI that runs the bindings: they seem wrong...

src/backend/tests/it/contacts.rs
```diff
@@ -147,11 +150,12 @@ fn test_create_contact_should_fail_with_leading_and_trailing_whitespace_name() {
         "create_contact",
         CreateContactRequest {
             name: "   Leading Whitespace".to_string(),
+            image: None,
         },
     );
     assert!(
-        wrapped_result.is_err(),
-        "Leading whitespace should be rejected"
```
bitdivine: Ditto

src/backend/tests/it/contacts.rs
```diff
@@ -108,36 +109,38 @@ fn test_create_contact_should_succeed_with_valid_name() {
...
         caller,
         "create_contact",
         CreateContactRequest {
             name: "  ".to_string(),
+            image: None,
         },
     );
     assert!(
-        wrapped_result.is_err(),
-        "String with multiples whitespaces should be rejected"
```
bitdivine: This change seems as if it's not related to storing images.  If there is a change in policy regarding whitespace, we would normally recommend that the corresponding change, like this, is put into a dedicated PR so that it can be reverted independently of the other changes, if needed.

src/backend/tests/it/contacts.rs
```diff
@@ -160,11 +164,12 @@ fn test_create_contact_should_fail_with_leading_and_trailing_whitespace_name() {
         "create_contact",
         CreateContactRequest {
             name: "Trailing Whitespace   ".to_string(),
+            image: None,
         },
     );
     assert!(
-        wrapped_result.is_err(),
-        "Trailing whitespace should be rejected"
```
bitdivine: Ditto

src/backend/tests/it/contacts.rs
```diff
@@ -173,11 +178,12 @@ fn test_create_contact_should_fail_with_leading_and_trailing_whitespace_name() {
...
         CreateContactRequest {
             name: "   Leading and Trailing Whitespace   ".to_string(),
+            image: None,
         },
     );
     assert!(
-        wrapped_result.is_err(),
-        "Leading and trailing whitespace should be rejected"
+        wrapped_result.is_ok(),
+        "Leading and trailing whitespace should be accepted"
```
bitdivine: Ditto

src/backend/tests/it/contacts.rs
```diff
@@ -478,7 +489,7 @@ fn test_update_contact_should_succeed_with_valid_data() {
 }
 
 #[test]
-fn test_update_contact_should_fail_with_whitespace_name() {
+fn test_update_contact_accepts_whitespace_name() {
     let pic_setup = setup();
     let caller: Principal = Principal::from_text(CALLER).unwrap();
```
bitdivine: Ditto

src/shared/src/impls.rs
```diff
@@ -541,6 +543,16 @@ impl Validate for Contact {
...
+        // Validate name does not have leading/trailing whitespace
+        validate_string_whitespace_padding(&self.name, "Contact.name")?;
+
+        // Reject names that are empty or only whitespace
+        if self.name.trim().is_empty() {
+            return Err(Error::msg(
+                "Contact.name cannot be empty or whitespace only",
+            ));
+        }
+
```
bitdivine: Ditto

src/shared/src/impls.rs
```diff
@@ -561,6 +573,17 @@ impl Validate for ContactAddressData {
     }
 }
 
+impl Validate for ContactImage {
+    fn validate(&self) -> Result<(), Error> {
+        // For now, we don't have specific validation rules for ContactImage
+        // In the future, we might want to validate:
+        // - Image size limits
+        // - MIME type consistency with the actual data
+        // - Image dimensions
```
bitdivine, line 584: Where is the image size checked, if not here?

It would be goo dto have a test that large images are rejected.  I don't see one in this PR.

src/shared/src/types/contact.rs
```diff
@@ -1,20 +1,63 @@
...
+
+    /// Returns a list of all supported MIME types
+    #[must_use]
+    pub fn supported_types() -> &'static [&'static str] {
+        &["image/jpeg", "image/png", "image/gif", "image/webp"]
+    }
+}
+
+/// Memory usage threshold (80%) above which new images cannot be added
+pub const MEMORY_USAGE_THRESHOLD: f64 = 0.8;
```
bitdivine, line 41: Do I understand correctly that the limit is defined but not enforced in this PR?  Will it be enforced in a follow-up PR?

bitdivine (approved):
The whitespace changes need to be split out into another PR and the code needs tests, but apart from that the code looks fine.  Thanks!

src/backend/tests/it/contacts.rs
SYInacta: Resolved.

src/backend/tests/it/contacts.rs
SYInacta: Resolved.

src/backend/tests/it/contacts.rs
SYInacta: Resolved.

src/shared/src/impls.rs
SYInacta: Resolved.

src/shared/src/types/contact.rs
SYInacta: Yes.

src/backend/tests/it/contacts.rs
SYInacta: Resolved.

src/shared/src/impls.rs
SYInacta: This will come in second PR.

src/backend/tests/it/contacts.rs
SYInacta: Resolved.

PR #7443, opened 2025-06-23
fix(frontend): Dapps translations tags

daviddecentage:
# Motivation

We missed to translate the tags for dapps. Here we add them to the english i18n file

# Changes

Changed occurrences for the tags to resolveText

# Tests

image


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7467, opened 2025-06-23
feat(backend)!: Add Arbitrum network

roman-nazaruk:
Adding Arbitrum networks to the types of the snapshots.


AntonioVentilii (approved):
Already approving since the CIs will solve/warn all missing pieces

src/frontend/src/lib/derived/user-networks.derived.ts
```diff
@@ -103,6 +103,15 @@ export const userNetworks: Readable<UserNetworks> = derived(
 				return POLYGON_AMOY_NETWORK_ID;
 			}
```
AntonioVentilii:
```suggestion
```

PR #7472, opened 2025-06-24
Update de.json

StefanBerger-DFINITY:
# Motivation

We introduce German as 2nd supported language.

# Changes

- translate all rewards related texts to german

# Tests
No functional changes. Just german texts instead of english.

### DIFF
--- a/src/frontend/src/lib/i18n/de.json
+++ b/src/frontend/src/lib/i18n/de.json
@@ -184,123 +184,123 @@
 		"categories": {
 			"defi": "DeFi",
 			"social_media": "Soziale Medien",
-			"verifiable_credentials": "Verifizierbare Nachweise",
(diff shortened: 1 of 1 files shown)


src/frontend/src/env/dapp-descriptions.json
```diff
@@ -93,7 +93,7 @@
 	{
 		"id": "uniswap",
 		"name": "dapps.descriptions.uniswap.name",
-		"oneLiner": "dapps.descriptions.uniswap.name",
+		"oneLiner": "dapps.descriptions.uniswap.one_liner",
```
AntonioVentilii, line 96: Good catch!

StefanBerger-DFINITY: was actually easy to see....from outside

image

PR #7479, opened 2025-06-24
feat(frontend): update WarningBanner component

DenysKarmazynDFINITY:
# Motivation

To make the WarningBanner component more versatile and re-usable, we extend it with some additional snippets as well as adjust its styles.

Screenshot 2025-06-24 at 12 07 26


daviddecentage (approved):
LGTM

PR #7485, opened 2025-06-24
feat(frontend): Menu theme selector

daviddecentage:
# Motivation

In preparation of the new user menu theme selector, we add a few components for it.

# Changes

- Added menu item component that renders the selector
- Added ThemeSelectorButton

# Tests

Functionality taken from current theme selector cards on settings page.

image
image
image

### DIFF
--- a/src/frontend/src/lib/components/core/ThemeSelectorButton.svelte
+++ b/src/frontend/src/lib/components/core/ThemeSelectorButton.svelte
@@ -6,7 +6,7 @@
 	interface Props {
 		label: string;
 		selected: boolean;
-		testId: string | undefined;
+		testId?: string;
 		onclick: MouseEventHandler<HTMLButtonElement>;
 		children: Snippet;
 	}


src/frontend/src/lib/components/core/MenuThemeSelector.svelte
```diff
@@ -0,0 +1,74 @@
...
+
+	// Here we just update the local variable above to update the selected card
+	const updateSelectedCard = (themeStore: ThemeStoreData) => {
+		selectedCard = isNullish(localStorage.getItem(THEME_KEY))
+			? THEME_SYSTEM
+			: (themeStore ?? THEME_SYSTEM);
+	};
+
+	$effect(() => {
+		updateSelectedCard($themeStore);
```
DenysKarmazynDFINITY, line 38: Can you please explain why we need to update `slectedCard` 2 times - the first update happens in `selectTheme` on button click, the second one - right after `themeStore` value gets updated. Can we spare the second update?

daviddecentage: As I remember it had to do when you select some theme coming from the system theme and back, then the system card wouldnt display being selected because the system theme being selected means the storage is empty.

I really just copied the logic from what it was before, and remember this was nessesary to get it working correctly.

Edit: I will reroute the tests for the original components to these new ones when we remove them in a later PR (#7486) then we can be sure it all works.

DenysKarmazynDFINITY: Ah ok, didn't realize it's an existing code. Let's proceed with it then.

src/frontend/src/lib/components/core/MenuThemeSelector.svelte
```diff
@@ -0,0 +1,74 @@
...
+	});
+</script>
+
+<div class="flex flex-row justify-between pl-3">
```
DenysKarmazynDFINITY: `flex-row` is the default value, so we can remove it

src/frontend/src/lib/components/core/MenuThemeSelector.svelte
```diff
@@ -0,0 +1,74 @@
...
+	<span class="mr-3 flex min-w-32 flex-row items-center text-tertiary">
```
DenysKarmazynDFINITY: same

src/frontend/src/lib/components/core/ThemeSelectorButton.svelte
```diff
@@ -0,0 +1,26 @@
+<script lang="ts">
+	import type { Snippet } from 'svelte';
+	import type { MouseEventHandler } from 'svelte/elements';
+	import Button from '$lib/components/ui/Button.svelte';
+
+	interface Props {
+		label: string;
+		selected: boolean;
+		testId: string | undefined;
```
DenysKarmazynDFINITY: I would make `testId` optional: `testId?: string`

src/frontend/src/lib/components/core/MenuThemeSelector.svelte
daviddecentage: updated

src/frontend/src/lib/components/core/MenuThemeSelector.svelte
daviddecentage: updated

src/frontend/src/lib/components/core/ThemeSelectorButton.svelte
daviddecentage: updated

PR #7486, opened 2025-06-24
refactor(frontend): Remove settings theme selector

daviddecentage:
# Motivation

# Changes

# Tests


src/frontend/src/lib/components/settings/ThemeSelector.svelte
```diff
@@ -1,68 +0,0 @@
-<script lang="ts">
-	import { themeStore, Theme, type ThemeStoreData } from '@dfinity/gix-components';
-	import { isNullish } from '@dfinity/utils';
-	import ThemeSelectorCard from '$lib/components/settings/ThemeSelectorCard.svelte';
-	import Img from '$lib/components/ui/Img.svelte';
-	import { THEME_SELECTOR_CARD } from '$lib/constants/test-ids.constants';
```
AntonioVentilii, line 6: can we remove this one too? or is it used elsewhere?

src/frontend/src/tests/lib/components/settings/MenuThemeSelector.spec.ts
```diff
@@ -94,70 +94,62 @@ describe('ThemeSelector', () => {
...
+		expect(getByTestId(testIdSystem).getAttribute('class')).toContain('primary');
 
 		await fireEvent.click(getByTestId(testIdDark));
 
-		expect(getByTestId(testIdDark).getAttribute('aria-checked')).toBe(JSON.stringify(true));
+		expect(getByTestId(testIdDark).getAttribute('class')).toContain('primary');
 
 		await fireEvent.click(getByTestId(testIdLight));
 
-		expect(getByTestId(testIdLight).getAttribute('aria-checked')).toBe(JSON.stringify(true));
```
AntonioVentilii, line 146: maybe for another PR, double-check that the new theme selector is a11y accessible (even adding `aria-checked` if necessary)

daviddecentage: But buttons dont take aria-checked attribute, thats why I changed the check for the tests to test for the primary button class

PR #7488, opened 2025-06-24
feat(frontend): Default language from browser

daviddecentage:
# Motivation

We want the browser language to be selected as the default language when loading translations. If the language has not been manually overridden we check if the brower language is a supported OISY app language and set it to that if it is.

# Changes

- Added `getDefaultLang()` utils function
- Made use of `getDefaultLang()` in i18n store

# Tests

Added tests for `getDefaultLang()`

### DIFF
--- a/src/frontend/src/lib/components/manage/AddTokenByNetwork.svelte
+++ b/src/frontend/src/lib/components/manage/AddTokenByNetwork.svelte
@@ -1,14 +1,13 @@
 <script lang="ts">
-	import { Dropdown, DropdownItem } from '@dfinity/gix-components';
 	import { isNullish, nonNullish, notEmptyString } from '@dfinity/utils';
 	import { createEventDispatcher } from 'svelte';
 	import { fade } from 'svelte/transition';
 	import EthAddTokenForm from '$eth/components/tokens/EthAddTokenForm.svelte';
 	import IcAddTokenForm from '$icp/components/tokens/IcAddTokenForm.svelte';
 	import type { AddTokenData } from '$icp-eth/types/add-token';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/utils/i18n.utils.ts
```diff
@@ -78,3 +79,14 @@ export const mergeWithFallback = ({
 
 	return merged;
 };
+
+export const getDefaultLang = () => {
```
DenysKarmazynDFINITY: Can you please add the return type?

src/frontend/src/lib/utils/i18n.utils.ts
```diff
@@ -78,3 +79,14 @@ export const mergeWithFallback = ({
...
+	const browserLocale = new Intl.Locale(navigator.language);
+	const browserLanguage = Object.keys(Languages).filter(
```
DenysKarmazynDFINITY: How about we make it a bit shorter:
```
return Object.keys(Languages).find(...) ?? Languages.ENGLISH;
```

AntonioVentilii (approved):
Thank you!!!

AntonioVentilii (changes requested):
Ops, too hasty, ignore my review, the one from @DenysKarmazynDFINITY is superseding. When he approves I unblock

src/frontend/src/lib/utils/i18n.utils.ts
daviddecentage: updated

src/frontend/src/lib/utils/i18n.utils.ts
daviddecentage: updated

PR #7490, opened 2025-06-24
feat(frontend): Implement canister method class for SOL RPC canister

AntonioVentilii:
# Motivation

Since we included the Solana RPC canister, we can create the class that holds the necessary methods for the frontend.

For now, we only create the `getAccountInfo` method: it is the first one that we integrate because it is the most used, so most visibility for potential errors.

# Changes

- Create error class specific for the SOL RPC canister, mapping the variants of type `RpcError` of the canister.
- Create constants for base configurations of the canister methods.
- New class `SolRpcCanister` with method `getAccountInfo` that queries the SOL RPC canister
- The methods (as the others that we will include after) takes as input some base parameters that we don't really modify. So for now we hard-code them in the class.

# Note

The SOL RPC canister will ALWAYS require to specify which Solana network we are querying. So, we create some base mapper to transform our `SolanaNetworkType` type into the canister network type.

# Tests

Created tests.


src/frontend/src/sol/canisters/sol-rpc.canister.ts
```diff
@@ -0,0 +1,99 @@
...
+			},
+			idlFactory: idlFactorySolRpc,
+			certifiedIdlFactory: idlCertifiedFactorySolRpc
+		});
+
+		return new SolRpcCanister(canisterId, service, certifiedService);
+	}
+
+	private baseParams: {
+		encoding: [] | [GetAccountInfoEncoding];
```
DenysKarmazynDFINITY, line 62: Shall we create a general type for nullable values? I imagine we can re-use it in different places

src/frontend/src/tests/lib/canisters/signer.canister.spec.ts
```diff
@@ -528,7 +528,7 @@ describe('signer.canister', () => {
 		});
 	});
 
-	describe('btc_caller_send', () => {
+	describe('sendBtc', () => {
```
DenysKarmazynDFINITY, line 531: I think the naming was correct before, or am I missing something?

DenysKarmazynDFINITY (approved):
2 minor things, otherwise LGTM 💪

src/frontend/src/sol/canisters/sol-rpc.canister.ts
AntonioVentilii: yes, i will for sure do it (I was surprised it didn't exist yet), but directly in `ic-js` (where `toNullable` is coded), WDYT?

DenysKarmazynDFINITY: Sounds good, thanks!

src/frontend/src/tests/lib/canisters/signer.canister.spec.ts
AntonioVentilii: well, theoretically we are testing the method `sendBtc` of the class `SignerCanister`, just to keep consistent with the other tests. P.S. I can do it in another PR

DenysKarmazynDFINITY: all good, let's merge it then

PR #7491, opened 2025-06-24
feat(frontend): extract network dropdown from AddTokenByNetwork

DenysKarmazynDFINITY:
# Motivation

We need to extract the dropdown code from `AddTokenByNetwork` so it can be re-used in the Edit ICRC token modal.

### DIFF
--- a/src/frontend/src/lib/components/manage/AddTokenByNetworkDropdown.svelte
+++ b/src/frontend/src/lib/components/manage/AddTokenByNetworkDropdown.svelte
@@ -1,0 +1,39 @@
+<script lang="ts">
+	import { Dropdown, DropdownItem } from '@dfinity/gix-components';
+	import { nonNullish } from '@dfinity/utils';
+	import Value from '$lib/components/ui/Value.svelte';
+	import { i18n } from '$lib/stores/i18n.store';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/manage/AddTokenByNetworkDropdown.svelte
```diff
@@ -0,0 +1,39 @@
...
+				{/each}
+			</Dropdown>
+		</div>
+	{/snippet}
+</Value>
+
+<style lang="scss">
+	:global(.disabled div.select) {
+		--input-border-size: 0;
+	}
```
AntonioVentilii, line 38: arent's some classes missing here? the ones form the original component:

```css
<style lang="scss">
	.hidden {
		display: none;
	}
	.network {
		--disable-contrast: rgba(0, 0, 0, 0.5);
	}
</style>
```

DenysKarmazynDFINITY: I actually removed them after doing some manual tests - they don't seem to have any effect. Also, `hidden` is already available as a tailwind class, so I'm not sure why it was there.

PR #7501, opened 2025-06-25
feat(frontend): API function for method `getAccountInfo` of SOL RPC canister

AntonioVentilii:
# Motivation

We now expose the `getAccountInfo` of the Solana RPC Canister as explicit function, as we usually do.

### DIFF
--- a/src/frontend/src/sol/api/sol-rpc.api.ts
+++ b/src/frontend/src/sol/api/sol-rpc.api.ts
@@ -1,0 +1,58 @@
+import type { AccountInfo } from '$declarations/sol_rpc/sol_rpc.did';
+import { SOL_RPC_CANISTER_ID } from '$lib/constants/app.constants';
+import type { SolAddress } from '$lib/types/address';
+import type { CanisterApiFunctionParams } from '$lib/types/canister';
+import { SolRpcCanister } from '$sol/canisters/sol-rpc.canister';
+import type { SolanaNetworkType } from '$sol/types/network';
+import { Principal } from '@dfinity/principal';
+import { assertNonNullish, isNullish } from '@dfinity/utils';
+
+let canister: SolRpcCanister | undefined = undefined;
+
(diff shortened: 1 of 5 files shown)


src/frontend/src/sol/api/sol-rpc.api.ts
```diff
@@ -0,0 +1,58 @@
...
+	const { getAccountInfo } = await solRpcCanister({ identity });
+
+	const accountInfo = await getAccountInfo(rest);
+
+	if (isNullish(accountInfo)) {
+		return accountInfo;
+	}
+
+	if (!('json' in accountInfo.data)) {
+		return accountInfo;
```
DenysKarmazynDFINITY: maybe let's make it just `return`, otherwise the return object won't be either `undefined` or a proper `AccountInfo`

src/frontend/src/sol/api/sol-rpc.api.ts
```diff
@@ -0,0 +1,58 @@
...
+}: CanisterApiFunctionParams<{
+	address: SolAddress;
+	network: SolanaNetworkType;
+}>): Promise<AccountInfo | undefined> => {
+	const { getAccountInfo } = await solRpcCanister({ identity });
+
+	const accountInfo = await getAccountInfo(rest);
+
+	if (isNullish(accountInfo)) {
+		return accountInfo;
```
DenysKarmazynDFINITY: Here I'd also keep just `return;`

src/frontend/src/sol/api/sol-rpc.api.ts
AntonioVentilii: good point! Changed

src/frontend/src/sol/api/sol-rpc.api.ts
AntonioVentilii: done!

PR #7502, opened 2025-06-25
fix(frontend): issue with Reward modal not being openable via Carousel

DenysKarmazynDFINITY:
# Motivation

It is currently not possible to open a Reward modal via Carousel. This started to be the case after some refactorings made for the airdrops feature a few months ago. To fix that, we need to do a few steps:

1. Move `RewardModal` to the global `Modals` component so it can be opened within any part of the app.
2. Extract the context initialisation into a separate component and re-use it in `Modals` and `AllRewardsList`.


AntonioVentilii (approved):
LGTM tks!

PR #7504, opened 2025-06-25
feat(frontend): Add Plausible event for Sprinkles banner on the login page

artkorotkikh-dfinity:
# Motivation

We'd like to track Sprinkles banner clicks 

# Changes

Added event 

# Tests
 
N/A

### DIFF
--- a/src/frontend/src/btc/services/btc-address.services.ts
+++ b/src/frontend/src/btc/services/btc-address.services.ts
@@ -11,6 +11,7 @@ import {
 	updateIdbBtcAddressMainnetLastUsage
 } from '$lib/api/idb-addresses.api';
 import { getBtcAddress as getSignerBtcAddress } from '$lib/api/signer.api';
+import { deriveBtcAddress } from '$lib/ic-pub-key/src/cli';
 import {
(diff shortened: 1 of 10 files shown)


src/frontend/src/lib/components/ui/InviteRewardsBanner.svelte
```diff
@@ -2,14 +2,26 @@
...
 	import Sprinkles from '$lib/components/sprinkles/Sprinkles.svelte';
+	import { TRACK_SPRINKLES_BANNER_CLICK } from '$lib/constants/analytics.contants';
+	import { trackEvent } from '$lib/services/analytics.services';
 	import { i18n } from '$lib/stores/i18n.store';
+
+	const handleSprinklesBannerClick = () => {
+		trackEvent({
+			name: TRACK_SPRINKLES_BANNER_CLICK,
+			metadata: {
+				source: 'login-page'
```
AntonioVentilii, line 13: Ideally, if we are going to use it more in the UI, we should use the `$page` store to get the source

AntonioVentilii (approved):
LGTM tks!

PR #7505, opened 2025-06-25
feat(frontend): Show all tokens if zero balance

daviddecentage:
# Motivation

We adjust the logic a bit when tokens in a group in the tokenlist should be truncated (show more button) or not. Until now we showed only native and ck token in a group if the groups total balance is 0. Now we want to show all tokens if total balance is 0.

# Changes

Adjusted condition to not check for ck and native token

# Tests

Manually tested by Stefan on FE1


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7513, opened 2025-06-25
feat(frontend): Enable more assets

daviddecentage:
# Motivation

We want to display the manage token list when we filter assets, so the user can quickly add another token that he is looking for but may have disabled.

- The list should reload only when filter changes
- The list should only show disabled tokens, with the exception of tokens that we enable via this list, so the user can revert it easily right after instead of having to go to the manage modal

# Changes

- Added the list
- Added filters, sorting etc for the list
- Make titles to StickyHeaders so in long lists we always see the "Apply" button and have a context
- Added translation for "Enable more assets"

# Tests

https://github.com/user-attachments/assets/5bba12dd-36aa-4af2-8373-79057984373c


src/frontend/src/lib/components/tokens/Tokens.svelte
```diff
@@ -5,22 +5,25 @@
...
 <div>
-	<div class="flex w-full flex-row justify-between">
-		<div class="grow-1 relative flex flex-row justify-between">
-			<TokensFilter>
-				{#snippet overflowableContent()}
-					<Header><span class="mt-2 flex">{$i18n.tokens.text.title}</span></Header>
-				{/snippet}
-			</TokensFilter>
+	<StickyHeader>
+		<div class="flex w-full flex-row justify-between">
```
DenysKarmazynDFINITY: I know it was there before but let's clean up a bit - `flex-row` is not needed

src/frontend/src/lib/components/tokens/Tokens.svelte
```diff
@@ -5,22 +5,25 @@
...
+			<div class="grow-1 relative flex flex-row justify-between">
```
DenysKarmazynDFINITY: also here

src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -46,6 +67,53 @@
 		getFilteredTokenList({ filter: $tokenListStore.filter, list: tokens ?? [] })
 	);
 
+	// To avoid strange behavior when the exchange data changes (for example, the tokens may shift
```
DenysKarmazynDFINITY: Not sure I got the idea here - we do want to have the tokens list dynamic, hence it reacting to changes in the rates sounds completely normal. Can you please explain why do you want to update this behavior?

src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -46,6 +67,53 @@
...
+				? pinEnabledTokensAtTop(
+						sortTokens({
+							$tokens: $allTokens,
+							$exchanges: exchangesStaticData,
+							$tokensToPin
+						})
+					)
+				: []
+			)
+				.filter(
```
DenysKarmazynDFINITY: let's replace `filter` + `map` a single `reduce` to avoid some additional looping

src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -46,6 +67,53 @@
...
+					(
+						t // hide enabled initially, but keep enabled ones that have just been enabled to let the user revert easily
```
DenysKarmazynDFINITY: maybe it's just my personal preference, but could you please use a full naming here, e.g. `token`/`listToken`?

src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -46,6 +67,53 @@
...
+					) =>
+						!t.enabled ||
+						(t.enabled && nonNullish(Object.values(modifiedTokens).find((s) => s.id === t.id)))
```
DenysKarmazynDFINITY: Also here - not really sure what `s` is

src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -54,6 +122,80 @@
...
+			toastsShow({
+				text: $i18n.tokens.manage.info.no_changes,
+				level: 'info',
+				duration: 5000
+			});
+
+			return;
+		}
+
+		await Promise.allSettled([
```
DenysKarmazynDFINITY: So this code (including all `save...` methods) is something that in the best case should be shared between this component and ManageTokensModal.

src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -54,6 +122,80 @@
...
+		};
+	};
+
+	const progress = () => ProgressStepsAddToken.DONE;
+
+	const saveIcrc = (tokens: SaveCustomTokenWithKey[]): Promise<void> =>
+		saveIcrcCustomTokens({
+			tokens,
+			progress,
+			modalNext: () => {},
```
DenysKarmazynDFINITY: Please update `save` services (separate PR if possible) to make all the modal-related stuff optional (`progress`, `modalNext`, etc) so we don't pass any empty/unecessary params

src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -93,6 +235,44 @@
...
 		{/if}
 
+		{#if $tokenListStore.filter !== '' && allTokensFilteredAndSorted.length > 0}
+			<div class="mb-3 mt-12 flex flex-col gap-3">
+				<StickyHeader>
+					<div class="flex items-center justify-between pb-4">
+						<h2 class="text-base">{$i18n.tokens.manage.text.enable_more_assets}</h2>
+						<div>
+							<Button
+								onclick={() => onSave()}
```
DenysKarmazynDFINITY:
```suggestion
								onclick={nSave}
```

DenysKarmazynDFINITY (commented):
A nice feature inside 💪

src/frontend/src/lib/components/tokens/Tokens.svelte
daviddecentage: removed

src/frontend/src/lib/components/tokens/Tokens.svelte
daviddecentage: removed

src/frontend/src/lib/components/tokens/TokensList.svelte
daviddecentage: According to Stefan, we decided we just sort the list alphabetically so I could remove all the balance/exchanges stuff there

src/frontend/src/lib/components/tokens/TokensList.svelte
daviddecentage: updated

src/frontend/src/lib/components/tokens/TokensList.svelte
daviddecentage: updated

src/frontend/src/lib/components/tokens/TokensList.svelte
daviddecentage: updated

src/frontend/src/lib/components/tokens/TokensList.svelte
daviddecentage: updated

src/frontend/src/lib/components/tokens/TokensList.svelte
daviddecentage: There we go: #7560

src/frontend/src/lib/components/tokens/TokensList.svelte
daviddecentage: There we go: #7560

src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -42,10 +50,64 @@
...
 
+	// Token list for enabling when filtering
+	let enableMoreTokensList: TokenUiOrGroupUi[] = $state([]);
+
+	const updateFilterList = (filter: string) => {
+		// hide enabled initially, but keep enabled (modified) ones that have just been enabled to let the user revert easily
+		// then we return it as a valid TokenUiOrGroupUi since the displaying cards require that type
+		const reducedTokens = ($allTokens ?? []).reduce<TokenUiOrGroupUi[]>((acc, token) => {
+			const isModified = nonNullish(
+				Object.values(modifiedTokens).find((modifiedToken) => modifiedToken.id === token.id)
```
DenysKarmazynDFINITY: `modifiedTokens` is Record<string, Token>, so why not just `const isModified = nonNullish(modifiedTokens[token.id])`?

src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -42,10 +50,64 @@
...
+					token: token as TokenUi
+				});
+			}
+			return acc;
+		}, []);
+
+		// sort alphabetally and apply filter
+		enableMoreTokensList = getFilteredTokenList({
+			filter,
+			list: reducedTokens.sort((a, b) =>
```
DenysKarmazynDFINITY, line 79: Would be really cool to move this function into a util and create a few unit tests just to understand what is an expected result. Let's do it in a follow-up PR?

src/frontend/src/lib/components/tokens/TokensList.svelte
DenysKarmazynDFINITY: Ah, sorry, just realized `string` is not a token id. Then, another suggestion - to prevent going through all modifiedTokens on every iteration of that reduce, can you please move it outside of the loop? You can generate a map to have Record<token.id, boolean>

DenysKarmazynDFINITY: P.P.S.: potentially, you still can use `modifiedTokens` directly if you generate keys here in the expected format [`${networkId.description}-${id.description}`]

daviddecentage: Im not sure the index is tokenId. I quickly tried it but it doesnt work so I assume we cannot use tokenId to index modifiedTokens.

Edit: just checked, its networkId.description + tokenId.description and adjusted it according to your suggestion

src/frontend/src/lib/components/tokens/TokensList.svelte
daviddecentage: Will do

PR #7516, opened 2025-06-25
feat(frontend): Token card togglable

daviddecentage:
# Motivation

In preparation to have a token list which is togglable when we have an active filter, we need to make the token cards togglable.

# Changes

- Added `ontoggle` prop
- Display it instead of the balance etc on the right if prop is passed

# Tests

With `ontoggle` passed:
image

Prop not passed:
image

Everything fine in privacy mode and visible skeleton as well:
image


src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -13,24 +13,39 @@
...
+	import { isSolanaToken } from '$sol/utils/token.utils';
+	import SolManageTokenToggle from '$sol/components/tokens/SolManageTokenToggle.svelte';
 
 	let {
 		data,
 		testIdPrefix = TOKEN_CARD,
 		asNetwork = false,
-		hover = false
+		hover = false,
+		ontoggle
```
AntonioVentilii:
```suggestion
		onToggle
```

src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -13,24 +13,39 @@
...
 	}: {
 		data: CardData;
 		testIdPrefix?: typeof TOKEN_CARD | typeof TOKEN_GROUP;
 		asNetwork?: boolean;
 		hover?: boolean;
+		ontoggle?: (t: CustomEvent<Token>) => void;
```
AntonioVentilii:
```suggestion
		onToggle?: (t: CustomEvent<Token>) => void;
```

src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -103,7 +120,17 @@
 
 		{#snippet descriptionEnd()}
 			<span class:text-sm={asNetwork} class="block min-w-12 text-nowrap">
-				{#if !$isPrivacyMode}
+				{#if nonNullish(ontoggle)}
```
AntonioVentilii:
```suggestion
				{#if nonNullish(onToggle)}
```

src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -103,7 +120,17 @@
...
+					{#if icTokenIcrcCustomToken(token)}
+						<IcManageTokenToggle {token} on:icToken={(t) => ontoggle(t)} />
+					{:else if isTokenEthereumUserToken(token) || isTokenSplToggleable(token)}
+						<ManageTokenToggle {token} on:icShowOrHideToken={(t) => ontoggle(t)} />
+					{:else if isBitcoinToken(token)}
+						<BtcManageTokenToggle />
+					{:else if isSolanaToken(token)}
+						<SolManageTokenToggle />
+					{/if}
```
AntonioVentilii: maybe this can already go in a different component (it's used twice, right?)? in another PR

src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -13,24 +13,39 @@
...
+		ontoggle?: (t: CustomEvent<Token>) => void;
 	} = $props();
 
 	const dispatch = createEventDispatcher();
 
 	let testId = $derived(
 		`${testIdPrefix}-${data.symbol}${nonNullish(data.network) ? `-${data.network.id.description}` : ''}`
 	);
+
+	let token: TokenToggleable<Token> = $derived(data as TokenToggleable<Token>);
```
AntonioVentilii: this is not always a token (it can be a group), so better to make it `TokenToggleable<Token> | undefined` using some new util (I don't recall an existing util), maybe even `zod` or a parser

daviddecentage: Maybe we can just use the same condition as for rendering? There needs to be a token if we wanna make it togglable, so maybe just `let token: TokenToggleable<Token> = $derived(nonNullish(onToggle) ? data as TokenToggleable<Token> : undefined);`

src/frontend/src/lib/components/tokens/TokenCard.svelte
daviddecentage: There we go https://github.com/dfinity/oisy-wallet/pull/7517

src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -1,36 +1,54 @@
...
 	} = $props();
 
 	const dispatch = createEventDispatcher();
 
 	let testId = $derived(
 		`${testIdPrefix}-${data.symbol}${nonNullish(data.network) ? `-${data.network.id.description}` : ''}`
 	);
+
+	let token: TokenToggleable<Token> | undefined = $derived(
+		nonNullish(onToggle) ? (data as TokenToggleable<Token>) : undefined
```
AntonioVentilii: what I meant, is to parse `data` as toggleable `token`, and if it fails, return `undefined` (not necessarily linked to `onToggle`)

daviddecentage: updated

src/frontend/src/lib/utils/token-card.utils.ts
```diff
@@ -23,3 +24,8 @@ export const mapHeaderData = ({
 	usdBalance,
 	tokenCount: tokens.length
 });
+
+export const isCardDataTogglableToken = (data: CardData): boolean => {
```
AntonioVentilii: nitpick: if you do

```suggestion
export const isCardDataTogglableToken = (data: CardData): data is TokenToggleable<Token> => {
```

you don't need to cast afterwards:

```ts
let token: TokenToggleable<Token> | undefined = $derived(
		isCardDataTogglableToken(data) ? data : undefined
	);
```

src/frontend/src/lib/components/tokens/TokenCard.svelte
```diff
@@ -75,16 +85,18 @@
...
-				<TokenBalance {data} hideBalance={$isPrivacyMode}>
-					{#snippet privacyBalance()}
-						<IconDots
-							variant={asNetwork ? 'sm' : 'md'}
-							styleClass={asNetwork ? 'my-4.25' : 'pb-4'}
-						/>
-					{/snippet}
-				</TokenBalance>
-			</span>
+			{#if isNullish(ontoggle)}
```
AntonioVentilii:
```suggestion
			{#if isNullish(onToggle)}
```

AntonioVentilii (approved):
LGTM tks!

PR #7518, opened 2025-06-25
feat(frontend): double click activates privacy mode

CvetelinaAngelova:
# Motivation

This PR enhances the user experience by:
Allowing users to toggle privacy mode (show/hide balance) via double-click on the balance text.
Providing an accessible tooltip that explains this interaction, ensuring discoverability and usability.

# Changes

Added ondblclick to the balance display to toggle privacy mode via setPrivacyMode.
Wrapped the balance and "hidden" text with the Tooltip component from @dfinity/gix-components.
Tooltip shows the localized string: "Double-click to hide/show balance".
Ensured accessibility.
Styled the tooltip using theme variables.

# Tests

Manual testing

Verified tooltip text displays correctly across languages.
Confirmed double-click toggles balance visibility (with toast).
Ensured keyboard focus works on balance element.
Tested tooltip positioning and arrow rendering in both hidden/visible balance states.
Checked for no accessibility or style regressions.
Screenshot 2025-06-25 at 19 54 41


src/frontend/src/lib/components/hero/Balance.svelte
```diff
@@ -46,18 +48,28 @@
...
-			/>
-		{:else}
-			<span class="flex items-center justify-center gap-2">
-				<IconEyeOff />
-				{$i18n.hero.text.hidden_balance}
-			</span>
-		{/if}
-	</span>
+	<div class="flex flex-col items-center">
+		<button
```
AntonioVentilii: is it possible to use the existing `<Button>` component and expand it? Or create a new component with tailwind classes instead of globals? We need to start migrating it anyway

And add an `aria-label` pls

CvetelinaAngelova: Change it to the transperant button and added support for the double click on the Button component

AntonioVentilii (approved):
LGTM tks!

PR #7520, opened 2025-06-26
feat(frontend): Get always the max fee for Ethereum/EVM

AntonioVentilii:
# Motivation

For all tokens in Ethereum and EVM, it is better to always get the max fee as the maximum between our sources: ck-minters, RPC fee data, and Infura fee API.

# Changes

- Create util to get the max between two `bigint` objects.
- Use the util in `FeeContext`.

# Tests

Added tests.

### DIFF
--- a/src/frontend/src/lib/utils/bigint.utils.ts
+++ b/src/frontend/src/lib/utils/bigint.utils.ts
@@ -1,0 +1,5 @@
+import { nonNullish } from '@dfinity/utils';
+
+// eslint-disable-next-line local-rules/prefer-object-params -- Visually, it is clearer to have two parameters for maxBigInt
+export const maxBigInt = (n1: bigint | null, n2: bigint | null): bigint | null =>
+	nonNullish(n1) && nonNullish(n2) ? (n1 > n2 ? n1 : n2) : (n1 ?? n2);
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/utils/bigint.utils.ts
```diff
@@ -0,0 +1,6 @@
+import type { Option } from '$lib/types/utils';
+import { nonNullish } from '@dfinity/utils';
+
+// eslint-disable-next-line local-rules/prefer-object-params -- Visually, it is clearer to have two parameters for maxBigInt
+export const maxBigInt = (n1: Option<bigint>, n2: Option<bigint>): Option<bigint> =>
+	nonNullish(n1) && nonNullish(n2) ? (n1 > n2 ? n1 : n2) : (n1 ?? n2);
```
DenysKarmazynDFINITY, line 6: potential `(n1 ?? ZERO) > (n2 ?? ZERO) ? n1 : n2` could also work?

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/lib/utils/bigint.utils.ts
AntonioVentilii: i wanted to include the negatives in this comparison (very wide approach), so theoretically i cannot default to `ZERO`, otherwise `null` vs `-10n` would make `null` the biggest (added tests for it)

DenysKarmazynDFINITY: makes sense 👍

PR #7523, opened 2025-06-26
fix(frontend): added new condition for network id before making a call to infura

roman-nazaruk:
# Motivation

When we open the EthTransactionModal, there is an error because the block is not loaded. This happens when token.network.id is undefined. To fix this, we added a check to make sure token.network.id is defined before loading the block

# Changes

Added a condition to check if token.network.id is not undefined before showing the block and transaction status in the list.

### DIFF
--- a/src/frontend/src/icp/services/icrc.services.ts
+++ b/src/frontend/src/icp/services/icrc.services.ts
@@ -53,14 +53,16 @@ const loadDefaultIcrcTokens = async () => {
 
 export const loadCustomTokens = ({
 	identity,
-	useCache = false
+	useCache = false,
+	onSuccess
 }: {
 	identity: OptionIdentity;
 	useCache?: boolean;
+	onSuccess?: () => void;
 }): Promise<void> =>
 	queryAndUpdate<IcrcCustomToken[]>({
 		request: (params) => loadIcrcCustomTokens({ ...params, useCache }),
-		onLoad: loadIcrcCustomData,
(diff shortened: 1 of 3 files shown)


src/frontend/src/eth/components/transactions/EthTransactionModal.svelte
```diff
@@ -147,7 +147,7 @@
 				</ListItem>
 			{/if}
 
-			{#if nonNullish(blockNumber) && nonNullish(token)}
+			{#if nonNullish(blockNumber) && nonNullish(token) && nonNullish(token.network.id)}
```
DenysKarmazynDFINITY: Interesting - `network` is a required param of `token`. Did you manage to reproduce a case where it was nullish?

StefanBerger-DFINITY: It's this error, right?

https://github.com/user-attachments/assets/ca633611-1b33-4208-9c29-8e339d7942e1

roman-nazaruk: Yeah, this one

roman-nazaruk: > Did you manage to reproduce a case where it was nullish?

Yes, with this new condition, I wasn’t able to reproduce the issue, but without it, I can

AntonioVentilii (approved):

PR #7526, opened 2025-06-26
feat(frontend): relaunch ICRC worker on indexCanisterId being added

DenysKarmazynDFINITY:
# Motivation

To make a newly added indexCanisterId being used by the ICRC wallet worker, we need to reset the related token store before updating it with the new data from BE (which will include indexCanisterId).


roman-nazaruk (approved):
LGTM, thnx

PR #7528, opened 2025-06-26
chore(frontend): Upgrade tokenlist svelte 5

daviddecentage:
# Motivation

Since we need to make some adjustments to the token list we upgrade it to svelte 5.

# Changes

Upgraded component


roman-nazaruk (approved):
LGTM, thnx

PR #7533, opened 2025-06-27
fix(frontend): Allow ERC20 tokens with same address and different networks

AntonioVentilii:
# Motivation

It may happen that some ERC20 share the same address thorough different networks. However, OISY was consider them duplicates. We fix that.

# Changes

- Change method setAll of user token store to identify ERC20 tokens both by address and chain ID.

# Tests

Added tests and practical example:

### Before

Screenshot 2025-06-27 at 11 35 03

### After

Screenshot 2025-06-27 at 11 34 46


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7534, opened 2025-06-27
feat(frontend): allow editing tokens with existing indexCanisterId

DenysKarmazynDFINITY:
# Motivation

We want to allow editing custom ICRC tokens that already have indexCanisterId set.

Screenshot 2025-06-27 at 11 38 28


AntonioVentilii (approved):
LGTM tks!

~~I would just add a test that that throws when giving an empty string to the index canister.~~

As discussed offline, such a case should not be possible

DenysKarmazynDFINITY:
As discussed "offline", the "save" button is disabled when user tries to set an empty value as indexCanisterId.

PR #7535, opened 2025-06-27
feat(frontend): Sticky header

daviddecentage:
# Motivation

On some places it might be neat to have some kind of sticky header which pins to the top of the page when scrolled past. This is useful when we want to keep some context info about a list etc.

# Changes

Added sticky header component

# Tests

https://github.com/user-attachments/assets/724ca2d8-c06a-49e5-81ad-eb73239d87f1


src/frontend/src/lib/components/ui/StickyHeader.svelte
```diff
@@ -0,0 +1,62 @@
...
+	};
+
+	$effect(() => {
+		calcSizes();
+	});
+</script>
+
+<svelte:window on:scroll={handleScroll} on:resize={() => calcSizes(true)} />
+
+<div bind:this={rootElement} class="relative block" style={`height: ${originalHeight ?? 0}px`}>
```
DenysKarmazynDFINITY: don't think `block` is needed

src/frontend/src/lib/components/ui/StickyHeader.svelte
```diff
@@ -0,0 +1,62 @@
...
+	<div
+		class="z-2 block"
```
DenysKarmazynDFINITY: here, too

src/frontend/src/lib/components/ui/StickyHeader.svelte
```diff
@@ -0,0 +1,62 @@
...
+	let scrolledSoon = $state(false);
+	let scrolledPast = $state(false);
+
+	const handleScroll = () => {
+		if (!rootElement) {
+			return;
+		}
+
+		const rect = rootElement.getBoundingClientRect();
+		scrolledSoon = rect.top <= 24 * 4;
```
DenysKarmazynDFINITY: Can we please introduce some consts for these vars? Just to make it a bit more readable + the "24" value can be reused below

src/frontend/src/lib/components/ui/StickyHeader.svelte
```diff
@@ -0,0 +1,62 @@
...
+			originalHeight = rootElement.children?.[0].clientHeight;
+		}
+	};
+
+	$effect(() => {
+		calcSizes();
+	});
+</script>
+
+<svelte:window on:scroll={handleScroll} on:resize={() => calcSizes(true)} />
```
DenysKarmazynDFINITY: let's add `debounce` to both `on:scroll` and `on:resize`

DenysKarmazynDFINITY (commented):
Just for a reference: we discussed "offline" why `position: sticky` does not work for us in this case - hence, a JS solution is needed.

src/frontend/src/lib/components/ui/StickyHeader.svelte
daviddecentage: I think debounce on the resize is a good idea, adding it. But for the scroll to remai smooth i tried different debounce delays but at the point where a debounce makes sense it turns very jittery and doesnt feel smooth.

src/frontend/src/lib/components/ui/StickyHeader.svelte
daviddecentage: done

src/frontend/src/lib/components/ui/StickyHeader.svelte
daviddecentage: updated

src/frontend/src/lib/components/ui/StickyHeader.svelte
daviddecentage: updated

src/frontend/src/lib/styles/global/body.scss
```diff
@@ -1,6 +1,6 @@
 body {
 	width: 100vw;
-	height: 100svh;
+	height: inherit; // needed for winow on:scroll to work
```
DenysKarmazynDFINITY: There is a type: `winow` -> `window`.

Also, please make sure it does not break anything by checking different browsers/breakpoints. Thanks!

daviddecentage: Tested on iOS safari and chrome (works really smooth to my suprise). Also tested on Chrome and Firefox on desktop and also in responsive mode. Seems to work just fine

PR #7536, opened 2025-06-27
feat(frontend): Create pseudo-network for ICRC test tokens

AntonioVentilii:
# Motivation

There are a few ICRC tokens that we define as "testnet" tokens, for example ckSepoliaETH or ckSepoliaUSDC. They are linked to their respective twin tokens but on testnets.

Since there is no test network for IC network, we first decided to show them in the IC network AND only if testnets are enabled. However, that is a bit confusing since we show a wrong balance (the sum of "normal" tokens and "testnet" tokens).

Since these tokens are really just for testing and developing purposes, the new proposal is to separate them from IC network and put them in a pseudo-test-network, that is simply a replica of IC network.

# Changes

- Create pseudo-network for ICRC "testnet" tokens: it is a copy of the existing IC network.
- Add it to the respective lists of networks.
- Expand util isNetworkIdICP (the pseudo-network should be treated exactly as the IC network).
- Map ICRC "testnet" tokens with the new pseudo-network.

# Tests

In local replica and test canister everything works as usual:

Screenshot 2025-06-27 at 12 26 32
Screenshot 2025-06-27 at 12 26 39
Screenshot 2025-06-27 at 12 26 47
Screenshot 2025-06-27 at 12 26 54
Screenshot 2025-06-27 at 12 33 55
Screenshot 2025-06-27 at 12 34 24
Screenshot 2025-06-27 at 13 30 39

### DIFF
--- a/src/frontend/src/icp/utils/icrc.utils.ts
+++ b/src/frontend/src/icp/utils/icrc.utils.ts
@@ -34,6 +34,32 @@ const CUSTOM_SYMBOLS_BY_LEDGER_CANISTER_ID: Record<LedgerCanisterIdText, string>
 	[ICONFUCIUS_LEDGER_CANISTER_ID]: 'ICONFUCIUS'
 };
 
+/**
+ * Determines which network a given ICRC token belongs to based on its ledger canister ID.
+ *
+ * Some tokens (e.g., `ckSepoliaETH`, `ckSepoliaUSDC`) are considered "testnet" tokens.
+ * These are tied to testnets of external chains (like Sepolia for Ethereum), but since the
+ * Internet Computer (IC) does not have a native testnet environment, we fake one.
+ *
+ * Originally, testnet tokens were shown as part of the IC network, but that led to confusion
+ * — especially when aggregating balances — as their balances would mix with mainnet tokens.
+ *
+ * To avoid this, we now separate these testnet tokens into a pseudo-test network:
+ * `ICP_PSEUDO_TESTNET_NETWORK`, which mirrors the IC network in behavior but is visually
+ * and logically distinct.
+ *
+ * This ensures:
+ * - Production tokens only show in the "real" IC network (`ICP_NETWORK`)
+ * - Testnet tokens only show in the pseudo-network when testnet mode is enabled
+ *
+ * @param ledgerCanisterId - The ledger canister ID of the token.
+ * @returns The appropriate network identifier for the token:
+ *          - `ICP_NETWORK` for "mainnet" tokens
+ *          - `ICP_PSEUDO_TESTNET_NETWORK` for known "testnet" tokens
+ */
+const mapIcNetwork = (ledgerCanisterId: LedgerCanisterIdText) =>
+	isTokenIcrcTestnet({ ledgerCanisterId }) ? ICP_PSEUDO_TESTNET_NETWORK : ICP_NETWORK;
+
 export const mapIcrcToken = ({
 	metadata,
 	icrcCustomTokens,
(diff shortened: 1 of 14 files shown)


src/frontend/src/lib/utils/user-networks.utils.ts
```diff
@@ -35,6 +35,8 @@ const networkIdToKey = (networkId: NetworkId): NetworkSettingsFor | undefined =>
 	switch (networkId) {
 		case ICP_NETWORK_ID:
 			return { InternetComputer: null };
```
peterpeterparker: You can remove this line

src/frontend/src/lib/derived/user-networks.derived.ts
```diff
@@ -126,7 +126,8 @@ export const userNetworks: Readable<UserNetworks> = derived(
 				return { ...acc, [networkId]: { enabled, isTestnet } };
 			}, {}),
 			// We always enable ICP network.
-			[ICP_NETWORK_ID]: { enabled: true, isTestnet: false }
+			[ICP_NETWORK_ID]: { enabled: true, isTestnet: false },
+			[ICP_PSEUDO_TESTNET_NETWORK_ID]: { enabled: true, isTestnet: true }
```
peterpeterparker, line 130: Not disabled by default?

src/frontend/src/icp/utils/icrc.utils.ts
```diff
@@ -58,7 +59,7 @@ export const mapIcrcToken = ({
 
 	return {
 		id: parseTokenId(symbol),
-		network: ICP_NETWORK,
+		network: isTokenIcrcTestnet({ ledgerCanisterId }) ? ICP_PSEUDO_TESTNET_NETWORK : ICP_NETWORK,
```
peterpeterparker: feel like this would be more readable with a dedicated function

```
network: mapICNetwork // + JSDOC
```

Or something

peterpeterparker (commented):
I kind of feel like it's a somewhat misleading use of the `network` field, which could potentially lead to issues  - even though we've migrated to rely solely on the `network` information provided by the `token`.

That said, if it's well-documented (i.e., it's clearly stated that `network` is not actually a network, maybe add some JSDocs or docs somewhere?), thoroughly tested, or even maybe renamed (?), then I suppose it's acceptable for `network` to contain values that don't correspond to real networks.

src/frontend/src/lib/derived/user-networks.derived.ts
AntonioVentilii: I think it is better to enable it by default (if the testnets are enabled), since ICP is always enabled with no chance of disabling, and since the tokens are technically the same network

src/frontend/src/lib/utils/user-networks.utils.ts
AntonioVentilii: right! good catch!

src/frontend/src/icp/utils/icrc.utils.ts
AntonioVentilii: yes, better, done with a little help from chatGPT eheh! have a look pls!

src/frontend/src/lib/derived/user-networks.derived.ts
peterpeterparker: Makes sense

DenysKarmazynDFINITY (approved):
Agree with David re. the `network` field becoming less "clear" with this change, but all in all the solution LGTM 👍

peterpeterparker (approved):
LGTM, thx

PR #7537, opened 2025-06-27
fix(frontend): issue with parsing a principal in Address picker

DenysKarmazynDFINITY:
# Motivation

This PR fixes an issue with parsing a principal in Address picker.

### DIFF
--- a/src/frontend/src/icp/derived/ic-contacts.derived.ts
+++ b/src/frontend/src/icp/derived/ic-contacts.derived.ts
@@ -2,8 +2,10 @@ import { ICP_TOKEN_ID } from '$env/tokens/tokens.icp.env';
 import { getAccountIdentifier } from '$icp/utils/icp-account.utils';
 import { isIcrcAddress } from '$icp/utils/icrc-account.utils';
 import { isTokenIcrc } from '$icp/utils/icrc.utils';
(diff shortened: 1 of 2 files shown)


src/frontend/src/icp/derived/ic-contacts.derived.ts
```diff
@@ -23,13 +25,20 @@ export const icNetworkContacts: Readable<NetworkContacts> = derived(
...
+						...acc,
+						[contact.id.toString()]: [
+							...(acc[contact.id.toString()] ?? []),
+							getAccountIdentifier(Principal.fromText(address)).toHex()
+						]
+					};
+				} catch (_: unknown) {
+					trackEvent({ name: TRACK_IC_NETWORK_CONTACTS_GENERATION_ISSUE });
+
+					// if address parsing failed, we ignore it and continue further
```
AntonioVentilii: shall we add something about that we arrive here if there is a sub-account?

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/icp/derived/ic-contacts.derived.ts
DenysKarmazynDFINITY: updated comment to be more specific

PR #7543, opened 2025-06-27
feat(frontend): allow resetting indexCanisterId for custom ICRC tokens

DenysKarmazynDFINITY:
# Motivation

The last missing piece - we now allow resetting indexCanisterId for custom ICRC tokens. Also, I adjusted the `onSuccess` callback so we only call it once after the query request succeeds.


AntonioVentilii (approved):
LGTM tks!

PR #7547, opened 2025-06-30
feat(frontend): added plausible event dapp filter button

roman-nazaruk:
# Motivation

Add plausible events tracking for filter button clicks on the dApps page.

# Changes

Added `trackEvent` inside `onFilterBtnClick` to log filtering button tags.


AntonioVentilii (approved):
LGTM tks!

Any chance of adding some tests?

roman-nazaruk:
> LGTM tks!
> 
> Any chance of adding some tests?

TBH, to test that properly we would need to create an external testing component or use the anti-pattern of exporting method from the component (which I don’t think is a good approach). So I’d prefer to leave it as is.

PR #7554, opened 2025-06-30
feat(frontend): handle subaccounts in icNetworkContacts

DenysKarmazynDFINITY:
# Motivation

Instead of just skipping principal addresses with subaccounts, we need to correctly handle such a case. To do that, we move the address parsing logic to a separate util `parseIcrcAccountToAccountIdentifierText`. Also, the temporary plausible event can now be removed.

### DIFF
--- a/src/frontend/src/icp/derived/ic-contacts.derived.ts
+++ b/src/frontend/src/icp/derived/ic-contacts.derived.ts
@@ -1,5 +1,5 @@
 import { ICP_TOKEN_ID } from '$env/tokens/tokens.icp.env';
-import { parseIcrcAccountToAccountIdentifierText } from '$icp/utils/icp-account.utils';
+import { tryToParseIcrcAccountStringToAccountIdentifierText } from '$icp/utils/icp-account.utils';
 import { isIcrcAddress } from '$icp/utils/icrc-account.utils';
 import { isTokenIcrc } from '$icp/utils/icrc.utils';
 import { contacts } from '$lib/derived/contacts.derived';
@@ -23,7 +23,7 @@ export const icNetworkContacts: Readable<NetworkContacts> = derived(
 			const { address, contact } = allIcNetworkContacts[key];
 
 			if (isIcrcAddress(address)) {
-				const accountIdentifierText = parseIcrcAccountToAccountIdentifierText(address);
+				const accountIdentifierText = tryToParseIcrcAccountStringToAccountIdentifierText(address);
 
 				return {
 					...acc,
--- a/src/frontend/src/icp/derived/ic-contacts.derived.ts
+++ b/src/frontend/src/icp/derived/ic-contacts.derived.ts
@@ -1,14 +1,12 @@
 import { ICP_TOKEN_ID } from '$env/tokens/tokens.icp.env';
-import { getAccountIdentifier } from '$icp/utils/icp-account.utils';
+import { parseIcrcAccountToAccountIdentifierText } from '$icp/utils/icp-account.utils';
 import { isIcrcAddress } from '$icp/utils/icrc-account.utils';
 import { isTokenIcrc } from '$icp/utils/icrc.utils';
-import { TRACK_IC_NETWORK_CONTACTS_GENERATION_ISSUE } from '$lib/constants/analytics.contants';
 import { contacts } from '$lib/derived/contacts.derived';
 import { tokenWithFallback } from '$lib/derived/token.derived';
-import { trackEvent } from '$lib/services/analytics.services';
 import type { NetworkContacts } from '$lib/types/contacts';
 import { getNetworkContacts } from '$lib/utils/contacts.utils';
-import { Principal } from '@dfinity/principal';
+import { nonNullish } from '@dfinity/utils';
 import { derived, type Readable } from 'svelte/store';
 
 export const icNetworkContacts: Readable<NetworkContacts> = derived(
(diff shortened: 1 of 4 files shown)


src/frontend/src/icp/utils/icp-account.utils.ts
```diff
@@ -54,3 +55,25 @@ export const getIcrcv2AccountIdString = (accountId: Icrcv2AccountId): string =>
 
 	return assertNever({ variable: accountId, typeName: 'Icrcv2AccountId' });
 };
+
+export const parseIcrcAccountToAccountIdentifierText = (address: Address): string | undefined => {
```
peterpeterparker: - maybe `tryToParse...` since we ignore errors?

- can you add JSDocs for providing context? Since I asked few question on Slack, might be useful maybe?

src/frontend/src/icp/utils/icp-account.utils.ts
```diff
@@ -54,3 +55,25 @@ export const getIcrcv2AccountIdString = (accountId: Icrcv2AccountId): string =>
...
+	try {
+		const { owner: principal, subaccount: unparsedSubAccount } = decodeIcrcAccount(address);
```
peterpeterparker: `unparsedSubAccount` -> `icrcSubaccount` maybe?

src/frontend/src/icp/utils/icp-account.utils.ts
```diff
@@ -54,3 +55,25 @@ export const getIcrcv2AccountIdString = (accountId: Icrcv2AccountId): string =>
...
+
+		const subAccount = nonNullish(unparsedSubAccount)
+			? SubAccount.fromBytes(new Uint8Array(unparsedSubAccount))
+			: undefined;
+
+		if (!(subAccount instanceof Error)) {
```
peterpeterparker, line 76: You might want to upgrade ic-js next first because in the latest version `fromBytes` throw an error.

src/frontend/src/icp/utils/icp-account.utils.ts
```diff
@@ -54,3 +55,25 @@ export const getIcrcv2AccountIdString = (accountId: Icrcv2AccountId): string =>
...
+			return AccountIdentifier.fromPrincipal({
+				principal,
+				...(nonNullish(subAccount) && {
```
peterpeterparker: In case interesting, here you can simplify the expression:

```
AccountIdentifier.fromPrincipal({principal, subAccount}).toHex()
```

src/frontend/src/icp/derived/ic-contacts.derived.ts
```diff
@@ -25,21 +23,14 @@ export const icNetworkContacts: Readable<NetworkContacts> = derived(
...
-				try {
-					return {
-						...acc,
-						[contact.id.toString()]: [
-							...(acc[contact.id.toString()] ?? []),
-							getAccountIdentifier(Principal.fromText(address)).toHex()
-						]
-					};
-				} catch (_: unknown) {
-					trackEvent({ name: TRACK_IC_NETWORK_CONTACTS_GENERATION_ISSUE });
```
peterpeterparker, line 37: The analytic entry got removed in the PR, is it expected?

DenysKarmazynDFINITY: Yes, we introduced it temporarily on Friday with the idea to remove it after the parsing correctly handles the subaccount case.

src/frontend/src/icp/utils/icp-account.utils.ts
DenysKarmazynDFINITY: renamed and added!

src/frontend/src/icp/utils/icp-account.utils.ts
DenysKarmazynDFINITY: Agreed

src/frontend/src/icp/utils/icp-account.utils.ts
DenysKarmazynDFINITY: true, updated

src/frontend/src/icp/utils/icp-account.utils.ts
DenysKarmazynDFINITY: [The ic-js upgrade didn't work](https://github.com/dfinity/oisy-wallet/actions/runs/14593319705/job/40933323682) - [the signer](https://github.com/dfinity/oisy-wallet-signer/blob/main/package.json#L97) should be updated first. 

I saw that the signer upgrade is already in progress, so I added a todo comment to update the util after oisy is also using the latest ic-js. Does this sound good to you?

src/frontend/src/icp/utils/icp-account.utils.ts
```diff
@@ -54,3 +54,31 @@ export const getIcrcv2AccountIdString = (accountId: Icrcv2AccountId): string =>
 
 	return assertNever({ variable: accountId, typeName: 'Icrcv2AccountId' });
 };
+
+/**
+ * Tries to parse an Icrc-1 account string to AccountIdentifier text.
```
peterpeterparker, line 59: Can you maybe add why because this is not common - i.e. I was suprised when you told me OISY does so.

src/frontend/src/icp/utils/icp-account.utils.ts
peterpeterparker: Please review following PRs:

- https://github.com/dfinity/oisy-wallet-signer/pull/592

- https://github.com/dfinity/oisy-wallet-signer/pull/593

That said, it still won't work. We will have to migrate to Zod v4, I can have a look tomorrow if not too complicated.

peterpeterparker: Let's move on with this PR and adapt afterwards then. Nevertheless if you can review those two PRs would be great.

peterpeterparker (approved):
LGTM, thanks

Note that I'm still not why the feature is required but, code wise, sounds good.

Before merge, please review the JSDocs. I think it would be great if we provide some context on why this is required.

src/frontend/src/icp/utils/icp-account.utils.ts
DenysKarmazynDFINITY: Good point - added.

PR #7559, opened 2025-06-30
feat(frontend): track successful token edit event

DenysKarmazynDFINITY:
# Motivation

We want to track when the "edit token" action is successfully completed.


src/frontend/src/lib/constants/analytics.contants.ts
```diff
@@ -93,6 +93,7 @@ export const TRACK_WELCOME_OPEN = 'welcome_open';
 
 // Delete Token
```
AntonioVentilii:
```suggestion
// Edit/Delete Token
```

AntonioVentilii (approved):
LGTM tks!

PR #7560, opened 2025-06-30
refactor(frontend): Save all custom tokens

daviddecentage:
# Motivation

We refactor the saving of all custom tokens a bit to make it reusable for another use case where we need to enable/disable tokens.

# Changes

- Adjusted `ManageTokens` to use the new function
- Added `saveAllCustomTokens` to utils
- Made all callbacks like `progress`, `modalNext`, `onSuccess` etc optional

# Tests

Added tests for new utils function and tested manually to ensure everything still works as expected

### DIFF
--- a/src/frontend/src/lib/services/manage-tokens.services.ts
+++ b/src/frontend/src/lib/services/manage-tokens.services.ts
@@ -21,15 +21,15 @@ import { isNullish, nonNullish } from '@dfinity/utils';
 import { get } from 'svelte/store';
 
 export interface ManageTokensSaveParams {
-	progress: (step: ProgressStepsAddToken) => void;
-	modalNext: () => void;
-	onSuccess: () => void;
-	onError: () => void;
+	progress?: (step: ProgressStepsAddToken) => void;
+	modalNext?: () => void;
+	onSuccess?: () => void;
+	onError?: () => void;
 	identity: OptionIdentity;
 }
 
 export interface SaveTokensParams<T> {
-	progress: (step: ProgressStepsAddToken) => void;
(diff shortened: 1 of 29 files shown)


src/frontend/src/lib/services/manage-tokens.services.ts
```diff
@@ -62,7 +62,9 @@ export const saveTokens = async <
 		return;
 	}
 
-	modalNext();
+	if (nonNullish(modalNext)) {
+		modalNext();
```
DenysKarmazynDFINITY: let's use here (and in all similar cases in this file) the shorter version (`modalNext?.();`), same as you did it in `spl-custom-tokens.service`

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -272,3 +278,72 @@ export const groupTogglableTokens = (
...
+
+		return;
+	}
+
+	await Promise.allSettled([
+		...(icrc.length > 0
+			? [
+					saveIcrcCustomTokens({
+						tokens: icrc.map((t) => ({ ...t, networkKey: 'Icrc' })),
+						progress,
```
DenysKarmazynDFINITY: let's create a const `params`, store all shared params there and reuse them in all 3 cases

src/frontend/src/lib/services/manage-tokens.services.ts
daviddecentage: updated

src/frontend/src/lib/utils/tokens.utils.ts
daviddecentage: updated

PR #7561, opened 2025-06-30
feat(build): Associate each env to its backend but only if exists

AntonioVentilii:
# Motivation

It is useful to associate each env to its own backend canister if any happens to exist.

### DIFF
--- a/src/frontend/src/lib/constants/app.constants.ts
+++ b/src/frontend/src/lib/constants/app.constants.ts
@@ -7,7 +7,9 @@ export const APP_VERSION = VITE_APP_VERSION;
 export const MODE = VITE_DFX_NETWORK;
 export const LOCAL = MODE === 'local';
 export const TEST_FE = MODE.startsWith('test_fe_');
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/constants/app.constants.ts
```diff
@@ -7,7 +7,9 @@ export const APP_VERSION = VITE_APP_VERSION;
 export const MODE = VITE_DFX_NETWORK;
 export const LOCAL = MODE === 'local';
 export const TEST_FE = MODE.startsWith('test_fe_');
-export const STAGING = MODE === 'staging' || TEST_FE || MODE === 'audit' || MODE === 'e2e';
+export const AUDIT = MODE === 'audit';
+export const E2E = MODE === 'e2e';
+export const STAGING = MODE === 'staging' || TEST_FE || AUDIT || E2E;
```
DenysKarmazynDFINITY, line 12: Since we now have separate `AUDIT` and `E2E` checks, shall we remove them from here?

DenysKarmazynDFINITY: Nevermind - I saw this const is used in other places too, so we need to keep all values.

PR #7564, opened 2025-06-30
feat(frontend): Sort tokenOrGroupUi alphabetically

daviddecentage:
# Motivation

According to Denys suggestion in [7513](https://github.com/dfinity/oisy-wallet/pull/7513), we move the sorting out to its own util with tests.

# Changes

- Added utils function
- Adjusted occurrence as well

# Tests

Added tests

### DIFF
--- a/src/frontend/src/btc/services/worker.btc-wallet.services.ts
+++ b/src/frontend/src/btc/services/worker.btc-wallet.services.ts
@@ -8,7 +8,7 @@ import {
 } from '$lib/stores/address.store';
 import type { OptionCanisterIdText } from '$lib/types/canister';
 import type { WalletWorker } from '$lib/types/listener';
-import type { PostMessage, PostMessageDataResponseError } from '$lib/types/post-message';
+import type { PostMessage } from '$lib/types/post-message';
 import type { Token } from '$lib/types/token';
 import {
 	isNetworkIdBTCMainnet,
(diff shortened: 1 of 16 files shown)


src/frontend/src/lib/utils/token-group.utils.ts
```diff
@@ -186,3 +186,25 @@ export const groupTokens = (tokens: TokenUi[]): TokenUiOrGroupUi[] => {
 
 	return Object.getOwnPropertySymbols(tokenGroupsMap).map((id) => tokenGroupsMap[id as TokenId]);
 };
+
+export const sortTokenOrGroupUiAlphabetically = (tokenOrGroup: TokenUiOrGroupUi[]) =>
```
AntonioVentilii: why not just `sortTokenOrGroupUi`?

AntonioVentilii (commented):
We are already sorting tokens, before grouping: should that be removed?

And in that logic does something different:
- we consider deprecated tokens;
- we sort by market cap if same name;
- we sort by network name if same name.

Is that logic deprecated? Does this new util re-sort them? Or is it used in a completely different context?

src/frontend/src/lib/utils/token-group.utils.ts
```diff
@@ -186,3 +186,25 @@ export const groupTokens = (tokens: TokenUi[]): TokenUiOrGroupUi[] => {
...
+						const bStartsValid = isAlphaNum(bName.charAt(0));
+						if (aStartsValid && !bStartsValid) {
+							return -1;
+						}
+						if (!aStartsValid && bStartsValid) {
+							return 1;
+						}
+						return aName.localeCompare(bName);
+					})()
+				: 1 // currently we have no plans to sort groups alphabetically so we fallback to not sorting it
```
AntonioVentilii: maybe add a todo?

src/frontend/src/lib/utils/token-group.utils.ts
daviddecentage: updated

src/frontend/src/lib/utils/token-group.utils.ts
daviddecentage: updated

daviddecentage:
> We are already sorting tokens, before grouping: should that be removed?
> 
> And in that logic does something different:
> 
> * we consider deprecated tokens;
> * we sort by market cap if same name;
> * we sort by network name if same name;
> * we pin specific tokens on top (native ones, ck-tokens).
> 
> Is that logic deprecated? Does this new util re-sort them? Or is it used in a completely different context?

Its an entirely different case. We only show enable/disable switch so sorting by anything that needs balance/exchange rates is not needed for this specific case. The only requirement is to sort alphabetically and everything that doesnt start with alphanumeric character gets put last.

Thats why I initially had alphabetically in the function name to make it clear its a different sort altogether. Maybe I should bring that back?

All other sorting is still used and should not be removed.

AntonioVentilii (approved):
Tks for the explanation!

PR #7565, opened 2025-06-30
fix(frontend): Save manage tokens modal

daviddecentage:
# Motivation

On splitting and moving logic out for saving tokens a bug snuck in which breaks the saving in the manage tokens modal. This was due to the change in how the event data is propagated between the `ManageTokens` and `ManageTokensModal` components.

# Changes

Fixed the param which was wrongly read

# Tests

Manually tested, works like before


DenysKarmazynDFINITY (approved):
Thanks!

PR #7568, opened 2025-07-01
feat(frontend): migrate to zod v4

peterpeterparker:
# Motivation

# Changes

# Tests


DenysKarmazynDFINITY (approved):
Massive job! The only thing I noticed is that our `z` imports are not aligned (in some files, we use `* as z`, in the other - `{ z }`), but that's just cosmetics, we can tackle it separately if we want to.

LGTM, thanks!

PR #7577, opened 2025-07-01
feat(backend):  add implementation btc_get_current_fee_percentiles endpoint

DecentAgeCoder:
# Motivation

This PR implements the `btc_get_current_fee_percentiles` endpoint, which is used by the frontend to calculate transaction fees. The primary goal is to reduce the execution time for sending BTC by loading fee data from a cache that is refreshed at a fixed interval of one minute.

By decoupling the fee calculation from real-time queries and relying on a periodically updated cache, the system can better handle varying traffic load, allowing more frequent calls to `btc_get_current_fee_percentiles` during low traffic, while reducing backend pressure during high traffic periods.

**Note:** Further details for this this change is described by this  [design proposal](https://github.com/dfinity/oisy-doc/blob/main/proposals/0001-improve-performance-for-btc-fee-calculation.md).

**Important background:**
- The execution time for `bitcoin_get_current_fee_percentiles` is between 6 and 8 seconds
- The newly introduced endpoint `bitcoin_get_current_fee_percentiles` is deliberately defined as query since a scheduler will load the fee percentile in a fixed interval in the background
- The existing endpoint `btc_select_user_utxos_fee` will be completely removed once all related backend and frontend changes have been merged.

# Changes

- Implemented a thread-local cache for fee percentiles stored per Bitcoin network.
- Added timer-based periodic updates for the fee percentile cache (`update_fee_percentiles_cache`).
- Modified `get_current_fee_percentiles` to use only cached data and return an error if cache data is unavailable.
- Introduced initialization of the fee percentiles cache (init_fee_percentiles_cache) during module init/upgrade 
- Replaced the `caller_is_controller` guard in `btc_get_current_fee_percentiles` with caller_is_not_anonymous.
- Updated `btc_get_current_fee_percentiles` to use the new cache logic for retrieving fee data.
- Removed unused imports across the affected files.

# Tests

Not avilable (Unfortunately it is not possible to provide canister tests for bitcoin related services).


src/backend/src/bitcoin_api.rs
```diff
@@ -51,12 +67,61 @@ pub async fn get_all_utxos(
...
+/// in the thread-local cache. Fetches current fee data from the bitcoin canister and stores
+/// it for quick access by other functions.
+async fn update_fee_percentiles_cache() -> Result<(), String> {
+    // Update for each network type
+    for network in &[
+        BitcoinNetwork::Mainnet,
+        BitcoinNetwork::Testnet,
+        BitcoinNetwork::Regtest,
+    ] {
+        match fetch_current_fee_percentiles(*network).await {
```
bitdivine: You can dispatch these in parallel, if you wish, with something like:
```
  let results = join_all(networks.map(fetch_current_fee_percentiles));
```
or
```
  let results = join_all(networks.map(|network| {
     match fetch_current_fee_percentiles(network) {
     ....
     }
  }));
```
Not required, though!

src/backend/src/bitcoin_api.rs
```diff
@@ -66,16 +131,103 @@ async fn get_current_fee_percentiles(
...
+    network: BitcoinNetwork,
+) -> Result<Vec<MillisatoshiPerByte>, String> {
+    // Try to get from cache first
+    let cached_percentiles =
+        FEE_PERCENTILES_CACHE.with(|cache| cache.borrow().get(&network).cloned());
+
+    match cached_percentiles {
+        Some(percentiles) if !percentiles.is_empty() => {
+            // Use cached values
+            ic_cdk::println!("Using cached fee percentiles for network {:?}", network);
```
bitdivine: You probably don't want high volume printouts like this in production.  Maybe in staging but delete it once you are confident that everything works as expected.

src/backend/src/bitcoin_api.rs
DecentAgeCoder: Thanks, this is an excellent hint. I have adapted the code to call the fetch_current_fee_percentiles endpoint in parallel.

src/backend/src/bitcoin_api.rs
DecentAgeCoder: I completely agree. This output is a leftover from debugging the code. I have removed the statement.

PR #7581, opened 2025-07-01
feat(frontend): Add ERC20 to custom token mapper

AntonioVentilii:
# Motivation

After having updated the backend, we can map the ERC20 tokens as `CustomToken`.

# Changes

- Change types of custom tokens to include ERC20 variant from backend. We make it more scalable with new backend types.
- Create mapper for ERC20 Custom Tokens.
- Adapt util `toCustomToken` to handle all cases and raise error to unhandled cases.

# Tests

Created new test.


roman-nazaruk (approved):
LGTM, thnx

PR #7587, opened 2025-07-01
refactor(frontend): Tokenlist filter save button

daviddecentage:
# Motivation

We do some final minor adjustments to the tokenlist filter enable/disable save button.

# Changes

- Added loading state to the save (apply) button
- Remove (0) in the save button when no elements are selected

# Tests

https://github.com/user-attachments/assets/d6e322db-6ea8-4c1a-9632-2c6a17beee38


src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -99,14 +99,19 @@
...
+		saveLoading = true;
 		await saveAllCustomTokens({ tokens: modifiedTokens, $authIdentity, $i18n });
 
 		// we need to update the filter list after a save to ensure the tokens got the newest backend "version"
 		updateFilterList($tokenListStore.filter);
+		saveLoading = false;
 	};
 
 	let modifiedTokens: Record<string, Token> = $state({});
+	let modifiedTokensLen: number = $derived(Object.keys(modifiedTokens).length);
```
AntonioVentilii:
```suggestion
	let modifiedTokensLen = $derived(Object.keys(modifiedTokens).length);
```

PR #7596, opened 2025-07-02
feat(frontend): Specify type for `getAccountInfo` method of Solana RPC canister

AntonioVentilii:
# Motivation

The Solana RPC canister returns an unparsed JSON for the info of the account data. Since we parse it during the request, we need to adapt the type for the correct output (the return is similar to what we are used with the Solana RPC requests done through Alchemy).


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7597, opened 2025-07-02
feat(frontend): Use Solana RPC canister when mapping instructions

AntonioVentilii:
# Motivation

We need to start somewhere to use the Solana RPC canister, with the method `getAccountInfo`.

However, different from what we used to do until now (query the Solana RPC through Alchemy), this method requires an `Identity` as input too.

That causes quite a lot of changes, since we need to pass down the `identity` through all the functions/components/workers.

# Changes

- Substitute `getAccountInfo` from Alchemy query with `getAccountInfo` from Solana RPC canister in util that parses instructions.
- Add `identity` as `OptionIdentity` as parameter to all the functions that at some point uses the mappers (this involves workers too, but they are already set to handle an identity parameter through the class `Scheduler`).

# Tests

The current (adapted) tests are sufficient. But we have a practical example in test canister (nothing changed):


DenysKarmazynDFINITY (approved):
I assume the nonNullish identity check happens somewhere inside the underlying methods/services?

The code LGTM, thanks!

PR #7606, opened 2025-07-02
feat(frontend): Accept minimum consensus from the Solana RPC canister

AntonioVentilii:
# Motivation

The Solana RPC canister could return an Inconsistent answer when queried, since one of the provider could not be available for that endpoint. To facilitate the current development, we allow a minimum of 2 consistent answers.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7607, opened 2025-07-02
fix(frontend): Landing page vertical alignment

daviddecentage:
# Motivation

When we made the body height `inherit` to enable scroll events, it lead to the landing page not being vertically centered anymore.

# Changes

Adding a fixed height of 100dvh to a wrapper on the app (only in logged out state)

For logged in state its not needed as all content is aligned on top.

# Tests

Before:
image

After:
image

Logged in still ok:
image

### DIFF
--- a/src/frontend/src/routes/(app)/+layout.svelte
+++ b/src/frontend/src/routes/(app)/+layout.svelte
@@ -45,44 +45,45 @@
 	});
 </script>
 
-<div
-	class="relative min-h-[640px] pb-5 md:pb-0 lg:flex lg:h-full lg:flex-col"
-	class:overflow-hidden={$authNotSignedIn}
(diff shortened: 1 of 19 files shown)


src/frontend/src/routes/(app)/+layout.svelte
```diff
@@ -54,6 +54,7 @@
 	class:md:flex={$authNotSignedIn}
 	class:md:flex-col={$authNotSignedIn}
 	class:md:h-full={$authNotSignedIn}
+	class:h-[100dvh]={$authNotSignedIn}
```
DenysKarmazynDFINITY:
```suggestion
	class:h-dvh={$authNotSignedIn}
```

DenysKarmazynDFINITY (approved):

PR #7608, opened 2025-07-02
fix(frontend): Solana SPL hide not working

daviddecentage:
# Motivation

The hide solana custom token in the TokenMenu currently can not work as its treated as an Erc20 Token.

# Changes

- Added own SolHideTokenModal
- Added modal to the store and Modals root file
- Adjusted condition when opening the modal to handle SPL tokens correctly

# Tests

https://github.com/user-attachments/assets/450c800f-943a-413e-aea6-319184738153


src/frontend/src/sol/components/tokens/SolHideTokenModal.svelte
```diff
@@ -0,0 +1,60 @@
...
+	import { i18n } from '$lib/stores/i18n.store';
+	import { toastsError } from '$lib/stores/toasts.store';
+	import { token } from '$lib/stores/token.store';
+	import { isNullishOrEmpty } from '$lib/utils/input.utils';
+	import { saveCustomTokens } from '$sol/services/spl-custom-tokens.services';
+	import type { SplCustomToken } from '$sol/types/spl-custom-token';
+
+	export let fromRoute: NavigationTarget | undefined;
+
+	let selectedToken: SplCustomToken;
```
DenysKarmazynDFINITY, line 21: Since it's just about updating 2 lines of code - let's make this component Svelte5 right away

daviddecentage: updated

PR #7610, opened 2025-07-02
feat(frontend): Use Sol RPC Canister getAccountInfo in the entire code

AntonioVentilii:
# Motivation

# Changes

# Tests


AntonioVentilii:
For now, we are not going to use the SOL RPC canister

PR #7612, opened 2025-07-02
refactor(frontend): Tokenlist placeholder

daviddecentage:
# Motivation

We adjust the tokenlist placeholder and spacings around it to make it not so long that the "Enable more assets" could be missed and also change the text.

# Changes

- Adjusted texts
- Adjusted/reduced spacings on the tokenlist
- Reduced size of placeholder icon

# Tests

image


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7625, opened 2025-07-03
feat(frontend): use assertIndexCanisterId for Icrc tokens editing

DenysKarmazynDFINITY:
# Motivation

We need to improve the token editing feature by asserting indexCanisterId before saving it. Mostly, it is needed to check whether an index id matches with the provided ledger id. 

Also, we need to improve the onSuccess callback of `loadCustomTokens` - previously, it was only triggered after query calls, but since we cannot guarantee that they succeed, it's better to call `onSuccess` on both certified and uncertified calls.

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokenModal.svelte
+++ b/src/frontend/src/lib/components/tokens/TokenModal.svelte
@@ -6,6 +6,7 @@
 	import { erc20UserTokensStore } from '$eth/stores/erc20-user-tokens.store';
 	import { isTokenErc20UserToken } from '$eth/utils/erc20.utils';
 	import IcAddTokenForm from '$icp/components/tokens/IcAddTokenForm.svelte';
+	import { assertIndexLedgerId } from '$icp/services/ic-add-custom-tokens.service';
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/tokens/TokenModal.svelte
```diff
@@ -229,9 +230,26 @@
 
 		try {
 			if (icTokenIcrcCustomToken(tokenToEdit)) {
+				loading = true;
 				gotoStep(TokenModalSteps.EDIT_PROGRESS);
 				progress(ProgressStepsAddToken.SAVE);
 
+				const { valid } = !isNullishOrEmpty(icrcTokenIndexCanisterId)
```
AntonioVentilii: nitpick: this could be `notEmptyString`

AntonioVentilii (approved):
LGTM tks

src/frontend/src/lib/components/tokens/TokenModal.svelte
DenysKarmazynDFINITY: true, good catch! update it here + in another place in this file

PR #7626, opened 2025-07-03
feat(frontend): New type for ERC20 Custom tokens

AntonioVentilii:
# Motivation

Starting to normalize the Custom tokens among the network, we first create a type for ERC20 custom tokens. It should be the same as the ICRC and SPL ones.

### DIFF
--- a/src/frontend/src/eth/types/erc20-custom-token.ts
+++ b/src/frontend/src/eth/types/erc20-custom-token.ts
@@ -1,0 +1,4 @@
+import type { Erc20Token } from '$eth/types/erc20';
+import type { UserToken } from '$lib/types/user-token';
+
+export type Erc20CustomToken = UserToken<Erc20Token>;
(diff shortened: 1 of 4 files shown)


src/frontend/src/eth/types/erc20-custom-token.ts
```diff
@@ -0,0 +1,4 @@
+import type { Erc20Token } from '$eth/types/erc20';
+import type { UserToken } from '$lib/types/user-token';
+
+export type Erc20CustomToken = UserToken<Erc20Token>;
```
DenysKarmazynDFINITY: Should it be `CustomToken<Erc20Token>`? Or is this change planned for a later point - then maybe a small TODO comment?

AntonioVentilii: You are completely right: the whole point is to use `CustomToken` eheh! Fixed

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7629, opened 2025-07-03
fix(frontend): Sticky headers

daviddecentage:
# Motivation

Theres two issues with the sticky headers that we need to fix:

1. If we have multiple sticky headers in one page, the lower ones should slide on top of the others that have been scrolled past. But if the upper ones have elements that might be wider than the usual content (e.g. the input for the search filter is 1-2px wider than the content) it can be visible because the sticky headers dont cover enough width and dont overlap all that is below.
2. If we calculate the original height of the headers, it can happen that we receive too high values leading in a lot of spacing. This is due to the breaks that happen in longer texts when initially rendering before we set the full width.

# Changes

1. Make sticky headers wider by 1 unit (1em) and introduce a padding of px-1 to shift the content to the correct alignment
2. Introduce whitespace-nowrap so initially longer texts cant wrap and the correct actual height is taken when determining the original height.

# Tests

Correct sizing:
image

No overlap anymore:
image

### DIFF
--- a/src/frontend/src/eth/types/erc20-custom-token.ts
+++ b/src/frontend/src/eth/types/erc20-custom-token.ts
@@ -1,0 +1,4 @@
+import type { Erc20Token } from '$eth/types/erc20';
+import type { CustomToken } from '$lib/types/custom-token';
+
+export type Erc20CustomToken = CustomToken<Erc20Token>;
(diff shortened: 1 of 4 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7633, opened 2025-07-03
chore(frontend): Add Italian support

StefanBerger-DFINITY:
# Motivation

We recently added german as additional language, and now we should add others asap.

# Changes

- add Italian

# Tests


AntonioVentilii (approved):
E andiamo!!! LGTM!

PR #7638, opened 2025-07-03
refactor(frontend): Extract type for load custom tokens params

AntonioVentilii:
# Motivation

To simplify the code, we create a specific type for the params of the functions that load custom tokens (and similar for user tokens).


src/frontend/src/sol/services/spl.services.ts
```diff
@@ -73,38 +70,19 @@ export const loadCustomTokens = ({
...
-	useCache?: boolean;
-}): Promise<SplCustomToken[]> => {
+const loadCustomTokensWithMetadata = async (
+	params: LoadCustomTokenParams
+): Promise<SplCustomToken[]> => {
 	const loadCustomContracts = async (): Promise<SplCustomToken[]> => {
-		if (isNullish(identity)) {
-			await nullishSignOut();
-			return [];
-		}
```
AntonioVentilii, line 105: this is already handled in the function `loadSplCustomTokens`

PR #7646, opened 2025-07-04
feat(frontend): Provide required utilities for BTC performance improvement 

DecentAgeCoder:
# Improve performance for BTC fee and UTXO calculation (5)

## Motivation

This PR adds new utilities related to the UTXO's and fees performance.

Note: 
- Further details for this this change is described by this [design proposal](https://github.com/dfinity/oisy-doc/blob/main/proposals/0001-improve-performance-for-btc-fee-calculation.md).
- The complete feature branch for better understanding this PR can be found here [here](https://github.com/dfinity/oisy-wallet/pull/7545)

**Important background:**

- The execution time for bitcoin_get_current_fee_percentiles is between 6 and 8 seconds
- The newly introduced endpoint bitcoin_get_current_fee_percentiles is deliberately defined as query since a scheduler will load the fee percentile in a fixed interval in the background
- The existing endpoint btc_select_user_utxos_fee will be completely removed once all related backend and frontend changes have been merged.

# Changes
**Adding btc-utxos.utils.ts**
- Utility functions to handle UTXOs: transaction ID conversion, UTXO selection, fee estimation, UTXO filtering, and fee calculation.

**Enhancing btc.utils.ts:**
- Added `getPendingTransactionIds` to retrieve locked UTXOs.
- New method `mapToBitcoinNetwork` to map Bitcoin networks.

**Updating backend.api.ts**
- Added API methods to fetch current BTC fee percentiles.

# Tests

New Tests `btc-utxos.utils.spec.ts` for testing the functionality for `btc-utxos.utils.ts`

### DIFF
--- a/src/frontend/src/icp/utils/btc.utils.ts
+++ b/src/frontend/src/icp/utils/btc.utils.ts
@@ -1,6 +1,6 @@
 import { btcPendingSentTransactionsStore } from '$btc/stores/btc-pending-sent-transactions.store';
 import type { PendingTransaction } from '$declarations/backend/backend.did';
-import { isNullish, uint8ArrayToHexString } from '@dfinity/utils';
+import { isNullish, notEmptyString, uint8ArrayToHexString } from '@dfinity/utils';
 import { get } from 'svelte/store';
 
 /**


src/frontend/src/btc/utils/btc-utxos.utils.ts
```diff
@@ -0,0 +1,176 @@
...
+	availableUtxos,
+	amountSatoshis,
+	feeRateSatoshisPerVByte
+}: {
+	availableUtxos: Utxo[];
+	amountSatoshis: bigint;
+	feeRateSatoshisPerVByte: bigint;
+}): UtxoSelectionResult => {
+	if (availableUtxos.length === 0) {
+		throw new Error('No UTXOs available for selection');
```
DenysKarmazynDFINITY, line 61: Do we really want to throw an error here? I'd say depending how do we want to handle it on the components side, the util can just return `undefined` if no available UTXOs found.

src/frontend/src/btc/utils/btc-utxos.utils.ts
```diff
@@ -0,0 +1,176 @@
...
+	// If we get here, we don't have sufficient funds
+	const finalTxSize = estimateTransactionSize({
+		numInputs: selectedUtxos.length,
+		numOutputs: 2
+	});
+	const finalEstimatedFee = BigInt(finalTxSize) * feeRateSatoshisPerVByte;
+	const totalRequired = amountSatoshis + finalEstimatedFee;
+
+	throw new Error(
+		`Insufficient funds: required ${totalRequired} satoshis (${amountSatoshis} and ${finalEstimatedFee} fee), available ${totalInputValue} satoshis`
```
DenysKarmazynDFINITY, line 109: In this case, I would suggest returning the numbers instead of the hardcoded untranslated string and let consumers (components) handle the data - e.g. by displaying a warning message inside the Send modal. By throwing an error, we are kind of forced to display the message in a toast which is not the best UX decision in this case.

src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -8,3 +13,43 @@ import { uint8ArrayToHexString } from '@dfinity/utils';
...
 	uint8ArrayToHexString(Uint8Array.from(txid).toReversed());
+
+/**
+ * Get pending transaction IDs from the store to exclude locked UTXOs
+ */
+export const getPendingTransactionIds = (address: string): string[] => {
+	const storeData = get(btcPendingSentTransactionsStore);
+	const pendingTransactions = storeData[address];
+
+	if (isNullish(pendingTransactions) || !pendingTransactions.data) {
```
```suggestion
	if (isNullish(pendingTransactions?.data)) {
```

src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -8,3 +13,43 @@ import { uint8ArrayToHexString } from '@dfinity/utils';
...
+
+			// Handle number array case (txid: number[])
+			if (Array.isArray(tx.txid)) {
+				return tx.txid.map((b: number) => b.toString(16).padStart(2, '0')).join('');
+			}
+
+			// Fallback, convert to string representation
+			return String(tx.txid);
+		})
+		.filter(Boolean);
```
DenysKarmazynDFINITY: `map` + `filter` can be replaced with one single `reduce`. This way we don't need to loop through the array 2 times.

src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -8,3 +13,43 @@ import { uint8ArrayToHexString } from '@dfinity/utils';
...
+};
+
+export const mapToBitcoinNetwork = ({
```
DenysKarmazynDFINITY: We already have such a util, please check `mapToSignerBitcoinNetwork` of `network.utils`.

src/frontend/src/btc/utils/btc-utxos.utils.ts
DecentAgeCoder: Throwing an error is the cleaner approach here. When there are no available UTXOs, this represents a genuine failure condition that prevents the transaction from proceeding, not just an alternative outcome that should be handled gracefully. I suggest that exceptional conditions should be handled via exceptions rather than special return values. If it's more about throwing an error and not being able to identify the error I rather suggest to return an optional error code that clearly defines an error state. What do you think?

src/frontend/src/btc/utils/btc-utxos.utils.ts
DecentAgeCoder: I agree in this case we should not throw an error. Do you mean with "numbers" the totalRequired funds? Or return something like "totalMissing" , handling it accordingly by the caller if not undefined? Or should we introduce an optional error code in UtxoSelectionResult?

src/frontend/src/icp/utils/btc.utils.ts
DecentAgeCoder: I have removed the duplicated function

src/frontend/src/btc/utils/btc-utxos.utils.ts
DenysKarmazynDFINITY: Ok, let's keep it as it is. I would probably translate the string since we now support several languages, but we can do it separately.

src/frontend/src/btc/utils/btc-utxos.utils.ts
DenysKarmazynDFINITY: Since this PR is already massive enough, let's keep this part also as it is. Handling the insufficient UTXOs error/warning gracefully can be considered as an additional improvement which can be done separately.

src/frontend/src/icp/utils/btc.utils.ts
DecentAgeCoder: Using reduce instead of map + filter now

src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -8,3 +11,38 @@ import { uint8ArrayToHexString } from '@dfinity/utils';
...
+		else if (Array.isArray(tx.txid)) {
+			txidString = tx.txid.map((b: number) => b.toString(16).padStart(2, '0')).join('');
+		}
+		// Fallback, convert to string representation
+		else {
+			txidString = String(tx.txid);
+		}
+
+		// Add to accumulator only if truthy (combines map + filter(Boolean))
+		return txidString ? [...acc, txidString] : acc;
```
```suggestion
		return notEmptyString(txidString) ? [...acc, txidString] : acc;
```

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7651, opened 2025-07-04
chore(frontend): Send input destination svelte 5

daviddecentage:
# Motivation

Since we need to make an adjustment in that component, we upgrade it to svelte 5.

# Changes

- Upgraded component to svelte 5

# Tests

Manually tested, working

### DIFF
--- a/src/frontend/src/lib/components/send/SendInputDestination.svelte
+++ b/src/frontend/src/lib/components/send/SendInputDestination.svelte
@@ -48,13 +48,9 @@
 
 	const convertContext = getContext<ConvertContext>(CONVERT_CONTEXT_KEY);
 
-	const sendTokenNetworkId = $derived(
-		nonNullish(sendContext) ? sendContext.sendTokenNetworkId : undefined
-	);
+	const sendTokenNetworkId = nonNullish(sendContext) ? sendContext.sendTokenNetworkId : undefined;
 
-	const destinationToken = $derived(
-		nonNullish(convertContext) ? convertContext.destinationToken : undefined
-	);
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/components/send/SendInputDestination.svelte
```diff
@@ -35,50 +48,60 @@
 
 	const convertContext = getContext<ConvertContext>(CONVERT_CONTEXT_KEY);
 
-	const sendTokenNetworkId = nonNullish(sendContext) ? sendContext.sendTokenNetworkId : undefined;
+	const sendTokenNetworkId = $derived(
```
roman-nazaruk: Should it use a derived store or state?

src/frontend/src/lib/components/send/SendInputDestination.svelte
```diff
@@ -35,50 +48,60 @@
...
+		nonNullish(sendContext) ? sendContext.sendTokenNetworkId : undefined
+	);
 
-	const destinationToken = nonNullish(convertContext) ? convertContext.destinationToken : undefined;
+	const destinationToken = $derived(
```
roman-nazaruk: Same here

src/frontend/src/lib/components/send/SendInputDestination.svelte
daviddecentage: Discussed offline: const without state is sufficient because its just the reference to the store

src/frontend/src/lib/components/send/SendInputDestination.svelte
daviddecentage: Discussed offline: const without state is sufficient because its just the reference to the store

PR #7654, opened 2025-07-04
feat(frontend): Provide review service implementation

DecentAgeCoder:
# Motivation

# Changes

# Tests


DenysKarmazynDFINITY:
Hey @DecentAgeCoder, here is my view on your qiestions:

* `Should the service logic currently defined in btc-review.services.ts be moved to either btc-utxos.service.ts or btc.service.ts for better cohesion?` - yes, `btc-utxos.service` sounds good to me
* `Is the name selectUtxosFee appropriate for the new function, or should we consider a more descriptive alternative such as prepareBtcSend(..)?` - `prepareBtcSend` 👍 
* `Is it ok to introduce new assertion functions primarily intended to enforce specific input states?` - also a good idea, we can merge those ones separately from this PR, as well as cover them with tests 
* `A new env variable CKBTC_MINTER_CANISTER_ID has been introduced which is used to retrieve the bitcoin canister id. What would the correct way be to retrieve the bitcoin canister ID?` - we already have `IC_CKBTC_MINTER_CANISTER_ID` (networks.icrc.env.ts) in place

src/frontend/src/btc/services/btc-utxos.service.ts
```diff
@@ -0,0 +1,148 @@
...
+	identity,
+	network,
+	amount,
+	source
+}: BtcReviewServiceParams): Promise<BtcReviewResult> => {
+	assertNonNullish(identity);
+	assertStringNotEmpty({ value: source, message: 'Source address is required' });
+	assertNonNullish(IC_CKBTC_MINTER_CANISTER_ID);
+
+	// assertAmount({ amount });
```
DenysKarmazynDFINITY:
```suggestion
	
```

src/frontend/src/btc/services/btc-utxos.service.ts
```diff
@@ -0,0 +1,148 @@
...
+ * Main orchestrator function that replaces the backend btc_select_user_utxos_fee call
+ * This function coordinates all the steps needed to select UTXOs and calculate fees
+ */
+export const prepareBtcSend = async ({
+	identity,
+	network,
+	amount,
+	source
+}: BtcReviewServiceParams): Promise<BtcReviewResult> => {
+	assertNonNullish(identity);
```
DenysKarmazynDFINITY: We have a standartized way of making such an assert, please use it here too:
```
			await nullishSignOut();
			return;
		}
```

src/frontend/src/lib/utils/asserts.ts
```diff
@@ -0,0 +1,84 @@
+import type { Amount } from '$lib/types/send';
+
+/**
+ * Asserts that an array has at least one element
+ * @param params - Object containing array and optional message
+ * @param params.array - The array to check
+ * @param params.message - Optional custom error message
+ * @throws Error if array is empty
+ */
+export const assertArrayNotEmpty = <T>(params: { array: T[]; message?: string }): void => {
```
DenysKarmazynDFINITY: Can you please add tests for all the assert functions? You can find some inspiration in `assert-amount.utils.spec.ts`

src/frontend/src/lib/utils/asserts.ts
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Asserts that a string is not empty or whitespace-only
+ * @param params - Object containing value and optional message
+ * @param params.value - The string to check
+ * @param params.message - Optional custom error message
+ * @throws Error if string is empty or whitespace-only
+ */
+export const assertStringNotEmpty = (params: { value: string; message?: string }): void => {
+	const { value, message = 'String must not be empty' } = params;
+	if (!value || value.trim().length === 0) {
```
DenysKarmazynDFINITY:
```suggestion
	if (isNullishOrEmpty(value) || value.trim().length === 0) {
```

src/frontend/src/lib/utils/asserts.ts
```diff
@@ -0,0 +1,84 @@
...
+};
+
+/**
+ * Asserts that a string is not empty or whitespace-only
+ * @param params - Object containing value and optional message
+ * @param params.value - The string to check
+ * @param params.message - Optional custom error message
+ * @throws Error if string is empty or whitespace-only
+ */
+export const assertStringNotEmpty = (params: { value: string; message?: string }): void => {
```
DenysKarmazynDFINITY:
```suggestion
export const assertStringNotEmpty = ({ value, message = 'String must not be empty' }: { value: string; message?: string }) => {
```

src/frontend/src/lib/utils/asserts.ts
```diff
@@ -0,0 +1,84 @@
...
+};
+
+/**
+ * Asserts that an amount is valid (positive number or parseable positive string)
+ * @param params - Object containing amount and optional message
+ * @param params.amount - The amount to check (string or number)
+ * @param params.message - Optional custom error message
+ * @throws Error if amount is not a positive number or parseable positive string
+ */
+export const assertAmount = (params: { amount: Amount; message?: string }): void => {
```
DenysKarmazynDFINITY:
```suggestion
export const assertAmount = ({ amount, message = 'Amount must be a positive number' }: { amount: Amount; message?: string }) => {
```

src/frontend/src/lib/utils/asserts.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Asserts that there are adequate funds for a transaction
+ * @param params - Object containing required amount, available balance, and optional message
+ * @param params.required - The required amount in satoshis
+ * @param params.available - The available balance in satoshis
+ * @param params.message - Optional custom error message
+ * @throws Error if inadequate funds
+ */
+export const assertAdequateBalance = (params: {
```
DenysKarmazynDFINITY: Maybe `assertBalance`? We already use such a naming in the code.

Also, please destruct `params` the same way as suggested for the other utils

src/frontend/src/btc/services/btc-utxos.service.ts
DenysKarmazynDFINITY: Also, according to the type, `identity` is always present. Why do we need to assert it then?  If you meant to make it optional, you can use `OptionIdentity`

src/frontend/src/btc/services/btc-utxos.service.ts
```diff
@@ -0,0 +1,148 @@
...
+	assertStringNotEmpty({ value: source, message: 'Source address is required' });
```
DenysKarmazynDFINITY: Same question here - `source` is not an optional param

src/frontend/src/btc/services/btc-utxos.service.ts
```diff
@@ -0,0 +1,148 @@
...
+	assertNonNullish(IC_CKBTC_MINTER_CANISTER_ID);
```
DenysKarmazynDFINITY: `IC_CKBTC_MINTER_CANISTER_ID` is a string, no need for assert

src/frontend/src/btc/services/btc-utxos.service.ts
DecentAgeCoder: I completly agree. The param identity is already mandatory. I have removed

src/frontend/src/btc/services/btc-utxos.service.ts
DecentAgeCoder: I completly agree. The param source is already mandatory. I have removed it as well

src/frontend/src/btc/services/btc-utxos.service.ts
DecentAgeCoder: I have also removed this assert as well

src/frontend/src/btc/services/btc-utxos.service.ts
DecentAgeCoder: I have removed it as suggested

DecentAgeCoder:
> Member

I will give it a second thought whether I will remove the asserts.ts altogether

src/frontend/src/lib/utils/asserts.ts
DecentAgeCoder: I’ve already looked into the existing assertBalance. The issue with this implementation is that it doesn’t throw an error as one would typically expect from an assertion.

src/frontend/src/lib/utils/asserts.ts
DecentAgeCoder: The assert.ts file has been removed, as the necessary assertions are already sufficiently covered

src/frontend/src/lib/utils/asserts.ts
DecentAgeCoder: The assert.ts file has been removed, as the necessary assertions are already sufficiently covered

src/frontend/src/lib/utils/asserts.ts
DecentAgeCoder: The assert.ts file has been removed, as the necessary assertions are already sufficiently covered

src/frontend/src/lib/utils/asserts.ts
DecentAgeCoder: The assert.ts file has been removed, as the necessary assertions are already sufficiently covered

src/frontend/src/lib/utils/asserts.ts
DecentAgeCoder: The assert.ts file has been removed, as the necessary assertions are already sufficiently covered

src/frontend/src/btc/services/btc-utxos.service.ts
```diff
@@ -0,0 +1,151 @@
+import { convertNumberToSatoshis } from '$btc/utils/btc-send.utils';
+
```
DenysKarmazynDFINITY: nitpick of the year: can you please remove the empty line :)

DenysKarmazynDFINITY (approved):
LGTM, thanks!

src/frontend/src/btc/services/btc-utxos.service.ts
DecentAgeCoder: I have removed the empty line

PR #7655, opened 2025-07-04
feat(frontend): Completely terminate ck-related workers on destroy

AntonioVentilii:
# Motivation

When we destroy a component that initiate a worker, we stop such worker. However, we do not destroy it.

To fix it, we create a new function called `destroy`.

In this PR, we tackle the ck-related workers.

# Changes

- Refactor a sub-function for `stop` in the workers.
- Create `destroy` method, using `worker.terminate`.

# Tests

Existing and new tests suffice.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7657, opened 2025-07-04
chore(frontend): Send input destination reset button

daviddecentage:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/components/send/SendInputDestination.svelte
+++ b/src/frontend/src/lib/components/send/SendInputDestination.svelte
@@ -49,13 +49,9 @@
 
 	const convertContext = getContext<ConvertContext>(CONVERT_CONTEXT_KEY);
 
-	const sendTokenNetworkId = $derived(
-		nonNullish(sendContext) ? sendContext.sendTokenNetworkId : undefined
-	);
+	const sendTokenNetworkId = nonNullish(sendContext) ? sendContext.sendTokenNetworkId : undefined;
 
-	const destinationToken = $derived(
-		nonNullish(convertContext) ? convertContext.destinationToken : undefined
-	);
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/send/SendInputDestination.svelte
```diff
@@ -121,9 +124,20 @@
 			on:focus={onFocus}
 			on:blur={onBlur}
 			on:nnsInput
+			bind:inputElement
 		>
 			{#snippet innerEnd()}
-				<span class="flex">
+				<span class="flex gap-1 bg-primary">
+					{#if notEmptyString(destination)}
+						<ButtonReset
```
AntonioVentilii, line 132: aria label? and a test maybe?

src/frontend/src/lib/components/send/SendInputDestination.svelte
```diff
@@ -121,9 +124,20 @@
...
+							onclick={() => {
+								destination = '';
```
AntonioVentilii, line 134: not for this PR:

I was thinking, what if we bind the value directly? the scope of `ButtonReset` is clear, so we could just do `bind:value={destination}` and handle the reset directly inside the button, no?
However, we should be able to handle any other callback (for example in this case the focus on the input)

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/components/send/SendInputDestination.svelte
daviddecentage: The default will be fine

src/frontend/src/lib/components/send/SendInputDestination.svelte
daviddecentage: I can have a look at it, should have some time later

PR #7659, opened 2025-07-04
refactor(frontend): getDisabledOrModifiedTokens utils

daviddecentage:
# Motivation

We extract some logic to utils so the full feature is tested.

# Changes

- Moved filtering logic for enabled/disabled tokens to utils
- Adjusted usage by using the new utils function

# Tests

Added tests for new utils function


AntonioVentilii (approved):
LGTM tks!

PR #7669, opened 2025-07-07
fix(frontend): Pass data from WalletConnect to fee context

AntonioVentilii:
# Motivation

When WalletConnect send the requests, for ERC20, it may happen that the first transaction is an approval. In that case, to estimate the gas (the one that we show in the frontend review UI), we need to pass the data of the transaction.

# Changes

- Allow `EthFeeContext` component to receive optional `data` params.
- Extract data from the transaction in WalletConnect flow.
- Pass it through to calculate the estimated gas.

# Tests

### Before

Screenshot 2025-07-07 at 13 45 29

### After

Screenshot 2025-07-07 at 13 43 12


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7672, opened 2025-07-07
feat(frontend): integrating prepareBtcSend in UI's

DecentAgeCoder:
# Improve performance for BTC fee and UTXO calculation (7)

## Motivation

This PR removes the deprecated selectUtxosFee(..) function and updates the relevant Svelte components to use the new prepareBtcSend(..) function. The backend endpoint previously used by selectUtxosFee(..) will remain temporarily in place to enable a quick frontend rollback if any issues occur. It will be removed in a final cleanup PR once the new solution has proven to be stable.

Note:
- Further details for this this change is described by this [design proposal](https://github.com/dfinity/oisy-doc/blob/main/proposals/0001-improve-performance-for-btc-fee-calculation.md).
- The complete feature branch for better understanding this PR can be found here [here](https://github.com/dfinity/oisy-wallet/pull/7545)

**Important background:**

- The execution time for `bitcoin_get_current_fee_percentiles` is between 6 and 8 seconds
- The newly introduced endpoint bitcoin_get_current_fee_percentiles is deliberately defined as query since a scheduler will load the fee percentile in a fixed interval in the background
- The existing endpoint `btc_select_user_utxos_fee` will be completely removed once all related backend and frontend changes have been merged.

# Changes

- Refactored BTC fee estimation logic to include an additional source parameter in components (`BtcConvertTokenWizard`, `UtxosFeeContext`, `BtcSendReview`, and `BtcUtxosFee`).
- Replaced `selectUtxosFee` API calls with a new preparation service `prepareBtcSend` for improved fee processing.
- Updated error-handling and validation in fee-related logic (e.g., source parameter checks).

# Tests

- Adapted tests in `BtcConvertTokenWizard.spec.ts` to align with the new `prepareBtcSend`
- Adapted tests in `BtcUtxosFee.spec.ts` to align with the new `prepareBtcSend`
- Adapted tests in `BtcUtxosFeeContext.spec.ts` to align with the new `prepareBtcSend`


src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
```diff
@@ -43,6 +44,8 @@
 		if (
 			amountError ||
 			isNullish(networkId) ||
+			isNullish(source) ||
+			source === '' ||
```
DenysKarmazynDFINITY: Please replace these 2 checks with `isNullishOrEmpty` (even thought it cannot be nullish, for convenience let's use that method)

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
```diff
@@ -31,20 +34,26 @@
 	const selectUtxosFee = async () => {
 		try {
 			// all required params should be already defined at this stage
-			if (isNullish(amount) || isNullish(networkId) || isNullish($authIdentity)) {
+			if (
+				isNullish(amount) ||
+				isNullish(networkId) ||
+				isNullish($authIdentity) ||
+				isNullish(source)
```
DenysKarmazynDFINITY: `source` is a required prop and cannot be nullish

src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
```diff
@@ -60,28 +63,32 @@
...
 
-		const utxosFee = nonNullish(network)
-			? await selectUtxosFeeApi({
-					amount: parsedAmount,
-					network,
-					identity: $authIdentity
-				})
-			: undefined;
+		if (nonNullish(network)) {
+			const result = await prepareBtcSend({
```
DenysKarmazynDFINITY: `prepareBtcSend` can throw an error, therefore you need to add try/catch here. Also, right now, the error messages generated inside that function are ignored - we need to display this information to the users either via toast or via a warning message inside modal. Please align with the Design team what is the better UX approach here.

src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
DecentAgeCoder: I can not apply the `isNullishOrEmpty` to the networkId (only to source), since it expects an `OptionString` (which is `string | null | undefined`), but `networkId` is of type `NetworkId | undefined`

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
DecentAgeCoder: Nice one! I have removed the check for source

src/frontend/src/btc/components/fee/UtxosFeeContext.svelte
DecentAgeCoder: I have simplified and centralized the handling of insufficient funds errors within BtcSendReview. The business logic has now been moved into the service layer, where it belongs, instead of residing in the Svelte component (see the removal of the reactive insufficientFundsForFee variable)

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
```diff
@@ -36,15 +39,16 @@
...
+				? await prepareBtcSend({
+						identity: $authIdentity,
 						network,
-						identity: $authIdentity
+						amount,
+						source
 					})
 				: undefined;
 		} catch (err: unknown) {
+			console.error('Error selecting utxos fee', err);
```
DenysKarmazynDFINITY: Not sure if we need a console.error on top of the toast - if you find it useful, we can keep it

src/frontend/src/btc/services/btc-utxos.service.ts
```diff
@@ -95,56 +88,60 @@ export const prepareBtcSend = async ({
 		feeRateSatoshisPerVByte
 	});
 
-	// Step 5: Calculate the final fee based on selected UTXOs
-	const feeSatoshis = calculateFinalFee({ selection, amountSatoshis });
```
DenysKarmazynDFINITY: So this step is now integrated into the `getFeeRateFromPercentiles` method?

src/frontend/src/btc/services/btc-utxos.service.ts
```diff
@@ -95,56 +88,60 @@ export const prepareBtcSend = async ({
...
 	const response = await getCurrentBtcFeePercentiles({
 		identity,
-		network: bitcoinNetwork
+		network: mappedNetwork
 	});
 
-	// Get fee percentiles array
-	const feePercentiles = response.fee_percentiles;
-	if (isNullish(feePercentiles) || feePercentiles.length === 0) {
+	const { fee_percentiles } = response;
```
DenysKarmazynDFINITY: you can destruct it directly on the line 116

src/frontend/src/btc/services/btc-utxos.service.ts
```diff
@@ -95,56 +88,60 @@ export const prepareBtcSend = async ({
...
+
+	if (!fee_percentiles || fee_percentiles.length === 0) {
```
DenysKarmazynDFINITY:
```suggestion
	if (isNullish(fee_percentiles) || fee_percentiles.length === 0) {
```

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
DecentAgeCoder: I will remove it

src/frontend/src/btc/services/btc-utxos.service.ts
DecentAgeCoder: Done

src/frontend/src/btc/services/btc-utxos.service.ts
DecentAgeCoder: - No, `getFeeRateFromPercentiles` is no longer integrated within `calculateFinalFee`. Instead, it’s now used inside the `calculateUtxoSelection` function:
```
const estimatedFee = BigInt(txSize) * feeRateSatoshisPerVByte;
const totalRequired = amountSatoshis + estimatedFee;
```
- This new approach is more direct and straightforward, UTXO selection and fee calculation are now handled together in a cohesive step.
- Additionally, MIN_FEE_RATE and MAX_FEE_RATE have been introduced to `getFeeRateFromPercentiles` as a safeguard against anomalies. This ensures that the resulting `feeRateSatsPerVByte` always stays within a safe and non-critical range.

DecentAgeCoder: - No, getFeeRateFromPercentiles it's not integrated in calculateFinalFee but is now contained in the `calculateUtxoSelection` function:
```
const estimatedFee = BigInt(txSize) * feeRateSatoshisPerVByte;
const totalRequired = amountSatoshis + estimatedFee;
```
This new approach is more direct and simple. In a nutshell the utxo selection and fee calculation go hand in hand.
- The MIN_FEE_RATE and MAX_FEE_RATE added to the getFeeRateFromPercentiles. This is an extra measure to prevent anomalises from beeing accepted. The feeRateSatsPerVByte will therfore always between a non critical fee value MIN_FEE_RATE and MAX_FEE_RATE

DecentAgeCoder:
The source code for this solution was already merged by the other other PR's

PR #7674, opened 2025-07-07
style(frontend): Correct background for toasts

AntonioVentilii:
# Motivation

If there is at least an error-styled toast in the list of toasts, all of them will be error-styled. We correct this with a specific CSS selector.

### Before

Screenshot 2025-07-07 at 15 58 56

### After

Screenshot 2025-07-07 at 15 45 50


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7677, opened 2025-07-07
feat(frontend): Safe estimate gas for Ethereum and EVMs

AntonioVentilii:
# Motivation

Since we accept an nullish estimated gas in the code, we can easily avoid raising error in case of estimation failure. Any issue with nullish gas fee is already handled.

### DIFF
--- a/src/frontend/src/eth/providers/infura.providers.ts
+++ b/src/frontend/src/eth/providers/infura.providers.ts
@@ -34,15 +34,13 @@ export class InfuraProvider {
 		try {
 			return await this.estimateGas(params);
 		} catch (err: unknown) {
-			// Log the error to console for debugging purposes
-			console.error(`Error estimating gas for network ${this.network}:`, err);
-
 			trackEvent({
 				name: TRACK_ETH_ESTIMATE_GAS_ERROR,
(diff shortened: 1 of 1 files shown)


src/frontend/src/eth/providers/infura.providers.ts
```diff
@@ -28,6 +28,16 @@ export class InfuraProvider {
 
 	estimateGas = (params: GetFeeData): Promise<bigint> => this.provider.estimateGas(params);
 
+	safeEstimateGas = async (params: GetFeeData): Promise<bigint | undefined> => {
+		try {
+			return await this.estimateGas(params);
+		} catch (error: unknown) {
+			// Log the error to console for debugging purposes
+			console.error('Error estimating gas:', error);
```
DenysKarmazynDFINITY: how about a Plausible event?

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/eth/providers/infura.providers.ts
AntonioVentilii: good one! included!

PR #7678, opened 2025-07-07
refactor(frontend): Merge nullish check for Solana network in mapper util

AntonioVentilii:
# Motivation

When we use util `mapNetworkIdToNetwork` we always check for undefined network in the end, or ignore an undefined network. So, to avoid repetition of code, we merge the logic directly inside the util.


DenysKarmazynDFINITY (approved):
Nice, thanks!

PR #7680, opened 2025-07-07
fix(frontend): No loader if networks disabled

daviddecentage:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/services/loader.services.ts
+++ b/src/frontend/src/lib/services/loader.services.ts
@@ -108,9 +108,6 @@ export const initLoader = async ({
 		return;
 	}
 
-	// We are loading the addresses from the backend. Consequently, we aim to animate this operation and offer the user an explanation of what is happening. To achieve this, we will present this information within a modal.
-	setProgressModal(true);
-
 	const { success: initSignerAllowanceSuccess } = await initSignerAllowance();
 
 	if (!initSignerAllowanceSuccess) {
@@ -131,6 +128,12 @@ export const initLoader = async ({
 	const networkIds: NetworkId[] = errorNetworkIds.filter((networkId) =>
 		enabledNetworkIds.includes(networkId)
 	);
+
+	// if no addresses need to be loaded, we dont show the progress modal
+	if (networkIds.length > 0) {
+		// We are loading the addresses from the backend. Consequently, we aim to animate this operation and offer the user an explanation of what is happening. To achieve this, we will present this information within a modal.
+		setProgressModal(true);
+	}
 
 	const { success: addressSuccess } = await loadAddresses(networkIds);
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
 				"@dfinity/principal": "^2.4.0",
 				"@dfinity/utils": "^2.13.0",
 				"@dfinity/verifiable-credentials": "^0.0.4",
-				"@dfinity/zod-schemas": "^0.0.2",
+				"@dfinity/zod-schemas": "^1.0.0",
 				"@metamask/detect-provider": "^2.0.0",
 				"@noble/hashes": "^1.8.0",
 				"@noble/secp256k1": "^2.3.0",
@@ -38,7 +38,7 @@
 				"idb-keyval": "^6.2.2",
 				"plausible-tracker": "^0.3.9",
 				"svelte-confetti": "^2.3.1",
-				"zod": "^3.25.56"
+				"zod": "^3.25.75"
 			},
 			"devDependencies": {
 				"@dfinity/eslint-config-oisy-wallet": "^0.1.14",
@@ -613,13 +613,13 @@
 			}
 		},
 		"node_modules/@dfinity/zod-schemas": {
-			"version": "0.0.2",
-			"resolved": "https://registry.npmjs.org/@dfinity/zod-schemas/-/zod-schemas-0.0.2.tgz",
-			"integrity": "sha512-sPVR9vLx4Vxx9MloCDvdS07k/2mGIiGophxT9TSwP1aFVHmibu6YzBRzdaji9h02FIgqrpS+LHX4deadf6M+Rg==",
+			"version": "1.0.0",
+			"resolved": "https://registry.npmjs.org/@dfinity/zod-schemas/-/zod-schemas-1.0.0.tgz",
+			"integrity": "sha512-5ApkpRO8hqTb7B9GH4H8FljY/r6hh3zpA/HFeeozIHieyebAzB748+4T9/oL6T7udkvlfWPMulbmjSHerm3B9A==",
 			"license": "Apache-2.0",
 			"peerDependencies": {
 				"@dfinity/principal": "^2.0.0",
-				"zod": "^3"
+				"zod": "^3.25"
 			}
 		},
 		"node_modules/@esbuild/aix-ppc64": {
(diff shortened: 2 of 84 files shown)


src/frontend/src/lib/services/loader.services.ts
```diff
@@ -132,6 +129,12 @@ export const initLoader = async ({
 		enabledNetworkIds.includes(networkId)
 	);
 
+	// if no addresses need to be loaded, we dont show the progress modal
+	if (networkIds.length > 0) {
+		// We are loading the addresses from the backend. Consequently, we aim to animate this operation and offer the user an explanation of what is happening. To achieve this, we will present this information within a modal.
+		setProgressModal(true);
```
AntonioVentilii: not really sure this is the solution: the progress modal is not only for the addresses

Most probably, we just need to not show the step where it loads the addresses, or make it really faster. I mean, is it not already like this?

IDB will throw errors for the missing addresses, but the initializer handles them based on enabled networks already:

```ts
	const errorNetworkIds: NetworkId[] = err?.map(({ networkId }) => networkId) ?? [];

	// We can fetch these values imperatively because these stores were just updated at the beginning of this same function, when loading the user profile.
	const enabledNetworkIds: NetworkId[] = [
		...(get(networkBitcoinMainnetEnabled) ? [BTC_MAINNET_NETWORK_ID] : []),
		...(get(networkEthereumEnabled) || get(networkEvmMainnetEnabled) ? [ETHEREUM_NETWORK_ID] : []),
		...(get(networkSolanaMainnetEnabled) ? [SOLANA_MAINNET_NETWORK_ID] : [])
	];

	// We don't need to load the addresses of the disabled networks.
	const networkIds: NetworkId[] = errorNetworkIds.filter((networkId) =>
		enabledNetworkIds.includes(networkId)
	);
	
	const { success: addressSuccess } = await loadAddresses(networkIds);
```

So, if the Bitcoin network is disabled, theoretically, even if IDB throws, we should not load the address. If this is not correct, we need to fix this part.

P.S. in general, how is this solution linked with bitcoin network?

AntonioVentilii: Just as info: I tried in local replica with bitcoin disabled and incognito mode, and it works as expected

Screenshot 2025-07-08 at 10 28 12

AntonioVentilii: tried with all networks disabled:

Screenshot 2025-07-08 at 10 30 53

daviddecentage: Discussed offline: Only loading the addresses for enabled networks deemed to be the cleanest solution to this problem.

src/frontend/src/lib/services/addresses.services.ts
```diff
@@ -32,13 +38,30 @@ export const loadAddresses = async (networkIds: NetworkId[]): Promise<ResultSucc
...
+
+	if (get(networkEthereumEnabled)) {
+		promisesList.push(loadIdbEthAddress());
+	}
+
+	if (get(networkSolanaMainnetEnabled)) {
+		promisesList.push(loadIdbSolAddressMainnet());
+	}
+
+	let results = await Promise.all(promisesList);
```
AntonioVentilii: maybe this? note that I included EVM networks too

```suggestion
const promisesList: Promise<ResultSuccess<LoadIdbAddressError>>[] = [
	...(get(networkBitcoinMainnetEnabled) ? [loadIdbBtcAddressMainnet()] : []),
	...(get(networkEthereumEnabled) || get(networkEvmMainnetEnabled) ? [loadIdbEthAddress()] : []),
	...(get(networkSolanaMainnetEnabled) ? [loadIdbSolAddressMainnet()] : [])
];

const results: NonEmptyArray<ResultSuccess<LoadIdbAddressError>> = await Promise.all(promisesList);
```

src/frontend/src/lib/services/addresses.services.ts
```diff
@@ -32,13 +38,30 @@ export const loadAddresses = async (networkIds: NetworkId[]): Promise<ResultSucc
...
+	if (results.length === 0) {
+		results = [{ success: true, err: undefined }];
+	}
 
-	const { success, err } = reduceResults<LoadIdbAddressError>(results);
+	const { success, err } = reduceResults<LoadIdbAddressError>(
+		results as [ResultSuccess<LoadIdbAddressError>, ...ResultSuccess<LoadIdbAddressError>[]]
+	);
 
 	return { success, err };
```
AntonioVentilii: some tests for this new logic?

daviddecentage: What new logic? I just added a cast

src/frontend/src/lib/services/addresses.services.ts
daviddecentage: updated

src/frontend/src/lib/services/addresses.services.ts
```diff
@@ -32,13 +38,17 @@ export const loadAddresses = async (networkIds: NetworkId[]): Promise<ResultSucc
...
+		...(get(networkBitcoinMainnetEnabled) ? [loadIdbBtcAddressMainnet()] : []),
+		...(get(networkEthereumEnabled) ? [loadIdbEthAddress()] : []),
+		...(get(networkSolanaMainnetEnabled) ? [loadIdbSolAddressMainnet()] : [])
+	];
+
+	let results = await Promise.all(promisesList);
 
-	const { success, err } = reduceResults<LoadIdbAddressError>(results);
+	const { success, err } = reduceResults<LoadIdbAddressError>(
+		results as [ResultSuccess<LoadIdbAddressError>, ...ResultSuccess<LoadIdbAddressError>[]]
```
AntonioVentilii: i am thinking now: this is not necessary true anymore, right? what happens if all the networks are disabled? the list is empty, so how does `reduceResults` reacts?

Before this change, the list was never empty, now it is

src/frontend/src/lib/services/addresses.services.ts
AntonioVentilii: i mean tests for disabled networks: all disabled, only a few, all enabled

src/frontend/src/lib/services/addresses.services.ts
daviddecentage: Then the initial reduce value is returned `{ success: true, err: undefined }`. We could basically directly return and not call reduceResults in that case, but we dont save much as theres no iteration to be had and the initial value is returned directly

AntonioVentilii: then, let's modify `reduceResults` to accept empty lists too (and adapt tests). Otherwise we are casting incorrectly

AntonioVentilii: here:

- https://github.com/dfinity/oisy-wallet/pull/7705

AntonioVentilii: merged! no need to cast anymore

src/frontend/src/lib/services/addresses.services.ts
```diff
@@ -31,14 +31,18 @@ export const loadAddresses = async (networkIds: NetworkId[]): Promise<ResultSucc
...
-		loadIdbEthAddress(),
-		loadIdbSolAddressMainnet()
+		...(networkIds.includes(BTC_MAINNET_NETWORK_ID) ? [loadIdbBtcAddressMainnet()] : []),
+		...(networkIds.includes(ETHEREUM_NETWORK_ID) ? [loadIdbEthAddress()] : []),
+		...(networkIds.includes(SOLANA_MAINNET_NETWORK_ID) ? [loadIdbSolAddressMainnet()] : [])
 	]);
 
-	const { success, err } = reduceResults<LoadIdbAddressError>(results);
+	const { success, err } = reduceResults<LoadIdbAddressError>(
+		results as ResultSuccess<LoadIdbAddressError>[]
```
AntonioVentilii: is it still required to cast here?

src/frontend/src/lib/services/loader.services.ts
```diff
@@ -96,7 +96,14 @@ export const initLoader = async ({
...
 
-	const { success: addressIdbSuccess, err } = await loadIdbAddresses();
+	// We can fetch these values imperatively because these stores were just updated at the beginning of this same function, when loading the user profile.
+	const enabledNetworkIds: NetworkId[] = [
+		...(get(networkBitcoinMainnetEnabled) ? [BTC_MAINNET_NETWORK_ID] : []),
+		...(get(networkEthereumEnabled) || get(networkEvmMainnetEnabled) ? [ETHEREUM_NETWORK_ID] : []),
+		...(get(networkSolanaMainnetEnabled) ? [SOLANA_MAINNET_NETWORK_ID] : [])
+	];
+
+	const { success: addressIdbSuccess, err } = await loadIdbAddresses(enabledNetworkIds);
```
AntonioVentilii, line 106: amazing! this way is smarter

src/frontend/src/lib/services/addresses.services.ts
daviddecentage: no, adjusted it. I need to cast the empty arrays though

src/frontend/src/lib/services/addresses.services.ts
daviddecentage: added tests now

AntonioVentilii (approved):
Thank you! The implementation of the service improved a lot tks to this PR!

PR #7687, opened 2025-07-08
fix(frontend): Format numbers with precision above the maximum

AntonioVentilii:
# Motivation

The util `formatToken` does not properly format big numbers with precision above the 17. That is because transforming a string/bigint to number accepts a precision of 15-17 digits.

For example:

`Number(0.999999999999999876n)` or `(+"0.999999999999999876")` prints `0.9999999999999999`
`Number(0.999999999999999812n)` or `(+"0.999999999999999812")` prints `0.9999999999999998`

Such rounding is causing issues when defining the amounts provided by the user in an input. For example, if the max balance is the first number in the example, and we click the max button, we would try to send more than the balance.

To fix it, we cannot rely on `Number()` nor to `(+...)`. We need to create our own function.

# Changes

- Changed util to have its own flow of transforming a string into the format that the parameters require.

# Tests

Added tests and practical one too where we know that the account has `999999999999999843n ARB`:

### Before

Screenshot 2025-07-08 at 13 56 59

### Before

Screenshot 2025-07-08 at 14 50 46


DenysKarmazynDFINITY (approved):
Some heavy stuff inside 🤯 It's quite hard to comment something on the function body, but since the tests look fine, I think we should be fine with merging it and testing it in the "real" world.

PR #7693, opened 2025-07-08
feat(frontend): custom popover

CvetelinaAngelova:
# Motivation

Creating reusable popover menu component that can render arbitrary trigger elements and a dynamic list of logo-based menu items, while handling positioning, outside clicks, and responsive behavior without pulling in a heavy external library.
Needed basic change for the menu to be shown next to the button trigger eather that it is into a 

# Changes

Dynamic menu items
Accepts an items prop array where each item supplies a logo component, a title, an action callback, and an optional testId.
Renders each item using the shared LogoButton component

Manual positioning logic
On open (via tick()), measures the trigger’s bounding rect and sets CSS custom properties (--popover-top, --popover-left, --popover-right) on the menu for placement.
Listens for window.resize to recompute positioning.
Renders a full-screen transparent <button class="backdrop"> behind the menu to catch outside clicks.
Also attaches a global document.click listener to close the menu when clicking outside both the trigger and the menu.

Accessibility & testing
Adds role="menu" and tabindex="-1" to the menu container.
Applies data-tid attributes for POPOVER_MENU and POPOVER_MENU_ITEM to support end-to-end and unit testing.

# Tests

Unit tests

Toggles menu visibility when trigger is clicked
Renders title and executes item action then closes
Closes menu when clicking outside

Menual tests
Confirm that arbitrary trigger content renders and toggles the menu.
Verify that LogoButton items display their logos and titles correctly.
Screenshot 2025-07-08 at 17 41 37


AntonioVentilii (changes requested):
Why not using the existing `Popover` component from GIX-cmp, or something similar to the `Dropdown` component in OISY? Is it not adaptable to the scope?

CvetelinaAngelova:
> Why not using the existing Popover component from GIX-cmp, or something similar to the Dropdown component in OISY? Is it not adaptable to the scope?

GIX Popover computes the available viewport space—measuring the window’s height and width—before rendering. So it breacks the design and does not attach to the Icon into the modal.

PR #7704, opened 2025-07-09
feat(backend)!: adds erc721 custom token to backend api

BonomoAlessandro:
# Motivation

We want to be able to save custom erc721 tokens in the backend.

BREAKING CHANGE: The backend interfaces changed to include the new type.

# Changes

- adds custom erc721 tokens to backend api

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "@dfinity/oisy-wallet",
-	"version": "1.5.8",
+	"version": "1.5.9",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "@dfinity/oisy-wallet",
-			"version": "1.5.8",
+			"version": "1.5.9",
 			"license": "Apache-2.0",
 			"dependencies": {
 				"@dfinity/agent": "^2.4.0",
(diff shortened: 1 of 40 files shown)


src/shared/src/types/token_standard.rs
```diff
@@ -20,6 +20,13 @@ pub enum TokenStandard {
     ///
     /// - [ERC20 Spec](https://eips.ethereum.org/EIPS/eip-20)
     ERC20,
+    /// Standard for the Ethereum blockchain
+    ///
+    /// # Used for:
+    /// - Most tokens on ETH and other EVM networks.
```
AntonioVentilii: this description should change to consider NFT, no?

src/backend/tests/it/custom_token.rs
```diff
@@ -63,12 +64,24 @@ static ERC20_TOKEN: LazyLock<CustomToken> = LazyLock::new(|| CustomToken {
     enabled: true,
     version: None,
 });
+static ERC721_TOKEN_ID: LazyLock<ErcTokenId> =
+    LazyLock::new(|| ErcTokenId("0x8821bee2ba0df28761afff119d66390d594cd280".to_string()));
+static ERC721_CHAIN_ID: LazyLock<ChainId> = LazyLock::new(|| 8453);
```
AntonioVentilii: just to test something different can you change the Chain ID? for example 137 (polygon mainnet)

AntonioVentilii (approved):
LGTM tks!

PR #7705, opened 2025-07-09
refactor(frontend): Adapt `reduceResults` to accept empty lists

AntonioVentilii:
# Motivation

Util `reduceResults` does not necessarily need to have non-empty lists as inputs. It could just return the default value, that is a successful one.


daviddecentage (approved):
LGTM, thx!

PR #7706, opened 2025-07-09
feat(frontend): improve error handling in calculateUtxoSelection

DecentAgeCoder:
# Motivation

We are now returning specific error types to provide users with clear, actionable feedback when the BTC preperation fails (instead of throwing errors). Furthermore the  business logic in calculateUtxoSelection has been simplified.

# Changes

- Replace `selectUtxosFee` backend API with new `prepareBtcSend` function in `btc-utxos.service.ts`
- Remove deprecated `selectUtxosFee` function from `btc-send.services.ts`
- Enhance UTXO selection algorithm with better validation in `btc-utxos.utils.ts`
- Update `BtcUtxosFee.svelte` to use `prepareBtcSend` instead of deprecated API
- Add `getFeeRateFromPercentiles` function for dynamic fee calculation

# Tests

- Updated `btc-utxos.service.spec.ts` to test the new `prepareBtcSend` function with various UTXO scenarios including insufficient balance cases
- Enhanced `btc-utxos.utils.spec.ts` to test improved UTXO selection algorithm with edge cases and validation scenarios
- Modified `BtcUtxosFee.spec.ts` to test component integration with the new frontend service instead of backend API calls


src/frontend/src/btc/services/btc-utxos.service.ts
```diff
@@ -68,10 +58,11 @@ export const prepareBtcSend = async ({
 	const allUtxos = response.utxos;
 
 	if (allUtxos.length === 0) {
-		throw new Error(
-			// We can't send BTC when no UTXOs are available
-			`InsufficientBalance: No UTXO's available for address ${source} with ${requiredMinConfirmations} confirmations`
-		);
+		return {
+			feeSatoshis: 0n,
```
DenysKarmazynDFINITY: small hint: for cases like this one, we created a variable called `ZERO`, please use it in this file

src/frontend/src/btc/services/btc-utxos.service.ts
```diff
@@ -68,10 +58,11 @@ export const prepareBtcSend = async ({
...
+			utxos: [],
+			error: BtcPrepareSendError.InsufficientBalance
```
DenysKarmazynDFINITY, line 65: Not that I'm against it, but seems like we don't need to have this specific `error` field:
1. `utxos.length <= 0` - means the insufficient balance error.
2. `utxos.length > 0 && feeSatoshis === ZERO` - means the insufficient balance for fee error.

If you think this field helps to simplify some things on the components side, we can keep it. Otherwise, we can remove it.

src/frontend/src/btc/utils/btc-utxos.utils.ts
```diff
@@ -58,7 +60,13 @@ export const calculateUtxoSelection = ({
 	feeRateSatoshisPerVByte: bigint;
 }): UtxoSelectionResult => {
 	if (availableUtxos.length === 0) {
-		throw new Error('No UTXOs available for selection');
+		return {
+			selectedUtxos: [],
+			totalInputValue: 0n,
```
DenysKarmazynDFINITY: all those values can also be `ZERO`

src/frontend/src/btc/services/btc-utxos.service.ts
DecentAgeCoder: For example, we may need to distinguish between multiple error cases such as `InsufficientBalance` and `InsufficientBalanceForFe`, or potentially others in the future. If we used ZERO for `feeSatoshis`, we’d be limited to signaling only a single, generic error case.

The current approach also results in cleaner error handling, as the error code can be used directly as a translation key. For these reasons, I’d strongly prefer to keep the current implementation as is

DenysKarmazynDFINITY: Ok, totally fine with keeping it

src/frontend/src/btc/utils/btc-utxos.utils.ts
DecentAgeCoder: I have replaced all 0n with ZERO (efen in tests)

src/frontend/src/btc/services/btc-utxos.service.ts
DecentAgeCoder: Done

PR #7709, opened 2025-07-09
feat(frontend): Close modal on failed Solana confirmation

AntonioVentilii:
# Motivation

In the Solana send flow, if the confirmation fails but sending is successful, we don't necessarily know if the issue was with the transaction or with other factors (for example connection lost with the listener).

In case of failure in this case, we prefer to close the modal, since the current behaviour is to go to previous step in the wizard modal (that could cause the wrong impression that the user could re-do the transaction).

# Changes

- Change Solana send wizard modal to exit the modal in case of error at step `CONFIRM`.

# Tests

https://github.com/user-attachments/assets/bdf1f821-1bac-4dcc-89d1-168944c03609


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7710, opened 2025-07-09
feat(frontend): Derive SOL address in the frontend

AntonioVentilii:
# Motivation

We will derive the SOL address in the frontend, using the library [ic-pub-key](https://github.com/dfinity/ic-pub-key) (that for now it is just copied code, see PR #7464).

However, since we are aware of the public key ONLY for the `production Chain Fusion Signer`, we can make it work ONLY for the environments that use such canister, that means `Beta` and `Prod`.

For the other environments, that use `staging Chain Fusion Signer`, we will have to find a better solution, once we discover the correct root key and the correct public key.

# Changes

- When the feature flag is turned on, use the function form the library to derive the address.

# Tests

Manually tested deploying this branch in Beta, and the SOL address was the same as Prod:

### DIFF
--- a/src/frontend/src/lib/ic-pub-key/src/cli.ts
+++ b/src/frontend/src/lib/ic-pub-key/src/cli.ts
@@ -2,6 +2,10 @@
 /* istanbul ignore file */
 /* v8 ignore start */
 
+import { schnorr_ed25519_derive } from '$lib/ic-pub-key/src/schnorr/ed25519';
+import { mapDerivationPath } from '$lib/utils/signer.utils';
+import { SOLANA_DERIVATION_PATH_PREFIX } from '$sol/constants/sol.constants';
+import type { SolanaNetworkType } from '$sol/types/network';
 import type { BitcoinNetwork } from '@dfinity/ckbtc';
 import { Principal } from '@dfinity/principal';
 import { networks, payments, type Network } from 'bitcoinjs-lib';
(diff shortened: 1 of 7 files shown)


src/frontend/src/lib/ic-pub-key/src/cli.ts
```diff
@@ -68,4 +70,25 @@ export const deriveBtcAddress = async (user: string, network: BitcoinNetwork): P
...
 
+/* istanbul ignore next */
+export const deriveSolAddress = async (user: string, derivationPath: string[]): Promise<string> => {
+	const pubkey = '476374d9df3a8af28d3164dc2422cff894482eadd1295290b6d9ad92b2eeaa5c';
+	const chaincode = '0000000000000000000000000000000000000000000000000000000000000000';
+
+	const principal = Principal.fromText(user);
+
+	let derivationPathObj = new DerivationPath([
+		Uint8Array.from([0, 0, 0, 0, 2, 48, 0, 113, 1, 1]),
```
bitdivine, line 81: This is the chain fusion signer canister ID.  If you have that as a variable, which you probably do as you have to make API calls to the signer, this can be made to work on staging and prod alike.  On staging we also use the test key, the moment, so in my opinion this is not worth doing in this PR.

bitdivine (approved):

src/frontend/src/lib/ic-pub-key/src/cli.ts
AntonioVentilii: I am enabling the frontend-derivation ONLY for PROD and BETA for now, that is why I would prefer to treat it as an entire feature, and raise some "mistake" incase we talk about STAGING (or other envs)

PR #7718, opened 2025-07-10
style(frontent): Fix BTC send review warnings

daviddecentage:
# Motivation

The warnings for example "Insufficient balance to cover fees" are not correctly styled.

# Changes

- Used MessageBox component instead of warning banner
- Adjusted width so it takes full space horizontally

# Tests

image


AntonioVentilii (approved):
Fantastic!

P.S. After this PR, is `WarningBanner` still used?

PR #7721, opened 2025-07-10
fix(frontend): display fee when fee amount is insufficient

DecentAgeCoder:
# Motivation

When the error "Insufficient balance to cover the fees" occurs, the fees are not currently displayed. This fix ensures that the fee details are also shown when the error code `InsufficientBalanceForFee` is returned.

# Changes

- Updated the service to return calculated fees even in insufficient balance scenarios by using `selection.feeSatoshis` instead of hardcoded `ZERO`

# Tests

- Fixed test case "should return insufficient funds when not enough UTXOs"


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7723, opened 2025-07-10
test(frontend): Use `TokenId` type in mock for IC transactions

AntonioVentilii:
# Motivation

For consistency, we can use directly the `TokenId` type in the mocks.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7725, opened 2025-07-10
feat(frontend): Services to clear IDB transactions stores

AntonioVentilii:
# Motivation

When the user logs out, we need to clean the IDB stores that cached the transactions.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7732, opened 2025-07-10
style(frontend): Use Responsive component for mobile navigation

AntonioVentilii:
# Motivation

To improve performance, we should avoid rendering the mobile navigation in DOM, if not needed. Until now we were rendering it, but hiding it too. However, with Responsive component, we can use the Svelte `{#if}` statement and avoid rendering it, until necessary.

Note that we combine both. This ensures:

- It's only in the DOM when it should be
- And it behaves responsively too

### Results

No changes in behaviour:

https://github.com/user-attachments/assets/fb049972-2245-4905-8a52-de8316140a0f


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7742, opened 2025-07-11
feat(frontend): loads custom erc721 tokens

BonomoAlessandro:
# Motivation

We want to be able to load custom `erc721` tokens from the backend service.

# Changes

- loads custom `erc721` tokens


AntonioVentilii (approved):
@BonomoAlessandro good point, it is already a lot generalized, it just seemed as repeated code at first sight... for sure we can make it even more modular, but for now all good, tks!

PR #7743, opened 2025-07-11
feat(frontend): adds erc721 tokens to all tokens

BonomoAlessandro:
# Motivation

# Changes

# Tests


src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
```diff
@@ -9,7 +10,9 @@
 	export let tokens: TokenUiOrGroupUi[] | undefined = undefined;
 
 	let groupedTokens: TokenUiOrGroupUi[];
-	$: groupedTokens = groupTokensByTwin($combinedDerivedSortedNetworkTokensUi);
+	$: groupedTokens = groupTokensByTwin(
+		$combinedDerivedSortedNetworkTokensUi.filter((token) => !isTokenErc721(token))
+	);
```
AntonioVentilii: Not really sure we should do this: it means that we have to remove it custom everywhere they should not be treated like other tokens. It become quite unmaneagable

BonomoAlessandro: @AntonioVentilii I see your point. Do you think it would be better to create a new `derived` function that only provides the tokens that we want to display on that token list?

src/frontend/src/eth/derived/erc721.derived.ts
```diff
@@ -0,0 +1,33 @@
...
+import { erc721CustomTokensStore } from '$eth/stores/erc721-custom-tokens.store';
+import type { Erc721CustomToken } from '$eth/types/erc721-custom-token';
+import type { Erc721TokenToggleable } from '$eth/types/erc721-token-toggleable';
+import { enabledEvmNetworksIds } from '$evm/derived/networks.derived';
+import { derived, type Readable } from 'svelte/store';
+
+/**
+ * The list of ERC721 tokens the user has added, enabled or disabled.
+ */
+export const erc721CustomTokens: Readable<Erc721CustomToken[]> = derived(
```
AntonioVentilii, line 11: some tests for this one?

BonomoAlessandro: @AntonioVentilii i just provided some test cases

AntonioVentilii (approved):
Thank you!

PR #7748, opened 2025-07-11
feat(frontend): expose LLM canister methods

DenysKarmazynDFINITY:
# Motivation

We can now expose LLM canister methods, particularly `v1_chat` method.


src/frontend/src/lib/api/llm.api.ts
```diff
@@ -0,0 +1,36 @@
+import type { chat_request_v1, chat_response_v1 } from '$declarations/llm/llm.did';
+import { LlmCanister } from '$lib/canisters/llm.canister';
+import { LLM_CANISTER_ID } from '$lib/constants/app.constants';
+import type { CanisterApiFunctionParams } from '$lib/types/canister';
+import { Principal } from '@dfinity/principal';
+import { assertNonNullish, isNullish } from '@dfinity/utils';
+
+let canister: LlmCanister | undefined = undefined;
```
AntonioVentilii, line 8: not sure, but I think we need to create some tests (mostly mocks) for coverage even here for the `.api` module

AntonioVentilii (approved):
LGTM tks

PR #7769, opened 2025-07-14
feat(frontend): Change ETH transactions store to have certified transactions

AntonioVentilii:
# Motivation

We are trying to simplify the code and mostly it means to standardize the stores when it concerns different networks.

Similar to what we did for the tokens store, we want to normalize the Ethereum/EVM transactions store with the ones used for the other networks (IC, Solana and Bitcoin).

As first step, we need to provide the transactions as "certified" transactions. Obviously, we do not certify them, but we can set them in a structure that define them as certified.

# Note

We set `certified` as `false`, because it is not technically certified in the sense that IC transactions are certified for example.
We do the same thing for Bitcoin and Solana transactions.

# Changes

- Change current type of Ethereum/EVM transactions store to reflect the same os the ones for the other stores (it is basically a copy paste).
- Adapt the usage throughout the code, bot for write and read methods.

# Tests

Tests work once adapted to the read/write changes.


DenysKarmazynDFINITY (approved):
Makes sense!

LGTM, thanks

PR #7772, opened 2025-07-14
feat(frontend): implement ai-assistant.derived

DenysKarmazynDFINITY:
# Motivation

The next step is to implement ai-assistant.derived.

### DIFF
--- a/src/frontend/src/lib/derived/ai-assistant.derived.ts
+++ b/src/frontend/src/lib/derived/ai-assistant.derived.ts
@@ -2,6 +2,7 @@ import type { chat_message_v1 } from '$declarations/llm/llm.did';
 import { AI_ASSISTANT_SYSTEM_PROMPT } from '$lib/constants/ai-assistant.constants';
 import { aiAssistantStore } from '$lib/stores/ai-assistant.store';
 import type { ChatMessage } from '$lib/types/ai-assistant';
+import { toNullable } from '@dfinity/utils';
 import { derived, type Readable } from 'svelte/store';
 
 export const aiAssistantConsoleOpen: Readable<boolean> = derived(
@@ -41,8 +42,8 @@ export const aiAssistantLlmMessages: Readable<chat_message_v1[]> = derived(
 					...acc,
 					{
 						assistant: {
-							content: [content],
-							tool_calls: []
+							content: toNullable(content),
+							tool_calls: toNullable()
 						}
 					}
 				];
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/derived/ai-assistant.derived.ts
```diff
@@ -0,0 +1,75 @@
...
+					}
+				];
+			}
+			if (role === 'assistant') {
+				return [
+					...acc,
+					{
+						assistant: {
+							content: [content],
+							tool_calls: []
```
AntonioVentilii: `toNullable` for both?
```suggestion
							content: toNullable(content),
							tool_calls: toNullable()
```

src/frontend/src/lib/derived/ai-assistant.derived.ts
```diff
@@ -0,0 +1,75 @@
...
+	[aiAssistantStore],
+	([$aiAssistantStore]) => {
+		let includesSystemMessage = false;
+
+		// Get last 10 messages from chat history
+		const recentHistory = ($aiAssistantStore?.chatHistory ?? []).slice(-10);
+
+		// Parse chat messages into LLM-compatible messages
+		const messages = recentHistory.reduce<chat_message_v1[]>((acc, { role, content }) => {
+			if (role === 'system') {
```
AntonioVentilii, line 28: just a doubt: it is guaranteed that the system message is first? if not, MUST it go first? or it doesn't matter at all?

src/frontend/src/lib/stores/ai-assistant.store.ts
```diff
@@ -35,6 +38,10 @@ const initAiAssistantStore = (): AiAssistantStore => {
 			update((state) => ({ ...state, isOpen: false }));
 		},
 
+		reset: () => {
+			update(() => initialState);
```
AntonioVentilii: nitpick: you can use directly `set(initialState)` from:

`const { subscribe, update, set } = writable<AiAssistant>(initialState);`

src/frontend/src/lib/derived/ai-assistant.derived.ts
DenysKarmazynDFINITY: done!

src/frontend/src/lib/derived/ai-assistant.derived.ts
DenysKarmazynDFINITY: The order doesn't really matter, we just need to keep the system message in the array.

src/frontend/src/lib/stores/ai-assistant.store.ts
DenysKarmazynDFINITY: good point, updated

AntonioVentilii (approved):
LGTM tkS

PR #7773, opened 2025-07-14
feat(frontend): improves Loader test cases

BonomoAlessandro:
# Motivation

Some mocks of the `Loader` test cases were not correctly implemented. We fix it with this PR.

# Changes

- fixes test cases


BonomoAlessandro:
@AntonioVentilii yes the test cases also work without the mocks but the load functions will be called. With the correct mocks, the load functions are not called which improves the speed of the tests.

AntonioVentilii (approved):
Thank you!

PR #7775, opened 2025-07-14
fix(frontend): Remove reference to workers objects

AntonioVentilii:
# Motivation

As per @DecentAgeCoder , we are going to completely delete the reference to the worker objects in our worker services.

If we frequently create and destroy Web Workers this could leave behind a residual memory footprint, possibly outside the JS heap. This could be due to delayed garbage collection, where the browser holds on to memory for a few seconds (e.g., around 10 seconds) after a worker is terminated.

As long as that reference exists, the garbage collector may delay reclaiming the memory — even if the worker’s thread has been shut down. This could potentially lead to memory staying around longer than needed.

# Disclaimer

This is mostly the assumption done observing the current behaviour.


DecentAgeCoder (commented):
LGTM! Thanks!

PR #7776, opened 2025-07-14
feat(frontend): implements isErc721 infura method

BonomoAlessandro:
# Motivation

We want to be able to find out if a contract address belongs to an `erc721` contract or not. 

# Changes

- implements `isErc721` function

### DIFF
--- a/src/frontend/src/tests/eth/providers/infura-erc721.providers.spec.ts
+++ b/src/frontend/src/tests/eth/providers/infura-erc721.providers.spec.ts
@@ -132,6 +132,17 @@ describe('infura-erc721.providers', () => {
 				expect(result).toBeTruthy();
 			});
 
+			it('should return false on error', async () => {
+				const errorMessage = 'Error fetching metadata';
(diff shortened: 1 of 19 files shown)


src/frontend/src/tests/eth/providers/infura-erc721.providers.spec.ts
```diff
@@ -108,6 +108,49 @@ describe('infura-erc721.providers', () => {
...
+				expect(mockContract).toHaveBeenCalledOnce();
+
+				expect(mockContract).toHaveBeenNthCalledWith(
+					1,
+					...expectedContractParams,
+					new mockProvider()
+				);
+
+				expect(mockSupportsInterface).toHaveBeenCalledOnce();
+			});
```
AntonioVentilii, line 162: just a test in case of error (it should return false)?

BonomoAlessandro: You are right! This would make sense.

PR #7783, opened 2025-07-15
chore(frontend): Fix Portuguese dapp category names

StefanBerger-DFINITY:
# Motivation

Tags in the Dapp Explorer are too long in Portuguese.

image

# Changes

- Shorten them. Also don't translate idioms.

Sources:
-  https://pt.wikipedia.org/wiki/M%C3%ADdias_sociais
- image

# Tests

image


AntonioVentilii (approved):
Thanks for the fix!

PR #7785, opened 2025-07-15
refactor(frontend): Rename util to format currency

AntonioVentilii:
# Motivation

We are going to use more currencies than USD, so, we first rename the util to format the currency amounts.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7797, opened 2025-07-15
fix(frontend): worker memory leak fixes

DecentAgeCoder:
# Motivation
Workers are not being properly terminated, causing memory leaks. The main issues are:
1. Exchange worker isn't properly terminated on destroy
2. Worker cleanup in `cleanWorkers` only called stop but not `destroy()`

If I understand correctly the `ExchangeWorker.svelte` gets "terminated" when `ExchangeWorker.svelte` gets unmounted or the user navigates to to a page that doesn't include the component `ExchangeWorker.svelte`

**The missing exchange worker termination could be the primary source of the memory leak**

# Changes

**Fixed Exchange Worker (`worker.exchange.services.ts`)**
- Changed worker type:  From `const exchangeWorker: Worker` to `let exchangeWorker: Worker | null` 
- Added optional chaining: Used `?.` for all `postMessage` calls to prevent errors when worker is null
- Enhanced destroy method: Added proper worker termination:

**Improved Worker Cleanup Utility (`wallet.utils.ts`)**

### DIFF
--- a/src/frontend/src/lib/services/worker.exchange.services.ts
+++ b/src/frontend/src/lib/services/worker.exchange.services.ts
@@ -18,7 +18,7 @@ let errorMessages: { msg: string; timestamp: number }[] = [];
 
 export const initExchangeWorker = async (): Promise<ExchangeWorker> => {
 	const ExchangeWorker = await import('$lib/workers/workers?worker');
-	const exchangeWorker: Worker = new ExchangeWorker.default();
+	let exchangeWorker: Worker | null = new ExchangeWorker.default();
 
 	exchangeWorker.onmessage = ({
 		data: dataMsg
@@ -87,13 +87,13 @@ export const initExchangeWorker = async (): Promise<ExchangeWorker> => {
 	};
 
 	const stopTimer = () =>
-		exchangeWorker.postMessage({
+		exchangeWorker?.postMessage({
 			msg: 'stopExchangeTimer'
 		});
 
 	return {
 		startExchangeTimer: (data: PostMessageDataRequestExchangeTimer) => {
-			exchangeWorker.postMessage({
+			exchangeWorker?.postMessage({
 				msg: 'startExchangeTimer',
 				data
 			});
@@ -101,6 +101,10 @@ export const initExchangeWorker = async (): Promise<ExchangeWorker> => {
 		stopExchangeTimer: stopTimer,
 		destroy: () => {
 			stopTimer();
+			if (exchangeWorker) {
+				exchangeWorker.terminate();
+				exchangeWorker = null;
+			}
 			errorMessages = [];
 		}
 	};
--- a/src/frontend/src/lib/utils/wallet.utils.ts
+++ b/src/frontend/src/lib/utils/wallet.utils.ts
@@ -18,6 +18,26 @@ export const waitAndTriggerWallet = async () => {
 };
 
 /**
+ * Properly destroy a worker and remove it from the workers map.
+ *
+ * @param workers The map of workers defined as `Map<TokenId, WalletWorker>`.
+ * @param tokenId The token ID of the worker to destroy.
+ */
+export const destroyWorker = ({
+	workers,
+	tokenId
+}: {
+	workers: Map<TokenId, WalletWorker>;
+	tokenId: TokenId;
+}) => {
+	const worker = workers.get(tokenId);
+	if (worker) {
+		worker.destroy();
+		workers.delete(tokenId);
+	}
+};
+
+/**
  * Clean the workers that are not used anymore based on a given list of tokens.
  *
  * @param workers The map of workers defined as `Map<TokenId, WalletWorker>`.
@@ -33,9 +53,10 @@ export const cleanWorkers = ({
 	Array.from(workers.keys())
 		.filter((workerTokenId) => !tokens.some(({ id: tokenId }) => workerTokenId === tokenId))
 		.forEach((tokenId) => {
-			// TODO: use a more functional approach instead of deleting the worker from the map.
+			// Trigger the stop event to properly cleanup the worker thread
 			workers.get(tokenId)?.stop();
-			workers.delete(tokenId);
+			// terminate and clean up worker
+			destroyWorker({ workers, tokenId });
 		});
 
 /**


src/frontend/src/lib/services/worker.exchange.services.ts
```diff
@@ -87,20 +87,24 @@ export const initExchangeWorker = async (): Promise<ExchangeWorker> => {
...
 			});
 		},
 		stopExchangeTimer: stopTimer,
 		destroy: () => {
 			stopTimer();
+			if (exchangeWorker) {
+				exchangeWorker.terminate();
+				exchangeWorker = null;
+			}
 			errorMessages = [];
```
AntonioVentilii: good catch!!!! I would just suggest to keep the same implementations of the other `destroy` methods, for consistency. If you want to change them (for example adding checks), we can always do it in another PR

```suggestion
			if (isDestroying) {
				return;
			}
			isDestroying = true;
			stopTimer();
			exchangeWorker?.terminate();
			exchangeWorker = null;
			isDestroying = false;
errorMessages = [];
```

src/frontend/src/lib/utils/wallet.utils.ts
```diff
@@ -33,9 +53,10 @@ export const cleanWorkers = ({
 	Array.from(workers.keys())
 		.filter((workerTokenId) => !tokens.some(({ id: tokenId }) => workerTokenId === tokenId))
 		.forEach((tokenId) => {
-			// TODO: use a more functional approach instead of deleting the worker from the map.
+			// Trigger the stop event to properly cleanup the worker thread
```
AntonioVentilii, line 58: let's keep the TODO, it's still not solved

```suggestion
			// Trigger the stop event to properly cleanup the worker thread
			// TODO: use a more functional approach instead of deleting the worker from the map.
```

src/frontend/src/lib/utils/wallet.utils.ts
```diff
@@ -17,6 +17,26 @@ export const waitAndTriggerWallet = async () => {
...
+ */
+export const destroyWorker = ({
+	workers,
+	tokenId
+}: {
+	workers: Map<TokenId, WalletWorker>;
+	tokenId: TokenId;
+}) => {
+	const worker = workers.get(tokenId);
+	if (worker) {
```
AntonioVentilii:
```suggestion
	if (nonNullish(worker)) {
	worker.stop()
```

src/frontend/src/lib/utils/wallet.utils.ts
```diff
@@ -17,6 +17,26 @@ export const waitAndTriggerWallet = async () => {
 	waitForMilliseconds(INDEX_RELOAD_DELAY).then(() => emit({ message: 'oisyTriggerWallet' }));
 };
 
+/**
+ * Properly destroy a worker and remove it from the workers map.
+ *
+ * @param workers The map of workers defined as `Map<TokenId, WalletWorker>`.
+ * @param tokenId The token ID of the worker to destroy.
+ */
+export const destroyWorker = ({
```
AntonioVentilii:
```suggestion
const destroyWorker = ({
```

src/frontend/src/lib/utils/wallet.utils.ts
```diff
@@ -17,6 +17,26 @@ export const waitAndTriggerWallet = async () => {
...
+	workers,
+	tokenId
+}: {
+	workers: Map<TokenId, WalletWorker>;
+	tokenId: TokenId;
+}) => {
+	const worker = workers.get(tokenId);
+	if (worker) {
+		worker.destroy();
+		workers.delete(tokenId);
```
AntonioVentilii, line 38: guaranteed that this modify the original map, right? maybe some tests?

src/frontend/src/lib/utils/wallet.utils.ts
```diff
@@ -33,9 +53,10 @@ export const cleanWorkers = ({
 	Array.from(workers.keys())
 		.filter((workerTokenId) => !tokens.some(({ id: tokenId }) => workerTokenId === tokenId))
 		.forEach((tokenId) => {
-			// TODO: use a more functional approach instead of deleting the worker from the map.
+			// Trigger the stop event to properly cleanup the worker thread
 			workers.get(tokenId)?.stop();
```
AntonioVentilii: better to put it in the util 

```suggestion
```

src/frontend/src/lib/utils/wallet.utils.ts
```diff
@@ -33,9 +53,10 @@ export const cleanWorkers = ({
...
-			workers.delete(tokenId);
+			// terminate and clean up worker
```
AntonioVentilii:
```suggestion
```

AntonioVentilii (commented):
Good catch! A few comments, but let's anyway split the PR into the two fixes and add some tests

src/frontend/src/lib/services/worker.exchange.services.ts
```diff
@@ -87,21 +87,33 @@ export const initExchangeWorker = async (): Promise<ExchangeWorker> => {
...
 	const stopTimer = () =>
-		exchangeWorker.postMessage({
+		exchangeWorker?.postMessage({
 			msg: 'stopExchangeTimer'
 		});
 
 	return {
 		startExchangeTimer: (data: PostMessageDataRequestExchangeTimer) => {
-			exchangeWorker.postMessage({
+			exchangeWorker?.postMessage({
```
AntonioVentilii, line 96:
```suggestion
});


let isDestroying = false

return {
startExchangeTimer: (data: PostMessageDataRequestExchangeTimer) => {
			exchangeWorker?.postMessage({
```

src/frontend/src/lib/services/worker.exchange.services.ts
```diff
@@ -87,20 +87,32 @@ export const initExchangeWorker = async (): Promise<ExchangeWorker> => {
...
 				msg: 'startExchangeTimer',
 				data
```
AntonioVentilii, line 98:
```suggestion
};

const stopTimer = () =>
		exchangeWorker?.postMessage({
msg: 'stopExchangeTimer'
});


let isDestroying = false

return {
startExchangeTimer: (data: PostMessageDataRequestExchangeTimer) => {
			exchangeWorker?.postMessage({
msg: 'startExchangeTimer',
data
```

src/frontend/src/lib/utils/wallet.utils.ts
DecentAgeCoder: Added test for worker cleanup to verify that stop and destroy is called and the worker is removed from the workers map in wallet.utils.spec.ts

AntonioVentilii (approved):
LGTM tks!

Please update the description before merging

PR #7798, opened 2025-07-15
fix(frontend): Completely terminate Exchange workers on destroy

AntonioVentilii:
# Credits

This PR is just a split of @DecentAgeCoder 's PR https://github.com/dfinity/oisy-wallet/pull/7797 .

> # Motivation
> 
> Workers are not being properly terminated, causing memory leaks. In this PR we fix the issue of the Exchange worker that isn't properly terminated on destroy.
> 
> `ExchangeWorker.svelte` gets "terminated" when `ExchangeWorker.svelte` gets unmounted or the user navigates to to a page that doesn't include the component `ExchangeWorker.svelte`
> Wallet worker lifecycle is bound to tokens array
> The missing worker termination for exchange or wallet workers could be the primary source of the memory leak
> 
> # Changes
> 
> - Changed worker type: From `const exchangeWorker: Worker` to `let exchangeWorker: Worker | null`.
> - Added optional chaining: Used `?.` for all `postMessage` calls to prevent errors when worker is `null`.
> - Enhanced `destroy` method: Added proper worker termination.

# Tests

Since we are here, we included tests.

### DIFF
--- a/src/frontend/src/lib/services/worker.exchange.services.ts
+++ b/src/frontend/src/lib/services/worker.exchange.services.ts
@@ -18,7 +18,7 @@ let errorMessages: { msg: string; timestamp: number }[] = [];
 
 export const initExchangeWorker = async (): Promise<ExchangeWorker> => {
 	const ExchangeWorker = await import('$lib/workers/workers?worker');
-	const exchangeWorker: Worker = new ExchangeWorker.default();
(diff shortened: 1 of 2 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7801, opened 2025-07-15
feat(frontend): Util to get currency symbol

AntonioVentilii:
# Motivation

To allow more currencies in the UI, we require to know the symbol of such currencies in the user's selected language.

So, we create an util for that, using ISO 4217 codes and the existing languages.

# Changes

- Create util to get the currency symbol by symbol and language, by formatting zero to a currency.

# Tests

Added tests.


src/frontend/src/lib/utils/currency.utils.ts
```diff
@@ -12,3 +12,21 @@ export const getCurrencyName = ({
...
 };
+
+export const getCurrencySymbol = ({
+	currency,
+	language
+}: {
+	currency: Currencies;
+	language: Languages;
+}): string | undefined =>
+	(0)
```
DenysKarmazynDFINITY: I think it can also be achieved by using something like this:
```
new Intl.NumberFormat('de', { style: 'currency', currency: 'USD' })
  .formatToParts(0)
  .find(p => p.type === 'currency').value;
```
Also a tricky one, but maybe a bit more readable. If the tests fail with the implementation or you prefer your approach - no hard opinion here.

AntonioVentilii: Amazing! Changed, thank you!

PR #7802, opened 2025-07-15
feat(frontend): Service to calculate FX rates USDXXX

AntonioVentilii:
# Motivation

We want to display the FIAT amounts in more currencies than USD. So, as _escamotage_, we leave all amounts in USD and multiply by the USDXXX FX rate of the user's selected currency.

This should work well, since we do not expect to have multiple FIAT currencies at the same time. So, we can use the unique FX rate for all the amounts.

However, to use existing services, we will derive this rate by cross-referencing the prices of a very liquid asset (BTC in our case) against the two currencies (USD and XXX). Then we obtain the USDXXX rate by BTCUSD / BTCXXX.

# Changes

- Adapt Coingecko `simplePrice` service to accept multiple currencies as request (see [documentation](https://docs.coingecko.com/reference/simple-price)) by having them as comma-separated list.
- Create service to fetch the prices of USD and a generic currency XXX.
- Cross those prices to obtain the final FX rate.

# Tests

Added unitests.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -590,7 +590,7 @@ if (
 export const exchangeRateBNBToUsd = (): Promise<CoingeckoSimplePriceResponse | null> =>
 	simplePrice({
 		ids: 'binancecoin',
-		vs_currencies: 'usd'
+		vs_currencies: Currencies.USD
 	});
 ```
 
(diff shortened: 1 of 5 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7803, opened 2025-07-15
refactor(frontend): Use `Currencies` enum in all Coingecko references

AntonioVentilii:
# Motivation

The Coingecko requests will have to be adapted to the list of supported currencies of OISY. As first step we define the correct types.

### DIFF
--- a/src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
+++ b/src/frontend/src/btc/components/transactions/BtcTransactionModal.svelte
@@ -11,6 +11,7 @@
 	import TransactionContactCard from '$lib/components/transactions/TransactionContactCard.svelte';
 	import ButtonCloseModal from '$lib/components/ui/ButtonCloseModal.svelte';
 	import ContentWithToolbar from '$lib/components/ui/ContentWithToolbar.svelte';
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/types/coingecko.ts
```diff
@@ -21,9 +22,9 @@ export type CoingeckoPlatformId =
 	| 'polygon-pos'
 	| 'arbitrum-one';
 
-// We only support conversion in USD for now, therefore not an exhaustive list.
+// Please, cross-reference the OISY supported currencies with the Coingecko API for supported currencies.
 // *refers to curl -l https://api.coingecko.com/api/v3/simple/supported_vs_currencies
-export type CoingeckoCurrency = 'usd';
+export type CoingeckoCurrency = `${Currencies}`;
```
DenysKarmazynDFINITY, line 27: `Currencies.USD`?

AntonioVentilii: in this PR we already expand for all of them, but since they are `enum`, i need that notation

DenysKarmazynDFINITY: got it

PR #7811, opened 2025-07-16
feat(frontend): implements tokenInventory function for etherscan provider

BonomoAlessandro:
# Motivation

We want to be able to find out with `tokenIds` a specific user holds of a specific erc721 contract.

# Changes

- implements `tokenInventory` function


src/frontend/src/eth/providers/etherscan.providers.ts
```diff
@@ -181,6 +181,27 @@ export class EtherscanProvider {
...
+		const params = {
+			action: 'addresstokennftinventory',
+			address,
+			contractaddress: contractAddress,
+			startblock: 0,
+			endblock: 99999999,
+			sort: 'desc'
+		};
+
+		const result = await this.provider.fetch('account', params);
```
AntonioVentilii: typically we try to "document" the expected results. For example, look at 

```ts
const result: EtherscanProviderTokenTransferTransaction[] | string = await this.provider.fetch(
			'account',
			params
		);
```

that allows us to avoid possible future issues with types when bumping libraries (we had some of them when bumping `ethers`)

AntonioVentilii (commented):
Just for more clarity, may you please create a `type NftId = number` in another PR? So that we know when we are talking about that.

Even better if you brand it with `zod` (see what is done with `TokenId` for example)

`export const NftIdSchema = z.number().brand<'NftId'>();`

or any name you want eheh

src/frontend/src/eth/providers/etherscan.providers.ts
BonomoAlessandro: Good catch!

BonomoAlessandro:
@AntonioVentilii Yes, i will create a separate PR based on this one to add a type for the `NftId`.

PR #7812, opened 2025-07-16
feat(frontend): implements getNftMetadata function

BonomoAlessandro:
# Motivation

We want to have an option to load the nft metadata for a specific nft.

# Changes

- implements `getNftMetadata`


src/frontend/src/eth/providers/infura-erc721.providers.ts
```diff
@@ -43,6 +44,48 @@ export class InfuraErc721Provider {
...
+		contractAddress,
+		tokenId
+	}: {
+		contractAddress: string;
+		tokenId: number;
+	}): Promise<NftMetadata> => {
+		const erc721Contract = new Contract(contractAddress, ERC721_ABI, this.provider);
+
+		const resolveResourceUrl = (url: string): string => {
+			const IPFS_PREFIX = 'ipfs://';
```
AntonioVentilii, line 60: just as info, this will not work in any env (apart from local) until we open the CSP to have image-src from all source (or at least `ipfs://`)

I think better to ask @StefanBerger-DFINITY if prodsec is ok with opening it. Related PR:

- https://github.com/dfinity/oisy-wallet/pull/7160

StefanBerger-DFINITY: Hi @AntonioVentilii 
We discussed this with Prodsec. Sending the link in a DM.

We have a ticket to open the CSP for specific image types (and metadata), and add logic to further block fungible tokens from loading images, so it behaves as now. (oisy-1834)
For NFTs, this would for now work, but it's not yet live. We then add a consent process per collection, where the user must approve to download the image, knowing the risk (exposure, phishing, legal).
And only when this is done, we can enable NFTs.

AntonioVentilii: Ah perfect! Then, before this PR, we should open the CSP, @BonomoAlessandro , with the discussed measures

AntonioVentilii: As discussed offline, we will work in parallel on opening the CSP

src/frontend/src/eth/providers/infura-erc721.providers.ts
```diff
@@ -43,6 +44,48 @@ export class InfuraErc721Provider {
...
+	getNftMetadata = async ({
+		contractAddress,
+		tokenId
+	}: {
+		contractAddress: string;
+		tokenId: number;
+	}): Promise<NftMetadata> => {
+		const erc721Contract = new Contract(contractAddress, ERC721_ABI, this.provider);
+
+		const resolveResourceUrl = (url: string): string => {
```
AntonioVentilii: maybe the type should be `Url`? or parsed as Url at any point? we have `UrlSchema`

BonomoAlessandro: @AntonioVentilii you are right. To use `URL` would be better. I just updated the code.

src/frontend/src/eth/providers/infura-erc721.providers.ts
```diff
@@ -43,6 +44,48 @@ export class InfuraErc721Provider {
...
+		const resolveResourceUrl = (url: URL): URL => {
+			const IPFS_PREFIX = 'ipfs://';
+			if (url.href.startsWith(IPFS_PREFIX)) {
+				return new URL(url.href.replace(IPFS_PREFIX, 'https://ipfs.io/ipfs/'));
+			}
+
+			return url;
+		};
+
+		const tokenUri = new URL(await erc721Contract.tokenURI(tokenId));
```
AntonioVentilii: nitpick: you can use the safeParse to guarantee the correct type:

```ts
safeParse({
	schema: UrlSchema,
	value: uri
});
```

src/frontend/src/eth/providers/infura-erc721.providers.ts
```diff
@@ -43,6 +44,48 @@ export class InfuraErc721Provider {
...
+	}: {
+		contractAddress: string;
+		tokenId: number;
+	}): Promise<NftMetadata> => {
+		const erc721Contract = new Contract(contractAddress, ERC721_ABI, this.provider);
+
+		const resolveResourceUrl = (url: URL): URL => {
+			const IPFS_PREFIX = 'ipfs://';
+			if (url.href.startsWith(IPFS_PREFIX)) {
+				return new URL(url.href.replace(IPFS_PREFIX, 'https://ipfs.io/ipfs/'));
```
AntonioVentilii, line 62: same

PR #7813, opened 2025-07-16
style(frontend): Correct colors for warn toast

AntonioVentilii:
# Motivation

After PR #7674, we are defining the colors of a `Toast` message according to their level.

In this PR we fix the colors of the `warn` level, to be the same as `error` level (as it was before the aforementioned PR).

Thanks @DenysKarmazynDFINITY for spotting it!

### Before

Screenshot 2025-07-16 at 09 34 56

### After

Screenshot 2025-07-16 at 09 31 57


DenysKarmazynDFINITY (approved):
Thanks!

PR #7818, opened 2025-07-16
refactor(frontend): Get language directly as params in util `formatSecondsToDate`

AntonioVentilii:
# Motivation

To simplify the code, we pass directly the language to the util `formatSecondsToDate`.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7820, opened 2025-07-16
feat(frontend): implements nft store

BonomoAlessandro:
# Motivation

We want to be able to store our loaded nfts.

# Changes

- implements nft store

### DIFF
--- a/src/frontend/src/lib/components/tokens/Tokens.svelte
+++ b/src/frontend/src/lib/components/tokens/Tokens.svelte
@@ -1,12 +1,22 @@
 <script lang="ts">
+	import { nonNullish } from '@dfinity/utils';
 	import { fade } from 'svelte/transition';
+	import ManageTokensModal from '$lib/components/manage/ManageTokensModal.svelte';
 	import ManageTokensButton from '$lib/components/tokens/ManageTokensButton.svelte';
 	import TokensFilter from '$lib/components/tokens/TokensFilter.svelte';
 	import TokensList from '$lib/components/tokens/TokensList.svelte';
 	import TokensMenu from '$lib/components/tokens/TokensMenu.svelte';
 	import Header from '$lib/components/ui/Header.svelte';
+	import MessageBox from '$lib/components/ui/MessageBox.svelte';
 	import StickyHeader from '$lib/components/ui/StickyHeader.svelte';
+	import { modalManageTokens, modalManageTokensData } from '$lib/derived/modal.derived';
 	import { i18n } from '$lib/stores/i18n.store';
+
+	let { initialSearch, message } = $derived(
+		nonNullish($modalManageTokensData)
+			? $modalManageTokensData
+			: { initialSearch: undefined, message: undefined }
+	);
 </script>
 
 <div>
@@ -31,3 +41,15 @@
 		<ManageTokensButton />
 	</div>
 </div>
+
+{#if $modalManageTokens}
+	<ManageTokensModal {initialSearch}>
+		{#snippet infoElement()}
+			{#if nonNullish(message)}
+				<MessageBox level="info">
+					{message}
+				</MessageBox>
+			{/if}
+		{/snippet}
+	</ManageTokensModal>
+{/if}
--- a/src/frontend/src/lib/components/tokens/TokensList.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensList.svelte
@@ -6,7 +6,6 @@
 	import { goto } from '$app/navigation';
 	import { erc20UserTokensNotInitialized } from '$eth/derived/erc20.derived';
 	import Listener from '$lib/components/core/Listener.svelte';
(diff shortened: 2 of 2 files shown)


src/frontend/src/lib/stores/nft.store.ts
```diff
@@ -0,0 +1,61 @@
...
+						(nft) => nft.id === current.id && nft.contract.address === current.contract.address
+					)
+				) {
+					acc.push(current);
+				}
+				return acc;
+			}, []);
+
+			update((currentNfts) => {
+				if (!currentNfts) {
```
AntonioVentilii: `isNullish`

src/frontend/src/lib/stores/nft.store.ts
```diff
@@ -0,0 +1,61 @@
...
+			const uniqueNfts = nfts.reduce<Nft[]>((acc, current) => {
+				if (
+					!acc.some(
+						(nft) => nft.id === current.id && nft.contract.address === current.contract.address
+					)
+				) {
+					acc.push(current);
+				}
+				return acc;
+			}, []);
```
AntonioVentilii: just to avoid too many cycles, we can suppose that the input list is already unique (taken care by the consumer). We don't do this check in the other stores either.

What would be the case where the input has repeated elements? It seems like a simple safe-guard and not a practical situation

src/frontend/src/lib/stores/nft.store.ts
```diff
@@ -0,0 +1,61 @@
...
+								existingNft.id === newNft.id &&
+								existingNft.contract.address === newNft.contract.address
+						)
+				);
+
+				return [...currentNfts, ...newNfts];
+			});
+		},
+		getTokenIds: (contractAddress: string) => {
+			const currentNfts = get(nftStore);
```
AntonioVentilii: use `update` and not imperative `get`

src/frontend/src/lib/stores/nft.store.ts
```diff
@@ -0,0 +1,61 @@
...
+
+			if (!currentNfts) {
+				return [];
+			}
+
+			return currentNfts
+				.filter((nft) => nft.contract.address.toLowerCase() === contractAddress.toLowerCase())
+				.map((nft) => nft.id);
```
AntonioVentilii: `.reduce()` ?

src/frontend/src/lib/stores/nft.store.ts
BonomoAlessandro: You are right, good catch!

src/frontend/src/lib/stores/nft.store.ts
BonomoAlessandro: I would prefer to stay with the current solution because this is a very simple case and like this it is much more readable than using `reduce()`.

AntonioVentilii: two loops though... and in general the nft list is not that small

performance-wise, it would be better only one loop

BonomoAlessandro: Good argument, i will implement a `reduce()`

src/frontend/src/lib/stores/nft.store.ts
```diff
@@ -0,0 +1,61 @@
...
+						acc.push(nft.id);
+					}
+					return acc;
+				}, []);
+
+				return nfts;
+			});
+
+			return tokenIds;
+		},
```
AntonioVentilii: sorry, right now this is clearer: this function should not be here, right? it should either be a derived or a service that calls the store. Otherwise it is imperatively using the values (no reactivity) and I am not sure that is the use-case that we will require

BonomoAlessandro: So you think it is better to have a `service` function that filters the stored `nfts` by contract address than the store that provides a function to load a subset of the stored data?

I mean in the end we want to have a way to get a subset of the stored data that depends on the contract address.

AntonioVentilii: the problem is that `getTokenIds` will always get it imperatively, and not reactively. So, most probably, whenever you are going to consume it, you are going to require the reactivity somehow.

Typically we pass the store to some util/service so that they can extract informations, or we use derived if we don't need params.

I find it difficult to think of an use case where you will need this imperatively

BonomoAlessandro: @AntonioVentilii I need it in my `service` implementation that is coming soon. There i load some `nfts` and store them. The next time the service is running i want to be able to check which `nfts` are already in the store and then only load the `nfts` that are not loaded yet. So for this case it does not need to be reactive.

AntonioVentilii: then let's add it in the PR when you create the service: hard to believe that you need these values imperatively

BonomoAlessandro: @AntonioVentilii so i removed it from this PR and will add it with the PR that will add the `service` implementation.

PR #7824, opened 2025-07-16
refactor(frontend): Use generic transactions store interface in ETH transactions store

AntonioVentilii:
# Motivation

It makes sense to use the generic transactions store in the Ethereum transaction store initializer, to avoid repetition of code.


roman-nazaruk (approved):
LGTM, thnx

PR #7829, opened 2025-07-16
refactor(frontend): simplify types in `initAddressStore`

AntonioVentilii:
# Motivation

In the address store initializer we do not really need to import `StorageStoreData` type, since we just need the `Option` modifier, and the `Storage` prefix does not really make sense in this context.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7831, opened 2025-07-16
feat(frontend): Currency store

AntonioVentilii:
# Motivation

We require a currency store to keep track of the current currency selected by the user and the respective exchange rate to USD.

# Changes

- Create store based on storage store, with two methods:
  - `switchCurrency`: it changes the current currency and nullify the exchange rate (to avoid wrong data shown in the UI)
  - `setExchangeRate`: it just modifies the exchange rate in the store

# Tests

Added tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7842, opened 2025-07-17
feat(frontend): Create IDB methods for balances store

AntonioVentilii:
# Motivation

Very similar to what we did with transactions store, we create IDB methods to cache the balances store in IDB. They will be used in the wallet services for the initial loader of balances.

# Changes

- Create IDB method to set the balances store in IDB: this is by token, since the wallet services work by token.
- Create IDB method to get and delete the cached balances.

# Tests

Added tests.


src/frontend/src/tests/lib/api/idb-balances.api.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+	createStore: vi.fn(() => ({
+		/* mock store implementation */
+	})),
+	set: vi.fn(),
+	get: vi.fn(),
+	del: vi.fn(),
+	update: vi.fn()
+}));
+
+vi.mock('$app/environment', () => ({
```
DenysKarmazynDFINITY, line 23: Should we also mock nullishSignOut? I remember we had some issues with the tests because of that

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/tests/lib/api/idb-balances.api.spec.ts
AntonioVentilii: good catch! included!

PR #7845, opened 2025-07-17
feat(frontend): Derived store for current currency and exchange rate

AntonioVentilii:
# Motivation

To facilitate and simplify the code, we create derived stores based on `currencyStore`.

# Changes

- Derived store for pure currency value.
- Derived store for the exchange rate to USD.

# Tests

Added tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7852, opened 2025-07-17
feat(frontend): added new params to approveToken before send

roman-nazaruk:
# Motivation

To support flexible approval flows, in this PR was introduced conditions to handle approvals using a custom spender when required.

# Changes

- Extended types to support approveRequired condition

- Introduced logic to optionally use to as the spender for approval

### DIFF
--- a/src/frontend/src/eth/services/send.services.ts
+++ b/src/frontend/src/eth/services/send.services.ts
@@ -279,6 +279,7 @@ export const send = async ({
 
 	const { transactionNeededApproval, nonce } = await approve({
 		progress,
+		progressSteps: ProgressStepsSend,
 		token,
 		...rest
 	});
@@ -427,6 +428,7 @@ const prepareAndSignApproval = async ({
 	gas,
 	sourceNetwork,
 	identity,
+	progressSteps = ProgressStepsSend,
 	progress,
 	...rest
 }: SignAndApproveParams): Promise<
@@ -445,11 +447,11 @@ const prepareAndSignApproval = async ({
 		networkId
 	});
 
-	progress?.(ProgressStepsSend.SIGN_APPROVE);
+	progress?.(progressSteps.SIGN_APPROVE);
 
 	const rawTransaction = await signTransaction({ identity, transaction: approve });
 
-	progress?.(ProgressStepsSend.APPROVE);
+	progress?.(progressSteps.APPROVE);
 
 	const { sendTransaction } = infuraProviders(networkId);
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/eth/services/send.services.ts
```diff
@@ -445,11 +447,11 @@ const prepareAndSignApproval = async ({
 		networkId
 	});
 
-	progress?.(ProgressStepsSend.SIGN_APPROVE);
+	progress?.(progressSteps.SIGN_APPROVE);
```
DenysKarmazynDFINITY, line 450: How about instead of passing separately both `progress` and `progressSteps`, we change our approach - the send methods will accept just a single callback called `onProgress`. Then, consumers can decide which step should be set as next. The benefit is that the service does not need to know about UI stuff. We already have such a pattern in place, you can find it in `BtcConvertTokenWizard` and `BtcSendTokenWizard`.

DenysKarmazynDFINITY: We discussed it offline - implementing this approach would lead to a lot of changes. Therefore, we leave a TODO to implement it a later stage, but for now we'll proceed with the current code.

roman-nazaruk: As discussed, this is a solid proposal, but it will be implemented later in a dedicated refactor PR, since it requires a lot of changes across multiple files.

roman-nazaruk: Added TODO comments. Thanks

src/frontend/src/eth/services/send.services.ts
```diff
@@ -533,30 +538,39 @@ const approve = async ({
...
 
-	if (
-		isNullish(erc20HelperContractAddress) ||
-		!shouldSendWithApproval({
-			to,
-			tokenId: token.id,
-			erc20HelperContractAddress
-		})
-	) {
+	const shouldSkipApproval = nonNullish(approveRequired)
```
DenysKarmazynDFINITY: I am a bit unsure about the param name - `approveRequired` implies that if it's undefined or false, the approve call is not needed at all. But in reality, the approval will still be called in that case if erc20HelperContractAddress is present and shouldSendWithApproval returns true. Also, it's not really clear why we change a spender based on this param on the line 554. I suggest we change the name to something swap-related that will be clearly used for skipping the approval or using `to` as a spender if the flag is true.

DenysKarmazynDFINITY (approved):
Just a small suggestion on the naming, otherwise LGTM

src/frontend/src/eth/services/send.services.ts
roman-nazaruk: Renamed param to shouldSwapWithApproval

PR #7857, opened 2025-07-17
feat(frontend): added new swap eth service

roman-nazaruk:
# Motivation

As we aim to support sending swap transactions with custom parameters (typically returned from SDKs), this new service provides flexibility and control over swap flows

# Changes

- Added new swap ETH service method to send transactions using custom parameters from SDKs (to, data, value, gasLimit).


src/frontend/src/eth/services/swap.services.ts
```diff
@@ -0,0 +1,116 @@
+import type { EthSignTransactionRequest } from '$declarations/signer/signer.did';
+import { infuraProviders } from '$eth/providers/infura.providers';
+import { processTransactionSent } from '$eth/services/eth-transaction.services';
+import type { SwapParams } from '$eth/types/swap';
+import { signTransaction } from '$lib/api/signer.api';
+import { ProgressStepsSwap } from '$lib/enums/progress-steps';
+import { i18n } from '$lib/stores/i18n.store';
+
```
AntonioVentilii:
```suggestion
```

src/frontend/src/eth/services/swap.services.ts
```diff
@@ -0,0 +1,116 @@
...
+		throw new Error(erc20_data_undefined);
+	}
+
+	return {
+		to,
+		max_fee_per_gas,
+		max_priority_fee_per_gas,
+		nonce: BigInt(nonce),
+		data: [data],
```
AntonioVentilii: `toNullable(data)`

src/frontend/src/eth/services/swap.services.ts
```diff
@@ -0,0 +1,116 @@
...
+	maxPriorityFeePerGas,
+	sourceNetwork,
+	identity,
+	nonce
+}: Pick<TransferParams, 'from' | 'to'> &
+	Omit<SwapParams, 'lastProgressStep'> &
+	Omit<RequiredTransactionFeeData, 'gas'> & { nonce: number } & {
+		transaction: TransactionParams;
+	}): Promise<TransactionResponse> => {
+	const { id: networkId } = sourceNetwork;
```
AntonioVentilii, line 61: maybe, just to be sure, i would put a filter here for the sourceNetwork? Something like "if it's not one of the supported swap network, then raise an error"

I know that it should never happen, but just in case? WDYT?

src/frontend/src/eth/services/swap.services.ts
```diff
@@ -0,0 +1,116 @@
...
+	const rawTransaction = await signTransaction({
+		identity,
+		transaction: transactionParams,
+		nullishIdentityErrorMessage: get(i18n).auth.error.no_internet_identity
+	});
+
+	return await sendTransaction(rawTransaction);
+};
+
+export const swap = async ({
```
AntonioVentilii, line 82: no test?

src/frontend/src/eth/services/swap.services.ts
roman-nazaruk: Done.

src/frontend/src/eth/services/swap.services.ts
roman-nazaruk: As we discussed, this check will be handled in the service that calls the swap method, so there's no need to include the condition here. Also, the type for the sourceNetwork prop is already defined as EthereumNetwork.

src/frontend/src/eth/services/swap.services.ts
roman-nazaruk: Added tests

PR #7858, opened 2025-07-17
refactor(frontend): Use `parseBoolEnvVar` for feature flag on earnings

AntonioVentilii:
# Motivation

For consistency, we use `parseBoolEnvVar` util for the feature flag of earnings.


src/frontend/src/env/earning.ts
```diff
@@ -1,6 +1,5 @@
-import { LOCAL, STAGING } from '$lib/constants/app.constants';
+import { parseBoolEnvVar } from '$lib/utils/env.utils';
 
 // Enabled on Staging and Local if not set
 // Todo: remove once the feature has been completed
-export const EARNING_ENABLED =
-	JSON.parse(import.meta.env.VITE_EARNING_ENABLED ?? (STAGING || LOCAL)) === true;
+export const EARNING_ENABLED = parseBoolEnvVar(import.meta.env.VITE_EARNING_ENABLED);
```
BonomoAlessandro, line 5: But why remove the `?? (STAGING || LOCAL)`? This activates the feature per default if nothing else is configured.

AntonioVentilii: nono, by default this is always `false`, if the feature flag is not set. It should be controlled via CI. Unless you change the variable in the `.env` file or in the CI secrets.

If you need it on by default in STAGING and LOCAL, we can add it after:

`export const EARNING_ENABLED = parseBoolEnvVar(import.meta.env.VITE_EARNING_ENABLED) || (STAGING || LOCAL)`

but it does not really make sense, since you can control it in both environment via `.env` file

PR #7859, opened 2025-07-17
feat(frontend): added new derived for all cc swap tokens

roman-nazaruk:
# Motivation

As we need list of all supported tokens for cross-chain swaps, this PR introduces a new derived store that filters them accordingly.

# Changes

Added a new allCrossChainSwapTokens derived store to provide a filtered list of tokens supported in cross-chain swaps.

# Tests

Covered new logic with a tests.


BonomoAlessandro (approved):
LGTM

PR #7863, opened 2025-07-17
feat(frontend): added new modal network list store

roman-nazaruk:
# Motivation

To enable selecting different networks for cross-chain swaps, we need a store to manage the list of networks shown in the modal and update filters accordingly.

# Changes

Added new modalNetworksListStore to manage network selection for the modal.

# Tests

Covered the new store with unit tests.

### DIFF
--- a/src/frontend/src/lib/stores/modal-networks-list.store.ts
+++ b/src/frontend/src/lib/stores/modal-networks-list.store.ts
@@ -1,20 +1,23 @@
 import type { Network, NetworkId } from '$lib/types/network';
-import { nonNullish } from '@dfinity/utils';
 import { derived, writable, type Readable } from 'svelte/store';
 
 export interface ModalNetworksListData {
 	networks: Network[];
-	allowedNetworkIds?: NetworkId[];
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/stores/modal-networks-list.store.ts
```diff
@@ -0,0 +1,36 @@
+import type { Network, NetworkId } from '$lib/types/network';
+import { nonNullish } from '@dfinity/utils';
+import { derived, writable, type Readable } from 'svelte/store';
+
+export interface ModalNetworksListData {
+	networks: Network[];
+	allowedNetworkIds?: NetworkId[];
```
AntonioVentilii: just a doubt: does it make sense to leave it non-nullish and force it to be an empty list? for example, you can call `setAllowedNetworkIds()` and it does not really make sense, right? Better to do something like `setAllowedNetworkIds([])` or `resetAllowedNetworkIds()`

roman-nazaruk: Agree, fixed

PR #7864, opened 2025-07-17
feat(frontend): added new filter by networks id to modal-tokens-list store

roman-nazaruk:
# Motivation

To enable selecting different networks for cross-chain swaps, we need to update modal-tokens list store to manage the list of tokens filtered by networks ids

# Changes

Added new filterNetworksIds to manage tokens selection for the modal.

# Tests

Covered the new logic with tests.


AntonioVentilii (commented):
May you please add some test for hedge cases? like `filterNetworksIds` is nullish, or empty

AntonioVentilii (approved):
THank you! LGTM!

PR #7866, opened 2025-07-17
refactor(frontend): Create zod schemas for types of currency stores

AntonioVentilii:
# Motivation

We will need the interfaces and types of the currency stores to be used by the workers types too. Since the latter are set in zod, it seems a good time to migrate the others to zod too.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7870, opened 2025-07-17
feat(frontend): Add custom currency to util `formatCurrency`

AntonioVentilii:
# Motivation

We will let the users choose their own currency. So, we need to start formatting it correctly (some of them have 3 decimals, other no decimals at all).

The `Intl` library knows these rules when `currency` is selected in method `NumberFormat`, so we can rely on it (see [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#currency_formatting)).

# Changes

- Add currency param in util `formatCurrency`.
- Use it to define the currency in `Intl.NumberFormat()`.
- Adapt the consumers to use the store `currentCurrency`.
- Remove constraints of minimum and maximum digits: the library knows exactly how many digits each currency should have.

# Tests

Added test cases.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7873, opened 2025-07-17
feat(frontend): Sync balances from IDB cache during initial wallet sync

AntonioVentilii:
# Motivation

Similar to what we did for transactions, we sync the IDB cache with our balances store when there is the first initialization sync.

# Changes

- Created service to get a balance from IDB cache by token.
- Use the new service in `syncWalletFromIdbCache`.

# Tests

New tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7876, opened 2025-07-17
feat(frontend): component to save the IDB store for balances

AntonioVentilii:
# Motivation

Similar to what we did for the transactions stores, we will start using the IDB services that store the balances in the cache.

The idea is that each change in the balances store would trigger caching it. That is because the balances store is handled in various places in the code, and it becomes quite un-maintainsble to add the code in each use. 

# Changes

- Create component `BalancesIdbSetter` to cache the balances (very similar to the transacitons one): we separate the reactivity and we limit it only to the balances store. This will avoid useless updates.
- Add the `BalancesIdbSetter` to `Loaders`.

# Tests

Manual test worked too:

Screenshot 2025-07-18 at 01 29 01


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7878, opened 2025-07-18
refactor(frontend): Rename enum of supported currencies

AntonioVentilii:
# Motivation

It makes more sense to rename the enum `Currencies` to `Currency` (before it is used even more).


DenysKarmazynDFINITY (approved):
👍

PR #7879, opened 2025-07-18
feat(frontend): Debounce wallet workers trigger

AntonioVentilii:
# Motivation

To avoid too many calls on loading, we debounce the trigger of the wallet workers.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7884, opened 2025-07-18
feat(frontend): Do not wait for Signer allowance if derivation is in frontend

AntonioVentilii:
# Motivation

When we initialize the UI, we are requesting an allowance to the Chain Fusion signer. This is useful for most of the calls that we do to the Chain Fusion signer, for example deriving the addresses, or sign a transaction.

However, if we are deriving the addresses in the frontend, we do not need tov await for the call to finish. We can accept the fact that the user will take some time until the next transaction signing, enough for the call to complete.

In any case, even if the call is not completed, the error received will just tell the user to re-try. And if there is an error, the call service will sign-out.

# Results

Initialization in beta canister is much faster:

https://github.com/user-attachments/assets/8d0a500d-e47d-4813-91d3-91c8711f59f4


DenysKarmazynDFINITY (approved):
Nice 🚀

PR #7886, opened 2025-07-18
feat(frontend): implements nft service

BonomoAlessandro:
# Motivation

We want to have a service that loads the `nfts` from the blockchain.

# Changes

- implements `nft` service


src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+
+export const loadNfts = ({
+	tokens,
+	loadedNfts,
+	walletAddress
+}: {
+	tokens: Erc721CustomToken[];
+	loadedNfts: Nft[];
+	walletAddress: string;
+}) => {
```
AntonioVentilii: return type missing?

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+	token,
+	loadedNfts,
+	walletAddress
+}: {
+	etherscanProvider: EtherscanProvider;
+	infuraProvider: InfuraErc721Provider;
+	token: Erc721CustomToken;
+	loadedNfts: Nft[];
+	walletAddress: string;
+}) => {
```
AntonioVentilii: same

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+	let tokenIds: number[];
```
AntonioVentilii: can we do it functional? meaning, all inside the try-catch, or creating a sub-service for `tokenIds`

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+};
+
+const loadNftsOfBatch = async ({
+	infuraProvider,
+	token,
+	tokenIds
+}: {
+	infuraProvider: InfuraErc721Provider;
+	token: Erc721CustomToken;
+	tokenIds: number[];
```
AntonioVentilii, line 86: isn't this NftId? Or am i mistaken?

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+};
+
+const loadNftsMetadata = async ({
+	infuraProvider,
+	contractAddress,
+	tokenIds
+}: {
+	infuraProvider: InfuraErc721Provider;
+	contractAddress: string;
+	tokenIds: number[];
```
AntonioVentilii, line 107: same?

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+}) => {
+	const nftsMetadata: NftMetadata[] = [];
+	for (const tokenId of tokenIds) {
+		nftsMetadata.push(await loadNftMetadata({ infuraProvider, contractAddress, tokenId }));
+	}
+
+	return nftsMetadata;
+};
```
AntonioVentilii: how are errors handled for a single failure? everything fails? if it's ok, let's use `Promise.all`, otherwise `Promise.allSettled`

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+
+const loadNftMetadata = async ({
+	infuraProvider,
+	contractAddress,
+	tokenId
+}: {
+	infuraProvider: InfuraErc721Provider;
+	contractAddress: string;
+	tokenId: number;
```
AntonioVentilii, line 134: NftId?

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+}) => {
+	await new Promise((resolve) => setTimeout(resolve, 200));
```
AntonioVentilii: why this?

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+
+	try {
+		return await infuraProvider.getNftMetadata({
+			contractAddress,
+			tokenId
+		});
+	} catch (_: unknown) {
+		return { id: parseNftId(tokenId) };
```
AntonioVentilii, line 147: maybe an explanation on why we don't care about errors?

P.S. is it worth a `console.warn` or it really does not matter?

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+	loadedNfts: Nft[];
+	walletAddress: string;
+}) => {
+	let tokenIds: number[];
+	try {
+		tokenIds = await etherscanProvider.erc721TokenInventory({
+			address: walletAddress,
+			contractAddress: token.address
+		});
+	} catch (_: unknown) {
```
AntonioVentilii: same question

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+		tokenIds = [];
+	}
+
+	const loadedTokenIds: NftId[] = loadedNfts.map((nft) => nft.id);
+	const tokenIdsToLoad = tokenIds.filter((id: number) => !loadedTokenIds.includes(parseNftId(id)));
+
+	const tokenIdBatches = createBatches({ tokenIds: tokenIdsToLoad, batchSize: 10 });
+	for (const tokenIds of tokenIdBatches) {
+		const nfts = await loadNftsOfBatch({ infuraProvider, token, tokenIds });
```
AntonioVentilii: how are failures handled here?

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -0,0 +1,143 @@
...
+	walletAddress: string;
+}) => {
+	const etherscanProvider = etherscanProviders(ETHEREUM_NETWORK.id);
+	const infuraProvider = infuraErc721Providers(ETHEREUM_NETWORK.id);
+
+	const loadedNftsByNetwork: NftsByNetwork = getLoadedNftsByTokens({ tokens, loadedNfts });
+
+	return Promise.all(
+		tokens.map((token) => {
+			let loadedNfts: Nft[] = [];
```
AntonioVentilii: this can be defined functional, right?

src/frontend/src/lib/services/nft.services.ts
BonomoAlessandro: No, here it is `number`, because we are working with the result that we got from the `etherscanProvider`.

src/frontend/src/lib/services/nft.services.ts
BonomoAlessandro: same as above

src/frontend/src/lib/services/nft.services.ts
BonomoAlessandro: The errors are not handled here. The errors are handled in the `loadNftMetadata` function.

src/frontend/src/lib/services/nft.services.ts
BonomoAlessandro: same as the other two

src/frontend/src/lib/services/nft.services.ts
BonomoAlessandro: This is to not send too many requests at once to infura. Like this we load 5 nfts per second.

src/frontend/src/lib/services/nft.services.ts
BonomoAlessandro: A `console.warn` could make sense, but otherwise we have no way to handle the error, so we load the nft without the metadata.

src/frontend/src/lib/services/nft.services.ts
BonomoAlessandro: If we cannot load the `tokenIds` that the user is holding of a specific contract, we skip this contract and load the `nfts` of the other contracts.

src/frontend/src/lib/services/nft.services.ts
BonomoAlessandro: No failures are handled here. The `loadNftMetadata` function handles the failures.

src/frontend/src/lib/services/nft.services.ts
AntonioVentilii: Let's add some comment for this

AntonioVentilii (approved):
LGTM tks!

PR #7893, opened 2025-07-18
feat(frontend): Get all SPL token info in a single service

AntonioVentilii:
# Motivation

We are going to require more information on an SPL token: for example, we need the `mintAuthority` to do a `transferChecked` for Token Program 2022.

In the meanwhile, we already query the information for the decimals and the owner.

To avoid multiple calls, we can reduce it into a single service: we take the service that was fetching the token owner and expand it to take all the necessary info too.

# Changes

- Expand type of `SplToken` to have the `mintAuthority` and the `freezeAuthority` (optional, since not all tokens have it).
- Refactor service `getTokenOwner` to be `getTokenInfo`: it will take the `decimals` and the other info, beyond the `owner`.
- The nullish checks are the same as the `getTokenDecimals` service, and the fallbacks too.
- Adapt consumers.
- Remove deprecated service `getTokenDecimals`.

# Tests

Added tests.


src/frontend/src/sol/api/solana.api.ts
```diff
@@ -197,44 +197,40 @@ export const estimatePriorityFee = async ({
...
 
 	const { value } = await getAccountInfo(token, { encoding: 'jsonParsed' }).send();
 
-	if (nonNullish(value) && 'parsed' in value.data) {
-		const {
-			data: {
-				parsed: { info }
-			}
-		} = value;
+	const owner = value?.owner?.toString();
```
roman-nazaruk: WDYT about destructuring it? const { value: { owner, data } }

src/frontend/src/sol/api/solana.api.ts
```diff
@@ -197,44 +197,40 @@ export const estimatePriorityFee = async ({
...
-export const getTokenOwner = async ({
-	address,
-	network
-}: {
-	address: SplTokenAddress;
-	network: SolanaNetworkType;
-}): Promise<SplTokenAddress | undefined> => {
-	const { getAccountInfo } = solanaHttpRpc(network);
-	const token = solAddress(address);
+	if (isNullish(value.data.parsed?.info)) {
```
roman-nazaruk: Maybe as well here as we use data.parsed in 2 places. Row 227 f.e.

roman-nazaruk (approved):
LGTM, thnx

src/frontend/src/sol/api/solana.api.ts
AntonioVentilii: done!

src/frontend/src/sol/api/solana.api.ts
AntonioVentilii: done!

PR #7897, opened 2025-07-21
chore(frontend): Add Czech Koruna currency

StefanBerger-DFINITY:
# Motivation

We are adding Czech as language, so we can also add the currency.

# Changes

- add currency (ISO 4217 code: CZK)

# Tests
image


roman-nazaruk (approved):
LGTM, thnx

PR #7904, opened 2025-07-21
feat(frontend): Derived store with current currency symbol

AntonioVentilii:
# Motivation

It is useful to have a direct derived store for the current currency symbol:

Screenshot 2025-07-21 at 11 36 35


BonomoAlessandro (approved):
LGTM

PR #7907, opened 2025-07-21
feat(frontend): triggers loading of nfts

BonomoAlessandro:
# Motivation

The `Loader` component should trigger the loading of the `nfts`.

# Changes

- triggers loading of nfts

### DIFF
--- a/src/frontend/src/lib/services/nft.services.ts
+++ b/src/frontend/src/lib/services/nft.services.ts
@@ -20,9 +20,9 @@ export const loadNfts = ({
 	tokens: Erc721Token[];
 	loadedNfts: Nft[];
 	walletAddress: OptionEthAddress;
-}): void[] => {
+}): Promise<void[]> => {
 	if (isNullish(walletAddress)) {
-		return;
+		return Promise.resolve([]);
 	}
 
 	const loadedNftsByNetwork: NftsByNetwork = getNftsByNetworks({ tokens, nfts: loadedNfts });


src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -151,6 +155,19 @@
 		}
 	}
 
+	const debounceLoadNfts = debounce(() => {
+		if (nonNullish($ethAddressStore?.data)) {
```
AntonioVentilii: you can use derived `$ethAddress`

but the nullish value should be handled inside the service, and the component should just call the service

src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -151,6 +155,19 @@
...
+			const tokensList = $erc721CustomTokensStore?.map((entry) => entry.data) ?? [];
+			const loadedNfts = $nftStore ?? [];
+
+			loadNfts({ tokens: tokensList, loadedNfts, walletAddress: $ethAddressStore.data });
+		}
+	});
+
+	$: if ($erc721CustomTokensStore) {
```
AntonioVentilii: let's create a derived that is like `erc20UserTokensInitialized`, something like `erc721CustomTokensInitialized`

src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -151,6 +155,19 @@
 		}
 	}
 
+	const debounceLoadNfts = debounce(() => {
+		if (nonNullish($ethAddressStore?.data)) {
+			const tokensList = $erc721CustomTokensStore?.map((entry) => entry.data) ?? [];
+			const loadedNfts = $nftStore ?? [];
+
+			loadNfts({ tokens: tokensList, loadedNfts, walletAddress: $ethAddressStore.data });
```
AntonioVentilii:
```suggestion
			loadNfts({ tokens: tokensList, loadedNfts:$nftStore ?? [], walletAddress: $ethAddressStore.data });
```

src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -151,6 +155,19 @@
...
+		}
+	});
+
+	$: if ($erc721CustomTokensStore) {
+		debounceLoadNfts();
```
AntonioVentilii, line 166: you probably have to add the reactivity of the dependencies of debounceLoadNfts here:

$: $ethAddress, .... , if....

src/frontend/src/lib/components/loaders/Loader.svelte
BonomoAlessandro: to use `$ethAddress` is a good idea, but i think it is better to handle the nullish value here. The service implementation just says that it wants to get a nonNullish walletAddress, so the caller has to handle it.

src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -151,6 +155,18 @@
 		}
 	}
 
+	const debounceLoadNfts = debounce(() => {
+		if (nonNullish($ethAddress)) {
+			const tokensList = $erc721CustomTokensStore?.map((entry) => entry.data) ?? [];
```
AntonioVentilii: should you not use `erc721CustomTokens` ? it is filtered by enabled networks too

src/frontend/src/lib/components/loaders/Loader.svelte
AntonioVentilii: indeed, we should not have this check in the component: better to put it in the service (similar to the other loaders of tokens)

AntonioVentilii (commented):
Please add some test too

src/frontend/src/lib/components/loaders/Loader.svelte
BonomoAlessandro: Yes i guess i should use `erc721Tokens`.

src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -151,6 +154,18 @@
...
 
+	const debounceLoadNfts = debounce(() => {
+		loadNfts({
+			tokens: $erc721Tokens ?? [],
+			loadedNfts: $nftStore ?? [],
+			walletAddress: $ethAddress
+		});
+	});
+
+	$: if ($erc721CustomTokensInitialized && nonNullish($ethAddress) && $erc721Tokens) {
```
AntonioVentilii, line 165:
```suggestion
	$:$erc721Tokens, $nftStore, $erc721CustomTokensInitialized, $ethAddress, if ($erc721CustomTokensInitialized && nonNullish($ethAddress)) {
```

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -18,8 +19,12 @@ export const loadNfts = ({
 }: {
 	tokens: Erc721Token[];
 	loadedNfts: Nft[];
-	walletAddress: string;
+	walletAddress: OptionEthAddress;
 }): Promise<void[]> => {
+	if (isNullish(walletAddress)) {
+		return Promise.resolve([]);
```
AntonioVentilii, line 25:
```suggestion
		return;
```

src/frontend/src/lib/services/nft.services.ts
```diff
@@ -18,8 +19,12 @@ export const loadNfts = ({
 }: {
 	tokens: Erc721Token[];
 	loadedNfts: Nft[];
-	walletAddress: string;
+	walletAddress: OptionEthAddress;
 }): Promise<void[]> => {
```
AntonioVentilii, line 23: just noticed, since we are here, you don't need to return the promises, right? you can just call them, and reutrn nothing?

BonomoAlessandro: We want to return promises. The caller should be able to await the loading of the nfts even if we don't use it yet.

PR #7910, opened 2025-07-21
feat(frontend): improves loading of nft metadata image

BonomoAlessandro:
# Motivation

The metadata of an `nft` can contain the url to the image as `image` or `image_url`. We want to be able to support both options.

# Changes

- improves loading of metadata image

### DIFF
--- a/src/frontend/src/eth/providers/infura-erc721.providers.ts
+++ b/src/frontend/src/eth/providers/infura-erc721.providers.ts
@@ -89,7 +89,7 @@ export class InfuraErc721Provider {
 		const response = await fetch(metadataUrl);
 		const metadata = await response.json();
 
-		const imageUrl = extractImageUrl(metadata.image);
+		const imageUrl = extractImageUrl(metadata.image ?? metadata.image_url);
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/eth/providers/infura-erc721.providers.ts
```diff
@@ -89,7 +89,7 @@ export class InfuraErc721Provider {
 		const response = await fetch(metadataUrl);
 		const metadata = await response.json();
 
-		const imageUrl = extractImageUrl(metadata.image);
+		const imageUrl = extractImageUrl(metadata.image ?? metadata.image_url);
```
AntonioVentilii, line 92: just a generic doubt: which one should have priority if they have both? or they never have both?

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/eth/providers/infura-erc721.providers.ts
BonomoAlessandro: In theory it is possible that both are defined, but it does not make sense. We asume that if both are defined, `image` has priority.

PR #7912, opened 2025-07-21
feat(frontend): replaces isErc721 check with isErc20 check

BonomoAlessandro:
# Motivation

Not all `erc721` contracts implement the `supportsInterface` function which means we cannot rely on this. To have a more reliable solution we replace the `isErc721` function with the `isErc20` function that checks if the contracts provides a `decimals` function.

# Changes

- improves way to determine if a contract is `erc20` or `erc721`


src/frontend/src/eth/providers/infura-erc20.providers.ts
```diff
@@ -114,6 +114,17 @@ export class InfuraErc20Provider implements Erc20Provider {
...
+	isErc20 = async ({ contractAddress }: { contractAddress: string }): Promise<boolean> => {
+		const erc20Contract = new Contract(contractAddress, ERC20_ABI, this.provider);
+
+		try {
+			await erc20Contract.decimals();
+			return true;
+		} catch (_: unknown) {
+			return false;
+		}
+	};
```
AntonioVentilii, line 129: funny enough you can transform this in `isErc721`, inverting the condition, eheh! So bizarre... Kidding aside: I imagine we have no other choice for now, right? Can you add some comment above this function for posterity, just to be clear that is a workaround?

"We use this one to differentiate between ERC721 and ERC20, because...."

AntonioVentilii (approved):
I imagine, no other choice, eh?

src/frontend/src/eth/providers/infura-erc20.providers.ts
BonomoAlessandro: Yes currently we have no other choice than doing this...
Yes, I will add a comment.

PR #7865, opened 2025-07-17
feat(frontend): set new avatar

CvetelinaAngelova:
# Motivation

Enable users to upload, preview and replace a contact’s avatar directly within the Edit Contact flow. This enhances the user experience by providing an intuitive, in-place avatar management UI.

# Changes
Implement menu with PopOver
Implemented handleFileChange to:
reead the selected file and compress it (≤100 KB, ≤200 px) via imageCompression
Add handleAddAvatar to update the image to the Api

# Tests

Manual test and Unit test for EditContact component
Verified behavior on both desktop and mobile viewport sizes.

https://github.com/user-attachments/assets/ccabf818-7c0d-48fe-a670-998f50d6baf2

PR #7915, opened 2025-07-21
feat(frontend): add handleAddAvatar Avatar to the AddressBookModal

CvetelinaAngelova:
# Motivation

This Pr include part of the  set Avatar logic 
https://github.com/dfinity/oisy-wallet/pull/7865/

# Changes

Includes handleAddAvatar function into AddressBookModal

# Tests


src/frontend/src/lib/components/address-book/EditContactStep.svelte
```diff
@@ -35,11 +37,14 @@
...
 		onClose,
 		onEdit,
+		onAvatarEdit,
 		onEditAddress,
 		onAddAddress,
 		onDeleteContact,
 		onDeleteAddress
 	}: Props = $props();
+
+	void onAvatarEdit;
```
AntonioVentilii, line 47:
```suggestion
```

src/frontend/src/lib/components/address-book/AddressBookModal.svelte
```diff
@@ -244,6 +245,15 @@
...
 		gotoStep(AddressBookSteps.SHOW_CONTACT);
 	};
+	const handleAddAvatar = async (image: ContactImage) => {
+		if (isNullish(currentContact)) {
+			return;
+		}
+		const contact = {
+			...currentContact
+		};
+		await callUpdateContact({ contact, image });
```
AntonioVentilii, line 255: The errors are correctly handled here, right? For example, going back, showing the toast, or something else?

src/frontend/src/lib/components/address-book/EditContactStep.svelte
CvetelinaAngelova: This is only to avoid int errors and pass the jobs. Will remove it in the next Pr and use the function appropriate.

src/frontend/src/lib/components/address-book/AddressBookModal.svelte
CvetelinaAngelova: All the calls here uses the same util function (wrapCallWith ) which is handling the error messages. It's the same one for CreateContact, UpdateContact and DeleteContact

AntonioVentilii: but are the steps handled too? For example, in case of error, to which steps it goes? Same, back, next?

CvetelinaAngelova: The same step, we do not change the Steps on the AvatarUpload that's why I have new call and not reuse onEdit.

CvetelinaAngelova: https://github.com/dfinity/oisy-wallet/blob/main/src/frontend/src/lib/services/utils.services.ts

PR #7916, opened 2025-07-21
feat(frontend): use current currency in the input of the send flow

AntonioVentilii:
# Motivation

In the send flow, we need to use the current currency when the user prefers to input a FIAT amount.

So we focus on modifying the `TokenInputCurrencyFiat` component.

# Changes

- Use derived store `currentCurrencySymbol` instead of dollar sign.
- Convert the FIAT input amount from USD to the current currency.
- Adapt the format util to normalize the separators: this is important when we use `Number()` to re-convert the input string to a value.

# Tests

Added test. And a practical one:

https://github.com/user-attachments/assets/3523f54c-fd5f-4097-9884-d035f709df55


src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -196,15 +198,31 @@ export const formatCurrency = ({
 
 	const convertedValue = value / exchangeRateToUsd;
 
-	const formatted = new Intl.NumberFormat(locale, {
+	const nf = new Intl.NumberFormat(locale, {
```
roman-nazaruk: WDYT about renaming it to something more explicit like formattedNumber?
nf feels unclear and might not be understandable. But it’s up to you,  just a suggestion for readability.

roman-nazaruk (approved):
LGTM, thnx

src/frontend/src/lib/utils/format.utils.ts
AntonioVentilii: Good idea! Changed

PR #7919, opened 2025-07-21
feat(frontend): Use `TransferChecked` in Solana send flow

AntonioVentilii:
# WIP - Needs to clean the debbuging

# Motivation

According to [Solana documentation](https://solana.com/docs/tokens/basics/transfer-tokens), a more safe way of transferring tokens between accounts is `TransferChecked` instead of the simple `Transfer` method.

The difference is that the checked method verify the mint authority and the decimals too, guaranteeing that it is indeed the token that the user wants to transfer.

However, that entails that not all tokens can use it: not all of them have a defined `mintAuthority`.

So, for now, we use it ONLY when the mintAuthority is defined.

# Changes

- Refactor service `createSplTokenTransactionMessage` to use method `getTransferCheckedInstruction` when the `mintAuthority` of the token is non-nullish.
- Adapt the rest of the code.

# Tests

Tests adapted and a practical one:

### DIFF
--- a/src/frontend/src/sol/services/sol-send.services.ts
+++ b/src/frontend/src/sol/services/sol-send.services.ts
@@ -13,20 +13,23 @@ import {
 import type { SolanaNetworkType } from '$sol/types/network';
 import type { SolTransactionMessage } from '$sol/types/sol-send';
 import type { SolSignedTransaction } from '$sol/types/sol-transaction';
-import type { SplTokenAddress } from '$sol/types/spl';
+import type { SplToken } from '$sol/types/spl';
 import { safeMapNetworkIdToNetwork } from '$sol/utils/safe-network.utils';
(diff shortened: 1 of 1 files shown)


src/frontend/src/sol/services/sol-send.services.ts
```diff
@@ -324,8 +337,10 @@ export const sendSol = async ({
 			rpc
 		});
 
-	const computeUnitsEstimate =
-		await getComputeUnitEstimateForTransactionMessage(transactionMessage);
+	const computeUnitsEstimate = await getComputeUnitEstimateForTransactionMessage(
+		transactionMessage,
+		{ commitment: 'confirmed' }
```
AntonioVentilii:
```suggestion
```

PR #7929, opened 2025-07-22
feat(frontend): Add new component HiddenFileInput

CvetelinaAngelova:
# Motivation

Add new HiddenFileInput component to use in the Ui

# Changes

Add svelte component and style for it

# Tests


AntonioVentilii (commented):
Maybe I was not able to explain correctly: my point was to merge the `EditAvatar` and this one in a single new component, since they seem very linked (one cannot exists without the other)

CvetelinaAngelova:
@AntonioVentilii The HiddenInput could be used separately from EditAvatar. For example we could make the Avatar itself clickable with this hidden Input without having EditAvatar menu with popOver menu with options. Into Create contact step we have single Avatar and in the feature we could provide upload image only.

src/frontend/src/lib/components/ui/HiddenFileInput.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
+	interface Props {
+		onclick?: () => void;
```
AntonioVentilii:
```suggestion
		onClick?: () => void;
```

src/frontend/src/lib/components/ui/HiddenFileInput.svelte
```diff
@@ -0,0 +1,23 @@
...
+	};
+</script>
+
+<input
+	bind:this={fileInput}
+	type="file"
+	accept="image/*"
+	data-tid={testId}
+	class="visually-hidden"
+	on:change
```
AntonioVentilii: this is deprecated right? it will not work, unless you pass it as prop? `onchange={onChange)`

src/frontend/src/lib/components/ui/HiddenFileInput.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
+	interface Props {
+		onclick?: () => void;
+		testId?: string;
+	}
+
+	let { onclick, testId }: Props = $props();
+	let fileInput: HTMLInputElement;
```
AntonioVentilii:
```suggestion

	let fileInput = $state< HTMLInputElement  | undefined>()
```

src/frontend/src/lib/styles/global.scss
```diff
@@ -94,4 +94,14 @@
 			}
 		}
 	}
+	input {
```
AntonioVentilii: let's not put it in global, we should deprecate it either way. Create a mixin, or even directly inside the component

CvetelinaAngelova: add mixin and separate it

src/frontend/src/lib/components/ui/HiddenFileInput.svelte
CvetelinaAngelova: add it to the prop

src/frontend/src/lib/components/ui/HiddenFileInput.svelte
```diff
@@ -0,0 +1,25 @@
+<script lang="ts">
+	interface Props {
+		onClick?: () => void;
+		onchange?: (e: Event) => void;
```
AntonioVentilii:
```suggestion
		onChange?: (e: Event) => void;
```

src/frontend/src/lib/components/ui/HiddenFileInput.svelte
```diff
@@ -0,0 +1,25 @@
...
+	};
+</script>
+
+<input
+	bind:this={fileInput}
+	type="file"
+	accept="image/*"
+	data-tid={testId}
+	class="visually-hidden"
+	{onchange}
```
AntonioVentilii:
```suggestion
	onchange={() => onChange?.()}
```

src/frontend/src/lib/components/ui/HiddenFileInput.svelte
```diff
@@ -0,0 +1,25 @@
+<script lang="ts">
+	interface Props {
+		onClick?: () => void;
+		onChange?: (e: Event) => void;
+		testId?: string;
+	}
+
+	let { onClick, onchange, testId }: Props = $props();
```
AntonioVentilii:
```suggestion
	let { onClick, onChange, testId }: Props = $props();
```

src/frontend/src/lib/components/ui/HiddenFileInput.svelte
```diff
@@ -0,0 +1,25 @@
...
+	};
+</script>
+
+<input
+	bind:this={fileInput}
+	type="file"
+	accept="image/*"
+	data-tid={testId}
+	class="visually-hidden"
+	onchange={() => onChange?.()}
```
AntonioVentilii:
```suggestion
	onchange={(e: Event) => onChange?.(e)}
```

PR #7935, opened 2025-07-22
feat(frontend): added new Swap Error class

roman-nazaruk:
# Motivation

As we introduced new translations, we needed to improve error handling in swap services. This PR introduces a new SwapError class that extends the native Error object, enabling more structured handling of error codes and messages.

# Changes

- Added a SwapError class to represent swap failures with specific error codes and localized messages.

- Updated error-handling logic in services to use the new SwapError for better integration with translations.

# Tests

Added unit tests to cover the new error-handling logic and ensure correct behavior of SwapError.

### DIFF
--- a/src/frontend/src/lib/services/swap-errors.services.ts
+++ b/src/frontend/src/lib/services/swap-errors.services.ts
@@ -1,0 +1,16 @@
+import { i18n } from '$lib/stores/i18n.store';
+import type { SwapErrorKey } from '$lib/types/swap';
+import { get } from 'svelte/store';
+
+export const getSwapErrorMessage = (key: keyof I18n['swap']['error']) => get(i18n).swap.error[key];
+
+export const throwSwapError = (code: SwapErrorKey): never => {
+	throw new SwapError(code);
+};
+
+export class SwapError extends Error {
+	constructor(public readonly code: SwapErrorKey) {
+		super(get(i18n).swap.error[code]);
+		this.name = 'SwapError';
+	}
+}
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/services/swap-errors.services.ts
```diff
@@ -0,0 +1,16 @@
+import { i18n } from '$lib/stores/i18n.store';
+import type { SwapErrorKey } from '$lib/types/swap';
+import { get } from 'svelte/store';
+
+export const getSwapErrorMessage = (key: keyof I18n['swap']['error']) => get(i18n).swap.error[key];
```
AntonioVentilii: you can use `SwapErrorKey` here?

furthermore, we should give the `i18n` store as input, and it MUST be reactive in the consumer, since now we have different languages

src/frontend/src/lib/services/swap-errors.services.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const throwSwapError = (code: SwapErrorKey): never => {
+	throw new SwapError(code);
+};
+
+export class SwapError extends Error {
+	constructor(public readonly code: SwapErrorKey) {
+		super(get(i18n).swap.error[code]);
```
AntonioVentilii: it must be passed in the consumer, otherwise it cannot be reactive

src/frontend/src/lib/services/swap-errors.services.ts
roman-nazaruk: Nice catch, updated

src/frontend/src/lib/services/swap-errors.services.ts
roman-nazaruk: Updated. Nice tip

src/frontend/src/lib/services/swap-errors.services.ts
roman-nazaruk: I removed this method, we do not need it.

PR #7939, opened 2025-07-22
feat(frontend): adjusted errors handling for swap in service and component

roman-nazaruk:
# Motivation

Adjusted Swap service to throw and handle errors in the Swap component

# Changes
- Refactored Swap service to throw errors using throwSwapError
- Updated error handling logic in the Swap component
- Added new properties to Plausible tracking events

### DIFF
--- a/src/frontend/src/env/i18n.ts
+++ b/src/frontend/src/env/i18n.ts
@@ -8,5 +8,6 @@ export const LANGUAGES = {
 	[Languages.GERMAN]: 'Deutsch',
 	[Languages.ITALIAN]: 'Italiano',
 	[Languages.PORTUGUESE]: 'Português',
+	[Languages.VIETNAMESE]: 'Tiếng việt',
 	[Languages.CHINESE_SIMPLIFIED]: '中文 (简体)'
 };
--- a/src/frontend/src/lib/enums/languages.ts
+++ b/src/frontend/src/lib/enums/languages.ts
@@ -3,5 +3,6 @@ export enum Languages {
 	GERMAN = 'de',
 	ITALIAN = 'it',
 	PORTUGUESE = 'pt',
+	VIETNAMESE = 'vi',
 	CHINESE_SIMPLIFIED = 'zh-CN'
 }
--- a/src/frontend/src/lib/i18n/vi.json
+++ b/src/frontend/src/lib/i18n/vi.json
@@ -184,13 +184,13 @@
 			"website": "Hình ảnh của $dAppName"
 		},
 		"categories": {
-			"defi": "Tài chính phí tập trung",
+			"defi": "DeFi",
 			"social_media": "Mạng xã hội",
-			"verifiable_credentials": "Thông tin xác thực có thể xác minh",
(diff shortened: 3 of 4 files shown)


src/frontend/src/lib/components/swap/SwapWizard.svelte
```diff
@@ -113,7 +114,7 @@
 				}
 			});
 
-			setTimeout(() => close(), 750);
+			setTimeout(() => close(), 50);
```
AntonioVentilii: why this change?

src/frontend/src/lib/components/swap/SwapWizard.svelte
```diff
@@ -127,23 +128,26 @@
...
-						url: `https://app.icpswap.com/swap?input=${$sourceToken.ledgerCanisterId}&output=${$destinationToken.ledgerCanisterId}`,
-						text: 'icpswap.com'
-					}
-				});
-			} else if (errorDetail && errorDetail.startsWith('Swap failed.')) {
-				failedSwapError.set({ message: errorDetail, variant: 'error' });
+			}
+
+			if (err instanceof SwapError) {
+				if (err.code === 'withdraw_failed') {
```
AntonioVentilii: can we have enums for the codes?

AntonioVentilii (commented):
Some tests? Even for hedge cases

src/frontend/src/lib/components/swap/SwapWizard.svelte
roman-nazaruk: I added this change because we had a long loading spinner. Since we call triggerWallet right after a successful swap, it was causing two spinners to appear briefly.

roman-nazaruk:
> Some tests? Even for hedge cases

TBH, it's almost impossible, since the swap method is called inside a child component and isn't exported. I have no clear way to mock or test it.

src/frontend/src/lib/components/swap/SwapWizard.svelte
AntonioVentilii: then, maybe just `debounce` the call? does it work?

src/frontend/src/tests/lib/utils/swap.utils.spec.ts
```diff
@@ -450,4 +452,25 @@ describe('swap utils', () => {
...
+
+			expect(isSwapError(error)).toBeFalsy();
+		});
+
+		it('should return false for non-error values', () => {
+			expect(isSwapError(null)).toBeFalsy();
+			expect(isSwapError(undefined)).toBeFalsy();
+			expect(isSwapError('string')).toBeFalsy();
+			expect(isSwapError({ code: 'deposit_error' })).toBeFalsy();
+		});
```
AntonioVentilii, line 474: nice!!!

src/frontend/src/lib/components/swap/SwapWizard.svelte
roman-nazaruk: Nice catch, I took a dipper look on this and fix was not related to this.

PR #7949, opened 2025-07-22
refactor(frontend): Specify standard for all ERC contract type

AntonioVentilii:
# Motivation

We are going to add more and more standard of ERC contracts. To help use type-wise, we specify the standard for each single one.


roman-nazaruk (approved):
LGTM, thnx

PR #7957, opened 2025-07-23
chore(test): Remove vitest global imports from setup file

AntonioVentilii:
# Motivation

No need to import the global `vi` in vitest setup file.


roman-nazaruk (approved):
LGTM, thnx

PR #7961, opened 2025-07-23
feat(backend)!: avatar flow, add tests and types.

SYInacta:
# Motivation

This pull request introduces integration tests for the contact image functionality. With the backend now supporting image attachments for contacts, it's crucial to have automated tests to verify that adding, updating, and removing images works correctly and to prevent future regressions.

BREAKING CHANGE: The backend API is now incompatible with production.

# Changes

**New Integration Tests:**  
  Added two new integration tests to `src/backend/tests/it/contacts.rs`:
    - `test_update_contact_image_png_and_remove`:  
      Verifies that a PNG image can be successfully added to a contact and later removed.
    - `test_update_contact_image_jpeg`:  
      Confirms that a JPEG image can be added to a contact.

 **Improved `UpdateContactRequest` Type:**  
   The `UpdateContactRequest` struct in `shared/src/types/contact.rs` included `update_timestamp_ns`, which should only be set by the backend.  
   Removed this field to make the API clearer and prevent misuse.

# Tests
The tests follow these steps:

1. **Create** a new contact.
2. **Call** `update_contact` with image data.
3. **Assert** that the returned contact contains the correct image.
4. (For the removal test) **Call** `update_contact` again with the image set to `None`.
5. **Assert** that the image has been removed.

This ensures the full lifecycle of a contact's image is handled correctly by the canister.


src/backend/src/contacts.rs
```diff
@@ -137,9 +137,11 @@ pub fn update_contact(request: UpdateContactRequest) -> Result<Contact, ContactE
...
             addresses: request.addresses,
             update_timestamp_ns: current_time,
-            image: request.image.or_else(|| existing_contact.image.clone()), /* Use request image
-                                                                              * or preserve
-                                                                              * existing */
+            image: if request.image.is_none() {
+                None
+            } else {
+                request.image.or_else(|| existing_contact.image.clone())
+            },
```
AntonioVentilii: this small part sounds more like a fix: I would prefer to move it to a separate PR, so that we can track the issue

(am I right saying that this fixes the "delete" issue?)

SYInacta: Added separate PR:
https://github.com/dfinity/oisy-wallet/pull/7968

bitdivine (approved):

PR #7964, opened 2025-07-23
feat(devops): Separate Alchemy API key for Test ENVs

AntonioVentilii:
# Motivation

To avoid overworking the Alchemy API key for staging, we decide to use a different one for the test envs.


roman-nazaruk (approved):
LGTM, thnx

PR #7972, opened 2025-07-23
chore(frontend): Add ICP error texts in German

StefanBerger-DFINITY:
# Motivation

We show some ICP specific warnings

# Changes

- add german translations

# Tests


roman-nazaruk (approved):
LGTM, thnx

PR #7973, opened 2025-07-23
feat(frontend): add new trackiing message

CvetelinaAngelova:
# Motivation

Enable avatar updates and deletions to send specific Plausible tracking events.

# Changes

Change handleAddAvatar function to send dinamic messages depends on the event 

# Tests

### DIFF
--- a/src/frontend/src/lib/components/address-book/AddressBookModal.svelte
+++ b/src/frontend/src/lib/components/address-book/AddressBookModal.svelte
@@ -24,7 +24,11 @@
 		TRACK_CONTACT_DELETE_ERROR,
 		TRACK_CONTACT_DELETE_SUCCESS,
 		TRACK_CONTACT_UPDATE_ERROR,
-		TRACK_CONTACT_UPDATE_SUCCESS
+		TRACK_CONTACT_UPDATE_SUCCESS,
+		TRACK_AVATAR_UPDATE_SUCCESS,
+		TRACK_AVATAR_UPDATE_ERROR,
+		TRACK_AVATAR_DELETE_SUCCESS,
+		TRACK_AVATAR_DELETE_ERROR
 	} from '$lib/constants/analytics.contants';
 	import { ADDRESS_BOOK_MODAL } from '$lib/constants/test-ids.constants';
 	import { authIdentity } from '$lib/derived/auth.derived';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/address-book/AddressBookModal.svelte
```diff
@@ -249,10 +253,25 @@
 		if (isNullish(currentContact)) {
 			return;
 		}
-		const contact = {
-			...currentContact
+
+		const contact = { ...currentContact };
+		const isDeleting = isNullish(image);
```
AntonioVentilii, line 259: is this correct? what if the avatar is already absent and the user is changing only the name? or this function is used ONLY for avatar?

src/frontend/src/lib/components/address-book/AddressBookModal.svelte
```diff
@@ -249,10 +253,25 @@
 		if (isNullish(currentContact)) {
```
AntonioVentilii, line 254: some tests for this entire method `handleAddAvatar`? this component became quite huge: ideally we separate the services, and test them independently. Or we test the component, whatever you prefer

src/frontend/src/lib/components/address-book/AddressBookModal.svelte
roman-nazaruk: Just to confirm, does the handleAddAvatar method handle both updating and deleting the avatar?

CvetelinaAngelova: This function will be only for the Avatar. 
callUpdateContact wll stay unchanged and cover name and address change.

CvetelinaAngelova: @roman-nazaruk  
handleAddAvatar should update the contact image with the new one or with null if we want to remove it

src/frontend/src/lib/components/address-book/AddressBookModal.svelte
CvetelinaAngelova: I'll include a test for handleAddAvatar for now as qucker option. But I agree we need to refactor the AddressBookModal and separate the services.

PR #7976, opened 2025-07-23
refactor(frontend): Use `NftId` type in the NFT services

AntonioVentilii:
# Motivation

To be more type-safe, we can use directly `NftId` type in all the NFT services.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7983, opened 2025-07-24
feat(frontend): Util for fungible and not fungible ERC tokens

AntonioVentilii:
# Motivation

We need some utils to distinguish between fungible and not fungible ERC tokens.


roman-nazaruk (approved):
LGTM, thnx

PR #7985, opened 2025-07-24
Feat(frontend) fix update contact

CvetelinaAngelova:
# Motivation

Ensure the contact's image is always updated to the latest version during contact updates.

# Changes

Updated updateContact to always use the latest provided image.
Added fallback to retain the current image if a new one is not provided.

# Tests

Manually tested updating a contact with a new image — confirmed image is updated in the UI and backend.

### DIFF
--- a/src/frontend/src/lib/services/manage-contacts.service.ts
+++ b/src/frontend/src/lib/services/manage-contacts.service.ts
@@ -1,5 +1,4 @@
 import type { ContactImage } from '$declarations/backend/backend.did';
-import { nonNullish } from '@dfinity/utils';
 import {
 	createContact as createContactApi,
 	deleteContact as deleteContactApi,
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/services/manage-contacts.service.ts
```diff
@@ -48,13 +49,15 @@ export const updateContact = async ({
 }): Promise<ContactUi> => {
 	const contactWithSortedAddresses = {
 		...contact,
-		image,
+		image: nonNullish(image) ? image : contact.image,
```
AntonioVentilii: if image is `null`, we should pass it too, no?

DenysKarmazynDFINITY: As far as I can see, it won't be possible to remove an existing image with such an approach. Until the BE is updated, you need to distinguish "image removal" vs "empty image" cases on the frontend: e.g. `undefined` = use fallback value, `null` - user removed the existing avatar so we need to send it to BE.

CvetelinaAngelova: Correct, change it to undefined check

CvetelinaAngelova: yes,that's right-> change it to undefined check

AntonioVentilii (approved):
LGTM tks!

PR #7993, opened 2025-07-24
feat(frontend): Sync ETH balances from cache

AntonioVentilii:
# Motivation

We should sync the ETH balances from the IDB cache, like the other networks.

# Changes

Adapt component `LoaderEthBalances` to sync IDB cache on mounting.

# Tests

Added tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #7995, opened 2025-07-25
feat(frontend): Sync ETH transactions from cache in component `LoaderEthTransactions`

AntonioVentilii:
# Motivation

Similar to what we did for balances, we load transactions for Ethereum/EVM networks from the IDB cache at the first try.

# Changes

- In component `LoaderEthTransactions`, we sync the token transactions from the IDB cache.
- We do not need to do it if the transaction store is already initialized for that specific token.

# Tests

Added tests


roman-nazaruk (approved):
LGTM, thnx

PR #7996, opened 2025-07-25
feat(frontend): add loading to avatar

CvetelinaAngelova:
# Motivation

A lightweight and flexible avatar loading spinner component was needed to indicate loading states in a consistent and accessible way. 

# Changes

Adding new component which could be reused as a LoaderWithOverlay
Implemented a simple SVG-based circular spinner with semantic role and accessibility tags.

# Tests

Manually tested component rendering in various usage contexts.
Confirmed spinner displays and animates as expected.
Adding Unit tests for rendering and style applies.
Screenshot 2025-07-25 at 10 33 59

### DIFF
--- a/src/frontend/src/lib/components/ui/LoaderWithOverlay.svelte
+++ b/src/frontend/src/lib/components/ui/LoaderWithOverlay.svelte
@@ -1,0 +1,24 @@
+<div class="avatar-spinner-overlay" role="status" aria-label="Uploading image">
+	<svg
+		id="app-spinner"
+		preserveAspectRatio="xMidYMid meet"
+		focusable="false"
+		aria-hidden="true"
+		data-tid="spinner"
+		viewBox="0 0 100 100"
+	>
+		<circle cx="50%" cy="50%" r="45" />
+	</svg>
+</div>
+
+<style>
+	.avatar-spinner-overlay {
+		position: absolute;
+		inset: 0;
+		border-radius: 50%;
+		background: rgba(255, 255, 255, 0.4);
+		display: flex;
+		justify-content: center;
+		align-items: center;
+	}
+</style>
--- a/src/frontend/src/lib/components/ui/LoaderWithOverlay.svelte
+++ b/src/frontend/src/lib/components/ui/LoaderWithOverlay.svelte
@@ -1,22 +1,26 @@
 <script lang="ts">
-import { AVATAR_LOADER } from '$lib/constants/test-ids.constants';
-import { i18n } from '$lib/stores/i18n.store';
+	interface Props {
+		testId?: string;
+		ariaLabel?: string;
+		styleClass?: string;
+	}
 
+	const { testId, ariaLabel, styleClass }: Props = $props();
 </script>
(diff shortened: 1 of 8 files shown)


src/frontend/src/lib/components/ui/LoaderWithOverlay.svelte
```diff
@@ -0,0 +1,27 @@
...
+	const { testId, ariaLabel, styleClass }: Props = $props();
+</script>
+
+<div
+	class={`avatar-spinner-overlay ${styleClass ?? ''}`}
+	role="status"
+	data-tid={testId}
+	aria-label={ariaLabel}
+>
+	<svg
```
AntonioVentilii: there is the Spinner component from gix components, i think you can define the size for that one

CvetelinaAngelova: Updated with Gix component

src/frontend/src/lib/i18n/it.json
```diff
@@ -1289,7 +1289,8 @@
 		},
 		"avatar": {
 			"default": "Avatar predefinito",
-			"avatar_for": "Avatar per"
+			"avatar_for": "Avatar per",
+			"avatar_loading": ""
```
AntonioVentilii:
```suggestion
			"avatar_loading": "L'avatar si sta caricando"
```

src/frontend/src/lib/i18n/pt.json
```diff
@@ -1289,7 +1289,8 @@
 		},
 		"avatar": {
 			"default": "Avatar padrão",
-			"avatar_for": "Avatar para"
+			"avatar_for": "Avatar para",
+			"avatar_loading": ""
```
AntonioVentilii:
```suggestion
			"avatar_loading": "O avatar está carregando"
```

src/frontend/src/lib/styles/global/layout.scss
```diff
@@ -3,3 +3,10 @@
 header {
 	margin: 0 auto;
 }
+.avatar-spinner-overlay {
```
AntonioVentilii: let's move this to the specific new component, in the `style` block

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/styles/global/layout.scss
CvetelinaAngelova: I originally considered placing the styles in the component’s <style> block, but it currently throws a warning when running the lint command.

CvetelinaAngelova: Should I create a separate sass file, or move it to a different one ?

AntonioVentilii: i meant directly inside the component `LoaderWithOverlay`

```svelte
<style lang="scss">
...
```

CvetelinaAngelova: moved, thank you

PR #7998, opened 2025-07-25
feat(frontend): Replace standard of token store with page-token store

AntonioVentilii:
# Motivation

In a continuous effort to deprecated the `token` store, we replace the derived `tokenStandard` with the equivalent `pageTokenStandard`.


roman-nazaruk (approved):
LGTM, thnx

PR #7999, opened 2025-07-25
feat(frontend): Replace token store with page-token store in transactions components

AntonioVentilii:
# Motivation

The transactions components are used always inside a page that has a token defined in the URL. So, in a continuous effort to deprecate the `token` store, we can replace it with `pageToken` in all these component.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8001, opened 2025-07-25
feat(frontend): Replace Toggleable prop of token store with page-token store

AntonioVentilii:
# Motivation

In an effort to deprecate the token store, we replace the `tokenToggleable` derived store with the equivalent that is derived from the `pageToken`.

It is ok, since the consumers all refers to places where the page token is defined.

# Changes

- Create copy of `tokenToggleable` store as `pageTokenToggleable`.
- Adapt usages.
- Unrelated: we remove deprecated `tokenStandard` derived store.

# Tests

Added tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8003, opened 2025-07-25
refactor(frontend): Bulk migrate icons to Svelte 5

AntonioVentilii:
# Motivation

The icons components are easy to migrate to Svelte 5, so we do it in bulk in a single PR.


DenysKarmazynDFINITY (approved):
Thanks 👍

PR #8006, opened 2025-07-25
feat(frontend): Migrate `HeroContent` component to Svelte 5

AntonioVentilii:
# Motivation

Migrating `HeroContent` component to Svelte 5.

### DIFF
--- a/src/frontend/src/lib/components/hero/Actions.svelte
+++ b/src/frontend/src/lib/components/hero/Actions.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { nonNullish } from '@dfinity/utils';
-	import { page } from '$app/stores';
+	import { page } from '$app/state';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/hero/HeroContent.svelte
```diff
@@ -45,14 +45,15 @@
...
 
-	let pageTokenUi: OptionTokenUi;
-	$: pageTokenUi = nonNullish($pageToken)
-		? mapTokenUi({
-				token: $pageToken,
-				$balances: $balancesStore,
-				$exchanges
-			})
-		: undefined;
+	let pageTokenUi: OptionTokenUi = $derived(
```
DenysKarmazynDFINITY: I think we don't need the type with the $derived

AntonioVentilii: right! removed

PR #8012, opened 2025-07-25
chore(frontend): assert signer usage of new cbor library

peterpeterparker:
# Motivation

We must no necessarely introduce this change in OISY but, I would like to deploy it on staging to proceed with some manually testing of the signer. Therefore opening this PR to trigger a deploy.

# Changes

- Update OISY Wallet Signer next which uses the new `@dfinity/cbor` library.


peterpeterparker:
We deployed on staging and tests we successful. We'll release a version of the signer (https://github.com/dfinity/oisy-wallet-signer/pull/628) but, we don't have to rush to integrate it.

PR #8018, opened 2025-07-25
feat(frontend): Migrate component `InfoBoxWrapper` to Svelte 5

AntonioVentilii:
# Motivation

Migrating InfoBoxWrapper component (and its direct children) to Svelte 5.

### DIFF
--- a/src/frontend/src/lib/components/info/InfoBox.svelte
+++ b/src/frontend/src/lib/components/info/InfoBox.svelte
@@ -1,10 +1,17 @@
 <script lang="ts">
 	import { IconClose } from '@dfinity/gix-components';
+	import type { Snippet } from 'svelte';
 	import { slide } from 'svelte/transition';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/info/InfoBox.svelte
```diff
@@ -1,20 +1,29 @@
...
+	interface Props {
+		hideInfo: boolean;
+		children: Snippet;
+		onClick: () => void;
+	}
+
+	let { hideInfo, children, onClick }: Props = $props();
 </script>
 
 {#if !hideInfo}
```
CvetelinaAngelova, line 17: maybe isNullish

AntonioVentilii: Just checked: hideInfo is boolean (never nullish), so it is ok to use it like this

PR #8031, opened 2025-07-28
feat(frontend): Track network in ETH send flow

AntonioVentilii:
# Motivation

It is useful to track which network is being used in the ETH send flow.

### DIFF
--- a/src/frontend/src/eth/components/send/EthSendTokenWizard.svelte
+++ b/src/frontend/src/eth/components/send/EthSendTokenWizard.svelte
@@ -184,7 +184,8 @@
 			trackEvent({
 				name: TRACK_COUNT_ETH_SEND_SUCCESS,
 				metadata: {
-					token: $sendToken.symbol
+					token: $sendToken.symbol,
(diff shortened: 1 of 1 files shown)


src/frontend/src/eth/components/send/EthSendTokenWizard.svelte
```diff
@@ -184,7 +184,8 @@
 			trackEvent({
 				name: TRACK_COUNT_ETH_SEND_SUCCESS,
 				metadata: {
-					token: $sendToken.symbol
+					token: $sendToken.symbol,
+					network: sourceNetwork.id.description ?? `${$sendToken.network.id.description}`
```
DenysKarmazynDFINITY, line 188: the second value also doesn't need to be casted to string, does it?

AntonioVentilii: i receive an error, since type-wise `networkId.description` could be `undefined`, and the metadata accepts non-nullish values.

DenysKarmazynDFINITY: ok, makes sense then

PR #8032, opened 2025-07-28
feat(frontend): Track BTC send flow events

AntonioVentilii:
# Motivation

We would like to understand how much is the Bitcoin send flow used.

### DIFF
--- a/src/frontend/src/sol/api/solana.api.ts
+++ b/src/frontend/src/sol/api/solana.api.ts
@@ -197,6 +197,18 @@ export const estimatePriorityFee = async ({
 	);
 };
 
+export const getAccountInfo = async ({
+	address,
+	network
+}: {
+	address: SolAddress;
+	network: SolanaNetworkType;
+}) => {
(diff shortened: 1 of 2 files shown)


src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
```diff
@@ -121,10 +126,26 @@
 				}
 			});
 
+			trackEvent({
+				name: TRACK_COUNT_BTC_SEND_SUCCESS,
+				metadata: {
+					token: $sendToken.symbol,
+					network: `${networkId.description}`
```
DenysKarmazynDFINITY, line 133: same as in another PR, I think networkId.description should be enough

AntonioVentilii: i receive an error, since type-wise `networkId.description` could be `undefined`, and the metadata accepts non-nullish values.

PR #8034, opened 2025-07-28
feat(frontend): Cache Solana RPC `getAccountInfo` calls

AntonioVentilii:
# Motivation

To reduce the amount of calls to the Solana RPC, we can try and cache-in-code the info that we receive per address from the call to `getAccountInfo`.

# Changes

- Create record to cache the account info per address and network, directly in the code.
- Use the record in the service `getAccountInfo`.

# Tests

Added tests.


roman-nazaruk (approved):
LGTM, thnx

PR #8035, opened 2025-07-28
feat(frontend): add lock button to the menu

CvetelinaAngelova:
# Motivation

Introduce a lock button into the menu to improve user control over session locking and logout behavior. This change provides a more accessible UI for secure session management, aligning with recent UX improvements.

# Changes

Added a new Lock button adjacent to the logout button within the menu layout.
Introduced MenuLockLogOut.svelte which wraps both buttons in a flex container to ensure equal width and spacing.

# Tests
Screenshot 2025-07-28 at 21 29 17

Manual test the logOut works properly 
Add Unit test to the menu that LogOut button is rendered

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapWizard.svelte
+++ b/src/frontend/src/lib/components/swap/SwapWizard.svelte
@@ -101,7 +101,15 @@
 				receiveAmount: $swapAmountsStore.selectedProvider.receiveAmount,
 				slippageValue,
 				sourceTokenFee,
-				isSourceTokenIcrc2: $isSourceTokenIcrc2
+				isSourceTokenIcrc2: $isSourceTokenIcrc2,
+				tryToWithdraw:
+					nonNullish($failedSwapError?.errorType) &&
+					($failedSwapError?.errorType === SwapErrorCodes.SWAP_FAILED_WITHDRAW_FAILED ||
+						$failedSwapError?.errorType === SwapErrorCodes.SWAP_SUCCESS_WITHDRAW_FAILED ||
+						$failedSwapError?.errorType === SwapErrorCodes.ICP_SWAP_WITHDRAW_FAILED),
+				withdrawDestinationTokens:
+					nonNullish($failedSwapError?.errorType) &&
+					$failedSwapError?.errorType === SwapErrorCodes.SWAP_SUCCESS_WITHDRAW_FAILED
 			});
 
 			progress(ProgressStepsSwap.DONE);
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/i18n/it.json
```diff
@@ -102,6 +102,7 @@
 			"title_part_1": "$oisy_name",
 			"title_part_2": "100% Decentralizzato",
 			"logout": "Logout",
+			"lock": "",
```
AntonioVentilii:
```suggestion
			"lock": "Blocca",
```

src/frontend/src/lib/i18n/pt.json
```diff
@@ -102,6 +102,7 @@
 			"title_part_1": "$oisy_name",
 			"title_part_2": "100% Descentralizado",
 			"logout": "Sair",
+			"lock": "",
```
AntonioVentilii:
```suggestion
			"lock": "Bloquear",
```

src/frontend/src/lib/components/icons/IconLock.svelte
```diff
@@ -0,0 +1,30 @@
+<script lang="ts">
```
AntonioVentilii, line 2:
```suggestion
<!-- source: DFINITY foundation -->
<script lang="ts">
```

src/frontend/src/lib/components/core/MenuLockLogOut.svelte
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	let { onHidePopover, hideText = true }: Props = $props();
+
+	const dispatch = createEventDispatcher();
+	const remainingTimeMilliseconds = $derived($authRemainingTimeStore);
+
+	const handleLogoutTriggered = () => {
+		dispatch('icLogoutTriggered');
+	};
```
AntonioVentilii: since we are using Svelte 5, let's remove dispatch events

AntonioVentilii (commented):
I understand the idea. My suggestion is that you first extract a `SignOutButton` (with its logout logic) from the `SignOut` component, so that it can be easy reusable. And a `RemaningTimeToLogout` component (or something similar) and reuse it in both `SignOut` and `MenuLockSignOut`.

P.S. shall we rename `MenuLockSignOut` to `LockOrSignOut` for consistency?

AntonioVentilii (approved):

src/frontend/src/lib/components/core/MenuLockLogOut.svelte
CvetelinaAngelova: This is copy paste logic from the old component but I will rewrite it in the new one

PR #8043, opened 2025-07-29
fix(frontend): Avoid tracking change of language every reload

AntonioVentilii:
# Motivation

We are tracking the change language event every reload because we call the tracker on initialization of `i18n` store.

We should not track it when it is in initialization.


roman-nazaruk (approved):
LGTM,thnx

PR #8057, opened 2025-07-29
fix(frontend): Do not listen to ETH transaction status if already finalised

AntonioVentilii:
# Motivation

Component `EthTransactionStatus` listens real-time to the block numbers, to verify the current status of the open transaction inside a Transaction modal. However, it is missing the logic that if the status is already finalised, it should not keep listening.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8060, opened 2025-07-30
fix(frontend): Turn listener off for Alchemy websocket providers

AntonioVentilii:
# Motivation

Alchemy websockets are buggy. So, we force one more layer of "destroying" for the websocket providers.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8064, opened 2025-07-30
feat(frontend): Extract service to fetch URI JSON for ERC721

AntonioVentilii:
# Motivation

In this PR we extract a service from `InfuraErc721Provider` class: the one responsible to fetch the JSON file linked to a token URI. It will be useful since we will re-use it for ERC1155.

# Changes

- Create types for ERC721 and ERC1155 JSON interfaces.
- Extract service `fetchMetadataFromUri` that is responsible to fetch the metadata JSON from a token URI and parsing the image URL correctly too.

# Tests

Added tests.


roman-nazaruk (approved):
LGTM, thnx

PR #8065, opened 2025-07-30
feat(frontend): Pass ICRC metadata for new Custom tokens

AntonioVentilii:
# Motivation

Similar to what we do with ERC and SPL tokens, we want to persist the metadata for ICRC custom tokens, since they are already being fetched.


src/frontend/src/icp/components/tokens/IcAddTokenReview.svelte
```diff
@@ -7,6 +7,7 @@
 		loadAndAssertAddCustomToken,
 		type ValidateTokenData
 	} from '$icp/services/ic-add-custom-tokens.service';
+	import type { IcTokenWithoutId } from '$icp/types/ic-token';
```
AntonioVentilii:
```suggestion
```

PR #8067, opened 2025-07-30
feat(frontend): Method to fetch metadata from ERC1155 URI

AntonioVentilii:
# Motivation

Similar to ERC721, we create a method to fetch the NFT metadata from the URI in class `InfuraErc1155Provider`.

# Changes

- Create method `getNftMetadata` that is basically a mirror of the same method in class `InfuraErc721Provider`.
- Adapt the attributes to consider the different properties of an ERC1155 interface.

# Tests

Included tests.


roman-nazaruk (approved):
LGTM, thnx

PR #8070, opened 2025-07-30
feat(frontend): added new step failed to progress Steps

roman-nazaruk:
# Motivation

To support displaying a failed step in the ProgressSteps component (e.g., when a step fails in the swap flow), we need to adjust the logic to check whether a specific step is marked as failed.

# Changes

- Added a new failedSteps prop that can be updated from the parent component
- Added a condition to identify and highlight failed steps
- Adjusted styles for failed step


AntonioVentilii (approved):
LGTM tks!

PR #8073, opened 2025-07-30
feat(frontend): adjusted swapError to have optional message and variant

roman-nazaruk:
# Motivation

The ICP Swap flow needs to raise SwapError without a message in some cases and with a severity level in others. To support this, SwapError should accept an optional message and a new variant field.

# Changes

- Made message optional.
- Added variant?: 'error' | 'warning' | 'info'.

# Tests

Covered new logix with the tests

### DIFF
--- a/src/frontend/src/lib/services/swap-errors.services.ts
+++ b/src/frontend/src/lib/services/swap-errors.services.ts
@@ -1,21 +1,27 @@
-import type { SwapErrorKey } from '$lib/types/swap';
+import type { SwapErrorCodes } from '$lib/types/swap';
 
 export const throwSwapError = ({
 	code,
-	message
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/services/swap-errors.services.ts
```diff
@@ -1,21 +1,27 @@
-import type { SwapErrorKey } from '$lib/types/swap';
+import type { SwapErrorCodes } from '$lib/types/swap';
 
 export const throwSwapError = ({
 	code,
-	message
+	message = undefined,
+	variant = undefined
```
AntonioVentilii:
```suggestion
	message,
	variant
```

PR #8076, opened 2025-07-30
feat(frontend): Map BTC self-transactions

AntonioVentilii:
# Motivation

Among our TODOs, there is the feature to map self-transactions in Bitcoin. This has not much application since it is quite expensive to send oneself BTC just to test. But in any case, it is useful for mapping everything.

# Changes

- Modify `mapBtcTransaction` util to consider a self-transaction if the input and ALL the outputs are the same address.
- Add some comments about the steps.

# Tests

Added a test, plus a practical one:

### Before

Screenshot 2025-07-30 at 15 15 59

### After

Screenshot 2025-07-30 at 16 49 27


DenysKarmazynDFINITY (approved):
Thanks!

PR #8078, opened 2025-07-30
feat(frontend): added new withdraw methods for icpSwap

roman-nazaruk:
# Motivation

Added two new methods to support manual and automatic withdrawal flows for ICP Swap.

# Changes

- Implemented withdrawICPSwapAfterFailedSwap for automatic retry-based withdrawal after a failed swap
- Implemented performManualWithdraw for manual user-triggered withdrawal

# Tests

Unit tests added for both methods

### DIFF
--- a/src/frontend/src/lib/services/swap.services.ts
+++ b/src/frontend/src/lib/services/swap.services.ts
@@ -3,6 +3,7 @@ import { setCustomToken as setCustomIcrcToken } from '$icp-eth/services/custom-t
 import { approve } from '$icp/api/icrc-ledger.api';
 import { sendIcp, sendIcrc } from '$icp/services/ic-send.services';
 import { loadCustomTokens } from '$icp/services/icrc.services';
+import type { IcTokenToggleable } from '$icp/types/ic-token-toggleable';
 import { nowInBigIntNanoSeconds } from '$icp/utils/date.utils';
 import { isTokenIcrc } from '$icp/utils/icrc.utils';
 import { setCustomToken } from '$lib/api/backend.api';
@@ -18,6 +19,7 @@ import {
 	kongSwapTokensStore,
 	type KongSwapTokensStoreData
 } from '$lib/stores/kong-swap-tokens.store';
+import type { OptionIdentity } from '$lib/types/identity';
 import {
 	SwapErrorCodes,
 	SwapProvider,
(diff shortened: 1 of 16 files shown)


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -338,3 +341,114 @@ export const swapService = {
...
 };
+
+export const withdrawICPSwapAfterFailedSwap = async ({
+	identity,
+	canisterId,
+	token,
+	amount,
+	fee,
+	setFailedProgressStep
+}: {
```
DenysKarmazynDFINITY: both withdrawICPSwapAfterFailedSwap and performManualWithdraw share most of the params - let's create a type for that and re-use it

DenysKarmazynDFINITY: this params interface can potentially extend `ICPSwapDepositWithdrawParams`

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -338,3 +341,114 @@ export const swapService = {
...
+	identity: OptionIdentity;
+	canisterId: string;
+	token: string;
+	amount: bigint;
+	fee: bigint;
+	setFailedProgressStep?: (step: ProgressStepsSwap) => void;
+}): Promise<{ code: SwapErrorCodes; message?: string; variant?: 'error' | 'warning' | 'info' }> => {
```
DenysKarmazynDFINITY: same with the return type - let's re-use it

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -338,3 +341,114 @@ export const swapService = {
...
+	try {
+		await withdraw({
```
DenysKarmazynDFINITY: let's create a const for the `withdraw` params and re-use it in both calls

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Good point. Updated, thnx

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Updated, thnx

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Updated, thnx

PR #8079, opened 2025-07-30
feat(frontend): added new tryToWithdraw condition in swap icpSwap

roman-nazaruk:
# Motivation

Added a new tryToWithdraw condition to handle manual withdrawal using performManualWithdraw when swapping with ICP Swap.

# Changes

Introduced tryToWithdraw condition for ICP Swap flow


roman-nazaruk:
> not?

Yes, this properties would come if we would call swapService if the firsr time it was failed. This property is not very important here, and I would create another PR to this service, where it would make more sense. 

> My understanding is that `tryToWithdraw` and `withdrawDestinationTokens` are provided by something that is absolutely not linked to this services, otherwise we could have them directly here.
> 
> For example, if you have `withdrawDestinationTokens` as `true`, means that the swap failed, so theoretically, you could do `setFailedProgressStep?.(ProgressStepsSwap.SWAP)` in the place where the swap failed. Or not?

Yeah, you're right. I'll create a separate PR for setFailedProgressStep, where it will make more sense. I've removed setFailedProgressStep from this PR. And yes, tryToWithdraw and withdrawDestinationTokens will be provided via params.

PR #8095, opened 2025-07-31
chore(frontend): Add xStock ABBVx Solana token

StefanBerger-DFINITY:
# Motivation

Tokenized shares from https://xstocks.com/products

# Changes

- enable token ABBVx

**Note:** The token was already added earlier, just not yet enabled because we had no price for it yet.

# Tests
image

### DIFF
--- a/src/frontend/src/btc/components/receive/BtcReceive.svelte
+++ b/src/frontend/src/btc/components/receive/BtcReceive.svelte
@@ -13,28 +13,27 @@
 	import {
 		btcAddressMainnetStore,
 		btcAddressRegtestStore,
-		btcAddressTestnetStore,
-		type AddressStoreData
+		btcAddressTestnetStore
 	} from '$lib/stores/address.store';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { modalStore } from '$lib/stores/modal.store';
-	import type { BtcAddress } from '$lib/types/address';
-	import type { Token } from '$lib/types/token';
 	import { isNetworkIdBTCRegtest, isNetworkIdBTCTestnet } from '$lib/utils/network.utils';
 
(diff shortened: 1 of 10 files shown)


src/frontend/src/env/tokens/tokens.spl.env.ts
```diff
@@ -44,9 +45,7 @@ import type { RequiredSplToken } from '$sol/types/spl';
 
 const SPL_TOKENS_MAINNET: RequiredSplToken[] = [
 	AAPLX_TOKEN,
-	// TODO: temporarily disable these 2 tokens, because we don't have a price for them yet. We can enable them once they get traded and habe a price.
-	// ABBVX_TOKEN,
-	// ABTX_TOKEN,
+	ABBVX_TOKEN,
```
AntonioVentilii, line 48: shall we leave `ABTX_TOKEN` as commented out? or will it be removed?

StefanBerger-DFINITY: I removed the comment, because I think it's risky to have it there, but when uncommenting, you also need to add it above in the imports...
So I'd rather not have it here anymore. I will add it as soon as ABTX has a price on coinbase

AntonioVentilii (approved):

src/frontend/src/env/tokens/tokens.spl.env.ts
StefanBerger-DFINITY: image

soon.....

StefanBerger-DFINITY: ABTx is now in: https://github.com/dfinity/oisy-wallet/pull/8221

PR #8113, opened 2025-07-31
feat(frontend): adjusted logic of icpSwap to set swapSucced prop

roman-nazaruk:
# Motivation

For icpSwap, we need to track whether the swap was successful in order to determine whether to retrieve the source token or the destination token. This status is now stored.

# Changes

Added logic to store swapSucceded status

### DIFF
--- a/src/frontend/src/lib/services/swap.services.ts
+++ b/src/frontend/src/lib/services/swap.services.ts
@@ -240,7 +240,7 @@ export const fetchIcpSwap = async ({
 
 		progress(ProgressStepsSwap.WITHDRAW);
 
-		const { code, message, variant } = await performManualWithdraw({
+		const { code, message, variant, swapSucceded } = await performManualWithdraw({
 			withdrawDestinationTokens,
 			setFailedProgressStep,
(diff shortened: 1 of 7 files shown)


AntonioVentilii (approved):
A bit strange that the error has a "succeded" property. Let's add a TODO to revisit this part later. 

Approved in the meanwhile. Tks!

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -462,15 +464,16 @@ export const performManualWithdraw = async ({
...
 			}
 		});
 
 		setFailedProgressStep?.(ProgressStepsSwap.WITHDRAW);
 
 		return {
 			code: SwapErrorCodes.ICP_SWAP_WITHDRAW_FAILED,
-			variant: 'error'
+			variant: 'error',
+			swapSucceded: withdrawDestinationTokens ? true : false
```
DenysKarmazynDFINITY:
```suggestion
			swapSucceded: withdrawDestinationTokens
```

PR #8117, opened 2025-08-02
fix(devops): Correct import of env variable if there is no secret

AntonioVentilii:
# Motivation

When an env variable is included in the CI workflow that deploys the frontend, it is being fetched from GitHub secrets. However, if it is not set there, defaults to an empty string, raising automatically an error during building when parsing with JSON.

This is still acceptable, since it would be inconsistent to tell the deployment workflow to use a variable that we have not set. If we don't want to set it, we simply can remove it from the workflow and it will default to `false` during building time.

However, it is useful to know why the building fails during the CI. So, in this PR we make the env variable parser more verbose.

# Changes

- Throw an error in `parseBoolEnvVar` when the input is an empty string. Explain the reason why it would fail and what to do to solve it.

# Tests

I added a mock env variable in this PR, adding it to all the CI workflows and the frontend, but without setting it among the GItHub secrets. As expected, the deployment failed (see https://github.com/dfinity/oisy-wallet/actions/runs/16690657397/job/47247923104), as usual. But this time it provided more info:

Screenshot 2025-08-02 at 08 16 47


src/frontend/src/lib/utils/env.utils.ts
```diff
@@ -4,8 +4,17 @@ import type { OptionString } from '$lib/types/string';
 // Example: `parseBoolEnvVar(import.meta.env.ENV_VARIABLE)` is type-accepted even if it should be `parseBoolEnvVar({ value: import.meta.env.ENV_VARIABLE })`
 // To avoid possible issues, we decided to give the parameters as array of objects and not as single object for this specific util.
 // eslint-disable-next-line local-rules/prefer-object-params
-export const parseBoolEnvVar = (value: OptionString, check = true): boolean =>
-	JSON.parse(value?.toLowerCase() ?? 'false') === check;
+export const parseBoolEnvVar = (value: OptionString, check = true): boolean => {
+	const normalised = value?.toLowerCase();
+	if (normalised === '') {
```
AntonioVentilii, line 9: i am not using `isEmptyString` here, since it would be `true` for `null` and `undefined`, while i need `true` ONLY for empty string.

DenysKarmazynDFINITY (approved):

PR #8122, opened 2025-08-04
feat(frontend): displays error message on unrecognised token standard

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/components/manage/ManageTokensModal.svelte
+++ b/src/frontend/src/lib/components/manage/ManageTokensModal.svelte
@@ -177,8 +177,10 @@
 			}
 		});
 
-		// TODO: Warn the user that if no interface is encountered, it falls back to standard ERC721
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/manage/ManageTokensModal.svelte
```diff
@@ -177,8 +177,10 @@
 			}
 		});
 
-		// TODO: Warn the user that if no interface is encountered, it falls back to standard ERC721
-		await saveErc721([newToken]);
+		// In case we are not able to determine the token standard, we display an error message
+		toastsError({
```
DenysKarmazynDFINITY, line 181: we only want to show the toast without closing the modal, right?

BonomoAlessandro: Yes, i think it is better to keep the modal open.

PR #8123, opened 2025-08-04
refactor(frontend): removes unused semicolon from abi definition

BonomoAlessandro:
# Motivation

When importing an erc1155 contract there is an warning displayed in the browsers console because there is an unexpected `;` in the abi defintion. To prevent this, we remove it from the definition.

# Changes

- removes unused `;` from abi


DenysKarmazynDFINITY (approved):
Thanks!

PR #8137, opened 2025-08-04
remove breaking the value for the token into the modal

CvetelinaAngelova:
# Motivation

The token selector UI currently breaks the token amount display when large token values are entered, causing visual misalignment.

# Changes

Remove break-all class that will make the text goes on the 2 lines

# Tests

Verified layout with various token name lengths and large numeric values
Checked responsive behavior on different screen sizes

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokenBalance.svelte
+++ b/src/frontend/src/lib/components/tokens/TokenBalance.svelte
@@ -21,7 +21,7 @@
 </script>
 
 <TokenBalanceSkeleton {data}>
-	<output class="break-all" data-tid={testId}>
+	<output data-tid={testId}>
 		{#if nonNullish(data.balance)}
 			{#if hideBalance}
 				{@render privacyBalance?.()}


src/frontend/src/lib/components/tokens/ModalTokensListItem.svelte
```diff
@@ -43,7 +43,9 @@
 	{/snippet}
 
 	{#snippet titleEnd()}
-		<TokenBalance data={token} />
+		<span class="block min-w-12 text-nowrap">
```
DenysKarmazynDFINITY: please use `div` - this way you won't need to manually set `block`

DenysKarmazynDFINITY:
@CvetelinaAngelova can you please attach before/after screenshot? Just to understand better what was the goal of the change.

src/frontend/src/lib/components/tokens/ModalTokensListItem.svelte
CvetelinaAngelova: Chnaged

src/frontend/src/lib/components/tokens/ModalTokensListItem.svelte
```diff
@@ -43,7 +43,9 @@
 	{/snippet}
 
 	{#snippet titleEnd()}
-		<TokenBalance data={token} />
+		<div class="min-w-12 text-nowrap">
```
DenysKarmazynDFINITY: Just checked your screenshot - let's add `ml-1` just to set some minimal space between the balance and the labels.

CvetelinaAngelova: add it

PR #8138, opened 2025-08-04
feat(frontend): extended address contacts derived

DenysKarmazynDFINITY:
# Motivation

Since the original contact address schema does not include any unique identifiers, we need to create an extended store where we manually add `id` to each address. This is needed for the AI assistant feature - we will be hiding the address values before feeding the data to LLM, and then parsing the response back by using these generated ids.

### DIFF
--- a/src/frontend/src/lib/derived/contacts.derived.ts
+++ b/src/frontend/src/lib/derived/contacts.derived.ts
@@ -1,5 +1,5 @@
 import { contactsStore } from '$lib/stores/contacts.store';
-import type { ContactUi } from '$lib/types/contact';
+import type { ContactUi, ExtendedAddressContactUiMap } from '$lib/types/contact';
 import { nonNullish } from '@dfinity/utils';
 import { derived, type Readable } from 'svelte/store';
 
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/derived/contacts.derived.ts
```diff
@@ -19,3 +19,22 @@ export const contacts: Readable<ContactUi[]> = derived([contactsStore], ([$conta
...
+	([$contacts]) =>
+		$contacts.reduce<ExtendedAddressContactUiMap>(
+			(acc, { addresses, id, ...rest }) => ({
+				...acc,
+				[`${id}`]: {
+					...rest,
+					id,
+					addresses: addresses.map((address) => ({
+						...address,
+						id: crypto.randomUUID().toString()
```
roman-nazaruk, line 34: I’m wondering about using crypto.randomUUID() here, is it safe in this context?

DenysKarmazynDFINITY: Yes, the method should be available in most of the modern browsers. Also, it's extensively used across the foundation repos (especially, in II).

roman-nazaruk: Great, thnx

PR #8146, opened 2025-08-05
feat(frontend): implements get nft ids for owner function

BonomoAlessandro:
# Motivation

We want to be able to load the nfts that are owned by a specific user of a specific `erc1155` contract. 

# Changes

- implements alchemy endpoint


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8147, opened 2025-08-05
feat(frontend): extend AI types and components with tools response

DenysKarmazynDFINITY:
# Motivation

As the next step of preparing our code for handling LLM tools, we need to extend the types and UI components with the `tool` field. The services that request and parse tool responses will be added in the next PRs.


roman-nazaruk (approved):
LGTM, thnx

PR #8149, opened 2025-08-05
feat(frontend): triggers reloading of nfts also for erc1155 contracts

BonomoAlessandro:
# Motivation

The reloading of the nfts should also be retriggered if the `erc1155` tokens changes.

### DIFF
--- a/src/frontend/src/lib/components/loaders/Loader.svelte
+++ b/src/frontend/src/lib/components/loaders/Loader.svelte
@@ -174,7 +174,11 @@
 	});
 
 	$effect(() => {
-		if (($erc721CustomTokensInitialized || $erc1155CustomTokensInitialized) && nonNullish($ethAddress) && $nonFungibleTokens.length > 0) {
+		if (
+			($erc721CustomTokensInitialized || $erc1155CustomTokensInitialized) &&
+			nonNullish($ethAddress) &&
+			$nonFungibleTokens.length > 0
+		) {
 			debounceLoadNfts();
 		}
 	});


src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -165,14 +167,19 @@
...
 		});
 	});
 
 	$effect(() => {
-		if ($erc721CustomTokensInitialized && nonNullish($ethAddress) && $erc721Tokens) {
+		if (
+			$erc721CustomTokensInitialized &&
+			$erc1155CustomTokensInitialized &&
+			nonNullish($ethAddress) &&
+			$nonFungibleTokens
```
DenysKarmazynDFINITY: maybe $nonFungibleTokens.length > 0?

src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -165,14 +167,19 @@
...
 			loadedNfts: $nftStore ?? [],
 			walletAddress: $ethAddress
 		});
 	});
 
 	$effect(() => {
-		if ($erc721CustomTokensInitialized && nonNullish($ethAddress) && $erc721Tokens) {
+		if (
+			$erc721CustomTokensInitialized &&
+			$erc1155CustomTokensInitialized &&
```
DenysKarmazynDFINITY: would it be better to start loading one if another is not initialized yet? or is it necessary to wait for both to start loading?

BonomoAlessandro: You are right, it is better to start as soon as possible.

src/frontend/src/lib/components/loaders/Loader.svelte
BonomoAlessandro: I don't think this is needed, because the `loadNfts` function can handle empty lists, but yes it might be more efficient

PR #8153, opened 2025-08-05
feat(frontend): added withdraw unused tokens from icpSwap pool

roman-nazaruk:
# Motivation

We encountered an issue with ICP Swap where, after a swap operation, a portion of both the sourceToken and destinationToken may remain in the pool under the user's balance. This PR introduces functionality to detect and withdraw any such unused token balances left in the pool.

# Changes

- Added `withdrawUserUnusedBalance `utility for checking and withdrawing unused user pool balances.
- Updated SwapParams type definitions to support new logic.
- Integrated withdrawUserUnusedBalance to handle manual withdrawal requests and after a successful swap.

# Tests

Covered new logic with tests


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -477,3 +472,39 @@ export const performManualWithdraw = async ({
...
+	IcpSwapManualWithdrawParams,
+	'setFailedProgressStep' | 'withdrawDestinationTokens'
+>): Promise<void> => {
+	const { balance0, balance1 } = await getUserUnusedBalance({
+		identity,
+		canisterId,
+		principal: identity.getPrincipal()
+	});
+
+	if (balance0 !== 0n) {
```
DenysKarmazynDFINITY:
```suggestion
	if (balance0 !== ZERO) {
```

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -477,3 +472,39 @@ export const performManualWithdraw = async ({
...
+		await withdraw({
+			identity,
+			canisterId,
+			token: destinationToken.ledgerCanisterId,
+			amount: balance0,
+			fee: destinationToken.fee
+		});
+	}
+
+	if (balance1 !== 0n) {
```
DenysKarmazynDFINITY:
```suggestion
	if (balance1 !== ZERO) {
```

DenysKarmazynDFINITY (approved):
LGTM, thanks

P.S.: in the test files, you can also ZERO :)

PR #8161, opened 2025-08-06
feat(frontend): displays nft balances

BonomoAlessandro:
# Motivation

Nfts of erc1155 contracts do have a balance which means we also want to display this balance for the nfts.

# Changes

- displays balances

# Tests

**before:**
image

**after:**
image


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8164, opened 2025-08-06
refactor(frontend): introduces collection schema

BonomoAlessandro:
# Motivation

Soon we will work with nft collections which means we need a collection schema.

# Changes

- introduces collection schema


roman-nazaruk (approved):
LGTM, thnx

PR #8165, opened 2025-08-06
feat(frontend): added loading state for swap not offered and added skeletons

roman-nazaruk:
# Motivation

This PR introduces loading states to improve user feedback. It covers scenarios such as when token fee data is being fetched for the Max button and when swap availability is still being determined.

# Changes

- Added a new showSwapNotOfferedError derived value to conditionally display swap errors
- Added SkeletonText placeholders for loading states in relevant components

### DIFF
--- a/src/frontend/src/lib/schema/nft.schema.ts
+++ b/src/frontend/src/lib/schema/nft.schema.ts
@@ -1,3 +1,5 @@
+import type { Network } from '$lib/types/network';
+import type { TokenId, TokenStandard } from '$lib/types/token';
 import * as z from 'zod';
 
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/swap/SwapForm.svelte
```diff
@@ -81,6 +84,14 @@
 			: false
 	);
 
+	let showSwapNotOfferedError = $derived(
```
BonomoAlessandro, line 87: This should be `const` i guess

BonomoAlessandro (approved):
LGTM

DenysKarmazynDFINITY (approved):

PR #8166, opened 2025-08-06
feat(frontend): added new usdsourceValue to plausible event for swap

roman-nazaruk:
# Motivation

To better understand user behavior and improve analytics, this PR adds a new usdSourceValue property to the swap tracking event.

# Changes 

Added usdSourceValue to the trackEvent payload for both successful and failed swap events

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapWizard.svelte
+++ b/src/frontend/src/lib/components/swap/SwapWizard.svelte
@@ -16,11 +16,14 @@
 		TRACK_COUNT_SWAP_SUCCESS
 	} from '$lib/constants/analytics.contants';
 	import { authIdentity } from '$lib/derived/auth.derived';
+	import { Currency } from '$lib/enums/currency';
+	import { Languages } from '$lib/enums/languages';
 	import { ProgressStepsSwap } from '$lib/enums/progress-steps';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/swap/SwapWizard.svelte
```diff
@@ -66,6 +75,17 @@
 		}
 	};
 
+	let sourceTokenUsdValue = $derived(
```
BonomoAlessandro, line 80: This should be `const` i guess

BonomoAlessandro (approved):
LGTM

src/frontend/src/lib/components/swap/SwapWizard.svelte
DenysKarmazynDFINITY: don't we want to use here currently selected currency?

roman-nazaruk: I believe we should always use the USD value, as it makes analyzing data much easier and more consistent on our side.

DenysKarmazynDFINITY: ah, I missed the "plausible" part, thought it's meant to be shown in UI :)

DenysKarmazynDFINITY (approved):

PR #8169, opened 2025-08-06
feat(frontend): implement askLlmToFilterContacts method

DenysKarmazynDFINITY:
# Motivation

We need to implement a method that is going to request a semantical contacts filter from LLM. It filters out all sensitive contacts info (addresses) before sending it to AI. After LLM replies, we parse the response to ContactUi[].


roman-nazaruk (approved):
LGTM, thnx

PR #8172, opened 2025-08-06
chore(frontend): fix main.js loader paths

peterpeterparker:
# Motivation

> TL;DR: we got lucky

I implemented a 404 page in the Juno Console. One thing led to another, and I ended up having to hash the filenames of `main.js` loaders (because 404.html and index.html both live in the same folder, at the root). That’s when I noticed an issue on mainnet: the loaders were no longer being loaded, and the frontend was crashing!

The root cause was that the loader was always being fetched from the root directory. This wouldn’t normally be an issue—*if* all loaders had the same content and were named `main.js` but, since I started hashing the filenames… yolo.

Long story short, I fixed the issue by using a relative path for the loaders (in this [PR](https://github.com/junobuild/juno/pull/1826)). While doing that, I realized the same issue exists in OISY. Fortunately, because all the loaders are currently identical, it hasn’t caused any problems yet. Still, for future-proofing, it’s probably best to address it now.

# Notes

I double-checked NNS dapp. It’s fine there, thanks to a PR from @bitdivine back in February 2023.

# Changes

- In `build.csk.mjs`, when `injectScriptLoader`, calculate a parent relative path and use it to load the `main.js` script loader from the related HTML page.

### DIFF
--- a/scripts/build.csp.mjs
+++ b/scripts/build.csp.mjs
@@ -3,7 +3,7 @@
 import { config } from 'dotenv';
 import { createHash } from 'node:crypto';
 import { readFileSync, writeFileSync } from 'node:fs';
-import { dirname, join } from 'node:path';
+import { dirname, join, relative } from 'node:path';
 import { ENV, findHtmlFiles } from './build.utils.mjs';
 
 config({ path: `.env.${ENV}` });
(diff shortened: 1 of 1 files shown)


DenysKarmazynDFINITY (approved):
Great catch - thanks for the fix!

PR #8174, opened 2025-08-07
chore(frontend): Add missing Czech texts

StefanBerger-DFINITY:
# Motivation

We added new labels and texts

# Changes

- also add the Czech translations

# Tests


StefanBerger-DFINITY:
Feedback from the Czech Hub

> hi @Stefan , in general looks good to me, possibly in the longer text some things might depend on a context, but I'd say to release as is an eventually do any updates later also based on public response...

DenysKarmazynDFINITY (approved):

PR #8181, opened 2025-08-08
feat(frontend): introducing locked page

CvetelinaAngelova:
# Motivation
Introducing Lock page to ensures a consistent user experience around session expiry, explicit locking, and logout, while encapsulating state in a simple browser flag.

When a user explicitly locks their session or their session expires, they land on a dedicated “Locked” page.
If they manually navigate back to any protected URL while locked or expired, they see the same “Locked” page.
We can distinguish a real logout (or post-login state) from an intentional lock, so that caching and UI flows behave correctly.

# Changes

locked flag in localStorage
Exposed through a Svelte store that reads on init and writes on updates.As part of login completion, clear any cached user data from previous sessions.
<LockedPage /> component
New UI in src/lib/components/LockedPage.svelte that shows a static background (not the real wallet).

# Tests
Screenshot 2025-08-08 at 10 11 40

### DIFF
--- a/src/frontend/src/lib/api/icp-swap-pool.api.ts
+++ b/src/frontend/src/lib/api/icp-swap-pool.api.ts
@@ -1,3 +1,4 @@
+import type { PoolMetadata } from '$declarations/icp_swap_pool/icp_swap_pool.did';
 import { ICPSwapPoolCanister } from '$lib/canisters/icp-swap-pool.canister';
 import type {
 	ICPSwapDepositWithdrawParams,
@@ -5,6 +6,7 @@ import type {
 	ICPSwapQuoteSwapParams
 } from '$lib/types/api';
 import type { CanisterApiFunctionParamsWithCanisterId } from '$lib/types/canister';
+import type { Identity } from '@dfinity/agent';
 import { Principal } from '@dfinity/principal';
 import { assertNonNullish } from '@dfinity/utils';
 
@@ -76,4 +78,12 @@ export const getUserUnusedBalance = async ({
 }> => {
 	const { getUserUnusedBalance } = await getPoolCanister({ identity, canisterId });
 	return getUserUnusedBalance(principal);
+};
+
+export const getPoolMetadata = async ({
+	identity,
+	canisterId
+}: CanisterApiFunctionParamsWithCanisterId<{ identity: Identity }>): Promise<PoolMetadata> => {
+	const { getPoolMetadata } = await getPoolCanister({ identity, canisterId });
+	return getPoolMetadata();
 };
--- a/src/frontend/src/lib/canisters/icp-swap-pool.canister.ts
+++ b/src/frontend/src/lib/canisters/icp-swap-pool.canister.ts
@@ -1,5 +1,6 @@
 import type {
 	DepositArgs,
+	PoolMetadata,
 	Result,
 	SwapArgs,
 	_SERVICE as SwapPoolService,
@@ -128,6 +129,23 @@ export class ICPSwapPoolCanister extends Canister<SwapPoolService> {
 	): Promise<{ balance0: bigint; balance1: bigint }> => {
 		const { getUserUnusedBalance } = this.caller({ certified: false });
 		const response = await getUserUnusedBalance(principal);
+
+		if ('ok' in response) {
+			return response.ok;
+		}
+
+		throw mapIcpSwapFactoryError(response.err);
+	};
+
+	/**
+	 * Retrieves metadata for the pool.
+	 *
+	 * @returns Object containing PoolMetadata object.
+	 * @throws CanisterInternalError if fetching metadata fails.
+	 */
+	getPoolMetadata = async (): Promise<PoolMetadata> => {
+		const { metadata } = this.caller({ certified: false });
+		const response = await metadata();
 
 		if ('ok' in response) {
 			return response.ok;
--- a/src/frontend/src/tests/lib/canisters/icp-swap-pool.spec.ts
+++ b/src/frontend/src/tests/lib/canisters/icp-swap-pool.spec.ts
@@ -235,4 +235,53 @@ describe('icp_swap_pool.canister', () => {
 			await expect(result).rejects.toThrow(mockResponseError);
 		});
 	});
+
+	describe('getPoolMetadata', () => {
+		it('returns pool metadata successfully', async () => {
+			const mockPoolMetadata = {
+				fee: BigInt(3000),
+				key: 'pool-icp-wtn',
+				sqrtPriceX96: BigInt('3950336'),
+				tick: BigInt(12345),
+				liquidity: BigInt(500000000000n),
+				token0: {
+					address: 'ryjl3-tyaaa-aaaaa-aaaba-cai',
+					standard: 'icrc1'
+				},
+				token1: {
+					address: 'qaa6y-5yaaa-aaaaa-aaafa-cai',
+					standard: 'icrc2'
+				},
+				maxLiquidityPerTick: BigInt('913129639935'),
+				nextPositionId: BigInt(42)
+			};
+			service.metadata.mockResolvedValue({ ok: mockPoolMetadata });
+			const { getPoolMetadata } = await createPool({ serviceOverride: service });
+			const result = await getPoolMetadata();
+
+			expect(result).toEqual(mockPoolMetadata);
+		});
+
+		it('throws CanisterInternalError on error variant', async () => {
+			service.metadata.mockResolvedValue({
+				err: { InternalError: 'Failed to get pool metadata' }
+			});
+			const { getPoolMetadata } = await createPool({ serviceOverride: service });
+			const result = getPoolMetadata();
+
+			await expect(result).rejects.toThrow(
+				new CanisterInternalError('Internal error: Failed to get pool metadata')
+			);
+		});
+
+		it('throws raw error if getPoolMetadata throws', async () => {
+			service.metadata.mockImplementation(() => {
+				throw mockResponseError;
+			});
+			const { getPoolMetadata } = await createPool({ serviceOverride: service });
+			const result = getPoolMetadata();
+
+			await expect(result).rejects.toThrow(mockResponseError);
+		});
+	});
 });


src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,104 @@
...
+
+	const ariaLabel = $derived(replaceOisyPlaceholders($i18n.auth.alt.preview));
+	const modalId = Symbol();
+
+	let src = $state<string | undefined>(undefined);
+	let width = $state<number>(window.innerWidth);
+
+	onMount(() => {
+		const handleResize = () => (width = window.innerWidth);
+		window.addEventListener('resize', handleResize);
```
DenysKarmazynDFINITY: For the cases where we need to display something for a specific breakpoint, we need to use the `Responsive` component, mostly because the `resize` event is quite a heavy one, as well as `addEventListener` is not a very svelte-like approach.

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,104 @@
...
+		return () => window.removeEventListener('resize', handleResize);
+	});
+
+	const resolution = $derived(width >= 1440 ? '1440' : width >= 480 ? '768' : '480');
```
DenysKarmazynDFINITY: We need to use already defined screen sizes in the code. If the existing values cannot be used for some reason, please talk to the design team to make the needed adjustments to the assets.

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,104 @@
...
+	const handleLogout = async () => {
+		lockedPageShow.set(false);
+		await signOut({ resetUrl: true });
+	};
+</script>
+
+<div class="background-overlay flex flex-col">
+	<div
+		class="fixed inset-0 z-[-1] flex items-center justify-center"
+		style="background-color: color-mix(in srgb, var(--color-background-page) 30%, transparent); backdrop-filter: blur(35px);"
```
DenysKarmazynDFINITY: The image seems to already have a background. Why do we need to define it once again in the code?

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,104 @@
...
+	>
+		{#if src}
+			<Img {src} alt={ariaLabel} styleClass="h-full object-contain mx-auto object-top" />
+		{/if}
+	</div>
+
+	<div class="flex h-screen flex-col items-center justify-center px-4">
+		<div
+			class="flex w-full max-w-md flex-col content-center items-center justify-center gap-5 rounded-[24px] bg-[var(--color-background-surface)] p-6 text-center text-[var(--color-text-primary)] shadow-lg transition-all duration-500 ease-in-out md:rounded-[28px] md:p-8"
```
DenysKarmazynDFINITY: the pixel values, as well as `var`s can be replaced with the values from the theme

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,104 @@
...
+		>
+			<div class="pointer-events-auto">
+				<OisyWalletLogoLink />
+			</div>
+			<div class="lock_title_wrapper my-7">
```
DenysKarmazynDFINITY: `lock_title_wrapper` seems to be not needed

DenysKarmazynDFINITY: also, I don't think my-7 class exists

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,104 @@
...
+		{#if src}
+			<Img {src} alt={ariaLabel} styleClass="h-full object-contain mx-auto object-top" />
+		{/if}
+	</div>
+
+	<div class="flex h-screen flex-col items-center justify-center px-4">
+		<div
+			class="flex w-full max-w-md flex-col content-center items-center justify-center gap-5 rounded-[24px] bg-[var(--color-background-surface)] p-6 text-center text-[var(--color-text-primary)] shadow-lg transition-all duration-500 ease-in-out md:rounded-[28px] md:p-8"
+		>
+			<div class="pointer-events-auto">
```
DenysKarmazynDFINITY: any reason for this css class?

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,104 @@
...
+				<OisyWalletLogoLink />
+			</div>
+			<div class="lock_title_wrapper my-7">
+				<h2 class="mb-2 text-2xl font-semibold"> {$i18n.lock.text.title_part_1}</h2>
+				<span class="text-gray-600 mb-6">{$i18n.lock.text.title_part_2}</span>
+			</div>
+			<div class="lock_button_wrapper w-full">
```
DenysKarmazynDFINITY: same with `lock_button_wrapper`

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,104 @@
...
+				<Button fullWidth colorStyle="secondary-light" onclick={handleLogout}
+					>{$i18n.lock.text.logout}
+					<IconLogout />
+				</Button>
+			</div>
+			<p class="text-gray-500 text-xs">
+				{$i18n.lock.text.logout_clear_cash_message}
+			</p>
+		</div>
+		<ExternalLink href="#" ariaLabel="Go to example.com" iconAsLast styleClass="mt-4">
```
DenysKarmazynDFINITY: if the button is just a mock, let's not add it yet to the code

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,104 @@
...
+			modalStore.openAuthHelp({ id: modalId, data: false });
+		}
+	};
+	const handleLogout = async () => {
+		lockedPageShow.set(false);
+		await signOut({ resetUrl: true });
+	};
+</script>
+
+<div class="background-overlay flex flex-col">
```
DenysKarmazynDFINITY: please replace `background-overlay` with tailwind classes

src/frontend/src/lib/components/core/LockOrSignOut.svelte
```diff
@@ -28,6 +30,12 @@
 	const handleLogoutTriggered = () => {
 		onHidePopover?.();
 	};
+
+	const handleLock = async () => {
+		onHidePopover?.();
+		lockedPageShow.set(true);
+		await lockSession({ resetUrl: true });
```
DenysKarmazynDFINITY: I think we discussed that we want to keep the url. Please double-check it with @StefanBerger-DFINITY (unless you have already done it).

src/frontend/src/lib/utils/locked.utils.ts
```diff
@@ -0,0 +1,12 @@
+import * as localStore from '$lib/utils/storage.utils';
+import { writable } from 'svelte/store';
+
+const KEY = 'lockedPageShow';
+
+const initial = localStore.get<boolean>({ key: KEY }) ?? false;
```
DenysKarmazynDFINITY: Please create a proper store for this data. As a reference, you can check how it's done in i18n.store.

src/frontend/src/lib/utils/locked.utils.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const lockedPageShow = writable<boolean>(initial);
```
DenysKarmazynDFINITY: After you create the store, you can create a derived for this variable.

src/frontend/src/lib/utils/locked.utils.ts
```diff
@@ -0,0 +1,12 @@
+import * as localStore from '$lib/utils/storage.utils';
+import { writable } from 'svelte/store';
+
+const KEY = 'lockedPageShow';
```
DenysKarmazynDFINITY: to be consistent with the auth states, I'd suggest something like `authLocked`

src/frontend/src/routes/(app)/+layout.svelte
```diff
@@ -55,59 +57,63 @@
...
-
-			<Modals />
-		</AuthGuard>
-
-		{#if $aiAssistantConsoleOpen}
-			<AiAssistantConsole />
-		{:else}
-			<Footer />
-		{/if}
+{#if $authNotSignedIn && $lockedPageShow}
```
DenysKarmazynDFINITY: Having the `$authNotSignedIn` check here seems to be a bit confusing - there won't be a case where locked page flag is true but `$authNotSignedIn` is false. I would suggest keeping just the newly renamed `authLocked` flag here. Inside the derived, you can use `$authNotSignedIn` if you think it's more preferable.

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: The source image is only 1440 px wide, so on screens wider than that it can’t fill the entire area when you use background-size: contain. The extra space around it needs to be filled somehow, so we specify a background color—otherwise you’d see whatever’s behind the element.

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: remove the unused classes

src/frontend/src/lib/components/core/LockOrSignOut.svelte
CvetelinaAngelova: changed to false

src/frontend/src/lib/utils/locked.utils.ts
CvetelinaAngelova: renamed to authLocked

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: Removed and implement the Responsive component, thanks for the suggestion

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: Removed and implement the Responsive component, thanks for the suggestion

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: Moved the most style in the css for the page

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: removed

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: I'm not sure what it should be, but we have it in the design

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: it contain style

src/frontend/src/lib/components/auth/LockPage.svelte
DenysKarmazynDFINITY: The problem is there might be a visual gap where the image ends and CSS background starts. When you will be creating a PR for these changes, please attach screenshots with the different breakpoints so we can review them.

src/frontend/src/lib/components/auth/LockPage.svelte
DenysKarmazynDFINITY: what I meant is that the styles that are inside that class should be migrated to tailwind classes

src/frontend/src/lib/utils/locked.utils.ts
CvetelinaAngelova: Update the storage use

src/frontend/src/routes/(app)/+layout.svelte
CvetelinaAngelova: add new derived values

src/frontend/src/lib/utils/locked.utils.ts
CvetelinaAngelova: add it as a separate file into derived folder

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: removed

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: migrate the style to to tailwind classes

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: We have real link provided

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,109 @@
...
+		const currentTheme = $themeStore ?? 'light';
+
+		import(`$lib/assets/lockpage_assets/lock-image-1440-${currentTheme}.webp`)
+			.then((mod) => (src = mod.default))
+			.catch(() => console.error('Failed to load background image'));
+	});
+
+	const handleUnlock = async () => {
+		const { success } = await signIn({});
+		authLocked.unlock({
```
DenysKarmazynDFINITY: I think the "unlock" function should be called only on `success`. If `signIn` is not successful, does it logout automatically?

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,109 @@
...
+
+	const ariaLabel = $derived(replaceOisyPlaceholders($i18n.auth.alt.preview));
+	const modalId = Symbol();
+
+	let src = $state<string | undefined>(undefined);
+
+	$effect(() => {
+		const currentTheme = $themeStore ?? 'light';
+
+		import(`$lib/assets/lockpage_assets/lock-image-1440-${currentTheme}.webp`)
```
DenysKarmazynDFINITY: Please implement it the way it's done in `Loader.tsx`, the line 208th. It should be 3 different imports (one per each Responsive) - when you just replacing the `size` it doesn't necessarily mean that the import got successfully resolved

src/frontend/src/lib/components/auth/LockPage.svelte
DenysKarmazynDFINITY: Those values still need to be replaced: `rounded-[24px] bg-[var(--color-background-surface)] text-[var(--color-text-primary)] md:rounded-[28px]`

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,109 @@
...
+
+	const handleLogout = async () => {
+		authLocked.unlock({
+			source: 'logout from lock page'
+		});
+		await signOut({ resetUrl: true });
+	};
+</script>
+
+<div class="z-4 fixed inset-0 flex h-full w-full flex-col bg-[var(--color-background-page)]">
```
DenysKarmazynDFINITY: `bg-[var(--color-background-page)]` - there should be a value in the theme file that you can use instead

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,109 @@
...
+	<div
+		class="fixed inset-0 z-[-1]
+		       bg-[color-mix(in_srgb,var(--color-background-page)_30%,transparent)] backdrop-blur-[35px]"
```
DenysKarmazynDFINITY: please move the bg value to the theme file - we have quite a few custom gradients already defined there, so you will find a lot of examples

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,109 @@
...
+	>
+		{#if src}
+			<Responsive up="xl">
+				<Img {src} alt={ariaLabel} styleClass="h-full object-contain mx-auto object-top" />
```
DenysKarmazynDFINITY: let's create a state variable for `stlyeClass` and re-use it in all 3 cases

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,109 @@
...
+		{/if}
+	</div>
+
+	<div class="flex h-screen flex-col items-center justify-center px-4">
+		<div
+			class="flex w-full max-w-md flex-col content-center items-center justify-center gap-5 rounded-[24px] bg-[var(--color-background-surface)] p-6 text-center text-[var(--color-text-primary)] shadow-lg transition-all duration-500 ease-in-out md:rounded-[28px] md:p-8"
+		>
+			<OisyWalletLogoLink />
+
+			<div class="my-7">
```
DenysKarmazynDFINITY: `my-7` class does not exist - either `my-6` or `my-8`

src/frontend/src/lib/components/auth/LockPage.svelte
```diff
@@ -0,0 +1,109 @@
...
+				</Button>
+			</div>
+
+			<p class="tertiary text-xs">
+				{$i18n.lock.text.logout_clear_cash_message}
+			</p>
+		</div>
+		<ExternalLink
+			href="https://docs.oisy.com/using-oisy-wallet/how-tos/locking-and-logging-out"
+			ariaLabel="Go to example.com"
```
DenysKarmazynDFINITY: `ariaLabel` needs to be translated

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: changed

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: changed

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: change the unlock to be if (success === 'ok')

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: changed

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: implement imgStyleClass

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: changed to 6

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: translated

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: change the awayt in every 	<Responsive

src/frontend/src/lib/components/auth/LockPage.svelte
CvetelinaAngelova: moved to the theme and create new bg class

DenysKarmazynDFINITY (approved):
LGTM, thanks - great job 💪

PR #8184, opened 2025-08-08
feat(frontend): ignore kong token metrics if price is 0

DenysKarmazynDFINITY:
# Motivation

In case kong token metrics does not have a price yet (e.g. there is no a pool yet created for the token), we want to ignore it completely instead of showing the token balance as 0.


roman-nazaruk (approved):
LGTM, thnx

PR #8186, opened 2025-08-08
feat(frontend): add minimum BTC amount validation and utility for satoshi conversion

DecentAgeCoder:
# Motivation

To fix the issue where BTC sending of small amounts creates failed transactions that block subsequent sends, we need to establish minimum amount validation. Users attempting to send very small amounts (e.g. 0.000001 BTC) experience silent failures where the transaction never appears on the block explorer, but the wallet shows a "pending" state for about an hour, preventing new transactions.

The first step is to establish the foundational constants and utilities needed for proper BTC amount validation.

# Changes

- Add `BTC_MINIMUM_AMOUNT` constant (700 satoshis) to prevent dust transactions
- Implement `convertSatoshisToBtc` utility function for displaying user-friendly amounts
- Add `invalidSendAmount` validation function to input utilities

# Tests

- Verified BTC_MINIMUM_AMOUNT constant matches network requirements (700 satoshis)
- Added test for `convertSatoshisToBtc`
- Tested `invalidSendAmount` properly validates amounts against minimum threshold


src/frontend/src/lib/utils/input.utils.ts
```diff
@@ -7,3 +8,6 @@ export const isNullishOrEmpty = (value: OptionString): value is Option<''> => is
 
 export const invalidAmount = (amount: OptionAmount): boolean =>
 	isNullish(amount) || Number(amount) < 0;
+
+export const invalidSendAmount = (amount: OptionAmount): boolean =>
+	isNullish(amount) || Number(amount) < BTC_MINIMUM_AMOUNT;
```
DenysKarmazynDFINITY: Since this method is purely BTC-related, can you please move it to `$btc/utils/input.utils.ts`? And let's add a basic unit test for that - just to not decrease the test coverage threshold.

DenysKarmazynDFINITY: Also, is it going to be used in the next PRs?

UPD: yeah, I guess I found the answer in the ticket description :)

DecentAgeCoder: Correct that will be used in one of the follow-up PR. I agree, this should be move to the corresponding btc util.

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8188, opened 2025-08-08
feat(frontend):  add lock page utils and derived value

CvetelinaAngelova:
# Motivation

Implement centralized session locking state management to:
Track lock/unlock events consistently
Persist lock state across sessions

# Changes

Created authLocked store with:
      Persistent storage via initStorageStore
      Analytics tracking for lock/unlock events
      setLock() and toggleLock() methods

Added derived state:
    isAuthLocked - Current lock status
    shouldShowLockPage - Combines with auth state

# Tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6461,9 +6461,9 @@
 			"integrity": "sha512-s0IrSOzLlbvX7yp4WBfPITzpAU8sqQcpsmwXDiKwrG4r491vwCO/XpejasRNl0piBMe/DvP4Tz0mIS/X1DPJBQ=="
 		},
 		"node_modules/bignumber.js": {
-			"version": "9.1.2",
-			"resolved": "https://registry.npmjs.org/bignumber.js/-/bignumber.js-9.1.2.tgz",
-			"integrity": "sha512-2/mKyZH9K85bzOEfhXDBFZTGd1CTs+5IHpeFQo9luiBG7hghdC851Pj2WAhb6E3R6b9tZj/XKhbg4fum+Kepug==",
+			"version": "9.3.1",
+			"resolved": "https://registry.npmjs.org/bignumber.js/-/bignumber.js-9.3.1.tgz",
+			"integrity": "sha512-Ko0uX15oIUS7wJ3Rb30Fs6SkVbLmPBAKdlm7q9+ak9bbIeFf0MwuBsQV6z7+X768/cHsfg+WlysDWJcmthjsjQ==",
 			"license": "MIT",
 			"engines": {
 				"node": "*"


src/frontend/src/lib/utils/locked.utils.ts
```diff
@@ -0,0 +1,46 @@
...
+				source: `${source}`
+			}
+		});
+
+		store.set({ key: 'authLocked', value: locked });
+	};
+
+	const toggleLock = (options: Omit<SetAuthLockOptions, 'locked'>) => {
+		const current = get(store);
+		setLock({ ...options, locked: !current });
```
DenysKarmazynDFINITY: Let's keep just the `toggleLock` method - no need to have both `toggle and `set` functions at the same time.

src/frontend/src/lib/utils/locked.utils.ts
```diff
@@ -0,0 +1,46 @@
...
+import { trackEvent } from '$lib/services/analytics.services';
+import { initStorageStore } from '$lib/stores/storage.store';
+import { get } from 'svelte/store';
+
+interface SetAuthLockOptions {
+	locked: boolean;
+	source?: string;
+}
+
+export interface AuthLockStore extends ReturnType<typeof initStorageStore<boolean>> {
```
DenysKarmazynDFINITY: Please move this file to the `stores` folder.

src/frontend/src/lib/derived/locked.derived.ts
```diff
@@ -0,0 +1,13 @@
+import { authNotSignedIn } from '$lib/derived/auth.derived';
+import { authLocked } from '$lib/utils/locked.utils';
+import { derived, type Readable } from 'svelte/store';
+
+export const isAuthLocked: Readable<boolean> = derived(
+	[authLocked],
+	([$authLocked]) => $authLocked
+);
+
+export const shouldShowLockPage: Readable<boolean> = derived(
```
DenysKarmazynDFINITY: Correct me if I'm wrong, but I think we should be fine with having just 1 store - I'd rename this one to `isAuthLocked` and remove the other one

DenysKarmazynDFINITY: Also, for this file, we also need some basic unit tests. You will find a lot of examples in the tests/derived folder.

src/frontend/src/lib/constants/analytics.contants.ts
```diff
@@ -131,6 +131,9 @@ export const MANAGE_TOKENS_MODAL_ROUTE = 'manage-tokens-modal';
 // Privacymode change
 export const TRACK_PRIVACY_MODE_CHANGE = 'privacy_mode_change';
 
+// Lock mode activated
+export const TRACK_lOCK_MODE_ACTIVATED = 'lock_mode_activated';
```
DenysKarmazynDFINITY: I noticed this event will be called for both "activated" and "deactivated". Can you please also add `TRACK_lOCK_MODE_DEACTIVATED` and use it accordingly in the function where it's called?

src/frontend/src/lib/utils/locked.utils.ts
CvetelinaAngelova: remove the seLock, there will be only toggleLock

src/frontend/src/lib/derived/locked.derived.ts
CvetelinaAngelova: changed and add unit tests

src/frontend/src/lib/utils/locked.utils.ts
CvetelinaAngelova: moved

src/frontend/src/lib/constants/analytics.contants.ts
CvetelinaAngelova: add TRACK_lOCK_MODE_DEACTIVATED

src/frontend/src/lib/utils/locked.utils.ts
CvetelinaAngelova: I set up the toggleLock but Also add lock and unlock to be more easy to use the store in the components

src/frontend/src/lib/stores/locked.store.ts
```diff
@@ -0,0 +1,45 @@
...
+	const updateLock = (newValue: boolean, source: string): void => {
+		trackEvent({
+			name: newValue ? TRACK_lOCK_MODE_ACTIVATED : TRACK_lOCK_MODE_DEACTIVATED,
+			metadata: { locked: String(newValue), source }
+		});
+		store.set({ key: 'authLocked', value: newValue });
+	};
+
+	const toggleLock = ({ source }: ToggleLockOptions): void => {
+		updateLock(!get(store), source ?? 'unknown');
```
DenysKarmazynDFINITY: nitpick: instead of adding the default value 3 times, you can define it on the `updateLock` function level

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/lib/stores/locked.store.ts
CvetelinaAngelova: changed

PR #8206, opened 2025-08-11
feat(frontend): improve mobile touch targets for language and currency dropdowns

artkorotkikh-dfinity:
# Motivation

Improves user experience on mobile devices by providing larger, easier-to-tap targets that meet accessibility guidelines for touch interfaces.

# Changes

- Add mobile-specific styles for language and currency selector dropdowns
- Increase minimum height to 44px for better accessibility on mobile devices
- Use CSS variables for maintainable and consistent styling
- Add --touch-target-min-height variable (44px) to global variables
- Use existing --padding-1_5x variable for consistent vertical padding
- Apply changes only on mobile devices (≤767px) when dropdowns become modals
- Targets .lang-selector and .currency-selector classes specifically

# Tests  

**Before:**  
staging oisy com_(iPhone XR) (1) 
staging oisy com_(iPhone XR) (2)  

**After:**  
localhost_5174_(iPhone XR) 
localhost_5174_(iPhone XR) (2)


daviddecentage (approved):
LGTM

PR #8229, opened 2025-08-12
feat(frontend): added new getWithdrawableToken util method for ICP Swap

roman-nazaruk:
# Motivation

We need to validate tokens from the pool against the unused balance in ICP Swap.
This PR introduces a utility method returns the withdrawable token based on data from the pool metadata and userUnusedBalances.

# Changes

Added new utility method getWithdrawableToken to identify the correct token using ICP Swap pool metadata.

# Tests

Added unit tests to cover the new method


BonomoAlessandro (approved):
LGTM

PR #8230, opened 2025-08-12
feat(frontend): added new functionality to btain quote result for EVMsupported networks

roman-nazaruk:
# Motivation

Provide an extensible, uniform way to fetch EVM swap quotes so it’s easy to add more DEX providers later (starting with Velora) without touching call sites.

# Changes

- Added fetchSwapAmountsEVM aggregator that returns a uniform SwapMappedResult[] and currently sources quotes from Velora.
- Normalized Velora responses (delta / market) via mappers; return [] when no route is found.

# Tests

Added test for new functionality

### DIFF
--- a/src/frontend/src/lib/constants/swap.constants.ts
+++ b/src/frontend/src/lib/constants/swap.constants.ts
@@ -18,3 +18,6 @@ export const SWAP_ETH_TOKEN_PLACEHOLDER = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeee
 
 export const OISY_DOCS_SWAP_WIDTHDRAW_FROM_ICPSWAP_LINK =
 	'https://docs.oisy.com/using-oisy-wallet/how-tos/swapping-tokens#manually-withdraw-funds-from-icpswap';
+
+export const SWAP_MODE = 'all';
+export const SWAP_SIDE = 'SELL';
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/constants/swap.constants.ts
```diff
@@ -18,3 +18,6 @@ export const SWAP_ETH_TOKEN_PLACEHOLDER = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeee
 
 export const OISY_DOCS_SWAP_WIDTHDRAW_FROM_ICPSWAP_LINK =
 	'https://docs.oisy.com/using-oisy-wallet/how-tos/swapping-tokens#manually-withdraw-funds-from-icpswap';
+
+export const SWAP_MODE = 'all';
+export const SWAP_SIDE = 'SELL';
```
BonomoAlessandro, line 23: Is `SELL` written with capital letters by purpose and `all` not?

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -480,3 +488,71 @@ export const performManualWithdraw = async ({
...
+	userAddress
+}: VeloraQuoteParams): Promise<SwapMappedResult[]> => {
+	const swapAmountsResults = await fetchVeloraSwapAmount({
+		sourceToken,
+		destinationToken,
+		amount,
+		userAddress
+	});
+
+	if (isNullish(swapAmountsResults)) {
```
BonomoAlessandro, line 518: this is the same as `return nonNullish(swapAmountsResults) ? [swapAmountsResults] : []`

BonomoAlessandro (approved):
LGTM

src/frontend/src/lib/constants/swap.constants.ts
roman-nazaruk: Yes, it was stated in the Velora documentation.
image

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Agree, but I prefer to have a simple if condition as it makes code more readable

PR #8231, opened 2025-08-12
feat(frontend): implements load erc721 transaction function from etherscan provider

BonomoAlessandro:
# Motivation

We want to be able to load the `erc721` transactions of users.

# Changes

- implements load function


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8232, opened 2025-08-12
feat(frontend): adds tokenId to transaction

BonomoAlessandro:
# Motivation

The `Transaction` object should also support nft transactions which means it needs an optional `tokenId` field.

# Changes

- adds `tokenId` to `Transaction`


roman-nazaruk (approved):
LGTM

PR #8239, opened 2025-08-12
feat(frontend): clear idb-balances methods

DenysKarmazynDFINITY:
# Motivation

Similar as with the deletion per principal, we now need all the methods for a complete "clear" of IDB balances.


roman-nazaruk (approved):
LGTM, thnx

PR #8243, opened 2025-08-12
feat(frontend): add possibility to clear all IDB on logout

DenysKarmazynDFINITY:
# Motivation

We now need to add an additional param to `logout` that will allow consumers to clear all IDB tables.


roman-nazaruk (approved):
LGTM, thnx

PR #8245, opened 2025-08-12
feat(frontend): fix menu buttons

CvetelinaAngelova:
# Motivation

The layout of the Lock and Log out buttons was inconsistent, especially with varying text lengths (e.g., localized versions like "Odhlásit se"). 

# Changes

Unified styling for Lock and Log out buttons to ensure consistent alignment, padding, and sizing.
Ensured icons remain vertically centered and do not shrink.

# Tests
Verified appearance in English (Log out) and other languages to confirm dynamic spacing works correctly.
Screenshot 2025-08-12 at 20 07 53


src/frontend/src/lib/components/core/LockOrSignOut.svelte
```diff
@@ -1,21 +1,22 @@
...
 
 	interface Props {
 		onHidePopover?: () => void;
-		hideText?: boolean;
 	}
-	let { onHidePopover, hideText = true }: Props = $props();
+	let { onHidePopover }: Props = $props();
 
 	let remainingTimeMs = $derived($authRemainingTimeStore);
+	const dispatch = createEventDispatcher();
```
DenysKarmazynDFINITY: The "dispatch" approach is deprecated in Svelte 5 - can you please change switch to passing onIcLogoutTriggered as a prop?

src/frontend/src/lib/components/ui/Button.svelte
```diff
@@ -71,7 +71,7 @@
 	aria-label={ariaLabel}
 >
 	<span
-		class="flex min-w-0 gap-2"
+		class="flex min-w-0 items-center justify-center gap-2"
```
DenysKarmazynDFINITY: It's too risky to add these classes to a very low-level component that is used across the whole app - if you cannot achieve the wanted result by applying the styles inside a consumer (e.g. a wrapper `div`), you can introduce a new prop called `innerStyleClass`.

src/frontend/src/lib/components/core/LockOrSignOut.svelte
CvetelinaAngelova: changed to prop

src/frontend/src/lib/components/ui/Button.svelte
CvetelinaAngelova: very nice Idea, I added the new prop

src/frontend/src/lib/components/core/LockOrSignOut.svelte
CvetelinaAngelova: removed it as it's not needed anymore

PR #8247, opened 2025-08-12
feat(frontend): added withdrawUserUnusedBalance for ICP Swap

roman-nazaruk:
# Motivation

We encountered an issue with ICP Swap where, after a swap operation, a portion of both the sourceToken and destinationToken may remain in the pool under the user's balance. This PR introduces functionality to detect and withdraw any such unused token balances left in the pool.

# Changes

- Added withdrawUserUnusedBalance utility for checking and withdrawing unused user pool balances.
- Updated SwapParams type definitions to support new logic.
- Integrated withdrawUserUnusedBalance to handle manual withdrawal requests and after a successful swap.

# Tests

Covered new logic with tests

### DIFF
--- a/src/frontend/src/lib/services/swap.services.ts
+++ b/src/frontend/src/lib/services/swap.services.ts
@@ -7,9 +7,20 @@ import { nowInBigIntNanoSeconds } from '$icp/utils/date.utils';
 import { isTokenIcrc } from '$icp/utils/icrc.utils';
 import { setCustomToken } from '$lib/api/backend.api';
 import { getPoolCanister } from '$lib/api/icp-swap-factory.api';
-import { deposit, depositFrom, swap as swapIcp, withdraw } from '$lib/api/icp-swap-pool.api';
+import {
+	deposit,
+	depositFrom,
+	getPoolMetadata,
+	getUserUnusedBalance,
+	swap as swapIcp,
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -480,3 +489,54 @@ export const performManualWithdraw = async ({
...
+>): Promise<void> => {
+	const { token0, token1 } = await getPoolMetadata({ identity, canisterId });
+	const { balance0, balance1 } = await getUserUnusedBalance({
+		identity,
+		canisterId,
+		principal: identity.getPrincipal()
+	});
+
+	if (balance0 === ZERO && balance1 === ZERO) {
+		throw new Error('No unused balance to withdraw');
```
DenysKarmazynDFINITY, line 510: Just curious - is it really an error if there are no unused balances to withdraw? To me, it seems like a normal state ("do nothing") rather than error, but I may miss some context.

roman-nazaruk: I throw an error here because it is wrapped in a try-catch block. If we do not have a withdraw balance, it means that something went wrong, and it’s better to terminate this method. When the error is thrown, we catch this state and return another SwapError with a custom message.

PR #8250, opened 2025-08-13
feat(frontend): extends Logo implementation

BonomoAlessandro:
# Motivation

The `Logo` should not always be displayed rounded. It should be configurable whether the `Logo` is rounded or not. 

# Changes

- updates `Logo`

### DIFF
--- a/src/frontend/src/lib/components/ui/Logo.svelte
+++ b/src/frontend/src/lib/components/ui/Logo.svelte
@@ -15,7 +15,15 @@
 		testId?: string;
 	}
 
-	let { src, alt = '', size = 'xxs', color = 'off-white', ring = false, rounded = true, testId }: Props = $props();
+	let {
+		src,
+		alt = '',
+		size = 'xxs',
+		color = 'off-white',
+		ring = false,
+		rounded = true,
+		testId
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/ui/Logo.svelte
```diff
@@ -23,11 +32,13 @@
...
 
 <div
-	class="flex items-center justify-center overflow-hidden rounded-full ring-primary"
+	class="flex items-center justify-center overflow-hidden ring-primary"
 	class:bg-off-white={color === 'off-white' && !isReady}
 	class:bg-white={color === 'white' && !isReady}
 	class:opacity-10={!isReady}
 	class:ring-2={ring}
+	class:rounded-full={rounded}
+	class:rounded-lg={!rounded}
```
DenysKarmazynDFINITY: I'd probably call the prop `circle` or something like this just because `class:rounded-lg={!rounded}` looks a bit off - for `!rounded` I would expect the class to be `rounded-none`.

DenysKarmazynDFINITY (approved):
Small observation, otherwise LGTM

PR #8257, opened 2025-08-13
feat(frontend): updates eth-transactions.services.ts to also load erc1155 token transactions

BonomoAlessandro:
# Motivation

We want that the transactions of `erc1155` tokens also are loaded

# Changes

- also loads `erc1155` token transactions


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8258, opened 2025-08-13
feat(frontend): implements findNft utils function

BonomoAlessandro:
# Motivation

We want to have a utils function `findNft`.

# Changes

- implements `findNft`


daviddecentage (approved):
LGTM

PR #8271, opened 2025-08-14
feat(frontend): NFT list store

daviddecentage:
# Motivation

For the upcoming NFT list we need to have a store to have filters and sorting settings persisted.

# Changes

Added store

### DIFF
--- a/src/frontend/src/lib/stores/nft-list.store.ts
+++ b/src/frontend/src/lib/stores/nft-list.store.ts
@@ -1,9 +1,9 @@
 import { writable } from 'svelte/store';
 
-export type NftListSortingType = {
+export interface NftListSortingType {
 	order: 'asc' | 'desc';
 	type: 'collection-name' | 'date';
-};
+}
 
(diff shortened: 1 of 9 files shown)


src/frontend/src/lib/stores/nft-list.store.ts
```diff
@@ -0,0 +1,19 @@
...
+	order: 'asc' | 'desc';
+	type: 'collection-name' | 'date';
+};
+
+interface NftListStoreData {
+	sort: NftListSortingType;
+	groupByCollection: boolean;
+}
+
+export const nftListStore = writable<NftListStoreData>({
```
DenysKarmazynDFINITY: we'll need to add some basic unit tests for this one just to prevent the test coverage score going down

daviddecentage: Will add!

BonomoAlessandro:
Wouldn't `nftListSettingsStore` be a better name for this?

PR #8278, opened 2025-08-14
feat(frontend): Get NFT and NFTCollections Utils

daviddecentage:
# Motivation

We need util functions to get the NFTs and NFT collections.

# Changes

- Added new type nftCollectionsUi which contains all collections and its NFTs
- Added getEnabledNfts
- Added getNftCollectionUi

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/utils/nfts.utils.ts
+++ b/src/frontend/src/lib/utils/nfts.utils.ts
@@ -1,6 +1,13 @@
 import { NftCollectionSchema } from '$lib/schema/nft.schema';
 import type { NftError } from '$lib/types/errors';
-import type { Nft, NftCollection, NftId, NftsByNetwork, NonFungibleToken } from '$lib/types/nft';
+import type {
+	Nft,
+	NftCollection,
+	NftCollectionUi,
+	NftId,
+	NftsByNetwork,
+	NonFungibleToken
+} from '$lib/types/nft';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -98,3 +105,38 @@ export const mapTokenToCollection = (token: NonFungibleToken): NftCollection =>
...
+	);
+
+export const getNftCollectionUi = ({
+	$nonFungibleTokens,
+	$nftStore
+}: {
+	$nonFungibleTokens: NonFungibleToken[];
+	$nftStore: Nft[] | undefined;
+}): NftCollectionUi[] =>
+	$nonFungibleTokens.map(mapTokenToCollection).map((coll) => ({
```
DenysKarmazynDFINITY: Can we please switch to a reduce function instead of map + map + filter?

daviddecentage:
> My understanding these utils will later be used for creating derived stores, right?

Not a derived as we just pass the store data as params and use it where we need the data

src/frontend/src/lib/utils/nfts.utils.ts
daviddecentage: Updated it

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8280, opened 2025-08-15
fix(frontend): add locked state to stop the timer of the Auth worker

CvetelinaAngelova:
# Motivation

When lock mode is enabled, the auth worker’s idle timer continues for one more tick and triggers idleSignOut(), causing an unintended session expired warning while the UI is locked.

# Changes

Gate idle timer by lock state: Extend syncAuthIdle to accept an optional locked?: boolean flag.
locked === true or missing auth.identity → post stopIdleTimer.
This chanes prevent the idleSignOut function to be triggered from the timer.

# Tests

Behavioral outcome: No more extra tick while locked; no idle sign-out or warning during lock.

### DIFF
--- a/src/frontend/src/lib/utils/nfts.utils.ts
+++ b/src/frontend/src/lib/utils/nfts.utils.ts
@@ -1,4 +1,5 @@
 import { NftCollectionSchema } from '$lib/schema/nft.schema';
+import type { NftListSortingType } from '$lib/stores/nft-list.store';
 import type { NftError } from '$lib/types/errors';
 import type {
 	Nft,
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/services/worker.auth.services.ts
```diff
@@ -22,15 +22,13 @@ export const initAuthWorker = async () => {
 	};
 
 	return {
-		syncAuthIdle: (auth: AuthStoreData) => {
-			if (!auth.identity) {
+		syncAuthIdle: ({ auth, locked = false }: { auth: AuthStoreData; locked?: boolean }) => {
+			if (locked || !auth.identity) {
```
DenysKarmazynDFINITY, line 29: does it mean that after calling the logout function `auth.identity` still stays defined for a short period of time? even though your suggested changes will most likely prevent the sign out toast from being shown, the root cause will stay in the code, and can backfire in some other places - I'd really try to solve it completely

CvetelinaAngelova: No it's not preventing the sign out toast. This is the initial logic of the worker to stop the IdleTimer and prevent multiple logout if the user is logOut and auth.identity is false. Now it will fire also on the locked page. Basicly this is the root cause to not do double logout.

DenysKarmazynDFINITY: As discussed offline, let's proceed with your fix, and add a TODO comment here to investigate and fix the underlying problem.

CvetelinaAngelova: Adding a TODO message to the worker

PR #8282, opened 2025-08-15
feat(frontend): NFTs and collections filter and sorting utils

daviddecentage:
# Motivation

We also need some util functions to sort and filter the NFTs and NFT collections.

We use a collator and dynamic comparator function, which is not needed since we only have one comparator right now. But its planned to add additional sorting types so we prepare for that.

# Changes

- Added filter and sort function for NFTs
- Added filter and sort functions for collections

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/utils/nfts.utils.ts
+++ b/src/frontend/src/lib/utils/nfts.utils.ts
@@ -185,10 +185,10 @@ interface NftCollectionFilterAndSortParams extends NftBaseFilterAndSortParams<Nf
 	items: NftCollectionUi[];
 }
 
-type FilterSortByCollection = {
+interface FilterSortByCollection {
 	(params: NftFilterAndSortParams): Nft[];
 	(params: NftCollectionFilterAndSortParams): NftCollectionUi[];
-};
+}
 
 // Single implementation (T is Nft or NftCollectionUi)
 export const filterSortByCollection: FilterSortByCollection = <T extends Nft | NftCollectionUi>({


src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -155,3 +156,66 @@ export const getNftCollectionUi = ({
 		return acc;
 	}, []);
 };
+
+const collator = new Intl.Collator(undefined, {
```
DenysKarmazynDFINITY: did you consider passing the current locale instead of undefined?

src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -155,3 +156,66 @@ export const getNftCollectionUi = ({
...
+	filter,
+	sort
+}: {
+	nfts: Nft[];
+	filter?: string;
+	sort?: NftListSortingType;
+}): Nft[] => {
+	if (nonNullish(filter)) {
+		nfts = [...nfts].filter(
+			(nft) => (nft?.collection?.name?.toLowerCase() ?? '').indexOf(filter.toLowerCase()) >= 0
```
DenysKarmazynDFINITY: let's use `''.includes` here

src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -155,3 +156,66 @@ export const getNftCollectionUi = ({
...
+		const dir = sort.order === 'asc' ? 1 : -1;
+		const comparator = cmpByCollectionName(dir);
+
+		nfts = [...nfts].sort((a, b) => comparator({ a, b }));
+	}
+
+	return nfts;
+};
+
+export const filterSortNftCollections = ({
```
DenysKarmazynDFINITY: so the functions are completely the same, just different types? then let's combine them and define params and return type with the help of generics

src/frontend/src/lib/utils/nfts.utils.ts
daviddecentage: updated

src/frontend/src/lib/utils/nfts.utils.ts
daviddecentage: updated

src/frontend/src/lib/utils/nfts.utils.ts
daviddecentage: right I didnt realize its the same because the types should be completely different :) Updated!

src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -155,3 +156,61 @@ export const getNftCollectionUi = ({
...
+
+// Single implementation (T is Nft or NftCollectionUi)
+export function filterSortByCollection<T extends Nft | NftCollectionUi>({
+	items,
+	filter,
+	sort
+}: {
+	items: T[];
+	filter?: string;
+	sort?: NftListSortingType;
```
DenysKarmazynDFINITY: nitpick: let's create a common interface for `filter` and `sort` as we have the exact same definition 3 times in this file

src/frontend/src/tests/lib/utils/nfts.utils.spec.ts
```diff
@@ -97,6 +98,21 @@ describe('nfts.utils', () => {
 		}
 	};
 
+	// Same collator settings as in the impl
+	const collator = new Intl.Collator(new Intl.Locale(navigator.language), {
```
DenysKarmazynDFINITY, line 102: I was thinking more about using `currentLanguage` store here

DenysKarmazynDFINITY (approved):
2 more minor comments - LGTM otherwise

src/frontend/src/tests/lib/utils/nfts.utils.spec.ts
daviddecentage: But language is not the same as the locale.

src/frontend/src/lib/utils/nfts.utils.ts
daviddecentage: updated

PR #8284, opened 2025-08-15
feat(frontend): NFT sorting menu

daviddecentage:
# Motivation

We need a responsive sorting menu for NFT pages similar to the settings popover/bottom sheet for the token list.

# Changes

Added component

# Tests

Trigger:
image

Desktop:
image

Mobile:
image

### DIFF
--- a/src/frontend/src/eth/components/transactions/EthTransaction.svelte
+++ b/src/frontend/src/eth/components/transactions/EthTransaction.svelte
@@ -20,6 +20,7 @@
 	let type: EthTransactionType;
 	let to: string | undefined;
 	let from: string | undefined;
+	let tokenId: number | undefined;
 
 	let pending: boolean;
 	$: pending = isTransactionPending(transaction);
@@ -27,7 +28,7 @@
 	let status: TransactionStatus;
 	$: status = pending ? 'pending' : 'confirmed';
 
-	$: ({ value, timestamp, displayTimestamp, type, to, from } = transaction);
+	$: ({ value, timestamp, displayTimestamp, type, to, from, tokenId } = transaction);
 
 	let ckTokenSymbol: string;
 	$: ckTokenSymbol = isSupportedEthToken(token)
@@ -80,6 +81,7 @@
 	{iconType}
 	{to}
 	{from}
+	{tokenId}
 >
 	{label}
 </Transaction>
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/nfts/NftSortMenu.svelte
```diff
@@ -0,0 +1,127 @@
...
+				disabled
+				onclick={() => setSorting({ order: 'asc', type: 'date' })}
+				fullWidth
+				alignLeft
+				styleClass="py-3 rounded-md text-primary underline-none pl-0.5 min-w-28"
+				colorStyle="tertiary-alt"
+				transparent
+			>
+				<span class="pt-0.75 w-[20px] text-brand-primary">
+					{#if $nftListStore.sort.type === 'date' && $nftListStore.sort.order === 'asc'}
```
DenysKarmazynDFINITY: How about creating derived for these values?

src/frontend/src/lib/components/nfts/NftSortMenu.svelte
```diff
@@ -0,0 +1,127 @@
...
+	const setSorting = (sort: NftListSortingType) => {
+		nftListStore.setSort(sort);
+	};
+</script>
+
+{#snippet content()}
+	<span class="mb-2 flex text-sm font-bold">{$i18n.nfts.text.sorting}</span>
+
+	<List noPadding>
+		<ListItem>
```
DenysKarmazynDFINITY: Let's extract this item into a separate component and re-use it in this file - 3 appearances deserve a dedicated file

src/frontend/src/lib/components/nfts/NftSortMenu.svelte
```diff
@@ -0,0 +1,127 @@
...
+	link={false}
+	colorStyle="muted"
+	styleClass={visible ? 'active' : ''}
+>
+	{#snippet icon()}
+		<IconArrowUpDown />
+	{/snippet}
+</ButtonIcon>
+
+<Responsive up="sm">
```
DenysKarmazynDFINITY: This part can also be extracted into a separate component - it's exactly the same in the sibling PR, and I think I saw it already somewhere :)

daviddecentage: done

src/frontend/src/lib/components/nfts/NftSortMenu.svelte
daviddecentage: #8288

src/frontend/src/lib/components/nfts/NftSortMenu.svelte
daviddecentage: done

PR #8291, opened 2025-08-18
feat(frontend): added new gradient border to token input component

roman-nazaruk:
# Motivation

For cross chain swaps, we need to display a gradient border that corresponds to the token’s network.

# Changes

Introduced a new component that applies a network-specific gradient border wrapper around the token, based on the selected network.

# Tests

image

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokenInputNetworkWrapper.svelte
+++ b/src/frontend/src/lib/components/tokens/TokenInputNetworkWrapper.svelte
@@ -2,7 +2,6 @@
 	import type { Snippet } from 'svelte';
 	import { isNetworkIdETH } from '$icp/utils/ic-send.utils';
 	import type { NetworkId } from '$lib/types/network';
-	
 	import {
 		isNetworkIdArbitrum,
 		isNetworkIdBase,
@@ -10,13 +9,13 @@
 		isNetworkIdICP,
 		isNetworkIdPolygon
 	} from '$lib/utils/network.utils';
-	
+
 	interface Props {
 		tokenNetworkId?: NetworkId;
 		tokenInput: Snippet;
 		showGradient: boolean;
 	}
-	
+
 	let { tokenNetworkId, tokenInput, showGradient }: Props = $props();
 </script>
 


src/frontend/src/lib/components/tokens/TokenInputNetworkWrapper.svelte
```diff
@@ -0,0 +1,43 @@
...
+		class:to-bsc-100={isNetworkIdBsc(tokenNetworkId)}
+		class:from-polygon-0={isNetworkIdPolygon(tokenNetworkId)}
+		class:to-polygon-100={isNetworkIdPolygon(tokenNetworkId)}
+		class:from-arbitrum-0={isNetworkIdArbitrum(tokenNetworkId)}
+		class:to-arbitrum-100={isNetworkIdArbitrum(tokenNetworkId)}
+		class:from-icp-0={isNetworkIdICP(tokenNetworkId)}
+		class:to-icp-100={isNetworkIdICP(tokenNetworkId)}
+	>
+		{@render tokenInput?.()}
+		<div class="-z-1 absolute bottom-0.5 left-1.5 right-0.5 top-0.5 rounded-xl bg-secondary"></div>
```
DenysKarmazynDFINITY: unless there is a specific need for that, let's maybe use either bottom or top and either left or right values? having 4 of those at the time is a bit unintuitive

src/frontend/src/lib/components/tokens/TokenInputNetworkWrapper.svelte
```diff
@@ -0,0 +1,43 @@
...
+	import type { Snippet } from 'svelte';
+	interface Props {
+		tokenNetworkId?: NetworkId;
+		tokenInput: Snippet;
+		showWrapper: boolean;
+	}
+	let { tokenNetworkId, tokenInput, showWrapper }: Props = $props();
+</script>
+
+{#if showWrapper}
```
DenysKarmazynDFINITY: Maybe `showGradient`? Since the component is called `*Wrapper` I'd expect `showWrapper` to hide the whole component, not just a part of it

src/frontend/src/lib/components/tokens/TokenInputNetworkWrapper.svelte
roman-nazaruk: Done. Thanks

src/frontend/src/lib/components/tokens/TokenInputNetworkWrapper.svelte
roman-nazaruk: Good idea. Agree. Updated

src/frontend/src/lib/components/tokens/TokenInputNetworkWrapper.svelte
```diff
@@ -0,0 +1,43 @@
...
+		isNetworkIdBsc,
+		isNetworkIdICP,
+		isNetworkIdPolygon
+	} from '$lib/utils/network.utils';
+	interface Props {
+		tokenNetworkId?: NetworkId;
+		tokenInput: Snippet;
+		showGradient: boolean;
+	}
+	let { tokenNetworkId, tokenInput, showGradient }: Props = $props();
```
daviddecentage:
```suggestion
	import type { NetworkId } from '$lib/types/network';
	
	import {
		isNetworkIdArbitrum,
		isNetworkIdBase,
		isNetworkIdBsc,
		isNetworkIdICP,
		isNetworkIdPolygon
	} from '$lib/utils/network.utils';
	
	interface Props {
		tokenNetworkId?: NetworkId;
		tokenInput: Snippet;
		showGradient: boolean;
	}
	
	let { tokenNetworkId, tokenInput, showGradient }: Props = $props();
```

PR #8293, opened 2025-08-18
feat(frontend): BgImg component

daviddecentage:
# Motivation

Since we need background images regularly for the new Nft pages, it might make sense to have a dedicated component for easily rendering bg images.

# Changes

Added component

# Tests

Added tests


src/frontend/src/lib/components/ui/BgImg.svelte
```diff
@@ -0,0 +1,42 @@
...
+		imageUrl,
+		ariaLabel,
+		shadow = 'inset',
+		size = 'auto',
+		styleClass,
+		children
+	}: Props = $props();
+
+	const { hasWidthClass, hasHeightClass } = $derived.by(() => ({
+		hasWidthClass: styleClass?.includes('w-') ?? false,
```
DenysKarmazynDFINITY, line 24: it would be probably a bit easier if we would have `height` and `width` as optional props

daviddecentage: but then we dont have the possibility to pass tailwind classes however we like and are limited to em or px values, or at least that was my idea with this approach. Wdyt?

DenysKarmazynDFINITY: yeah, I see your point - maybe then we set `h-full` and `w-full` by default, and consumers can overwrite it via `styleClass`?

daviddecentage: You cannot override it when they are set fixed, thats why I only add it when no w- or h- class is in the style classes. Ill merge it like this for now

PR #8301, opened 2025-08-18
feat(frontend): continuously update UtxosFee during review

DecentAgeCoder:
# Motivation

If the user does not immediately confirm a BTC send, the fee rate and available UTXOs may change over time. As a result, the displayed transaction data could become outdated and misleading by the time the user eventually clicks the Send button. Therefore, a continuous update of the UTXOS and fee information is required for as long as the review view is displayed

# Changes

- Added a recurring scheduler (`startScheduler` and `stopScheduler`) to periodically call `updatePrepareBtcSend` every 10 seconds.
- Replaced `selectUtxosFee` with `updatePrepareBtcSend` for better readability and enhanced functionality.
- Integrated cleanup logic on component destruction using `onDestroy` to stop the scheduler.
- Enhanced the initial logic in `onMount` to start the scheduler after loading required data.
- Improved code maintainability by introducing clear timeout management for the scheduler.

# Tests
- Verified the scheduler refreshes fee data every 10 seconds.
- Confirmed proper cleanup of the scheduler on component destruction.


src/frontend/src/btc/components/send/BtcUtxosFee.svelte
```diff
@@ -57,10 +60,41 @@
 		}
 	};
 
+	const startScheduler = () => {
+		// Clear any existing timer
+		if (schedulerTimer) {
+			clearTimeout(schedulerTimer);
```
DenysKarmazynDFINITY: let's use `stopScheduler` here

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
```diff
@@ -57,10 +60,41 @@
...
+		}
+
+		// Start the recurring scheduler
+		const scheduleNext = () => {
+			schedulerTimer = setTimeout(async () => {
+				await updatePrepareBtcSend();
+				scheduleNext();
+			}, 30000);
```
DenysKarmazynDFINITY: for the sake of readability, would be great to have this number defined in `btc/constants/btc.constants.ts`

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
```diff
@@ -1,7 +1,8 @@
 <script lang="ts">
 	import { isNullish, nonNullish } from '@dfinity/utils';
-	import { createEventDispatcher, getContext, onMount } from 'svelte';
+	import { createEventDispatcher, getContext, onMount, onDestroy } from 'svelte';
 	import { prepareBtcSend } from '$btc/services/btc-utxos.service';
+	import { UTXOS_FEE_CONTEXT_KEY, type UtxosFeeContext } from '$btc/stores/utxos-fee.store';
```
DenysKarmazynDFINITY: some unused imports :)

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
DecentAgeCoder: I have added a stopScheduler call

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
DecentAgeCoder: I have removed unused imports

src/frontend/src/btc/components/send/BtcUtxosFee.svelte
DecentAgeCoder: I have introduced constant BTC_UTXOS_FEE_UPDATE_INTERVAL

src/frontend/src/btc/constants/btc.constants.ts
```diff
@@ -15,3 +15,5 @@ export const BTC_CONVERT_FEE = 0n;
 export const DEFAULT_BTC_AMOUNT_FOR_UTXOS_FEE = 0.00001;
 export const BTC_AMOUNT_FOR_UTXOS_FEE_UPDATE_PROPORTION = 10;
 export const BTC_MINIMUM_AMOUNT = 700n;
+
+export const BTC_UTXOS_FEE_UPDATE_INTERVAL = 10000;
```
DenysKarmazynDFINITY, line 19: In the previous PR version, it was 30s. Was the change to 10s intended?

DecentAgeCoder: Yes because ethereum also uses 10 seconds. But since the blocktime of BTC is much higher, I suggest to set it to 30 seconds with the folow-up PR.

PR #8279, opened 2025-08-14
feat(frontend): add BTC send validation

DecentAgeCoder:
# Motivation

This PR enhances the Bitcoin sending functionality by introducing robust validations to prevent invalid transactions, improve error handling, and ensure accurate fee calculation. Sending Bitcoin transactions was prone to failures due to insufficient balance, locked UTXOs, and invalid UTXO data. By fixing these problems, users will experience fewer errors, smoother transactions, and better feedback for invalid states.

# Changes

- Added `validateBtcSend` function to robustly validate all aspects of Bitcoin transaction inputs before sending.
- Implemented detailed error handling using `BtcValidationError` to categorize and provide precise feedback for various types of invalid inputs or scenarios.
- Updated `BtcSendTokenWizard.svelte` to use the newly introduced validation, replacing multiple individual checks.
- Refactored and extended error messages for localization files (e.g., Czech, English, German) to include details for new validation errors like locked UTXOs, insufficient balance for fees, and invalid fee calculations.
- Introduced a shared `handleBtcValidationError` function for consistent error-handling logic across components.

# Tests
- Unit tests added for validateBtcSend to cover all validation scenarios extensively.
- Edge test cases for BtcSendWizard scenarios were verified.
- New mocks to simulate pending transactions, fee estimation, and UTXO scenarios for broad test coverage.
- Verified proper error messages based on validation results by introducing i18n-specific error tests.

PR #8309, opened 2025-08-18
feat(frontend): add BTC validation errors, constants, translations, and utils used for validating before btc send 

DecentAgeCoder:
# Motivation

This PR (1/2) enhances the Bitcoin sending functionality by introducing robust validations to prevent invalid transactions, improve error handling, and ensure accurate fee calculation. Sending Bitcoin transactions was prone to failures due to insufficient balance, locked UTXOs, and invalid UTXO data. By fixing these problems, users will experience fewer errors, smoother transactions, and better feedback for invalid states.

**The follow-up PR will provide the main validation logic and comprehensive tests.**

The PR containing all the changes for this PR-series can be found here: #8279 

# Changes

- Added new constants, including `BTC_SEND_FEE_TOLERANCE_PERCENTAGE`, for flexible fee tolerance adjustments.
- Implemented `handleBtcValidationError` function to centralize and manage Bitcoin transaction validation errors with detailed error handling.
- Introduced a `BtcValidationError` class and expanded `BtcSendValidationError` enum with error cases like `InvalidFeeCalculation`, `UtxoLocked`, and others.
- Updated the pending transaction utility logic with new functions like `convertPendingTransactionTxid`.
- Updated internationalization files (cs.json, de.json, en.json, etc.) to include additional error messages for Bitcoin transaction issues.
- Enhanced TypeScript typing in `i18n.d.ts` for newly added validation errors.

# Tests

- Confirmed updated translations for multiple languages work with the new validation errors.
- Ensured existing utilities like transaction ID retrieval still function after refactoring.


src/frontend/src/btc/services/btc-send.services.ts
```diff
@@ -24,6 +32,92 @@ export type SendBtcParams = BtcSendServiceParams & {
...
+ * This function handles the validation errors thrown by the validateUtxosForSend function
+ * It has been moved to a component so it can be shared between the BTC Send and Convert Wizard
+ * @param err BtcValidationError - The validation error that was thrown
+ * @param $i18n I18n - The i18n store containing translation strings
+ * @returns Promise<void> - Returns void if successful, may throw errors if validation fails
+ */
+export const handleBtcValidationError = async ({ err }: { err: BtcValidationError }) => {
+	// Safety check: ensure error object exists and has the expected structure
+	if (!err || typeof err !== 'object' || !('type' in err) || !err.type) {
+		console.error('Invalid error object passed to handleBtcValidationError:', err);
```
DenysKarmazynDFINITY: Let's not add a console.error when the same info can be taken from the toast

src/frontend/src/btc/services/btc-send.services.ts
```diff
@@ -24,6 +32,92 @@ export type SendBtcParams = BtcSendServiceParams & {
 	onProgress?: () => void;
 };
 
+/**
+ * This function handles the validation errors thrown by the validateUtxosForSend function
+ * It has been moved to a component so it can be shared between the BTC Send and Convert Wizard
```
DenysKarmazynDFINITY: You probably meant "moved to a service"?

src/frontend/src/btc/services/btc-send.services.ts
```diff
@@ -24,6 +32,92 @@ export type SendBtcParams = BtcSendServiceParams & {
...
+ * @param err BtcValidationError - The validation error that was thrown
+ * @param $i18n I18n - The i18n store containing translation strings
+ * @returns Promise<void> - Returns void if successful, may throw errors if validation fails
+ */
+export const handleBtcValidationError = async ({ err }: { err: BtcValidationError }) => {
+	// Safety check: ensure error object exists and has the expected structure
+	if (!err || typeof err !== 'object' || !('type' in err) || !err.type) {
```
DenysKarmazynDFINITY: If you expect `err` to be falsy in some cases, please update its type accordingly (hence making it optional)

also, we prefer to use `isNullish(err)` instead of `!err` (unless it's a boolean)

DecentAgeCoder: The cleanest approach would be to remove the checks entirely and trust the type system, what I did.

src/frontend/src/btc/services/btc-send.services.ts
DecentAgeCoder: Good catch! I have corrected the type.

src/frontend/src/btc/services/btc-send.services.ts
DecentAgeCoder: I have removed the console error statement

src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -13,36 +14,55 @@ export const utxoTxIdToString = (txid: Uint8Array | number[]): string =>
...
-			txidString = String(tx.txid);
-		}
-
-		// Add to accumulator only if truthy (combines map + filter(Boolean))
-		return notEmptyString(txidString) ? [...acc, txidString] : acc;
-	}, []);
+	// Use the utility function to convert txids and filter out nulls
+	return pendingTransactions.data
+		.map(convertPendingTransactionTxid)
+		.filter((txid): txid is string => txid !== null);
```
```suggestion
		.filter((txid): txid is string => nonNullish(txid);
```

src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -13,36 +14,55 @@ export const utxoTxIdToString = (txid: Uint8Array | number[]): string =>
...
+	}
+
+	// Return the txid string only if it's not empty
+	return notEmptyString(txidString) ? txidString : null;
+};
+
+export const getPendingTransactions = (
+	address: string
+): Omit<CertifiedData<Array<PendingTransaction> | null>, 'certified'> & {
+	certified: true;
```
DenysKarmazynDFINITY, line 46: wondering what's the reason for this type modifications?

DenysKarmazynDFINITY (approved):
Small question - otherwise LGTM (considering that we'll add tests in the follow-up PR)

src/frontend/src/icp/utils/btc.utils.ts
DecentAgeCoder: Did you mean why I defined the `Omit<CertifiedData<Array<PendingTransaction> | null>, 'certified'> & {`return type? If matches type of pendingTransactions
Or did you mean I have refactored the functions? Than: The main reason for this change is to avoid redundancy between `getPendingTransactions` and `getPendingTransactionIds`

src/frontend/src/icp/utils/btc.utils.ts
DecentAgeCoder: Thanks!

PR #8323, opened 2025-08-19
feat(frontend): Load SPL custom tokens with query and update

AntonioVentilii:
# Motivation

It does not make sense the only SPL tokens are being loaded with query. We should use the default strategy that is both query and update, like the other custom tokens.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8324, opened 2025-08-19
fix(frontend): Use mapper for IC errors during saving of custom tokens

AntonioVentilii:
# Motivation

To have an easier analysis of the errors, we use the mapper for the IC errors when saving custom tokens.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8325, opened 2025-08-19
feat(frontend): Nft details page

daviddecentage:
# Motivation

Here we add the Nft details page

# Changes

- Added component
- Added route
- Added translations
- Small route change on the nft cards
- Changed condition of layout hero to not display the default hero on NFT pages

# Tests

- Added tests

image


src/frontend/src/lib/components/nfts/Nft.svelte
```diff
@@ -0,0 +1,48 @@
...
+	const [networkId, collectionId, nftId] = $derived([
+		page.params.networkId,
+		page.params.collectionId,
+		page.params.nftId
+	]);
+
+	const nft: Nft | undefined = $derived(
+		($nftStore ?? []).find(
+			(nft) =>
+				String(nft.id) === nftId &&
```
BonomoAlessandro: It would be better if you check it like this: `nft.id === parseNftId(nftId)`

src/frontend/src/lib/components/nfts/Nft.svelte
```diff
@@ -0,0 +1,48 @@
...
+				nft.collection.address === collectionId &&
+				nft.collection.network.name === networkId
```
BonomoAlessandro, line 26: Is this check correct? It looks like you compare the networks `name` with its `id`

daviddecentage: The route params are intentionally always called id since we dont always want the id in the URL

src/frontend/src/lib/components/nfts/Nft.svelte
daviddecentage: updated

BonomoAlessandro (approved):
LGTM

PR #8328, opened 2025-08-19
feat(frontend): implements getUpdatedNfts utils function

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/utils/nfts.utils.ts
+++ b/src/frontend/src/lib/utils/nfts.utils.ts
@@ -7,7 +7,7 @@ import type {
 	NftCollectionUi,
 	NftId,
 	NftsByNetwork,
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -88,6 +89,36 @@ export const findRemovedNfts = ({
...
+	nfts,
+	token,
+	inventory
+}: {
+	nfts: Nft[];
+	token: NonFungibleToken;
+	inventory: OwnedNft[];
+}): Nft[] =>
+	(nfts ?? [])
+		.filter((nft) => {
```
daviddecentage: can you do a single reduce instead of filter + map?

PR #8329, opened 2025-08-19
chore(frontend): Extend `sprinkles_s1e4` campaign to 31 August

AntonioVentilii:
# Motivation

We want to extend the campaign `sprinkles_s1e4` to the 31st of August.


StefanBerger-DFINITY (approved):
s'all good man

PR #8344, opened 2025-08-20
feat(frontend): implements LoaderNfts

BonomoAlessandro:
# Motivation

When the application is running we want to check in an interval of 2 minutes if the user received some new erc721 nfts. If new nfts were received, we want to load them.

# Changes

- implements `LoaderNfts`


daviddecentage (approved):
LGTM

PR #8349, opened 2025-08-20
feat(frontend): Generalize ICP wallet workers for any token

AntonioVentilii:
# Motivation

We are going to onboard more tokens with standard `icp` (as defined in the the codebase). That means that we need to generalize all the services and workers for this type of standard, since right now they are "hard-coded" to be used only by ICP token.

In this PR, we generalize the input data for the ICP wallet worker.

# Changes

- Worker initializer now accepts a token as input (similar to the other IC workers).
- Remove hard-coded ICP token data.
- Adapt generic IC wallet worker initializer to accept `token` as input.

# Tests

Adapted tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8350, opened 2025-08-20
feat(frontend): Generalize transfer methods for ICP-standard tokens

AntonioVentilii:
# Motivation

We are going to onboard more tokens with standard `icp` (as defined in the the codebase). That means that we need to generalize all the services and workers for this type of standard, since right now they are "hard-coded" to be used only by ICP token.

In this PR, we generalize the transfer methods of the ICP-standard tokens (similar to the other IC methods).

# Changes

- Modify methods `transfer` and `icrc1Transfer` of ICP ledger to accept the ledger canister ID as param.
- Adapt initializer of the ICP ledger canister.
- Adapt send service for ICP tokens.

# Tests

Adapted current tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8360, opened 2025-08-20
feat(frontend): added new SwapEthFeeInfo component

roman-nazaruk:
# Motivation

Add support for displaying EthFeeInfo component in the swap modal.

# Changes

- Added new component `EthFeeInfo .svelte`

# Tests

Covered with the test new component


src/frontend/src/eth/components/swap/SwapEthFeeInfo.svelte
```diff
@@ -0,0 +1,41 @@
+<script lang="ts">
```
DenysKarmazynDFINITY, line 1: Since the component is almost identical with SendFeeInfo, let's maybe pass `token` and `styleClass` (the send flow needs that additional mt-6) and use a single component for both swap and send?

src/frontend/src/lib/components/swap/SwapProviderListModal.svelte
```diff
@@ -68,11 +68,11 @@
 					<SwapProviderListItem
 						amount={swap.receiveAmount}
 						dapp={dAppDescriptions.find(({ id }) => id === swap.provider.toLowerCase())}
-						destinationToken={$destinationToken}
+						destinationToken={$destinationToken as IcTokenToggleable}
```
DenysKarmazynDFINITY, line 71: small note: as discussed offline, the type casting will be changes in the follow up PRs

src/frontend/src/eth/components/swap/SwapEthFeeInfo.svelte
DenysKarmazynDFINITY: upd: there will be further changes to SwapEthFeeInfo in the next PRs, therefore we decided to keep it separated from the send flow

roman-nazaruk: Agreed. Thanks

src/frontend/src/lib/components/swap/SwapProviderListModal.svelte
roman-nazaruk: Thanks

PR #8361, opened 2025-08-20
feat(frontend): added new findSwapProvider util method

roman-nazaruk:
# Motivation

We would need to add new config for extedrnal Dapps, so in this Pr is a mutil method that allow to filter new Providers

# Changes

Added new util method findSwapProvider

# Tests

Covered new logic with the tests

### DIFF
--- a/src/frontend/src/lib/utils/swap.utils.ts
+++ b/src/frontend/src/lib/utils/swap.utils.ts
@@ -2,12 +2,18 @@ import type {
 	SwapAmountsReply,
 	SwapAmountsTxReply
 } from '$declarations/kong_backend/kong_backend.did';
+import { dAppDescriptions } from '$env/dapp-descriptions.env';
+import type { Erc20Token } from '$eth/types/erc20';
+import { isDefaultEthereumToken } from '$eth/utils/eth.utils';
+import type { IcTokenToggleable } from '$icp/types/ic-token-toggleable';
 import { isIcToken } from '$icp/validation/ic-token.validation';
 import { ZERO } from '$lib/constants/app.constants';
 import {
 	SWAP_DEFAULT_SLIPPAGE_VALUE,
 	SWAP_ETH_TOKEN_PLACEHOLDER
 } from '$lib/constants/swap.constants';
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/utils/swap.utils.ts
```diff
@@ -206,3 +207,27 @@ export const getWithdrawableToken = ({
 
 	throw new Error(`Unknown token address`);
 };
+
+export const findSwapProvider = (providerId: string) => {
```
DenysKarmazynDFINITY: can you please add all the possible return types?

src/frontend/src/lib/utils/swap.utils.ts
```diff
@@ -206,3 +207,27 @@ export const getWithdrawableToken = ({
...
+	const swapProviderDetails = swapProvidersDetails[normalizedId];
+
+	if (swapProviderDetails) {
+		return {
+			id: normalizedId,
+			...swapProviderDetails
+		};
+	}
+
+	return;
```
DenysKarmazynDFINITY: it can be removed

src/frontend/src/lib/utils/swap.utils.ts
```diff
@@ -206,3 +207,27 @@ export const getWithdrawableToken = ({
 
 	throw new Error(`Unknown token address`);
 };
+
+export const findSwapProvider = (providerId: string) => {
+	if (isNullish(providerId)) {
```
DenysKarmazynDFINITY: According to the providerId type, it's always a string - please either remove isNullish check or update the type

src/frontend/src/lib/utils/swap.utils.ts
roman-nazaruk: Done

src/frontend/src/lib/utils/swap.utils.ts
roman-nazaruk: done

src/frontend/src/lib/utils/swap.utils.ts
roman-nazaruk: Removed

PR #8362, opened 2025-08-20
feat(frontend): handles updated nfts

BonomoAlessandro:
# Motivation

When the application is running we want to check in an interval of 2 minutes if the user has some updated nfts. If nfts were updated, we want to also update them in the application.


daviddecentage (approved):
LGTM

PR #8366, opened 2025-08-20
feat(frontend): updated SwapProvider component

roman-nazaruk:
# Motivation

We would need to extend SwapProvider component to show Valora Dex info.

# Changes

Added new child component to render VeloraSwapDetails component


BonomoAlessandro (approved):
LGTM

PR #8371, opened 2025-08-20
feat(frontend): Nft image consent

daviddecentage:
# Motivation

Here we add the NftImageConsent component that wraps the media rendering elements and displays the review consent if no the collection has not been allowed first.

# Changes

- Added texts
- Added NftImageConsent component
- Adjusted all media rendering elements by wrapping them with the new component

# Tests

Adjusted all tests rendering media. We now mock the allow flag so we can test them for the case that media is allowed, as all cases for when its not allowed should be handled by the test added for this new component that I added in this PR.


BonomoAlessandro (approved):
Nice, LGTM

PR #8375, opened 2025-08-20
feat(frontend): Nft description

daviddecentage:
# Motivation

We want to display the NFTs description if available.

# Changes

- Added component
- Integrated component on Nft page

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftDescription.svelte
+++ b/src/frontend/src/lib/components/nfts/NftDescription.svelte
@@ -11,7 +11,7 @@
 	const { nft }: Props = $props();
 </script>
 
-{#if nonNullish(nft?.description)) || isNullish(nft)}
+{#if nonNullish(nft?.description) || isNullish(nft)}
 	<div class="mt-5 flex rounded-lg bg-primary p-5">
 		{#if nonNullish(nft)}
 			<div class="grow-1 flex flex-col">


src/frontend/src/lib/components/nfts/NftDescription.svelte
```diff
@@ -0,0 +1,39 @@
...
+	import type { Nft } from '$lib/types/nft';
+
+	interface Props {
+		nft?: Nft;
+	}
+
+	const { nft }: Props = $props();
+</script>
+
+{#if (nonNullish(nft) && nonNullish(nft.description)) || isNullish(nft)}
```
DenysKarmazynDFINITY:
```suggestion
{#if nonNullish(nft?.description)) || isNullish(nft)}
```

src/frontend/src/lib/components/nfts/NftDescription.svelte
```diff
@@ -0,0 +1,39 @@
...
+	<div class="mt-5 flex rounded-lg bg-primary p-5">
+		{#if nonNullish(nft)}
+			<div class="flex flex-col">
+				{#if nonNullish(nft.name)}
+					<h5 class="mb-3">{nft.name}</h5>
+				{/if}
+				{#if nonNullish(nft.name)}
```
DenysKarmazynDFINITY:
```suggestion
				{#if nonNullish(nft.description)}
```

src/frontend/src/lib/components/nfts/NftDescription.svelte
DenysKarmazynDFINITY: is this check really needed though? seems like it will always end up inside this if

DenysKarmazynDFINITY (approved):
Small questions - LGTM otherwise

src/frontend/src/lib/components/nfts/NftDescription.svelte
daviddecentage: Not always. When nft is not nullish and description is missing then we dont render the card at all

PR #8376, opened 2025-08-20
feat(frontend): add model and utilities for new BTC balance calculation

DecentAgeCoder:
# Motivation

This is one of several PRs introducing the new BTC wallet balance model, designed to enhance transparency and accuracy in Bitcoin balances while aligning with standardized Bitcoin practices.
This PR  adds new utilities for Bitcoin network handling. 

# Changes

- Added a new structured `BtcWalletBalance` interface to represent Bitcoin balances, including confirmed, unconfirmed, locked, and total balances.
- Implemented a `getBtcWalletBalance` function to calculate and structure Bitcoin balances, factoring in pending transactions and confirmation statuses.
- Added utility for mapping Bitcoin Token IDs to Network IDs (`mapTokenIdToNetworkId`) that will be used by `btc-listener.services.ts`  to obtain the `networkId` for the incoming `tokenId`.
- Introduced `getBtcSourceAddress` for retrieving source addresses based on network IDs that will be used by `btc-listener.services.ts to` obtain the source address from the incoming network.
- Updated imports and module dependencies to support network and balance functionalities.

# Tests

- The PR itself does not introduce any test cases, but the balance calculation functions and utilities are provided in the follow-up PR introduces unit test to ensure correctness.

### DIFF
--- a/src/frontend/src/eth/utils/eip712.utils.ts
+++ b/src/frontend/src/eth/utils/eip712.utils.ts
@@ -1,5 +1,6 @@
 import type { SignableDeltaOrderData } from '@velora-dex/sdk';
-import { Signature, TypedDataEncoder } from 'ethers';
+import { Signature } from 'ethers/crypto';
+import { TypedDataEncoder } from 'ethers/hash';
 
 /**
  * Generates the EIP-712 hash of typed structured data, used for signing with `eth_signTypedData`.
--- a/src/frontend/src/lib/constants/swap.constants.ts
+++ b/src/frontend/src/lib/constants/swap.constants.ts
@@ -26,6 +26,9 @@ export const ICP_SWAP_POOL_FEE = 3000n;
 
 export const SWAP_ETH_TOKEN_PLACEHOLDER = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE';
 
+export const SWAP_DELTA_TIMEOUT_MS = 5 * 60_000;
+export const SWAP_DELTA_INTERVAL_MS = 3_000;
+
 export const OISY_DOCS_SWAP_WIDTHDRAW_FROM_ICPSWAP_LINK =
 	'https://docs.oisy.com/using-oisy-wallet/how-tos/swapping-tokens#manually-withdraw-funds-from-icpswap';
 
--- a/src/frontend/src/lib/services/swap.services.ts
+++ b/src/frontend/src/lib/services/swap.services.ts
@@ -1,4 +1,6 @@
 import type { SwapAmountsReply } from '$declarations/kong_backend/kong_backend.did';
+import { approve as approveToken } from '$eth/services/send.services';
+import { getCompactSignature, getSignParamsEIP712 } from '$eth/utils/eip712.utils';
 import { setCustomToken as setCustomIcrcToken } from '$icp-eth/services/custom-token.services';
 import { approve } from '$icp/api/icrc-ledger.api';
 import { sendIcp, sendIcrc } from '$icp/services/ic-send.services';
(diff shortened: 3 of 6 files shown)


src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -1,9 +1,41 @@
...
 import { isNullish, nonNullish, notEmptyString, uint8ArrayToHexString } from '@dfinity/utils';
 import { get } from 'svelte/store';
 
+/**
+ * Get the NetworkId from a BTC TokenId
+ * @param tokenId - The BTC token ID
+ * @returns The corresponding NetworkId
+ */
+export const mapTokenIdToNetworkId = (tokenId: TokenId): NetworkId | undefined => {
+	if (tokenId === BTC_MAINNET_TOKEN_ID) {
```
DenysKarmazynDFINITY: please use the same functional approach with ternary operators as in getBtcSourceAddress

src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -64,3 +96,114 @@ export const getPendingTransactionIds = (address: string): string[] => {
...
+}): BtcWalletBalance => {
+	// Retrieve pending outgoing transactions from local store with safe fallback
+	// If the store access fails or returns invalid data, we'll default to empty state
+	let pendingTransactions: Array<PendingTransaction> = [];
+
+	try {
+		const pendingData = getPendingTransactions(address);
+
+		// Handle the various states the store data can be in
+		if (nonNullish(pendingData) && nonNullish(pendingData.data)) {
```
```suggestion
		if (nonNullish(pendingData?.data)) {
```

src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -64,3 +96,114 @@ export const getPendingTransactionIds = (address: string): string[] => {
...
+
+	// Calculate locked balance: UTXOs being used as inputs in pending outgoing transactions
+	// If pendingTransactions is empty (due to error or no data), locked balance will be 0
+	const lockedBalance = pendingTransactions.reduce((sum, tx) => {
+		try {
+			// Safely calculate UTXO sum with additional error handling
+			const txUtxoValue = nonNullish(tx.utxos)
+				? tx.utxos.reduce((utxoSum, utxo) => {
+						// Ensure utxo.value is valid before adding
+						const utxoValue = nonNullish(utxo?.value) ? BigInt(utxo.value) : 0n;
```
```suggestion
						const utxoValue = nonNullish(utxo?.value) ? BigInt(utxo.value) : ZERO;
```

DenysKarmazynDFINITY: and all other 0n appearances, please

src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -64,3 +96,114 @@ export const getPendingTransactionIds = (address: string): string[] => {
...
+	let pendingTransactions: Array<PendingTransaction> = [];
+
+	try {
+		const pendingData = getPendingTransactions(address);
+
+		// Handle the various states the store data can be in
+		if (nonNullish(pendingData) && nonNullish(pendingData.data)) {
+			pendingTransactions = pendingData.data;
+		}
+	} catch (error) {
```
DenysKarmazynDFINITY: I'm not sure what do we try to catch here - I don't think anything in the try block can fail with an error

src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -64,3 +96,114 @@ export const getPendingTransactionIds = (address: string): string[] => {
...
+			const txUtxoValue = nonNullish(tx.utxos)
+				? tx.utxos.reduce((utxoSum, utxo) => {
+						// Ensure utxo.value is valid before adding
+						const utxoValue = nonNullish(utxo?.value) ? BigInt(utxo.value) : 0n;
+						return utxoSum + utxoValue;
+					}, 0n)
+				: 0n;
+
+			return sum + txUtxoValue;
+		} catch (error) {
```
DenysKarmazynDFINITY: here, too - potentially, BigInt(*float number*) can throw an error but according to our types this should never happen

src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -64,3 +96,114 @@ export const getPendingTransactionIds = (address: string): string[] => {
...
+				try {
+					if (
+						tx.data.status === 'unconfirmed' &&
+						tx.data.type === 'receive' &&
+						nonNullish(tx.data.value)
+					) {
+						return sum + tx.data.value;
+					}
+					return sum;
+				} catch (error) {
```
DenysKarmazynDFINITY: same question

src/frontend/src/icp/utils/btc.utils.ts
DecentAgeCoder: I have adapted the code

src/frontend/src/icp/utils/btc.utils.ts
DecentAgeCoder: Yes that is a leftover from the direct canister call. I have removed the try catch.

src/frontend/src/icp/utils/btc.utils.ts
DecentAgeCoder: I fully agree

src/frontend/src/icp/utils/btc.utils.ts
DecentAgeCoder: I fully agree here as well

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8378, opened 2025-08-21
feat(frontend): UTXO should be selected with at least 6 confirmations

DecentAgeCoder:
# Motivation

The PR aims to ensure that the selection of  theUTXO when preparing the BTC send must have at least 6 confirmations. This change improves the reliability of BTC transaction handling.

# Changes

- Replaced the incorrect usage of `UNCONFIRMED_BTC_TRANSACTION_MIN_CONFIRMATIONS` with `CONFIRMED_BTC_TRANSACTION_MIN_CONFIRMATIONS` in the `btc-utxos.service.ts` file.
- Updated the test case in `btc-utxos.service.spec.ts` to align with the change, setting `minConfirmations` to 6 instead of 1.

# Tests

- Updated the test file `btc-utxos.service.spec.ts` to match the new constant by altering the expected value for `minConfirmations` to ensure proper validation of confirmed transactions.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8379, opened 2025-08-21
feat(frontend): added new SwapEthForm.svelte component

roman-nazaruk:
# Motivation

This PR introduces a new SwapEthForm component, that is main Wrapper for SwapForm but works only with Eth networks.

# Changes

Added new NewSwapForm component

# Tests

This component only renders another components that covered with the tests.


roman-nazaruk:
> let's add some unit tests for the new component

Of course, done.

src/frontend/src/eth/components/swap/SwapEthForm.svelte
```diff
@@ -0,0 +1,144 @@
...
+	import type { OptionAmount } from '$lib/types/send';
+	import type { Token } from '$lib/types/token';
+	import type { TokenActionErrorType } from '$lib/types/token-action';
+	import { formatToken } from '$lib/utils/format.utils';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+	import { parseToken } from '$lib/utils/parse.utils';
+
+	interface Props {
+		swapAmount: OptionAmount;
+		receiveAmount: number | undefined;
```
DenysKarmazynDFINITY:
```suggestion
		receiveAmount?: number;
```

src/frontend/src/eth/components/swap/SwapEthForm.svelte
```diff
@@ -0,0 +1,144 @@
...
+	const {
+		feeStore: storeFeeData,
+		minGasFee,
+		maxGasFee,
+		feeSymbolStore,
+		feeDecimalsStore,
+		feeTokenIdStore
+	} = getContext<EthFeeContext>(ETH_FEE_CONTEXT_KEY);
+
+	const customValidate = (userAmount: bigint): TokenActionErrorType | undefined => {
```
DenysKarmazynDFINITY, line 61: is this going to change in the next PRs? if not, we need to re-use it as a util (if it's not possible to use EthSendAmount component here) since it seems the same logic which we have in `EthSendAmount`

roman-nazaruk: Probably yes, but it is not similar to send customValidate as this logic return statuses as string and do not throw errors.

DenysKarmazynDFINITY: I see - still, there are some logic that can be moved into several smaller "assert" utils, similar as it's done in soma other places. No need to do it now, will be one more point for our backlog.

PR #8380, opened 2025-08-21
feat(frontend): handle pending transactions load failed or not initialized case

DecentAgeCoder:
# Motivation

We need to clearly distinguish between “pending transaction load failed or was not initialized” and “no pending transactions available” (i.e., no locked transactions). This distinction is essential for correctly implementing safeguards against double-spend issues.

# Changes

- **Improved Validation in `btc-send.services.ts`:**
    - Added a check to throw errors when pending transactions are missing or uninitialized during UTXO validation.
- **Error Handling in `btc-utxos.service.ts`:**
    - Added error handling to handle uninitialized pending transaction stores.
    - New logic to prevent null values from causing issues in UTXO-related operations.
- **Introduced New Bitcoin Wallet Balance Type (`btc.ts`):**
    - Created a structured `BtcWalletBalance` interface for detailed Bitcoin balance representation.
    - Documented usage and breakdowns for confirmed, unconfirmed, locked, and total balances.
- **Added Utilities in `btc-address.utils.ts` & `btc.utils.ts`:**
    - Added a mapping function for token IDs to network IDs.
    - Included functions to calculate Bitcoin wallet balances with validation for transaction data.
- **Tests Updates (`btc-utxos.service.spec.ts`, `btc.utils.spec.ts`):**
    - Added tests for error handling when pending transaction stores are uninitialized or invalid.
    - Extended test coverage for UTXO and transaction validation scenarios.

# Tests

- Extended unit tests in `btc-utxos.service.spec.ts` to cover new validation flows.
- Updated tests for `btc.utils.spec.ts` to reflect changes in error handling and return values for null cases.
- Verified balance calculations and transaction preparation logic with mock data.

### DIFF
--- a/src/frontend/src/icp/utils/btc.utils.ts
+++ b/src/frontend/src/icp/utils/btc.utils.ts
@@ -141,10 +141,10 @@ export const getPendingTransactionIds = (address: string): string[] | null => {
  * @returns Structured balance object with confirmed, unconfirmed, locked, and total amounts
  */
 export const getBtcWalletBalance = ({
-																			address,
-																			confirmedBalance,
-																			providerTransactions
-																		}: {
+	address,
+	confirmedBalance,
+	providerTransactions
+}: {
 	address: string;
 	confirmedBalance: bigint;
 	providerTransactions: CertifiedData<BtcTransactionUi>[] | null;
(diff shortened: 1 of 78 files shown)


src/frontend/src/icp/utils/btc.utils.ts
```diff
@@ -68,23 +68,53 @@ export const convertPendingTransactionTxid = (tx: PendingTransaction): string |
...
 	certified: true;
 } => {
 	const storeData = get(btcPendingSentTransactionsStore);
-	return storeData[address];
+
+	// Case 1: Store not initialized, return safe default
+	if (isNullish(storeData) || Object.keys(storeData).length === 0) {
+		return {
+			data: null,
+			certified: true
```
DenysKarmazynDFINITY: code-wise, it most likely does not make a difference, but I suggest we only save it as `certified` when it comes from BE

DecentAgeCoder: Than i rather completely change the return to Array<PendingTransaction>. There is no point in returning a CertifiedData data structure in that case.

PR #8384, opened 2025-08-21
feat(backend): contacts update image tests

SYInacta:
# Motivation

# Changes

# Tests


src/shared/src/types/tests.rs
```diff
@@ -171,17 +171,64 @@ mod bitcoin {
     );
 }
 
-mod custom_token {
-    //! Tests for the `custom_token` module.
+// removed empty custom_token module with unused imports
```
DecentAgeCoder: Remove this comment

src/shared/src/types/contact.rs
```diff
@@ -5,15 +5,15 @@ use serde_bytes::ByteBuf;
...
-pub const MAX_IMAGE_SIZE_BYTES: usize = 100 * 1024;
-
 /// Maximum number of images per principal (100)
 pub const MAX_IMAGES_PER_PRINCIPAL: usize = 100;
 
 /// Memory usage threshold (80%) above which new images cannot be added
 pub const MEMORY_USAGE_THRESHOLD: f64 = 0.8;
 
+// Re-export image size limit used by validation so tests and external modules can reference it
+pub use crate::impls::MAX_IMAGE_SIZE_BYTES;
```
DecentAgeCoder: I suggest to move the import to line 7

src/shared/src/types/tests.rs
SYInacta: Fixed. No need for second PR.

src/shared/src/types/contact.rs
SYInacta: Fixed. No need for second PR.

PR #8385, opened 2025-08-21
feat(frontend): new SwapIcpWizard component

roman-nazaruk:
# Motivation

This PR introduces a new SwapIcpWizard component, that is main Wrapper for SwapForm but works only with Icp network.

# Changes

Added new SwapIcpWizard component

# Tests

Covered component with tests

### DIFF
--- a/src/frontend/src/icp/components/swap/SwapIcpWizard.svelte
+++ b/src/frontend/src/icp/components/swap/SwapIcpWizard.svelte
@@ -1,0 +1,238 @@
+<script lang="ts">
+	import type { WizardStep } from '@dfinity/gix-components';
+	import { isNullish, nonNullish } from '@dfinity/utils';
+	import { getContext } from 'svelte';
+	import IcTokenFeeContext from '$icp/components/fee/IcTokenFeeContext.svelte';
+	import {
+		IC_TOKEN_FEE_CONTEXT_KEY,
+		type IcTokenFeeContext as IcTokenFeeContextType
+	} from '$icp/stores/ic-token-fee.store';
+	import SwapProgress from '$lib/components/swap/SwapProgress.svelte';
+	import SwapReview from '$lib/components/swap/SwapReview.svelte';
+	import {
+		TRACK_COUNT_SWAP_ERROR,
+		TRACK_COUNT_SWAP_SUCCESS
+	} from '$lib/constants/analytics.contants';
+	import { authIdentity } from '$lib/derived/auth.derived';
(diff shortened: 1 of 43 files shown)


src/frontend/src/icp/components/swap/SwapIcpWizard.svelte
```diff
@@ -0,0 +1,238 @@
...
+	import { errorDetailToString } from '$lib/utils/error.utils';
+	import { replaceOisyPlaceholders, replacePlaceholders } from '$lib/utils/i18n.utils';
+	import { isSwapError } from '$lib/utils/swap.utils';
+
+	interface Props {
+		swapAmount: OptionAmount;
+		receiveAmount?: number;
+		slippageValue: OptionAmount;
+		swapProgressStep: string;
+		swapFailedProgressSteps?: string[];
```
AntonioVentilii: these should be both type-related to `ProgressStepsSwap`, no?

src/frontend/src/icp/components/swap/SwapIcpWizard.svelte
```diff
@@ -0,0 +1,238 @@
...
+					dApp: $swapAmountsStore.selectedProvider.provider,
+					usdSourceValue: sourceTokenUsdValue ?? ''
+				}
+			});
+
+			setTimeout(() => onClose(), 2500);
+		} catch (err: unknown) {
+			const errorDetail = errorDetailToString(err);
+			// TODO: Add unit tests to cover failed swap error scenarios
+			if (nonNullish(errorDetail) && errorDetail.startsWith('Slippage exceeded.')) {
```
AntonioVentilii, line 165: if this logic is used more in the future, let's extract it in a service or util

src/frontend/src/icp/components/swap/SwapIcpWizard.svelte
roman-nazaruk: technically yes

src/frontend/src/icp/components/swap/SwapIcpWizard.svelte
roman-nazaruk: Nope, it will be only here and I am not going to extend it

src/frontend/src/icp/components/swap/SwapIcpWizard.svelte
roman-nazaruk: Fixed. Good catch. Thnx

PR #8386, opened 2025-08-21
style(frontend): Nft cards hover and focus states

daviddecentage:
# Motivation

We got a reworked design for NFT and collection cards from the design team that we are now implementing here.

# Changes

- Changed colors of card bg
- Changed hover behavior by making the root element a group
- Changed hover and focus state styles
- Adjusted spacings
- Adjusted animations on hover

# Tests

- Adjusted a test to not pass correct type undefined instead of null and adjusting the placeholder selector
- Adjusted two tests to check for correct bg-cover instead of bg-contain because it has been changed according to new design

https://github.com/user-attachments/assets/f1e8c9db-8144-486d-a551-f052b633f931


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8388, opened 2025-08-21
feat(frontend): added fetch Velora delta swap service

roman-nazaruk:
# Motivation

We aim to integrate with Velora DEX and for integration we need to add new execute swap with delta contract. 

# Changes

- Added new fetchVeloraDeltaSwap
- Added new checkDeltaOrderStatus  to track OrderStatus
- Added new isExecutedDeltaAuction 

# Tests

Covered new logic with tests.

### DIFF
--- a/src/frontend/src/tests/eth/utils/eip712.utils.spec.ts
+++ b/src/frontend/src/tests/eth/utils/eip712.utils.spec.ts
@@ -1,11 +1,15 @@
 import { getCompactSignature, getSignParamsEIP712 } from '$eth/utils/eip712.utils';
 import type { DeltaAuctionOrder, SignableDeltaOrderData } from '@velora-dex/sdk';
-import { Signature, TypedDataEncoder } from 'ethers';
+import { Signature } from 'ethers/crypto';
+import { TypedDataEncoder } from 'ethers/hash';
 
-vi.mock('ethers', () => ({
+vi.mock('ethers/hash', () => ({
 	TypedDataEncoder: {
 		hash: vi.fn()
-	},
+	}
+}));
+
+vi.mock('ethers/crypto', () => ({
 	Signature: {
 		from: vi.fn()
 	}


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -620,3 +636,128 @@ export const withdrawUserUnusedBalance = async ({
...
+	maxPriorityFeePerGas,
+	swapDetails
+}: SwapVeloraParams): Promise<void> => {
+	const parsedSwapAmount = parseToken({
+		value: `${swapAmount}`,
+		unitName: sourceToken.decimals
+	});
+
+	const sdk = constructSimpleSDK({
+		chainId: Number(sourceNetwork?.chainId),
```
DenysKarmazynDFINITY: Shall we add some fallback value here? Wondering how SDK will behave if chainId will be passed as `NaN`. 

P.S.: I checked the types, and seems `sourceNetwork` is not an optional param, so let's remove `?`

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -620,3 +636,128 @@ export const withdrawUserUnusedBalance = async ({
...
+		progress,
+		progressSteps: ProgressStepsSwap
+	});
+
+	progress(ProgressStepsSwap.SWAP);
+
+	const signableOrderData = await sdk.delta.buildDeltaOrder({
+		deltaPrice: swapDetails as DeltaPrice | BridgePrice,
+		owner: userAddress,
+		srcToken: sourceToken?.address,
```
DenysKarmazynDFINITY: tokens also seem to be required params

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -620,3 +636,128 @@ export const withdrawUserUnusedBalance = async ({
...
+		destToken: destinationToken?.address,
+		srcAmount: `${parsedSwapAmount}`,
+		destAmount: `${slippageMinimum}`,
+		destChainId: Number(destinationNetwork?.chainId)
```
DenysKarmazynDFINITY: same

src/frontend/src/lib/types/swap.ts
```diff
@@ -194,3 +203,25 @@ export interface SwapProvidersConfig {
 	logo: string;
 	website: string;
 }
+
+export interface SwapVeloraParams extends RequiredTransactionFeeData {
+	identity: Identity;
+	progress: (step: ProgressStep) => void;
+	sourceToken: Erc20Token;
```
DenysKarmazynDFINITY, line 210: wondering - it's only for ERC 20 tokens?

roman-nazaruk: I assume yes, but we also support native Eth token in this type

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Fixed

roman-nazaruk: About fallback, I assume no, we do not want to fetch tokens with wrong network id. and all method is wrapped as async await, so I assume we are good with that

src/frontend/src/eth/utils/eip712.utils.ts
```diff
@@ -1,5 +1,6 @@
 import type { SignableDeltaOrderData } from '@velora-dex/sdk';
-import { Signature, TypedDataEncoder } from 'ethers';
+import { Signature } from 'ethers/crypto';
+import { TypedDataEncoder } from 'ethers/hash';
```
DenysKarmazynDFINITY, line 3: these changes seem to be unrelated to the PR - rebasing issue?

DenysKarmazynDFINITY (approved):
LGTM, thanks

src/frontend/src/eth/utils/eip712.utils.ts
roman-nazaruk: Test issues, but all works as expected and it is related to logic of this PR

PR #8389, opened 2025-08-21
fix(frontend): Wallet connect modal z-index

daviddecentage:
# Motivation

# Changes

# Tests


src/frontend/src/lib/components/hero/Header.svelte
```diff
@@ -28,8 +28,8 @@
 	class:1.5xl:z-10={$authSignedIn}
 	class:pb-10={$authNotSignedIn}
 	class:sm:pb-8={$authNotSignedIn}
-	class:z-3={!menuOpen && !networkSwitcherOpen && !helpMenuOpen}
-	class:z-4={menuOpen || networkSwitcherOpen || helpMenuOpen}
+	class:z-3={!menuOpen && !networkSwitcherOpen && !helpMenuOpen && !$modalWalletConnect}
+	class:z-4={menuOpen || networkSwitcherOpen || helpMenuOpen || $modalWalletConnect}
```
DenysKarmazynDFINITY, line 32: can it happen with the other modals? if it's possible in theory, shall we check if any of the modals is open?

daviddecentage: It can only happen when any popover/modal (or generally anything absolute positioned) is placed inside the header, as the header creates a new stacking context because of explicit positioning with z-index.

You can read my detailed explanation https://github.com/dfinity/oisy-wallet/pull/8340

Short answer: No, I just solved it for the dropdowns/popover menus but forgot that wallet connect opens a modal which suffers from the same issue.

DenysKarmazynDFINITY: ok, thanks for the info 👍

DenysKarmazynDFINITY (approved):

PR #8398, opened 2025-08-22
fix(frontend): Avoid re-using existing identity keys for `AuthClient` instance

AntonioVentilii:
# Motivation

This PR addresses a security finding from Trail of Bits concerning how `@dfinity/auth-client` persists identity keys in IndexedDB and how those keys are reused on sign-in.

### The issue

When the user signs out (or is signed out), the current `AuthClient` instance removes the delegated identity from IndexedDB. That triggers the `onstorage` event, which calls syncAuthStore(). That function immediately creates a new `AuthClient` by calling `createAuthClient`, which in turn generates a fresh ECDSA identity key and saves it to IndexedDB.

This behavior is expected, and not an outright vulnerability on its own. However, it has the side effect that when the user next signs in, the new `AuthClient` will reuse whatever identity key is already in IndexedDB. This means that if an attacker with access to the browser can plant a key in IndexedDB before the next login, the `AuthClient` will use that key with Internet Identity to obtain a valid signed delegation.

### Why not fix in `agent-js` directly?

The reason keys are stored in IndexedDB at all is to persist the authentication state across browser sessions. If we generated a new in-memory key every time, the user would have to log in with Internet Identity on every page load — a significant degradation of user experience.

Thus, this cannot be solved in auth-client alone without forcing re-logins on every session.

### Solution (suggestion by @ilbertt )

Instead, we fix this on OISY's side by taking control of storage management.

# Changes

- Service `createAuthClient` will now accept a custom IDB storage to pass when creating a new `AuthClient` instance.
- Created new service `safeCreateAuthClient` that deleted any IDB-stored keys before creating a new `AuthClient` instance.
- Use the new service in the code where necessary (we don't use it during sign out, since the real issue may exists eithr on sign in or on sync.

# Test

Tested manually in a test canister and no changes perceived for the user's experience.

### DIFF
--- a/src/frontend/src/lib/stores/auth.store.ts
+++ b/src/frontend/src/lib/stores/auth.store.ts
@@ -7,7 +7,11 @@ import {
 } from '$lib/constants/app.constants';
 import type { OptionIdentity } from '$lib/types/identity';
 import type { Option } from '$lib/types/utils';
-import { createAuthClient, getOptionalDerivationOrigin } from '$lib/utils/auth.utils';
+import {
+	createAuthClient,
+	getOptionalDerivationOrigin,
+	safeCreateAuthClient
+} from '$lib/utils/auth.utils';
 import { popupCenter } from '$lib/utils/window.utils';
 import type { Identity } from '@dfinity/agent';
 import type { AuthClient } from '@dfinity/auth-client';
@@ -40,7 +44,7 @@ const initAuthStore = (): AuthStore => {
 		subscribe,
 
 		sync: async () => {
-			authClient = authClient ?? (await createAuthClient());
+			authClient = authClient ?? (await safeCreateAuthClient());
 			const isAuthenticated: boolean = await authClient.isAuthenticated();
 
 			set({
@@ -51,7 +55,7 @@ const initAuthStore = (): AuthStore => {
 		signIn: ({ domain }: AuthSignInParams) =>
 			// eslint-disable-next-line no-async-promise-executor
 			new Promise<void>(async (resolve, reject) => {
-				authClient = authClient ?? (await createAuthClient());
+				authClient = authClient ?? (await safeCreateAuthClient());
 
 				const identityProvider = nonNullish(INTERNET_IDENTITY_CANISTER_ID)
 					? /apple/i.test(navigator?.vendor)
--- a/src/frontend/src/lib/utils/auth.utils.ts
+++ b/src/frontend/src/lib/utils/auth.utils.ts
@@ -1,11 +1,17 @@
 import { AUTH_ALTERNATIVE_ORIGINS, AUTH_DERIVATION_ORIGIN } from '$lib/constants/app.constants';
 import { isNullishOrEmpty } from '$lib/utils/input.utils';
 import type { Identity } from '@dfinity/agent';
-import { AuthClient } from '@dfinity/auth-client';
+import {
+	AuthClient,
+	IdbStorage,
+	KEY_STORAGE_DELEGATION,
+	type AuthClientStorage
+} from '@dfinity/auth-client';
 import { notEmptyString } from '@dfinity/utils';
 
-export const createAuthClient = (): Promise<AuthClient> =>
+export const createAuthClient = (storage?: AuthClientStorage): Promise<AuthClient> =>
 	AuthClient.create({
+		storage,
 		idleOptions: {
 			disableIdle: true,
 			disableDefaultIdleCallback: true
@@ -13,11 +19,23 @@ export const createAuthClient = (): Promise<AuthClient> =>
 	});
 
 /**
+ * When the user signs out, we create a new AuthClient.
+ * Since agent-js persists identity keys in IndexedDB by default,
+ * if the old keys remain in storage, they could be tampered with and affect the next login.
+ * To ensure each session starts clean and safe, we clear the stored keys before creating a new AuthClient.
+ */
+export const safeCreateAuthClient = async (): Promise<AuthClient> => {
+	const idbStorage: IdbStorage = new IdbStorage();
+	await idbStorage.remove(KEY_STORAGE_DELEGATION);
+	return await createAuthClient(idbStorage);
+};
+
+/**
  * In certain features, we want to execute jobs with the authenticated identity without getting it from the auth.store.
- * This is notably useful for Web Workers which do not have access to the window.
+ * This is notably useful for Web Workers who do not have access to the window.
  */
 export const loadIdentity = async (): Promise<Identity | undefined> => {
-	const authClient = await createAuthClient();
+	const authClient = await safeCreateAuthClient();
 	const authenticated = await authClient.isAuthenticated();
 
 	// Not authenticated therefore we provide no identity as a result
--- a/src/frontend/src/lib/workers/auth.worker.ts
+++ b/src/frontend/src/lib/workers/auth.worker.ts
@@ -1,6 +1,6 @@
 import { AUTH_TIMER_INTERVAL, NANO_SECONDS_IN_MILLISECOND } from '$lib/constants/app.constants';
 import type { PostMessage, PostMessageDataRequest } from '$lib/types/post-message';
-import { createAuthClient } from '$lib/utils/auth.utils';
+import { safeCreateAuthClient } from '$lib/utils/auth.utils';
 import { IdbStorage, KEY_STORAGE_DELEGATION, type AuthClient } from '@dfinity/auth-client';
 import { DelegationChain, isDelegationValid } from '@dfinity/identity';
 
@@ -50,12 +50,12 @@ const onIdleSignOut = async () => {
 };
 
 /**
- * If user is not authenticated - i.e. no identity or anonymous and there is no valid delegation chain, then identity is not valid
+ * If the user is not authenticated - i.e. no identity or anonymous and there is no valid delegation chain, then identity is not valid
  *
  * @returns true if authenticated
  */
 const checkAuthentication = async (): Promise<boolean> => {
-	const authClient: AuthClient = await createAuthClient();
+	const authClient: AuthClient = await safeCreateAuthClient();
 	return authClient.isAuthenticated();
 };
 


src/frontend/src/lib/utils/auth.utils.ts
```diff
@@ -1,20 +1,33 @@
...
 	});
 
+/**
+ * When the user signs out, we create a new AuthClient.
+ * Since agent-js persists identity keys in IndexedDB by default,
+ * if the old keys remain in storage, they could be tampered with and affect the next login.
+ * To ensure each session starts clean and safe, we clear the stored keys before creating a new AuthClient.
+ */
+export const safeCreateAuthClient = async (): Promise<AuthClient> => {
+	await idbStorage.remove(KEY_STORAGE_DELEGATION);
```
ilbertt: What's important is to clean the session key rather than the delegation, because that's the one that signs messages.
```suggestion
	await idbStorage.remove(KEY_STORAGE_KEY);
```

AntonioVentilii: Ah right! Thank you! Yes, it was a missing commit, tks for the heads-up

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -51,7 +55,7 @@ const initAuthStore = (): AuthStore => {
 		signIn: ({ domain }: AuthSignInParams) =>
 			// eslint-disable-next-line no-async-promise-executor
 			new Promise<void>(async (resolve, reject) => {
-				authClient = authClient ?? (await createAuthClient());
+				authClient = authClient ?? (await safeCreateAuthClient());
```
peterpeterparker: is `safeCreateAuthClient` actually used here? At runtime, isn't the `authClient` already initialized by `authStore.sync` and per extension this call is neither performed because `authClient` isn't undefined?

If that's accurate, I fear this leads us to an issue. Does the sign-in actually works out in Safari if we add an asynchronous call in `signIn`?

Long story short:

1. Shouldn't `safeCreateAuthClient` be called here in any case because the `authClient` is already initialized by `authStore.sync`?

and

2. If we do so, does it still works in Safari?

src/frontend/src/lib/utils/auth.utils.ts
```diff
@@ -1,20 +1,33 @@
...
 	AuthClient.create({
+		storage: idbStorage,
 		idleOptions: {
 			disableIdle: true,
 			disableDefaultIdleCallback: true
 		}
 	});
 
+/**
+ * When the user signs out, we create a new AuthClient.
```
peterpeterparker: Not sure what's this sentence - "When the user signs out, we create a new AuthClient." - means in the context?

src/frontend/src/lib/utils/auth.utils.ts
```diff
@@ -1,20 +1,33 @@
...
 import { AUTH_ALTERNATIVE_ORIGINS, AUTH_DERIVATION_ORIGIN } from '$lib/constants/app.constants';
 import { isNullishOrEmpty } from '$lib/utils/input.utils';
 import type { Identity } from '@dfinity/agent';
-import { AuthClient } from '@dfinity/auth-client';
+import { AuthClient, KEY_STORAGE_KEY } from '@dfinity/auth-client';
 import { notEmptyString } from '@dfinity/utils';
 
 export const createAuthClient = (): Promise<AuthClient> =>
 	AuthClient.create({
+		storage: idbStorage,
```
peterpeterparker: Out of curiosity, never used it, why do we need to pass the `idbStorage` here?

AntonioVentilii: this is to tell AuthClient to use our "custom" storage, and be able to manage it. In our specific case, we will be able to delete the identity key

src/frontend/src/lib/utils/auth.utils.ts
AntonioVentilii: The issue is that when we sign out, we delete the delegation from the IDB store. Doing that, triggers `<svelte:window onstorage={syncAuthStore} />` that by itself will call:

```ts
sync: async () => {
			authClient = authClient ?? (await createAuthClient());
			const isAuthenticated: boolean = await authClient.isAuthenticated();

			set({
				identity: isAuthenticated ? authClient.getIdentity() : null
			});
		},
```
		

creating a new `AuthClient` since `authClient = authClient ?? (await createAuthClient())` will solve for right side of the condition.

Shall I modify the sentence? Make it more explainable?

src/frontend/src/lib/utils/auth.utils.ts
ilbertt: Exactly, the whole point of this PR is to stop using the default `AuthClient` internal storage, in order to be able to [clean it](https://github.com/dfinity/oisy-wallet/pull/8398/files#diff-11e4b75ae021c3aaa4a126e89227c620cfb20112812680af7f3da5c000f087c5R24) right before signing in again

src/frontend/src/lib/utils/auth.utils.ts
```diff
@@ -1,20 +1,33 @@
...
 		idleOptions: {
 			disableIdle: true,
 			disableDefaultIdleCallback: true
 		}
 	});
 
+/**
+ * When the user signs out, we create a new AuthClient.
+ * Since agent-js persists identity keys in IndexedDB by default,
```
ilbertt: How about this instead? Just for branding you know ;)
```suggestion
 * Since icp-js-core persists identity keys in IndexedDB by default,
```

src/frontend/src/lib/utils/auth.utils.ts
```diff
@@ -1,20 +1,33 @@
...
+ * if the old keys remain in storage, they could be tampered with and affect the next login.
```
ilbertt:
```suggestion
 * they could be tampered with and affect the next login.
```

src/frontend/src/lib/utils/auth.utils.ts
ilbertt: Agree it's not the best explanation here. I would rather remove the whole first sentence
```suggestion
```

src/frontend/src/lib/stores/auth.store.ts
ilbertt: Why not creating a new authClient every time here?
```suggestion
				authClient = await safeCreateAuthClient();
```

Otherwise, if the authClient is already created during `sync`, we don't clean the local storage and hence we do not solve the issue.
What am I missing?

src/frontend/src/lib/utils/auth.utils.ts
AntonioVentilii: I wanted to have some documentation on this issue, so maybe I shall put it where we use `safeCreateAuthClientas` suggested by @ilbertt

src/frontend/src/lib/stores/auth.store.ts
AntonioVentilii: So, a possible solution could be that sync becomes this:


```ts
sync: async () => {
	authClient = authClient ?? (await createAuthClient());
	const isAuthenticated: boolean = await authClient.isAuthenticated();

	if (isAuthenticated) {
		set({ identity: authClient.getIdentity() });

		return;
	}

	authClient = null;

	set({ identity: null });
},
```
		
and in `signIn` we do directly `authClient = await safeCreateAuthClient()`

right? WDYT?

peterpeterparker: Before we discuss the solution, we should probably verify that this doesn’t break the auth flow in **Safari**.

Can you test whether using `authClient = await safeCreateAuthClient()` in `signIn` still works in Safari?
Respectively test that adding an asynchronous call within the sign-in flow does not lead Safari to block the II popup?

AntonioVentilii: I tried it (Safari desktop), and it seems to be working correctly (I did the same tests of the description). If you want to check, I will leave it in [fe1.oisy.com](https://fe1.oisy.com) for today

src/frontend/src/lib/utils/auth.utils.ts
AntonioVentilii: I moved a bit the commenting, have a look pls

src/frontend/src/lib/stores/auth.store.ts
peterpeterparker: Thanks for deploying the test version. Unfortunately, my assumption proved to be accurate: adding this additional `async` call within the sign-in flow causes Safari to block the II popup. Therefore, given that to my knowledge interacting and ensuring a change take effect in IndexedDB requires async, I’m afraid this whole solution (PR) cannot really move forward, WDYT?

I’d also like to apologize for not having thought of this earlier when we first discussed the potential solution. Sorry that it only came to mind later on.

Capture d’écran 2025-08-28 à 11 25 07
Capture d’écran 2025-08-28 à 11 25 19

peterpeterparker: Discussed offline. We will resolve this issue by enforcing `signIn` to reuse an existing `authClient` (separe PR) and make `sync` reset the client if not signed-in.

src/frontend/src/lib/api/idb-auth.api.ts
```diff
@@ -0,0 +1,3 @@
+import { IdbStorage } from '@dfinity/auth-client';
+
+export const idbStorage: IdbStorage = new IdbStorage();
```
peterpeterparker: - if you like it: not sure you need to explicitely declare `: IdbStorage`, it's maybe interferred?

- I wonder if there is another solution than exposing `idbStorage`? The PR adds a reference from "utils" to "api" which seems a bit incorrect if you know what I mean? Maybe in another PR we could rename `idb-auth.api` into `auth-client.api` (or services?), move `createAuthClient` in it and remove the `export const`? Not that a strong opinion, just thinking out at loud. WDYT?

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -43,8 +47,21 @@ const initAuthStore = (): AuthStore => {
 			authClient = authClient ?? (await createAuthClient());
 			const isAuthenticated: boolean = await authClient.isAuthenticated();
 
+			if (!isAuthenticated) {
+				// When the user signs out, we modify the storage, triggering a call to `sync()` method through `<svelte:window onstorage={syncAuthStore} />` in the page.
```
peterpeterparker: Is that correct? I feel like those are two different things. Using signing-out is one, `onstorage` is another, a guard on manual changes on the storage but, I might be wrong.

src/frontend/src/lib/utils/auth.utils.ts
peterpeterparker: Yeah but, do we need to? I mean not against it, but we can as well just use the object in the cleanup without having to old a global value and passing it to the authclient no, that would be possible as well?

```
export const safeCreateAuthClient = async (): Promise<AuthClient> => {
    const storage = new IdbStorage();
	await idbStorage.remove(KEY_STORAGE_KEY);
	
	return await createAuthClient();
};
```

If we do not pass any options to `new IdbStorage();` we are just using the default option - i.e. default name, storeName and version.

BUT I might be missing something, and not against at all to the pattern proposed by this PR, just asking because we are adding a global variable, it's more for my understanding.

src/frontend/src/lib/stores/auth.store.ts
ilbertt: @peterpeterparker @AntonioVentilii I would anyway change the line 71 of `src/frontend/src/lib/stores/auth.store.ts` to:
```diff
-				authClient = authClient ?? (await createAuthClient());
+               // TODO: throw an error instead of re-creating the AuthClient 
+				authClient = authClient ?? (await safeCreateAuthClient());
```

src/frontend/src/tests/lib/components/guard/AddressGuard.spec.ts
```diff
@@ -71,22 +71,19 @@ describe('AddressGuard', () => {
 			expect(spy).toHaveBeenCalledOnce();
 		});
 
-		it('should sign out and ultimately reload the window if signer allowance fails', async () => {
```
peterpeterparker: Is that related to this PR? Did we remove a reload? I'm confused, sorry if it's a dumb question.

Reloading after sign-out is really important.

src/frontend/src/lib/stores/auth.store.ts
AntonioVentilii: Whatever you prefer. In this PR, I will remove it anyway:

- https://github.com/dfinity/oisy-wallet/pull/8544

@peterpeterparker your call here, no strong preference on my side

src/frontend/src/lib/utils/auth.utils.ts
ilbertt: The proposed `safeCreateAuthClient` function may work as well, but it's more opaque, as it doesn't really show that the storage used by the AuthClient is the one that has just been cleaned.

Constructing the storage directly inside the `safeCreateAuthClient` function and passing it to the `AuthClient` can work though

src/frontend/src/tests/lib/components/guard/AddressGuard.spec.ts
AntonioVentilii: Ye, indeed, but we don't test it here anymore: the problem is that i am required to entirely mock the signout function (because of the new IDB addition), so, we cannot call the spy on the reload, since theoretically the spy on the signout do not call that.

I can create tests for signOut, that's for sure.

I tried to mock the IDB functions, but still not enough. I will give it another try and let you know

src/frontend/src/lib/stores/auth.store.ts
AntonioVentilii: When we signOut, we delete something in the storage, and that may trigger the `onstorage` event. I just wanted to explain the relations between signOut and sync. That basically happens on the page reload too when signing out. Maybe I just write something like "signOut triggers sync, etc", WDYT?

src/frontend/src/lib/api/idb-auth.api.ts
AntonioVentilii: - done!

- I will rename it already in this PR, and in the next one do the moving as you suggested (both `createAuthClient` and `safeCreateAuthClient` that use it)

AntonioVentilii: Ah wait! it will be used in `checkDelegationChain` too in `auth.worker` so, it still needs to be exported.

Unless I revert that specific change:

### Before this PR

```ts
// auth.worker.ts
...

const idbStorage: IdbStorage = new IdbStorage();
const delegationChain: string | null = await idbStorage.get(KEY_STORAGE_DELEGATION);
```

### After this PR

```ts
// auth.worker.ts
import { idbStorage } from '$lib/api/auth-client.api';

...

const delegationChain: string | null = await idbStorage.get(KEY_STORAGE_DELEGATION);
```


I did the change because it seemed more correct to reference the specific IDB storage that we use (similar to @ilbertt 's comment [here](https://github.com/dfinity/oisy-wallet/pull/8398#discussion_r2308363134))

src/frontend/src/lib/stores/auth.store.ts
peterpeterparker: The listen on the storage is probably an artifact of the past, when the delegation was saved in the local storage, because it is triggered when the local storage or session storage is modified, so, I actually wonder if we should not omit it in this comment and in another PR even remove the listener?

src/frontend/src/lib/utils/auth.utils.ts
peterpeterparker: I see your point, I'm not necesseraly convinced that it's less opaque - it's just a reference on some undocumented global variable which has a generic name and which is the same storage as the one used by default by the authclient - and I'm not sure it really expresses the fact that it has been cleaned, but works for me.

src/frontend/src/lib/api/auth-client.api.ts
```diff
@@ -0,0 +1,3 @@
+import { IdbStorage } from '@dfinity/auth-client';
+
+export const idbStorage = new IdbStorage();
```
peterpeterparker: - Should we rename it to express a bit better it's purpose, like `authClientStorage` or something of that sort?

- Should we add a short comment on why we delcare a storage? Even just `@see safeCreateAuthClient`?

AntonioVentilii: yes on both!

src/frontend/src/tests/lib/components/guard/AddressGuard.spec.ts
AntonioVentilii: Ok, I think now it's better, have a look please

src/frontend/src/lib/stores/auth.store.ts
AntonioVentilii: WDYT of this?

```ts
// When the user signs out, we trigger a call to `sync()`.
// The `sync()` method creates a new `AuthClient` (since the previous one was nullified on sign out), causing the creation of new identity keys in IndexedDB.
// To avoid using such keys (or tampered ones) for the next login, we use method `safeCreateAuthClient()` which clears any stored keys before creating a new `AuthClient`.
// We do it only if the user is not authenticated, because if it is, then it is theoretically already safe (or at least, it is out of our control to make it safer).
```

src/frontend/src/lib/api/idb-auth.api.ts
peterpeterparker: ok

src/frontend/src/lib/utils/auth.utils.ts
peterpeterparker: In addition, I just noticed that the AuthClient prints a console warning "You are using a custom storage provider..." whenever the storage parameter is used. Aside from my aversion to printing anything in the console of a production application, I also find this unfriendly to users: any curious developer will notice that when using OISY they get a warning upon signing in, which, in my opinion, does not convey a strong sense of security.

Just leaving this comment for posterity.

peterpeterparker (approved):
LGTM, thx.

Note for posterity: I'm not convinced about the global variable and I'm displeased with the enforced console warning printed by the AuthClient, but I suppose I can live with it and we can iterate if necessary.

src/frontend/src/lib/stores/auth.store.ts
peterpeterparker: Sounds good!

src/frontend/src/lib/api/idb-auth.api.ts
AntonioVentilii: I moved both `createAuthClient` and `safeCreateAuthClient` in PR:

- https://github.com/dfinity/oisy-wallet/pull/8398#discussion_r2309713548

PR #8399, opened 2025-08-22
feat(frontend): added snippet for SwapFees in ReviewModal

roman-nazaruk:
# Motivation

We would need to render different fees for icp and evm networks, so we need to pass this component as Snippet.

# Changes

Added ability to render SwapFees via snippet.


AntonioVentilii (commented):
it seems this PR has two scopes: migrating to Svelte 5 (we need to remove the `dispatch`, too in that case) and adding a message box.

Let's split it

PR #8401, opened 2025-08-22
fix(frontend): Signer text overflow

daviddecentage:
# Motivation

Long text in the signer consent message screen can overflow the content border.

# Changes

Added break-all class to make long words break to new lines

# Tests

Before:
image

After:
image


roman-nazaruk (approved):
LGTM, thnx

PR #8419, opened 2025-08-25
feat(frontend): update SendTokenReview component

DenysKarmazynDFINITY:
# Motivation

We need to extend `SendTokenReview` with some additional props to make it re-usable for the AI assistant console.


roman-nazaruk (approved):
LGTM, thnx

PR #8421, opened 2025-08-25
feat(frontend): added snippet for SwapFees in ReviewModal

roman-nazaruk:
# Motivation

We would need to render different fees for icp and evm networks, so we need to pass this component as Snippet.

# Changes

Added ability to render SwapFees via snippet.

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapReview.svelte
+++ b/src/frontend/src/lib/components/swap/SwapReview.svelte
@@ -2,7 +2,6 @@
 	import { Html } from '@dfinity/gix-components';
 	import { isEmptyString, nonNullish } from '@dfinity/utils';
 	import { getContext, type Snippet } from 'svelte';
-	import SwapFees from '$lib/components/swap/SwapFees.svelte';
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/swap/SwapReview.svelte
```diff
@@ -98,9 +100,20 @@
 
 	<div class="flex flex-col gap-3">
 		<SwapProvider {slippageValue} />
-		<SwapFees />
+		{@render swapFees()}
 	</div>
 
+	{#if nonNullish($destinationToken) && nonNullish($sourceToken) && $sourceToken.network.id !== $destinationToken.network.id}
```
DenysKarmazynDFINITY: it seems this part is not related to this PR, or am I missing something?

roman-nazaruk: Yeah, it is not so related to this PR

roman-nazaruk: Removed

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8427, opened 2025-08-25
refactor(frontend): filters displayed collections

BonomoAlessandro:
# Motivation

We want that the displayed collections can be filtered by the `network` filter.


daviddecentage (approved):
LGTM

PR #8439, opened 2025-08-25
feat(frontend): Markdown Sidebar

daviddecentage:
# Motivation

For the new Terms of Use, License agreement pages, we add a sidebar component that renders a table of contents for a list of MarkdownBlocks.

# Changes

Added component

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/components/ui/MarkdownSidebar.svelte
+++ b/src/frontend/src/lib/components/ui/MarkdownSidebar.svelte
@@ -12,10 +12,12 @@
 	const { headings }: Props = $props();
 
 	let activeId: string | undefined = $state();
+	let observer: IntersectionObserver | undefined;
+	const observed = new Set<Element>();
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/ui/MarkdownSidebar.svelte
```diff
@@ -0,0 +1,51 @@
...
+
+	interface Props {
+		headings: MarkdownBlockType[];
+	}
+
+	const { headings }: Props = $props();
+
+	let activeId: string | undefined = $state();
+
+	onMount(() => {
```
roman-nazaruk, line 18: Should you clean Observer when the component unmounts?

daviddecentage: good point, updated it

roman-nazaruk (approved):
LGTM

PR #8442, opened 2025-08-25
feat(frontend): added support supportedNetworks for ModalNetworksFilter component

roman-nazaruk:
# Motivation

We need to show only networks that support swaps. So this PR adds an option to pass supported network IDs and filters the list accordingly (including cross-network swap scenarios).

# Changes

Filter networks against supportedNetworkIds

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapAmountsContext.svelte
+++ b/src/frontend/src/lib/components/swap/SwapAmountsContext.svelte
@@ -1,8 +1,6 @@
 <script lang="ts">
 	import { debounce, isNullish, nonNullish } from '@dfinity/utils';
 	import { getContext, type Snippet } from 'svelte';
-	import type { Erc20Token } from '$eth/types/erc20';
-	import type { IcToken } from '$icp/types/ic-token';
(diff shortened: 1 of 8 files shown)


src/frontend/src/lib/components/tokens/ModalNetworksFilter.svelte
```diff
@@ -30,10 +44,12 @@
 
 <ContentWithToolbar>
 	<NetworkSwitcherList
+		{allNetworksEnabled}
 		delayOnNetworkSelect={false}
 		labelsSize="lg"
 		onSelected={onNetworkSelect}
 		selectedNetworkId={$filterNetwork?.id}
+		supportedNetworks={$filteredNetworks}
```
BonomoAlessandro, line 52: Wouldn't it be better to remove the `allNetworksEnabled` and only use the `supportedNetworks`? Like this you can provide all networks or just a filtered subset.

roman-nazaruk: These are two separate props: one controls whether all networks are visible, and the other filters them. Additionally, setting allNetworksEnabled shows the “All networks” label.

BonomoAlessandro (approved):
LGTM

PR #8447, opened 2025-08-26
refactor(frontend): refactors LoaderCollections reduce functions

BonomoAlessandro:
# Motivation

To be more consistent, we want to handle the return value of the `reduce` functions the same way as anywhere else in the code.


daviddecentage (approved):
LGTM

PR #8453, opened 2025-08-26
feat(frontend): prepare getBtcWalletBalance() integration

DecentAgeCoder:
# Motivation

This Pull Request  introduces user-friendly translation for addition errors, e.g  when pending transactions are unavailable. Furthermore it ensures that externally fetched transactions are recalculated until the required confirmations are met, keeping the number of confirmations in the UI up to date.

# Changes

  - Updated logic for filtering and including transactions based on confirmation status and the new constant `CONFIRMED_BTC_TRANSACTION_MIN_CONFIRMATIONS`.
  - Added new localized error messages for various Bitcoin transaction-related issues in multiple languages (ar, cs, de, en, fr, hi, it, ja, pl, pt, ru, vi, zh-CN).
  - Updated `i18n.d.ts` to include the new key for `pending_transactions_not_available`.
  - Adjusted expectations for the number of postMessage calls in corresponding test assertions.

# Tests

- Updated test in `btc-listener.services.spec.ts` so it matches the expected post message events with the for the changed filter condition in `loadBtcTransactionsData(..)`

### DIFF
--- a/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
+++ b/src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
@@ -95,7 +95,7 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc> 
 				}
 
 				// If block_height is null (unconfirmed/pending), include the transaction
-				if (transaction.block_height === null) {
+				if (isNullish(transaction.block_height)) {
 					return true;
 				}
 
--- a/src/frontend/src/tests/btc/workers/btc-wallet.worker.spec.ts
+++ b/src/frontend/src/tests/btc/workers/btc-wallet.worker.spec.ts
@@ -167,10 +167,10 @@ describe('btc-wallet.worker', () => {
 
 					await vi.advanceTimersByTimeAsync(WALLET_TIMER_INTERVAL_MILLIS);
 
-					expect(postMessageMock).toHaveBeenCalledTimes(10); // Change from 8 to 10
-					expect(postMessageMock).toHaveBeenNthCalledWith(8, mockPostMessageStatusInProgress); // Change from 7 to 8
-					expect(postMessageMock).toHaveBeenNthCalledWith(9, mockPostMessageUncertified); // Add this line
-					expect(postMessageMock).toHaveBeenNthCalledWith(10, mockPostMessageStatusIdle); // Change from 8 to 10
+					expect(postMessageMock).toHaveBeenCalledTimes(10);
+					expect(postMessageMock).toHaveBeenNthCalledWith(8, mockPostMessageStatusInProgress);
+					expect(postMessageMock).toHaveBeenNthCalledWith(9, mockPostMessageUncertified);
+					expect(postMessageMock).toHaveBeenNthCalledWith(10, mockPostMessageStatusIdle);
 				});
 
 				it('should start the scheduler with an interval', async () => {


src/frontend/src/tests/btc/workers/btc-wallet.worker.spec.ts
```diff
@@ -160,15 +160,17 @@ describe('btc-wallet.worker', () => {
...
-					expect(postMessageMock).toHaveBeenNthCalledWith(6, mockPostMessageStatusIdle);
+					expect(postMessageMock).toHaveBeenNthCalledWith(6, mockPostMessageUncertified);
+					expect(postMessageMock).toHaveBeenNthCalledWith(7, mockPostMessageStatusIdle);
 
 					await vi.advanceTimersByTimeAsync(WALLET_TIMER_INTERVAL_MILLIS);
 
-					expect(postMessageMock).toHaveBeenCalledTimes(8);
-					expect(postMessageMock).toHaveBeenNthCalledWith(7, mockPostMessageStatusInProgress);
-					expect(postMessageMock).toHaveBeenNthCalledWith(8, mockPostMessageStatusIdle);
+					expect(postMessageMock).toHaveBeenCalledTimes(10); // Change from 8 to 10
```
DenysKarmazynDFINITY: can you please remove these comments?

src/frontend/src/lib/i18n/fr.json
```diff
@@ -728,7 +728,8 @@
 			"btc_insufficient_balance_for_fee": "Solde Bitcoin insuffisant pour couvrir les frais de transaction",
 			"btc_invalid_utxo_data": "Données de transaction Bitcoin invalides",
 			"btc_utxo_locked": "Les fonds Bitcoin sont verrouillés par des transactions en attente",
-			"btc_invalid_fee_calculation": "Échec de la validation des frais de transaction Bitcoin"
+			"btc_invalid_fee_calculation": "Échec de la validation des frais de transaction Bitcoin",
+			"pending_transactions_not_available": "Le service de transactions en attente n'est actuellement pas disponible. Veuillez réessayer plus tard."
```
DenysKarmazynDFINITY, line 732: Did you translate this key to all languages by yourself?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -81,12 +84,28 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
+			// Keep transactions that are either unconfirmed (no block_height) or have fewer than the required number of confirmations.
+			// This ensures proper recalculation of confirmation status until final confirmation is reached.
+			const newTransactions = fetchedTransactions.filter((transaction) => {
+				// Include transactions that are NOT already in the store
+				if (isNullish(this.store.transactions[`${transaction.hash}`])) {
+					return true;
+				}
+
+				// If block_height is null (unconfirmed/pending), include the transaction
+				if (transaction.block_height === null) {
```
DenysKarmazynDFINITY: shouldn't we compare block_height with the transaction from the store and update only if it's different? not sure I understand logic behind the `transaction.block_height === null` check

src/frontend/src/tests/btc/workers/btc-wallet.worker.spec.ts
DecentAgeCoder: Done

src/frontend/src/lib/i18n/fr.json
DecentAgeCoder: No, I have translated them with chatgpt :wink:

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DecentAgeCoder: No, this will not work. The confirmation is calculated within mapBtcTransaction, and the block_height only changes from undefined to 1 with the first confirmation. As a result, the confirmation will not be recalculated once the transaction is in the store (you would need to reload the page). With the current condition, transactions are treated as newTransaction until they reach 6 confirmations and will thereby have an updated confirmation value..

So, transaction.block_height === null only occurs while the transaction is still pending.
I am therefore changing the condition to isNullish(transaction.block_height)

PR #8455, opened 2025-08-26
feat(feat): enable parallel UTXO locked transactions for BTC send

DecentAgeCoder:
# Motivation

# Changes

# Tests


src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
```diff
@@ -50,7 +46,8 @@
 
 	const progress = (step: ProgressStepsSendBtc) => (sendProgressStep = step);
 
-	let utxosFee: UtxosFee | undefined = undefined;
+	let utxosFee: UtxosFee | undefined;
+	$: utxosFee;
```
DenysKarmazynDFINITY: not sure what this line means, is this some leftover?

src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
```diff
@@ -136,31 +133,20 @@
...
 		try {
+			progress(ProgressStepsSendBtc.SEND);
 			await sendBtc({
 				destination,
 				amount,
 				utxosFee,
 				network,
 				source,
-				identity: $authIdentity,
-				onProgress: () => {
```
DenysKarmazynDFINITY, line 157: any reason why we should call progress in the component and not in the service (the way how it was)?

src/frontend/src/btc/components/send/BtcSendWarnings.svelte
```diff
@@ -37,6 +38,68 @@
 			>
 		</MessageBox>
 	</div>
+{:else if utxosFee?.error === BtcPrepareSendError.UtxoLocked}
+	<div class="w-full" in:fade>
```
DenysKarmazynDFINITY, line 42: the wrapper div is the same for all the cases - please move it outside of the `if`

DenysKarmazynDFINITY: most likely, we'll loose the animation effect if we do so - so let's keep it as it is

src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
DecentAgeCoder: I looked into this issue with David. He recommended to add the reactive variable this way. Are you sure I can remove it?

DenysKarmazynDFINITY: The way it is defined now, it’s a no-op reactive statement, so it can be definitely removed.

DenysKarmazynDFINITY: Or was intended just to declare it as reactive? I haven't seen such a pattern, nor do we have it in the code. Would have been much easier if this component would be in Svelte 5.

src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
DecentAgeCoder: Previously, the status was not displayed correctly, which is why I had to change it. Since I explicitly set the progress using progress(..), I don’t see the need for a manual onProgress definition. In my understanding, it only advances to the next step (which is probably desired if there is a state change within sendBtc(..) that should be displayed). But it might also be due to my lack of understanding of how this component is intended to be used.

src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
DecentAgeCoder: Now I got it. Thanks

src/frontend/src/btc/components/send/BtcSendTokenWizard.svelte
DecentAgeCoder: The INITIALIZATION phase is now in sync with the validation.

DenysKarmazynDFINITY (approved):

PR #8457, opened 2025-08-26
feat(frontend): update contacts tool

DenysKarmazynDFINITY:
# Motivation

We need to upgrade the show_contacts tool, so the LLM can return a small message if no contacts were found.


roman-nazaruk (approved):
LGTM, thnx

PR #8458, opened 2025-08-26
feat(frontend): increase chat history length

DenysKarmazynDFINITY:
# Motivation

We can now increase the chat history length to the max allowed by LLM canister number.


roman-nazaruk (approved):
LGTM, thnx

PR #8462, opened 2025-08-26
refactor(frontend): Use ERC utils to check the standard

AntonioVentilii:
# Motivation

For better manageability, we use the existing utils to check the standard of an ERC token.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8470, opened 2025-08-26
style(frontend): Capitalize first letter of currency name

AntonioVentilii:
# Motivation

Some languages have the name of the currencies lowercase by default. So, to make it more proper, we make the first letter uppercase in the dropdown list of currencies.

### Before

Screenshot 2025-08-26 at 11 20 20

### After

Screenshot 2025-08-26 at 11 19 44

### DIFF
--- a/src/frontend/src/lib/components/currency/CurrencyDropdown.svelte
+++ b/src/frontend/src/lib/components/currency/CurrencyDropdown.svelte
@@ -18,6 +18,7 @@
 	import { currencyStore } from '$lib/stores/currency.store';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { getCurrencyName, getCurrencySymbol } from '$lib/utils/currency.utils';
+	import { capitalizeFirstLetter } from '../../../tests/utils/string-utils';
 
 	let dropdown = $state<Dropdown>();
 
@@ -90,7 +91,7 @@
 								{/if}
 							</span>
 							<div class="flex w-full flex-row justify-between gap-5">
-								{name}
+						{capitalizeFirstLetter(name)}
 								<span class="text-right text-tertiary">{symbol}</span>
 							</div>
 						</Button>


src/frontend/src/lib/components/currency/CurrencyDropdown.svelte
```diff
@@ -90,7 +91,7 @@
 								{/if}
 							</span>
 							<div class="flex w-full flex-row justify-between gap-5">
-								{name}
+								{capitalizeFirstLetter(name)}
```
DenysKarmazynDFINITY: how about we use CSS instead: `<span class='capitalize '>{name}</span>`

src/frontend/src/lib/components/currency/CurrencyDropdown.svelte
```diff
@@ -1,5 +1,6 @@
 <script lang="ts">
 	import { nonNullish } from '@dfinity/utils';
+	import { capitalizeFirstLetter } from '../../../tests/utils/string-utils';
```
DenysKarmazynDFINITY: We should be able to do it with CSS, but if for some reasons not - we'll need to move this util from `tests` to `lib`

src/frontend/src/lib/components/currency/CurrencyDropdown.svelte
AntonioVentilii: Good catch! Yes, done! (i used `first-letter:` modifier because otherwise it would uppercase all first letters)

src/frontend/src/lib/components/currency/CurrencyDropdown.svelte
AntonioVentilii: I was going to do it in another PR:

- https://github.com/dfinity/oisy-wallet/pull/8471

But we don't need it anymore

src/frontend/src/lib/components/currency/CurrencyDropdown.svelte
```diff
@@ -18,6 +18,7 @@
 	import { currencyStore } from '$lib/stores/currency.store';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { getCurrencyName, getCurrencySymbol } from '$lib/utils/currency.utils';
+	import { capitalizeFirstLetter } from '../../../tests/utils/string-utils';
```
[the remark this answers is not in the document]
AntonioVentilii: we don't need to do that anymore

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8471, opened 2025-08-26
refactor(frontend): Move util `capitalizeFirstLetter` in the `lib` module

AntonioVentilii:
# Motivation

Since we are using the util `capitalizeFirstLetter` outside the tests, it make sense to move it.

In a follow-up we will create some tests.


AntonioVentilii:
We don't need this PR anymore, since we are not using it in `lib` anymore.

PR #8486, opened 2025-08-26
feat(frontend): integrate getBtcWalletBalance()

DecentAgeCoder:
# Motivation

This Pull Request integrates `getBtcWalletBalance() `into `btc-listener.services.ts`. It retrieves btcWalletBalance.total and stores the value in the BTC balance store. Additionally, it improves error handling and introduces a user-friendly message for cases where pending transactions are unavailable.

# Changes

- Integrated structured balance calculation `getBtcWalletBalance()` in `btc-listener.services.ts`
- If the send button was not properly disabled during an error, and `validateBtcSend()` is called we abort and return the same error
- Added new error codes (`PendingTransactionsNotAvailable`, `UtxoLocked`) in `btc-send.ts` to facilitate better error reporting.
- Added translations for new error messages in multiple language files to ensure localization support.

# Tests
Updated and added several tests in `btc-listener.services.spec.ts` and btc-utxos.service.spec.ts to validate:
- Proper transaction and balance storage logic.
- Graceful handling of missing pending transactions
- Mocked dependencies to simulate actual service behavior.
- Adjustments in tests for mocked services like `btc-pending-sent-transactions.ts` and `btc-address.utils`


src/frontend/src/btc/services/btc-listener.services.ts
```diff
@@ -24,22 +30,67 @@ export const syncWallet = ({
...
+
+	// Only store transactions when we have actual transaction data (certified === false)
+	if (nonNullish(providerTransactions)) {
+		btcTransactionsStore.prepend({
+			tokenId,
+			transactions: providerTransactions
+		});
+	}
+	// Skip the certified event since it does not contain the required provider transactions
+	if (certified) {
```
DenysKarmazynDFINITY: we are rejecting certified balances with this approach, any reason for that?

src/frontend/src/btc/services/btc-listener.services.ts
```diff
@@ -24,22 +30,67 @@ export const syncWallet = ({
...
+		return;
+	}
 	if (nonNullish(balance)) {
+		/*
+		 * Balance calculation is performed here in the main thread rather than in the worker (btc-wallet.scheduler.ts)
+		 * because the pending transactions store (btcPendingSentTransactionsStore) is not accessible from the worker context.
+		 * The worker provides the confirmed balance from the Bitcoin canister, and we calculate the structured
+		 * balance (confirmed, unconfirmed, total) using newTransactions data to determine confirmation states.
+		 */
+		const identity = get(authIdentity);
```
DenysKarmazynDFINITY: `syncWallet` is just about parsing and assigning data from workers to the stores, having all the additional balance calculations here as well as accessing all the stores does not feel right - can you please think about moving it to another place?

src/frontend/src/btc/services/btc-listener.services.ts
```diff
@@ -24,22 +30,67 @@ export const syncWallet = ({
...
+
+		// Get networkId from tokenId (same way as SendContext derives it from token)
+		const networkId = mapTokenIdToNetworkId(tokenId);
+
+		// Get the source address using the same logic as other components (BtcConvertTokenWizard, etc.)
+		const sourceAddress = getBtcSourceAddress(networkId);
+
+		// Wait for pending transactions to be loaded before calculating balance
+		await loadBtcPendingSentTransactions({
```
DenysKarmazynDFINITY: is it really necessary to load it on every worker tick? this data only gets updated when user initiates a new tx

src/frontend/src/btc/types/btc-send.ts
```diff
@@ -20,6 +22,7 @@ export enum BtcSendValidationError {
 	InsufficientBalanceForFee = 'InsufficientBalanceForFee',
 	InvalidUtxoData = 'InvalidUtxoData',
 	UtxoLocked = 'UtxoLocked',
+	PendingTransactionsNotAvailable = 'pending_transactions_not_available',
```
DenysKarmazynDFINITY: please use the same format as all other values

src/frontend/src/btc/services/btc-listener.services.ts
DecentAgeCoder: I will introduce a new function calculateBtcWalletBalance(..) in wallet-btc.service.ts and invoke it from btc-listener.service.ts. Does that approach make sense?

src/frontend/src/btc/services/btc-listener.services.ts
DecentAgeCoder: - Updated transactions are only available when certified != null (since the certified request arrives after the uncertified one, which would otherwise overwrite the previously calculated and stored BtcWalletBalance).

- When we recalculate the BTC balance, btc_get_pending_transactions is called in very short succession in both cases. 

What would you suggest?

src/frontend/src/btc/services/btc-listener.services.ts
DecentAgeCoder: No, this would be triggered on every tick as long as the pending transactions have fewer than 6 confirmations (due to the updated filter logic in the BTC scheduler). To avoid unnecessary postMessage events for changed pending transactions, we would need to calculate the confirmations on the btc-listener.service.ts side.

I’m also not sure about the delay between the detection of a new BTC block and the time at which those block changes are reflected in the fetched transactions from the external provider, which is required to have a consistend and up-to-date data for the BTC wallet.

src/frontend/src/btc/services/btc-listener.services.ts
DenysKarmazynDFINITY: The way it looks now, this function does not have any effect when certified flag is true. Which means the "update" call to get a certified balance is just being ignored. I get the idea of generating the very "up-to-date" BTC balance, but I'm not sure if it's a good idea to always rely on manually calculated balance.

If such an approach was discussed and agreed with @StefanBerger-DFINITY and the security team, we'll need to clean up the "certified" part of the BTC worker to avoid unnecessary requests and code.

src/frontend/src/btc/services/btc-listener.services.ts
DenysKarmazynDFINITY: yes, we need to extract this logic

src/frontend/src/btc/services/btc-listener.services.ts
DecentAgeCoder: I have updated the code as follows:

1. The `confirmedBalance` is now calculated as:
   balance - lockedBalance
  `balance`: the certified balance retrieved by the BTC worker
   `lockedBalance`: calculated using the fetched pending transactions

   This still aligns with the documented definition in `BtcWalletBalance`:
   *“Represents UTXOs that are confirmed on the blockchain and considered safe. This is the baseline amount from which other balances are calculated.”*

2. The `confirmedBalance` from `btcWalletBalance` is now stored in the `balances` store (instead of the total).

For now, this solution is sufficient. Once we refactor the `balancesStore` to a similar data structure, we will gain more flexibility in choosing which property to use for a given use case.

DenysKarmazynDFINITY (approved):

PR #8488, opened 2025-08-26
refactor(frontend): License agreement

daviddecentage:
# Motivation

We refactor the layout and way of rendering the Licence Agreement. For now, we only add english and keep the old version if the feature flag is not enabled.

# Changes

- Changed styles in the layout
- Conditionally render the original or new version with the table of contents

# Tests

image


BonomoAlessandro (approved):
LGTM

PR #8491, opened 2025-08-26
feat(frontend): AiAssistantReviewSendBtcToken component

DenysKarmazynDFINITY:
# Motivation

This PR includes a new component for displaying the BTC "review_send_token" tool.

Screenshot 2025-08-26 at 16 44 11


src/frontend/src/lib/components/ai-assistant/AiAssistantReviewSendBtcToken.svelte
```diff
@@ -0,0 +1,249 @@
...
+	let loading = $state(false);
+
+	let parsedAmount = $derived(
+		parseToken({
+			value: `${amount}`,
+			unitName: $sendTokenDecimals
+		})
+	);
+
+	let amountErrorMessage = $derived.by(() => {
```
AntonioVentilii, line 70: BTW, not for this PR, if this logic is used in the BTC send flow too, we can always extract it to reduce code (passing the reactive variable)

I imagine it will be the case for the other networks too

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/components/ai-assistant/AiAssistantReviewSendBtcToken.svelte
DenysKarmazynDFINITY: Yes, we should consider re-using this part. We'll need to align the validation formats (in the send flow, it returns errors, here - strings), but it should be doable. Noted 👍

PR #8492, opened 2025-08-26
fix(devops): Correct output name for steps in backend tests

AntonioVentilii:
# Motivation

In the backend tests CIs we were using the wrong output for the job `breaking-interface`


BonomoAlessandro (approved):
LGTM

PR #8496, opened 2025-08-27
feat(frontend): maps new section field from custom token

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/eth/services/erc1155.services.ts
+++ b/src/frontend/src/eth/services/erc1155.services.ts
@@ -13,8 +13,9 @@ import type { LoadCustomTokenParams } from '$lib/types/custom-token';
 import type { OptionIdentity } from '$lib/types/identity';
 import type { NetworkId } from '$lib/types/network';
 import { parseCustomTokenId } from '$lib/utils/custom-token.utils';
-import { assertNonNullish, fromNullable, queryAndUpdate } from '@dfinity/utils';
+import { assertNonNullish, fromNullable, nonNullish, queryAndUpdate } from '@dfinity/utils';
 import { get } from 'svelte/store';
+import { CustomTokenSection } from '$lib/enums/custom-token-section';
 
 export const isInterfaceErc1155 = async ({
 	networkId,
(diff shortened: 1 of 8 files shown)


src/frontend/src/eth/services/erc1155.services.ts
```diff
@@ -105,7 +107,10 @@ const loadCustomTokensWithMetadata = async (
 						standard: 'erc1155' as const,
 						category: 'custom' as const,
 						enabled,
-						version
+						version,
+						...(nonNullish(section) && {
+							section: 'Spam' in section ? CustomTokenSection.SPAM : CustomTokenSection.HIDDEN
```
AntonioVentilii: maybe just to be assertive?

```suggestion
							section: section === 'Spam' ? CustomTokenSection.SPAM : CustomTokenSection.HIDDEN
```

src/frontend/src/eth/services/erc721.services.ts
```diff
@@ -106,7 +108,10 @@ const loadCustomTokensWithMetadata = async (
 						standard: 'erc721' as const,
 						category: 'custom' as const,
 						enabled,
-						version
+						version,
+						...(nonNullish(section) && {
+							section: 'Spam' in section ? CustomTokenSection.SPAM : CustomTokenSection.HIDDEN
```
AntonioVentilii: same (P.S. you could create a small util `mapTokenSection`, since it is already used twice and it is an hard-coded value)

```suggestion
							section: section === 'Spam' ? CustomTokenSection.SPAM : CustomTokenSection.HIDDEN
```

src/frontend/src/eth/services/erc1155.services.ts
BonomoAlessandro: This does not work. `section` is a value that looks like the following `{ Spam: null }`

src/frontend/src/lib/utils/custom-token-section.utils.ts
```diff
@@ -0,0 +1,13 @@
+import type { TokenSection } from '$declarations/backend/backend.did';
+import { CustomTokenSection } from '$lib/enums/custom-token-section';
+
+export const mapTokenSection = (section: TokenSection): CustomTokenSection | undefined => {
```
AntonioVentilii, line 4: nitpick: you can even pass `TokenSection | undefined` and remove the check in the components

BonomoAlessandro: I think it is more clear to only allow real objects when calling a mapping function.

PR #8501, opened 2025-08-27
test(frontend): Module for balances derived for tokens list

AntonioVentilii:
# Motivation

Creating some tests for the balances derived stores. Note that for test purposes we include a `reinitialize` function in the `initCertifiedStore`.


roman-nazaruk (approved):
LGTM, thnx

PR #8504, opened 2025-08-27
feat(frontend): Terms and privacy links

daviddecentage:
# Motivation

Same as the LicenseLink we add link components for the new Terms of Use and Privacy Policy.

# Changes

Added components
Added texts


src/frontend/src/lib/components/privacy-policy/PrivacyPolicyLink.svelte
```diff
@@ -0,0 +1,20 @@
...
+	interface Props {
+		noUnderline?: boolean;
+		testId?: string;
+	}
+
+	let { noUnderline = false, testId }: Props = $props();
+</script>
+
+<a
+	class:no-underline={noUnderline}
```
AntonioVentilii, line 14: these components seems very similar: is it worth creating a generic one for them?

daviddecentage: Will check it and create another PR if it makes sense

PR #8505, opened 2025-08-27
feat(frontend): Terms of use

daviddecentage:
# Motivation

We add the new Terms of Use

# Changes

- Added conditional redirect in case the feature is not enabled
- Added english draft version
- Render it in the respective component


BonomoAlessandro (approved):
LGTM

PR #8506, opened 2025-08-27
chore(frontend): Env agreements

daviddecentage:
# Motivation

Adds the json with the update dates for the diffrent agreements.

# Changes

- Added zod schema
- Added TS type
- Added json with initial dates set to today


src/frontend/src/env/types/env-agreements.ts
```diff
@@ -0,0 +1,4 @@
+import type { EnvAgreementsSchema } from '$env/schema/env-agreements.schema';
+import type * as z from 'zod/v4/index';
```
AntonioVentilii:
```suggestion
import type * as z from 'zod/v4';
```

AntonioVentilii (approved):
LGTM tks!

PR #8508, opened 2025-08-27
feat(frontend): vvuthuuljlinddtlhtnlgikggbrvbuuuulddfgkjldnk

daviddecentage:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/env/agreements.env.ts
+++ b/src/frontend/src/env/agreements.env.ts
@@ -1,5 +1,26 @@
+import agreementsJson from '$env/agreements.json';
+import { EnvAgreementsSchema } from '$env/schema/env-agreements.schema';
+import type { EnvAgreement } from '$env/types/env-agreements';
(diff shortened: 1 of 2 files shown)


src/frontend/src/env/agreements.env.ts
```diff
@@ -1,5 +1,25 @@
...
+import { get } from 'svelte/store';
+import * as z from 'zod/v4';
 
 export const NEW_AGREEMENTS_ENABLED = parseBoolEnvVar(
 	import.meta.env.VITE_FRONTEND_NEW_AGREEMENTS_ENABLED
 );
+
+const parseAgreementsJson = () => z.parse(EnvAgreementsSchema, agreementsJson);
+
+export const getAgreementLastUpdated = (type: keyof EnvAgreement) =>
```
AntonioVentilii: this can be moved to a `agreements.utils` (or `agreements.services`) module if you want; but here is good too

daviddecentage: done

PR #8510, opened 2025-08-27
feat(frontend): added check for allowance before swap

roman-nazaruk:
# Motivation

We’ve encountered errors indicating insufficient allowance before the swap, so we now verify the allowance prior to executing the Velora swap.

# Changes

Added a pre-swap allowance check before creating txParams.

# Tests

Fixed tests.

### DIFF
--- a/src/backend/backend.did
+++ b/src/backend/backend.did
@@ -188,7 +188,11 @@ type DefiniteCanisterSettingsArgs = record {
 	compute_allocation : nat
 };
 type DeleteContactResult = variant { Ok : nat64; Err : ContactError };
-type ErcToken = record { token_address : text; chain_id : nat64 };
+type ErcToken = record {
+	token_address : text;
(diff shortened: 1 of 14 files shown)


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -849,6 +850,22 @@ export const fetchVeloraMarketSwap = async ({
...
+				const currentAllowance = await erc20ContractAllowance({
+					token: sourceToken,
+					owner: userAddress,
+					spender: TokenTransferProxy,
+					networkId: sourceNetwork.id
+				});
+				if (currentAllowance < parsedSwapAmount) {
+					throw new Error(get(i18n).swap.error.unexpected);
+				}
+				return currentAllowance;
```
AntonioVentilii:
```suggestion
```

PR #8513, opened 2025-08-27
feat(frontend): review_send_token tool component

DenysKarmazynDFINITY:
# Motivation

This is the final missing piece to start displaying review_send_token tool in the AI console.

### DIFF
--- a/src/frontend/src/lib/components/ai-assistant/AiAssistantToolResults.svelte
+++ b/src/frontend/src/lib/components/ai-assistant/AiAssistantToolResults.svelte
@@ -3,7 +3,7 @@
 	import AiAssistantReviewSendTokenTool from '$lib/components/ai-assistant/AiAssistantReviewSendTokenTool.svelte';
 	import AiAssistantShowContactsTool from '$lib/components/ai-assistant/AiAssistantShowContactsTool.svelte';
 	import SendTokenContext from '$lib/components/send/SendTokenContext.svelte';
-	import type { ToolResult } from '$lib/types/ai-assistant';
+	import { type ToolResult, ToolResultType } from '$lib/types/ai-assistant';
 
 	interface Props {
 		results: ToolResult[];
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/ai-assistant/AiAssistantToolResults.svelte
```diff
@@ -15,6 +17,10 @@
 	{#each results as { result, type }, index (index)}
 		{#if type === 'show_contacts' && nonNullish(result) && 'contacts' in result}
 			<AiAssistantShowContactsTool {...result} {onSendMessage} />
+		{:else if type === 'review_send_tokens' && nonNullish(result) && 'token' in result}
```
roman-nazaruk: WDYT to move these types to constants to have ability to reuse it?

src/frontend/src/lib/components/ai-assistant/AiAssistantReviewSendTokenTool.svelte
```diff
@@ -0,0 +1,86 @@
...
+	} from '$lib/utils/network.utils';
+
+	type Props = ReviewSendTokensToolResult;
+
+	let { amount, contact, address, contactAddress }: Props = $props();
+
+	const { sendToken, sendTokenExchangeRate, sendTokenNetworkId } =
+		getContext<SendContext>(SEND_CONTEXT_KEY);
+
+	let destination = $derived(contactAddress?.address ?? address ?? '');
```
roman-nazaruk, line 36: Why are we checking contract.address and then using a different address variable? Is that intended?

DenysKarmazynDFINITY: After parsing the send tool LLM response, we are getting either a contact or direct address - if it's the former, we prefer to use it. It is not expected to fallback to an empty string, as we always gonna have either a contact or an address, but it is needed for TS.

roman-nazaruk: Wow, sorry, I thought that you are using ternary operator. My bad.

src/frontend/src/lib/components/ai-assistant/AiAssistantToolResults.svelte
DenysKarmazynDFINITY: yep, sounds good - added

roman-nazaruk: Thanks

roman-nazaruk (approved):
LGTM, thnx

PR #8514, opened 2025-08-27
fix(frontend): Open agreements new tab

daviddecentage:
# Motivation

We need to open the new agreements in a new tab.

# Changes

Changed agreement links


BonomoAlessandro (approved):
LGTM

PR #8515, opened 2025-08-27
feat(backend): Add agreements data to user profile

AntonioVentilii:
# Motivation

We will require the user to accept some agreements to use OISY (for example, License Agreement or Terms of Use).

And we want to store if and when the user accepts (with the timestamp of each agreement version).

# Changes

- Create new Agreement-related structures in the backend.
- Add a new field to the user profile.

# Tests

Current tests should suffice, for now. We will create some more when we have the methods to modify the new field.


src/shared/src/types/agreement.rs
```diff
@@ -0,0 +1,26 @@
...
+pub struct Agreements {
+    pub license_agreement: Agreement,
+    pub terms_of_use: Agreement,
+    pub privacy_policy: Agreement,
+}
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq, Default)]
+pub struct UserAgreements {
+    pub agreements: Agreements,
+}
```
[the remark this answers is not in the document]
AntonioVentilii: we could add more settings for agreements, so we leave it like this

src/shared/src/types/agreement.rs
```diff
@@ -0,0 +1,28 @@
...
+/// Per-agreement status/metadata.
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq, Default)]
+pub struct UserAgreement {
+    /// Whether the user has accepted this agreement (true), rejected it (false), or has not yet
+    /// responded (null).
+    pub accepted: Option<bool>,
+    /// When the user last accepted this agreement (nanos since epoch).
+    pub last_accepted_at: Option<Timestamp>,
+    /// When the agreement itself was last updated (nanos since epoch).
+    pub last_updated_at: Option<Timestamp>,
```
DecentAgeCoder: Why are this two fields optional? They should mandator for the accept and update usecase.

src/declarations/backend/backend.did
```diff
@@ -350,12 +351,23 @@ type TopUpCyclesLedgerResult = variant {
 	Ok : TopUpCyclesLedgerResponse;
 	Err : TopUpCyclesLedgerError
 };
+type UserAgreement = record {
+	last_accepted_at : opt nat64;
+	last_updated_at : opt nat64;
+	accepted : opt bool
+};
```
DecentAgeCoder, line 358: When a user can accept multiple agreements I suggest to introduce an additional field e.g. ref_id that defines which agreement has been accepted.

src/shared/src/types/agreement.rs
AntonioVentilii: So, we currently have 3 states: whether the user has accepted this agreement (true), rejected it (false), or has not yet responded (null).

So, the optionality can tell us if the user still has not even saw the agreements. In that case, we show it to the user, so that it can accept or reject.

src/declarations/backend/backend.did
AntonioVentilii: I am doing it through the `UserAgreements` structure, that has a field for each agreement that needs to be signed (for now we have 3 of them). We decided to go with an object and not an array of agreements because it is easier to manage in the frontend.

```rust
/// The user agreements tracked by the system.
#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq, Default)]
pub struct UserAgreements {
    pub license_agreement: UserAgreement,
    pub terms_of_use: UserAgreement,
    pub privacy_policy: UserAgreement,
}
```

DecentAgeCoder:
@AntonioVentilii I think we could improve the current model so its usage and data integrity can be improved: 

STATE -> UserProfile -> UserAgreements -> UserAgreement 
STATE -> Config --> Agreement < ---------------------|

The relation between Config and Agreement would be: 0 to many (using a hashset or hashmap) with the following entries:

The UserAgreement:
- accepted
- ast_accepted_at
- agreement (immutable reference)

- The Agreement (unique key [type, version]):
  - version
  - type (e.g. user agreement or termrs of use etc.) 
  - name
  - last_updated_at
  - content (extension in the future)


This way:
- each agreement can be agreed independently from each other
- the admin would just need add a new Agreement to the config
- on user login, we can check whether the user has accepted the latest version for each agreement type.
- a user may still reference an older version if they haven’t logged in since an agreement was updated.
- we simplify the user model by avoiding the need for individual models per agreement type.

DecentAgeCoder:
Ah, my mistake. I had assumed the oisy wallet would also have to generate KYC reports for audit purposes in the future (e.g., showing which users agreed to which terms at which date). But if that's not the case, then all good. Approved on my end.

DecentAgeCoder (approved):
Besides the conceptual Q&A, it now LGTM!

PR #8524, opened 2025-08-27
fix(frontend): Fixes agreements scroll intersector

daviddecentage:
# Motivation

The active item of an agreement wasnt always selected correctly in the table of contents.

# Changes

Reduced threshold of when it should be selected to make it more precise (especially for short paragraphs)

# Tests

https://github.com/user-attachments/assets/3b53b07f-74a8-4177-ae48-e7a3f0349495


roman-nazaruk (approved):
LGTM, thnx

PR #8525, opened 2025-08-28
feat(frontend): address book refactor

CvetelinaAngelova:
# Motivation

The AddressBookModal grew to 500+ lines, which made it harder to reason about navigation, async side-effects, and view updates. This refactor splits UI, state, and effects into clear layers so it’s easier to maintain and test. We also add unit tests around the Contacts flows to prevent regressions.

# Changes

Extracted modal logic into three layers:
Stores: addressbook-modal store writable & derived stores for modal state; 
Controller/Services: addressBookModal service. all async handlers: create/update/delete contact, add/save/delete address, avatar updates). Wraps calls with existing wrapCallWith and toggles loading.
UI shell: AddressBookModal.svelte 
Add Config for the AddressBookModal wizard steps

# Tests

Add unit test for the services - addressBookModal.services.spec.ts
Check that the existing test is working corectly.
Manual test ll the functionality and navigation.


src/frontend/src/lib/components/address-book/AddressBookModal.svelte
```diff
@@ -35,96 +35,67 @@
...
-			})
-		)
-	);
+	let modal: WizardModal<AddressBookSteps> | undefined;
+	let editContactNameStep = $state<EditContactNameStep>();
 
 	const steps = $derived(addressBookWizardSteps({ i18n: $i18n }));
 
-	let currentStep: WizardStep<AddressBookSteps> | undefined = $state();
+	const modalData = $derived($modalStore?.data as AddressBookModalParams | undefined);
```
AntonioVentilii:
```suggestion
	let modalData = $derived($modalStore?.data as AddressBookModalParams | undefined);
```

AntonioVentilii:
Large refactor with no activity since Oct 2025; the address book code in main has since evolved. Closing — please reopen with a refresh if still desired.

PR #8531, opened 2025-08-28
fix(frontend): Normalize nft names

daviddecentage:
# Motivation

Sometimes the Nfts include the ID in the Nft name itself. In that case the ID will be displayed twice.

# Changes

We normalize the Nft name by checking if the name includes the ID, if it does we dont add it at the end.

# Tests

Before:
image

After:
image


roman-nazaruk (approved):
LGTM

PR #8534, opened 2025-08-28
feat(backend)!: Methods to update user agreements

AntonioVentilii:
# Motivation

Very similar to what we did for the user settings, we create methods to update the user agreements.

BREAKING CHANGE: the interface has a new method `update_user_agreements` (and related types).

# Changes

- Extend impl methods adding `with_agreements` that verifies if there is any of the inputs that should update the stored values of the user agreements.
- Create methods  to update the user profile.
- Open method in the interface.

# Tests

Added tests.


src/declarations/backend/backend.factory.certified.did.js
```diff
@@ -514,6 +518,7 @@ export const idlFactory = ({ IDL }) => {
 			[]
 		),
 		update_contact: IDL.Func([Contact], [GetContactResult], []),
+		update_user_agreements: IDL.Func([SaveAgreementsRequest], [SetUserShowTestnetsResult], []),
```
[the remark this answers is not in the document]
AntonioVentilii: this is a known issue in the generate script, but it does not affect the final results, since the types are identical

src/declarations/backend/backend.did
```diff
@@ -573,6 +577,19 @@ service : (Arg) -> {
...
+	// set to `now`.
+	//
+	// # Returns
+	// - Returns `Ok(())` if the agreements were saved successfully, or if they were already set to the
+	// same value.
+	//
+	// # Errors
+	// - Returns `Err` if the user profile is not found, or the user profile version is not up-to-date.
+	update_user_agreements : (SaveAgreementsRequest) -> (
+		SetUserShowTestnetsResult
```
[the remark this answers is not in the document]
AntonioVentilii: this is a known issue in the generate script, but it does not affect the final results, since the types are identical

src/declarations/backend/backend.factory.did.js
```diff
@@ -519,6 +523,7 @@ export const idlFactory = ({ IDL }) => {
 			[]
 		),
 		update_contact: IDL.Func([Contact], [GetContactResult], []),
+		update_user_agreements: IDL.Func([SaveAgreementsRequest], [SetUserShowTestnetsResult], []),
```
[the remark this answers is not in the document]
AntonioVentilii: this is a known issue in the generate script, but it does not affect the final results, since the types are identical

src/shared/src/types/agreement.rs
```diff
@@ -26,3 +26,15 @@ pub struct UserAgreements {
...
+
+#[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
+pub enum SaveAgreementsSettingsError {
+    UserNotFound,
+    VersionMismatch,
+}
+
+#[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
+pub struct SaveAgreementsRequest {
+    pub current_user_version: Option<Version>,
```
DecentAgeCoder, line 38: Is this version referring to a single agreement, all agreements, or the user profile? I'm used to handle versioning in the backend: whenever an entity is saved, the backend increments the latest version. The frontend should not be able to overwrite an existing version.

src/backend/src/lib.rs
```diff
@@ -696,7 +699,36 @@ pub fn add_user_hidden_dapp_id(request: AddHiddenDappIdRequest) -> AddUserHidden
...
+///
+/// # Returns
+/// - Returns `Ok(())` if the agreements were saved successfully, or if they were already set to the
+///   same value.
+///
+/// # Errors
+/// - Returns `Err` if the user profile is not found, or the user profile version is not up-to-date.
+#[update(guard = "caller_is_not_anonymous")]
+#[must_use]
+pub fn update_user_agreements(request: SaveAgreementsRequest) -> UpdateUserAgreementsResult {
```
DecentAgeCoder, line 714: If we stick to the CRUD naming pattern I suggest to rename `SaveAgreementsRequest` to `UpdateUserAgreementsRequest`.

DecentAgeCoder (changes requested):
Some minor issues regarding naming and a question

src/shared/src/types/agreement.rs
AntonioVentilii: this is the same version that is used in the other existing methods, so it refers to the user profile that is stored. The versioning is to maintain consistency and give a sort of "race order" to the calls. There is a test to check a bad version call.

Imagine that an user has several windows opened with the same user, or if it does different calls to the user profile in the same instance. Ideally, we want always to approve the call if the version that the user is "using" is the latest

src/backend/src/lib.rs
AntonioVentilii: I am using the same convention that exists for example for the settings. IIRC, it was because theoretically we can update or overwrite the data. This specific method updates it, but if at some point we want to overwrite them too, the parameters will be the same, so we call them generically as "Save" paramenters

AntonioVentilii: If you prefer, I can rename all of them in another PR

src/shared/src/types/agreement.rs
DecentAgeCoder: Yes, that makes perfect sense. I mistakenly thought the version applied to the agreements rather than to the profile.

DecentAgeCoder (approved):
LGTM and thanks for the input you have provided!

PR #8539, opened 2025-08-28
feat(frontend): enabledNetworksSymbols derived store

DenysKarmazynDFINITY:
# Motivation

For LLM to know which networks are currently enabled, we create a special derived store.


AntonioVentilii (approved):
LGTM tks!

PR #8545, opened 2025-08-28
fix(frontend): added hr for swapReview modal

roman-nazaruk:
# Motivation

We need to add some spaces between component in SwapReview Modal

# Changes

Added Hr component in SwapReviewModal


daviddecentage (approved):
LGTM

PR #8553, opened 2025-08-29
feat(frontend): Derived stores for analysis of user agreements

AntonioVentilii:
# Motivation

To simplify our code and the analysis of the user agreements, we create some useful derived stores.

# Changes

- Derived store to check if the user has never visualized the agreements: it is useful, since we show a different message in this case.
- Derived store with the list of outdated and/or rejected agreements.
- Derived store to check if the user has any outdated agreements: useful since we will show ONLY the outdated agreements.
- Derived store to check if all agreements are accepted and up-to-date: this will be the main guard to proceed with the usage of OISY.

# Tests

Added tests.


daviddecentage (approved):
LGTM, thanks!

PR #8554, opened 2025-08-29
refactor(frontend): Agreements json bigint

daviddecentage:
# Motivation

We are getting the timestamp for agreements accepted in bigint type, therefor we need to adjust our frontend to deal with big ints.

# Changes

Added transform utils
Adjusted all types to bigint
Adjusted agreements json to bigint

# Tests

Adjusted and extended tests


AntonioVentilii (approved):
LGTM tks!

PR #8556, opened 2025-08-29
refactor(frontend): Move `transformAgreementsJsonBigint` to specific module

AntonioVentilii:
# Motivation

Util `transformAgreementsJsonBigint` is used to set agreementsData, that by itself it will be used inside the `agreements.utils` module, risking a looped called.

To avoid it, we move it to a different module, specific for it.


daviddecentage (approved):
LGTM

PR #8557, opened 2025-08-29
feat(frontend): principal/accountId flag for LLM to filter contacts

DenysKarmazynDFINITY:
# Motivation

For LLM to properly filter contacts in the send IC/ICRC tokens flow, we need to introduce an additional flag `isPrincipal` and update the prompts accordingly.


AntonioVentilii (approved):
LGTM tks!

PR #8566, opened 2025-08-29
refactor(frontend): Move `AuthClient` creation methods to API module

AntonioVentilii:
# Motivation

As per suggestion in https://github.com/dfinity/oisy-wallet/pull/8398#discussion_r2309713548, we move the creation methods of AuthClient inside the auth-client API module.


src/frontend/src/lib/utils/auth.utils.ts
```diff
@@ -1,29 +1,9 @@
-import { authClientStorage } from '$lib/api/auth-client.api';
+import { createAuthClient } from '$lib/api/auth-client.api';
```
peterpeterparker: Should everything move to the api?
i.e. not having references on api from utils at all

AntonioVentilii: No preference here: I just thought we still wanted to segregate the `loadIdentity` util, just to have it in a different scope, boh

What do you prefer?

peterpeterparker: I would move `loadIdentity` to the api as well and keep the stateless `isAlternativeOrigin` and `getOptionalDerivationOrigin` in the utils.

AntonioVentilii: Done!

PR #8572, opened 2025-08-30
feat(frontend): render AI console inside the footer

DenysKarmazynDFINITY:
# Motivation

We need to move the AI console inside the footer so the soical icons can be properly rendered when the console is open. It also solves the positioning issues.

Screenshot 2025-08-30 at 12 01 24

Screenshot 2025-08-30 at 12 01 17


daviddecentage (approved):
LGTM

PR #8579, opened 2025-09-01
feat(frontend): implements mapCustomTokenSection function

BonomoAlessandro:
# Motivation

Since we are able to map `TokenSection` to `CustomTokenSection` we also want to be able to map it back.

# Changes

- implements `mapCustomTokenSection` function


daviddecentage (approved):
LGTM

PR #8591, opened 2025-09-01
test(backend): Test if a user agreement is updated with the date

AntonioVentilii:
# Motivation

A small test to verify that we really modify the user agreements even if just the updated-date is modified.


src/shared/src/impls.rs
```diff
@@ -199,21 +200,22 @@ impl TokenVersion for StoredUserProfile {
...
 }
 
+impl UserAgreement {
+    fn without_accept_ts(mut self) -> Self {
+        self.last_accepted_at_ns = None;
+        self
+    }
+}
+
 impl UserAgreements {
```
DecentAgeCoder: You should be able to remove this dedicated equals implementation.

You can basically just call:
        if current.eq(&new_agreements) {
            return Ok(self.clone());
        }

AntonioVentilii: changed

DecentAgeCoder (approved):
LGTM! Thanks

PR #8592, opened 2025-09-01
refactor(backend): Create result type for method `update_user_network_settings`

AntonioVentilii:
# Motivation

To be more consistent, we can use a specific result type for method `update_user_network_settings`.


src/shared/src/types/result_types.rs
```diff
@@ -139,16 +139,31 @@ impl From<Result<Vec<Contact>, ContactError>> for GetContactsResult {
...
     /// The user's show testnets was set successfully.
     Ok(()),
     /// The user's show testnets was not set due to an error.
     Err(SetTestnetsSettingsError),
 }
-
-impl From<Result<(), SetTestnetsSettingsError>> for SetUserShowTestnetsResult {
-    fn from(result: Result<(), SetTestnetsSettingsError>) -> Self {
+impl From<Result<(), SaveTestnetsSettingsError>> for SetUserShowTestnetsResult {
+    fn from(result: Result<(), SaveTestnetsSettingsError>) -> Self {
```
AntonioVentilii:
```suggestion
impl From<Result<(), SetTestnetsSettingsError >> for SetUserShowTestnetsResult {
    fn from(result: Result<(), SetTestnetsSettingsError >) -> Self {
```

DecentAgeCoder (approved):
LGTM! Thanks

PR #8604, opened 2025-09-01
feat(frontend): implements NftCollectionActionButtons

BonomoAlessandro:
# Motivation

We want to display the different possible actions, like marking the collaction as spam or hidden, on the nft collection page.

# Tests

image

image

image


daviddecentage (approved):
LGTM

PR #8606, opened 2025-09-01
fix(frontend): Correct strings when showing agreements modal

AntonioVentilii:
# Motivation

THe strings for never-seen and updated agreements were inverted in the modal.


daviddecentage (approved):
LGTM

PR #8607, opened 2025-09-01
refactor(backend): Expand agreements when updating

AntonioVentilii:
# Motivation

Small refactoring to catch all possible fields of agreements when updating.


DecentAgeCoder (approved):
LGTM!

PR #8619, opened 2025-09-02
fix(frontend): Remove flickering of Agreements modal

AntonioVentilii:
# Motivation

If the user already accepted the agreements and logs in, there is an instant where the Agreements modal shows. That is because the user profile is still not loaded. We adjust the User profile loader guard to await a loaded user profile.

### Before

https://github.com/user-attachments/assets/907779c2-1032-474a-b72b-ca24ab18af66

### After

https://github.com/user-attachments/assets/b8498aae-6dec-4c75-a26a-13d48cba0ca1


BonomoAlessandro (approved):
LGTM

PR #8634, opened 2025-09-02
test(e2e): fix homepage e2e test

SYInacta:
# Motivation

Fix e2e tests so they follow and run 

# Changes

Made changes to homepage e2e tests.
Added script to run e2e tests with all checks.

# Tests

Homepage tests.


e2e/utils/pages/homepage.page.ts
```diff
@@ -101,6 +101,22 @@ abstract class Homepage {
 		this.#isMobile = isMobile;
 	}
 
+	private async closeBanners(): Promise<void> {
+		const betaBannerCloseButton = this.#page
```
AntonioVentilii, line 105: do we have the banner in E2E env? i don't think that is the intention, so we should remove it at source then

e2e/utils/pages/homepage.page.ts
```diff
@@ -101,6 +101,22 @@ abstract class Homepage {
...
+			.locator('div:has-text("This is a beta release")')
+			.locator('[aria-label="Close"]');
+		if (await betaBannerCloseButton.isVisible()) {
+			await betaBannerCloseButton.click();
+		}
+
+		const testBannerCloseButton = this.#page
```
AntonioVentilii, line 112: same

AntonioVentilii:
Stale E2E fix from 2025 with no activity for >6 months; the affected tests/pages have changed since. Closing — please reopen against current main if still needed.

PR #8639, opened 2025-09-02
feat(frontend): Persist WalletConnect connection on reload

AntonioVentilii:
# Motivation

When we reload the page, we would like to persist the WalletConnect connection. 

# Changes

- Add method `getActiveSessions` to WalletConnect `listener` object.
- Reactively, if all the necessary addresses are loaded, try to reconnect, without deleting existing sessions.
- If no active sessions present, we reset the `listener` for good measure.

# Tests

Manual test:

https://github.com/user-attachments/assets/25716c6c-e2d2-432a-a90d-9d282118f33c

### DIFF
--- a/src/frontend/src/eth/derived/erc721.derived.ts
+++ b/src/frontend/src/eth/derived/erc721.derived.ts
@@ -3,7 +3,6 @@ import { erc721CustomTokensStore } from '$eth/stores/erc721-custom-tokens.store'
 import type { Erc721CustomToken } from '$eth/types/erc721-custom-token';
 import type { Erc721TokenToggleable } from '$eth/types/erc721-token-toggleable';
 import { enabledEvmNetworksIds } from '$evm/derived/networks.derived';
-import { nonNullish } from '@dfinity/utils';
 import { derived, type Readable } from 'svelte/store';
 
 /**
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/wallet-connect/WalletConnectSession.svelte
```diff
@@ -364,6 +371,55 @@
 
 	$: walletConnectPaired.set(nonNullish(listener));
 
+	const reconnect = async () => {
+		// If the listener is already initialized, we don't need to do anything.
+		if (nonNullish(listener)) {
+			return;
+		}
+
+		if ($loading || isNullish($ethAddress) || isNullish($solAddressMainnet)) {
```
DenysKarmazynDFINITY: `initWalletConnect` expects sol address as an optional param - do we want to return early if it's null here?

AntonioVentilii: Good catch! I copied the same logic as the connection, but it is wrong there too... Basically, it should raise an error if ETH address is nullish AND (not OR) SOL mainnet address is nullish


I will modify it here and adjust the rest in another PR.

PR #8653, opened 2025-09-02
feat(frontend): reset AI chat history button

DenysKarmazynDFINITY:
# Motivation

We need to add a button that allows user to reset AI chat history.

Screenshot 2025-09-02 at 16 58 02


src/frontend/src/lib/components/ai-assistant/AiAssistantConsole.svelte
```diff
@@ -112,43 +113,59 @@
 
 		<h5 class="mx-2 w-full">{replaceOisyPlaceholders($i18n.ai_assistant.text.title)}</h5>
 
+		<button
```
AntonioVentilii, line 116: can we use the existing `<Button>` component?

src/frontend/src/lib/components/ai-assistant/AiAssistantConsole.svelte
```diff
@@ -112,43 +113,59 @@
...
+			class:text-tertiary={!loading}
+			class:text-tertiary-inverted={loading}
+			aria-label={$i18n.ai_assistant.text.reset_chat_history}
+			disabled={loading}
+			onclick={aiAssistantStore.resetChatHistory}
+		>
+			<IconRepeat size="18" />
+		</button>
+
 		<button
```
AntonioVentilii, line 128: just noticed: here too maybe?

src/frontend/src/lib/components/ai-assistant/AiAssistantConsole.svelte
DenysKarmazynDFINITY: the problem was that Button comes with some styles that I would need to overwrite here with `!`, therefore I thought it was easier not to use it at all

AntonioVentilii (approved):

PR #8663, opened 2025-09-03
feat(frontend): enables show spam toggle

BonomoAlessandro:
# Motivation

The `Show spam` toggle should be enabled and the made selection should be saved.

# Tests

https://github.com/user-attachments/assets/abf2adcf-b4a3-49b5-9fc6-7ed89e757312


src/frontend/src/lib/components/nfts/NftSettingsMenu.svelte
```diff
@@ -21,6 +23,11 @@
 	const setGrouping = (grouping: boolean) => {
 		nftListStore.setGroupByCollection(grouping);
 	};
+
+	const toggleShowSpam = () => {
+		document.dispatchEvent(new CustomEvent('show-spam'));
```
AntonioVentilii: we don't typically use this. Is there a reason? In Svelte 5 we don't dispatch anymore, we pass it as prop, like `onShowSpam` or something similar

EDIT: just checked and we use it in `document.dispatchEvent(new CustomEvent('toggleHideZeros'))` too. However, that is wrong, we should use the `dispatchEvent` util (if not in Svelte 5). I will correct it in another PR

src/frontend/src/lib/components/nfts/NftSettingsMenu.svelte
```diff
@@ -21,6 +23,11 @@
...
+		emit({ message: 'oisyToggleShowSpam' });
```
AntonioVentilii, line 32: why emitting this one too? is it not already handled by the consumer?

src/frontend/src/lib/components/nfts/NftsShowSpamToggle.svelte
```diff
@@ -0,0 +1,14 @@
...
+	import { showSpam } from '$lib/derived/settings.derived';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { showSpamStore } from '$lib/stores/settings.store';
+
+	let checked = $derived($showSpam);
+
+	const toggleShow = () => showSpamStore.set({ key: 'show-spam', value: { enabled: !checked } });
+</script>
+
+<svelte:window on:oisyToggleShowSpam={toggleShow} />
```
AntonioVentilii, line 12: why this one? isn't the trigger handled only in this flow? Do you expect the event o be emitted somewhere else too?

AntonioVentilii (changes requested):

src/frontend/src/lib/components/nfts/NftSettingsMenu.svelte
AntonioVentilii: Ah ok, I understand why: if we click in the component that has the toggle as child, we want to trigger the toggle too. it is a bit of a workaround, but it is elsewhere in the codebase already. I will think of another solution, but for this PR, all good

src/frontend/src/lib/components/nfts/NftsShowSpamToggle.svelte
AntonioVentilii: Same answer given in:

- https://github.com/dfinity/oisy-wallet/pull/8663#discussion_r2318370127

src/frontend/src/lib/components/nfts/NftSettingsMenu.svelte
BonomoAlessandro: Like this the flag is also set when the button surrounding the toggle element is clicked.

src/frontend/src/lib/components/nfts/NftsShowSpamToggle.svelte
BonomoAlessandro: We listen on the event that is thrown that i mentioned above

BonomoAlessandro:
@AntonioVentilii This PR has to be closed. We need to implement that feature in another way.

src/frontend/src/lib/components/nfts/NftSettingsMenu.svelte
AntonioVentilii: Here the PR to remove it in the other component:

- https://github.com/dfinity/oisy-wallet/pull/8665

AntonioVentilii: I think you can simply remove this line, does it work?

BonomoAlessandro: You are right, i can just remove it.

src/frontend/src/lib/components/nfts/NftsShowSpamToggle.svelte
```diff
@@ -0,0 +1,14 @@
...
+
+<Toggle ariaLabel={$i18n.nfts.text.show_spam} bind:checked on:nnsToggle={toggleShow} />
```
AntonioVentilii: if `checked` is a `$derived`, i think you can unbind the value

```suggestion
<Toggle ariaLabel={$i18n.nfts.text.show_spam} {checked} on:nnsToggle={toggleShow} />
```

BonomoAlessandro: You are right

PR #8664, opened 2025-09-03
feat(frontend): enables show hidden toggle

BonomoAlessandro:
# Motivation

The `Show hidden` toggle should be enabled and the made selection should be saved.

# Tests

https://github.com/user-attachments/assets/85c0afe4-00d5-4034-aa17-061b35adffde


src/frontend/src/lib/components/nfts/NftsShowHiddenToggle.svelte
```diff
@@ -0,0 +1,15 @@
...
+
+	let checked = $derived($showHidden);
+
+	const toggleHidden = () =>
+		showHiddenStore.set({ key: 'show-hidden', value: { enabled: !checked } });
+</script>
+
+<svelte:window on:oisyToggleShowHidden={toggleHidden} />
+
+<Toggle ariaLabel={$i18n.nfts.text.show_hidden} bind:checked on:nnsToggle={toggleHidden} />
```
AntonioVentilii: checked is a `$derived`, so i think you can unbind it

```suggestion
<Toggle ariaLabel={$i18n.nfts.text.show_hidden} {checked} on:nnsToggle={toggleHidden} />
```

PR #8671, opened 2025-09-03
feat(frontend): Single IDB cache for custom tokens

AntonioVentilii:
# Motivation

The IDB cache for custom tokens is divided in networks. However, we have several kind of standard (or backend variant) that are not being saved correctly: since we filter the backend custom tokens by standard (or backend variant), for example, when saving ERC721, we overwrite ERC20, and viceversa.

It is not scalable to have a single IDB store for each variant or type of token. And it does not make sense to save all networks custom tokens in each network IDB cache.

Since we have a single service to load tokens from the backend (and consequently saving them in the IDB cache) and a single service to load them for all the networks, we can have a single IDB cache for all of them.

# Changes

- Create method to set all custom tokens in the cache and use it in service `loadCustomTokensFromBackend`.
- Create method to get all custom tokens in the IDB cache and use it in service `loadNetworkCustomTokens` when the use of cache is enabled (note that we need to filter afterwards for the specific network/variant).
- Create methods to delete/clean the IDB cache and use it in the appropriate service.
- Remove all deprecated IDB methods for the rest of the custom tokens.

# Tests

Tests adapted.

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftList.svelte
+++ b/src/frontend/src/lib/components/nfts/NftList.svelte
@@ -1,0 +1,31 @@
+<script lang="ts">
+	import type { Snippet } from 'svelte';
+	import NftCard from '$lib/components/nfts/NftCard.svelte';
+	import type { Nft } from '$lib/types/nft';
+
+	interface Props {
+		title: string;
+		icon?: Snippet;
+		nfts: Nft[];
+		hidden?: boolean;
+		spam?: boolean;
+		testId?: string;
+	}
+
+	let { title, icon, nfts, hidden, spam, testId }: Props = $props();
+</script>
+
+{#if nfts.length > 0}
+	<div data-tid={testId}>
+		<div class="mt-5 flex items-center gap-2">
+			{@render icon?.()}
+			<h5>{title}</h5>
+		</div>
+
+		<div class="grid grid-cols-3 gap-3 gap-y-4 py-4">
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/services/custom-tokens.services.ts
```diff
@@ -12,43 +12,36 @@ interface LoadCustomTokensFromBackendParams {
...
 		nullishIdentityErrorMessage: get(i18n).auth.error.no_internet_identity
 	});
 
-	// We filter the custom tokens, since the backend "Custom Token" potentially supports other types
-	const filteredTokens = tokens.filter(filterTokens);
-
 	// Caching the custom tokens in the IDB if update call
-	if (certified && filteredTokens.length > 0) {
-		await setIdbTokens({ identity, tokens: filteredTokens });
+	if (certified && tokens.length > 0) {
```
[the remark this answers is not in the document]
AntonioVentilii: in this PR we will not change the logic, but yes, well noticed, it should save empty tokens too. Will do in another PR

.github/workflows/pr-checks.yml
```diff
@@ -8,7 +8,6 @@ on:
       - reopened
       - synchronize
       - labeled
-  merge_group:
```
DenysKarmazynDFINITY: I guess it's not from this PR, but if you want to include it, I don't mind :)

AntonioVentilii: nono! it was for another PR! removed, good catch

PR #8676, opened 2025-09-03
feat(frontend): AI console automatic scrolling

DenysKarmazynDFINITY:
# Motivation

We want the following behaviour to be implemented: 
* when the log is at the bottom, and new content comes in (user entry or LLM), then scroll down automatically
* f the user manually scrolls up a bit, avoid automating scrolling to the bottom
* user types and sends - always scroll to the bottom

### DIFF
--- a/src/frontend/src/lib/components/ai-assistant/AiAssistantConsole.svelte
+++ b/src/frontend/src/lib/components/ai-assistant/AiAssistantConsole.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { IconClose } from '@dfinity/gix-components';
-	import { isNullish } from '@dfinity/utils';
+	import { isNullish, nonNullish } from '@dfinity/utils';
 	import { fade } from 'svelte/transition';
 	import AiAssistantActionButton from '$lib/components/ai-assistant/AiAssistantActionButton.svelte';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/ai-assistant/AiAssistantConsole.svelte
```diff
@@ -39,6 +41,29 @@
...
+
+		// 10 - a small tolerance in scroll measurements
+		shouldScrollMessagesContainer =
+			messagesContainer.scrollTop + messagesContainer.clientHeight >=
+			messagesContainer.scrollHeight - 10;
+	};
+
+	$effect(() => {
+		// add messages as a dependency to trigger the scroll when new content arrives
+		messagesToDisplay;
```
AntonioVentilii, line 57: theoretically you could add the condition `messagesToDisplay.length >0` ?

AntonioVentilii (approved):
LGTM tks

src/frontend/src/lib/components/ai-assistant/AiAssistantConsole.svelte
DenysKarmazynDFINITY: yes, good idea - replaced

PR #8679, opened 2025-09-03
refactor(frontend): updates NftCardSkeleton component

BonomoAlessandro:
# Motivation

The `NftCardSkeleton` should be more dynamic. Now it takes a parameter that indicates how many times the skeleton should be display.


AntonioVentilii (approved):
Nice!

PR #8682, opened 2025-09-03
fix(frontend): Pass correct address to Solana WalletConnect sign modal

AntonioVentilii:
# Motivation

To solve a pending TODO, we pass the source address directly to the Solana WalletConnect modal.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8690, opened 2025-09-03
feat(lint): Add `ls-lint` configuration

AntonioVentilii:
# Motivation

As suggested by @peterpeterparker, we want to start using the linter for naming the files and the folders too. To help us, there is the library [ls-lint](https://ls-lint.org/). 

As first step, we add some basic configuration file as per tutorial, and we run it manually locally, fixing any issue.


DenysKarmazynDFINITY (approved):
Good one!

PR #8691, opened 2025-09-04
feat(frontend): Open Nft in fullscreen

daviddecentage:
# Motivation

# Changes

# Tests


src/frontend/src/lib/components/nfts/NftHero.svelte
```diff
@@ -64,7 +66,19 @@
...
+						<button
+							class="block h-auto max-h-full w-auto max-w-full border-0"
+							onclick={() =>
+								modalStore.openNftFullscreenDisplay({
+									id: Symbol('nft-fullscreen-display'),
+									data: nft
+								})}
+						>
+							<Img
+								src={nft?.imageUrl}
```
DenysKarmazynDFINITY: Shall we maybe check if nft?.imageUrl nonNullish before calling modalStore.openNftFullscreenDisplay?

src/frontend/src/lib/components/nfts/NftHero.svelte
```diff
@@ -64,7 +66,19 @@
...
+								styleClass="block h-auto w-auto max-h-full max-w-full object-contain"
```
DenysKarmazynDFINITY: maybe we can simplify the classes a bit - do both button and img need `h-auto w-auto max-h-full max-w-full`?

src/frontend/src/lib/components/nfts/NftHero.svelte
daviddecentage: We do a few lines above, but I adjusted to remove the optional chaining `?`

DenysKarmazynDFINITY (approved):
Minor comments, approving already

src/frontend/src/lib/components/nfts/NftHero.svelte
daviddecentage: Yes, I was able to strip some classes

PR #8692, opened 2025-09-04
feat(frontend): Fullscreen img modal

daviddecentage:
# Motivation

I thought it would be nice to display the Nft in a fullscreen modal to be able to have a better look at the Nft image.

For that we add a fullscreen image modal component.

# Changes

Added component

# Tests

Added tests

Large screens
image

Mobile:
image

Horizontal images:
image


src/frontend/src/lib/components/ui/FullscreenImgModal.svelte
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	let { imageSrc }: Props = $props();
+</script>
+
+<div class="fixed bottom-0 left-0 right-0 top-0 z-10" transition:fade>
+	<div class="pointer-events-none absolute right-3 top-3 z-10 md:right-8 md:top-8"
+		><IconClose /></div
+	>
+	<div class="z-9 pointer-events-none absolute inset-0 grid place-items-center">
```
DenysKarmazynDFINITY, line 19: just wondering - any particular reason to use grid here instead of flex?

src/frontend/src/lib/components/ui/FullscreenImgModal.svelte
```diff
@@ -0,0 +1,39 @@
...
+				<Img src={imageSrc} />
+			</div>
+		</div>
+	</div>
+	<Backdrop on:nnsClose={() => modalStore.close()} />
+</div>
+
+<style lang="scss">
+	/* Ensure images (or videos, etc.) shrink but never upscale */
+	:global(.fullscreen-modal img) {
```
DenysKarmazynDFINITY: `Img` component accepts `styleClass` prop - can we use it instead of scss?

src/frontend/src/lib/components/ui/FullscreenImgModal.svelte
daviddecentage: Yeah so I had flex originally, but I couldnt get it to work with centering + restricting the max size of the image to 90vh/vw AND making it grow to that maximum restrictions AND keeping proportions of the image. It was quite tricky, maybe there is a way with flex but its so many different factors to get it to behave correctly that Im honestly happy to use grid here :)

src/frontend/src/lib/components/ui/FullscreenImgModal.svelte
```diff
@@ -0,0 +1,39 @@
...
+		max-width: 90dvw;
```
DenysKarmazynDFINITY: Since we already assigned width/height values to the parent, can we inherit it via width: 100%/height: 100% instead of setting it again?

src/frontend/src/lib/components/ui/FullscreenImgModal.svelte
```diff
@@ -0,0 +1,39 @@
...
+	let { imageSrc }: Props = $props();
+</script>
+
+<div class="fixed bottom-0 left-0 right-0 top-0 z-10" transition:fade>
+	<div class="pointer-events-none absolute right-3 top-3 z-10 md:right-8 md:top-8"
+		><IconClose /></div
+	>
+	<div class="z-9 pointer-events-none absolute inset-0 grid place-items-center">
+		<div class="fullscreen-modal max-h-[90vh] max-w-[90vw] overflow-hidden">
+			<div class="overflow-hidden rounded-lg">
```
DenysKarmazynDFINITY: can we pass `rounded-lg` as `styleClass` to Img and avoid having this div?

src/frontend/src/lib/components/ui/FullscreenImgModal.svelte
daviddecentage: Good call, I was able to strip away quite a lot. I intended it to be used with children snippet first to not make it exclusively about images, so I tried making it so it will always work for children. Now that we just render the image in the component itself we can strip all the compatibility wrappers and pass it to the img instead!

src/frontend/src/lib/components/ui/FullscreenImgModal.svelte
daviddecentage: Yes, see my comment above

src/frontend/src/lib/components/ui/FullscreenImgModal.svelte
daviddecentage: Yes, see my comment below

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5064, opened 2025-03-06
test(frontend): compare real Solana transactions with total balance

AntonioVentilii:
# WIP

# Motivation

# Changes

# Tests

PR #8698, opened 2025-09-04
test(frontend): Create fixtures for Solana integration tests

AntonioVentilii:
# Motivation

As suggested by prodsec, we should have some integration test on Solana that checks if our mapping/parsing of the Solana/SPL transactions matches the balance fetched from the provider.

We will create the tests in PR https://github.com/dfinity/oisy-wallet/pull/5064. However, the tests could use the Alchemy provider extensively, and it has rate limit for requests.

It does not really make sense to continuously fetch real data, if we can take a snapshot now that is more that satisfactory for the tests.

So, we create fixtures for that specific test. In this PR, we create a script that saves the fixtures (all public data) and we run it, saving the results.

# Changes

- Create list of addresses and ATA addresses to use for the fixtures and the tests.
- Create script to save the fixtures:
  - for addresses:
    - transactions
    - signatures
    - SOL balances
  - for ATA addresses:
    - transactions
    - SPL balances

# Tests

We manually ran the script and saved the results.


src/frontend/src/tests/fixtures/solana/5Dqoon9MdWRgwmJ839FJ2ZTpTAcc1MMprZeNyaxpaV1Q/tokens/Xsc9qvGR1efVDFGLrVsmkzv3qi45LTBjeUKSPmx9qEh/transactions/before-5hA1inhymrZEGiugT5TNYQTAjJ7QXaXskPvmhUm9MiEFExuAS9avk4CS6Cnr4o2A1Qdw9hcJ5LcrSn7Cqha9JUG9.json
```diff
@@ -0,0 +1 @@
+[]
```
DenysKarmazynDFINITY, line 1: is it correctly empty?

AntonioVentilii: yes, it's the last request according `before` parameter: there are not more signatures/transactions

PR #8708, opened 2025-09-08
feat(frontend): prevents loading nfts

BonomoAlessandro:
# Motivation

As long as the `NFT_ENABLED` feature flag is disabled no nft related data should be loaded.


AntonioVentilii (approved):
Thank you!

PR #8721, opened 2025-09-08
feat(backend)!: experimental features

DenysKarmazynDFINITY:
# Motivation

We need to extend user's settings with an additional param - experimental features. At the moment, there will be just one possible entry for the AI assistant feature. The idea is to make as flexible as possible, therefore for the data structure we use the same approach as it's done for the network settings - each experimental feature can be prefixed in the future with an additional property (e.g. `FeatureBeta`, `FeatureAlpha`, and so on) similar to the networks (`BitcoinMainnet`, `BitcoinTestnet`, etc.). The only option at the moment is gonna be `AiAssistantBeta`.

BREAKING CHANGE: includes new functionality.

### DIFF
--- a/src/frontend/src/eth/providers/alchemy.providers.ts
+++ b/src/frontend/src/eth/providers/alchemy.providers.ts
@@ -3,13 +3,11 @@ import { SUPPORTED_ETHEREUM_NETWORKS } from '$env/networks/networks.eth.env';
 import { ALCHEMY_API_KEY } from '$env/rest/alchemy.env';
 import type { AlchemyProviderContracts } from '$eth/types/alchemy-contract';
 import type { AlchemyProviderOwnedNfts } from '$eth/types/alchemy-nfts';
-import type { Erc1155ContractAddress } from '$eth/types/erc1155';
-import type { Erc721ContractAddress } from '$eth/types/erc721';
 import { i18n } from '$lib/stores/i18n.store';
 import type { EthAddress } from '$lib/types/address';
 import type { WebSocketListener } from '$lib/types/listener';
 import type { NetworkId } from '$lib/types/network';
-import type { Nft, NonFungibleToken, OwnedContract, OwnedNft } from '$lib/types/nft';
+import type { Nft, NonFungibleToken, OwnedContract } from '$lib/types/nft';
 import type { TokenStandard } from '$lib/types/token';
 import type { TransactionResponseWithBigInt } from '$lib/types/transaction';
 import { replacePlaceholders } from '$lib/utils/i18n.utils';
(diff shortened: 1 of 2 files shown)


src/shared/src/types/experimental_feature.rs
```diff
@@ -0,0 +1,67 @@
+//! Level 0 experimental eatures
```
AntonioVentilii: can this be removed?

src/shared/src/types/result_types.rs
```diff
@@ -174,14 +175,14 @@ impl From<Result<(), SetTestnetsSettingsError>> for SetUserShowTestnetsResult {
 #[derive(CandidType, Deserialize, Clone, Eq, PartialEq, Debug)]
 pub enum GetUserProfileResult {
     /// The user's profile was retrieved successfully.
-    Ok(UserProfile),
+    Ok(Box<UserProfile>),
```
DenysKarmazynDFINITY, line 178: This change was forced by linter - it is most likely related to the `settings` growing in size. The tests are running fine, but please let me know if it had to be fixed somehow differently.

src/shared/src/types/experimental_feature.rs
```diff
@@ -0,0 +1,65 @@
...
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq, Default)]
+pub struct ExperimentalFeatureSettings {
+    pub enabled: bool,
+}
+
+/// A flat list of logical experimental features.
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq, Default, Ord, PartialOrd)]
+pub enum ExperimentalFeatureSettingsFor {
+    #[default]
```
AntonioVentilii: no expert, but i think we can remove the `Default` trait here and the `#default` too, since we don't initialize it anywhere as default, right? The `ExperimentalFeatureSettingsMap` defaults to empty, or not?

src/shared/src/types/experimental_feature.rs
```diff
@@ -0,0 +1,65 @@
...
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub enum ExperimentalFeatureId {
+    AiAssistant(AiAssistantExperimentalFeatureId),
+}
+impl ExperimentalFeature for ExperimentalFeatureId {}
+
+#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq, Default)]
+#[repr(u64)]
+pub enum AiAssistantExperimentalFeatureId {
+    #[default]
```
AntonioVentilii: same here maybe?

src/shared/src/types/experimental_feature.rs
DenysKarmazynDFINITY: Yes, by default it's empty - removed it. Thanks!

src/shared/src/types/experimental_feature.rs
DenysKarmazynDFINITY: yep, removed

PR #8723, opened 2025-09-08
feat(frontend): implements getNftsByOwner alchemy api endpoint

BonomoAlessandro:
# Motivation

We want to have a new `getNftsByOwner` alchemy api function to be able to load the nft metadata.

### DIFF
--- a/src/frontend/src/tests/eth/providers/alchemy.providers.spec.ts
+++ b/src/frontend/src/tests/eth/providers/alchemy.providers.spec.ts
@@ -186,13 +186,13 @@ describe('alchemy.providers', () => {
 					id: parseNftId(1),
 					collection: {
 						...mapTokenToCollection(mockValidErc1155Token)
-					},
+					}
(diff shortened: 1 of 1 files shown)


src/frontend/src/tests/eth/providers/alchemy.providers.spec.ts
```diff
@@ -80,6 +81,77 @@ describe('alchemy.providers', () => {
...
+
+			const nfts = await provider.getNftsByOwner({
+				address: mockEthAddress,
+				token: mockValidErc1155Token
+			});
+
+			expect(Alchemy.prototype.nft.getNftsForOwner).toHaveBeenCalledOnce();
+			expect(nfts).toStrictEqual(expectedTokenIds);
+		});
+	});
```
AntonioVentilii, line 216: can you add some tests that shows how it handles errors, and another for missing data. Just for coverage

PR #8724, opened 2025-09-08
feat(devops): Custom env file for deployment

AntonioVentilii:
# Motivation

For test environments, it is useful to provide a custom env file, that override any default secret.

# Tests

Manually tried in https://github.com/dfinity/oisy-wallet/actions/runs/17552460777/job/49848005009


daviddecentage (approved):
LGTM

PR #8727, opened 2025-09-08
feat(frontend): rewrites LoaderNfts to use alchemy

BonomoAlessandro:
# Motivation

To be more efficient we want to use alchemy to load the nft metadata. To achieve this, the `LoaderNfts` component needs a rework.

### DIFF
--- a/src/frontend/src/lib/components/loaders/LoaderNfts.svelte
+++ b/src/frontend/src/lib/components/loaders/LoaderNfts.svelte
@@ -52,10 +52,10 @@
 		}
 
 		for (const token of $enabledNonFungibleTokens) {
-			const alchemyProvider = alchemyProviders(token.network.id);
+			const { getNftsByOwner } = alchemyProviders(token.network.id);
 
 			try {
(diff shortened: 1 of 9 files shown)


src/frontend/src/lib/components/loaders/LoaderNfts.svelte
```diff
@@ -133,16 +47,27 @@
...
 		for (const token of $enabledNonFungibleTokens) {
-			if (isTokenErc721(token)) {
-				await handleErc721(token);
-			}
-			if (isTokenErc1155(token)) {
-				await handleErc1155(token);
+			const alchemyProvider = alchemyProviders(token.network.id);
+
+			try {
+				const nfts = await alchemyProvider.getNftsByOwner({ address: $ethAddress, token });
```
AntonioVentilii:
```suggestion
			const {getNftsByOwner} = alchemyProviders(token.network.id);

			try {
				const nfts = await getNftsByOwner({ address: $ethAddress, token });
```

PR #8731, opened 2025-09-09
feat(frontend): updates nft service implementation

BonomoAlessandro:
# Motivation

To be more efficient we want to use alchemy to load the nft metadata. To achieve this, the `nft.services` implementation needs a rework.


AntonioVentilii (approved):
LGTM tks!

PR #8734, opened 2025-09-09
feat(frontend): the BTC wallet balance calculation to btc wallet worker

DecentAgeCoder:
# Motivation

# Changes

# Tests


src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -57,6 +64,20 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
 		latestBitcoinBlockHeight: undefined
 	};
 
+	/**
+	 * Convert ckBTC BitcoinNetwork type to backend BitcoinNetwork variant type
+	 */
+	private mapToBackendBitcoinNetwork(network: BitcoinNetwork): BackendBitcoinNetwork {
```
DenysKarmazynDFINITY: I think you can use `mapToSignerBitcoinNetwork` here, can you please check it?

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -122,37 +175,57 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
...
-			};
-		}
+		certified = true,
+		pendingTransactions = [],
+		uncertifiedTransactions = []
+	}: Omit<LoadBtcWalletParams, 'shouldFetchTransactions'> & {
+		pendingTransactions: CertifiedData<PendingTransaction>[];
+		uncertifiedTransactions?: CertifiedData<BtcTransactionUi>[];
+	}): Promise<CertifiedData<BtcWalletBalance | null>> => {
+		let confirmedBalance: bigint | null;
```
DenysKarmazynDFINITY:
```suggestion
		let confirmedBalance: BtcWalletBalance['confirmed'] | null;
```

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
```diff
@@ -224,7 +308,9 @@ export class BtcWalletScheduler implements Scheduler<PostMessageDataRequestBtc>
 	}) => {
 		const newBalance =
 			isNullish(this.store.balance) ||
-			this.store.balance.data !== balance.data ||
+			// Compare balance data properly using JSON comparison for structured balance
+			JSON.stringify(this.store.balance.data, jsonReplacer) !==
```
DenysKarmazynDFINITY: To be honest, I would prefer comparing objects field-by-field instead of the string approach. If you are sure that JSON.stringify is sufficient in this particular case, we can proceed with that.

DenysKarmazynDFINITY (approved):
A few comments, otherwise LGTM.

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DecentAgeCoder: Introduce equals function to compare two instances of BtcWalletBalance

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DecentAgeCoder: Done

src/frontend/src/btc/schedulers/btc-wallet.scheduler.ts
DecentAgeCoder: There is no corresponding mapper available. I have renamed it and moved it to network.utils.ts

PR #8762, opened 2025-09-09
refactor(frontend): Nft list

daviddecentage:
# Motivation

Refactoring Nft List component so it can easily be reused.

# Changes

- Adjusted list component to take list item prop as we do for ModalTokenList
- Adjusted isSpam and isHidden props of NftCard to make it more readable

# Tests
Adjusted test by passing correct parametrized snippets via a testhost component


roman-nazaruk (approved):
LGTM, thnx

PR #8763, opened 2025-09-09
refactor(frontend): Nft card selectable

daviddecentage:
# Motivation

# Changes

# Tests


src/frontend/src/lib/components/nfts/NftCard.svelte
```diff
@@ -15,9 +15,21 @@
...
+		selectable?: boolean;
+		onSelect?: (nft: Nft) => void;
 	}
 
-	let { nft, testId, disabled, isHidden, isSpam }: Props = $props();
+	let { nft, testId, disabled, isHidden, isSpam, selectable, onSelect }: Props = $props();
+
+	const onClick = (e: Event) => {
+		if (disabled || selectable) {
+			e.preventDefault();
```
AntonioVentilii: you can use the util `onclick={preventDefault(onClick)}` from gix-cmp

src/frontend/src/lib/components/nfts/NftImageConsent.svelte
```diff
@@ -54,23 +54,25 @@
...
+					class:lg:max-h-0={type === 'card'}
+					class:lg:opacity-0={type === 'card'}
 				>
-					{$i18n.nfts.text.review_button}
-				</Button>
-			</span>
+					<Button
+						colorStyle="secondary-light"
+						onclick={(e) => {
+							e.preventDefault();
```
AntonioVentilii: same here

src/frontend/src/lib/components/nfts/NftCard.svelte
```diff
@@ -26,14 +38,16 @@
 	class:hover:-translate-y-1={!disabled}
 	class:hover:bg-primary={!disabled}
 	data-tid={testId}
-	href={`${AppPath.Nfts}${nft.collection.network.name}-${nft.collection.address}/${nft.id}`}
-	onclick={disabled ? (e) => e.preventDefault() : undefined}
+	href={selectable && nonNullish(onSelect)
+		? '#'
```
AntonioVentilii: ah wait, now i understand: better to find another way. We prefer not to use `<a>` elements for non-links.

Theoretically, this should be a button (or another selectable element), it is incorrect that it is an `<a>` element

daviddecentage: adjusted it to be a button

src/frontend/src/lib/components/nfts/NftImageConsent.svelte
daviddecentage: not needed anymore now thats a  button

src/frontend/src/lib/components/nfts/NftCard.svelte
daviddecentage: not needed anymore now thats a  button

PR #8764, opened 2025-09-09
refactor(frontend): Button with modal

daviddecentage:
# Motivation

In order to make certain modals reusable, we adjust the sendButtonWithModal a little.

# Changes

- Renamed and relocated to be more generic so we can reuse it for Receive modal etc
- Upgraded to svelte 5
- Adjusted occurrences


src/frontend/src/lib/components/send/Send.svelte
```diff
@@ -1,12 +1,22 @@
...
 
-<SendButtonWithModal isOpen={$modalSend} open={modalStore.openSend}>
-	<SendModal slot="modal" {isTransactionsPage} on:nnsClose />
-</SendButtonWithModal>
+<ButtonWithModal isOpen={$modalSend} onOpen={modalStore.openSend}>
+	{#snippet button(onclick)}
+		<SendButton {onclick} />
+	{/snippet}
+	{#snippet modal()}
+		<SendModal {isTransactionsPage} on:nnsClose />
```
AntonioVentilii: in Svelte 5 we don't have the events anymore, this should become `onClose` in `Props`

src/frontend/src/lib/components/send/Send.svelte
```diff
@@ -1,12 +1,22 @@
...
+
+	const { isTransactionsPage }: Props = $props();
 </script>
 
-<SendButtonWithModal isOpen={$modalSend} open={modalStore.openSend}>
-	<SendModal slot="modal" {isTransactionsPage} on:nnsClose />
-</SendButtonWithModal>
+<ButtonWithModal isOpen={$modalSend} onOpen={modalStore.openSend}>
+	{#snippet button(onclick)}
+		<SendButton {onclick} />
```
AntonioVentilii, line 17: why do you need to pass it twice? doesn't the snippet already open the modal? may you please check that the `onclick` is executed only once? an additional test would be nice too

daviddecentage: Im not calling it again, its just passing the onclick from the ButtonWithModal component to the button snippet as a param, as outside we cannot access the onclick function which is logic that lives inside the ButtonWithModal.

src/frontend/src/lib/components/send/Send.svelte
daviddecentage: I removed the dispatch completely as its unused

src/frontend/src/lib/components/send/Send.svelte
AntonioVentilii: Thank you for the exaplanation

PR #8766, opened 2025-09-09
test(frontend): Simplify tests for IC wallet workers

AntonioVentilii:
# Motivation

We are doing more and more specific tests for the ICRC wallet workers. So, for manageability and to simplify the code we refactor the tests to avoid using functions to call them: they are not so generic anymore.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #5457, opened 2025-03-26
feat(frontend): service to check if an Index canister is sleepy or awake

AntonioVentilii:
# Motivation

We had a few cases where the Index canister is a particular state: it returns zero balance and no transactions, even if the ledger canister shows a balance. It may happen that if it is low in cycles, it goes in "power-saving" mode and in a "sleepy" state, where it does not update its state.

In these cases, however, it does not raise any error, nor we have any flag that point it out.

We need a workaround to know if we need to use only the Ledger canister and ignore the Index canister when the latter is "sleepy". One way of doing it is to compare the balances: if, after a certain amount of time (enough to run a syncing), the balance of the Index canister is zero (and no transaction), BUT the Ledger canister balance is not zero, then we have this particular state.

# Changes

- Create a service isIndexCanisterAwake to check if an Index canister is not updating its state (`sleepy` or `awake`):
  - if balance is not zero or if we have transactions, then it is `awake`;
  - if balance is zero BUT Ledger canister balance is zero too, than we can suppose it is `awake` (in any case it is ok);
  - if after 5 seconds (enough to sync), the balance is not different from zero or there are still no transactions, then it is sleepy (because the Ledger canister balance was not zero, so, something must change).

# Tests

Included tests.

# Future Improvements

A better check would be to compare the number of blocks synced by the Index canister and check how many there are in the Ledger canister. Given an acceptable range, if the two values diverge too much, we can suppose that the Index canister is "sleepy". We are working on it, awaiting for the endpoints to be open in `ic-js` with the next release.

PR #8767, opened 2025-09-09
feat(frontend): Raise an error when the Index canister is sleepy

AntonioVentilii:
# Motivation

We had a few cases where the Index canister is in a particular state: it returns zero balance and no transactions, even if the ledger canister shows a balance. It may happen that if it is low in cycles, it goes in "power-saving" mode and in a "sleepy" state, where it does not update its state.

In these cases, however, it does not raise any error, nor we have any flag that point it out.

With PR #5457, we create a workaround service to verify the status of the Index canister, using the synced blocks.

In this PR, we use such a service, raising an error when the Index canister is "sleepy". It is ok to raise it, since it will be treated as any other Index canister error, falling back to using only the Ledger canister.

# Changes

- For ICRC, when fetching the balance and transactions, if the index balance is different from the ledger balance, we check if the index canister is awake.
- If it is not awake, we raise an error (handled elsewhere).

# Tests

Created tests.


BonomoAlessandro (approved):
LGTM

PR #8771, opened 2025-09-10
feat(frontend): displays collection banner if exists

BonomoAlessandro:
# Motivation

When loading the nft and collection data with alchemy, there is a chance that we also get a collection banner. If so, we want to display that banner instead of the first nft of the collection.

# Changes

- displays collection banner

# Tests

**New:**

CollectionHero:
image

Consent:
image


src/frontend/src/lib/components/nfts/NftCollectionHero.svelte
```diff
@@ -29,12 +29,14 @@
 	const { token, nfts }: Props = $props();
 
 	const breadcrumbItems = $derived([{ label: $i18n.navigation.text.tokens, url: AppPath.Nfts }]);
+
+	const bannerUrl = $derived(nfts?.[0].collection.bannerImageUrl);
```
AntonioVentilii: what happens if the list is empty? isn't better to split it into a derived first?

`const firstNft = $derived(nfts[0])`

`const bannerUrl = $derived(nonNullish(firstNft) ? ....);`

Plus, it seems that `nfts` is never nullish according to `Props`

AntonioVentilii (commented):
Let's split this PR for the two components

src/frontend/src/lib/components/nfts/NftCollectionHero.svelte
BonomoAlessandro: I made the changes, and yes, `nfts` can never be undefined.

AntonioVentilii (approved):

PR #8785, opened 2025-09-10
feat(frontend): Send NFT list

daviddecentage:
# Motivation

We need a list of NFTs to select from on the send modal.

# Changes

Added component
Added texts

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/components/send/SendNftsList.svelte
+++ b/src/frontend/src/lib/components/send/SendNftsList.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { IconExpandMore } from '@dfinity/gix-components';
 	import { nonNullish, notEmptyString } from '@dfinity/utils';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/send/SendNftsList.svelte
```diff
@@ -0,0 +1,78 @@
...
+	import { isDesktop } from '$lib/utils/device.utils';
+	import { findNftsByNetwork } from '$lib/utils/nfts.utils';
+
+	interface Props {
+		onSelect: (nft: Nft) => void;
+	}
+
+	let { onSelect }: Props = $props();
+
+	const dispatch = createEventDispatcher();
```
BonomoAlessandro: `createEventDispatcher` is depricated and should not be used.

daviddecentage: updated

PR #8792, opened 2025-09-10
feat(frontend): Nft services send

daviddecentage:
# Motivation

We add a central nft send function that can be called for any network (only ETH for now)

# Changes

Added service

# Tests

Added tests


src/frontend/src/lib/services/nft.services.ts
```diff
@@ -69,3 +76,60 @@ export const loadNftsByNetwork = async ({
...
+	fromAddress: string;
+	identity: OptionIdentity;
+	gas: bigint;
+	maxFeePerGas: bigint;
+	maxPriorityFeePerGas: bigint;
+	progress?: (step: ProgressStepsSend) => void;
+}) => {
+	if (isNullish(identity)) {
+		await nullishSignOut();
+	} else {
```
BonomoAlessandro: You could also do something like this:

`if (isNullish(identity)) {
  await nullishSignOut();
  return;
}
`

Like this you don't need the `else` part anymore.

BonomoAlessandro (approved):
LGTM

src/frontend/src/lib/services/nft.services.ts
daviddecentage: updated

PR #8705, opened 2025-09-05
feat(frontend): Enable Nft flow for Send modal

daviddecentage:
# Motivation

This enables the send NFT flow for the send modal.

# Changes

- Added new props and flags for NFT flow
- Added function that sets the selected Nft on the select NFT screen

# Tests

Manually tested and verified on fe3:
- Manual token send
- NFT send

PR #8795, opened 2025-09-10
feat(frontend): Send button NFT page

daviddecentage:
# Motivation

Adding the Send Modal with button to the Nft Hero.

# Changes

- Add send button with modal to new NftActionButtons component and added it to the NFT hero
- Extend send modal to take nft prop and preselect it when set

# Tests

Extended tests to ensure the send modal correctly opens on clicking the send button
Added new tests for NftActionButtons

### DIFF
--- a/src/frontend/src/env/caip10-chains.env.ts
+++ b/src/frontend/src/env/caip10-chains.env.ts
@@ -21,3 +21,5 @@ export const CAIP10_CHAINS: Record<
 		}
 	})
 };
+
+export const CAIP10_CHAINS_KEYS = Object.keys(CAIP10_CHAINS);
--- a/src/frontend/src/lib/providers/wallet-connect.providers.ts
+++ b/src/frontend/src/lib/providers/wallet-connect.providers.ts
@@ -1,3 +1,4 @@
+import { CAIP10_CHAINS_KEYS } from '$env/caip10-chains.env';
 import { EIP155_CHAINS_KEYS } from '$env/eip155-chains.env';
 import { SOLANA_MAINNET_NETWORK } from '$env/networks/networks.sol.env';
 import {
@@ -115,7 +116,7 @@ export const initWalletConnect = async ({
 				...(nonNullish(solAddress)
 					? {
 							solana: {
-								chains: [solMainnetNamespace],
+								chains: CAIP10_CHAINS_KEYS,
 								methods: [
 									SESSION_REQUEST_SOL_SIGN_TRANSACTION,
 									SESSION_REQUEST_SOL_SIGN_AND_SEND_TRANSACTION


src/frontend/src/lib/components/send/SendModal.svelte
```diff
@@ -61,6 +62,7 @@
 
 	export let isTransactionsPage: boolean;
 	export let isNftsPage: boolean;
+	export let nft: Nft | undefined = undefined;
```
AntonioVentilii, line 65:
```suggestion
	export let nft?: Nft;
```

src/frontend/src/lib/components/nfts/NftHero.svelte
```diff
@@ -92,6 +93,10 @@
 				<h1 class="flex-1 truncate">
 					{normalizedNftName}
 				</h1>
+
+				{#if nonNullish(nft)}
+					<NftActionButtons {nft} />
+				{/if}
```
AntonioVentilii, line 99: at this point `nft` is already `nonNullish`, no? I don't mean type-wise, but as logic. If yes, the upper condition shall be changed to confirm that no?

src/frontend/src/lib/components/send/SendModal.svelte
daviddecentage: that doesnt work its still a svelte 4 component

src/frontend/src/lib/components/send/SendModal.svelte
```diff
@@ -182,6 +184,12 @@
 			}
 		});
 	};
+
+	onMount(() => {
+		if (nonNullish(nft) && isNftsPage) {
+			selectNft(nft);
+		}
+	});
```
AntonioVentilii, line 192: this should go in another PR? and BTW, is this correct? We were not calling anything onMount of the modal, so we should not change that logic. Maybe there is a context to be initialized and/or modified?

daviddecentage: it belongs to opening the modal with a preselected nft

src/frontend/src/lib/components/nfts/NftHero.svelte
daviddecentage: No, the Nft can be not loaded if we refresh the page

src/frontend/src/lib/components/send/SendModal.svelte
AntonioVentilii: So, it should have not been done in this PR, and the logic is not correct. You should use the context to select the token. May you please fix it in another PR?

AntonioVentilii: In fact, I am trying to remove it, and I found out that you are doing this (PR #8705):

```ts
	const selectNft = (nft: Nft) => {
		selectedNft = nft;
		loadTokenAndRun({
			token: findNonFungibleToken({
				tokens: $nonFungibleTokens,
				networkId: nft.collection.network.id,
				address: nft.collection.address
			}) as Token,
			callback: async () => {
				await goToStep(WizardStepsSend.DESTINATION);
			}
		});
	};
```

This is not correct, since the function `loadTokenAndRun` sets the `token` store (deprecated) that affects quite a lot of derived.

It is required that whatever is set there is not cast. Is it safe to save a `NonFungibleToken` into a `Token` type? Conceptually it is wrong, but since we are already here, may you please check that everything works as intended? And I suggest to really try everything, or even better find a way to deprecate the store finally.

Furthermore, the `SendModal` component should not receive NFT objects as input, since it does not do that for fungible tokens either. So, we either change it entirely to receive both, or we adapt to what the fungible tokens are doing.
Typically, once the `SendModal` opens and we want to give a certain context to it (like the NFT), we do it through the Send context.

PR #8801, opened 2025-09-11
feat(frontend): Nft fee context

daviddecentage:
# Motivation

We extend the EthFeeContext for the new NFT send feature.

# Changes

Added logic for fee calculation

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/derived/ai-assistant.derived.ts
+++ b/src/frontend/src/lib/derived/ai-assistant.derived.ts
@@ -45,9 +45,9 @@ export const aiAssistantSystemMessage: Readable<chat_message_v1> = derived(
 export const aiAssistantLlmMessages: Readable<chat_message_v1[]> = derived(
 	[aiAssistantStore, aiAssistantSystemMessage],
 	([$aiAssistantStore, $aiAssistantSystemMessage]) => {
-		// Get last 100 messages from chat history
+		// Get the last 100 messages from chat history and deduct 1 slot reserved for the system message
 		const recentHistory = ($aiAssistantStore?.chatHistory ?? []).slice(
-			-MAX_SUPPORTED_AI_ASSISTANT_CHAT_LENGTH
+			-MAX_SUPPORTED_AI_ASSISTANT_CHAT_LENGTH + 1
 		);
 
 		return [


src/frontend/src/tests/eth/components/fee/EthFeeContextProps.ts
```diff
@@ -0,0 +1,18 @@
+import type { EthereumNetwork } from '$eth/types/network';
+import type { Network } from '$lib/types/network';
+import type { Nft } from '$lib/types/nft';
+import type { OptionAmount } from '$lib/types/send';
+import type { Token, TokenId } from '$lib/types/token';
+
+export interface EthFeeContextProps {
```
AntonioVentilii: this should not be in the test folder, right? and even in another folder should be named otherwise, since it is a type module

daviddecentage: It is only used by the tests though, thats why I put it there. It cannot be exported from the svelte test host so I had to move it to its own file to use it in the Test host component and in the Test itself

AntonioVentilii: i think you don't need to create a specific interface if it used only by the tests. You can have this directly in the test components.

But, BTW< I would not even create a test component, I would just mock the context directly in the main component: we do this in other tests too

daviddecentage: I cannot mock the context when Im trying to test exactly that?

Also, as I said, I cannot put the interface into the component as it fails importing an interface from a svelte component - not possible

AntonioVentilii: It seems quite a workaround and usually workaround are not good, you know what I mean?

About the context, I don't understand what you mean: you are creating the context and passing it to the component. Why don't you do it like in the other components? Search for `mockContext` in the test folder. Is it somehow applicable?

For example, we do

```ts
		const { getByTestId } = render(BtcConvertForm, {
			props,
			context: mockContext({ utxosFeeStore: store })
		});
```

daviddecentage: Got it, I refactored it, I forgot you can pass context to the render. Now I was able to remove the testhost and the type

src/frontend/src/eth/components/fee/EthFeeContext.svelte
```diff
@@ -125,6 +131,33 @@
 				return;
 			}
 
+			if (nonNullish(sendNft)) {
+				const { to, data } =
+					sendNft.collection.standard === 'erc721'
```
AntonioVentilii: i think there is an util for this (at least there is for `isCollectionErc1155`)

src/frontend/src/eth/components/fee/EthFeeContext.svelte
```diff
@@ -125,6 +131,33 @@
...
+						? encodeErc721SafeTransfer({
+								contractAddress: sendNft.collection.address,
+								from: $ethAddress,
+								to: destination,
+								tokenId: sendNft.id
+							})
+						: encodeErc1155SafeTransfer({
```
AntonioVentilii: if this is the case for ERC1155 let's treat it as another IF, so that it is not fallback for all nft

src/frontend/src/tests/eth/components/fee/EthFeeContext.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+	type EthFeeStore,
+	type FeeStoreData
+} from '$eth/stores/eth-fee.store';
+import type { EthereumNetwork } from '$eth/types/network';
+import type { Network } from '$lib/types/network';
+import type { Nft } from '$lib/types/nft';
+import type { OptionAmount } from '$lib/types/send';
+import type { Token, TokenId } from '$lib/types/token';
+import { mockValidErc721Token } from '$tests/mocks/erc721-tokens.mock';
+import { mockValidErc721Nft } from '$tests/mocks/nfts.mock';
```
AntonioVentilii:
```suggestion
import { render } from '@testing-library/svelte';
import { readable, writable, type Writable } from 'svelte/store';
import * as ckethStoreMod from '$icp-eth/stores/cketh.store';
import * as addressDerived from '$lib/derived/address.derived';
import * as infuraMod from '$eth/providers/infura.providers';
import * as infuraGasRestMod from '$eth/rest/infura.rest';
import * as listenerServices from '$eth/services/eth-listener.services';
import * as feeServices from '$eth/services/fee.services';
import * as ethUtils from '$eth/utils/eth.utils';
import * as tokenUtils from '$eth/utils/token.utils';
import * as evmNativeUtils from '$evm/utils/native-token.utils';
import * as networkUtils from '$lib/utils/network.utils';
import { ETHEREUM_NETWORK } from '$env/networks/networks.eth.env';
import { ETHEREUM_TOKEN } from '$env/tokens/tokens.eth.env';
import EthFeeContext from '$eth/components/fee/EthFeeContext.svelte';
import * as nftSend from '$eth/services/nft-send.services';
import {
	ETH_FEE_CONTEXT_KEY,
	type EthFeeStore,
	type FeeStoreData
} from '$eth/stores/eth-fee.store';
import type { EthereumNetwork } from '$eth/types/network';
import type { Network } from '$lib/types/network';
import type { Nft } from '$lib/types/nft';
import type { OptionAmount } from '$lib/types/send';
import type { Token, TokenId } from '$lib/types/token';
import { mockValidErc721Token } from '$tests/mocks/erc721-tokens.mock';
import { mockValidErc721Nft } from '$tests/mocks/nfts.mock';
```

src/frontend/src/tests/eth/components/fee/EthFeeContext.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+	destination: string;
+	amount: OptionAmount;
+	data: string | undefined;
+	sourceNetwork: EthereumNetwork;
+	targetNetwork: Network | undefined;
+	nativeEthereumToken: Token;
+	sendToken: Token;
+	sendTokenId: TokenId;
+	sendNft: Nft | undefined;
+}
```
AntonioVentilii: let's remove this; If it must exists, it should go into the `types` folder in `src/frontend/src` and used in the original component too (another PR for the creation of the type if it is really required)

src/frontend/src/eth/components/fee/EthFeeContext.svelte
daviddecentage: Updated

src/frontend/src/eth/components/fee/EthFeeContext.svelte
daviddecentage: Updated

src/frontend/src/tests/eth/components/fee/EthFeeContext.spec.ts
daviddecentage: I removed it

src/frontend/src/eth/components/fee/EthFeeContext.svelte
```diff
@@ -125,6 +133,36 @@
...
+						? encodeErc1155SafeTransfer({
+								contractAddress: sendNft.collection.address,
+								from: $ethAddress,
+								to: destination,
+								tokenId: sendNft.id,
+								amount: 1n,
+								data: '0x'
+							})
+						: (() => {
+								throw new Error('Could not calculate fee: Unsupported collection type');
```
AntonioVentilii: maybe this should be localized

src/frontend/src/tests/eth/components/fee/EthFeeContext.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+		destination: string;
+		amount: OptionAmount;
+		data: string | undefined;
+		sourceNetwork: EthereumNetwork;
+		targetNetwork: Network | undefined;
+		nativeEthereumToken: Token;
+		sendToken: Token;
+		sendTokenId: TokenId;
+		sendNft: Nft | undefined;
+	} = {
```
AntonioVentilii, line 58: do you need this type specification? what is the type-error that you receive if you remove it?

daviddecentage: The data is initialized with undefined, so it infers undefined type for a lot of the optional stuff

src/frontend/src/eth/components/fee/EthFeeContext.svelte
daviddecentage: updated

src/frontend/src/tests/eth/components/fee/EthFeeContext.spec.ts
AntonioVentilii: ah! understood! tks for the explanation

BTW I am not against creating a specific type for the original component and use it there and in the tests (I think we do it for some other components too)

PR #8807, opened 2025-09-11
feat(frontend): util for parsing AI contacts tool

DenysKarmazynDFINITY:
# Motivation

Before we can introduce a new tool for filtering contacts, we need to implement a util for parsing it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
 		"build": "tsc --noEmit && vite build && npm run build:post-process",
 		"prepare": "svelte-kit sync",
 		"preview": "vite preview",
-		"check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json",
+		"check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --fail-on-warnings",
 		"check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
 		"test": "tsc --project tsconfig.spec.json --noEmit && vitest",
 		"test:coverage": "tsc --project tsconfig.spec.json --noEmit && vitest run --coverage",


src/frontend/src/lib/utils/ai-assistant.utils.ts
```diff
@@ -45,6 +46,51 @@ export const parseFromAiAssistantContacts = ({
...
+	const { addressIdsFilter } = filterParams.reduce<{
+		addressIdsFilter: string;
+	}>(
+		(acc, { value, name }) => ({
+			addressIdsFilter: name === 'addressIds' ? value : acc.addressIdsFilter
+		}),
+		{
+			addressIdsFilter: '[]'
+		}
+	);
```
AntonioVentilii: nitpick: this can be done with a `filter`, no? and fallback to `'[]'` in case of not find value

`const { addressIdsFilter } = filterParams.filter(({name}) => name === 'addressIds')[0]?.name ?? '[]'`

does this work?

AntonioVentilii: or if you need the last one always 

`const { addressIdsFilter } = filterParams.filter(({name}) => name === 'addressIds').at(-1)?.name ?? '[]'`

or something similar?

AntonioVentilii: i mean, either `find` or `filter`. Maybe `find` is better if you accept the first one found

DenysKarmazynDFINITY: Good idea - changed to `const addressIdsFilter = filterParams.find(({ name }) => name === 'addressIds')?.value ?? '[]';`

PR #8841, opened 2025-09-12
refactor(frontend): Nft heroes

daviddecentage:
# Motivation

Refactoring the Nft heroes some more.

# Changes

Added token ID and collection name to NftHero
Added texts

# Tests

image

image


src/frontend/src/lib/components/nfts/NftImageConsentPreference.svelte
```diff
@@ -39,25 +36,10 @@
...
-			{#snippet icon()}
-				<IconPenLine />
-			{/snippet}
-		</ButtonIcon>
-	</Responsive>
+	<Button
+		colorStyle="secondary-light"
+		onclick={openConsentModal}
+		styleClass="w-auto p-0 grow-0 text-nowrap hover:bg-inherit ml-2"
+		transparent>{$i18n.nfts.text.review_preference}</Button
```
AntonioVentilii: BTW, here we could add the `ariaLabel`, ideally, for `a11y`

daviddecentage: updated

src/frontend/src/lib/components/nfts/NftMetadataList.svelte
```diff
@@ -51,6 +51,18 @@
 			</span>
 		{/if}
 	</ListItem>
+	{#if nonNullish(nft)}
+		<ListItem>
+			<span class="text-tertiary">{$i18n.nfts.text.token_id}</span>
+			{#if nonNullish(nft?.id)}
+				{nft?.id}
```
AntonioVentilii: at this point the nft is non nullish, right? so you don't even need the skeleton. Or did i misread? Can the ID be nullish?

```suggestion
			{#if nonNullish(nft.id)}
				{nft.id}
```

daviddecentage: yes updated

PR #8842, opened 2025-09-12
feat(frontend): Use Svelte `page` store to derive the NFT page values

AntonioVentilii:
# Motivation

Similar to what we do with fungible tokens, we can use the Svelte `page` store to fetch the params of the page and use it as derived.

# Changes

- Create route derived to split NFT-related values: network, collection and NFT IDs.
- Create page derived to match the route parameters with the current NFT store.
- Use the new derived in the code.

# Tests

Added tests.

### DIFF
--- a/src/frontend/src/icp/workers/icrc-wallet.worker.ts
+++ b/src/frontend/src/icp/workers/icrc-wallet.worker.ts
@@ -20,7 +20,6 @@ import type {
 	PostMessageDataRequestIcrc,
 	PostMessageDataRequestIcrcStrict
 } from '$lib/types/post-message';
-import { emit } from '$lib/utils/events.utils';
 import type {
 	IcrcIndexNgGetTransactions,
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/derived/page-nft.derived.ts
```diff
@@ -0,0 +1,44 @@
...
+);
+
+export const pageNft: Readable<Nft | undefined> = derived(
+	[pageCollectionNfts, routeNft],
+	([$pageCollectionNfts, $routeNft]) => {
+		if (isNullish($pageCollectionNfts) || $pageCollectionNfts.length === 0) {
+			return;
+		}
+
+		// Check if route NFT is a possible number
```
AntonioVentilii:
```suggestion
```

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8844, opened 2025-09-13
feat(frontend): Remove emitted event from ICRC wallet workers

AntonioVentilii:
# Motivation

Emitting a DOM event does not really work in the ICRC wallet workers, since they are not in the document. We will move it elsewhere.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8859, opened 2025-09-15
feat(frontend): Support Solana legacy Genesis Hash in WalletConnect

AntonioVentilii:
# Motivation

Some DEXs still use the legacy genesis hash of Solana networks for WalletConnect (for example https://app.hubbleprotocol.io/).

So, we want to support it, to avoid disruptions for the users.

# Note

The official documentation of Solana states that the genesis hash can be directly found via RPC method [getGenesisHash](https://www.quicknode.com/docs/solana/getGenesisHash).

Plus it seems it changed before 22 April 2020: the oldest commit I found is [here](https://github.com/solana-labs/solana/commit/5b8d963ee26700521bd954807fd60a49def0c2cf#diff-fe856881d95ff6c4da2db925ff9bdb728008595800f9095c7a9cef8da9eb9f59R116), but I didn't dig further.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #8862, opened 2025-09-15
chore(frontend): Temporarily disable Japanese support

StefanBerger-DFINITY:
# Motivation

Akeel wants to enable Polish first, and then Japanese in the next release.

# Changes

- Remove Japanese from the languages options

# Tests


src/frontend/src/lib/enums/languages.ts
```diff
@@ -5,7 +5,6 @@ export enum Languages {
 	FRENCH = 'fr',
 	HINDI = 'hi',
 	ITALIAN = 'it',
```
roman-nazaruk, line 7: I suggest commenting this variable for now and adding a TODO: uncomment when needed

@AntonioVentilii WDYT?

roman-nazaruk: // TODO: ....
// JAPANESE = 'ja',

StefanBerger-DFINITY: does not work. if not defined, then the store will fail.
And if we comment it out in env/i18n, then a validation fails.
Seemed fastest to remove it for now (I can add it back in 5 mins).

roman-nazaruk: I see. It probably has dependencies. Approved

PR #8863, opened 2025-09-15
feat(frontend): Adapt custom OISY events to Svelte 5

AntonioVentilii:
# Motivation

In Svelte 5, the events callbacks are called with `onevent` instead of `on:event`. So we adapt our custom events.


src/frontend/src/app.d.ts
```diff
@@ -19,20 +19,20 @@ declare const VITE_GIT_BRANCH_NAME: string;
...
-		'on:oisyCkEthMinterInfoStatus'?: (event: CustomEvent<any>) => void;
-		'on:oisyCkEthereumPendingTransactions'?: (event: CustomEvent<any>) => void;
-		'on:oisyIndexCanisterBalanceOutOfSync'?: (event: CustomEvent<any>) => void;
-		'on:oisyValidateAddresses'?: (event: CustomEvent<any>) => void;
-		'on:oisyRefreshUserProfile'?: (event: CustomEvent<any>) => void;
-		'on:oisyToggleZeroBalances'?: (event: CustomEvent<any>) => void;
-		'on:oisyRefreshContacts'?: (event: CustomEvent<any>) => void;
-		'on:oisyToggleShowHidden'?: (event: CustomEvent<any>) => void;
-		'on:oisyToggleShowSpam'?: (event: CustomEvent<any>) => void;
+		onoisyTriggerWallet?: (event: CustomEvent<any>) => void;
```
DenysKarmazynDFINITY, line 22: For consistency, I would personally prefer to follow the camel case - `onOisy...`. However, I understand we want to keep it close to the Svelte events naming convention, so tough choice.

DenysKarmazynDFINITY (approved):

src/frontend/src/app.d.ts
AntonioVentilii: @DenysKarmazynDFINITY my doubt was on something like this

`emit({ message: 'oisyRefreshUserProfile' });`

that would become

`emit({ message: 'OisyRefreshUserProfile' });`

(same reason why svelte left `on:click` as `onclick`)

DenysKarmazynDFINITY: ok, good point, let's keep it as it is now then

PR #8868, opened 2025-09-15
feat(frontend): add noBorder prop to the List context

DenysKarmazynDFINITY:
# Motivation

To reuse the List component/context in the AI console, we need to extend it with an additional prop - `noBorder`.


roman-nazaruk (approved):
LGTM

PR #8876, opened 2025-09-15
feat(frontend): Make `EthFeeStoreContext` accept even undefined tokens

AntonioVentilii:
# Motivation

We are going to remove the fallback form the derived store for the native Ethereum tokens. So, we need to manage possible nullish values in the consumers. In this specific case we modify `EthFeeStoreContext`.

### DIFF
--- a/src/frontend/src/eth/components/fee/EthFeeStoreContext.svelte
+++ b/src/frontend/src/eth/components/fee/EthFeeStoreContext.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+	import { isNullish } from '@dfinity/utils';
 	import { setContext, type Snippet } from 'svelte';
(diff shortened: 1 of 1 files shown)


src/frontend/src/eth/components/fee/EthFeeStoreContext.svelte
```diff
@@ -54,6 +59,10 @@
 			return;
 		}
 
+		if (isNullish(token)) {
```
DenysKarmazynDFINITY: maybe a bit shorter version?
```
if (!isEthNetwork || isNullish(token)) {
  return;
}
```

AntonioVentilii: nice!

PR #8877, opened 2025-09-15
feat(frontend): updates RewardDateBadge and its position

BonomoAlessandro:
# Motivation

# Changes

# Tests


src/frontend/src/lib/components/icons/lucide/IconCalendarDays.svelte
```diff
@@ -0,0 +1,70 @@
+<!-- source: ISC Lucide - please visit https://lucide.dev/license -->
+<script lang="ts">
+	interface Props {
+		size?: string;
+	}
```
AntonioVentilii:
```suggestion
	}
	
```

src/frontend/src/lib/components/rewards/RewardDateBadge.svelte
```diff
@@ -15,7 +16,10 @@
 
 <DateBadge
 	{date}
+	inlineStyleClass="flex items-center gap-1.5 text-sm"
 	prefix={hasEnded ? $i18n.rewards.text.ended_date : $i18n.rewards.text.active_date}
+	styleClass="rounded-full py-1 pl-2 pr-3"
```
AntonioVentilii: both `inlineStyleClass` and `styleClass` may probably go directly inside `<DateBadge>`

there are not more usages. Same for the calendar icon

BonomoAlessandro: Hmm i guess it is better like this because a Badge normally looks different.

Badge:
image


When i add the `styleClasse` and the `inlineStyleClass` we are still able to use the `DateBadge` as a "Badge".

AntonioVentilii (approved):

PR #8878, opened 2025-09-15
feat(frontend): added partner for Valora calls

roman-nazaruk:
# Motivation

We need to provide a partner for Velora calls.


AntonioVentilii (changes requested):
Let's wait for the ENV variable to be available. It is coming soon, no?

roman-nazaruk:
> Let's wait for the ENV variable to be available. It is coming soon, no?

Yeah, I am waiting for a feedback

PR #8880, opened 2025-09-15
feat(frontend): updates RewardDateBadge appearance

BonomoAlessandro:
# Motivation

The `RewardDatebadge` got an appearance rework.

# Tests

**New:**

RewardCard:
image

Modal:
image


src/frontend/src/lib/components/rewards/RewardDateBadge.svelte
```diff
@@ -15,7 +16,10 @@
 
 <DateBadge
 	{date}
+	inlineStyleClass="flex items-center gap-1.5 text-sm"
 	prefix={hasEnded ? $i18n.rewards.text.ended_date : $i18n.rewards.text.active_date}
+	styleClass="rounded-full py-1 pl-2 pr-3"
```
BonomoAlessandro: See discussion on [https://github.com/dfinity/oisy-wallet/pull/8877](#8877)

AntonioVentilii: Continuing discussion https://github.com/dfinity/oisy-wallet/pull/8877/files#r2348774499

I still think you can put it directly here, since the component is specific for DataBadge, it is not really just a badge anymore, right?

BonomoAlessandro: But that is why we have the `DateBadge` which is a badge and the `RewardDateBadge` which is an additional abstraction of it.

AntonioVentilii: ah understood! then maybe just the Calendar icon can be sort of a feature in DataBadge? like `showIcon`

BonomoAlessandro: This would mean we cannot use the `DateBadge` with another icon. In Figma there are many other possible `date`, `calendar` icons that possibly can be used.
What i can do is, instead of using `children`, using `icon` as parameter.

AntonioVentilii: wait, so we already have more applications for the `DateBadge`? Will they have the same style and only different icons?

BonomoAlessandro: No, we have a `Badge` component that is used at several points in the application and has a specific look. We also have a `DateBadge` which looks exactly the same as the `Badge` but is able to display dates. And we also do have the `RewardDateBadge` which is another abstraction layer of the `DateBadge` and defines the Logo etc. But this component is only needed for the reward specific part of the application, while the `DateBadge` is a generic component to use everywhere else. Even when it is not used directly anymore.

AntonioVentilii: Understood, tks for the explanation!

If you already plan to use the `DateBadge` with another icon and other styles in the next PRs, then all good; otherwise let's put the calendar icon and the style directly inside `DateBadge` with an optional prop `showIcon`?

AntonioVentilii (approved):

PR #8882, opened 2025-09-15
feat(frontend): added OISY_DOMAIN during build

roman-nazaruk:
# Motivation

We need to derive OISY_DOMAIN during the build and expose it as an environment variable.

# Changes
- In vite.config, get the domain from the current DFX_NETWORK.
- Inject it as an environment variable VITE_OISY_DOMAIN


src/frontend/src/lib/components/swap/SwapModal.svelte
```diff
@@ -7,6 +7,7 @@
 	import SwapTokensList from '$lib/components/swap/SwapTokensList.svelte';
 	import SwapWizard from '$lib/components/swap/SwapWizard.svelte';
 	import { swapWizardSteps } from '$lib/config/swap.config';
+	import { OISY_URL } from '$lib/constants/oisy.constants';
```
AntonioVentilii:
```suggestion
```

PR #8885, opened 2025-09-15
refactor(frontend): Rename derived store for native Ethereum/EVM token

AntonioVentilii:
# Motivation

Just to avoid confusion, we rename the derived store `ethereumToken` to `nativeEthereumToken`.


roman-nazaruk (approved):
LGTM, thnx

PR #8886, opened 2025-09-15
refactor(frontend): Correct type for the native required tokens

AntonioVentilii:
# Motivation

For consistency, we use the type `RequiredToken` for all the native tokens.


roman-nazaruk (approved):
LGTM, thnx

PR #8887, opened 2025-09-15
feat(frontend): Use EVMs too in the native token for Ethereum/EVM networks

AntonioVentilii:
# Motivation

In Derived store `nativeEthereumToken` we need to include the EVMs networks.


src/frontend/src/tests/lib/components/swap/NewSwapWizard.spec.ts
```diff
@@ -5,23 +5,14 @@ import NewSwapWizard from '$lib/components/swap/NewSwapWizard.svelte';
...
-	const actual = await importOriginal();
-	return {
-		...(actual as Record<string, unknown>),
-		isNullish: vi.fn()
-	};
-});
-
-vi.mock('$lib/utils/network.utils', () => ({
-	isNetworkIdICP: vi.fn()
+vi.mock('$lib/services/auth.services', () => ({
```
DenysKarmazynDFINITY, line 14: Wondering if the changes here are related to the derived store update above?

AntonioVentilii: yes, because now we are using utils from these modules, and they are not mocked. And ultimately, there is no need to mock them, since we want them to work as expected

PR #8888, opened 2025-09-15
feat(frontend): update design and logic of ShowContactsToolItem

DenysKarmazynDFINITY:
# Motivation

We want to update the ShowContactsToolItem component's logic - it now should behave similar to what we have in the Address book modal. Similar as with the previous implementation, we do not re-use the existing component but create a separate one since the usecases (and therefore params and props) are quite different.

Screenshot 2025-09-15 at 15 56 43


AntonioVentilii (approved):
Let's split it into design and logic please: is it possible or they go hand-in-hand?

src/frontend/src/lib/components/ai-assistant/AiAssistantShowContactsTool.svelte
```diff
@@ -13,39 +15,29 @@
...
+					{contact}
+					onClick={async (address: ContactAddressUiWithId) =>
+						!loading &&
+						(await onSendMessage({
+							messageText: replacePlaceholders($i18n.ai_assistant.text.send_to_message, {
+								$contact_name: contact.name,
+								$address_info: `${isEmptyString(address.label) ? '' : `${address.label}: `}${shortenWithMiddleEllipsis({ text: address.address })}`
+							}),
+							context: `Send destination information: "selectedContactAddressId" - ${address.id}; "addressType": ${address.addressType}.`
+						}))}
```
AntonioVentilii, line 36: maybe we can move this function in the script at some point

DenysKarmazynDFINITY: yep, agreed - will do it with the next iteration

PR #8893, opened 2025-09-16
chore(frontend): Add Sprinkles s1e5 configuration

StefanBerger-DFINITY:
# Motivation

We will start with Sprinkles s1e5

# Changes

- add configuration and initial english texts.

Notes:
- the texts are not yet final. We will likely will get an update on some texts in a later PR, but we want to have the initial config in, so we can start reviewing them in the app directly.
- We will update the final start date before the release. For now, we keep a day in the past, so all deployments to staging from the main branch keep the episode card active.

# Tests
image


BonomoAlessandro (approved):
LGTM

PR #8894, opened 2025-09-16
feat(frontend): adds EligibilityBadge component

BonomoAlessandro:
# Motivation

We want to have another abstraction of the `Badge` which enables us to meet the reward specific requirements.

# Tests

**New:**

eligible:
image

not eligible:
image


src/frontend/src/lib/components/rewards/EligibilityBadge.svelte
```diff
@@ -0,0 +1,34 @@
...
+</script>
+
+<span class="inline-flex">
+	<Badge
+		styleClass="rounded-full py-1 pl-2 pr-3"
+		testId={nonNullish(testId) ? `${testId}-badge` : undefined}
+		variant={isEligible ? 'eligible' : 'not-eligible'}
+	>
+		{#if isEligible}
+			<div class="flex items-center gap-1.5 text-sm">
```
AntonioVentilii: this div can be put outside the if condition, since it's the same in both cases?

BonomoAlessandro: You are right

AntonioVentilii (approved):
LGTM tks!

PR #8895, opened 2025-09-16
chore(frontend): remove buffer and node polyfill

peterpeterparker:
# Motivation

Thanks to the upgrade to AgentJS v3, polyfying buffer and node should not be required anymore. 

However, it seems that buffer is used by `@solana/web3.js` so this PR needs to be asserted or serves as exploration.

**Update**: `@solana/web3.js` is actually used by `alchemy-sdk` which itself is deprecated. So, when we will have migrated it, there is a good chance that `buffer` won't be referenced anymore - as the solana kit does not seem to use it - and per extension we will be able to remove the buffer and node polyfill.

# Changes

# Tests


AntonioVentilii:
Blocked on the alchemy-sdk / @solana/web3.js migration; will be re-opened (or recreated) once that lands and buffer is no longer referenced.

PR #8898, opened 2025-09-16
fix(frontend): Use correct type for IC token fee context

AntonioVentilii:
# Motivation

In component SwapFees, the type for the IC token fee context was mistakenly using the component instead of the context type.


roman-nazaruk (approved):
Wow, thanks

PR #8900, opened 2025-09-16
feat(frontend): `ethers` instead of `alchemy-sdk` when formatting numbers

AntonioVentilii:
# Motivation

In an effort to deprecate `alchemy-sdk`, we replace the usage of `Utils.formatUnits` of `alchemy-sdk` (that is inherited from the old version of `ethers`, alias `@ethersproject`) with `formatUnits` from `ethers/utils`.

The tests are quite extensive

# Changes

- Replace `Utils.formatUnits` with `formatUnits`.
- The parameter for the units does not accept stringified numbers anymore, so we need to parse it first.

# Tests

The tests for this util are quite extensive, so they should still work as intended.


src/frontend/src/lib/utils/format.utils.ts
```diff
@@ -29,7 +28,14 @@ export const formatToken = ({
 	trailingZeros = false,
 	showPlusSign = false
 }: FormatTokenParams): AmountString => {
-	const res = Utils.formatUnits(value, unitName);
+	const parsedUnitName: BigNumberish =
```
roman-nazaruk, line 31: I am just wondering, maybe we should move it to utils. WDYT?

roman-nazaruk (approved):
LGTM

PR #8901, opened 2025-09-16
feat(frontend): implements RewardNetworkBonus component

BonomoAlessandro:
# Motivation

We want to have a component that is able to display the correct network bonus image.

# Tests

**New:**
image


daviddecentage (approved):
LGTM, thanks!

PR #8903, opened 2025-09-16
refactor(frontend): Send nft modal

daviddecentage:
# Motivation

# Changes

# Tests


src/frontend/src/lib/components/send/SendModal.svelte
```diff
@@ -173,23 +177,21 @@
...
-				networkId: nft.collection.network.id,
-				address: nft.collection.address
-			}) as Token,
-			callback: async () => {
-				await goToStep(WizardStepsSend.DESTINATION);
-			}
+		const token = findNonFungibleToken({
+			tokens: $nonFungibleTokens,
+			networkId: nft.collection.network.id,
+			address: nft.collection.address
```
AntonioVentilii, line 183: isn't this the new derived `pageNonFungibleToken`?

src/frontend/src/routes/(app)/+layout.svelte
```diff
@@ -34,14 +34,14 @@
 	let tokensRoute = $derived(isRouteTokens(page));
 
 	let nftsRoute = $derived(isRouteNfts(page));
-	let nftsCollectionRoute = $derived(isRouteNfts(page) && nonNullish(routeCollection));
+	let nftsCollectionRoute = $derived(isRouteNfts(page) && nonNullish($routeCollection));
```
AntonioVentilii: can you do this change in another PR, since it is a fix?

src/frontend/src/tests/mocks/page.store.mock.ts
```diff
@@ -35,6 +35,17 @@ const initPageStoreMock = () => {
...
+			data,
+			params
+		}: {
+			data: Partial<RouteParams>;
+			params: { [key: string]: string };
+		}) => {
+			set({ params, data });
+			page.params = params;
+			page.data = data;
+		},
```
AntonioVentilii: can you modify `mockDynamicRoutes()` to use optionally data?

or actually, i would use in the tests first `mock()` and then `mockDynamicRoutes()`

daviddecentage: Right I call both now, no need for a new function

src/frontend/src/lib/components/send/SendModal.svelte
daviddecentage: No, that is the method for manually selecting an Nft, which only comes up when accessing the send modal from the root of the Nft pages ($pageNonFungibleToken is not available there)

src/frontend/src/routes/(app)/+layout.svelte
daviddecentage: There we go https://github.com/dfinity/oisy-wallet/pull/8916

src/frontend/src/lib/components/send/SendModal.svelte
AntonioVentilii: right! good point!

src/frontend/src/tests/lib/components/nfts/NftActionButtons.spec.ts
```diff
@@ -1,16 +1,20 @@
...
 describe('NftActionButtons', () => {
 	const sendButtonSelector = `button[data-tid="${NFT_ACTION_SEND}"]`;
 
 	it('should render the send button', async () => {
-		const { container } = render(NftActionButtons, {
-			nft: mockValidErc1155Nft
+		vi.spyOn(pageNft, 'subscribe').mockImplementation((fn) => {
+			fn(mockValidErc1155Nft);
+			return () => {};
 		});
```
AntonioVentilii, line 14: theoretically here you could use the `mockPage` methods no?

daviddecentage: I was not able to figure it out quickly since its a composition of quite a few derived/stores. To save some time I copied what was working for another test. Shouldnt matter too much since we really just wanna test the rendering of the components

PR #8904, opened 2025-09-16
feat(frontend): Adds derived pageNonFungibleToken

daviddecentage:
# Motivation

Adding a derived pageNonFungibleToken to be used in the refactoring of the Nft send flow.

# Changes

Added derived
Adjusted current usage in Nft.svelte to avoid duplicate logic


src/frontend/src/lib/components/nfts/NftBadges.svelte
```diff
@@ -0,0 +1,25 @@
+<script lang="ts">
```
AntonioVentilii: is this file supposed for this PR?

src/frontend/src/lib/derived/page-token.derived.ts
```diff
@@ -96,3 +100,15 @@ export const pageTokenToggleable: Readable<boolean> = derived([pageToken], ([$pa
 
 	return false;
 });
+
+export const pageNonFungibleToken: Readable<NonFungibleToken | undefined> = derived(
+	[pageNft, nonFungibleTokens],
+	([$pageNft, $nonFungibleTokens]) =>
+		nonNullish($pageNft) && nonNullish($pageNft.collection)
+			? findNonFungibleToken({
```
AntonioVentilii, line 108: i know we have tests already for `findNonFungibleToken` util, but some tests for this derived would be useful too, even if repeated

src/frontend/src/lib/components/nfts/NftBadges.svelte
daviddecentage: no sorry, don know how that slipped in

src/frontend/src/lib/derived/page-token.derived.ts
daviddecentage: added some small tests

PR #8908, opened 2025-09-16
feat(frontend): Map Solana ATA instructions

AntonioVentilii:
# Motivation

As suggested in the official repo in https://github.com/solana-program/associated-token-account/issues/135#issuecomment-3284456574, we map the Solana ATA instructions using the existing methods in `@solana-program/token`.


roman-nazaruk (approved):
LGTM

PR #8910, opened 2025-09-16
feat(frontend): provide more info about OISY to LLM via system prompt

DenysKarmazynDFINITY:
# Motivation

We want to teach LLM about OISY so it can support users with some basic questions about the wallet.


roman-nazaruk (approved):
LGTM,thnx

PR #8912, opened 2025-09-16
feat(build): Avoid redirect from www sub-domains

AntonioVentilii:
# Motivation

This issue was flagged:
> The DNS configuration is a bit problematic for the custom domain ([oisy.com](http://oisy.com/)): we have a redirect setup that goes from [www.oisy.com](http://www.oisy.com/) to [oisy.com](http://oisy.com/). This redirect is handled by Cloudflare and requires Cloudflare to obtain certificates. This interferes with the HTTP gateways obtaining certificates for [oisy.com](http://oisy.com/). So, ideally, we can move the redirect to the canister or handle everything by Cloudflare.

Credits to @peterpeterparker 

# Changes

- Add the domain with `www.` sub-domain prefix to `ic-domains` file.

# Tests

Checked deploying it in a test canister and the derivationp ath is the same

# Additional offline TO-DOs

- [ ] Add DNS entries
- [ ] Trigger Boundary Nodes (BN) HTTPS post to register the


AntonioVentilii:
Superseded by:

- https://github.com/dfinity/oisy-wallet/pull/12658

PR #8919, opened 2025-09-16
feat(frontend): Start deprecating `alchemy-sdk` in `AlchemyProvider`

AntonioVentilii:
# Motivation

It was announced that [alchemy-sdk](https://github.com/alchemyplatform/alchemy-sdk-js) will be deprecated soon:

> Warning
> 
> The Alchemy SDK JS has been deprecated and will be archived January 2026. The software will only receive minimal support until then. New and existing users are encourage to use the following tools instead:

In this PR we create a new provider using `ethers` library and use it in parallel in `AlchemyProvider` class for method `getTransaction`.

# Changes

- Create new provider in `AlchemyProvider` using the Alchemy provider class from `ethers` library.
- Substitute the old provider in method `getTransaction`.
- Adapt the subsequent types.

# Tests

Adapted tests.


AntonioVentilii:
We need to wait for `ethers` library to include BSC among the Alchemy networks:

- https://github.com/ethers-io/ethers.js/issues/5040

PR #8928, opened 2025-09-17
feat(frontend): updates rewards canister api

BonomoAlessandro:
# Motivation

We want to use the updated `reward` canister api (0.5.9).

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -112,6 +112,7 @@ A list of useful faucets:
 - Bitcoin: [Coinfaucet](https://coinfaucet.eu/en/btc-testnet/)
 - SOL: [Solana Foundation Faucet](https://faucet.solana.com/) or [Sol Faucet](https://solfaucet.com/)
 - TESTICP: [TESTICP Faucet](https://nqoci-rqaaa-aaaap-qp53q-cai.icp0.io/)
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/utils/rewards.utils.ts
```diff
@@ -109,7 +109,9 @@ export const mapEligibilityReport = (eligibilityReport: EligibilityReport): Camp
 			campaignId,
 			available: eligibility.available,
 			eligible: eligibility.eligible,
-			criteria
+			criteria,
+			probabilityMultiplierEnabled: eligibility.probability_multiplier_enabled,
+			probabilityMultiplier: Number(eligibility.probability_multiplier)
```
AntonioVentilii: isn't this already a `number`?

AntonioVentilii: are you sure? the type says number:

```ts
export interface CampaignEligibility {
	probability_multiplier_enabled: boolean;
	probability_multiplier: number;
	available: boolean;
	eligible: boolean;
	criteria: Array<CriterionEligibility>;
}
```

but in any case, if it's a `bigint`, let's leave it as `bigint`, better for math operations

BonomoAlessandro: You are right. I just removed it.

PR #8931, opened 2025-09-17
refactor(frontend): Media consent preference

daviddecentage:
# Motivation

Refactoring the display preference component for upcoming and already merged changes.

# Changes

- Removed responsive wrapper as lists are always responsive now
- Added ariaLabel
- Changed prop to take collection directly because the media preference is a property of the token, not a specific Nft

# Tests

Adjusted test


AntonioVentilii (approved):
Thank you!

PR #8942, opened 2025-09-17
feat(frontend): adds NetworkBonusImage to RewardCard

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/components/rewards/RewardCard.svelte
+++ b/src/frontend/src/lib/components/rewards/RewardCard.svelte
@@ -13,7 +13,9 @@
 		type RewardEligibilityContext
 	} from '$lib/stores/reward.store';
 	import { replacePlaceholders, resolveText } from '$lib/utils/i18n.utils';
-	import { isEndedCampaign } from '$lib/utils/rewards.utils';
+	import { isEndedCampaign, normalizeNetworkMultiplier } from '$lib/utils/rewards.utils';
+	import EligibilityBadge from '$lib/components/rewards/EligibilityBadge.svelte';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/rewards/RewardCard.svelte
```diff
@@ -29,6 +30,10 @@
 
 	const campaignEligibility = getCampaignEligibility(reward.id);
 	const isEligible = $derived($campaignEligibility?.eligible ?? false);
+	const hasNetworkBonus = $derived($campaignEligibility?.probabilityMultiplierEnabled ?? false);
+	const networkBonusMultiplier = $derived(
+		normalizeNetworkMultiplier($campaignEligibility?.probabilityMultiplier ?? 1)
```
roman-nazaruk: I would move 1 to constants, as maybe in the future we would need to change it. But not a strong opinion. WDYT?

src/frontend/src/lib/components/rewards/RewardCard.svelte
```diff
@@ -29,6 +30,10 @@
 
 	const campaignEligibility = getCampaignEligibility(reward.id);
```
roman-nazaruk, line 32: As it was discussed offline, it should be wrapped with $derived

BonomoAlessandro: I will fix it in antother PR, because i need to fix it at several locations.

src/frontend/src/lib/components/rewards/RewardCard.svelte
BonomoAlessandro: Good point

PR #8946, opened 2025-09-17
feat(frontend): adds NetworkBonusImage to RewardModal

BonomoAlessandro:
# Motivation

# Changes

# Tests

### DIFF
--- a/src/frontend/src/lib/components/rewards/RewardsRequirements.svelte
+++ b/src/frontend/src/lib/components/rewards/RewardsRequirements.svelte
@@ -1,29 +1,40 @@
 <script lang="ts">
 	import RewardRequirement from '$lib/components/rewards/RewardRequirement.svelte';
-	import Badge from '$lib/components/ui/Badge.svelte';
 	import { REWARDS_REQUIREMENTS_STATUS } from '$lib/constants/test-ids.constants';
 	import { i18n } from '$lib/stores/i18n.store';
 	import type { CampaignCriterion } from '$lib/types/reward';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/rewards/RewardsRequirements.svelte
```diff
@@ -1,29 +1,40 @@
...
 	import RewardRequirement from '$lib/components/rewards/RewardRequirement.svelte';
-	import Badge from '$lib/components/ui/Badge.svelte';
 	import { REWARDS_REQUIREMENTS_STATUS } from '$lib/constants/test-ids.constants';
 	import { i18n } from '$lib/stores/i18n.store';
 	import type { CampaignCriterion } from '$lib/types/reward';
 
 	interface Props {
 		isEligible: boolean;
+		hasNetworkBonus: boolean;
+		networkBonusMultiplier: 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8;
```
AntonioVentilii, line 13: these values have no check, since the type (limit of 8) is not checked in the consumer. Shall we accepts any number here and create a limitative logic directly in the component?

BonomoAlessandro: To prevent this problem i implemented the `normalizeNetworkMultiplier` function which makes sure that the correct numbers are provided.

AntonioVentilii: Tks for the explanation

AntonioVentilii (approved):
LGTM tks!

PR #8957, opened 2025-09-17
refactor(frontend): Make util `parseSolComputeBudgetInstruction` more type-safe

AntonioVentilii:
# Motivation

To make util `parseSolComputeBudgetInstruction` more type-safe, we use a forced casting to never to discover if we forgot to track some specific variants.

Since we are here, we add tests.


roman-nazaruk:
As discussed offline, it’s better to keep the switch statement and simply return a type error in the default case, rather than rewriting the whole logic as if statements. Refers to these PRs: 

1. [PR-1](https://github.com/dfinity/oisy-wallet/pull/8957)
2. [PR-2](https://github.com/dfinity/oisy-wallet/pull/8958)
3. [PR-3](https://github.com/dfinity/oisy-wallet/pull/8959)
4. [PR-4](https://github.com/dfinity/oisy-wallet/pull/8964)
5. [PR-5](https://github.com/dfinity/oisy-wallet/pull/8965)

AntonioVentilii:
Thank you @roman-nazaruk! Very good suggestion!

PR #8964, opened 2025-09-17
refactor(frontend): Make util `parseSolTokenInstruction` more type-safe

AntonioVentilii:
# Motivation

To make util `parseSolTokenInstruction` more type-safe, we use a forced casting to never to discover if we forgot to track some specific variants.

Since we are here, we add tests.


roman-nazaruk (approved):
LGTM, thnx

PR #8971, opened 2025-09-17
refactor(frontend): Start deprecating `alchemy-sdk` in `AlchemyProvider`

AntonioVentilii:
# Motivation

It was announced that [alchemy-sdk](https://github.com/alchemyplatform/alchemy-sdk-js) will be deprecated soon:

> Warning
> 
> The Alchemy SDK JS has been deprecated and will be archived January 2026. The software will only receive minimal support until then. New and existing users are encourage to use the following tools instead:

So, we need to start taking care of the migration.

In this first PR we just rename the current provider of class `AlchemyProvider`, so that we can include a new one after and slowly migrate.

### DIFF
--- a/src/frontend/src/eth/providers/alchemy.providers.ts
+++ b/src/frontend/src/eth/providers/alchemy.providers.ts
@@ -118,17 +118,18 @@ export const initPendingTransactionsListener = ({
 };
 
 export class AlchemyProvider {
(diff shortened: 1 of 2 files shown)


roman-nazaruk (approved):
I like the idea. Thanks. LGTM

PR #8973, opened 2025-09-18
chore(frontend): Add Polish Złoty Currency

StefanBerger-DFINITY:
# Motivation

We add Polish as language. We should also support its currency: PLN - Polish Złoty

# Changes

- add pln

# Tests
image


AntonioVentilii (approved):
LGTM tks!!!

PR #8975, opened 2025-09-18
fix(frontend): Add hedge cases to util `parseIcErrorMessage`

AntonioVentilii:
# Motivation

We treat more hedge cases in util `parseIcErrorMessage` and create tests to check.


src/frontend/src/lib/utils/error.utils.ts
```diff
@@ -102,7 +102,7 @@ export const replaceIcErrorFields = (err: unknown): string | undefined =>
 
 const stripHttpDetails = (text: string): string =>
 	text
-		// Remove from "HTTP d'-0tails:" + "{" up to the next closing brace on its own line
+		// Remove from "HTTP details:" + "{" up to the next closing brace on its own line
```
roman-nazaruk, line 105: Wow, thanks

PR #8983, opened 2025-09-18
feat(frontend): add transperant Network logos

CvetelinaAngelova:
# Motivation

Bugfix : AddressType icon rendered instead of the network logo.
Provide circle-free network logos

# Changes

Schema: add optional iconTransparent to NetworkSchema.
Transaction card: render NetworkLogo (transparent variant) instead of AddressType.

# Tests

Manual/visual: verify transparent logos for networks in transaction cards to be the same as transaction popOverv
Screenshot 2025-09-18 at 12 13 14


CvetelinaAngelova:
No we couldn't use the IconAddressType.svelte as it have only the AddressType networks which leads to the bugs like this one https://dfinity.atlassian.net/browse/OISY-2106
The Icon AddressType have only "Btc" | "Eth" | "Sol" | "Icrcv2" - then it leads to the problems like 
Cards always show the Ethereum Icon for all EVM networks

src/frontend/src/lib/components/networks/NetworkLogo.svelte
```diff
@@ -1,27 +1,30 @@
...
 	}
 
-	let { network, size = 'xxs', color = 'off-white', addressType, testId }: Props = $props();
+	let { network, size = 'xxs', color = 'off-white', testId, transparent = false }: Props = $props();
 </script>
 
-{#if nonNullish(addressType)}
-	<div data-tid={`${testId}-transparent`}>
-		<IconAddressType {addressType} size="20" transparent />
+{#if transparent && network.iconTransparent}
```
daviddecentage:
```suggestion
{#if transparent && nonNullish(network.iconTransparent)}
```

daviddecentage (approved):
LGTM, thanks

PR #8984, opened 2025-09-18
fix(frontend): Network settings list

daviddecentage:
# Motivation

There seem to be a few unwanted side effects of a PR from last week https://github.com/dfinity/oisy-wallet/pull/8914.

That PR aimed to make ListItems responsive in general, but it assumes that the content of the list items always consists of one element, which we cannot know.

To solve this issue we revert the change, and also remove the original flex-row from the styled list items (as that is the default anyways). That way, we can override responsive styling rules for list items on demand.

# Changes

- Added new prop itemStyleClass so we can set a styleClass for list items via the single List component (instead of passing it for each list item)
- Adjusted responsive usecase in NftMetadataList

# Tests

Network settings list looks correct on small screens again:
image

Nft Metadata still responsive after applying the styles directly where needed:
image

On desktop nothing has changed:
image
image

### DIFF
--- a/src/frontend/src/eth/components/swap/SwapEthWizard.svelte
+++ b/src/frontend/src/eth/components/swap/SwapEthWizard.svelte
@@ -6,7 +6,7 @@
 	import SwapEthForm from './SwapEthForm.svelte';
 	import EthFeeContext from '$eth/components/fee/EthFeeContext.svelte';
 	import EthFeeDisplay from '$eth/components/fee/EthFeeDisplay.svelte';
-	import { nativeEthereumTokenWithFallback } from '$eth/derived/token.derived';
+	import { enabledEthereumTokens } from '$eth/derived/tokens.derived';
 	import {
 		ETH_FEE_CONTEXT_KEY,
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/nfts/NftMetadataList.svelte
```diff
@@ -38,7 +38,7 @@
 	);
 </script>
 
-<List condensed styleClass="text-sm text-primary">
+<List condensed itemStyleClass="flex-col sm:flex-row" styleClass="text-sm text-primary">
```
DenysKarmazynDFINITY, line 41: Can we just pass `styleClass` to ListItem instead?

daviddecentage: We can, but as described in the PR description, its less verbose if we can pass it for the whole list once instead of for each element seperately. The option to pass it to single elements still exists if we want to do that. But I think its nice to have the option to set it once for each element as we can with `noPadding` `noBorder` etc

PR #8992, opened 2025-09-18
fix(frontend): fix evm eth sade estimate gas value

roman-nazaruk:
# Motivation

We got issue typeError: Cannot convert 0.001 to a BigInt when building the value for Infura’s safeEstimateGas call. 

# Changes

- EthFeeContext: Replace direct BigInt(amount) conversion with parseToken(...)
- WalletConnectSendModal: Add formatToken(...) to sanitize and validate amount


src/frontend/src/eth/components/fee/EthFeeContext.svelte
```diff
@@ -102,11 +102,14 @@
 			if (isSupportedEthTokenId(sendTokenId) || isSupportedEvmNativeTokenId(sendTokenId)) {
 				// We estimate gas only when it is not a ck-conversion (i.e. target network is not ICP).
 				// Otherwise, we would need to emulate the data that are provided to the minter contract address.
+
```
AntonioVentilii:
```suggestion
```

PR #9026, opened 2025-09-22
feat(frontend): updates RewardDateBadge

BonomoAlessandro:
# Motivation

The **RewardDateBadge** should be able to display upcoming, active and ended dates with the corresponding texts.

# Tests

**New:**

image

image

image


daviddecentage (approved):
LGTM

PR #9036, opened 2025-09-22
feat(frontend): use send status from store in SendTokenTool

DenysKarmazynDFINITY:
# Motivation

We need to apply the last changes in order to start using send statuses from store instead of component's state in SendTokenTool.

### DIFF
--- a/src/frontend/src/lib/components/ai-assistant/AiAssistantReviewSendBtcToken.svelte
+++ b/src/frontend/src/lib/components/ai-assistant/AiAssistantReviewSendBtcToken.svelte
@@ -47,9 +47,10 @@
 		destination: Address;
 		sendCompleted: boolean;
 		sendEnabled: boolean;
+		onSendCompleted: () => void;
 	}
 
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/ai-assistant/AiAssistantReviewSendBtcToken.svelte
```diff
@@ -47,9 +47,10 @@
 		destination: Address;
 		sendCompleted: boolean;
 		sendEnabled: boolean;
+		onSendCompleted: () => void;
 	}
 
-	let { amount, destination, sendCompleted = $bindable(), sendEnabled }: Props = $props();
+	let { amount, destination, sendCompleted, onSendCompleted, sendEnabled }: Props = $props();
```
AntonioVentilii, line 53: do you still need `sendComplete` if it is not bindable anymore?

DenysKarmazynDFINITY: yes, it is not bindable because the value now comes from the store, it gets updated when `onSendCompleted` is triggered

PR #9046, opened 2025-09-22
feat(frontend): should only load and import collections of active networks

BonomoAlessandro:
# Motivation

In our `LoaderCollection` component we only want to load and import the collections of active eth or evm networks.


AntonioVentilii (approved):
Nice! Tks for the fix!

PR #9049, opened 2025-09-22
feat(frontend): added isIcrcTokenSupportsIcrc2 util

roman-nazaruk:
# Motivation

We need an util method that checks is icrc token supports icrc2 standard

# Changes

Added new util method isIcrcTokenSupportIcrc2

# Tests

Covered new logic with tests.

### DIFF
--- a/src/frontend/src/icp/utils/icrc.utils.ts
+++ b/src/frontend/src/icp/utils/icrc.utils.ts
@@ -6,8 +6,15 @@ import {
 	ICONFUCIUS_LEDGER_CANISTER_ID,
 	ODINDOG_LEDGER_CANISTER_ID
 } from '$env/networks/networks.icrc.env';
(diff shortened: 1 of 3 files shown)


src/frontend/src/icp/utils/icrc.utils.ts
```diff
@@ -169,6 +177,21 @@ export const icTokenIcrcCustomToken = (token: Partial<IcrcCustomToken>): token i
...
+}: {
+	identity: OptionIdentity;
+	ledgerCanisterId: CanisterIdText;
+}) => {
+	const supportedStandards = await icrc1SupportedStandards({
+		identity,
+		ledgerCanisterId
+	});
+
+	return supportedStandards.some((standard) => standard.name === IcTokenStandards.icrc2);
```
AntonioVentilii:
```suggestion
	return supportedStandards.some(({name}) => name === IcTokenStandards.icrc2);
```

PR #9050, opened 2025-09-22
chore(frontend): Remove index canisters from AAA, BURN, and CLOUD

StefanBerger-DFINITY:
# Motivation

We have 3 curated non-SNS tokens where the index canister is dead for a while:
- AAA
- BURN
- CLOUD

# Changes

- Remove the index from their configs

# Tests
image


AntonioVentilii (approved):
Thank you!!!

PR #9052, opened 2025-09-22
fix(frontend): load kongswap tokens

roman-nazaruk:
# Motivation

We encountered an issue retrieving tokens from the KongSwap canister due to a response-size limit. As a hotfix, we switched to fetching tokens individually.

# Changes

- Implemented Promise.allSettled in allIcrcTokens to request tokens one by one.
- For each supported token, retrieve metadata from kong_swap.

# Tests

Fixed tests.

### DIFF
--- a/src/frontend/src/lib/components/swap/Swap.svelte
+++ b/src/frontend/src/lib/components/swap/Swap.svelte
@@ -58,7 +58,7 @@
 		try {
 			await loadKongSwapTokensService({
 				identity: $authIdentity,
-				allIcrcTokens: [{ ...ICP_TOKEN }, ...$allIcrcTokens]
+				allIcrcTokens: [ICP_TOKEN, ...$allIcrcTokens]
 			});
 
 			return 'ready';
--- a/src/frontend/src/lib/components/swap/Swap.svelte
+++ b/src/frontend/src/lib/components/swap/Swap.svelte
@@ -43,7 +43,7 @@
 		store: icTokenFeeStore
 	});
 
-	const isDisabled = (): boolean => isNullish($kongSwapTokensStore) || isNullish($allIcrcTokens);
+	const isDisabled = (): boolean => isNullish($kongSwapTokensStore);
 
 	const loadKongSwapTokens = async (): Promise<'ready' | undefined> => {
 		if (isNullish($authIdentity)) {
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/components/swap/Swap.svelte
```diff
@@ -53,7 +57,8 @@
 
 		try {
 			await loadKongSwapTokensService({
-				identity: $authIdentity
+				identity: $authIdentity,
+				allIcrcTokens: [{ ...ICP_TOKEN }, ...$allIcrcTokens]
```
AntonioVentilii: maybe here you can use derived store `defaultIcpTokens`

`allIcrcTokens: [...$defaultIcpTokens, ...$allIcrcTokens]`

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -158,20 +158,36 @@ export const fetchKongSwap = async ({
...
+}: {
+	identity: Identity;
+	allIcrcTokens: IcToken[];
+}): Promise<void> => {
+	const kongSwapTokens = await Promise.allSettled(
+		allIcrcTokens.map((token: IcToken) =>
+			kongTokens({
+				identity,
+				tokenLedgerCanisterId: token.ledgerCanisterId
+			})
```
AntonioVentilii, line 173:
```suggestion
	const kongSwapTokens = await Promise.allSettled(
		allIcrcTokens.map(({ledgerCanisterId: tokenLedgerCanisterId}: IcToken) =>
			kongTokens({
				identity,
				tokenLedgerCanisterId
			})
```

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -158,20 +158,36 @@ export const fetchKongSwap = async ({
...
-					: acc,
-			{}
+export const loadKongSwapTokens = async ({
+	identity,
+	allIcrcTokens
+}: {
+	identity: Identity;
+	allIcrcTokens: IcToken[];
+}): Promise<void> => {
+	const kongSwapTokens = await Promise.allSettled(
```
AntonioVentilii, line 168: don't we want `Promise.all`? So that errors are handled elsewhere (for example lack of nodes, or any other non-kongswap error). The method returns empty if the token is not supported, right

AntonioVentilii (approved):
LGTM tks! A small doubt:

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: As it was discussed offline, we should keep it as Promise.allSettled

PR #9055, opened 2025-09-22
feat(frontend): add transparent background to QR code scanner

artkorotkikh-dfinity:
# Motivation
QR scanner used old colors 

# Changes
- Remove blue-purple overlay from QR scanner camera view
- Use elegant single-line CSS solution: color: transparent
- Improve user experience with clear, unobstructed camera view for QR scanning

# Tests
Before: 
image

After: 
image

### DIFF
--- a/src/frontend/src/lib/components/qr/QrCodeScanner.svelte
+++ b/src/frontend/src/lib/components/qr/QrCodeScanner.svelte
@@ -51,7 +51,6 @@
 
 <style lang="scss">
 	.qr-code-wrapper {
-		--primary-rgb: 50, 20, 105;
-		color: rgba(var(--primary-rgb), 0.6);
+		color: transparent;
 	}
 </style>


src/frontend/src/lib/components/qr/QrCodeScanner.svelte
```diff
@@ -51,7 +51,6 @@
 
 <style lang="scss">
 	.qr-code-wrapper {
-		--primary-rgb: 50, 20, 105;
-		color: rgba(var(--primary-rgb), 0.6);
+		color: transparent;
```
peterpeterparker, line 54: Do we need to set transparent or we can actually remove the all style block?

peterpeterparker: Tested offline, it needs to be transparent.

peterpeterparker (approved):
LGTM, thx

PR #9058, opened 2025-09-22
style(frontend): Layout refactoring

daviddecentage:
# Motivation

# Changes

# Tests


AntonioVentilii (approved):
I have to believe in your tests eheh!

Kidding aside, approving, but, before merging, may you please add just a small video or some pictures of the several views? For example, mobile (big and small), wide screen, small-width screen, small-height screen (I remember this being an issue); and all of them both logged in and logged out

PR #9060, opened 2025-09-22
fix(frontend): Center the loading message in the signer flow

AntonioVentilii:
# Motivation

The loading message in the signer flow was not centered anymore because we use now `AgreementGuard` to wrap the login: this component has a `<div> `element for fading that was not inheriting the `flex` class. However, there is not really need for a fading transition here, so we can remove it.

### Before

Screenshot 2025-09-22 at 21 31 47

### After

Screenshot 2025-09-22 at 21 52 07

### Entire flow

https://github.com/user-attachments/assets/aa7672d6-b722-40a2-a929-6caeb2e7d1c9

### DIFF
--- a/src/frontend/src/lib/components/qr/QrCodeScanner.svelte
+++ b/src/frontend/src/lib/components/qr/QrCodeScanner.svelte
@@ -51,7 +51,6 @@
 
 <style lang="scss">
 	.qr-code-wrapper {
-		--primary-rgb: 50, 20, 105;
-		color: rgba(var(--primary-rgb), 0.6);
+		color: transparent;
 	}
 </style>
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/guard/AgreementsGuard.svelte
```diff
@@ -16,7 +16,7 @@
 </script>
 
 {#if $hasAcceptedAllLatestAgreements}
-	<div in:fade>
+	<div class="flex flex-1 flex-col" in:fade>
```
peterpeterparker: I wonder if `contents` style would work as well? Assuming this guard is used around other elements that just the signer, to avoid side effect like fixing the the spinner in the signer but, misaligning something else.

AntonioVentilii: As suggested offline, I am removing the `div` altogether

PR #9071, opened 2025-09-23
feat(frontend): parseToAiAssistantTokens util

DenysKarmazynDFINITY:
# Motivation

We need to update the way we generate an available tokens list for LLM - similar to the contacts, we create a separate util for that. It filters out NFTs tokens (not possible to send via the console yet) and returns all required data.

### DIFF
--- a/src/frontend/src/lib/types/ai-assistant.ts
+++ b/src/frontend/src/lib/types/ai-assistant.ts
@@ -1,6 +1,6 @@
 import type { Address } from '$lib/types/address';
 import type { ContactAddressUiWithId, ExtendedAddressContactUi } from '$lib/types/contact';
-import type { Token, TokenStandard } from '$lib/types/token';
+import type { Token } from '$lib/types/token';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/types/ai-assistant.ts
```diff
@@ -63,3 +63,10 @@ export interface AiAssistantContactUi
 }
 
 export type AiAssistantContactUiMap = Record<string, AiAssistantContactUi>;
+
+export interface AiAssistantToken {
+	name: string;
+	symbol: string;
+	standard: TokenStandard;
+	networkId: string;
+}
```
AntonioVentilii: does it make sense to get it directly from the `Token` type? like `Pick<Token,...`

DenysKarmazynDFINITY: yep, makes sense - everything except networkId is now picked from Token

PR #9074, opened 2025-09-23
feat(frontend): extend AI contact addresses with acceptedTokenStandards

DenysKarmazynDFINITY:
# Motivation

We need to extend AI contact addresses with acceptedTokenStandards. This field is gonna help LLM to understand whether a token can be sent to an address. It's especially useful in some particular cases: e.g. `Icrcv2` address can be a principal or an account id - if it's the former, than both ICP and ICRC tokens can be sent to it, if the latter - only ICP.

The system prompt will be updated in the follow-up PR.

### DIFF
--- a/src/frontend/src/lib/utils/ai-assistant.utils.ts
+++ b/src/frontend/src/lib/utils/ai-assistant.utils.ts
@@ -32,8 +32,10 @@ export const parseToAiAssistantContacts = (
 									: ['icp', 'icrc']
 								: addressType === 'Sol'
 									? ['solana', 'spl']
-									: // We do not include NFT standards here until the console can handle sending them
-										['ethereum', 'erc20', 'dip20'],
+									: addressType === 'Eth'
+										? // We do not include NFT standards here until the console can handle sending them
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/utils/ai-assistant.utils.ts
```diff
@@ -21,7 +22,21 @@ export const parseToAiAssistantContacts = (
...
+						addressType === 'Btc'
+							? ['bitcoin']
+							: addressType === 'Icrcv2'
+								? !isIcrcAddress(address)
+									? ['icp']
+									: ['icp', 'icrc']
+								: addressType === 'Sol'
+									? ['solana', 'spl']
+									: // We do not include NFT standards here until the console can handle sending them
+										['ethereum', 'erc20', 'dip20'],
```
AntonioVentilii: would it be better to have an empty as fallback? just for insurance

DenysKarmazynDFINITY: oh yes, I meant to add a specific check for Eth addressType - thanks!

PR #9079, opened 2025-09-23
feat(frontend): update system prompt to reflect the recent changes

DenysKarmazynDFINITY:
# Motivation

With all the changes made for AI contacts and tokens, we can now update the information we send to LLM via system prompt.


src/frontend/src/lib/constants/ai-assistant.constants.ts
```diff
@@ -118,7 +120,7 @@ export const getAiAssistantToolsDescription = ({
 			function: {
 				name: 'show_filtered_contacts',
 				description: toNullable(
-					'Filter the provided contacts list by semantic meaning and return only matching addressIds. Always return only { "addressIds": [...] }. If no matches, return an empty array. Never include any other fields.'
```
AntonioVentilii, line 121: you removed `Always return only { "addressIds": [...] }` intentionally, right?

DenysKarmazynDFINITY: yes, this part was not really needed, so I cleaned it up

PR #9091, opened 2025-09-23
fix(frontend): `LoaderMultipleEthTransactions` should not stop loading

AntonioVentilii:
# Motivation

Component `LoaderMultipleEthTransactions` stops loading the transactions, since we track which tokens are already loaded.

This was useful when we had no control over the loading of the transactions being called multiple times.

However, now we have the `loading` state that fixes this issue, so we don't need the list of tokens already loaded anymore.


src/frontend/src/eth/components/loaders/LoaderMultipleEthTransactions.svelte
```diff
@@ -42,10 +35,10 @@
 
 		loading = true;
 
-		const loader = batchLoadTransactions({ tokens, tokensAlreadyLoaded });
+		const loader = batchLoadTransactions({ tokens });
 
-		for await (const results of loader) {
-			tokensAlreadyLoaded = [...tokensAlreadyLoaded, ...batchResultsToTokenId(results)];
+		for await (const _ of loader) {
```
DenysKarmazynDFINITY, line 40: do we still need the loop?

AntonioVentilii: yes, we need to await the results since `loader` is an async generator

PR #9103, opened 2025-09-24
feat(frontend): disable to install oisy as PWA

BonomoAlessandro:
# Motivation

When OISY is installed as a PWA standalone it is not possible to connect other DEX's that are running in PWA standalone mode to OISY. To resolve this issue we have changed the `display-mode` setting to `browser`, which prevents OISY from being installed as a PWA anymore.
Users who have already installed OISY as a PWA are still affected. For them, we display a warning message to inform them that they should reinstall it.

# Changes

- activate browser mode

# Tests

**New:**
image


src/frontend/src/lib/utils/device.utils.ts
```diff
@@ -15,3 +15,11 @@ export const isMobile = (): boolean => {
 };
 
 export const isDesktop = () => !isMobile();
+
+export const isPWAStandalone = () => {
+	if ('standalone' in navigator && nonNullish(navigator.standalone)) {
+		return navigator.standalone === true;
```
roman-nazaruk: Do we really need this?

src/frontend/src/lib/components/core/Banner.svelte
```diff
@@ -28,6 +30,17 @@
 			</button>
 		</WarningBanner>
 	</div>
+{:else if isPWAStandalone()}
```
roman-nazaruk: Is it right that we should show it only on PROD env?

src/frontend/src/lib/utils/device.utils.ts
BonomoAlessandro: Yes, that is to detect it on IPhones

src/frontend/src/lib/components/core/Banner.svelte
BonomoAlessandro: No it should be displayed everywhere. You are right

roman-nazaruk (approved):

PR #9110, opened 2025-09-24
feat(frontend): Adds network to transaction modals

daviddecentage:
# Motivation

We want to also display the network in the list of transaction metadata in the transaction modals.

# Changes

Added network metadata display to 
- Btc
- Eth
- Ic
- Sol

transaction modals

# Tests

Extended tests for new field.

image


BonomoAlessandro (approved):
LGTM

PR #9114, opened 2025-09-24
fix(frontend): Dont persist search modal tokenlist

daviddecentage:
# Motivation

If we leave the tokenlist in modals we should not persist the last search.

# Changes

Achieved by not using the context flterQuery value as the initial state value

# Tests

https://github.com/user-attachments/assets/415b2021-b4e6-4dea-aefd-617b614a775b


AntonioVentilii (approved):
LGTM tks!

PR #9117, opened 2025-09-24
feat(frontend): Wait for WalletConnect disconnection on logout

AntonioVentilii:
# Motivation

Just for insurance, we await for WalletConnect to disconnect on logout (or up to 10 seconds).


src/frontend/src/lib/services/auth.services.ts
```diff
@@ -233,6 +235,17 @@ const clearSessionStorage = async () => {
 	sessionStorage.clear();
 };
 
+const disconnectWalletConnect = async () => {
```
roman-nazaruk, line 238: Could you please some tests to this method?

AntonioVentilii: Good point! Included

PR #9125, opened 2025-09-24
feat(frontend): improves nfts list view

BonomoAlessandro:
# Motivation

If there are no `nfts` displayed in the "default" section but nfts that are marked as `spam` or `hidden` (and active), we also want to display the "default" section with an additional logo.

# Tests

**New:**

NFT view:
image
image

Collection view:
image
image


src/frontend/src/lib/components/nfts/NftList.svelte
```diff
@@ -9,24 +10,27 @@
 		icon?: Snippet;
 		nfts: Nft[];
 		testId?: string;
+		isMain?: boolean;
```
AntonioVentilii: what is `isMain`? it is not quite clear what `main` refers to

AntonioVentilii (commented):
From what I see, all its

src/frontend/src/lib/components/nfts/NftCollectionList.svelte
```diff
@@ -1,36 +1,37 @@
...
-	<div data-tid={testId}>
+<div data-tid={testId}>
+	{#if notEmptyCollections.length > 0 || isMain}
 		<div class="mt-2 flex items-center gap-2">
 			{@render icon?.()}
 			<h5>{title}</h5>
 		</div>
+	{/if}
 
+	{#if notEmptyCollections.length > 0}
```
AntonioVentilii: these conditions can be merged right?

{#if notEmptyCollections.length > 0}
 {#if  isMain}
		<div class="mt-2 flex items-center gap-2">
			{@render icon?.()}
			<h5>{title}</h5>
		</div>
{/if}
......

src/frontend/src/lib/components/nfts/NftList.svelte
BonomoAlessandro: This parameter should "mark" the `main` section. I will rename the parameter to `asMainSection`.

src/frontend/src/lib/components/nfts/NftCollectionList.svelte
BonomoAlessandro: It cannot be merged, but a little bit simplified.

PR #9136, opened 2025-09-24
fix(frontend): fixes rewards context

BonomoAlessandro:
# Motivation

To prevent flakyness and be more efficient, the reward context should only be created once.


AntonioVentilii (approved):
Nice one! Tks!

PR #9144, opened 2025-09-25
refactor(frontend): Migrate `SolSendTokenWizard` to Svelte 5

AntonioVentilii:
# Motivation

Migrating component `SolSendTokenWizard` to Svelte 5.


src/frontend/src/sol/components/send/SolSendTokenWizard.svelte
```diff
@@ -53,27 +51,45 @@
...
 	} from '$sol/stores/sol-fee.store';
 
-	export let currentStep: WizardStep | undefined;
-	export let destination = '';
-	export let amount: OptionAmount = undefined;
-	export let sendProgressStep: string;
-	export let selectedContact: ContactUi | undefined = undefined;
+	interface Props {
+		currentStep?: WizardStep;
+		destination: string;
```
DenysKarmazynDFINITY: if you want to make `destination` required, maybe let's make it `destination: string | undefined;` as I assume it's not defined when user just opens the modal. `destination?: string` is also totally fine

src/frontend/src/sol/components/send/SolSendTokenWizard.svelte
```diff
@@ -53,27 +51,45 @@
...
+		onNext,
+		onSendBack,
+		onTokensList
+	}: Props = $props();
 
 	const { sendToken, sendTokenDecimals } = getContext<SendContext>(SEND_CONTEXT_KEY);
 
-	let network: Network | undefined = undefined;
-	$: ({ network } = $sendToken);
+	let { network } = $derived($sendToken);
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: I would vote for `let network = $derived($sendToken.network);` to keep it consistent with other $derived

src/frontend/src/sol/components/send/SolSendTokenWizard.svelte
DenysKarmazynDFINITY: or, since in the parent component it's typed as `destination: string`, let's just remove the fallback value

AntonioVentilii: nono, you are right! let's just keep it consistent with the refactoring: if it was optional before, it should be now too

AntonioVentilii:
```suggestion
		destination?: string;
```

PR #9149, opened 2025-09-25
fix(frontend): Ic transaction seperate fee for send and approve

daviddecentage:
# Motivation
We wat to display the fees seperately from the amount for ICP and ICRC send and approve transactions.

# Changes

- Seperate amount and fee in Transaction modal
- Show cumulated amount + fee in transaction list like before
- Show fee in transaction list for approve transactions
- Display allowance amount in hero header for approve transactions
- Display fee for approve transactions

# Tests

Added and adjusted tests + manually tested on FE3.


src/frontend/src/tests/icp/services/ic-transactions.services.spec.ts
```diff
@@ -178,7 +178,8 @@ describe('ic-transactions.services', () => {
 			incoming: true,
 			fromExplorerUrl: `${ICP_EXPLORER_URL}/account/${transaction.from}`,
 			toExplorerUrl: `${ICP_EXPLORER_URL}/account/${mockIdentity.getPrincipal().toString()}`,
-			txExplorerUrl: `${ICP_EXPLORER_URL}/transaction/${transaction.id}`
+			txExplorerUrl: `${ICP_EXPLORER_URL}/transaction/${transaction.id}`,
+			fee: 0n
```
AntonioVentilii:
```suggestion
			fee: ZERO
```

src/frontend/src/lib/components/transactions/Transaction.svelte
```diff
@@ -58,9 +59,14 @@
 		tokenId,
 		children,
 		onClick,
+		fee,
 		approveSpender
 	}: Props = $props();
 
+	const amountWithFee = $derived(
+		nonNullish(cardAmount) && nonNullish(fee) ? cardAmount + fee : cardAmount
```
AntonioVentilii: is this correct? what if it is incoming transaction? Is it handled when setting the `fee`

src/frontend/src/icp/utils/icrc-transactions.utils.ts
```diff
@@ -121,11 +119,7 @@ export const mapIcrcTransaction = ({
 	const transferFee = fromNullishNullable(fromNullable(transfer)?.fee);
 
 	// for approve we shows the fee value
-	const value = isApprove
-		? approveFee
-		: nonNullish(data?.amount)
-			? data.amount + (isTransfer && source.incoming === false ? (transferFee ?? ZERO) : ZERO)
-			: undefined;
+	const value = isApprove ? approveFee : nonNullish(data?.amount) ? data.amount : undefined;
```
AntonioVentilii:
```suggestion
	const value = isApprove ? approveFee : data?.amount;
```

src/frontend/src/icp/utils/icp-transactions.utils.ts
```diff
@@ -85,26 +85,18 @@ export const mapIcpTransaction = ({
...
-		amount: Tokens;
-	}): bigint => amount.e8s + (incoming === false ? fee.e8s : ZERO);
-
 	if ('Approve' in operation) {
 		const approve = operation.Approve;
-		const approveValue = approve.fee.e8s;
-		const approveExpiresAt = fromNullable(approve.expires_at)?.timestamp_nanos;
+		const approveValue = approve.allowance.e8s;
+		const approveFee = approve?.fee?.e8s;
+		const approveExpiresAt = fromNullable(approve?.expires_at)?.timestamp_nanos;
```
AntonioVentilii:
```suggestion
		const approveExpiresAt = fromNullable(approve.expires_at)?.timestamp_nanos;
```

src/frontend/src/icp/utils/icp-transactions.utils.ts
```diff
@@ -85,26 +85,18 @@ export const mapIcpTransaction = ({
...
-		fee: Tokens;
-		amount: Tokens;
-	}): bigint => amount.e8s + (incoming === false ? fee.e8s : ZERO);
-
 	if ('Approve' in operation) {
 		const approve = operation.Approve;
-		const approveValue = approve.fee.e8s;
-		const approveExpiresAt = fromNullable(approve.expires_at)?.timestamp_nanos;
+		const approveValue = approve.allowance.e8s;
+		const approveFee = approve?.fee?.e8s;
```
AntonioVentilii:
```suggestion
		const approveFee = approve.fee?.e8s;
```

src/frontend/src/icp/components/transactions/IcTransactionModal.svelte
```diff
@@ -113,6 +113,21 @@
 					/>
 				</span>
 			</ListItem>
+
+			{#if nonNullish(fee) && nonNullish(token) && (type === 'send' || type === 'approve')}
```
AntonioVentilii: is more correct to use `incoming` here, WDYT?

`...  && (!incoming)}`

or something similar

src/frontend/src/lib/components/transactions/Transaction.svelte
daviddecentage: Then it will be just the amount as it was before, since we only add the fee in the cases nessesary.

src/frontend/src/icp/components/transactions/IcTransactionModal.svelte
daviddecentage: The task was to add it for send and approve, not exactly sure if it would make more sense to use `!incoming`.

I think we would have to add the fee in the mapping for the other types as well to make use of it. It would still only display for send and approve without adding to the mapping.

Preferences from your side?

AntonioVentilii: my train of thought was: when do you show the fee? When it is not an incoming transaction basically, if we follow the logic of the `fee`. if we make it specific for type it not that scalable and the `incoming` value loses a bit of significance

src/frontend/src/lib/components/transactions/Transaction.svelte
AntonioVentilii: you are setting the same fee for both sending and receive, while we should not show the fee for a receive transaction, I suppose

src/frontend/src/icp/components/transactions/IcTransactionModal.svelte
daviddecentage: Good points. I updated it

src/frontend/src/lib/components/transactions/Transaction.svelte
daviddecentage: It wont be displayed anyways.

If we want the incoming flag in here for this condition, I would have to create another PR first to pass it as a prop to this component from BtcTransaction, SolTransacion etc etc.

Your call

AntonioVentilii: i am lost... in `mapIcrcTransaction` you are setting the fee for the `Transfer` transaction. Here is the part of the code that you included

```ts
return {
  ...
  ...(nonNullish(transferFee) && { fee: transferFee }),
  ...
};
```

So, for ALL `Transfer` transaction (send and receive) you are adding the fee to the amount. Isn't it wrong? It should be added ONLY when you have a send transaction, because the user paid that fee. If it is a receive transaction, why are you adding the fee?

AntonioVentilii: if I read this component, I read "always sum amount and fee if they are not undefined, independently of the type of transaction", meaning that all transactions that have a fee and an amount should display the sum of both. Conceptually, this is wrong.

Agnostically, I cannot rely on the fact that we call "correctly" the component: if i call this component and I see the prop `fee`, I will obviously pass the `fee` of the transaction. I cannot suppose that I MUST pass what I suppose the `fee` is, since I am passing the `type` of transaction too

You know what I mean?

The logic of Transaction should be clear: what is the amount that I should show? On what logic is based? I sum fee and amount when exactly?

daviddecentage: I see, right I can just not set the fee for the cases we dont wanna display it. Updated

AntonioVentilii (approved):
Thank you!

PR #9153, opened 2025-09-25
fix(frontend): Guarantee synced key pair and delegation for `authClient`

AntonioVentilii:
# Motivation

When multiple OISY tabs are open in the same browser, each creates a new `authClient` object with its own key pair. Since keys are stored in IndexedDB (IDB) as a security measure to avoid key injection, the last loaded tab overwrites the cached keys.

This causes a problem: if a user logs in on a tab that is not the "latest" one (without refreshing first), the existing `authClient` in that tab uses a key pair different from the one currently cached in IDB. It will then request a delegation for those non-cached keys, and store that delegation in IDB too. As a result, the cache ends up with a delegation that does not match the cached key.

Later, if we recreate the `authClient` (e.g. inside a worker) or refresh another tab, the mismatch between keys and delegation leads to invalid signatures and errors.

To prevent this, we must ensure that after login, the correct key is always cached in IDB, overwriting any previously stored key. This guarantees that delegation and keys remain in sync.

# Credits

Big thanks to @ilbertt for debugging and providing a solution! Hopefully directly in `AuthClient` in the future, eheh!

# Changes

- Use the login callback to get the key pair of the current `authClient` object (the one used to fetch a delegation).
- Cache such key pair, overwriting any existing one in IDB cache.

# Note

This does not affect the single-async-call restriction of some browsers, since the callback happens AFTER the login is successful.

# Tests

There are several ways to test. The simplest reproduction is:

1. Open tab 1
2. Open tab 2
3. Login in tab 1 (without refreshing)
4. ERROR!

### Before

https://github.com/user-attachments/assets/da7412fc-81b9-4ff2-81ad-7604426e6627

### After

https://github.com/user-attachments/assets/a0ad866f-f585-408e-afe2-f8ce0c23eb9c


AntonioVentilii:
Pending security review

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -99,7 +104,26 @@ const initAuthStore = (): AuthStore => {
...
+						// As a result, the cache ends up with a delegation that does not match the cached key.
+						//
+						// Later, if we recreate the `authClient` (e.g. inside a worker) or refresh another tab, the mismatch
+						// between keys and delegation leads to invalid signatures and errors.
+						//
+						// To prevent this, we must ensure that after login, the correct key is always cached in IDB,
+						// overwriting any previously stored key. This guarantees that delegation and keys remain in sync.
+						//
+						// TODO: Remove this when authClient will do it by itself during login
+						const key = authClient?.['_key'];
```
peterpeterparker: - Maybe add a `nonNullish` check?
- Maybe extract this to a separate local function - like we did for `pickAuthClient`?

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -99,7 +104,26 @@ const initAuthStore = (): AuthStore => {
 
 				await authClient.login({
 					maxTimeToLive: AUTH_MAX_TIME_TO_LIVE,
-					onSuccess: () => {
+					onSuccess: async () => {
+						// When multiple OISY tabs are open in the same browser, each creates a new `authClient` object
```
peterpeterparker: Given it's a hacky workaround covered by no tests, shall we add a comment like:

```
// ⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️
// ⚠️          **Warning:**       ⚠️
// ⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️
```

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -99,7 +104,26 @@ const initAuthStore = (): AuthStore => {
...
+						//
+						// Later, if we recreate the `authClient` (e.g. inside a worker) or refresh another tab, the mismatch
+						// between keys and delegation leads to invalid signatures and errors.
+						//
+						// To prevent this, we must ensure that after login, the correct key is always cached in IDB,
+						// overwriting any previously stored key. This guarantees that delegation and keys remain in sync.
+						//
+						// TODO: Remove this when authClient will do it by itself during login
+						const key = authClient?.['_key'];
+						await authClientStorage.set(KEY_STORAGE_KEY, (key as ECDSAKeyIdentity).getKeyPair());
```
peterpeterparker: What happens if this fails and throws an error? Should we allow the user to use the session or catch the error and sign-out?

AntonioVentilii: It is better to signout, I think. I modified the function, have a look please

src/frontend/src/lib/stores/auth.store.ts
AntonioVentilii: Done!

src/frontend/src/lib/stores/auth.store.ts
AntonioVentilii: Done on both counts

AntonioVentilii:
> To address this, we should e.g. implement a [broadcast channel](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) and reload the authclient/identity in any tabs if one tab logs in.

@peterpeterparker, as per your suggestion, working on this in:

 - https://github.com/dfinity/oisy-wallet/pull/9194

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -54,6 +61,48 @@ const initAuthStore = (): AuthStore => {
...
+		}
+
+		try {
+			const key = authClient['_key'];
+
+			await authClientStorage.set(KEY_STORAGE_KEY, (key as ECDSAKeyIdentity).getKeyPair());
+		} catch (_: unknown) {
+			// In case of error in this flow, we prefer to log out the user to avoid possible further conflicts.
+			// The logout will in any case refresh the page.
+			await warnSignOut(get(i18n).auth.warning.reload_and_retry);
```
peterpeterparker: This creates a recursive imports between the store and the service. Maybe we can just call `signOut()` and `console.error` from the store assuming it's an edge case?

My main concerns are handling the error in a callback and ensuring the auth store identity has the state we expect.

AntonioVentilii: right! but I will leave the toast error too, to have some feedback for the user, WDYT?

AntonioVentilii: However it is not that nice, since it is basically repeated code:

```ts
const overwriteStoredIdentityKey = async () => {
  ...
  } catch (_: unknown) {
	  // In case of error in this flow, we prefer to log out the user to avoid possible further conflicts.
	  // The logout will in any case refresh the page.
  
	  busy.start();
  
	  await authStore.signOut();
  
	  const PARAM_MSG = 'msg';
	  const PARAM_LEVEL = 'level';
  
	  const appendMsgToUrl = (msg: ToastMsg) => {
		  if (typeof window === 'undefined') {
			  return;
		  }
  
		  const { text, level } = msg;
  
		  const url: URL = new URL(window.location.href);
  
		  url.searchParams.append(PARAM_MSG, encodeURI(text));
		  url.searchParams.append(PARAM_LEVEL, level);
  
		  replaceHistory(url);
	  };
  
	  appendMsgToUrl({
		  text: get(i18n).auth.warning.reload_and_retry,
		  level: 'warn'
	  });
  
	  window.location.reload();
  }
};
```

AntonioVentilii: Discussed offline: i will remove the user feedback

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -54,6 +60,50 @@ const initAuthStore = (): AuthStore => {
...
+	 *
+	 * To prevent this, we must ensure that after login, the correct key is always cached in
+	 * IDB, overwriting any previously stored key. This guarantees that delegation and keys
+	 * remain in sync.
+	 *
+	 * TODO: Remove this when `authClient` will handle it by itself during login.
+	 */
+	const overwriteStoredIdentityKey = async () => {
+		// This does not happen at this point, since the method is called after the login is successful.
+		// But we add this check to make is type-safe.
```
peterpeterparker: "to make is type-safe" -> "it"?

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -54,6 +60,50 @@ const initAuthStore = (): AuthStore => {
...
+		if (isNullish(authClient)) {
+			return;
+		}
+
+		try {
+			const key = authClient['_key'];
+
+			await authClientStorage.set(KEY_STORAGE_KEY, (key as ECDSAKeyIdentity).getKeyPair());
+		} catch (_: unknown) {
+			// In case of error in this flow, we prefer to log out the user and refresh the page to avoid possible further conflicts.
```
peterpeterparker, line 97: Maybe "In the unlikely event of an error while setting a value in IndexedDB, we log out the user and refresh the page to prevent potential conflicts."

peterpeterparker (approved):
LGTM, thx

src/frontend/src/lib/stores/auth.store.ts
peterpeterparker: Main concern was recursive imports between stores or stores and services.

src/frontend/src/lib/stores/auth.store.ts
AntonioVentilii: Changed

src/frontend/src/lib/stores/auth.store.ts
AntonioVentilii: 🤦‍♂️

ilbertt:
@AntonioVentilii @peterpeterparker I've opened a PR with the fix: https://github.com/dfinity/icp-js-auth/pull/34

PR #9165, opened 2025-09-25
chore(deps): Remove `jimp`

AntonioVentilii:
# Motivation

While bumping `zod` to v4 we encountered a library that was sort of causing dependencies issue: `jimp`.

After a quick analysis, it is used only in the E2E tests, that are either way being skipped for now and under review.

So, to prioritize the maintenance of the repo, we comment such tests and uninstall `jimp`.


e2e/utils/qr-code.utils.ts
```diff
@@ -2,23 +2,28 @@ import { nonNullish } from '@dfinity/utils';
 import { Jimp } from 'jimp';
 import jsQR from 'jsqr';
 
+// TODO: uncomment and replace Jimp with native functions
```
peterpeterparker: Maybe add "unless the library bumps its dependency to zod" or something?

AntonioVentilii: done!

PR #9193, opened 2025-09-26
chore(scripts): remove unused update-agent.sh script

peterpeterparker:
# Motivation

The script `update-agent.sh` seems unused and therefore can be deleted.


AntonioVentilii (approved):
Tks!

PR #9194, opened 2025-09-26
feat(frontend): Broadcast login success among tabs to resync auth

AntonioVentilii:
# Motivation

When multiple OISY tabs are open in the same browser, each creates a new `authClient` object with its own key pair. Since keys are stored in IndexedDB (IDB) as a security measure to avoid key injection, the last loaded tab overwrites the cached keys.

This causes a problem: if a user logs in on a tab that is not the "latest" one (without refreshing first), the existing `authClient` in that tab uses a key pair different from the one currently cached in IDB. It will then request a delegation for those non-cached keys, and store that delegation in IDB too. As a result, the cache ends up with a delegation that does not match the cached key.

Later, if we recreate the `authClient` (e.g. inside a worker) or refresh another tab, the mismatch between keys and delegation leads to invalid signatures and errors.

Part of the issue is being fixed in AuthClient directly, but we already merged a temporary workaround (PR https://github.com/dfinity/oisy-wallet/pull/9153).

To try and solve it, we implement a [broadcast channel](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) and force-reload the authclient/identity in any tabs if one tab logs in.

# Credits

Big thanks to @peterpeterparker for the solution!

# Changes

- Create method to force synchronization in the `authStore`.
- Create class `AuthBroadcastChannel` that allows to initialize and close an OISY broadcast channel, plus it can post the "login successful" message.
- The handling of the messages will sync the `authStore`, as explained, but ONLY for same origin and correct message.
- Adapt the main layout to create the channel on mounting and destroying it on unmounting.
- Trigger the posting of the message after every successful login.

# Tests

Added some unittests, mocking directly the `BroadcastChannel` class.

### DIFF
--- a/scripts/update-agent.sh
+++ b/scripts/update-agent.sh
@@ -1,9 +1,0 @@
-#!/usr/bin/env bash
-
-npm rm @dfinity/agent @dfinity/auth-client @dfinity/candid @dfinity/principal @dfinity/ckbtc @dfinity/cketh @dfinity/ledger-icp @dfinity/ledger-icrc @dfinity/utils
-
-npm rm @dfinity/identity-secp256k1 @dfinity/ic-management -D
-
-npm i @dfinity/agent@1.4.0 @dfinity/auth-client@1.4.0 @dfinity/candid@1.4.0 @dfinity/principal@1.4.0 @dfinity/ckbtc@latest @dfinity/cketh@latest @dfinity/ledger-icp@latest @dfinity/ledger-icrc@latest @dfinity/utils@latest
-
-npm i @dfinity/identity-secp256k1@1.4.0 @dfinity/ic-management@latest -D


src/frontend/src/routes/+layout.svelte
```diff
@@ -33,9 +37,9 @@
 
 	const init = async () => {
 		/**
-		 * We use `Promise.allSettled` to ensure that all initialization functions run,
+		 * We use `Promise.allSettled` to ensure that all initialisation functions run,
```
peterpeterparker: Can you undo the unrelated typo fixes to improve the readability of the review?

src/frontend/src/routes/+layout.svelte
```diff
@@ -108,6 +112,21 @@
 		const spinner = document.querySelector('body > #app-spinner');
 		spinner?.remove();
 	});
+
+	let bc = $state<AuthBroadcastChannelInterface | undefined>();
```
peterpeterparker: Do we need a state?

src/frontend/src/routes/+layout.svelte
```diff
@@ -108,6 +112,21 @@
...
+
+	const openBc = () => {
+		if (isNullish(bc)) {
```
peterpeterparker: When would it be non nullish?

src/frontend/src/routes/+layout.svelte
```diff
@@ -108,6 +112,21 @@
...
+			bc = initAuthBroadcastChannel();
+		}
+
+		return () => {
+			bc?.close();
+			bc = undefined;
```
peterpeterparker: Why reset to undefined?

src/frontend/src/lib/services/auth-broadcast.services.ts
```diff
@@ -0,0 +1,60 @@
...
+	close = () => {
+		this.bc.close();
+	};
+
+	postLoginSuccess = () => {
+		this.bc.postMessage(AUTH_BROADCAST_MESSAGE_LOGIN_SUCCESS);
+	};
+}
+
+export const initAuthBroadcastChannel = (): AuthBroadcastChannel => {
```
peterpeterparker: If you use a class why not a static method?

src/frontend/src/lib/services/auth-broadcast.services.ts
```diff
@@ -0,0 +1,60 @@
...
+	close: () => void;
+	postLoginSuccess: () => void;
+}
+
+// If the user has more than one tab open in the same browser,
+// there could be a mismatch of the cached delegation chain vs the identity key of the `authClient` object.
+// This causes the `authClient` to be unable to correctly sign calls, raising Trust Errors.
+// To mitigate this, we use a `BroadcastChannel` to notify other tabs when a login has occurred, so that they can sync their `authClient` object.
+class AuthBroadcastChannel implements AuthBroadcastChannelInterface {
+	private bc: BroadcastChannel;
```
peterpeterparker: If you like a bit more modern:

```
readonly #bc: BroadcastChannel
```

src/frontend/src/lib/services/auth-broadcast.services.ts
```diff
@@ -0,0 +1,60 @@
+import { OISY_URL } from '$lib/constants/oisy.constants';
+import { authStore } from '$lib/stores/auth.store';
+
+export const AUTH_BROADCAST_CHANNEL = 'oisy_auth_channel';
+
+export const AUTH_BROADCAST_MESSAGE_LOGIN_SUCCESS = 'authClientLoginSuccess';
+
+export interface AuthBroadcastChannelInterface {
```
peterpeterparker: What's the goal of the interface?

src/frontend/src/lib/services/auth-broadcast.services.ts
```diff
@@ -0,0 +1,60 @@
...
+	private onLoginSuccess = (handler: () => Promise<void>) => {
+		this.bc.onmessage = async (event) => {
+			if (event.origin === OISY_URL && event.data === AUTH_BROADCAST_MESSAGE_LOGIN_SUCCESS) {
+				await handler();
+			}
+		};
+	};
+
+	init = () => {
+		this.onLoginSuccess(authStore.forceSync);
```
peterpeterparker: I'm not sure why we have a generic implementation - class + onLoginSuccess - but `authStore.forceSync` "hardcoded"?

src/frontend/src/lib/services/auth-broadcast.services.ts
```diff
@@ -0,0 +1,60 @@
...
+// To mitigate this, we use a `BroadcastChannel` to notify other tabs when a login has occurred, so that they can sync their `authClient` object.
+class AuthBroadcastChannel implements AuthBroadcastChannelInterface {
+	private bc: BroadcastChannel;
+
+	constructor() {
+		this.bc = new BroadcastChannel(AUTH_BROADCAST_CHANNEL);
+	}
+
+	private onLoginSuccess = (handler: () => Promise<void>) => {
+		this.bc.onmessage = async (event) => {
```
peterpeterparker: nitpick: can you prefix or rename `event` like `$event`. As a best practice I think it's good to avoid reserved keywords.

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -102,30 +104,38 @@ const initAuthStore = (): AuthStore => {
 		}
 	};
 
-	return {
-		subscribe,
+	const sync = async ({ forceSync }: { forceSync: boolean }) => {
+		authClient = forceSync ? await createAuthClient() : await pickAuthClient();
```
peterpeterparker, line 107: Do we have a test for this?

src/frontend/src/lib/services/auth-broadcast.services.ts
```diff
@@ -0,0 +1,60 @@
...
+			if (event.origin === OISY_URL && event.data === AUTH_BROADCAST_MESSAGE_LOGIN_SUCCESS) {
```
peterpeterparker: IMO asserting the `origin` is the most important feature. Do we have a test for it?

src/frontend/src/routes/+layout.svelte
AntonioVentilii: removed, it was a late commit

src/frontend/src/routes/+layout.svelte
AntonioVentilii: removed, it is not needed

src/frontend/src/routes/+layout.svelte
AntonioVentilii: same as above

src/frontend/src/lib/services/auth-broadcast.services.ts
AntonioVentilii: you mean to use directly the class?

My idea was just to use methods instead of class, but no preference there. You prefer a static method in the class?

src/frontend/src/lib/services/auth-broadcast.services.ts
AntonioVentilii: nice! changed

src/frontend/src/lib/services/auth-broadcast.services.ts
AntonioVentilii: no real goal actually, i removed it

src/frontend/src/lib/services/auth-broadcast.services.ts
AntonioVentilii: done!

src/frontend/src/lib/stores/auth.store.ts
AntonioVentilii: still no test for the entire `authStore` class... it is in my todo list still

src/frontend/src/lib/services/auth-broadcast.services.ts
AntonioVentilii: yep, there is a unittest that fails if the origin is different

src/frontend/src/routes/+layout.svelte
AntonioVentilii: done, it is the auto-correct tool eheh

src/frontend/src/lib/services/auth-broadcast.services.ts
```diff
@@ -0,0 +1,54 @@
...
+	init = () => {
+		this.onLoginSuccess(authStore.forceSync);
+	};
+
+	close = () => {
+		this.#bc.close();
+	};
+
+	postLoginSuccess = () => {
+		this.#bc.postMessage(AUTH_BROADCAST_MESSAGE_LOGIN_SUCCESS);
```
peterpeterparker: Maybe I'm thinking to far in the future but, a bit like post message, should we:

1. use Zod to type the messages
2. already use an object that contains a message? in case in the future we would like to exchange msg type and data?

AntonioVentilii: ah, like the `postData`, yes! I can do it in this PR, if you are ok, otherwise another PR (before or after)

src/frontend/src/lib/stores/auth.store.ts
peterpeterparker: ok

src/frontend/src/lib/stores/auth.store.ts
```diff
@@ -155,6 +165,13 @@ const initAuthStore = (): AuthStore => {
 							identity: authClient?.getIdentity()
 						}));
 
+						// If the user has more than one tab open in the same browser,
+						// there could be a mismatch of the cached delegation chain vs the identity key of the `authClient` object.
+						// This causes the `authClient` to be unable to correctly sign calls, raising Trust Errors.
+						// To mitigate this, we use a BroadcastChannel to notify other tabs when a login has occurred, so that they can sync their `authClient` object.
+						const bc = new AuthBroadcastChannel();
```
peterpeterparker: what if it throws an error? should we silence it?

src/frontend/src/routes/+layout.svelte
```diff
@@ -108,6 +109,18 @@
 		const spinner = document.querySelector('body > #app-spinner');
 		spinner?.remove();
 	});
+
+	const openBc = () => {
```
peterpeterparker, line 113: what if it throws an error? should we silence it?

src/frontend/src/lib/services/auth-broadcast.services.ts
peterpeterparker: Discussed offline.

src/frontend/src/lib/services/auth-broadcast.services.ts
peterpeterparker: Discussed offline.

src/frontend/src/lib/services/auth-broadcast.services.ts
peterpeterparker: It's not generic it's about the auth

src/frontend/src/lib/services/auth-broadcast.services.ts
peterpeterparker: Great

src/frontend/src/lib/stores/auth.store.ts
AntonioVentilii: i think we should silence it, since we will not really logout in case of `BroadcastChannel` failure

src/frontend/src/routes/+layout.svelte
AntonioVentilii: i think we should silence it, since we will not really logout in case of `BroadcastChannel` failure

src/frontend/src/lib/services/auth-broadcast.services.ts
AntonioVentilii: This was done by @peterpeterparker in:

- https://github.com/dfinity/oisy-wallet/pull/10271

PR #9206, opened 2025-09-26
refactor(frontend): No zero fallback for fees and valus icp/icrc transactions

daviddecentage:
# Motivation

We dont fall back to zero values if the fee or value/amount is undefined. Instead, the UI should not display it if absent.

# Changes

Removed zero fallbacks and adjusted mapping to not include the properties at all if undefined.

### DIFF
--- a/.github/workflows/backend-checks.yml
+++ b/.github/workflows/backend-checks.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Check if backend files changed
         id: changes
         uses: ./.github/actions/check-backend-changes
-      - uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+      - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         if: steps.changes.outputs.backend == 'true'
         with:
           path: |
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -69,7 +69,7 @@ jobs:
         id: changes
         uses: ./.github/actions/check-backend-changes
 
-      - uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+      - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         if: steps.changes.outputs.backend == 'true'
         with:
           path: |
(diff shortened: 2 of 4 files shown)


src/frontend/src/icp/utils/icp-transactions.utils.ts
```diff
@@ -97,16 +97,16 @@ export const mapIcpTransaction = ({
...
 		return {
 			...tx,
 			type: 'approve',
 			...mapFrom(operation.Approve.from),
-			value: approveValue,
-			approveSpender: approve?.spender,
-			...(nonNullish(approveExpiresAt) && { approveExpiresAt })
+			...(nonNullish(approveValue) && { value: approveValue }),
+			...(nonNullish(approveExpiresAt) && { approveExpiresAt }),
+			approveSpender: approve.spender
```
AntonioVentilii, line 109: Can you do the same for approveSpender?

src/frontend/src/icp/utils/icrc-transactions.utils.ts
```diff
@@ -152,6 +152,6 @@ export const mapIcrcTransaction = ({
 		timestamp,
 		status: 'executed',
 		approveSpender,
-		approveExpiresAt
+		...(nonNullish(approveExpiresAt) && { approveExpiresAt })
```
AntonioVentilii, line 155: Here too

src/frontend/src/icp/utils/icp-transactions.utils.ts
```diff
@@ -97,16 +97,16 @@ export const mapIcpTransaction = ({
 
 	if ('Approve' in operation) {
 		const approve = operation.Approve;
-		const approveValue: bigint = approve?.fee?.e8s ?? ZERO;
+		const approveValue: bigint | undefined = approve?.fee?.e8s;
```
AntonioVentilii:
```suggestion
		const approveValue = approve?.fee?.e8s;
```

src/frontend/src/icp/utils/icp-transactions.utils.ts
daviddecentage: why? it can not be nullish. Type is non-optional string

src/frontend/src/icp/utils/icrc-transactions.utils.ts
daviddecentage: here yes, updated

src/frontend/src/icp/utils/icp-transactions.utils.ts
AntonioVentilii: you are right! thanks for checking!

PR #9215, opened 2025-09-26
refactor(frontend): Migrate `TokenInputCurrencyToken` to Svelte 5

AntonioVentilii:
# Motivation

Migrate component `TokenInputCurrencyToken` to Svelte 5.


src/frontend/src/lib/components/ui/InputCurrency.svelte
```diff
@@ -13,6 +13,7 @@
 		decimals?: number;
 		testId?: string;
 		autofocus?: boolean;
+		onInput?: () => void;
```
DenysKarmazynDFINITY: Shall we align it with TokenInputCurrencyToken component and make the prop required?

AntonioVentilii: The problem is that it is used without `onInput` in `SwapSlippage`. But I agree, it should be compulsory, since it is defined as input. Let me investigate more

PR #9246, opened 2025-09-29
feat(frontend): uses alchemy to load nft contract metadata

BonomoAlessandro:
# Motivation

We want to use `alchemy` to load the `erc721` and `erc1155` contract metadata instead of `infura`.


AntonioVentilii (approved):
As discussed offline, we are going to deprecate `alchemy-sdk` soon, but we currently need more metadata. So, for now, we go with this solution, while we find a substitution

PR #9247, opened 2025-09-29
feat(frontend): Nft filter changes

daviddecentage:
# Motivation

We need to make some adjustments to the Nft search filtering.

- The search should match all collections by collection name or if any Nft in that collection matches the query in Nft name or Nft token ID.

- The search should match all Nfts by collection name, Nft name or Nft token ID.

- Nft collection cards should display only the matched Nfts.

# Changes

- Adjusted the filter method in Nfts util with some utility functions to clean the filter method up
- Also use the filter method in NftCollectionCard

# Tests

Adjusted/extended filter utils tests to cover all different new cases

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
+++ b/src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
@@ -7,6 +7,9 @@
 	import { i18n } from '$lib/stores/i18n.store';
 	import type { NftCollectionUi } from '$lib/types/nft';
 	import { replacePlaceholders } from '$lib/utils/i18n.utils.js';
+	import { filterSortByCollection } from '$lib/utils/nfts.utils';
+	import { tokenListStore } from '$lib/stores/token-list.store';
+	import { nftSortStore } from '$lib/stores/settings.store';
 
 	interface Props {
 		collection: NftCollectionUi;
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
```diff
@@ -38,7 +41,7 @@
 				</span>
 				<span class="absolute z-0 h-full w-full bg-secondary-alt"></span>
 
-				{#each collection.nfts as nft, index (`${nft.id}-${index}`)}
+				{#each filterSortByCollection( { items: collection.nfts, filter: $tokenListStore.filter, sort: $nftSortStore } ) as nft, index (`${nft.id}-${index}`)}
```
roman-nazaruk, line 44:
```suggestion
				{#each filterSortByCollection({ items: collection.nfts, filter: $tokenListStore.filter, sort: $nftSortStore } ) as nft, index (`${nft.id}-${index}`)}
```

BonomoAlessandro: I think it would be more clear to move the function call to the `<script>` part as a `derived` variable.

BonomoAlessandro (approved):
LGTM

src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
daviddecentage: will do

PR #9249, opened 2025-09-29
feat(frontend): Show token in WalletConnect sign review

AntonioVentilii:
# Motivation

We want to display the token in the WalletConnect sign approval review, IF it is provided in the message.

# Changes

- Extract the token details (if present) from the request, parsing according to the `eth_signTypedData_v4` schema.
- Search the token and its network among the enabled tokens.
- Render the token data, if it is not nullish.

# Tests

Added some tests and a practical example:

Screenshot 2025-09-29 at 14 45 49


roman-nazaruk (approved):
LGTM, thnx

PR #9257, opened 2025-09-29
feat(frontend): AiAssistantShowBalanceToolCard component

DenysKarmazynDFINITY:
# Motivation

We need to implement a component for displaying the show balance tool list item. It should handle 4 different cases: network + token, just token, just network, no network + no token.

Screenshot 2025-09-29 at 16 14 44


AntonioVentilii (approved):
LGTM tks!

PR #9270, opened 2025-09-30
feat(frontend): maps collection description into model

BonomoAlessandro:
# Motivation

Since we are loading the collections metadata from alchemy, we get sometimes a collection description. If so, we want to map it into our data model to be able to display it on the frontend.


daviddecentage (approved):
LGTM, ywbm

PR #9272, opened 2025-09-30
feat(frontend): Transaction time only

daviddecentage:
# Motivation

We want to be able to get `formatSecondsToDate` to return time only. Since `toLocale**Date**String` doesnt support time only we need to pass a flag that switches the usage to `toLocale**Time**String`.

# Changes

Adjusted implementation
Make use of the new flag in the transactions

# Tests

Added tests and ensured that current tests still run without changes to the tests, with the changed implementation.


BonomoAlessandro (approved):
LGTM

PR #9277, opened 2025-09-30
fix(frontend): maps explorerUrl when mapping Network to NftCollectionSchema

BonomoAlessandro:
# Motivation

When mapping a Network to `NftCollectionSchema` we lost the `explorerUrl` definition, but we want to keep it.


daviddecentage (approved):
LGTM

PR #9285, opened 2025-09-30
feat(frontend): Show nfts manage tokens

daviddecentage:
# Motivation

We want to display only Nfts on the manage tokens modal if were on the Nft routes, and only tokens if not.

# Changes

- Extended filtering inside of modal token list context
- Added labels for modal, buttons etc to correctly display "Import Nft collection" etc
- Pass if it should render Nfts or not to the context

# Tests

Added test to ensure the filter prop in the ModalListContext works

image

image

### DIFF
--- a/src/frontend/src/eth/providers/alchemy.providers.ts
+++ b/src/frontend/src/eth/providers/alchemy.providers.ts
@@ -275,6 +275,9 @@ export class AlchemyProvider {
 		return {
 			...(nonNullish(maybeName) && { name: maybeName }),
 			...(nonNullish(result.symbol) && { symbol: result.symbol }),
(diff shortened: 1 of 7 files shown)


src/frontend/src/lib/components/manage/ManageTokens.svelte
```diff
@@ -24,7 +24,11 @@
 	import type { Token } from '$lib/types/token';
 	import { pinEnabledTokensAtTop, sortTokens } from '$lib/utils/tokens.utils';
 
-	let { initialSearch, infoElement }: { initialSearch?: string; infoElement?: Snippet } = $props();
+	let {
```
roman-nazaruk, line 27: Colud you please create interface Props

PR #9290, opened 2025-09-30
feat(frontend): Hide loader modal if derivation is done on frontend

AntonioVentilii:
# Motivation

The frontend derivation is way faster that the backend one. Since that causes a bit of a strange effect for the loader modal (it is almost instantly hidden), we can remove it.

Suggestion by @peterpeterparker .

### Before

https://github.com/user-attachments/assets/cb249f7f-c2e2-46f5-ae7a-afffe7e44f26

### After

https://github.com/user-attachments/assets/79300c0f-c3dc-4437-93f6-0b30b14d6d61


src/frontend/src/lib/components/loaders/Loader.svelte
```diff
@@ -256,7 +257,7 @@
 	});
 </script>
 
-{#if $initialLoading}
+{#if $initialLoading && !FRONTEND_DERIVATION_ENABLED}
```
peterpeterparker, line 260: Not strictly related to your PR, just an idea that crossed my mind: if the constant is suppose to stay in the future you might want to add the alias `FRONTEND_DERIVATION_DISABLED` to improve readability.

peterpeterparker (approved):
Cool! 🥳

LGTM, thx.

That said, we can probably smoothness the loading even a bit more in other PRs.

For example, assuming you updated the resulting video provided in the PR after the first iteration, would removing the fade on line 285 even make the boot more smooth?

src/frontend/src/lib/components/loaders/Loader.svelte
AntonioVentilii: Thankfully it is not intended to stay: it is just a temporary feature flag awaiting the possibility to derive the addresses in the frontend for STAGING and LOCAL using [`ic-pub-key`](https://github.com/dfinity/ic-pub-key).

BETA and PROD already can.

PR #9297, opened 2025-09-30
feat(devops): Increase the coverage thresholds margin

AntonioVentilii:
# Motivation

I was too aggressive in the margin calculation of the coverage thresholds: the calculation has an error margin up to 15bps...


daviddecentage (approved):
LGTM

PR #7409, opened 2025-06-19
feat(frontend): finalizing Pow feature

DecentAgeCoder:
# Motivation

This PR completes the PoW protection system by adding the user interface components and integrating them into the loader hierarchy.

# Changes

- **PowProtector Component**: Comprehensive UI component with modal interface, progress tracking, and cycle checking logic
- **Loader Integration**: Added `PowProtector` to the component hierarchy in to wrap the entire application `Loaders.svelte`
- **User Experience**: Features include:
    - Real-time progress updates during PoW challenges
    - Automatic cycle checking with polling mechanism
    - Graceful timeout handling with user sign-out
    - Responsive modal design with theme support
    - Internationalized progress messages
    
# Tests

Tested with VITE_POW_FEATURE_ENABLED enabled and disabled on fe2

image

PR #9299, opened 2025-09-30
feat(frontend): adding PowProtector component

DecentAgeCoder:
# Motivation

This PR finalizes the PoW protection system by introducing the user interface component.
The complete feature, including all related changes, is tracked in PR #7409.

image

# Changes
- adds PowProtector modal
- adds support for dynamic progress steps
- includes localization updates for multiple languages.

# Tests
- Introduced `PowProtector.spec.ts `to test `PowProtector.svelte`


src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,216 @@
...
 	}
 
 	let { children }: Props = $props();
 
-	if (POW_FEATURE_ENABLED) {
-		let powWorker: PowProtectorWorkerInitResult;
+	// Use let with $state() for variables that need to be reassigned
+	let hasCycles = $state(false);
+	let loading = $state(false);
+	let checkInterval: ReturnType<typeof setInterval> | undefined;
```
AntonioVentilii:
```suggestion
	let checkInterval = $state<NodeJS.Timeout | undefined>();
```

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,216 @@
...
+	let checkAttempts = $state(0);
+	let powWorker: PowProtectorWorkerInitResult | undefined;
```
AntonioVentilii:
```suggestion
	let powWorker= $state<PowProtectorWorkerInitResult | undefined>();
```

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,216 @@
...
 
-		onMount(async () => {
-			// Initialize the worker
-			powWorker = await initPowProtectorWorker();
-			// Start the worker
-			powWorker.start();
-		});
+	const MAX_CHECK_ATTEMPTS = 100;
+	const CHECK_INTERVAL_MS = 7000;
```
AntonioVentilii: shall we put this in a separate module in the `constants` folder?

src/frontend/src/lib/services/loader.services.ts
```diff
@@ -34,7 +34,7 @@ import { get } from 'svelte/store';
  * @returns {Promise<boolean>} A promise resolving to `true` if the required cycles are met or exceeded,
  * otherwise `false` if insufficient cycles are detected or an error occurs during processing.
  */
-export const handleInsufficientCycles = async (): Promise<boolean> => {
+export const hasEnoughCycles = async (): Promise<boolean> => {
```
AntonioVentilii: let's do this refactor in another PR first

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,216 @@
...
 
 	interface Props {
 		children?: Snippet;
 	}
 
 	let { children }: Props = $props();
 
-	if (POW_FEATURE_ENABLED) {
-		let powWorker: PowProtectorWorkerInitResult;
+	// Use let with $state() for variables that need to be reassigned
```
AntonioVentilii, line 26:
```suggestion
```

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,216 @@
...
+			step: ProgressStepsPowProtectorLoader.SOLVE_CHALLENGE,
+			text: $i18n.pow_protector.text.solve_challenge,
+			state: 'completed'
+		} as ProgressStep,
+		{
+			step: ProgressStepsPowProtectorLoader.GRANT_CYCLES,
+			text: $i18n.pow_protector.text.grant_cycles,
+			state: 'completed'
+		} as ProgressStep
+	] as NonEmptyArray<ProgressStep | StaticStep>);
```
AntonioVentilii: typically we put the steps structures in a separate file in the `config` folder (see `addressBookWizardSteps` for example)

AntonioVentilii (commented):
I did not finish the review: let's first split this PR into smaller scopes:

- create a component with the basic structure: `onMount` and `onDestroy`
- add the modal without the `InProgress` component
- add the InProgress
- enable it ONLY for the feature flag

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: Nice. Thanks

src/frontend/src/lib/services/loader.services.ts
DecentAgeCoder: I have reverted the  renaming of the handleInsufficientCycles(). It will be renamed in the follow-up PR.

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: I have put them to the existing `pow.constants.ts` file and post-fixed them with 'POW_'

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: I have moved the steps to src/frontend/src/lib/config/pow.config.ts

src/frontend/src/lib/components/pow/PowProtector.svelte
AntonioVentilii: this is not resolved

src/frontend/src/lib/components/pow/PowProtector.svelte
AntonioVentilii: this was not resolved

src/frontend/src/lib/components/pow/PowProtector.svelte
AntonioVentilii: this was not resolved

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,197 @@
...
+			try {
+				// Initial check
+				hasCycles = await handleInsufficientCycles();
+			} catch (error) {
+				// Log error but continue (assume no cycles on error)
+				console.error('Error checking initial cycles:', error);
+				hasCycles = false;
+			}
+
+			loading = true;
```
AntonioVentilii: shouldn't `loading = true` be on top of the `onMount`? and where is it set to false again?

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,197 @@
...
+				clearInterval(checkInterval);
+				checkInterval = undefined;
+			}
+
+			// Stop the worker if it was started
+			if (powWorker) {
+				powWorker.destroy();
+			}
+		}
+	});
```
AntonioVentilii, line 150: instead of clearing the interval in `onDestroy`, return it as function to `onMount`. For example, look at `IntervalLoader`

BTW, I think you could use `IntervalLoader` directly, no?

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,197 @@
...
-{@render children?.()}
+{#if !POW_FEATURE_ENABLED}
+	<!-- POW feature is globally disabled. So we bypass all protection logic and render the app content directly -->
+	{@render children?.()}
+{:else if loading}
+	<!-- POW feature is enabled and component has started loading -->
+	{#if hasCycles}
+		<!-- User has sufficient cycles so, user can proceed normally -->
+		{@render children?.()}
+	{:else}
```
AntonioVentilii: what happens if it is not loading but it has Cycles?

AntonioVentilii (commented):
Can you please put a video of the entire experience?

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,197 @@
...
+		<!-- 
+			User lacks sufficient cycles for POW, so we display modal with progress indicator while cycles are being obtained
+			This modal will be displayed until either:
+			- User obtains sufficient cycles (checkCycles polling succeeds)
+			- Maximum retry attempts reached (user gets signed out)
+		-->
+		<div class="insufficient-cycles-modal">
```
AntonioVentilii, line 161: why a specific class for this modal? is it different from the default?

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,197 @@
...
+			<Modal testId="pow-protector-modal">
```
AntonioVentilii: let's put the test id in the test id module as constant

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,197 @@
...
+
+					<h3 class="my-3">{$i18n.pow_protector.text.title}</h3>
+					<p class="mt-3">{$i18n.pow_protector.text.description}</p>
+
+					<InProgress {progressStep} {steps} />
+				</div>
+			</Modal>
+		</div>
+	{/if}
+{/if}
```
AntonioVentilii, line 190: what happens if the user is in the middle of a transfer (or any other modal-related situation) and it runs out of cycles? Meaning, what happens if there is a modal already open? Or if there are processes running? Does this process at some point risk to log out the user?

src/frontend/src/lib/config/pow.config.ts
```diff
@@ -0,0 +1,24 @@
+import { ProgressStepsPowProtectorLoader } from '$lib/enums/progress-steps';
+import type { StaticStep, WizardStepsParams } from '$lib/types/steps';
+import type { NonEmptyArray } from '$lib/types/utils';
+import type { ProgressStep } from '@dfinity/gix-components';
+
+export const powProtectorSteps = ({
+	i18n
+}: WizardStepsParams): NonEmptyArray<ProgressStep | StaticStep> => [
```
AntonioVentilii:
```suggestion
}: WizardStepsParams): NonEmptyArray<ProgressStep> => [
```

src/frontend/src/lib/config/pow.config.ts
```diff
@@ -0,0 +1,24 @@
...
+	{
+		step: ProgressStepsPowProtectorLoader.REQUEST_CHALLENGE,
+		text: i18n.pow_protector.text.request_challenge,
+		state: 'completed'
+	} as ProgressStep,
```
AntonioVentilii: is the casting required?

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,197 @@
...
+			switch (progressStoreData.progress) {
+				case 'REQUEST_CHALLENGE':
+					progressStep = ProgressStepsPowProtectorLoader.REQUEST_CHALLENGE;
+					break;
+				case 'SOLVE_CHALLENGE':
+					progressStep = ProgressStepsPowProtectorLoader.SOLVE_CHALLENGE;
+					break;
+				case 'GRANT_CYCLES':
+					progressStep = ProgressStepsPowProtectorLoader.GRANT_CYCLES;
+					break;
```
AntonioVentilii, line 48: these cases should be from the enum of `ProgressStepsPowProtectorLoader` right?

`case ProgressStepsPowProtectorLoader.REQUEST_CHALLENGE.....`

src/frontend/src/lib/components/pow/PowProtector.svelte
```diff
@@ -1,27 +1,197 @@
...
+	let powWorker: PowProtectorWorkerInitResult | undefined;
 
-		onMount(async () => {
-			// Initialize the worker
-			powWorker = await initPowProtectorWorker();
-			// Start the worker
-			powWorker.start();
-		});
+	// Initialize with default value, but it will be reactively updated from the store
+	let progressStep = $state(ProgressStepsPowProtectorLoader.REQUEST_CHALLENGE);
```
AntonioVentilii, line 33: i am thinking that maybe it would be better to initialize it as undefined, and then set it to `ProgressStepsPowProtectorLoader.REQUEST_CHALLENGE` when you need to request the challenge, or not?

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: Now prioritizing hasCycles condition over loading state to ensure children always render when user has sufficient cycles.

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: Using POW_PROTECTOR_MODAL variable from test-ids.constants now

src/frontend/src/lib/config/pow.config.ts
DecentAgeCoder: I have removed the redundant cast

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: I have removed the comment

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: I have cleaned-up and simplified the code. The loading variable is not used anymore and has therfore been removed.

src/frontend/src/lib/config/pow.config.ts
DecentAgeCoder: I have changed the the type definition as suggested

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: I have refactored `PowProtector` to replace manual interval checks with `IntervalLoader`

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: The store returns string literal types, so we need to compare against those string literals in the switch cases

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: This class uses the same custom styling as the Loader modal (max-width: 90vw, custom border-radius, etc. I could extract this to a shared CSS class or use the Modal's default styling if these customizations aren't necessary. What do you think?

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: This class uses the same custom styling as the Loader modal (max-width: 90vw, custom border-radius, etc. I could extract this to a shared CSS class or use the Modal's default styling if these customizations aren't necessary. What do you think?

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: Good suggestion! I have changed `progressStep` to initialize as `undefined`. Now it only has a value when the store provides actual progress data, avoiding false UI states.

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: The UI is designed to appear only as a last resort. Under normal circumstances, new cycles are obtained in the background as soon as the allowance balance falls below a minimum threshold > POW_MIN_CYCLES_THRESHOLD = 200B cycles). This process is completely transparent to the use, they don't even notice it's happening.
However, if the user's allowance balance somehow drops even lower, which will almost never happen ( < POW_ZERO_CYCLES_THRESHOLD 50B cycles) , approaching zero , the PoW Modal is displayed. This modal shows the progress of the current PoW process and blocks all other activities that could potentially consume cycles. This is why the PowProtector uses a hard switch to display the modal rather than an overlay, it needs to enforce that no cycle-consuming operations can be triggered until the balance is restored ot avoid any errors. 
Even if the user has zero cycles when they log in, the PoWProtector will block all other activities by displaying the UI until the PoW challenge is successfully requested, completed, and cycles are granted. Importantly, the user is not logged out during this process, the modal remains active until the cycle balance is restored.

src/frontend/src/lib/components/pow/PowProtector.svelte
DecentAgeCoder: Has been resolved now

src/frontend/src/lib/components/pow/PowProtector.svelte
AntonioVentilii: eh, the store should use the enum too...

src/frontend/src/lib/components/pow/PowProtector.svelte
AntonioVentilii: let's create tests for all hedge cases; typically are the ones that require to release an urgent hotfix

src/frontend/src/lib/components/pow/PowProtector.svelte
AntonioVentilii: why a different styling? shouldn't we be consistent with the style through the UI? May you please double-confirm with design?

src/frontend/src/lib/components/pow/PowProtector.svelte
AntonioVentilii: is there a test for all the cases?

PR #9310, opened 2025-10-01
feat(backend): finalizing backend before enablement

DecentAgeCoder:
# Motivation

This PR finalizes the PoW protection system on the backend side.
The complete feature, including all related changes, is tracked in PR https://github.com/dfinity/oisy-wallet/pull/7409.

# Changes

- When auto-adjustment is disabled, the `START_DIFFICULTY` is used to initialize the difficulty (instead of taking the adjusted difficulty from the previous challenge).
- Updated comments for better explanation of logic and added examples.

**asserts.rs:**
- Replaced panic-based underflow handling with a safer saturating subtraction approach to prevent runtime errors.
- Simplified overflow handling by removing redundant checks and relying on direct computations.
- Improved error reporting.

# Tests


src/backend/src/pow.rs
```diff
@@ -97,6 +97,7 @@ pub async fn create_pow_challenge() -> Result<StoredChallenge, CreateChallengeEr
         return Err(CreateChallengeError::MissingUserProfile);
     }
 
+    // Retrieve or initialize new challenge
```
AntonioVentilii:
```suggestion
```

PR #9312, opened 2025-10-01
style(frontend): add text highlighting to the AI console messages

DenysKarmazynDFINITY:
# Motivation

Currently, the selected/highlighted text that is placed inside an element with "primary" background color is not really visible (happens in most of the browsers). It is accepted in most of the cases, but in the AI console it would be really great to allow users to do so. The solution is to add a custom styling for the highlighted text.

Before:
Screenshot 2025-10-01 at 12 48 01

After:
Screenshot 2025-10-01 at 12 44 39


AntonioVentilii (approved):
Tks!

P.S. shall this be a global setting for the styles? I have to imagine it happens in the rest of the code too, no?

PR #9314, opened 2025-10-01
refactor(frontend): Spin-off `LoaderTokens` from `Loader` component

AntonioVentilii:
# Motivation

To improve the manageability and reduce the scope of each component, we spin-off the part of `Loader` that takes care of the tokens only: it will be called `LoaderTokens`.

As per design, the two component are independent basically, but, as the rest, `LoaderTokens` must be a children of `Loader`.


roman-nazaruk (approved):
LGTM, thnx

PR #9317, opened 2025-10-01
refactor(frontend): Extract service to accept user agreements

AntonioVentilii:
# Motivation

Since we are going to reuse it, we extract the service to accept the user agreements.


roman-nazaruk (approved):
LGTM, thnx

PR #9320, opened 2025-10-01
feat(frontend): Display approve spender

daviddecentage:
# Motivation

Based on #9280.

We want to display the spender account for Icp/Icrc transactions.

# Changes

- Refactored mappings for transactions
- Adjusted fee and amount definitions so it is consistent across the different transaction types
- Display spender for approve transactions in modal and transaction card

# Tests

All new additions should be tested fully:

- Transaction cards tested by #9346 which ensures that transactions of the other networks are not affected by the changes
- Displaying spender in contact cards tested by tests in #9342
- Added tests for ensuring transaction card label is correct including allowance for approve
- Extended tests to ensure contacts are displayed in transaction cards and that it shows the aprove spender contact for approve transactions

Manually tested as well on FE3


src/frontend/src/lib/components/transactions/Transaction.svelte
```diff
@@ -65,8 +65,14 @@
...
-	const amountWithFee = $derived(
-		nonNullish(cardAmount) && nonNullish(fee) && !incoming ? cardAmount + fee : cardAmount
+	const displayAmount = $derived(
+		nonNullish(cardAmount)
+			? type === 'approve' && nonNullish(fee)
+				? fee * -1n
+				: !incoming && nonNullish(fee)
+					? cardAmount + fee * -1n
+					: cardAmount
+			: undefined
```
AntonioVentilii: let's maintain the old logic: `incoming` is calculated based on the `type` already, and when reading the component, one should read "if it is NOT an incoming transaction, the fee should be added"

daviddecentage: For approve we want to display the fee instead of the approved amount. It needs to be treated as a seperate case as `!incoming` includes more than just approved

src/frontend/src/lib/components/transactions/Transaction.svelte
```diff
@@ -182,6 +203,8 @@
 							<span class="shrink-0">{$i18n.transaction.text.to}</span>
 						{:else if type === 'receive'}
 							<span class="shrink-0">{$i18n.transaction.text.from}</span>
+						{:else if type === 'approve'}
```
AntonioVentilii: let's do this in a separate PR, with some screenshot

src/frontend/src/lib/components/transactions/Transaction.svelte
```diff
@@ -109,10 +115,25 @@
...
-					{type === 'send' ? $i18n.transaction.type.send : $i18n.transaction.type.receive}
+					{type === 'send'
+						? $i18n.transaction.type.send
+						: type === 'approve'
+							? $i18n.transaction.type.approve
+							: $i18n.transaction.type.receive}
 				{:else}
 					{@render children?.()}
 				{/if}
+				{#if type === 'approve' && nonNullish(cardAmount)}
```
AntonioVentilii: let's add it in another PR (maybe after this one)

src/frontend/src/icp/components/transactions/IcTransactionModal.svelte
```diff
@@ -139,6 +142,17 @@
 					</output>
 				</ListItem>
 			{/if}
+			{#if nonNullish(approveExpiresAt)}
```
AntonioVentilii: let's add it in another PR

src/frontend/src/lib/components/transactions/Transaction.svelte
daviddecentage: Okay, removed

src/frontend/src/lib/components/transactions/Transaction.svelte
daviddecentage: okay, removed

src/frontend/src/icp/components/transactions/IcTransactionModal.svelte
daviddecentage: okay, removed

src/frontend/src/lib/components/transactions/Transaction.svelte
AntonioVentilii: correct, but the logic should not be in this component: the logic of this component is:"if NOT an incoming, add the fee"

And incoming consider really all the "incoming transactions, so all good for approve, since it is not incoming.

Now, `amount` and `fee` should set accordingly elsewhere for each transaction

src/frontend/src/lib/components/transactions/Transaction.svelte
```diff
@@ -100,11 +100,7 @@
 			<span
 				class="relative inline-flex items-center gap-1 whitespace-nowrap first-letter:capitalize"
 			>
-				{#if nonNullish(contact)}
-					{type === 'send' ? $i18n.transaction.type.send : $i18n.transaction.type.receive}
-				{:else}
-					{@render children?.()}
-				{/if}
+				{@render children?.()}
```
AntonioVentilii, line 103: is this change expected? we are not showing the contact anymore?

AntonioVentilii (approved):

PR #9322, opened 2025-10-01
feat(frontend): Automatically accepts agreements if never visioned

AntonioVentilii:
# Motivation

There will be a shift on how we manage the agreements and how we want the user to accept it.

The idea is be "passive":

- the login make the user automatically/implicitly accept all agreements;
- in case of outdated agreements, we additionally show a banner to make the user acknowledge the changed agreements.

In this PR we treat first the case where the agreements were never visioned (meaning that the user never logged in after we introduced the agreements).

# Changes

- Create derived store for a list of agreements that need to be accepted.
- In case of no agreements yet visioned, do not show the modal, but render the children.
- Reactively, if no agreement was ever visioned, accept all of them.

# Tests

Created tests for all new logic.

### DIFF
--- a/src/frontend/src/lib/derived/user-agreements.derived.ts
+++ b/src/frontend/src/lib/derived/user-agreements.derived.ts
@@ -78,13 +78,14 @@ export const outdatedAgreements: Readable<Partial<UserAgreements>> = derived(
 
 export const agreementsToAccept: Readable<AgreementsToAccept> = derived(
 	[outdatedAgreements],
-	([$outdatedAgreements]) => {
-		const result: AgreementsToAccept = {};
-		Object.keys($outdatedAgreements).forEach((agreementType) => {
-			result[agreementType as keyof EnvAgreements] = true;
-		});
-		return result;
-	}
+	([$outdatedAgreements]) =>
+		Object.keys($outdatedAgreements).reduce<AgreementsToAccept>(
+			(acc, agreementType) => ({
+				...acc,
+				[agreementType as keyof EnvAgreements]: true
+			}),
+			{}
+		)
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/derived/user-agreements.derived.ts
```diff
@@ -75,6 +76,17 @@ export const outdatedAgreements: Readable<Partial<UserAgreements>> = derived(
 		)
 );
 
+export const agreementsToAccept: Readable<AgreementsToAccept> = derived(
+	[outdatedAgreements],
+	([$outdatedAgreements]) => {
```
DenysKarmazynDFINITY: We could probably shorten it a bit if we directly return the result of reduce: 
```
export const agreementsToAccept: Readable<AgreementsToAccept> = derived(
	[outdatedAgreements],
	([$outdatedAgreements]) => Object.keys($outdatedAgreements).reduce(...)
);
```

AntonioVentilii: That was my first iteration, but then I read this comment from Copilot

https://github.com/dfinity/oisy-wallet/pull/9322#discussion_r2395497427

And made me wonder... WDYT?

I still prefer `reduce`, BTW!

DenysKarmazynDFINITY: Oh, I see - in our case, the N number will be really low (3?), so almost no difference in terms of performance, but we can still stick to the for-of pattern.

AntonioVentilii: changed back!

PR #9347, opened 2025-10-02
fix(frontend): Remove concurrency of accepting agreements

AntonioVentilii:
# Motivation

When mounting the component `AgreementsGuard`, the user is auto-accepting the agreements.

Since it depends on derived stores of the user profile, it could be called several times consecutively. It is not an issue, but we can spare calls.

The issue is that accept `acceptAgreements` is asynchronous, but the rune `$effect` does not accept `async` function, so it will not await `acceptAgreements` to finish. So, even using a flag to avoid duplicate calls is not that easy: the flag could be set back to `false` BEFORE `acceptAgreements` is finished.

However, we have a very useful service called `queryAndUpdate` that executes canisters calls (query and/or update) and has callbacks in case of success or failure.

# Changes

- Use `queryAndUpdate` to run `acceptAgreements`:
  - only update call;
  - set the accepting flag to false in case both of success or error;
  - avoid running `acceptAgreements` if it is already running (`accepting === true`).
- For good measure, we avoid running it completely if there are no agreements to accept.

# Tests

Created a new test that was failing before this PR:

Screenshot 2025-10-02 at 14 23 55

### DIFF
--- a/src/frontend/src/lib/components/guard/AgreementsGuard.svelte
+++ b/src/frontend/src/lib/components/guard/AgreementsGuard.svelte
@@ -38,10 +38,10 @@
 				return;
 			}
 
-            // The $effect rune cannot await async functions (in this case acceptAgreements).
-            // That causes the `accepting` flag to be set to `false` before the async function is finished.
-            // To avoid it, we use the service queryAndUpdate,
-            // which handles the state of the request internally and has callbacks in case of awaited success or failure.
+			// The $effect rune cannot await async functions (in this case acceptAgreements).
+			// That causes the `accepting` flag to be set to `false` before the async function is finished.
+			// To avoid it, we use the service queryAndUpdate,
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/guard/AgreementsGuard.svelte
```diff
@@ -18,16 +19,49 @@
 
 	let { children }: Props = $props();
 
+	let accepting = false;
```
roman-nazaruk, line 22: maybe $state?

AntonioVentilii: the issue is that if i put it as `$state` and change it in the `$effect`, it will re-trigger `$effect` (infinite loop). I tested it as first iteration, and that was the issue

PR #9351, opened 2025-10-02
feat(backend)!: Save SHA256 of the agreements files

AntonioVentilii:
# Motivation

We add a field to store the SHA256 in the agreements structure in the backend: it will be useful to match the versions.

BREAKING CHANGE: adding a new field to the type `Agreements` changes the interface.

### DIFF
--- a/src/frontend/src/tests/lib/components/guard/AgreementsGuard.spec.ts
+++ b/src/frontend/src/tests/lib/components/guard/AgreementsGuard.spec.ts
@@ -148,7 +148,9 @@ describe('AgreementsGuard', () => {
 							accepted: toNullable(false),
 							last_accepted_at_ns: toNullable(1677628800n),
 							last_updated_at_ms: toNullable(1677542400n),
-							text_sha256: toNullable('3c1f186b9348d5cdf17d438f98eb29909bb6bcb74f6f3b554b4abd4942064424')
+							text_sha256: toNullable(
+								'3c1f186b9348d5cdf17d438f98eb29909bb6bcb74f6f3b554b4abd4942064424'
+							)
 						}
 					}
 				})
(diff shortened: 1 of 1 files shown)


src/shared/src/types/agreement.rs
```diff
@@ -12,6 +17,23 @@ pub struct UserAgreement {
...
     pub last_updated_at_ms: Option<Timestamp>,
+    /// SHA256 hash of the agreement text, to detect changes.
+    pub text_sha256: Option<String>,
+}
+impl Validate for UserAgreement {
+    /// Verifies that agreement text SHA256 is a valid hex of expected length, if provided.
+    fn validate(&self) -> Result<(), candid::Error> {
+        if let Some(ref hash) = self.text_sha256 {
+            if hash.len() != SHA256_HEX_LENGTH {
+                return Err(candid::Error::msg(format!(
```
DenysKarmazynDFINITY, line 28: Just for my understanding - is there any reason for returning here the error as a string instead of an error type (we already have a few inside `UpdateAgreementsError`)? In the other similar places, it's sometime the former option, sometime the latter.

AntonioVentilii: No real reason to be honest: I am following what we are doing in other methods when we treat validation for "type". For example the length of some input, or the min/max number that an input can reach.

If you prefer I can use a specific error variant: I agree it is always better eheh!

DenysKarmazynDFINITY: Thanks for the explanations! I'm fine with keeping it as it is - we indeed use the string approach in most of the cases when it comes to validating. Maybe at some point we can align all similar places to use a single approach, but it's of course just cosmetics, so no rush at all.

PR #9362, opened 2025-10-03
feat(frontend): Use util `assertNever` from`@dfinity/utils`

AntonioVentilii:
# Motivation

With @peterpeterparker 's PR https://github.com/dfinity/ic-js/pull/1081, a new util `assertNever` was released:

> /**
>  * Utility to enforce exhaustiveness checks in TypeScript.
>  *
>  * This function should only be called in branches of a `switch` or conditional
>  * that should be unreachable if the union type has been fully handled.
>  *
>  * By typing the parameter as `never`, the compiler will emit an error if
>  * a new variant is added to the union but not covered in the logic.
>  *
>  * @param _ - A value that should be of type `never`. If this is not the case,
>  *            the TypeScript compiler will flag a type error.
>  * @param message - Optional custom error message to include in the thrown error.
>  * @throws {Error} Always throws when invoked at runtime.
>  *
>  */

We can use it in a few places in the codebase, since we are applying a similar logic.

# Changes

- Add the usage of `assertNever` from `@dfinity/utils` in all places with `const ...: never = ...`.
- Replace existing `assertNever` with the one from the library.

# Tests

Current tests work as expected.

### DIFF
--- a/src/frontend/src/btc/utils/btc-address.utils.ts
+++ b/src/frontend/src/btc/utils/btc-address.utils.ts
@@ -5,13 +5,13 @@ import {
 	btcAddressTestnet
 } from '$lib/derived/address.derived';
 import type { NetworkId } from '$lib/types/network';
-import { assertNever } from '$lib/types/utils';
 import { isNetworkIdBTCRegtest, isNetworkIdBTCTestnet } from '$lib/utils/network.utils';
 import {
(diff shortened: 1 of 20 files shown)


peterpeterparker (approved):
Nice, LGTM; thx!

PR #9367, opened 2025-10-03
feat(frontend): Confirmation modal

daviddecentage:
# Motivation

We need a confirmation modal for an upcoming feature.

# Changes

- Adds Trigger with Modal component
- Adjusted gix variables in SCSS to correctly style paddings for the modal in role "alert"

# Tests

Tests added covering the functionality

image


src/frontend/src/lib/components/ui/ConfirmButtonWithModal.svelte
```diff
@@ -0,0 +1,45 @@
...
+	{#snippet title()}
+		<div class="p-3">{@render innerTitle()}</div>
+	{/snippet}
+
+	{@render children()}
+
+	{#snippet footer()}
+		<div class="my-3 flex w-full justify-between gap-3">
+			<ButtonCancel onclick={onCancelHandler} testId={`${testId}-cancel`} />
+			<Button onclick={onConfirmHandler} testId={`${testId}-confirm`}
```
AntonioVentilii: i would create the test ID for each button, but not strictly necessary, your call

src/frontend/src/lib/styles/global/gix.scss
```diff
@@ -71,8 +71,8 @@
 	--alert-max-width: 100vw;
 	--alert-max-height: 100dvh;
 	--alert-border-radius: var(--border-radius-xs);
-	--alert-padding-y: 0;
-	--alert-padding-x: 0;
+	--alert-padding-y: var(--padding-2x);
+	--alert-padding-x: var(--padding-3x);
```
AntonioVentilii, line 75: doesn't this impact other components?

src/frontend/src/lib/styles/global/gix.scss
```diff
@@ -95,6 +95,7 @@
 		--dialog-width: 560px;
 		--alert-max-width: calc(100vw - var(--padding-4x));
 		--alert-max-height: initial;
+		--alert-border-radius: calc(var(--border-radius-sm) * 3);
```
AntonioVentilii, line 98: same question

src/frontend/src/lib/styles/global/gix.scss
daviddecentage: Nope, it was unused

src/frontend/src/lib/styles/global/gix.scss
daviddecentage: Nope, it was unused

src/frontend/src/lib/components/ui/ConfirmButtonWithModal.svelte
daviddecentage: Ive seen it usually done this way I think its quite handy having it to pass once.

src/frontend/src/lib/styles/global/gix.scss
AntonioVentilii: they are both used by GIX components (hence the `gix.scss` module); they are used intrinsically in the GIX components that we import. Are you sure it does not affect any other component?

AntonioVentilii: Just as example, it is used by all the modals and wizard modals. Is the change intended for all of them?

daviddecentage: All other modals we have in OISY either use `role="dialog"` or no role at all which defaults to `"dialog"`. The variables I adjusted target only the `role="alert"` styles which havent been used until this PR. I checked also gix and oisy repo for usages of it and its really just used for `role="alert"`.

The exception is for `--dialog-max-width/height` which we dont touch in my change here and  `--dialog-border-radius` which uses the `--alert-border-radius` but we override it anyways and to the same value.

AntonioVentilii: thank you for checking

AntonioVentilii (approved):

PR #9369, opened 2025-10-03
feat(backend): enable PoW with low protection

DecentAgeCoder:
# Motivation

This PR contains the necessary changes and information to enable the PoW protection for granting cycles to user account used for signing operations.

**Important:**
In order to test this PR on a fe environment the variable VITE_POW_FEATURE_ENABLED has to be set (see: "Enabling Pow feature")

**This PR requires the following PR's to be approved:**
- [9302](https://github.com/dfinity/oisy-wallet/pull/9302)
- [9310](https://github.com/dfinity/oisy-wallet/pull/9310)

The configuration in this PR:
- will grant **0.6 T cycles** to the user when the challenge is solved
- new cycles are obtained through the Pow process when the available cycles fall below the threshold **0.2 T cycles**
- the modal should never show-up (to test it, read the "Configuration rules" below)
- the challenge has to be solved within **2 minutes** and a new one can be obtained after this time window

### How does the PoW behave:
The PoW UI  modal is designed to appear only as a last resort. Under normal circumstances, new cycles are obtained in the background as soon as the allowance balance falls below a minimum threshold (`POW_MIN_CYCLES_THRESHOLD`). This process is completely transparent to the user.

However, if the user's allowance balance somehow drops even lower, which will almost never happen ( < `POW_ZERO_CYCLES_THRESHOLD` ) , approaching zero , the PoW Modal is displayed. This modal shows the progress of the current PoW process and blocks all other activities that could potentially consume cycles. This is why the PowProtector uses a hard switch to display the modal rather than an overlay, it needs to enforce that no cycle-consuming operations can be triggered until the balance is restored to avoid any errors.

Even if the user has zero cycles when they log-in, the PoWProtector will block all other activities by displaying the UI until the PoW challenge is successfully requested, completed, and cycles are granted. Importantly, the user is not logged out during this process, the modal remains active until the cycle balance is restored.

### Configuring the Pow protection feature

Relevant source files incl. documentation can be found here:
- [pow.rs](https://github.com/dfinity/oisy-wallet/blob/feat/frontend/pow/protect-allow-signing-7-8-4/src/shared/src/types/pow.rs) - backend settings
- [pow.constants.ts](https://github.com/dfinity/oisy-wallet/blob/feat/frontend/pow/protect-allow-signing-7-8-4/src/frontend/src/lib/constants/pow.constants.ts) - frontend settings

**Enabling Pow feature:**
- Set constant POW_ENABLED in **backend** code `src/shared/src/types/pow.rs to true
- Set the environment variable `VITE_POW_FEATURE_ENABLED` for the **frontend** to true. **When deploying through github the following value has to be added to input field "Optional .env content": VITE_POW_FEATURE_ENABLED=true**

_The backend and frontend enablement flags must both be set to the same value (either both true or both false)._

**Configuration rules:**
- Set the value `POW_MIN_CYCLES_THRESHOLD` in `pow.constants.ts` to the product of  `START_DIFFICULTY` x `CYCLES_PER_DIFFICULTY` which are defined in `pow.rs`
- In order to explicitly  test the modal UI the  `POW_ZERO_CYCLES_THRESHOLD` can temp. be set to the same values as `POW_MIN_CYCLES_THRESHOLD`. For more details, see "How does the PoW modal behave" above)

** Debugging:**
The PR [9373](https://github.com/dfinity/oisy-wallet/pull/9373) contains the changes that can be used for testing purposes


AntonioVentilii:
Depends on #9302 and #9310, which are no longer open; the PoW configuration approach has since changed. Closing — reopen on top of current PoW code if still needed.

PR #9385, opened 2025-10-03
refactor(frontend): Extract types of Token UI to separate modules

AntonioVentilii:
# Motivation

To avoid circular reference, we extract the types for Token UI and Token Group UI to separate modules.

### DIFF
--- a/src/frontend/src/lib/types/token-card.ts
+++ b/src/frontend/src/lib/types/token-card.ts
@@ -1,5 +1,4 @@
 import type { Network } from '$lib/types/network';
-
 import type { TokenUi } from '$lib/types/token-ui';
 
 export type CardData = Pick<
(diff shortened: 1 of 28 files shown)


src/frontend/src/lib/types/token-card.ts
```diff
@@ -1,5 +1,6 @@
 import type { Network } from '$lib/types/network';
-import type { TokenUi } from '$lib/types/token';
+
```
AntonioVentilii:
```suggestion
```

src/frontend/src/tests/lib/components/tokens/TokensDisplayHandlerTest.svelte
```diff
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import TokensDisplayHandler from '$lib/components/tokens/TokensDisplayHandler.svelte';
-	import type { TokenUiOrGroupUi } from '$lib/types/token-group';
+
```
AntonioVentilii:
```suggestion
```

PR #9408, opened 2025-10-06
feat(dependabot): Let it bump lint packages

AntonioVentilii:
# Motivation

We will allow dependabot to bump the lint packages, to keep up with current updates.


AntonioVentilii:
Overridden by:

- https://github.com/dfinity/oisy-wallet/pull/9409

PR #9434, opened 2025-10-06
refactor(frontend): move GldtStaking component

DenysKarmazynDFINITY:
# Motivation

The idea is to move the GldtStaking component to `$icp/components/gldt-staking` to maintain the structure that we have for sending/converting/etc:
- All specific to the GLDT staking components will go into `$icp/components/gldt-staking` 
- All general to the staking feature components (wizards, forms, etc.) will go into `$lib/components/staking`


AntonioVentilii (approved):
LGTM!

As suggestion i would even nest them more:

`icp/components/staking/gldt/`

So that every staking component in IC is under the same roof

PR #9436, opened 2025-10-06
refactor(frontend): Move WalletConnect method inside `EthWalletConnectMessage`

AntonioVentilii:
# Motivation

To keep everything in one place, we move the method of WalletConnect inside the component `EthWalletConnectMessage`.

### DIFF
--- a/src/frontend/src/eth/components/wallet-connect/EthWalletConnectMessage.svelte
+++ b/src/frontend/src/eth/components/wallet-connect/EthWalletConnectMessage.svelte
@@ -20,6 +20,8 @@
 
 	let { request }: Props = $props();
 
+	let method = $derived(request.params.request.method);
+
 	let json = $derived.by(() => {
 		try {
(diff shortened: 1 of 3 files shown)


src/frontend/src/eth/components/wallet-connect/EthWalletConnectMessage.svelte
```diff
@@ -98,6 +100,9 @@
 	});
 </script>
 
+<p class="mb-0 font-bold">{$i18n.wallet_connect.text.method}:</p>
```
DenysKarmazynDFINITY: shouldn't it be `mb-0.5` like in all other entries in this component?

AntonioVentilii: it was the original one but yes, makes sense, changed!

At some point I will use a separate component, for maneagability

PR #9441, opened 2025-10-07
feat(frontend): added loadAllIcrcTokensWithSupportedStandards

roman-nazaruk:
# Motivation

We need a new service that accepts a list of ICRC tokens and returns those tokens with an additional property, isIcrc2.

# Changes

Added a new loadAllIcrcTokensWithSupportedStandards service method.

# Tests

Added test coverage for the new service method.

### DIFF
--- a/src/frontend/src/lib/services/swap.services.ts
+++ b/src/frontend/src/lib/services/swap.services.ts
@@ -10,7 +10,7 @@ import { sendIcp, sendIcrc } from '$icp/services/ic-send.services';
 import { loadCustomTokens } from '$icp/services/icrc.services';
 import type { IcToken } from '$icp/types/ic-token';
 import { nowInBigIntNanoSeconds } from '$icp/utils/date.utils';
-import { isTokenIcrc } from '$icp/utils/icrc.utils';
+import { isIcrcTokenSupportIcrc2, isTokenIcrc } from '$icp/utils/icrc.utils';
 import { setCustomToken } from '$lib/api/backend.api';
 import { getPoolCanister } from '$lib/api/icp-swap-factory.api';
 import {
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -190,6 +191,34 @@ export const loadKongSwapTokens = async ({
...
+			const isIcrc2 = await isIcrcTokenSupportIcrc2({
+				identity,
+				ledgerCanisterId: token.ledgerCanisterId
+			});
+
+			return { ...token, isIcrc2 };
+		})
+	);
+
+	const supportedTokens = tokens.reduce<Array<IcToken & { isIcrc2: boolean }>>((acc, result) => {
```
AntonioVentilii: wasn't there a type for this token? `SwappableToken`?

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -190,6 +191,34 @@ export const loadKongSwapTokens = async ({
...
 
+export const loadAllIcrcTokensWithSupportedStandards = async ({
+	allTokens,
+	identity
+}: {
+	allTokens: IcToken[];
+	identity: Identity;
+}): Promise<void> => {
+	const tokens = await Promise.allSettled(
+		allTokens.map(async (token: IcToken) => {
```
AntonioVentilii: and here let's put the return type

`async (token: IcToken): SwappableToken` maybe?

AntonioVentilii (approved):
LGTM tks!

Some obs

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Done.

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Added new IcToken schema to reuse in the future

PR #9446, opened 2025-10-07
feat(frontend): updated value of isSourceTokenIsIcrc2 in swap store

roman-nazaruk:
# Motivation

Since we now have the swap-icrc-tokens-store as a fallback for swap tokens, we need to update the swapStore to check whether a token is ICRC2, using the fallback value from allSwappableTokensDerived.

# Changes

Updated the condition for isSourceTokenIcrc2 in swapStore.

### DIFF
--- a/src/frontend/src/lib/stores/swap.store.ts
+++ b/src/frontend/src/lib/stores/swap.store.ts
@@ -54,21 +54,14 @@ export const initSwapContext = (swapData: SwapData = {}): SwapContext => {
 				return false;
 			}
 
-			if (
-				nonNullish($kongSwapTokensStore) &&
-				nonNullish($kongSwapTokensStore[$sourceToken.symbol])
-			) {
-				return $kongSwapTokensStore[$sourceToken.symbol].icrc2;
-			}
+			const isKongSwapTokenIcrc2 =
+				nonNullish($kongSwapTokensStore?.[$sourceToken.symbol]) &&
+				$kongSwapTokensStore[$sourceToken.symbol].icrc2;
+			const isSwappableTokenIcrc2 = $allSwappableTokensDerived.find(
+				(token) => token.symbol === $sourceToken.symbol
+			)?.isIcrc2;
 
-			if (nonNullish($allSwappableTokensDerived)) {
-				return (
-					$allSwappableTokensDerived.find((token) => token.symbol === $sourceToken.symbol)
-						?.isIcrc2 ?? false
-				);
-			}
-
-			return false;
+			return isKongSwapTokenIcrc2 || isSwappableTokenIcrc2;
 		}
 	);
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/stores/swap.store.ts
```diff
@@ -47,12 +48,28 @@ export const initSwapContext = (swapData: SwapData = {}): SwapContext => {
 	);
 
 	const isSourceTokenIcrc2 = derived(
-		[kongSwapTokensStore, sourceToken],
-		([$kongSwapTokensStore, $sourceToken]) =>
-			nonNullish($sourceToken) &&
-			nonNullish($kongSwapTokensStore) &&
-			nonNullish($kongSwapTokensStore[$sourceToken.symbol]) &&
-			$kongSwapTokensStore[$sourceToken.symbol].icrc2
+		[kongSwapTokensStore, sourceToken, allSwappableTokensDerived],
```
DenysKarmazynDFINITY, line 51: Maybe it's just my personal preference, but I like the previous (short) version a bit more. With your change, it would look also quite neat:
```
const isSourceTokenIcrc2 = derived(
		[kongSwapTokensStore, sourceToken, allSwappableTokensDerived],
		([$kongSwapTokensStore, $sourceToken, $allSwappableTokensDerived]) =>
			nonNullish($sourceToken) &&
			((nonNullish($kongSwapTokensStore) &&
				nonNullish($kongSwapTokensStore[$sourceToken.symbol]) &&
				$kongSwapTokensStore[$sourceToken.symbol].icrc2) ||
				$allSwappableTokensDerived.find((token) => token.symbol === $sourceToken.symbol)?.isIcrc2)
	);
```

DenysKarmazynDFINITY: After I put the brackets correctly, I understood why you wanted to break it down 😄 

how about:
```
const isSourceTokenIcrc2 = derived(
		[kongSwapTokensStore, sourceToken, allSwappableTokensDerived],
		([$kongSwapTokensStore, $sourceToken, $allSwappableTokensDerived]) => {
			if (isNullish($sourceToken)) {
				return false;
			}

			const isKongSwapTokenIcrc2 =
				nonNullish($kongSwapTokensStore?.[$sourceToken.symbol]) &&
				$kongSwapTokensStore[$sourceToken.symbol].icrc2;
			const isSwappableTokenIcrc2 = $allSwappableTokensDerived.find(
				(token) => token.symbol === $sourceToken.symbol
			)?.isIcrc2;

			return isKongSwapTokenIcrc2 || isSwappableTokenIcrc2;
		}
	);
```

roman-nazaruk: Looks better, agree. Thanks

roman-nazaruk: but this value can not be undefined so, idk, I would assume that prev implementation looks as it is clear what happens and really easy to understand code

PR #9450, opened 2025-10-07
feat(frontend): Expand text component

daviddecentage:
# Motivation

We need a component where we can expand the text.

# Changes

Added component and translations

# Tests

Added tests

image

image

### DIFF
--- a/src/frontend/src/lib/components/transactions/TransactionsDateGroup.svelte
+++ b/src/frontend/src/lib/components/transactions/TransactionsDateGroup.svelte
@@ -3,6 +3,7 @@
 	import BtcTransaction from '$btc/components/transactions/BtcTransaction.svelte';
 	import EthTransaction from '$eth/components/transactions/EthTransaction.svelte';
 	import IcTransaction from '$icp/components/transactions/IcTransaction.svelte';
+	import StickyHeader from '$lib/components/ui/StickyHeader.svelte';
 	import { SLIDE_DURATION } from '$lib/constants/transition.constants';
 	import type { AllTransactionUiWithCmpNonEmptyList } from '$lib/types/transaction-ui';
 	import SolTransaction from '$sol/components/transactions/SolTransaction.svelte';
@@ -18,7 +19,11 @@
 
 {#if transactions.length > 0}
 	<div class="mb-5 flex flex-col gap-3" data-tid={testId}>
-		<span class="text-lg font-medium text-tertiary first-letter:capitalize">{formattedDate}</span>
+		<StickyHeader>
+			<span class="mb-3 flex text-lg font-medium text-tertiary first-letter:capitalize"
+				>{formattedDate}</span
+			>
+		</StickyHeader>
 
 		{#each transactions as transactionUi, index (`${transactionUi.transaction.id}-${index}`)}
 			{@const { component, token, transaction } = transactionUi}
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/ui/ExpandText.svelte
```diff
@@ -0,0 +1,28 @@
...
+	const truncated = $derived(
+		wordsSplit.length > maxWords ? `${wordsSplit.slice(0, maxWords).join(' ')}…` : text
+	);
+</script>
+
+<div class="inline-block">
+	{expanded ? text : truncated}
+
+	{#if wordsSplit.length > maxWords}
+		<Button link onclick={() => (expanded = !expanded)} styleClass="inline-block">
```
AntonioVentilii: some aria label here? maybe as optional input?

src/frontend/src/lib/components/ui/ExpandText.svelte
```diff
@@ -0,0 +1,28 @@
...
+		maxWords?: number;
+	}
+
+	const { text, maxWords = 30 }: Props = $props();
+
+	let expanded = $state(false);
+
+	const wordsSplit = $derived(text.split(' '));
+	const truncated = $derived(
+		wordsSplit.length > maxWords ? `${wordsSplit.slice(0, maxWords).join(' ')}…` : text
```
AntonioVentilii: just an ignorant question:

shall the `"..."` be localized too? meaning, do all languages do the same for `...`?

src/frontend/src/lib/components/ui/ExpandText.svelte
```diff
@@ -0,0 +1,28 @@
...
+	interface Props {
+		text: string;
+		maxWords?: number;
+	}
+
+	const { text, maxWords = 30 }: Props = $props();
+
+	let expanded = $state(false);
+
+	const wordsSplit = $derived(text.split(' '));
```
DenysKarmazynDFINITY, line 14: Shall we maybe rely on symbols instead of words? "Rechtsschutzversicherungsgesellschaften" - 1 word but 39 symbols, "he" - 1 word but 2 symbols.

daviddecentage: I tried to avoid cutting off words (as can happen if were going by character count). This component allows for multiline texts even in the truncated version so in the case of longer/shorter words it would just lead to the excerpt being a line more or less. I think the main point of the component is to just not have paragraphs full of text displayed by default which is solved by this in any case. What do you think?

DenysKarmazynDFINITY: oh yes, cutting words is a good point - all good then

src/frontend/src/lib/components/ui/ExpandText.svelte
daviddecentage: According to Chatgpt most languages use the unicode ellipsis, which Im using. Variations of it include simple three dots (`...`) or Chinese, Japanese, Korean (CJK) which use double ellipsis (`……`).

Ill add a translation for it just in case.

src/frontend/src/lib/components/ui/ExpandText.svelte
daviddecentage: Updated

PR #9453, opened 2025-10-07
refactor(frontend): Nft spam hide buttons separation

daviddecentage:
# Motivation

For upcoming PRs we need to refactor the Spam and Hide buttons for Nfts.

# Changes

- Created seperate components for spam and hide
- Extracted saving section logic to services

# Tests

Added test for that service


src/frontend/src/tests/lib/services/nft.services.spec.ts
```diff
@@ -242,4 +247,40 @@ describe('nft.services', () => {
 			expect(transfer1155Spy).not.toHaveBeenCalled();
 		});
 	});
+
+	describe('updateNftSection', () => {
```
AntonioVentilii, line 250: maybe some more tests? al the conditions/branches

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/tests/lib/services/nft.services.spec.ts
daviddecentage: added a bunch more

PR #9484, opened 2025-10-08
feat(frontend): fixed debounce in SwapAmountsContext

roman-nazaruk:
# Motivation

We had some problems with loading SwapAmounts if user changed value to fast. So in scope of this PR was fixed debounce value.

# Changes

Removed debounce and added timeout with clearDebounce.

# Tests

Added tests for new logic


src/frontend/src/lib/components/swap/SwapAmountsContext.svelte
```diff
@@ -44,8 +44,7 @@
 	const { store } = getContext<SwapAmountsContext>(SWAP_AMOUNTS_CONTEXT_KEY);
 
 	let timer: NodeJS.Timeout | undefined;
-
-	let isFetching = $state(false);
+	let debounceTimer: NodeJS.Timeout | undefined;
```
AntonioVentilii: `$state` (and `untrack` in `$effect`)

roman-nazaruk: Fixed

PR #9485, opened 2025-10-08
feat(frontend): added manual loading if isIcrc2 token is Swap

roman-nazaruk:
# Motivation

We no longer need to obtain the icrc2 value from Kongswap. Instead, we should call the ledger directly to retrieve the supported standards for ICRC tokens.

# Changes

- Removed the dependency on the general store.
- Updated the component to call the ledger directly to determine isIcrc2.

### DIFF
--- a/src/frontend/src/icp/components/swap/SwapIcpForm.svelte
+++ b/src/frontend/src/icp/components/swap/SwapIcpForm.svelte
@@ -15,6 +15,8 @@
 		swapAmount: OptionAmount;
 		receiveAmount?: number;
 		slippageValue: OptionAmount;
+		isSourceTokenIcrc2?: boolean;
+
 		sourceTokenFee?: bigint;
 		isSwapAmountsLoading: boolean;
 		onShowTokensList: (tokenSource: 'source' | 'destination') => void;
@@ -26,6 +28,7 @@
 		swapAmount = $bindable(),
 		receiveAmount = $bindable(),
 		slippageValue = $bindable(),
+		isSourceTokenIcrc2,
 		sourceTokenFee,
 		isSwapAmountsLoading,
 		onShowTokensList,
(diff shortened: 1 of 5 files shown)


src/frontend/src/icp/components/swap/SwapIcpForm.svelte
```diff
@@ -15,6 +15,8 @@
 		swapAmount: OptionAmount;
 		receiveAmount?: number;
 		slippageValue: OptionAmount;
+		isSourceTokenIcrc2?: boolean;
+
```
AntonioVentilii:
```suggestion
```

src/frontend/src/icp/components/swap/SwapIcpWizard.svelte
```diff
@@ -55,6 +58,8 @@
 		slippageValue = $bindable(),
 		swapProgressStep = $bindable(),
 		swapFailedProgressSteps = $bindable([]),
+		isSourceTokenIcrc2 = $bindable(),
+
```
AntonioVentilii:
```suggestion
```

src/frontend/src/lib/components/swap/SwapFees.svelte
```diff
@@ -14,7 +14,12 @@
 	import { formatToken, formatCurrency } from '$lib/utils/format.utils';
 	import { getTokenDisplaySymbol } from '$lib/utils/token.utils';
 
-	const { destinationToken, sourceToken, sourceTokenExchangeRate, isSourceTokenIcrc2 } =
+	interface Props {
+		isSourceTokenIcrc2?: boolean;
+	}
+	let { isSourceTokenIcrc2 }: Props = $props();
```
AntonioVentilii, line 20:
```suggestion
	let { isSourceTokenIcrc2 }: Props = $props();
	
```

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/components/swap/SwapFees.svelte
AntonioVentilii, line 21:
```suggestion

	let { isSourceTokenIcrc2 }: Props = $props();
```

PR #9489, opened 2025-10-08
feat(frontend): changed Promise.all to Promise.allSettled for loading exchanges and balances

roman-nazaruk:
# Motivation

We need to replace Promise.all with Promise.allSettled when loading exchanges and balances for ICRC tokens to ensure that one failed request does not prevent others from completing.

# Changes

Replaced Promise.all with Promise.allSettled for loading exchanges and balances of ICRC tokens.

# Tests

This logic is already covered by existing tests.

### DIFF
--- a/src/frontend/src/icp/services/icrc.services.ts
+++ b/src/frontend/src/icp/services/icrc.services.ts
@@ -17,7 +17,7 @@ import {
 	mapTokenOisySymbol,
 	type IcrcLoadData
 } from '$icp/utils/icrc.utils';
-import { TRACK_COUNT_IC_LOADING_ICRC_CANISTER_ERROR } from '$lib/constants/analytics.contants';
+import { TRACK_COUNT_IC_LOADING_ICRC_CANISTER_ERROR } from '$lib/constants/analytics.constants';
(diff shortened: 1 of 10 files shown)


src/frontend/src/icp/services/icrc.services.ts
```diff
@@ -261,37 +261,45 @@ const loadIcrcCustomData = ({
...
 				identity,
 				owner: identity.getPrincipal(),
 				ledgerCanisterId
 			});
 
+			return { id, icrcTokenBalance };
+		})
+	);
+
+	results.forEach((result) => {
```
AntonioVentilii, line 285: is it possible to have only one loop?

roman-nazaruk: It’s possible, but it looks odd and hard to read. I tried using queryAndUpdate for that, but it caused some type errors.

AntonioVentilii: may you please add 2 TODOs?

one for the query and update, and one for the reducing the loops

roman-nazaruk: Done

PR #9503, opened 2025-10-08
refactor(frontend): Nft Media Consent Modal

daviddecentage:
# Motivation

We need to make some adjustments on the Nft Media Consent Modal

# Changes

- Show 'Keep disabled' button if no preference has been saved yet
- Fixed icon fill
- Display some collection info
- Moved docs URL to constants

# Tests

Added and improved components tests

image
image
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "@dfinity/oisy-wallet",
-	"version": "1.7.4",
+	"version": "1.7.5",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "@dfinity/oisy-wallet",
-			"version": "1.7.4",
+			"version": "1.7.5",
 			"license": "Apache-2.0",
 			"dependencies": {
 				"@dfinity/agent": "^3.2.6",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@dfinity/oisy-wallet",
-	"version": "1.7.4",
+	"version": "1.7.5",
 	"private": true,
 	"license": "Apache-2.0",
 	"repository": {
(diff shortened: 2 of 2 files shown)


src/frontend/src/lib/components/icons/IconImageDownload.svelte
```diff
@@ -7,37 +7,48 @@
 	let { size = '52' }: Props = $props();
 </script>
 
-<svg height={size} viewBox="0 0 54 52" width={size} xmlns="http://www.w3.org/2000/svg">
+<svg
```
roman-nazaruk, line 10: Could you please move it to another PR

src/frontend/src/lib/components/nfts/NftImageConsentModal.svelte
```diff
@@ -34,12 +38,12 @@
...
 	const { collection, testId }: Props = $props();
 
-	const hasConsent = $derived(
+	const allowMedia = $derived(
 		getAllowMediaForNft({
 			tokens: $nonFungibleTokens,
 			networkId: collection.network.id,
 			address: collection.address
-		}) ?? false
+		}) ?? undefined
```
roman-nazaruk: Is it correct that you check value is not nullish and set undefined?

src/frontend/src/lib/components/nfts/NftImageConsentModal.svelte
```diff
@@ -106,125 +110,162 @@
...
 			</div>
-			<div class="flex w-full flex-col justify-between md:flex-row">
-				<span class="text-tertiary">{$i18n.networks.network}</span><span
-					><NetworkWithLogo network={collection.network} /></span
+
+			<p class="mb-5">
+				{$i18n.nfts.text.review_description}
+				<ExternalLink
+					ariaLabel={$i18n.nfts.text.learn_more}
+					href="https://docs.oisy.com/using-oisy-wallet/how-tos/nfts"
```
roman-nazaruk: Up to you but maybe it is better to move to constants?

src/frontend/src/lib/components/icons/IconImageDownload.svelte
daviddecentage: done

src/frontend/src/lib/components/nfts/NftImageConsentModal.svelte
daviddecentage: Good point, updated

src/frontend/src/lib/components/nfts/NftImageConsentModal.svelte
daviddecentage: Yes, since its a boolean and all we do is is it for conditions it doesnt matter if its false or undefined for that matter.

Edit: acually I dont need a fallback, it returns what I need already

PR #9524, opened 2025-10-09
feat(frontend): Get media status nft util

daviddecentage:
# Motivation

We need to check if Nft images are too big or invalid in any way.

# Changes

Added utils function to retrieve file size and file type

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/utils/nfts.utils.ts
+++ b/src/frontend/src/lib/utils/nfts.utils.ts
@@ -331,7 +331,7 @@ export const getMediaStatus = async (mediaUrl?: string): Promise<NftMediaStatusE
 			// 1MB
 			return NftMediaStatusEnum.FILESIZE_LIMIT_EXCEEDED;
 		}
-	} catch (e: unknown) {
+	} catch (_: unknown) {
 		return NftMediaStatusEnum.INVALID_DATA;
 	}
 


src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -310,3 +310,30 @@ export const getAllowMediaForNft = (params: {
...
+
+		if (nonNullish(type) && !type.startsWith('image/')) {
+			return NftMediaStatusEnum.NON_SUPPORTED_MEDIA_TYPE;
+		}
+
+		if (nonNullish(size) && Number(size) > 1024 * 1024) {
+			// 1MB
+			return NftMediaStatusEnum.FILESIZE_LIMIT_EXCEEDED;
+		}
+	} catch (e: unknown) {
```
AntonioVentilii:
```suggestion
	} catch (_: unknown) {
```

src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -310,3 +310,30 @@ export const getAllowMediaForNft = (params: {
...
+export const getMediaStatus = async (mediaUrl?: string): Promise<NftMediaStatusEnum> => {
+	try {
+		const url = new URL(mediaUrl ?? '');
+		const response = await fetch(url.href, { method: 'HEAD' });
+
+		const type = response.headers.get('Content-Type');
+		const size = response.headers.get('Content-Length');
+
+		if (isNullish(type) || isNullish(size)) {
+			throw new Error('Invalid response headers');
```
AntonioVentilii: this error is caught locally, no?

src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -310,3 +310,30 @@ export const getAllowMediaForNft = (params: {
 	address: EthAddress;
 	networkId: NetworkId;
 }): boolean | undefined => findNonFungibleToken(params)?.allowExternalContentSource;
+
+export const getMediaStatus = async (mediaUrl?: string): Promise<NftMediaStatusEnum> => {
+	try {
+		const url = new URL(mediaUrl ?? '');
```
AntonioVentilii, line 316: should there be a check for empty/nullish string? meaning, do we accept empty/nullish strings?

`if (notEmptystring(mediaUrl)) {}`

daviddecentage: I dont think its nessesary, it will throw when instancing a new URL which is fine because it will fall into the Invalid Data case

src/frontend/src/lib/utils/nfts.utils.ts
daviddecentage: Ahh yeah actually I can just return the correct enum here

PR #9532, opened 2025-10-09
refactor(frontend): Get nft for owner add media status check

daviddecentage:
# Motivation

We add the media status check when we load the Nfts.

# Changes

Adjusted method getNftsByOwner in alchemy provider

# Tests

Adjusted/extended test to ensure mapping includes the new field

### DIFF
--- a/src/frontend/src/icp/components/stake/gldt/GldtStakeForm.svelte
+++ b/src/frontend/src/icp/components/stake/gldt/GldtStakeForm.svelte
@@ -1,0 +1,43 @@
+<script lang="ts">
+	import { getContext } from 'svelte';
+	import GldtStakeFees from '$icp/components/stake/gldt/GldtStakeFees.svelte';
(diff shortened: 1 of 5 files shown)


src/frontend/src/eth/providers/alchemy.providers.ts
```diff
@@ -214,11 +222,17 @@ export class AlchemyProvider {
...
+				},
+				mediaStatus
 			};
 
-			return [...acc, nft];
-		}, []);
+			return nft;
+		});
+
+		const nfts = await Promise.all(nftPromises);
```
AntonioVentilii: why not doing it in the reduce? this creates 2 loops

daviddecentage: Initially I did it because it got to messy with the new promises inside, but I realized I can basically use reduce to accummulate the promises instead and still use promise.all.

Updated

PR #9537, opened 2025-10-09
feat(frontend): add StakeWizard component

DenysKarmazynDFINITY:
# Motivation

The next Stake flow component we need to implement is `StakeWizard`.

### DIFF
--- a/src/frontend/src/lib/components/stake/StakeWizard.svelte
+++ b/src/frontend/src/lib/components/stake/StakeWizard.svelte
@@ -39,5 +39,5 @@
 {#if isGldtToken}
 	<GldtStakeWizard {currentStep} {onBack} {onClose} {onNext} bind:amount bind:stakeProgressStep />
 {:else}
-	<div class="mt-6"><MessageBox>{$i18n.stake.text.unsupported_token_staking}</MessageBox></div>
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/stake/StakeWizard.svelte
```diff
@@ -0,0 +1,43 @@
...
+		onNext,
+		onBack
+	}: Props = $props();
+
+	const { sendToken } = getContext<SendContext>(SEND_CONTEXT_KEY);
+
+	let isGldtToken = $derived(
+		isTokenIcrc($sendToken) &&
+			($sendToken as IcrcCustomToken).ledgerCanisterId === GLDT_LEDGER_CANISTER_ID
+	);
```
AntonioVentilii, line 36: there is an `isGLDTToken` util, however i see that in other parts of the code we use this logic with the canister (that makes more sense too)

for example in `VipRewardStateModal`

```ts
	const goldToken = $derived(
		$enabledIcTokens.find((token) => token.ledgerCanisterId === GLDT_LEDGER_CANISTER_ID)
	);
```

shall we maybe adapt the util?

DenysKarmazynDFINITY: yeah, good catch - I can do it in a separate PR

PR #9538, opened 2025-10-09
feat(frontend): Sort Nft and NftCollections by date

daviddecentage:
# Motivation

Addin the missing sort by date functionality to the Nft/collection list.

# Changes

- Adjusted Nft/Collection schema, seems aquiredAt was accidentily added to the collection instead to the Nft (I have checked the mapping in alchemy provider and what alchemy delivers, this should indeed be a property of Nft, not collection)
- Added compare function and registered it in the filterSort functions condition

# Tests

Added a bunch of tests to ensure sorting works as intended


src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -206,6 +206,29 @@ const cmpByCollectionName =
...
+					const ts = nft.acquiredAt?.getTime() ?? 0;
+					return ts > max ? ts : max;
+				}, 0);
+			}
+			return 0;
+		};
+
+		const an = getNewestAcquiredAt(a);
+		const bn = getNewestAcquiredAt(b);
+
```
AntonioVentilii, line 232: furthermore, here we can almost reduce something: as soon as you have the data from `a`, you can use that to loop in `b`, without the need to loop all of `b` (as soon as there is a max date, return)

src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -206,6 +206,29 @@ const cmpByCollectionName =
...
+				// take the max acquiredAt among nfts in the collection
+				return item.nfts.reduce((max, nft) => {
+					const ts = nft.acquiredAt?.getTime() ?? 0;
+					return ts > max ? ts : max;
+				}, 0);
+			}
+			return 0;
+		};
+
+		const an = getNewestAcquiredAt(a);
```
AntonioVentilii, line 230: yes, most probably this information should be "saved" in the NftCollection directly when it is mapped

daviddecentage: Updated

src/frontend/src/lib/utils/nfts.utils.ts
daviddecentage: Updated

src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -181,6 +181,12 @@ export const getNftCollectionUi = ({
 			const entry = index.get(k);
 			if (entry) {
```
AntonioVentilii: nonNullish

daviddecentage: updated

PR #9539, opened 2025-10-09
feat(frontend): Add placeholders to nft media guard

daviddecentage:
# Motivation

We want to display placeholders in case the Nft images are broken, invalid or exceed a certain filesize.

# Changes

- Added placeholders
- Added the placeholder images as icons so we can color them using CSS

# Tests

Added tests and adjusted some other tests and testdata

### DIFF
--- a/src/frontend/src/lib/components/icons/nfts/FilesizeLimitExceeded.svelte
+++ b/src/frontend/src/lib/components/icons/nfts/FilesizeLimitExceeded.svelte
@@ -1,3 +1,4 @@
+<!-- source: DFINITY foundation -->
 <script lang="ts">
 	import { NFT_PLACEHOLDER_FILESIZE } from '$lib/constants/test-ids.constants';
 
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/icons/nfts/FilesizeLimitExceeded.svelte
```diff
@@ -0,0 +1,53 @@
+<script lang="ts">
```
roman-nazaruk, line 2: Could you please add source?

src/frontend/src/lib/components/icons/nfts/InvalidData.svelte
```diff
@@ -0,0 +1,53 @@
+<script lang="ts">
```
roman-nazaruk, line 2: Same

src/frontend/src/lib/components/icons/nfts/UnsupportedMediaType.svelte
```diff
@@ -0,0 +1,48 @@
+<script lang="ts">
```
roman-nazaruk, line 2: same

PR #9545, opened 2025-10-10
feat(frontend): add StakeModal component

DenysKarmazynDFINITY:
# Motivation

We now can add the most parent stake modal component - `StakeModal`.

### DIFF
--- a/src/frontend/src/lib/components/stake/StakeModal.svelte
+++ b/src/frontend/src/lib/components/stake/StakeModal.svelte
@@ -6,6 +6,7 @@
 	import { ProgressStepsStake } from '$lib/enums/progress-steps';
 	import { WizardStepsStake } from '$lib/enums/wizard-steps';
 	import { i18n } from '$lib/stores/i18n.store';
+	import type { OptionAmount } from '$lib/types/send';
 	import type { Token } from '$lib/types/token';
 	import { closeModal } from '$lib/utils/modal.utils';
 	import { getTokenDisplaySymbol } from '$lib/utils/token.utils';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/stake/StakeModal.svelte
```diff
@@ -0,0 +1,64 @@
...
+	interface Props {
+		token: Token;
+	}
+
+	let { token }: Props = $props();
+
+	let modal: WizardModal<WizardStepsStake> | undefined = $state();
+	let currentStep: WizardStep<WizardStepsStake> | undefined = $state();
+	let stakeProgressStep: string = $state(ProgressStepsStake.INITIALIZATION);
+	let amount: number | undefined = $state();
```
roman-nazaruk: I am not sure but you can use OptionAmount

src/frontend/src/lib/components/stake/StakeModal.svelte
```diff
@@ -0,0 +1,64 @@
...
+
+<SendTokenContext {token}>
+	<WizardModal
+		bind:this={modal}
+		disablePointerEvents={currentStep?.name === WizardStepsStake.STAKING}
+		onClose={close}
+		{steps}
+		bind:currentStep
+	>
+		{#snippet title()}{currentStep?.title ?? ''}{/snippet}
```
roman-nazaruk, line 54: Is it possible that we can not have currentStep?

src/frontend/src/lib/components/stake/StakeModal.svelte
DenysKarmazynDFINITY: yeah, we definitely can

src/frontend/src/lib/components/stake/StakeModal.svelte
DenysKarmazynDFINITY: yep - on the initial render, the value of this state var is undefined

roman-nazaruk: Okay, thnaks for update. Approved

PR #9546, opened 2025-10-10
feat(frontend): show GLDT stake modal content card

DenysKarmazynDFINITY:
# Motivation

We can now add the GLDT stake modal content card (incl. the button that shows it) to the respective page. The card itself is not styled yet.

### DIFF
--- a/src/frontend/src/icp/components/stake/gldt/GldtStakeContentCard.svelte
+++ b/src/frontend/src/icp/components/stake/gldt/GldtStakeContentCard.svelte
@@ -1,0 +1,119 @@
+<script lang="ts">
+	import { isNullish, nonNullish } from '@dfinity/utils';
+	import {
+		enabledIcrcTokens,
+		icrcTokens,
+		icrcCustomTokensNotInitialized,
+		icrcCustomTokensInitialized
+	} from '$icp/derived/icrc.derived';
+	import { loadCustomTokens } from '$icp/services/icrc.services';
+	import { setCustomToken } from '$icp-eth/services/custom-token.services';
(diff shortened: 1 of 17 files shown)


src/frontend/src/icp/components/stake/gldt/GldtStakeContentCard.svelte
```diff
@@ -0,0 +1,119 @@
...
+	import { autoLoadSingleToken } from '$lib/services/token.services';
+	import { balancesStore } from '$lib/stores/balances.store';
+	import { currencyExchangeStore } from '$lib/stores/currency-exchange.store';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+	import { formatCurrency, formatToken } from '$lib/utils/format.utils';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
+	import { calculateTokenUsdAmount, getTokenDisplaySymbol } from '$lib/utils/token.utils';
+
+	let gldtToken = $derived($enabledIcrcTokens.find((token) => isGLDTToken(token)));
```
roman-nazaruk, line 32: What do you think to move this logic to Context as I assume we would need it in the future: token, balance

DenysKarmazynDFINITY: the token (as well as its balance, and all other additional info) is already a part of the send context, it will be initialized after user opens the staking modal

roman-nazaruk: Super, thanks

PR #9549, opened 2025-10-10
feat(frontend): store cache for swap tokens that supports icrc2

roman-nazaruk:
# Motivation

To avoid unnecessary re-renders and calls to ledger_canister, this PR stores the isIcrc2 flag in the swap store.

# Changes

Store the token’s ICRC-2 support as isIcrc2 in the swap store.

# Tests

Added tests covering the new store behavior and isIcrc2 logic.

### DIFF
--- a/src/frontend/src/lib/stores/swap.store.ts
+++ b/src/frontend/src/lib/stores/swap.store.ts
@@ -23,7 +23,7 @@ type IsTokensIcrc2Map = Record<string, boolean>;
 
 export const initSwapContext = (swapData: SwapData = {}): SwapContext => {
 	const data = writable<SwapData>(swapData);
-	const isTokensIcrc2 = writable<IsTokensIcrc2Map | undefined>(undefined);
+	const isTokensIcrc2 = writable<IsTokensIcrc2Map | undefined>();
 	const { update } = data;
 
 	const sourceToken = derived([data], ([{ sourceToken }]) => sourceToken);


src/frontend/src/lib/stores/swap.store.ts
```diff
@@ -47,12 +50,13 @@ export const initSwapContext = (swapData: SwapData = {}): SwapContext => {
...
-		[kongSwapTokensStore, sourceToken],
-		([$kongSwapTokensStore, $sourceToken]) =>
-			nonNullish($sourceToken) &&
-			nonNullish($kongSwapTokensStore) &&
-			nonNullish($kongSwapTokensStore[$sourceToken.symbol]) &&
-			$kongSwapTokensStore[$sourceToken.symbol].icrc2
+		[isTokensIcrc2, sourceToken],
+		([$isTokensIcrc2, $sourceToken]) => {
+			if (isNullish($sourceToken) || !isIcToken($sourceToken) || isNullish($isTokensIcrc2)) {
+				return undefined;
```
DenysKarmazynDFINITY:
```suggestion
				return;
```

src/frontend/src/lib/stores/swap.store.ts
```diff
@@ -19,8 +19,11 @@ export interface SwapData {
 	destinationToken?: Token;
 }
 
+type IsTokensIcrc2 = Record<string, boolean>;
```
DenysKarmazynDFINITY: let's add something like "map" to highlight that this store is actually an object: `IsTokensIcrc2Map`

src/frontend/src/lib/stores/swap.store.ts
```diff
@@ -19,8 +19,11 @@ export interface SwapData {
...
+
 export const initSwapContext = (swapData: SwapData = {}): SwapContext => {
 	const data = writable<SwapData>(swapData);
+	const isTokensIcrc2 = writable<IsTokensIcrc2 | undefined>(undefined);
```
DenysKarmazynDFINITY:
```suggestion
	const isTokensIcrc2 = writable<IsTokensIcrc2 | undefined>();
```

PR #9553, opened 2025-10-10
feat(build): add VITE_EARNING_ENABLED to CI

DenysKarmazynDFINITY:
# Motivation

VITE_EARNING_ENABLED was not added to the CI build workflows.


.env.example
```diff
@@ -13,3 +13,4 @@ VITE_AUTH_ALTERNATIVE_ORIGINS=
 VITE_ONRAMPER_API_KEY_DEV=pk_test_
 VITE_ONRAMPER_API_KEY_PROD=pk_prod_
 VITE_AI_ASSISTANT_CONSOLE_ENABLED=
+VITE_EARNING_ENABLED=
```
AntonioVentilii: i think we can remove it from here, we will change them anyway to be per-provider, no?

DenysKarmazynDFINITY: right, updated

PR #9589, opened 2025-10-13
feat(frontend): Abstract class for workers

AntonioVentilii:
# Motivation

Copying the pattern from @peterpeterparker 's [Juno for the workers](https://github.com/junobuild/juno/blob/main/src/frontend/src/lib/services/workers/_worker.services.ts), we create an abstract class for all the workers.

# Changes

- Copy the `_worker.services.ts` module from Juno.
- Use it in the AuthWorker.
- Adapt the code.

# Tests

Created new tests.

### DIFF
--- a/src/frontend/src/lib/services/_worker.services.ts
+++ b/src/frontend/src/lib/services/_worker.services.ts
@@ -1,8 +1,8 @@
 export abstract class AppWorker {
-	protected _worker: Worker;
+	readonly #worker: Worker;
 
 	protected constructor(worker: Worker) {
-		this._worker = worker;
+		this.#worker = worker;
 	}
 
 	static async getInstance(): Promise<Worker> {
@@ -10,7 +10,11 @@ export abstract class AppWorker {
 		return new Workers.default();
 	}
 
+	postMessage = <T>(data: T) => {
+		this.#worker.postMessage(data);
+	};
+
 	terminate = () => {
-		this._worker.terminate();
+		this.#worker.terminate();
 	};
 }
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/services/_worker.services.ts
```diff
@@ -0,0 +1,16 @@
+export abstract class AppWorker {
+	protected _worker: Worker;
```
peterpeterparker: can `_worker` be `readonly` actually?

src/frontend/src/lib/services/worker.auth.services.ts
```diff
@@ -1,37 +1,42 @@
...
 import { idleSignOut } from '$lib/services/auth.services';
 import { authRemainingTimeStore, type AuthStoreData } from '$lib/stores/auth.store';
 import type { PostMessage, PostMessageDataResponseAuth } from '$lib/types/post-message';
+import { isNullish } from '@dfinity/utils';
 
-export const initAuthWorker = async () => {
-	const AuthWorker = await import('$lib/workers/workers?worker');
-	const authWorker: Worker = new AuthWorker.default();
+export class AuthWorker extends AppWorker {
+	constructor(worker: Worker) {
```
peterpeterparker, line 8: likewise about readonly

src/frontend/src/lib/services/_worker.services.ts
AntonioVentilii: done!

src/frontend/src/lib/services/worker.auth.services.ts
peterpeterparker: Nah my bad, not here.

PR #9593, opened 2025-10-13
feat(frontend): Send snapshots for tokens without exchange rate

AntonioVentilii:
# Motivation

We want to send the snapshots even for tokens that do now have an exchange rate. However, the Rewards Canister does not accept a nullish value as input. For now, we send them with an exchange rate of 0, even if theoretically wrong.

### DIFF
--- a/src/frontend/src/lib/services/user-snapshot.services.ts
+++ b/src/frontend/src/lib/services/user-snapshot.services.ts
@@ -246,7 +246,7 @@ const takeAccountSnapshots = (timestamp: bigint): AccountSnapshotFor[] => {
 			return acc;
 		}
 
-		// We want to send the snapshots even for tokens that do now have an exchange rate.
+		// We want to send the snapshots even for tokens that do not have an exchange rate.
 		// However, the Rewards Canister does not accept a nullish value as input.
 		// For now, we send them with an exchange rate of 0, even if theoretically wrong.
 		// TODO: Remove the nullish check and the zero-fallback once the Rewards Canister accepts nullish values.


src/frontend/src/lib/services/user-snapshot.services.ts
```diff
@@ -246,7 +246,11 @@ const takeAccountSnapshots = (timestamp: bigint): AccountSnapshotFor[] => {
 			return acc;
 		}
 
-		const exchangeRate = exchangeRates[token.id]?.usd;
+		// We want to send the snapshots even for tokens that do now have an exchange rate.
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: 👁️ 🦅

src/frontend/src/lib/services/user-snapshot.services.ts
```diff
@@ -246,7 +246,11 @@ const takeAccountSnapshots = (timestamp: bigint): AccountSnapshotFor[] => {
...
+		// However, the Rewards Canister does not accept a nullish value as input.
+		// For now, we send them with an exchange rate of 0, even if theoretically wrong.
+		// TODO: Remove the nullish check and the zero-fallback once the Rewards Canister accepts nullish values.
```
DenysKarmazynDFINITY, line 252: the nullish check, we can already remove, right?

AntonioVentilii: i would prefer to leave it for now. I am not 100% sure this is just a temporary request or if we really are going to modify the RC

PR #9609, opened 2025-10-13
feat(frontend): Accept NFT images status for missing headers

AntonioVentilii:
# Motivation

Since we don't have a better solution for now, if the server does not return informations about the files size, we accept to show the image either way.

### DIFF
--- a/src/frontend/src/tests/eth/providers/alchemy.providers.spec.ts
+++ b/src/frontend/src/tests/eth/providers/alchemy.providers.spec.ts
@@ -110,7 +110,7 @@ describe('alchemy.providers', () => {
 					...mapTokenToCollection(mockValidErc1155Token)
 				},
 				description: 'lorem ipsum',
-				mediaStatus: NftMediaStatusEnum.INVALID_DATA
+				mediaStatus: NftMediaStatusEnum.OK
 			},
 			{
 				id: parseNftId(3),


src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -359,7 +359,11 @@ export const getMediaStatus = async (mediaUrl?: string): Promise<NftMediaStatusE
 		const size = response.headers.get('Content-Length');
 
 		if (isNullish(type) || isNullish(size)) {
-			return NftMediaStatusEnum.INVALID_DATA;
```
DenysKarmazynDFINITY, line 362: if one value is not available, does it mean that the other is also absent? Otherwise, we check either `type` or `size` if at least one is there.

AntonioVentilii: the issue is that we need both, and in general yes, they go hand-in-hand (we cannot proceed without one or the other for the size check)

AntonioVentilii: the example is this one

```bash
> curl -I https://i2.seadn.io/base/0xf60a44920d51f4264ba0a8eb164a4bc15dc34e92/b10c0929e78ba6641ccf3692b8c7faf6.jpeg

HTTP/2 200 
server: CloudFront
content-type: image/avif
date: Thu, 19 Jun 2025 07:43:51 GMT
cache-control: public, max-age=31536000
x-amzn-trace-id: Root=1-6853bfb7-208fcd8b2d6c05e51c22953f;Parent=5b510e0358cdc0fc;Sampled=0;Lineage=1:39de85f2:0
via: 1.1 2d22f2597665c5eef97a8ac324518692.cloudfront.net (CloudFront)
alt-svc: h3=":443"; ma=86400
age: 10051432
vary: Origin
x-amzn-requestid: b94ca93f-0160-4075-a5fc-34859643e127
x-cache: Hit from cloudfront
x-amz-cf-pop: ZRH52-P1
x-amz-cf-id: 8Vfkew8d1ZdmZhSXkeDUt92R65VrWuoxSzu_AiTUZZhWExaStk_FWw==
```


we can load this one, but we have no info about the size, and we have it about the type

DenysKarmazynDFINITY: I see, thanks for the explanation - quite an important TODO we have here

PR #9617, opened 2025-10-14
feat(frontend): Create destroy method in abstract class `AppWorker`

AntonioVentilii:
# Motivation

It is useful to have a method that completely destroys the worker. Since we will use it for all workers, it makes sense to add it in class `AppWorker` (the abstract one that is used by the other workers).

# Changes

- Add abstract method to stop the timer of a worker.
- Add callback method to be overridden.
- Create public `destroy` method that terminates and stops the timer.

# Tests

Current tests should work as expected.

### DIFF
--- a/src/frontend/src/lib/services/_worker.services.ts
+++ b/src/frontend/src/lib/services/_worker.services.ts
@@ -10,7 +10,7 @@ export abstract class AppWorker {
 		return new Workers.default();
 	}
 
-postMessage = <T>(data: T) => {
+	postMessage = <T>(data: T) => {
 		this.#worker.postMessage(data);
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/services/_worker.services.ts
```diff
@@ -10,11 +10,31 @@ export abstract class AppWorker {
 		return new Workers.default();
 	}
 
-	postMessage = <T>(data: T) => {
+	protected postMessage = <T>(data: T) => {
```
AntonioVentilii:
```suggestion
postMessage = <T>(data: T) => {
```

PR #9619, opened 2025-10-14
feat(frontend): implement gldt_stake APY context

DenysKarmazynDFINITY:
# Motivation

To fetch APY value from the gldt_stake canister, we create a respective context.

### DIFF
--- a/src/frontend/src/icp/components/stake/gldt/GldtStakeApyContext.svelte
+++ b/src/frontend/src/icp/components/stake/gldt/GldtStakeApyContext.svelte
@@ -9,7 +9,7 @@
 	import { authIdentity } from '$lib/derived/auth.derived';
 
 	interface Props {
-		children?: Snippet;
+		children: Snippet;
 	}
 
 	let { children }: Props = $props();
@@ -34,4 +34,4 @@
 	});
 </script>
 
-{@render children?.()}
+{@render children()}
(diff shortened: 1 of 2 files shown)


src/frontend/src/icp/components/stake/gldt/GldtStakeApyContext.svelte
```diff
@@ -0,0 +1,39 @@
...
+	import { getContext, type Snippet } from 'svelte';
+	import {
+		GLDT_STAKE_APY_CONTEXT_KEY,
+		type GldtStakeApyContext
+	} from '$icp/stores/gldt-stake-apy.store';
+	import { getApyOverall } from '$lib/api/gldt_stake.api';
+	import { authIdentity } from '$lib/derived/auth.derived';
+
+	interface Props {
+		children?: Snippet;
```
AntonioVentilii: shall the `children` be non-optional?

src/frontend/src/icp/components/stake/gldt/GldtStakeApyContext.svelte
```diff
@@ -0,0 +1,39 @@
...
+
+		try {
+			gldtStakeApyStore.setApy(await getApyOverall({ identity: $authIdentity }));
+		} catch (_err: unknown) {
+			gldtStakeApyStore.reset();
+		}
+	};
+
+	$effect(() => {
+		if (isNullish($gldtStakeApyStore?.apy)) {
```
AntonioVentilii: doesn't this create a loop in case of error?

1. is nullish
2. load APY
3. raise error
4. reset APY store
5. effect triggered again
6. is nullish
7. load APY
8. raise error
9. ......

src/frontend/src/icp/components/stake/gldt/GldtStakeApyContext.svelte
DenysKarmazynDFINITY: yes, let's do it required - updated

src/frontend/src/icp/components/stake/gldt/GldtStakeApyContext.svelte
DenysKarmazynDFINITY: Good point - the check actually does not make sense as the store is not global, so it will be anyway nullish when someone closes and opens the gldt stake page.

PR #9661, opened 2025-10-15
style(frontend): Auto-hide content in transactions list

AntonioVentilii:
# Motivation

For long lists, we can enable the auto-visibility of contents that will avoid rendering the list elements that are out of sight.


AntonioVentilii:
this is not ideal

PR #9674, opened 2025-10-15
chore(frontend): Remove deprecated component `LoaderEthTransactions`

AntonioVentilii:
# Motivation

Component `LoaderEthTransactions` is not used anymore since we rely on `LoaderMultipleEthTransactions`.


roman-nazaruk (approved):
Thnx

PR #9685, opened 2025-10-15
fix(frontend): Avoid repeated tokens when managing enabled and disabled

AntonioVentilii:
# Motivation

When we manage tokens, we are currently indexing them with a string that defines the token ID and the network ID, both `symbols` but parsed as `string`.

However, it may happen to have tokens with same network and same ID (for example some NFTs).

We can directly use instead the `TokenId` type that is a `symbol`: each `symbol`, even if it has the same `description`/`string`, is treated as a different unique object. So, using it for indexing guarantees no problems in confusing one with the other.

# Changes

- Change the `Record` of modified tokens to be a `Map` in `ManageTokens` component.
- Adapt the flow to add/remove/update a modified token.

# Tests

### Before

https://github.com/user-attachments/assets/0af388ee-188c-4dca-a82d-b4e0dd4c2ee9

### After

https://github.com/user-attachments/assets/510559f8-09bd-48e0-860a-eee4ad5dbfb7

### DIFF
--- a/src/frontend/src/tests/lib/utils/token-list.utils.spec.ts
+++ b/src/frontend/src/tests/lib/utils/token-list.utils.spec.ts
@@ -15,6 +15,8 @@ import {
 	getFilteredTokenGroup,
 	getFilteredTokenList
 } from '$lib/utils/token-list.utils';
+import { parseNetworkId } from '$lib/validation/network.validation';
+import { parseTokenId } from '$lib/validation/token.validation';
 
 // Mock data for tokens
 const token1: TokenUi = BTC_MAINNET_TOKEN;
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/manage/ManageTokensModal.svelte
```diff
@@ -41,7 +41,7 @@
 	import { toastsError } from '$lib/stores/toasts.store';
 	import type { SaveCustomTokenWithKey } from '$lib/types/custom-token';
 	import type { Network } from '$lib/types/network';
-	import type { Token, TokenMetadata } from '$lib/types/token';
+	import type { Token, TokenId, TokenMetadata } from '$lib/types/token';
```
AntonioVentilii:
```suggestion
	import type { Token, TokenMetadata } from '$lib/types/token';
```

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #9686, opened 2025-10-15
refactor(frontend): Nft Metadata adjustments

daviddecentage:
# Motivation

We want to adjust some fields in the Nft metadata

# Changes

- Added copy icon and truncate class for Collection name and Token Id
- Minor styling changes to enhance responsive/long text behavior

# Tests

image

### DIFF
--- a/src/frontend/src/lib/i18n/ar.json
+++ b/src/frontend/src/lib/i18n/ar.json
@@ -1128,6 +1128,8 @@
 			"collections": "",
 			"all_assets": "",
 			"address_copied": "",
+			"collection_name_copied": "",
+			"id_copied": "",
 			"spam": "",
 			"not_spam": "",
 			"spam_warning": "",
--- a/src/frontend/src/lib/i18n/cs.json
+++ b/src/frontend/src/lib/i18n/cs.json
@@ -1128,6 +1128,8 @@
 			"collections": "Kolekce",
 			"all_assets": "Všechna aktiva",
 			"address_copied": "Adresa $address zkopírována",
+			"collection_name_copied": "",
+			"id_copied": "",
 			"spam": "Spam",
 			"not_spam": "Ne spam",
 			"spam_warning": "",
--- a/src/frontend/src/lib/i18n/de.json
+++ b/src/frontend/src/lib/i18n/de.json
@@ -1128,6 +1128,8 @@
 			"collections": "Kollektionen",
 			"all_assets": "Alle Assets",
 			"address_copied": "Adresse $address kopiert",
+			"collection_name_copied": "",
+			"id_copied": "",
 			"spam": "Spam",
 			"not_spam": "Kein Spam",
 			"spam_warning": "",
--- a/src/frontend/src/lib/i18n/fr.json
+++ b/src/frontend/src/lib/i18n/fr.json
@@ -1128,6 +1128,8 @@
 			"collections": "Collections",
 			"all_assets": "Tous les actifs",
 			"address_copied": "Adresse $address copiée",
+			"collection_name_copied": "",
+			"id_copied": "",
 			"spam": "Spam",
 			"not_spam": "Pas de spam",
 			"spam_warning": "",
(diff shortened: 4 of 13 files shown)


src/frontend/src/lib/components/nfts/NftMetadataList.svelte
```diff
@@ -25,14 +27,45 @@
 	const collection: NftCollection | undefined = $derived(
 		nft?.collection ?? (nonNullish(token) ? mapTokenToCollection(token) : undefined)
 	);
+
+	const allowMedia = $derived(
```
AntonioVentilii: this property should be created with the mapping of the collection

src/frontend/src/lib/components/nfts/NftMetadataList.svelte
```diff
@@ -25,14 +27,45 @@
...
+	</ListItem>
 	<ListItem>
-		<span class="text-tertiary">{$i18n.nfts.text.collection_name}</span>
+		<span class="flex whitespace-nowrap text-tertiary">{$i18n.nfts.text.collection_name}</span>
 		{#if nonNullish(collection?.name)}
-			<span class="flex items-center">
-				{shortenWithMiddleEllipsis({ text: collection?.name })}
+			<span class="inline-flex min-w-0 items-center">
+				<span class="truncate"
+					>{collection?.name} very very long collection name very very long collection name very very
```
AntonioVentilii: ?

src/frontend/src/lib/components/nfts/NftMetadataList.svelte
```diff
@@ -42,12 +75,18 @@
 	</ListItem>
 	{#if nonNullish(nft)}
 		<ListItem>
-			<span class="text-tertiary">{$i18n.nfts.text.token_id}</span>
-			{nft.id}
+			<span class="flex whitespace-nowrap text-tertiary">{$i18n.nfts.text.token_id}</span>
+			<span class="inline-flex min-w-0 items-center">
+				<span class="truncate"
+					>{nft?.id} very very long collection name very very long collection name very very long collection
```
AntonioVentilii: ?

AntonioVentilii (changes requested):

src/frontend/src/lib/components/nfts/NftMetadataList.svelte
daviddecentage: This is a dynamic property that needs to be reactive though.

I made a comment for that utils function back when I implemented it:

`We offer this util so we don't mistakenly take the value from the nfts collection prop, as it is not updated after updating the consent. Going through this function ensures no stale data`

src/frontend/src/lib/components/nfts/NftMetadataList.svelte
daviddecentage: Ahh yes sorry this was to test long strings with truncate/ellipsis. Removed

src/frontend/src/lib/components/nfts/NftMetadataList.svelte
daviddecentage: Ahh yes sorry this was to test long strings with truncate/ellipsis. Removed

src/frontend/src/lib/components/nfts/NftMetadataList.svelte
AntonioVentilii:
```ts
export const getAllowMediaForNft = (params: {
	tokens: NonFungibleToken[];
	address: EthAddress;
	networkId: NetworkId;
}): boolean | undefined => findNonFungibleToken(params)?.allowExternalContentSource;

```

the implementation basically says: re-loop the tokens and check this property. But collection is already mapped somewhere, and this property should be mapped there, no? It is not reactive, it is set when the users allows or not, so it should already go in the mapping of NTFs/collection, or similar. Basically, it should be a property in `collection`

I mean, it is already in `$nonFungibleTokens`, that most probably is already connected somehow with the current `collection` object.

This logic creates an additional loop that can be avoided

daviddecentage: Solved in https://github.com/dfinity/oisy-wallet/pull/9712

AntonioVentilii (commented):
Let's split this PR:

1. Moving the network snippet
2. Adding the `{#if nonNullish(nft)}` snippet
3. Adjust layout,style,classes

all of them can be done at the same time

PR #9688, opened 2025-10-15
feat(frontend): added new trackEvent for nft spam/hide buttons

roman-nazaruk:
# Motivation

We aim to track nft flow with trackEvents.

# Changes

Added new trackEvent nft-spam-hide-action

# Tests

Covered new logic with tests


daviddecentage (approved):
LGTM

PR #9693, opened 2025-10-15
refactor(frontend): Hero buttons skeleton variant

daviddecentage:
# Motivation

In #9637 we removed the the loadingAsSkeleton prop and general style variant for loading buttons.

Design wants the skeleton button back in the Hero. We still keep the spinner for normal button loading states but this way we can still have both and apply on demand.

# Changes

- Added back prop
- Added prop to HeroButton

# Tests

image


AntonioVentilii:
@daviddecentage my point is exactly that: the "loading" status should be the same everywhere, no? I understand that the spinner is actually used to say something like "we are finishing doing the flow that you clicked", but for that we have the `InProgressWizard`. So we should remove that too, if we are using only the spinner

src/frontend/src/lib/components/ui/Button.svelte
```diff
@@ -30,6 +31,7 @@
 		type = 'submit',
 		disabled,
 		loading = false,
+		loadingAsSkeleton = false,
```
AntonioVentilii: i think we need to split the topics: one is loading because we are loading something (no user interaction), and the other loading is after a click, and the user is waiting for something. They should not be considered the same state

So, one can be called `loading` and the other `working` maybe, or `awaiting`? Any better idea on names? Or even invert it, boh...

And the pulsating state takes priority on the spinning, since it cannot be both

Priority:
- disabled
- loading (pulsating)
- loading (after click)

daviddecentage: Updated

AntonioVentilii: was it like this before? we didn't have a "spinning" state before, I may have lost something here. We want 3 states right?

1. disabled
2. pulsating
3. spinning

right? the 3rd one is new (it was recently added)

Or not?

PR #9694, opened 2025-10-15
feat(frontend): add "destination" to StakeForm

DenysKarmazynDFINITY:
# Motivation

We need to extend the StakeForm component with "destination" field.

Screenshot 2025-10-15 at 18 17 18


src/frontend/src/icp/components/stake/gldt/GldtStakeWizard.svelte
```diff
@@ -31,6 +33,8 @@
 		onBack
 	}: Props = $props();
 
+	let destination = $derived(Principal.fromText(GLDT_STAKE_CANISTER_ID).toString());
```
AntonioVentilii: this is a constant, not? it will change only by ENV, so is `$derived` required?

DenysKarmazynDFINITY: very good point - not needed indeed

PR #9696, opened 2025-10-16
refactor(frontend): Migrate `ReceiveAddressModal` to Svelte 5

AntonioVentilii:
# Motivation

Migrating component `ReceiveAddressModal` to Svelte 5.

### DIFF
--- a/src/frontend/src/lib/components/receive/ReceiveAddressModal.svelte
+++ b/src/frontend/src/lib/components/receive/ReceiveAddressModal.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
 	import { WizardModal, type WizardStep, type WizardSteps } from '@dfinity/gix-components';
-	import { nonNullish } from '@dfinity/utils';
-	import { createEventDispatcher } from 'svelte';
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/components/receive/ReceiveAddressModal.svelte
```diff
@@ -32,33 +36,41 @@
...
 		copyAriaLabel = undefined;
 	};
-
-	const dispatch = createEventDispatcher();
 </script>
 
 <WizardModal
 	bind:this={modal}
-	onClose={() => closeModal(() => dispatch('nnsClose'))}
+	onClose={() => closeModal(() => onClose?.())}
```
DenysKarmazynDFINITY, line 73: Can we change it to smth like `onClose={() => closeModal(onClose)}`? Or that function doesn't accept `undefined`?

AntonioVentilii: it doesn't accept it, but good point, i will change it in another PR

PR #9705, opened 2025-10-16
refactor(frontend): Switch interval to timeout in Exchange worker

AntonioVentilii:
# Motivation

Using `setTimeout` (recursive) is generally safer than `setInterval` when the callback is asynchronous, because it waits for the previous run to finish before scheduling the next one. It guarantees serial execution and avoids race conditions.

### With `setInterval`

- The timer ticks at fixed intervals no matter how long the callback takes.
- If the callback takes longer than `interval`, the next call starts before the previous one finishes → overlapping executions (race conditions, unexpected states, or API overload).

### With recursive `setTimeout`

- Each cycle is scheduled after the previous `await` completes.
- The delay is still roughly the same, but ensures no overlap and makes execution timing predictable.

In this PR we migrate the worker for Exchange rates.

### DIFF
--- a/src/frontend/src/lib/workers/exchange.worker.ts
+++ b/src/frontend/src/lib/workers/exchange.worker.ts
@@ -58,7 +58,17 @@ const startExchangeTimer = async (data: PostMessageDataRequestExchangeTimer | un
 	// We sync now but also schedule the update afterward
 	await sync();
 
-	timer = setInterval(sync, SYNC_EXCHANGE_TIMER_INTERVAL);
+	const scheduleNext = (): void => {
+		timer = setTimeout(async () => {
+			await sync();
+
+			if (nonNullish(timer)) {
+				scheduleNext();
+			}
+		}, SYNC_EXCHANGE_TIMER_INTERVAL);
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/workers/exchange.worker.ts
```diff
@@ -58,15 +58,25 @@ const startExchangeTimer = async (data: PostMessageDataRequestExchangeTimer | un
 	// We sync now but also schedule the update afterward
 	await sync();
 
-	timer = setInterval(sync, SYNC_EXCHANGE_TIMER_INTERVAL);
+	const scheduleNext = (): void => {
+		timer = setTimeout(async () => {
+			await sync();
+
+			if (nonNullish(timer)) {
```
DenysKarmazynDFINITY, line 65: In the siblings PR, we used `timer !== undefined` - maybe let's keep it like that here too

DenysKarmazynDFINITY: or actually the other way around, I will comment in that PR

PR #9707, opened 2025-10-16
feat(frontend): implement gldt-stake service

DenysKarmazynDFINITY:
# Motivation

We need to create a service for staking GLDT. It's gonna be called by the staking wizard component.


src/frontend/src/icp/services/gldt-stake.services.ts
```diff
@@ -0,0 +1,52 @@
...
+}: {
+	identity: Identity;
+	gldtToken: IcToken;
+	amount: bigint;
+	stakeCompleted: () => void;
+	progress?: (step: ProgressStepsStake) => void;
+}): Promise<StakePositionResponse | undefined> => {
+	progress?.(ProgressStepsStake.APPROVE);
+
+	await approve({
```
AntonioVentilii, line 27: I imagine that the errors are caught in the consumer?

DenysKarmazynDFINITY: yep, inside the wizard

PR #9710, opened 2025-10-16
feat(frontend): adjust StakeProgress

DenysKarmazynDFINITY:
# Motivation

We need to adjust `StakeProgress` and related progress steps.

### DIFF
--- a/src/frontend/src/lib/components/stake/StakeProgress.svelte
+++ b/src/frontend/src/lib/components/stake/StakeProgress.svelte
@@ -17,18 +17,8 @@
 			state: 'in_progress'
 		},
 		{
-			step: ProgressStepsStake.SIGN_APPROVE,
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/stake/StakeProgress.svelte
```diff
@@ -16,21 +16,11 @@
 			text: $i18n.send.text.initializing,
 			state: 'in_progress'
 		},
-		{
-			step: ProgressStepsStake.SIGN_APPROVE,
-			text: $i18n.send.text.signing_approval,
```
AntonioVentilii, line 21: shall we remove the strings from `i18n` too? both of them?

DenysKarmazynDFINITY: these are the strings from the `send` flow, so they are still in use

PR #9711, opened 2025-10-16
style(frontend): Disable flip animations for iOS

AntonioVentilii:
# Motivation

The Svelte animation `flip` weighs quite a lot on iOS devices. It may cause to trigger a memory (Jetsam) kill.

So, for now, we simply disable it for all iOS devices.

# Changes

- Create util to recognize an iOS device.
- Add disclaimer for devices that pretend to be Macs but are indeed iOS (like some iPads).
- Make the animation dependent on the device.

# Tests

Created tests.

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensList.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensList.svelte
@@ -118,7 +118,7 @@
 
 	let ios = $derived(isIos());
 
-	let flipParams = $derived({ duration: ios ? 0 : 250 })
+	let flipParams = $derived({ duration: ios ? 0 : 250 });
 </script>
 
 <TokensDisplayHandler {animating} bind:tokens>


src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -126,7 +129,7 @@
 					onanimationend={handleAnimationEnd}
 					onanimationstart={handleAnimationStart}
 					transition:fade
-					animate:flip={{ duration: 250 }}
+					animate:flip={{ duration: ios ? 0 : 250 }}
```
DenysKarmazynDFINITY: let's create another var for this settings and re-use it in both places

AntonioVentilii: done!

PR #9712, opened 2025-10-16
refactor(frontend): Allow external content source

daviddecentage:
# Motivation

We simplify the access for allow external content source flag on collections.

Previously, it has been taken from the token and via util, which is not nessesary and creates unnessesary loops.

# Changes

- Map the prop directly from token to collection
- Adjust usages
- Removed getAllowMedia utils function

# Tests

Adjusted test

### DIFF
--- a/src/frontend/src/lib/utils/nfts.utils.ts
+++ b/src/frontend/src/lib/utils/nfts.utils.ts
@@ -336,9 +336,6 @@ export const findNonFungibleToken = ({
 }): NonFungibleToken | undefined =>
 	tokens.find((token) => token.address === address && token.network.id === networkId);
 
-export const getAllowMediaForNft = (collection: NftCollection): boolean | undefined =>
-	collection.allowExternalContentSource;
-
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/utils/nfts.utils.ts
```diff
@@ -333,13 +336,8 @@ export const findNonFungibleToken = ({
...
 	tokens.find((token) => token.address === address && token.network.id === networkId);
 
-// We offer this util so we don't mistakenly take the value from the nfts collection prop,
-// as it is not updated after updating the consent. Going through this function ensures no stale data
-export const getAllowMediaForNft = (params: {
-	tokens: NonFungibleToken[];
-	address: EthAddress;
-	networkId: NetworkId;
-}): boolean | undefined => findNonFungibleToken(params)?.allowExternalContentSource;
+export const getAllowMediaForNft = (collection: NftCollection): boolean | undefined =>
```
AntonioVentilii: i think this can be removed already, no?

daviddecentage: yes right, removed it

PR #9714, opened 2025-10-16
refactor(frontend): Nft card labels

daviddecentage:
# Motivation

We want to change the way Nft card labels display.

On the collection page, it should stay as is.

On the list view Nft home page it should display the Nft name after the ID and render the collection name above.

# Changes

Adjusted component

# Tests

Adjusted text

List view:
image

Collection page:
image

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftCard.svelte
+++ b/src/frontend/src/lib/components/nfts/NftCard.svelte
@@ -113,15 +113,11 @@
 		</span>
 	</span>
 
-	<span class="flex w-full flex-col gap-1 px-2 pb-2">
-		<span
-			class="truncate text-sm font-bold"
-			class:text-disabled={disabled}
-			class:text-primary={!disabled}
-		>
+	<span class="flex w-full flex-col gap-1 px-2 pb-2" class:text-disabled={disabled}>
+		<span class="truncate text-sm font-bold" class:text-primary={!disabled}>
 			{source !== NFT_LIST_ROUTE ? nft.name : nft.collection.name}
 		</span>
-		<span class="text-xs" class:text-disabled={disabled} class:text-tertiary={!disabled}>
+		<span class="text-xs" class:text-tertiary={!disabled}>
 			#{nft.id}
 			{#if source === NFT_LIST_ROUTE}
 				&ndash; {nft.name}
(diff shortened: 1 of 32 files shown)


src/frontend/src/lib/components/nfts/NftCard.svelte
AntonioVentilii: something like this?

```suggestion
		source?:'default' | NFT_COLLECTION_ROUTE | NFT_LIST_ROUTE;
```

daviddecentage: That doesnt work, those are const not enums

AntonioVentilii: you can use `typeof NFT_COLLECTION_ROUTE` etc etc

daviddecentage: updated

src/frontend/src/lib/components/nfts/NftCard.svelte
```diff
@@ -116,10 +117,15 @@
...
 			class:text-disabled={disabled}
-			class:text-primary={!disabled}>{nft.name}</span
-		>
-		<span class="text-xs" class:text-disabled={disabled} class:text-tertiary={!disabled}
-			>#{nft.id}</span
+			class:text-primary={!disabled}
 		>
+			{source !== NFT_LIST_ROUTE ? nft.name : nft.collection.name}
+		</span>
+		<span class="text-xs" class:text-disabled={disabled} class:text-tertiary={!disabled}>
```
AntonioVentilii: this can go directly in the parent `<span>` i think, for both of them

AntonioVentilii:
```svelte
<span
	class="flex w-full flex-col gap-1 px-2 pb-2"
	class:text-disabled={disabled}
>
	<span class="truncate text-sm font-bold" class:text-primary={!disabled}>
		{source !== NFT_LIST_ROUTE ? nft.name : nft.collection.name}
	</span>
	<span class="text-xs" class:text-tertiary={!disabled}>
		#{nft.id}
		{#if source === NFT_LIST_ROUTE}
			&ndash; {nft.name}
		{/if}
	</span>
</span>
```

daviddecentage: #9728

PR #9715, opened 2025-10-16
feat(frontend): use gldt-stake service in the wizard

DenysKarmazynDFINITY:
# Motivation

We can now connect the wizard component with the staking service.


src/frontend/src/lib/constants/analytics.constants.ts
```diff
@@ -178,6 +178,10 @@ export const TRACK_BUY_TOKEN = 'buy_token';
 export const TRACK_OPEN_ASSETS_TAB = 'open_assets_tab';
 export const TRACK_NFT_SETTINGS_CHANGE = 'nft_settings_change';
 export const TRACK_NFT_SORT_CHANGE = 'nft_sort_change';
+
+// Stake
+export const TRACK_COUNT_STAKE_SUCCESS = 'stake_success';
```
AntonioVentilii: i think that the `Stake` block introduces itself into the `NFT` block eheh

DenysKarmazynDFINITY: rebasing... 😄

PR #9724, opened 2025-10-17
refactor(frontend): Nft metadata adds new fields

daviddecentage:
# Motivation

We want to add the media URL and acquired at date to the Nft metadatalist.

# Changes

Added the two fields

# Tests

Added tests for the new fields

image

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftMetadataList.svelte
+++ b/src/frontend/src/lib/components/nfts/NftMetadataList.svelte
@@ -32,6 +32,16 @@
 
 <List condensed itemStyleClass="flex-col sm:flex-row" styleClass="text-sm text-primary">
 	<ListItem>
+		<span class="text-tertiary">{$i18n.networks.network}</span>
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/nfts/NftMetadataList.svelte
```diff
@@ -25,6 +26,10 @@
 	const collection: NftCollection | undefined = $derived(
 		nft?.collection ?? (nonNullish(token) ? mapTokenToCollection(token) : undefined)
 	);
+
+	const allowMedia = $derived(
+		nonNullish(collection) ? collection.allowExternalContentSource : undefined
```
DenysKarmazynDFINITY: just `const allowMedia = $derived(collection?. allowExternalContentSource)` should also work

daviddecentage: updated

PR #9727, opened 2025-10-17
feat(frontend): update StakeProvider design

DenysKarmazynDFINITY:
# Motivation

The first step of implement the final design of Stake form is to update how the provider info looks like. The component will be extended with further data in the follow-up PRs.

Before:
Screenshot 2025-10-17 at 10 15 41

After:
Screenshot 2025-10-17 at 10 14 57

### DIFF
--- a/src/frontend/src/lib/components/stake/StakeProvider.svelte
+++ b/src/frontend/src/lib/components/stake/StakeProvider.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
 	import ExternalLink from '$lib/components/ui/ExternalLink.svelte';
 	import Logo from '$lib/components/ui/Logo.svelte';
-	import ModalValue from '$lib/components/ui/ModalValue.svelte';
 	import { stakeProvidersConfig } from '$lib/config/stake.config';
 	import {
 		STAKE_PROVIDER_EXTERNAL_URL,
@@ -9,6 +8,7 @@
 	} from '$lib/constants/test-ids.constants';
 	import { i18n } from '$lib/stores/i18n.store';
 	import type { StakeProvider } from '$lib/types/stake';
+	import { replacePlaceholders } from '$lib/utils/i18n.utils';
 
 	interface Props {
 		provider: StakeProvider;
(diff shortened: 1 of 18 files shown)


src/frontend/src/lib/i18n/en.json
```diff
@@ -1714,7 +1714,8 @@
 			"review": "Review transaction",
 			"executing_transaction": "Executing transaction...",
 			"unsupported_token_staking": "Staking for the selected token is not supported.",
-			"provider": "Staking provider",
+			"provider": "Staking terms provided by $provider",
```
AntonioVentilii, line 1717: since this string was changed, WDYT of resetting it in the other languages too? So that we can check changes

src/frontend/src/lib/components/stake/StakeProvider.svelte
```diff
@@ -17,27 +17,31 @@
...
 
-<ModalValue>
-	{#snippet label()}
-		{$i18n.stake.text.provider}
-	{/snippet}
-
-	{#snippet mainValue()}
-		<div class="flex items-center gap-1">
+<div class="my-4 rounded-lg border border-disabled bg-secondary px-2 py-3">
+	<div class="flex justify-between">
```
AntonioVentilii, line 21: doubt: do you need a nested `div` here? or can it be a single one?

src/frontend/src/lib/i18n/en.json
DenysKarmazynDFINITY: oh, very good point

src/frontend/src/lib/components/stake/StakeProvider.svelte
DenysKarmazynDFINITY: yes, the component will grow, so all the current divs are needed

src/frontend/src/lib/i18n/en.json
DenysKarmazynDFINITY: luckily, it wasn't translated to any of the languages, so we're good

PR #9730, opened 2025-10-17
feat(frontend): update AI button appearance

DenysKarmazynDFINITY:
# Motivation

To make things cleaner, as well as to fix the issue when the AI button does not render on some screens, we need to update the layout and footer - just one `AiAssistantConsoleButton` should be enough to cover all the requirements.


AntonioVentilii (approved):
Nice! Thank you!

PR #9731, opened 2025-10-17
feat(frontend): fixed velora delta swap destAmount

roman-nazaruk:
# Motivation

We need to fix Velora delta swap receive amount

# Changes

Start using new field from velora response with fallback.

We need to use `destAmountAfterBridge` in `bridgeInfo` for delta swaps, if available, since this value is not mentioned in the documentation. However, after discussions with the Velora team, we confirmed that this value is already formatted correctly with the required token amount precision.

# Tests

Adjusted tests

### DIFF
--- a/src/frontend/src/lib/components/core/Footer.svelte
+++ b/src/frontend/src/lib/components/core/Footer.svelte
@@ -2,7 +2,6 @@
 	import { IconGitHub } from '@dfinity/gix-components';
 	import { page } from '$app/state';
 	import AiAssistantConsole from '$lib/components/ai-assistant/AiAssistantConsole.svelte';
-	import AiAssistantConsoleButton from '$lib/components/ai-assistant/AiAssistantConsoleButton.svelte';
 	import IconDfinity from '$lib/components/icons/IconDfinity.svelte';
 	import IconHeart from '$lib/components/icons/IconHeart.svelte';
 	import IconTwitter from '$lib/components/icons/IconTwitter.svelte';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/utils/swap.utils.ts
```diff
@@ -170,10 +171,13 @@ export const formatReceiveOutMinimum = ({
 	});
 };
 
-export const mapVeloraSwapResult = (swap: DeltaPrice): SwapMappedResult => ({
+export const mapVeloraSwapResult = (swap: QuoteWithDeltaPriceAndBridgePrice): SwapMappedResult => ({
 	provider: SwapProvider.VELORA,
-	receiveAmount: BigInt(swap.destAmount),
-	swapDetails: swap as VeloraSwapDetails,
+	receiveAmount:
+		'bridgeInfo' in swap.delta
```
AntonioVentilii, line 183: please add an extensive comment on the reason we are doing it, since it is not documented in the Velora official docs either. And update the description

roman-nazaruk: Added

AntonioVentilii: what about

```ts
// Velora does not always return the destination amount in the precision of the destination token (as we would expect).
// For example, if we request a swap from USDC-BSC (18 digits) to USDC-BASE (6 digits), the destination amount is returned as if it has 18 digits instead of 6.
// This causes issues in the normal formatting of our code, since we expect each amount to be strictly related to its reference token.
// To avoid this issue, we could use the `bridgeInfo` data that Velora adds for this specific cases: it specify the correct amount to look at when the bridge is treating tokens with scaling factor.
// This is not documented anywhere in Velora documentation, it was the result of a direct conversations with them.
// TODO: remove this disclaimer where Velora fixes this issue on their side.
```

roman-nazaruk: Updated

PR #9738, opened 2025-10-17
refactor(frontend): Make keys in tokens list deterministic

AntonioVentilii:
# Motivation

In the tokens list, if the keys change when we sort/filter, Svelte will treat items as new and run intro/outro/FLIP again (→ the “detached” pileup). We want stable, primitive keys that never change across renders.

So, we avoid `symbol` and use a `string` that’s deterministic.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #9769, opened 2025-10-20
refactor(frontend): refactored spam hide nft buttons events

roman-nazaruk:
# Motivation

WE aim to refactor plausible events for nfts.

# Changes

Refactored events for hide/spam buttons for nfts

# Tests

Updated tests

### DIFF
--- a/src/frontend/src/icp/canisters/gldt_stake.canister.ts
+++ b/src/frontend/src/icp/canisters/gldt_stake.canister.ts
@@ -8,7 +8,8 @@ import { idlFactory as idlFactoryGldtStake } from '$declarations/gldt_stake/gldt
 import { mapGldtStakeCanisterError } from '$icp/canisters/gldt_stake.errors';
 import { getAgent } from '$lib/actors/agents.ic';
 import type { CreateCanisterOptions } from '$lib/types/canister';
-import { Canister, createServices } from '@dfinity/utils';
+import type { Principal } from '@dfinity/principal';
+import { Canister, createServices, fromNullable } from '@dfinity/utils';
 
 export class GldtStakeCanister extends Canister<GldtStakeService> {
 	static async create({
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/nfts/NftHideButton.svelte
```diff
@@ -33,24 +33,41 @@
 		nonNullish($nftStore) ? findNftsByToken({ nfts: $nftStore, token }).length > 1 : false
 	);
 
+	const trackNftCategorizeEvent = ({
+		value,
+		metadata
+	}: {
+		value?: CustomTokenSection;
+		metadata: Record<string, string>;
```
AntonioVentilii: this can be directly result success, right?

src/frontend/src/lib/components/nfts/NftSpamButton.svelte
```diff
@@ -34,22 +34,39 @@
 
 	let loading = $state(false);
 
+	const trackNftCategorizeEvent = ({
+		value,
+		metadata
+	}: {
+		value?: CustomTokenSection;
+		metadata: Record<string, string>;
```
AntonioVentilii: same question

src/frontend/src/lib/components/nfts/NftHideButton.svelte
roman-nazaruk: yes, but we can extend it in the future

roman-nazaruk: Updated just to have status

src/frontend/src/lib/components/nfts/NftSpamButton.svelte
roman-nazaruk: Fixed

PR #9772, opened 2025-10-20
refactor(frontend): refactored list update settings for nft events

roman-nazaruk:
# Motivation

We aim to refactor nft events.

# Changes

Refactored updating list settings for nft events

# Tests

Updated tests

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftImageConsentModal.svelte
+++ b/src/frontend/src/lib/components/nfts/NftImageConsentModal.svelte
@@ -15,12 +15,15 @@
 	import ContentWithToolbar from '$lib/components/ui/ContentWithToolbar.svelte';
 	import ExpandText from '$lib/components/ui/ExpandText.svelte';
 	import ExternalLink from '$lib/components/ui/ExternalLink.svelte';
-	import { TRACK_NFT_CONSENT_GIVEN } from '$lib/constants/analytics.constants';
 	import { OISY_NFT_DOCS_URL } from '$lib/constants/oisy.constants';
 	import { authIdentity } from '$lib/derived/auth.derived';
 	import { nonFungibleTokens } from '$lib/derived/tokens.derived';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/constants/analytics.constants.ts
```diff
@@ -187,3 +187,29 @@ export const TRACK_NFT_SPAM_HIDE_ACTION = 'nft_spam_hide_action';
 // Stake
 export const TRACK_COUNT_STAKE_SUCCESS = 'stake_success';
 export const TRACK_COUNT_STAKE_ERROR = 'stake_error';
+
+// New Plausible Events
```
AntonioVentilii: should this go in another module?

src/frontend/src/tests/lib/components/nfts/NftSettingsMenu.spec.ts
```diff
@@ -182,10 +197,13 @@ describe('NftsSettingsMenu', () => {
...
 
 		expect(trackEventSpy).toHaveBeenCalledWith({
-			name: TRACK_NFT_SETTINGS_CHANGE,
-			metadata: expect.objectContaining({
-				show_spam: expect.stringMatching(/enabled|disabled/)
-			})
+			name: PLAUSIBLE_EVENTS.LIST_SETTINGS_CHANGE,
+			metadata: {
+				event_context: PLAUSIBLE_EVENT_CONTEXTS.NFT,
+				event_subcontext: 'spam',
```
AntonioVentilii, line 203: shouldn't this be an enum too?

roman-nazaruk: TBH, no, as it is just 1 case where we should use it and I don't think that we should create a huge enum with all the values that we will never reuse in the future

PR #9828, opened 2025-10-21
feat(frontend): Make component `LoaderCollections` more reactive

AntonioVentilii:
# Motivation

Component `LoaderCollections` should load the NFT collecitons in a more reactive way instead of doing it at intervals.

### DIFF
--- a/src/frontend/src/lib/components/loaders/LoaderCollections.svelte
+++ b/src/frontend/src/lib/components/loaders/LoaderCollections.svelte
@@ -181,7 +181,7 @@
 		}
 	};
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/loaders/LoaderCollections.svelte
```diff
@@ -182,8 +180,14 @@
 			});
 		}
 	};
+
+	const debounceLoad = debounce(onLoad, 1000);
```
AntonioVentilii:
```suggestion
	const debounceLoad = debounce(onLoad);
```

PR #9830, opened 2025-10-21
fix(frontend): fixed nft card events propertirs

roman-nazaruk:
# Changes

Fixed nft card event properties

# Tests

Updated tests

### DIFF
--- a/src/frontend/src/lib/components/ui/Tabs.svelte
+++ b/src/frontend/src/lib/components/ui/Tabs.svelte
@@ -35,7 +35,7 @@
 					metadata: {
 						event_context: PLAUSIBLE_EVENT_CONTEXTS.ASSETS_TAB,
 						event_value: id,
-						location_source: PLAUSIBLE_EVENT_SOURCES.HOMEPAGE
+						location_source: PLAUSIBLE_EVENT_SOURCES.ASSETS_PAGE
 					}
 				});
 			}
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/nfts/NftCard.svelte
```diff
@@ -56,7 +56,8 @@
 					token_network: nft.collection.network.name,
 					token_address: nft.collection.address,
 					token_symbol: nft.collection.symbol ?? '',
-					token_name: nft.collection.name ?? ''
+					token_name: nft.name ?? '',
+					token_id: String(nft.id)
```
AntonioVentilii: if ID is a `symbol`, this will not work. You have to use `nft.id.description ?? ''`

AntonioVentilii: just checked, it is not a symbol, all good

AntonioVentilii:
```suggestion
					token_id:nft.id
```

roman-nazaruk: nft.id is a Symbol and we can not use typeof Symbol for plausible events

AntonioVentilii: ah, then [my first comment](https://github.com/dfinity/oisy-wallet/pull/9830#discussion_r2448146603) stays: if the ID is a `symbol` object, you should use `nft.id.description ?? ''`

AntonioVentilii: but if it is a `string` object, then all good, and you don't need to do `String(nft.id)`

AntonioVentilii: i think it is a string, no?

`export const NftIdSchema = z.string().brand<'NftId'>();`

roman-nazaruk: Yeah, you are right. Thank you. Good catch

PR #9833, opened 2025-10-21
fix(frontend): fixed nft page-open event_value

roman-nazaruk:
# Motivation

Updated nft page-open event  event_value

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
+++ b/src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
@@ -62,7 +62,14 @@
 	}}
 >
 	<div class="relative h-full w-full">
-		<NftDisplayGuard nft={previewNft} type="card">
+		<NftDisplayGuard
+			location={{
+				source: PLAUSIBLE_EVENT_SOURCES.NFTS_PAGE,
+				subSource: 'card'
+			}}
+			nft={previewNft}
+			type="card"
+		>
 			<div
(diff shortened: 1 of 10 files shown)


src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
```diff
@@ -51,7 +51,7 @@
 			name: PLAUSIBLE_EVENTS.PAGE_OPEN,
 			metadata: {
 				event_context: PLAUSIBLE_EVENT_CONTEXTS.NFT,
-				event_value: PLAUSIBLE_EVENT_VALUES.NFT,
+				event_value: PLAUSIBLE_EVENT_SOURCES.NFT_COLLECTION,
```
AntonioVentilii: doubt: why an event value has a SOURCES enum?

roman-nazaruk: That’s a case where we need this value, since the event itself is to open a page. The event value represents the source, so I decided to reuse this enum value.

AntonioVentilii: why not adding a value to `PLAUSIBLE_EVENT_VALUES`? it may avoid confusions. Or if you will have repeated cross-values, maybe better to refactor the enums directly, creating a basic one that can be reused by scope-specific enums

roman-nazaruk: Updated

PR #9834, opened 2025-10-21
feat(frontend): Build nft url util

daviddecentage:
# Motivation

Adding a utility to easily build Nft URLs.

# Changes

Added utility and query param consts for the new params

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/utils/nav.utils.ts
+++ b/src/frontend/src/lib/utils/nav.utils.ts
@@ -171,34 +171,31 @@ export const switchNetwork = async (networkId: Option<NetworkId>) => {
 	await goto(url, { replaceState: true, noScroll: true });
 };
 
-export const buildNftSearchUrl = ({
-	nft,
-	collection,
-	fromRoute
-}: {
-	nft?: Nft;
-	collection?: NftCollection;
-	fromRoute: NavigationTarget | null;
-}): string | undefined => {
-	if (nonNullish(nft?.collection?.network?.id?.description)) {
-		fromRoute?.url.searchParams.set(NETWORK_PARAM, nft.collection.network.id.description);
-	} else if (nonNullish(collection?.network?.id?.description)) {
-		fromRoute?.url.searchParams.set(NETWORK_PARAM, collection.network.id.description);
-	} else {
-		const currentNetwork = fromRoute?.url.searchParams.get(NETWORK_PARAM);
-		if (nonNullish(currentNetwork)) {
-			fromRoute?.url.searchParams.set(NETWORK_PARAM, currentNetwork);
+export const nftsUrl = (
+	params: {
+		fromRoute: NavigationTarget | null;
+	} & (
+		| {
+				nft?: Nft;
+		  }
+		| {
+				collection?: NftCollection;
+		  }
+	)
+): string | undefined => {
+	const { fromRoute } = params;
+
+	if ('nft' in params && nonNullish(params.nft)) {
+		fromRoute?.url.searchParams.set(NFT_PARAM, params.nft.id);
+		fromRoute?.url.searchParams.set(COLLECTION_PARAM, params.nft.collection.address);
+		if (nonNullish(params.nft.collection.network.id.description)) {
+			fromRoute?.url.searchParams.set(NETWORK_PARAM, params.nft.collection.network.id.description);
 		}
-	}
-
-	if (nonNullish(nft?.collection)) {
-		fromRoute?.url.searchParams.set(COLLECTION_PARAM, nft.collection.address);
-	} else if (nonNullish(collection)) {
-		fromRoute?.url.searchParams.set(COLLECTION_PARAM, collection.address);
-	}
-
-	if (nonNullish(nft)) {
-		fromRoute?.url.searchParams.set(NFT_PARAM, nft.id);
+	} else if ('collection' in params && nonNullish(params.collection)) {
+		fromRoute?.url.searchParams.set(COLLECTION_PARAM, params.collection.address);
+		if (nonNullish(params.collection.network.id.description)) {
+			fromRoute?.url.searchParams.set(NETWORK_PARAM, params.collection.network.id.description);
+		}
 	}
 
 	return fromRoute?.url.toString();
--- a/src/frontend/src/tests/lib/utils/nav.utils.spec.ts
+++ b/src/frontend/src/tests/lib/utils/nav.utils.spec.ts
@@ -14,7 +14,6 @@ import {
 import type { Network } from '$lib/types/network';
 import {
 	back,
-	buildNftSearchUrl,
 	gotoReplaceRoot,
 	isActivityPath,
 	isDappExplorerPath,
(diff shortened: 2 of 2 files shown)


src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -167,3 +170,31 @@ export const switchNetwork = async (networkId: Option<NetworkId>) => {
...
+	} else if (nonNullish(collection?.network?.id?.description)) {
+		url.searchParams.set(NETWORK_PARAM, collection.network.id.description);
+	}
+
+	if (nonNullish(collection)) {
+		url.searchParams.set(COLLECTION_PARAM, collection.address);
+	}
+
+	if (nonNullish(nft)) {
+		url.searchParams.set(NFT_PARAM, String(nft.id));
```
roman-nazaruk: I think it can be nft.id without String

src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -167,3 +170,31 @@ export const switchNetwork = async (networkId: Option<NetworkId>) => {
 
 	await goto(url, { replaceState: true, noScroll: true });
 };
+
+export const buildNftSearchUrl = ({
```
AntonioVentilii: I am confused: is it to filter the NFT or to select it (similar to what we do for fungible tokens)?

src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -167,3 +170,31 @@ export const switchNetwork = async (networkId: Option<NetworkId>) => {
...
+	nft,
+	collection,
+	network
+}: {
+	nft?: Nft;
+	collection?: NftCollection;
+	network?: Network;
+}) => {
+	const url = new URL(`${window.location.origin}${AppPath.Nfts}`);
```
AntonioVentilii: better to pass this, let's avoid taking it directly from the `window` object. Similar to `networkUrl`

src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -167,3 +170,31 @@ export const switchNetwork = async (networkId: Option<NetworkId>) => {
...
+
+	if (nonNullish(network?.id?.description)) {
+		url.searchParams.set(NETWORK_PARAM, network.id.description);
+	} else if (nonNullish(collection?.network?.id?.description)) {
+		url.searchParams.set(NETWORK_PARAM, collection.network.id.description);
+	}
```
AntonioVentilii: shouldn't the network be directly passed via nft param or via collection? for example

```suggestion
	if (nonNullish(nft?.network?.id?.description)) {
		url.searchParams.set(NETWORK_PARAM, network.id.description);
	} else if (nonNullish(collection?.network?.id?.description)) {
		url.searchParams.set(NETWORK_PARAM, collection.network.id.description);
	}
```

src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -167,3 +170,31 @@ export const switchNetwork = async (networkId: Option<NetworkId>) => {
...
+
+	if (nonNullish(collection)) {
+		url.searchParams.set(COLLECTION_PARAM, collection.address);
+	}
```
AntonioVentilii: I am wondering if the input param should be either nft non-nullish or collection non-nullish

for example:

```ts
params : {
	nft: Nft;
} | {
	collection: NftCollection;
}
```

because if you have the NFT, you don't need the collection, you can use `nft.collection`, no? and you need at least one of the two

src/frontend/src/lib/utils/nav.utils.ts
daviddecentage: I took inspiration from switchNetwork which does the same

src/frontend/src/lib/utils/nav.utils.ts
daviddecentage: updated it

src/frontend/src/lib/utils/nav.utils.ts
daviddecentage: updated it

src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -167,3 +170,36 @@ export const switchNetwork = async (networkId: Option<NetworkId>) => {
...
+
+export const buildNftSearchUrl = ({
+	nft,
+	collection,
+	fromRoute
+}: {
+	nft?: Nft;
+	collection?: NftCollection;
+	fromRoute: NavigationTarget | null;
+}): string | undefined => {
```
AntonioVentilii: what about

```suggestion
export const buildNftSearchUrl = (
	params: {
		fromRoute: NavigationTarget | null;
	} & (
		| {
				nft?: Nft;
		  }
		| {
				collection?: NftCollection;
		  }
	)
): string | undefined => {
```

src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -167,3 +170,36 @@ export const switchNetwork = async (networkId: Option<NetworkId>) => {
...
+	if (nonNullish(nft?.collection?.network?.id?.description)) {
+		fromRoute?.url.searchParams.set(NETWORK_PARAM, nft.collection.network.id.description);
+	} else if (nonNullish(collection?.network?.id?.description)) {
+		fromRoute?.url.searchParams.set(NETWORK_PARAM, collection.network.id.description);
+	} else {
+		const currentNetwork = fromRoute?.url.searchParams.get(NETWORK_PARAM);
+		if (nonNullish(currentNetwork)) {
+			fromRoute?.url.searchParams.set(NETWORK_PARAM, currentNetwork);
+		}
+	}
```
AntonioVentilii: is this part necessary? you are re-assigning the same value, it seems, am I reading it correctly?

src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -167,3 +170,36 @@ export const switchNetwork = async (networkId: Option<NetworkId>) => {
 
 	await goto(url, { replaceState: true, noScroll: true });
 };
+
+export const buildNftSearchUrl = ({
```
AntonioVentilii: will this function be ONLY for search? or for any generic NFT URL?

src/frontend/src/lib/utils/nav.utils.ts
daviddecentage: I had to refactor the whole function a bit for this

src/frontend/src/lib/utils/nav.utils.ts
daviddecentage: Any Nft URL, the naming was a bit misleading. Refactored to nftsUrl analogous to networkUrl

src/frontend/src/lib/utils/nav.utils.ts
daviddecentage: Right, not needed.

src/frontend/src/lib/utils/nav.utils.ts
daviddecentage: Its to build any nft page URL

PR #9838, opened 2025-10-21
feat(build): Unbundle i18n JSON files when chunking

AntonioVentilii:
# Motivation

As per @peterpeterparker's suggestion, we are going to lazy-load the i18n JSON files, instead of letting the user load them all at once. Before that we need to guarantee that such files are at least not bundled together in the same chunks.

### Before --> 2 chunks (biggest 5.4 MB)

Screenshot 2025-10-22 at 10 39 36

### After --> 16 chunks (biggest 4 MB)

Screenshot 2025-10-22 at 10 57 34

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -26,7 +26,7 @@ const config: UserConfig = {
 				manualChunks: (id) => {
 					const folder = dirname(id);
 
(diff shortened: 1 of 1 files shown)


AntonioVentilii:
Superseded by:

- https://github.com/dfinity/oisy-wallet/pull/9839

AntonioVentilii:
Reopening this PR, since we actually require smarter chunks

PR #9840, opened 2025-10-21
feat(frontend): Lazy-load the i18n JSON files

AntonioVentilii:
# Motivation

As per @peterpeterparker 's suggestion, inspired from https://github.com/junobuild/juno/blob/main/src/frontend/src/lib/stores/i18n.store.ts, we can avoid loading ALL i18n JSON files for each user, and actually lazy-load them according to the user's choice.


src/frontend/src/lib/components/core/LanguageDropdown.svelte
```diff
@@ -14,8 +14,8 @@
 
 	const currentLang: string = $derived(LANGUAGES[$currentLanguage]);
 
-	const handleLangChange = (lang: string) => {
-		i18n.switchLang(Languages[lang as keyof typeof Languages]);
+	const handleLangChange = async (lang: string) => {
```
peterpeterparker, line 17: Unrelated to the PR: shoudln't `lang: string` be typed to avoid the cast?

src/frontend/src/lib/stores/i18n.store.ts
```diff
@@ -22,95 +12,39 @@ export const enI18n = (): I18n => ({
...
-const frI18n = (): I18n => ({
-	...mergeWithFallback({ refLang: enI18n(), targetLang: fr as I18n }),
-	lang: Languages.FRENCH
-});
-
-const hiI18n = (): I18n => ({
-	...mergeWithFallback({ refLang: enI18n(), targetLang: hi as I18n }),
-	lang: Languages.HINDI
-});
+	const targetLang = langMod.default as I18n;
```
peterpeterparker: Can we avoid the cast? In Juno, I think I do not and it's really useful because letting TypeScript infer the types help detecting when keys have been declared in one language but not in the other.

Or, I'm not familiar with `mergeWithFallback`, is the goal to allow translations file to be incomplete? In that case, shouldn't  the cast be `Partial<I18n>`?

src/frontend/src/lib/stores/i18n.store.ts
```diff
@@ -22,95 +12,39 @@ export const enI18n = (): I18n => ({
...
+	const { subscribe, set } = writable<I18n>(enI18n());
+
+	const switchLang = async ({ lang, track = true }: { lang: Languages; track?: boolean }) => {
+		const language = await loadLang(lang);
 
-	const switchLang = ({ lang, track = true }: { lang: Languages; track?: boolean }) => {
-		set(loadLang(lang));
+		set(language);
 
 		if (track) {
```
peterpeterparker, line 47: Unrelated: is it expected to have an analytics event within the store? Maybe opiniated but, it makes the store impure which is a bit meh from my point of view.

src/frontend/src/lib/stores/i18n.store.ts
```diff
@@ -22,95 +12,39 @@ export const enI18n = (): I18n => ({
...
-			return enI18n();
+const loadLang = async (lang: Languages): Promise<I18n> => {
+	try {
+		return await loadLangI18n(lang);
+	} catch (_: unknown) {
+		return enI18n();
 	}
 };
 
 const saveLang = (lang: Languages) => set({ key: 'lang', value: lang });
```
peterpeterparker, line 32: Unrelated: is there any particular reason why you do not reflect the languague at the document level?

e.g. if I switch to French, the HTML still mention english `<html lang="en" />`.

src/frontend/src/lib/components/core/LanguageDropdown.svelte
AntonioVentilii: yep! it should use Languages: separate PR or this one?

src/frontend/src/lib/stores/i18n.store.ts
AntonioVentilii: I remember you suggested it: I still haven't found the time to do it honestly eheh! Will do as follow-up, tks for the suggestion!

src/frontend/src/lib/stores/i18n.store.ts
AntonioVentilii: Good point! I will move it in the consumer: after or before this PR? Any preference?

src/frontend/src/lib/stores/i18n.store.ts
AntonioVentilii: Removed casting, right!

src/frontend/src/lib/components/core/LanguageDropdown.svelte
peterpeterparker: I'm fine either way.

src/frontend/src/lib/stores/i18n.store.ts
peterpeterparker: Separate PR, it's unrelated to lazy loading translations.

src/frontend/src/lib/stores/i18n.store.ts
AntonioVentilii: `mergeWithFallback` --> actually we are putting empty strings in the missing translations, so the type should be still `I18n` to guarantee that ALL keys are present in the JSON file. The function will simply use the english values to overwrite the empty string.

It is useful for the translator to know which translations are missing.

It could even be done with missing keys, maybe another iteration

src/frontend/src/lib/stores/i18n.store.ts
peterpeterparker: Great! I'm just assuming it's useful for accessibility reasons, notably screen readers.

src/frontend/src/lib/components/core/LanguageDropdown.svelte
AntonioVentilii: Better in another PR, I see that it requires some more meddling around, enough to deserve its own PR

AntonioVentilii: Done here:

- https://github.com/dfinity/oisy-wallet/pull/9854

src/frontend/src/lib/stores/i18n.store.ts
AntonioVentilii: Done with:

- https://github.com/dfinity/oisy-wallet/pull/9856

src/frontend/src/lib/stores/i18n.store.ts
AntonioVentilii: Done with:

- https://github.com/dfinity/oisy-wallet/pull/9865
- https://github.com/dfinity/oisy-wallet/pull/9866

PR #9854, opened 2025-10-22
fix(frontend): Type-safe values when switching languages

AntonioVentilii:
# Motivation

As per suggestion https://github.com/dfinity/oisy-wallet/pull/9840/#discussion_r2450478419, we should avoid casting when switching languages and use directly the type `Languages` that collects all available values.


src/frontend/src/lib/components/core/LanguageDropdown.svelte
```diff
@@ -7,15 +7,15 @@
 	import Dropdown from '$lib/components/ui/Dropdown.svelte';
 	import { LANGUAGE_DROPDOWN } from '$lib/constants/test-ids.constants';
 	import { currentLanguage } from '$lib/derived/i18n.derived';
-	import { Languages } from '$lib/enums/languages';
+	import type { Languages } from '$lib/enums/languages';
```
peterpeterparker, line 10: TIL an enum can be imported as type, did not knew

peterpeterparker (approved):
LGTM, thx

PR #9855, opened 2025-10-22
fix(frontend): Use correct map when modifying tokens in Assets page

AntonioVentilii:
# Motivation

When we manage tokens in the Assets page through the filter, we are currently indexing them with the `TokenId` that is a `symbol` object. However it is quite complicated to extract the values from a `Record` indexed with symbols.

So, we can use a SvelteMap and simply using its built-in methods.

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensList.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensList.svelte
@@ -103,9 +103,9 @@
 
 	let tokensToBeSaved = $derived([...modifiedTokens.values()]);
 
-	let tokensToBeSavedLen = $derived(tokensToBeSaved.length);
+	let tokensToBeSavedLength = $derived(tokensToBeSaved.length);
 
-	let saveDisabled = $derived(tokensToBeSavedLen === 0);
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -91,31 +92,30 @@
...
 	};
 
-	let modifiedTokens: Record<TokenId, Token> = $state({});
+	const modifiedTokens = new SvelteMap<TokenId, Token>();
 
-	let modifiedTokensLen = $derived(Object.getOwnPropertySymbols(modifiedTokens).length);
+	let tokensToBeSaved = $derived([...modifiedTokens.values()]);
 
-	let saveDisabled = $derived(modifiedTokensLen === 0);
+	let tokensToBeSavedLen = $derived(tokensToBeSaved.length);
```
DenysKarmazynDFINITY: nitpick: I know it was like this before, but `tokensToBeSavedLength` is way more nicer for the eyes 😄

AntonioVentilii: very right! done

PR #9856, opened 2025-10-22
refactor(frontend): Move tracking of switch language to consumer

AntonioVentilii:
# Motivation

As per suggestion https://github.com/dfinity/oisy-wallet/pull/9840/#discussion_r2450488949, to have a cleaner `i18n` store, we should avoid doing the tracking of events directly in the function that switches languages.

Instead, it should be moved to the consumer. It even have the advantage to customize the event per-consumer.

NOTE: the event was being tracked ONLY when using method `i18n.switchLang` and not `i18n.init`

# Changes

- Move the tracking from the `i18n` store to the consumer `LanguageDropdown` component.
- Remove unnecessary `track` params from `switchLang` method.

# Tests

Current tests should work as expected: we have already tests in the component `LanguageDropdown` for the event tracking.

### DIFF
--- a/src/frontend/src/lib/stores/i18n.store.ts
+++ b/src/frontend/src/lib/stores/i18n.store.ts
@@ -59,7 +59,7 @@ const initI18n = (): I18nStore => {
 			await switchLang(lang);
 		},
 
-		switchLang: (lang: Languages) => switchLang(lang)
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
LGTM, thx

PR #9865, opened 2025-10-22
feat(frontend): Set language at HTML properties when switching

AntonioVentilii:
# Motivation

As suggested in https://github.com/dfinity/oisy-wallet/pull/9840#discussion_r2450495660 , we should set the language as property of the HTML document too, when switching.

Screenshot 2025-10-22 at 12 00 25

### DIFF
--- a/src/frontend/src/lib/stores/i18n.store.ts
+++ b/src/frontend/src/lib/stores/i18n.store.ts
@@ -26,9 +26,9 @@ const loadLang = async (lang: Languages): Promise<I18n> => {
 	}
 };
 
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
LGTM, thx!

Just sharing what crossed my mind:

- I asked myself if it would be nice to have a function for it, since we have `saveLang` what not an `updateHtmlLang`

- No test?

AntonioVentilii:
@peterpeterparker 

- good point! changed to have `updateHtmlLang`

- alas, no... still missing for the entire `i18n` store. We are testing it only in the consumers. I will provide them as follow-up. I just did a manual test (see screenshot).

PR #9866, opened 2025-10-22
feat(build): Set storage language when loading HTML document

AntonioVentilii:
# Motivation

It is useful to load the language at document level if it is already cached in the storage.


peterpeterparker (approved):
LGTM, thx

PR #9867, opened 2025-10-22
feat(frontend): Adds fromRoute to Nft components

daviddecentage:
# Motivation

In preparation for a future refactoring of the Nft routing, we add fromRoute props to Nft components that need it for building URLs to other Nft pages.

# Changes

Adds new props to several Nft components

### DIFF
--- a/src/frontend/src/lib/components/nfts/Nft.svelte
+++ b/src/frontend/src/lib/components/nfts/Nft.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 	import { isNullish, nonNullish } from '@dfinity/utils';
+	import type { NavigationTarget } from '@sveltejs/kit';
 	import { onMount } from 'svelte';
 	import { afterNavigate, goto } from '$app/navigation';
 	import { page } from '$app/state';
@@ -12,7 +13,6 @@
 	import { i18n } from '$lib/stores/i18n.store';
 	import { toastsError } from '$lib/stores/toasts.store';
 	import type { NonFungibleToken } from '$lib/types/nft';
-	import type { NavigationTarget } from '@sveltejs/kit';
 
 	const nft = $derived($pageNft);
 
--- a/src/frontend/src/lib/components/nfts/Nft.svelte
+++ b/src/frontend/src/lib/components/nfts/Nft.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { isNullish, nonNullish } from '@dfinity/utils';
 	import { onMount } from 'svelte';
-	import { goto } from '$app/navigation';
+	import { afterNavigate, goto } from '$app/navigation';
 	import { page } from '$app/state';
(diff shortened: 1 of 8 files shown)


src/frontend/src/lib/components/nfts/Nft.svelte
```diff
@@ -34,8 +35,14 @@
...
+
+	let fromRoute = $state<NavigationTarget | null>(null);
+
+	afterNavigate(({ from }) => {
+		fromRoute = from;
+	});
 </script>
 
-<NftHero {nft} {token} />
+<NftHero {fromRoute} {nft} {token} />
```
AntonioVentilii, line 46: why passing it though? can it not be done directly in `NftHero`?

src/frontend/src/lib/components/nfts/NftCollection.svelte
```diff
@@ -54,6 +61,7 @@
 	{#if collectionNfts.length > 0}
 		{#each collectionNfts as nft, index (`${nft.id}-${index}`)}
 			<NftCard
+				{fromRoute}
```
AntonioVentilii, line 64: same question

src/frontend/src/lib/components/nfts/NftCollectionList.svelte
```diff
@@ -27,7 +29,7 @@
 		{#if notEmptyCollections.length > 0}
 			<div class="grid grid-cols-2 gap-3 gap-y-4 py-4 md:grid-cols-3">
 				{#each notEmptyCollections as collection, index (`${String(collection.collection.id)}-${index}`)}
-					<NftCollectionCard {collection} />
+					<NftCollectionCard {collection} {fromRoute} />
```
AntonioVentilii, line 32: same question

src/frontend/src/lib/components/nfts/Nft.svelte
daviddecentage: I thought its a good idea so we dont have to call onNavigate each time on every single component.

Idea is we call it on the root "pages" of the Nft pages and pass it down.

Like this we will always have 3 states and 3 onNavigates, instead of multiple onNavigates on every place we need it.

AntonioVentilii: Good point! Does it require much computation, since it is done in `afterNavigate`? For example, can it be done in each component that calls `buildNftUrl`? Performance-wise is it better?

daviddecentage: No, we dont do anything in afterNavigate besides setting a state. And that fromRoute is then only used to set query params.

If we would do any computation inside afterNavigate, my approach here would be better since we only use afterNavigate on the root components and therfor reduce number of function calls by passing as props instead. But the amount of work done is negligible, so the only difference is that I can do this in a separate PR to make the main feature PR slimmer

PR #9870, opened 2025-10-22
feat(frontend): added event to track swap offer

roman-nazaruk:
# Motivation

We aim to track new events for swap offers with result_statuses[success, error]

# Changes

Added new events swap_offer for both IcSwaps and EvmSwaps with result_statuses[success, error]

# Tests

Covered new logic with tests

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,22 +7,20 @@ updates:
     directory: '/'
     schedule:
       interval: weekly
-    cooldown:
-      - default-days: 7
 
   - package-ecosystem: cargo
     directory: '/'
     schedule:
       interval: weekly
     cooldown:
-      - default-days: 7
+      default-days: 7
 
   - package-ecosystem: npm
     directory: '/'
     schedule:
       interval: weekly
     cooldown:
-      - default-days: 7
+      default-days: 7
     allow:
       - dependency-type: all
     ignore:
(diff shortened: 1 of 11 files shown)


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -664,19 +722,94 @@ const fetchVeloraSwapAmount = async ({
...
+		token2_id: String(destinationToken.id),
+		...(nonNullish(sourceTokenUsdValue) && {
+			token_usd_value: `${sourceTokenUsdValue * Number(sourceTokenToDecimals)}`
+		})
+	};
 
-	if ('market' in data) {
-		return mapVeloraMarketSwapResult(data.market);
-	}
+	try {
```
AntonioVentilii, line 751: is the try/catch a new condition? may you please do it in another PR?

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -260,9 +263,9 @@ const fetchSwapAmountsICP = async ({
 	tokens,
 	slippage,
 	isSourceTokenIcrc2
-}: Omit<FetchSwapAmountsParams, 'userEthAddress' | 'amount'> & { amount: bigint }): Promise<
-	SwapMappedResult[]
-> => {
+}: Omit<FetchSwapAmountsParams, 'userEthAddress' | 'amount'> & {
+	amount: bigint;
+}): Promise<SwapMappedResult[]> => {
```
AntonioVentilii, line 268: may you please do this change in another PR?

PR #9873, opened 2025-10-22
feat(lint): Temporary disable Svelte rule `no-navigation-without-resolve`

AntonioVentilii:
# Motivation

There is a new rule in the `recommended` suite of Svelte Lint plugin: [no-navigation-without-resolve](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-navigation-without-resolve/).

The rule enforces resolving the internal URLs. However, in our code it raises quite a lot of warnings (see CI https://github.com/dfinity/oisy-wallet/actions/runs/18709289138/job/53353612091?pr=9761).

We have an higher priority to bump the OISY signer, so for now we simply disable the rule, and work on the fixes step-by-step.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -32,6 +32,13 @@ export default [
 		}
 	},
 
+	// TODO: re-enable this rule when we fix all the warnings that it causes.
+	{
+		rules: {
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
THX A TON!

PR #9876, opened 2025-10-22
refactor(frontend): Extract PWA Banner to its own component

AntonioVentilii:
# Motivation

We are going to re-use the PWA Warning Banner: so for now, we simply extract it to its own component.

### DIFF
--- a/src/frontend/src/lib/components/core/PwaBanner.svelte
+++ b/src/frontend/src/lib/components/core/PwaBanner.svelte
@@ -1,6 +1,10 @@
 <script lang="ts">
 	import { Html, IconClose } from '@dfinity/gix-components';
 	import WarningBanner from '$lib/components/ui/WarningBanner.svelte';
+	import {
+		PWA_WARNING_BANNER_CLOSE_BUTTON_TEST_ID,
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/core/PwaBanner.svelte
```diff
@@ -0,0 +1,27 @@
...
+	import { isPWAStandalone } from '$lib/utils/device.utils';
+	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils.js';
+
+	let pwaBannerVisible = $state(true);
+
+	const closePwaBanner = () => (pwaBannerVisible = false);
+</script>
+
+<!-- TODO remove this WarningBanner again as soon a solution is found for enabling display type standalone  -->
+{#if isPWAStandalone() && pwaBannerVisible}
```
DenysKarmazynDFINITY, line 18: let's maybe add a minimal test case for the component, just for the sake of the coverage percentage

AntonioVentilii: good point! added!

PR #9877, opened 2025-10-22
feat(frontend): Move PWA Banner to sign page of OISY

AntonioVentilii:
# Motivation

Credits to @sbpublic :

We don't really need the PWA warning banner in the entire OISY, in case of a PWA. The real issue with PWA is born with missing connections during the OISY signer flow. So, we can show it ONLY in the `sign` routes.

Screenshot 2025-10-22 at 14 34 40

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,14 +13,14 @@ updates:
     schedule:
       interval: weekly
     cooldown:
-      - default-days: 7
+      default-days: 7
 
   - package-ecosystem: npm
     directory: '/'
     schedule:
       interval: weekly
     cooldown:
-      - default-days: 7
+      default-days: 7
     allow:
       - dependency-type: all
     ignore:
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/components/core/Banner.svelte
```diff
@@ -37,20 +32,7 @@
...
-			<span class="w-full px-2">
-				<Html text={replaceOisyPlaceholders($i18n.core.warning.standalone_mode)} />
-			</span>
-			<button aria-label={$i18n.core.text.close} onclick={closePwaBanner}>
-				<IconClose />
-			</button>
-		</WarningBanner>
-	</div>
-{/if}
+<PwaBanner />
```
AntonioVentilii:
```suggestion
```

src/frontend/src/lib/components/core/Banner.svelte
```diff
@@ -1,18 +1,13 @@
 <script lang="ts">
-	import { Html, IconClose, IconWarning } from '@dfinity/gix-components';
+	import { IconClose, IconWarning } from '@dfinity/gix-components';
+	import PwaBanner from '$lib/components/core/PwaBanner.svelte';
```
AntonioVentilii:
```suggestion
```

PR #9881, opened 2025-10-22
fix(frontend): Nft collection card button

daviddecentage:
# Motivation

To be consistent and for an upcoming refactoring of the routing, we need collection card components to be using button elements instead of links.

# Changes

- Adjusted element
- moved click handler function declaration
- added block styling to some inner spans that had to be changed from div (since div cannot be a child of button)

# Tests

image

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensList.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensList.svelte
@@ -2,6 +2,7 @@
 	import { debounce, isNullish, nonNullish } from '@dfinity/utils';
 	import { untrack } from 'svelte';
 	import { flip } from 'svelte/animate';
+	import { SvelteMap } from 'svelte/reactivity';
 	import { fade } from 'svelte/transition';
 	import { goto } from '$app/navigation';
 	import NoTokensPlaceholder from '$lib/components/tokens/NoTokensPlaceholder.svelte';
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
```diff
@@ -59,9 +53,19 @@
 				token_name: previewNft.collection.name ?? ''
 			}
 		});
-	}}
+
+		goto(`${AppPath.Nfts}${collection.collection.network.name}-${collection.collection.address}`);
+	};
+</script>
+
+<button
```
AntonioVentilii, line 61: does the existing `<Button>` component fits somehow?

daviddecentage: When I initially changed Nft card to a button I had tried it. The problem is our existing button applies base styles by default that clash with the styling we need here (and also the HTML structure inside).

The Button component is really optimized for normal button appearance.

AntonioVentilii (approved):

src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
AntonioVentilii: tks for checking!

PR #9889, opened 2025-10-22
feat(frontend): Nft assets page routing

daviddecentage:
# Motivation

We add the "routing" by query params to the assets page. This will currently have no effect but once the routing gets switched over, these act as the collection and nft pages.

# Changes

Added routing conditions and components

# Tests

Since the pages rely on the routing params, the functionality of the routing is already covered heavily. Also, this doesnt change the routing to query params yet, so we need to adjust the tests once the routing has switched to query params.

### DIFF
--- a/src/frontend/src/lib/services/swap.services.ts
+++ b/src/frontend/src/lib/services/swap.services.ts
@@ -664,19 +664,23 @@ const fetchVeloraSwapAmount = async ({
 		partner: OISY_URL_HOSTNAME
 	};
 
-	const data = await sdk.quote.getQuote(
-		srcChainId !== destChainId ? { ...baseParams, destChainId: Number(destChainId) } : baseParams
-	);
+	try {
+		const data = await sdk.quote.getQuote(
+			srcChainId !== destChainId ? { ...baseParams, destChainId: Number(destChainId) } : baseParams
+		);
 
-	if ('delta' in data) {
-		return mapVeloraSwapResult(data);
+		if ('delta' in data) {
+			return mapVeloraSwapResult(data);
+		}
+
+		if ('market' in data) {
+			return mapVeloraMarketSwapResult(data.market);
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/tokens/Assets.svelte
```diff
@@ -35,65 +38,75 @@
...
-				</div>
-				{#if tab === TokenTypes.TOKENS}
-					<div class="flex">
-						<TokensMenu />
-					</div>
-				{:else}
-					<div class="flex">
-						<NftSortMenu />
+{#if nonNullish($routeNft) && nonNullish($routeCollection) && nonNullish($routeNetwork)}
+	<Nft />
```
roman-nazaruk, line 42: Should we have FEATURE_FLAGS to have ability to disable this routes?

daviddecentage: Updated

src/frontend/src/lib/derived/page-nft.derived.ts
```diff
@@ -16,10 +16,17 @@ export const pageCollectionNfts: Readable<Nft[]> = derived(
...
 					address,
-					network: { name: networkName }
+					network: {
+						name: networkName,
+						id: { description: networkId }
+					}
 				}
-				// TODO: Confirm that `$routeNftNetwork` is the network name (not the ID) when comparing to `networkName` here.
-			}) => address === $routeCollection && networkName === $routeNftNetwork
+				// TODO: Remove check by network name once routing refactoring has been completed
```
roman-nazaruk, line 24: This logic should be in another PR, and what I assume you are already created PR for this

daviddecentage: yes it wasnt merged. Updated

PR #9900, opened 2025-10-22
fix(frontend): Switch to default language if not set

AntonioVentilii:
# Motivation

There is an issue when the user has the storage language set the same as the machine default language: the i18n store does not initialize with the correct language:

Capture d’écran 2025-10-22 à 17 31 52

To fix it we remove the condition that was impending the correct switch.

NOTE: `switchLang` already use `saveLang` internally.

Thanks to @peterpeterparker for the heads-up!

### DIFF
--- a/src/frontend/src/lib/stores/i18n.store.ts
+++ b/src/frontend/src/lib/stores/i18n.store.ts
@@ -54,6 +54,14 @@ const initI18n = (): I18nStore => {
 		init: async () => {
 			const lang = get<Languages>({ key: 'lang' }) ?? getDefaultLang();
 
+			// English is the default one in case no language is set.
+			// Or either way is what most users would have as default in their machines.
+			if (lang === Languages.ENGLISH) {
+				saveLang(lang);
+				// No need to reload the store, store is already initialised with the default
+				return;
+			}
+
 			await switchLang(lang);
 		},
 


src/frontend/src/lib/stores/i18n.store.ts
```diff
@@ -54,12 +54,6 @@ const initI18n = (): I18nStore => {
 		init: async () => {
 			const lang = get<Languages>({ key: 'lang' }) ?? getDefaultLang();
 
-			if (lang === getDefaultLang()) {
```
peterpeterparker, line 57: I was thinking doing `lang === 'en'` because it's the default -> avoid updating the state with it's default value. Would that work as well?

```
init: async () => {
			const lang = get<Languages>({ key: 'lang' }) ?? getDefaultLang();

			if (lang === "en") {
				saveLang(lang);
				// No need to reload the store, store is already initialised with the default
				return;
			}

			await switchLang(lang);
		},
```

AntonioVentilii: right, it makes sense, but somehow it feels like we are doing an exception because we know that exception, you know what I mean?

BUT, it is basically most of the users, so it is helpful

Boh, no strong opinion, so will go with the suggestion! Thank you!

AntonioVentilii: Done, have a look please

peterpeterparker: Yes but we know because we already made the choice by using english for default - the store does not start empty.

PR #9922, opened 2025-10-23
refactor(frontend): Remove from route nft collections

daviddecentage:
# Motivation

We wrongly used fromRoute to determine and set the query params for Nft pages. Here we remove it for the Nft collection components.

# Changes

Remove routeFrom prop and conditional checks for nullish urls.


AntonioVentilii (approved):
Thank you for that!

PR #9925, opened 2025-10-23
fix(frontend): Persist query params on assets tabs

daviddecentage:
# Motivation

Currently, on the assets page when we switch between tokens and Nfts, the network query param is lost.

# Changes

Add the url.search to the tabs paths.

### DIFF
--- a/src/frontend/src/lib/components/tokens/Assets.svelte
+++ b/src/frontend/src/lib/components/tokens/Assets.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
-	import { page } from '$app/stores';
 	import { nonNullish } from '@dfinity/utils';
 	import { fade } from 'svelte/transition';
+	import { page } from '$app/stores';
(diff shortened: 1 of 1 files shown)


AntonioVentilii (approved):

src/frontend/src/lib/components/tokens/Assets.svelte
```diff
@@ -58,9 +59,13 @@
 											{
 												label: $i18n.tokens.text.title,
 												id: TokenTypes.TOKENS,
-												path: AppPath.Tokens
+												path: `${AppPath.Tokens}${$page.url.search}`
```
AntonioVentilii, line 62: NITPICK, if you prefer you can create a derived for it

daviddecentage: Lets wait for the refactoring to be done. If we find other places that would benefit from it we can certainly add it as a derived.

PR #9934, opened 2025-10-24
feat(frontend): Nft pages context

daviddecentage:
# Motivation

We need an easy way to propagate values across all Nft pages, so we add a context.

# Changes

Added store for the context
Added context component

# Tests

Added tests

### DIFF
--- a/src/frontend/src/tests/lib/components/nfts/NftPagesContext.spec.ts
+++ b/src/frontend/src/tests/lib/components/nfts/NftPagesContext.spec.ts
@@ -17,7 +17,6 @@ const setOriginSelectedNetwork = vi.fn();
 const subscribe = vi.fn(() => () => {});
 
 vi.mock('$lib/stores/nft-pages.store', async (importOriginal) => {
-	//
 	const actual = (await importOriginal()) as typeof NftPagesStoreModule;
 	return {
 		...actual,


src/frontend/src/lib/components/nfts/NftPagesContext.svelte
```diff
@@ -0,0 +1,19 @@
...
+	interface Props {
+		children: Snippet;
+	}
+
+	const { children }: Props = $props();
+
+	const { store } = getContext<NftPagesContext>(NFT_PAGES_CONTEXT_KEY);
+
+	$effect(() => {
+		store.setOriginSelectedNetwork($selectedNetwork?.id);
```
roman-nazaruk: I would bettter make and isNullish and return here

daviddecentage: updated

src/frontend/src/lib/stores/nft-pages.store.ts
```diff
@@ -0,0 +1,31 @@
+import type { NetworkId } from '$lib/types/network';
```
roman-nazaruk: Also I would split it to have a Context and store in 2 diff PRs

daviddecentage: #9937

src/frontend/src/tests/lib/components/nfts/NftPagesContext.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+			return () => {};
+		}
+	}
+}));
+
+const setOriginSelectedNetwork = vi.fn();
+const subscribe = vi.fn(() => () => {});
+
+vi.mock('$lib/stores/nft-pages.store', async (importOriginal) => {
+	//
```
roman-nazaruk: I think it is not needed

PR #9937, opened 2025-10-24
feat(frontend): Nft pages store for context

daviddecentage:
# Motivation

Adding a store for Nft page data that will be used in a context #9934 

# Changes

Added store

# Tests

Added tests


src/frontend/src/lib/stores/nft-pages.store.ts
```diff
@@ -0,0 +1,31 @@
+import type { NetworkId } from '$lib/types/network';
+import type { Option } from '$lib/types/utils';
+import { writable, type Readable } from 'svelte/store';
+
+export type NftPagesStoreData = Option<{
+	originSelectedNetwork: NetworkId | undefined;
```
DenysKarmazynDFINITY:
```suggestion
	originSelectedNetwork: OptionNetworkId;
```

DenysKarmazynDFINITY: and in other places too, please

PR #9949, opened 2025-10-24
feat(frontend): Adds asset tab to nft pages store

daviddecentage:
# Motivation

In order to persist which tab the user was on on the assets page, we store it in the Nft page context.

# Changes

Added new assetsTab field

# Tests

Extended tests

### DIFF
--- a/src/frontend/src/lib/stores/nft-pages.store.ts
+++ b/src/frontend/src/lib/stores/nft-pages.store.ts
@@ -4,7 +4,7 @@ import type { Option } from '$lib/types/utils';
 import { writable, type Readable } from 'svelte/store';
 
 export type NftPagesStoreData = Option<{
-	assetsTab: TokenTypes | undefined;
+	assetsTab?: TokenTypes;
 	originSelectedNetwork: OptionNetworkId;
 }>;
 


src/frontend/src/lib/stores/nft-pages.store.ts
```diff
@@ -1,25 +1,41 @@
+import type { TokenTypes } from '$lib/enums/token-types';
 import type { OptionNetworkId } from '$lib/types/network';
 import type { Option } from '$lib/types/utils';
 import { writable, type Readable } from 'svelte/store';
 
 export type NftPagesStoreData = Option<{
+	assetsTab: TokenTypes | undefined;
```
AntonioVentilii:
```suggestion
	assetsTab?: TokenTypes;
```

src/frontend/src/lib/stores/nft-pages.store.ts
```diff
@@ -1,25 +1,41 @@
...
 
 export interface NftPagesStore extends Readable<NftPagesStoreData> {
+	setAssetsTab: (tab: TokenTypes) => void;
 	setOriginSelectedNetwork: (networkId: OptionNetworkId) => void;
 }
 
+const initialStore: NftPagesStoreData = {
+	assetsTab: undefined,
+	originSelectedNetwork: undefined
+};
```
AntonioVentilii, line 19: does this work type-wise?

```suggestion
const initialStore: NftPagesStoreData = {};
```

daviddecentage: No, because the ts checker doesnt get OptionNetworkId I think.

src/frontend/src/lib/stores/nft-pages.store.ts
daviddecentage: updated

PR #9951, opened 2025-10-24
feat(frontend): Urls with network query param only if user-set on nft pages

daviddecentage:
# Motivation

The network query param currently is always included in the breadcrumb links of Nft pages. We only want to include it if the user selected it manually. For this we add the NftPageContext.

# Changes

- Added NftPageContext to layout (will be needed on Assets page and Main navigation menu)
- Adjusted NftHero breadcrumb root Url with context network
- Adjusted NftCollection root Url with context network

# Tests

Added tests to NftHero and NftCollectionHero to ensure its working.

Also adjusted a bunch of tests and the mocking of context which is nessesary due to the changes in the heroes.

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftCollectionHero.svelte
+++ b/src/frontend/src/lib/components/nfts/NftCollectionHero.svelte
@@ -24,7 +24,7 @@
 
 	const { token, nfts }: Props = $props();
 
-	const { store } = getContext<NftPagesContext>(NFT_PAGES_CONTEXT_KEY);
+	const store = getContext<NftPagesContext>(NFT_PAGES_CONTEXT_KEY);
 	const originSelectedNetwork = $derived($store?.originSelectedNetwork ?? undefined);
 
 	const breadcrumbItems = $derived([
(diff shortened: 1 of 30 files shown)


src/frontend/src/lib/components/nfts/NftCollectionHero.svelte
```diff
@@ -21,7 +23,12 @@
 
 	const { token, nfts }: Props = $props();
 
-	const breadcrumbItems = $derived([{ label: $i18n.navigation.text.tokens, url: AppPath.Nfts }]);
+	const store = getContext<NftPagesContext>(NFT_PAGES_CONTEXT_KEY);
+	const originSelectedNetwork = $derived($store?.originSelectedNetwork ?? undefined);
```
AntonioVentilii: nitpick: you could even create a derived inside the context. Easier to re-use

daviddecentage: Great idea! #9958

src/frontend/src/routes/(app)/+layout.svelte
```diff
@@ -72,39 +73,41 @@
...
-										<DappsCarousel />
-									</div>
-								</Responsive>
-							{/if}
-						</NavigationMenu>
-					{/snippet}
-
-					{#if showHero}
-						<Hero />
+				<NftPagesContext>
```
AntonioVentilii, line 76: let's avoid putting it here: better to make it inside the NFT page

daviddecentage: I need the context in the MainNavigation

PR #9953, opened 2025-10-24
fix(frontend): Nft pages context/store

daviddecentage:
# Motivation

Destructuring the context value can lead to issues in tests when no context is set. For this we move the store inside the context to the root level of the context.

Since the store itself allows for multple values, this does not affect scalability in the future.

# Changes

Dont use subprop for store inside context


roman-nazaruk (approved):
Thank you so much!

PR #9954, opened 2025-10-24
feat(frontend): Open nft page on collection card if it has only 1 nft

daviddecentage:
# Motivation

When the collection has only 1 nft, the collection card should open the Nft page directly

# Changes

Added condition and goto

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
+++ b/src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
@@ -56,9 +56,9 @@
 
 		if (collection.nfts.length === 1) {
 			goto(nftsUrl({ nft: collection.nfts[0] }));
-		} else {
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
```diff
@@ -54,7 +54,11 @@
 			}
 		});
 
-		goto(nftsUrl({ collection: collection.collection }));
+		if (collection.nfts.length === 1) {
+			goto(nftsUrl({ nft: collection.nfts[0] }));
+		} else {
```
AntonioVentilii:
```suggestion
return
		} 

if {
```

AntonioVentilii (approved):
LGTM tks!

PR #9956, opened 2025-10-24
refactor(frontend): withCollectionLabel prop for NftCard

daviddecentage:
# Motivation

Instead of using the source prop (which is only used for plausible events), we add a dedicated prop for the collection label variation for the Nft card.

# Changes

Adjusted Nft card prop
Adjusted usages

# Tests

Adjusted and added tests


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #9958, opened 2025-10-25
feat(frontend): Add derived values to context Nft pages

daviddecentage:
# Motivation

For easier (one-liner) usage of the context, we add derived values to it returning readables of originSelectedNetwork and assetsTab.

# Changes

- Added derivations

# Tests

Extended tests

### DIFF
--- a/src/frontend/src/lib/stores/nft-pages.store.ts
+++ b/src/frontend/src/lib/stores/nft-pages.store.ts
@@ -21,7 +21,10 @@ const initialStore: NftPagesStoreData = {
 export const initNftPagesStore = (): NftPagesContext => {
 	const store = writable<NftPagesStoreData>(initialStore);
 
-	const originSelectedNetwork = derived([store], ([$store]) => $store?.originSelectedNetwork ?? undefined);
+	const originSelectedNetwork = derived(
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/stores/nft-pages.store.ts
```diff
@@ -18,28 +18,36 @@ const initialStore: NftPagesStoreData = {
 	originSelectedNetwork: undefined
 };
 
-export const initNftPagesStore = (): NftPagesStore => {
-	const { subscribe, update } = writable<NftPagesStoreData>(initialStore);
+export const initNftPagesStore = (): NftPagesContext => {
+	const store = writable<NftPagesStoreData>(initialStore);
+
+	const originSelectedNetwork = derived([store], ([$store]) => $store?.originSelectedNetwork);
```
AntonioVentilii: don't you need it like this?

```suggestion
	const originSelectedNetwork = derived([store], ([$store]) => $store?.originSelectedNetwork ?? undefined);
```

daviddecentage: yes thats right!

PR #9969, opened 2025-10-27
feat(frontend): Remove one loop from ERC1155 loader service

AntonioVentilii:
# Motivation

To try and improve performance, we remove one loop from the ERC1155 custom tokens loader.


AntonioVentilii:
we actually need to avoid raising the errors

PR #9981, opened 2025-10-27
feat(frontend): Nav stores

daviddecentage:
# Motivation

Were adding a nav store to store global navigation relation values, needed for some fixes to Nft routing refactoring.

# Changes

Added store
Added derived stores

# Tests

Added a bunch of tests for stores and derived

### DIFF
--- a/src/frontend/src/lib/derived/nav.derived.ts
+++ b/src/frontend/src/lib/derived/nav.derived.ts
@@ -3,6 +3,8 @@ import { TokenTypes as TokenTypesEnum, type TokenTypes } from '$lib/enums/token-
 import { navStore } from '$lib/stores/nav.store';
 import type { NetworkId } from '$lib/types/network';
 import type { OptionString } from '$lib/types/string';
+import { parseNetworkId } from '$lib/validation/network.validation';
+import { nonNullish } from '@dfinity/utils';
 import { derived, type Readable } from 'svelte/store';
 
 export const routeToken: Readable<OptionString> = derived(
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/stores/nav.store.ts
```diff
@@ -0,0 +1,44 @@
...
+	selectedAssetsTab: undefined,
+	userSelectedNetwork: undefined
+};
+
+const initNavStore = (): NavStore => {
+	const store = writable<NavStoreData>(initialStore);
+
+	return {
+		subscribe: store.subscribe,
+		reset: () => store.set({ ...initialStore }),
```
roman-nazaruk: I think you can just do set(initialStore). Not a strong opinion on it

src/frontend/src/lib/stores/nav.store.ts
```diff
@@ -0,0 +1,44 @@
...
+		setSelectedAssetsTab: (tab: TokenTypes) => {
+			store.update((curr) => ({
```
roman-nazaruk: maybe it is better to name it as activeTab?

roman-nazaruk: or what curr means?

src/frontend/src/lib/derived/nav.derived.ts
```diff
@@ -37,3 +40,13 @@ export const routeNft: Readable<OptionString> = derived(
 		}
 	]) => nft
 );
+
+export const userSelectedNetwork: Readable<NetworkId | undefined> = derived(
+	[navStore],
+	([$navStore]) => $navStore?.userSelectedNetwork ?? undefined
```
roman-nazaruk: Do we need here statement as ?? undefined as it is by default should be undefined

daviddecentage: Yes we need it it can be null as well and I want to unify the undefined case

src/frontend/src/lib/stores/nav.store.ts
daviddecentage: renamed.

curr is the current store element to make partial update

PR #9982, opened 2025-10-27
feat(frontend): Avoid toast message for ERC1155 metadata error

AntonioVentilii:
# Motivation

We prefer to avoid showing an error when the ERC1155 fails on fetching metadata. We will only track it and print it to console.

# Changes

- Create `safeLoadMetadata` sub-service for loading custom ERC1155 tokens.
- Filter from nullish results.

# Tests

Added more tests.

### DIFF
--- a/src/frontend/src/eth/services/erc1155.services.ts
+++ b/src/frontend/src/eth/services/erc1155.services.ts
@@ -93,13 +93,10 @@ const safeLoadMetadata = async ({
 				event_subcontext: PLAUSIBLE_EVENT_SUBCONTEXT_NFT.ERC1155,
 				token_address: address,
 				token_network: `${networkId.description}`
-			}
+			},
+			warning: `Error loading metadata for custom ERC1155 token ${address} on network ${networkId.description}. ${err}`
 		});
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/eth/services/erc1155.services.ts
```diff
@@ -61,6 +74,36 @@ const loadErc1155CustomTokens = async (params: LoadCustomTokenParams): Promise<C
...
+			name: PLAUSIBLE_EVENTS.LOAD_CUSTOM_TOKENS,
+			metadata: {
+				event_context: PLAUSIBLE_EVENT_CONTEXTS.NFT,
+				event_subcontext: PLAUSIBLE_EVENT_SUBCONTEXT_NFT.ERC1155,
+				token_address: address,
+				token_network: `${networkId.description}`
+			}
+		});
+
+		console.error(
```
roman-nazaruk: You can pass error message to trackEvent as warning param

AntonioVentilii: Nice, good one! Done!

PR #9983, opened 2025-10-27
feat(frontend): Avoid toast message for ERC721 metadata error

AntonioVentilii:
# Motivation

We prefer to avoid showing an error when the ERC721 fails on fetching metadata. We will only track it and print it to console.

# Changes

- Create `safeLoadMetadata` sub-service for loading custom ERC721 tokens.
- Filter from nullish results.

# Tests

Added more tests.

### DIFF
--- a/src/frontend/src/lib/enums/plausible.ts
+++ b/src/frontend/src/lib/enums/plausible.ts
@@ -16,7 +16,7 @@ export enum PLAUSIBLE_EVENT_CONTEXTS {
 }
 
 export enum PLAUSIBLE_EVENT_SUBCONTEXT_NFT {
-	ERC721 = 'erc721'
+	ERC721 = 'erc721',
 	ERC1155 = 'erc1155'
 }
 


src/frontend/src/eth/services/erc721.services.ts
```diff
@@ -61,6 +73,36 @@ const loadErc721CustomTokens = async (params: LoadCustomTokenParams): Promise<Cu
...
+			name: PLAUSIBLE_EVENTS.LOAD_CUSTOM_TOKENS,
+			metadata: {
+				event_context: PLAUSIBLE_EVENT_CONTEXTS.NFT,
+				event_subcontext: PLAUSIBLE_EVENT_SUBCONTEXT_NFT.ERC721,
+				token_address: address,
+				token_network: `${networkId.description}`
+			}
+		});
+
+		console.error(
```
roman-nazaruk: we do not need it, you can just add new params to trackEvent: warning

AntonioVentilii: Nice, good one! Done!

src/frontend/src/lib/enums/plausible.ts
```diff
@@ -16,6 +16,7 @@ export enum PLAUSIBLE_EVENT_CONTEXTS {
 }
 
 export enum PLAUSIBLE_EVENT_SUBCONTEXT_NFT {
+	ERC721 = 'erc721'
```
AntonioVentilii:
```suggestion
	ERC721 = 'erc721',
```

PR #9991, opened 2025-10-27
feat(frontend): Store active assets tab

daviddecentage:
# Motivation

We want to store what assets tab the user is on so we can later always keep that last tab open when he returns to that page.

# Changes

- Added store to settings
- Set current tab on assets page


src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -177,6 +178,8 @@ export const switchNetwork = async (networkId: Option<NetworkId>) => {
 		url.searchParams.set(NETWORK_PARAM, networkId.description);
 	}
 
+	userSelectedNetworkStore.set({ key: 'user-selected-network', value: networkId?.description });
+
```
AntonioVentilii:
```suggestion
	if (isNullish(networkId) || isNullish(networkId.description)) {
		url.searchParams.delete(NETWORK_PARAM);
		userSelectedNetworkStore.reset({ key: 'user-selected-network' });
	} else {
		url.searchParams.set(NETWORK_PARAM, networkId.description);
		userSelectedNetworkStore.set({ key: 'user-selected-network', value: networkId.description });
	}
	

```

src/frontend/src/lib/utils/nav.utils.ts
```diff
@@ -168,13 +169,21 @@ export const resetRouteParams = (): RouteParams => ({
 	[URI_PARAM]: null
 });
 
-export const switchNetwork = async (networkId: Option<NetworkId>) => {
+export const switchNetwork = async ({
+	networkId,
+	userSelectedNetworkStore
+}: {
+	networkId: Option<NetworkId>;
+	userSelectedNetworkStore: StorageStore<string | undefined>;
```
AntonioVentilii, line 177: why are you passing it as param? Will we have different stores?

daviddecentage: No but theres an module import issue if I try to import the store directly in the utils. Probably some naming clash because both methods are called set.

PR #10002, opened 2025-10-27
feat(frontend): show only evm networks for import nft collection modal select

roman-nazaruk:
# Motivation

Currently, we only support NFTs on EVM networks. Therefore, in the “Import NFT Collection” modal, the network selector should display only EVM networks.

# Changes

Added a new derived value to filter and show only EVM networks for nft


src/frontend/src/lib/schema/network.schema.ts
```diff
@@ -42,5 +42,6 @@ export const NetworkSchema = z.object({
 	name: z.string(),
 	icon: IconSchema.optional(),
 	exchange: NetworkExchangeSchema.optional(),
-	buy: NetworkBuySchema.optional()
+	buy: NetworkBuySchema.optional(),
+	nftEnabled: z.boolean().optional()
```
AntonioVentilii, line 46: may you please split this part in another PR?

PR #10011, opened 2025-10-28
refactor(frontend): Reorder util `filterTokens`

AntonioVentilii:
# Motivation

For readability and to expand it, we refactor util `filterTokens`.

### DIFF
--- a/src/frontend/src/lib/utils/tokens.utils.ts
+++ b/src/frontend/src/lib/utils/tokens.utils.ts
@@ -232,8 +232,12 @@ export const filterTokens = <T extends Token>({
 			return true;
 		}
 
-		if (icTokenIcrcCustomToken(token)) {
-			return (token.alternativeName ?? '').toLowerCase().includes(filter.toLowerCase());
+		if (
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -222,11 +222,22 @@ export const filterTokens = <T extends Token>({
...
+
+		if (
+			name.toLowerCase().includes(filter.toLowerCase()) ||
+			symbol.toLowerCase().includes(filter.toLowerCase())
+		) {
+			return true;
+		}
+
+		if (icTokenIcrcCustomToken(token)) {
+			return (token.alternativeName ?? '').toLowerCase().includes(filter.toLowerCase());
```
DenysKarmazynDFINITY: To be consistent with the first if statement, how about we move this condition to “if (…)” and just leave “return true”)?

AntonioVentilii: done!

PR #10013, opened 2025-10-28
feat(frontend): added new param to evm networks to filter supportsNft

roman-nazaruk:
# Motivation

We need the ability to filter networks based on whether they support NFTs, and to have a single configuration point for managing this property easily.

# Changes

Added a new property supportsNft to the EVM network configurations.

### DIFF
--- a/src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
+++ b/src/frontend/src/lib/components/networks/NetworksSwitcher.svelte
@@ -13,6 +13,7 @@
 	import { SettingsModalType } from '$lib/enums/settings-modal-types';
 	import { i18n } from '$lib/stores/i18n.store';
 	import { modalStore } from '$lib/stores/modal.store';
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/schema/network.schema.ts
```diff
@@ -42,5 +42,6 @@ export const NetworkSchema = z.object({
 	name: z.string(),
 	icon: IconSchema.optional(),
 	exchange: NetworkExchangeSchema.optional(),
-	buy: NetworkBuySchema.optional()
+	buy: NetworkBuySchema.optional(),
+	supportsNft: z.boolean().optional()
```
AntonioVentilii, line 46: I see that there are some tests for this schema: may you please add this case?

roman-nazaruk: Added

AntonioVentilii (approved):

PR #10022, opened 2025-10-28
fix(frontend): disabled import tokens button for non supported nft chains

roman-nazaruk:
# Motivation

We aim to disable import nft collection button if selected network doesn't support nfts in config

### DIFF
--- a/src/frontend/src/btc/services/worker.btc-wallet.services.ts
+++ b/src/frontend/src/btc/services/worker.btc-wallet.services.ts
@@ -5,7 +5,7 @@ import {
 } from '$btc/services/btc-listener.services';
 import type { BtcPostMessageDataResponseWallet } from '$btc/types/btc-post-message';
 import { STAGING } from '$lib/constants/app.constants';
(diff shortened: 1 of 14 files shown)


src/frontend/src/lib/components/manage/ManageTokens.svelte
```diff
@@ -149,7 +149,11 @@
 			</LogoButton>
 		{/snippet}
 		{#snippet toolbar()}
-			<Button colorStyle="secondary-light" disabled={$pseudoNetworkICPTestnet} onclick={onAddToken}
+			<Button
+				colorStyle="secondary-light"
+				disabled={$pseudoNetworkICPTestnet ||
+					(isNftsPage && nonNullish($selectedNetwork) && !$selectedNetwork?.supportsNft)}
```
AntonioVentilii:
```suggestion
					(isNftsPage && nonNullish($selectedNetwork) && !$selectedNetwork.supportsNft)}
```

PR #10038, opened 2025-10-29
feat(frontend): updated deltaOrder Approve logic

roman-nazaruk:
# Motivation

We aim to use Permit1 eip2612 for swap transaction. In this PR updated deltaOrder creating with default erc20 approve or with permit1.

# Changes

Updated fetchVeloradeltaOrder approve logic

# Tests

Covered logic with tests.

### DIFF
--- a/src/frontend/src/lib/services/swap.services.ts
+++ b/src/frontend/src/lib/services/swap.services.ts
@@ -907,6 +907,17 @@ export const fetchVeloraDeltaSwap = async ({
 
 	let signableOrderData;
 
+	const deltaOrderBaseParams = {
+		deltaPrice: swapDetails as DeltaPrice,
+		owner: userAddress,
+		srcToken: sourceToken.address,
+		destToken: destinationToken.address,
+		srcAmount: parsedSwapAmount.toString(),
+		destAmount: `${slippageMinimum}`,
+		destChainId: Number(destinationNetwork.chainId),
+		partner: OISY_URL_HOSTNAME
+	};
+
 	if (isGasless) {
 		progress(ProgressStepsSwap.APPROVE);
 
@@ -919,14 +930,7 @@ export const fetchVeloraDeltaSwap = async ({
 		});
 
 		signableOrderData = await sdk.delta.buildDeltaOrder({
-			deltaPrice: swapDetails as DeltaPrice,
-			owner: userAddress,
-			srcToken: sourceToken.address,
-			destToken: destinationToken.address,
-			srcAmount: parsedSwapAmount.toString(),
-			destAmount: `${slippageMinimum}`,
-			destChainId: Number(destinationNetwork.chainId),
-			partner: OISY_URL_HOSTNAME,
+			...deltaOrderBaseParams,
 			deadline,
 			nonce: Number(nonce),
 			permit: encodedPermit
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/services/swap.services.ts
```diff
@@ -903,36 +905,66 @@ export const fetchVeloraDeltaSwap = async ({
...
-		shouldSwapWithApproval: true,
-		maxPriorityFeePerGas,
-		progress,
-		progressSteps: ProgressStepsSwap
-	});
+	let signableOrderData;
 
-	progress(ProgressStepsSwap.SWAP);
+	if (isGasless) {
+		progress(ProgressStepsSwap.APPROVE);
```
DenysKarmazynDFINITY, line 922: shouldn't be approve be inside the "else" block?

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -903,36 +905,66 @@ export const fetchVeloraDeltaSwap = async ({
...
+		const { nonce, deadline, encodedPermit } = await createPermit({
+			token: sourceToken,
+			userAddress,
+			spender: deltaContract,
+			value: parsedSwapAmount.toString(),
+			identity
+		});
+
+		signableOrderData = await sdk.delta.buildDeltaOrder({
+			deltaPrice: swapDetails as DeltaPrice,
```
DenysKarmazynDFINITY: there are a lot of common params between the `sdk.delta.buildDeltaOrder` calls - let's create a const for them

src/frontend/src/lib/services/swap.services.ts
```diff
@@ -903,36 +905,66 @@ export const fetchVeloraDeltaSwap = async ({
...
+			srcToken: sourceToken.address,
+			destToken: destinationToken.address,
+			srcAmount: `${parsedSwapAmount}`,
+			destAmount: `${slippageMinimum}`,
+			destChainId: Number(destinationNetwork.chainId),
+			partner: OISY_URL_HOSTNAME
+		});
+	}
+
+	progress(ProgressStepsSwap.SWAP);
```
DenysKarmazynDFINITY: `ProgressStepsSwap.SWAP` is set 2 times

src/frontend/src/eth/components/swap/SwapEthWizard.svelte
```diff
@@ -203,6 +203,7 @@
 				gas,
 				maxFeePerGas,
 				maxPriorityFeePerGas,
+				isGasless: false,
```
DenysKarmazynDFINITY, line 206: just wondering - do you plan to add `isGasless: true` cases in the next PRs?

roman-nazaruk: Yes, I will add this logic in next PR

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: No, because create Permit is logix that responsible for approving token via Permit1

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: I thought about it, but probably we would need to adjust it

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Fixed. Nice catch

src/frontend/src/lib/services/swap.services.ts
roman-nazaruk: Updated and created baseParams

src/frontend/src/lib/services/swap.services.ts
DenysKarmazynDFINITY: and the actual `approveToken` does not need the approve step?

roman-nazaruk: it is inside of approveToken logic

PR #10039, opened 2025-10-29
feat(frontend): added new isPermitSupported and cache value in swap Context

roman-nazaruk:
# Motivation

We aim to use Permit1 eip2612 for swap transaction. In this Pr we cache does erc20token support permit or not.

# Changes

Updated swap context and cache value for supported permit1

# Tests

Covered logic with tests.

### DIFF
--- a/src/frontend/src/eth/components/swap/SwapEthWizard.svelte
+++ b/src/frontend/src/eth/components/swap/SwapEthWizard.svelte
@@ -203,6 +203,7 @@
 				gas,
 				maxFeePerGas,
 				maxPriorityFeePerGas,
+				isGasless: false,
 				swapDetails: $swapAmountsStore.swaps[0].swapDetails as VeloraSwapDetails
 			};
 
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/stores/swap.store.ts
```diff
@@ -23,8 +24,9 @@ type IsTokensIcrc2Map = Record<string, boolean>;
 
 export const initSwapContext = (swapData: SwapData = {}): SwapContext => {
 	const data = writable<SwapData>(swapData);
-	const isTokensIcrc2 = writable<IsTokensIcrc2Map | undefined>();
 	const { update } = data;
+	const isTokensIcrc2 = writable<IsTokensIcrc2Map | undefined>();
+	const isErc20SupportsPermit = writable<IsTokensIcrc2Map | undefined>();
```
DenysKarmazynDFINITY: The naming is a bit off grammatically: maybe `isErc20PermitSupported`?

DenysKarmazynDFINITY: it also goes to all other similar occurrences in the file

roman-nazaruk: Fixed

PR #10041, opened 2025-10-29
feat(frontend): added setIsTokenPermitSupported in swap Eth Wizard

roman-nazaruk:
# Motivation

We aim to make gaslees swaps with permit1 eip2612.

# Changes

Setting  isSourceTokenPermitSupported value  in Swap Eth Wizard


src/frontend/src/eth/components/swap/SwapEthWizard.svelte
```diff
@@ -122,6 +130,25 @@
 				: undefined;
 	});
 
+	$effect(() => {
+		if (isNullish($sourceToken) || !isTokenErc20($sourceToken) || isNullish($ethAddress)) {
+			return;
+		}
+		if (isNullish($isSourceTokenPermitSupported)) {
```
DenysKarmazynDFINITY: let's move this condition to the above if statement as `... || nonNullish($isSourceTokenPermitSupported)) {`

roman-nazaruk: Done

PR #10042, opened 2025-10-29
fix(frotnend): updated fee value for velora details component

roman-nazaruk:
# Motivation

Updated Fee value in Velora details swap component

image


src/frontend/src/lib/components/swap/SwapDetailsVelora.svelte
```diff
@@ -38,7 +38,7 @@
 		{/snippet}
 
 		{#snippet mainValue()}
-			{$currentCurrencySymbol}0
+			{$currentCurrencySymbol}0.00
```
DenysKarmazynDFINITY: Can you please use `formatCurrency({
				value: 0,
				currency: $currentCurrency,
				exchangeRate: $currencyExchangeStore,
				language: $currentLanguage
			})`, it should give you the wanted result.

roman-nazaruk: Updated

PR #10043, opened 2025-10-29
feat(frontend): adjust StakeForm component

DenysKarmazynDFINITY:
# Motivation

We need to make 2 adjustments to the StakeForm component:
1. `destination` should be optional.
2. Additional check if provided amount is zero.

### DIFF
--- a/src/frontend/src/lib/components/stake/StakeForm.svelte
+++ b/src/frontend/src/lib/components/stake/StakeForm.svelte
@@ -20,7 +20,7 @@
 
 	interface Props {
 		amount: OptionAmount;
-		destination: Address;
+		destination?: Address;
 		totalFee?: bigint;
 		onCustomValidate?: (userAmount: bigint) => TokenActionErrorType;
 		onClose: () => void;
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/stake/StakeForm.svelte
```diff
@@ -48,7 +48,7 @@
 	let exchangeValueUnit = $state<DisplayUnit>('usd');
 	let inputUnit = $derived<DisplayUnit>(exchangeValueUnit === 'token' ? 'usd' : 'token');
 
-	let invalid = $derived(invalidAmount(amount) || nonNullish(errorType));
+	let invalid = $derived(invalidAmount(amount) || Number(amount) === 0 || nonNullish(errorType));
```
AntonioVentilii, line 51: Not for this PR: i am now wondering if this check should go inside invalidAmount util... what is the use-case where we accept a zero-amount in the other flows, for example send and convert?

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/lib/components/stake/StakeForm.svelte
DenysKarmazynDFINITY: yeah, I was a bit surprised it's not a part of the util

PR #10044, opened 2025-10-29
feat(frontend): adjust RadioBox component

DenysKarmazynDFINITY:
# Motivation

We need to update the design of `RadioBox`. The component wasn't used before, therefore we can apply the changes without affecting other places.

Screenshot 2025-10-29 at 12 46 13

### DIFF
--- a/src/frontend/src/tests/lib/components/ui/RadioBox.spec.ts
+++ b/src/frontend/src/tests/lib/components/ui/RadioBox.spec.ts
@@ -1,20 +1,16 @@
 import RadioBox from '$lib/components/ui/RadioBox.svelte';
+import { createMockSnippet } from '$tests/mocks/snippet.mock';
 import { assertNonNullish } from '@dfinity/utils';
 import { fireEvent, render } from '@testing-library/svelte';
-import { createRawSnippet } from 'svelte';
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/tests/lib/components/ui/RadioBox.spec.ts
```diff
@@ -1,13 +1,20 @@
...
 		checked: false,
 		disabled: false,
-		label: 'RadioBox',
-		description: 'This is a radio box',
+		label: createRawSnippet(() => ({
+			render: () => `<span>${labelText}</span>`
+		})),
+		description: createRawSnippet(() => ({
+			render: () => `<span>${descriptionText}</span>`
+		})),
```
AntonioVentilii: Nitpick: we have an util called `createMockSnippet` that accepts a `testId` as input, that maybe is a bit more "secure" to test against

AntonioVentilii (approved):
LGTM tks!

src/frontend/src/tests/lib/components/ui/RadioBox.spec.ts
DenysKarmazynDFINITY: updated

PR #10057, opened 2025-10-29
feat(frontend): added new value if eth trx is gasless

roman-nazaruk:
# Motivation

We need to show Total fee is Gasless $0.00 is trx is gasless

# Changes

Updated EthFeeComponent

# Tests

Added tests

### DIFF
--- a/src/frontend/src/eth/components/fee/EthFeeDisplay.svelte
+++ b/src/frontend/src/eth/components/fee/EthFeeDisplay.svelte
@@ -7,13 +7,17 @@
 	interface Props {
 		label?: Snippet;
 		isApproveNeeded?: boolean;
-		isFeeGasless?: boolean;
 	}
 
-	let { label, isApproveNeeded, isFeeGasless }: Props = $props();
+	let { label, isApproveNeeded }: Props = $props();
 
-	const { maxGasFee, feeSymbolStore, feeDecimalsStore, feeExchangeRateStore }: EthFeeContext =
-		getContext<EthFeeContext>(ETH_FEE_CONTEXT_KEY);
+	const {
+		maxGasFee,
+		feeSymbolStore,
+		feeDecimalsStore,
+		feeExchangeRateStore,
+		isFeeGasless
+	}: EthFeeContext = getContext<EthFeeContext>(ETH_FEE_CONTEXT_KEY);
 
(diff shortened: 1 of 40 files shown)


src/frontend/src/eth/components/fee/EthFeeDisplay.svelte
```diff
@@ -7,9 +7,10 @@
 	interface Props {
 		label?: Snippet;
 		isApproveNeeded?: boolean;
+		isFeeGasless?: boolean;
 	}
 
-	let { label, isApproveNeeded }: Props = $props();
+	let { label, isApproveNeeded, isFeeGasless }: Props = $props();
```
AntonioVentilii: instead of passing it, why not putting it in the fee context?

roman-nazaruk: Because it’s not related to the fee context, it’s more about whether the token supports permit, and if it does, the transaction can be gasless.
Also, I don’t want to add too many fields to the fee context, as it already contains quite a lot of logic.

roman-nazaruk: +its calculated at a higher level, and even if a token supports Permit1, it doesn’t necessarily mean the transaction will be gasless. That’s only one of four cases where the transaction is fully gasless

AntonioVentilii: better to put it directly in the context, instead of passing it down. It gives more control and reduces the wieght of the nodes

roman-nazaruk: Sure

roman-nazaruk: Updated

src/frontend/src/eth/components/swap/SwapEthWizard.svelte
```diff
@@ -117,6 +118,12 @@
 		}
 	});
 
+	$effect(() => {
+		if (nonNullish($isSourceTokenPermitSupported)) {
+			isFeeGasless.set($isSourceTokenPermitSupported);
+		}
```
AntonioVentilii:
```suggestion
			isFeeGasless.set($isSourceTokenPermitSupported ?? false);
```

PR #10058, opened 2025-10-29
fix(frontend): fixed permit1 version value with fallback

roman-nazaruk:
# Motivation

To execute Permit1 correctly, we need to retrieve the token’s version from the contract.
If the contract doesn’t support version retrieval, a fallback value of 1 is used.

# Changes

Fixed token version retrieval logic with a fallback to 1.

# Tests

Updated tests

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapGaslessFee.svelte
+++ b/src/frontend/src/lib/components/swap/SwapGaslessFee.svelte
@@ -1,0 +1,27 @@
+<script lang="ts">
+	import ModalValue from '$lib/components/ui/ModalValue.svelte';
+	import { currentCurrency } from '$lib/derived/currency.derived';
+	import { currentLanguage } from '$lib/derived/i18n.derived';
+	import { currencyExchangeStore } from '$lib/stores/currency-exchange.store';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { formatCurrency } from '$lib/utils/format.utils';
+</script>
+
+<ModalValue>
+	{#snippet label()}
+		{$i18n.fee.text.total_fee}
+	{/snippet}
+
+	{#snippet mainValue()}
+		{$i18n.swap.text.gasless}
+	{/snippet}
+
+	{#snippet secondaryValue()}
+		{formatCurrency({
+			value: 0,
+			currency: $currentCurrency,
+			exchangeRate: $currencyExchangeStore,
+			language: $currentLanguage
+		})}
+	{/snippet}
+</ModalValue>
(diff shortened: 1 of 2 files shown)


src/frontend/src/eth/services/eip2612-permit.services.ts
```diff
@@ -28,10 +28,11 @@ const fetchPermitMetadata = async ({
 	tokenName
 }: FetchPermitMetadataParams): Promise<PermitMetadata> => {
 	const nonce = await tokenContract.nonces(userAddress);
+	const version = await tokenContract.version().catch(() => '1');
```
AntonioVentilii: let's use a `try catch` (but not `let`). You can use `Promise.allSettled`

roman-nazaruk: Added try catch block

src/frontend/src/eth/services/eip2612-permit.services.ts
```diff
@@ -27,14 +27,28 @@ const fetchPermitMetadata = async ({
...
+	} catch {
+		const nonce = await tokenContract.nonces(userAddress);
+
+		return {
+			name: tokenName,
+			version: '1',
+			nonce: nonce.toString(),
+			deadline: customDeadline ?? createDeadline()
+		};
+	}
```
AntonioVentilii: why not using Promise.allSettled and check the results? So you avoid calling `tokenContract.nonces(userAddress)` twice. Or at this point extract `tokenContract.version()` in a separate method where you always return something and do not raise an error, like `safeVersion`, boh! So you avoid looping the results of Promise.allSettled


```ts
const safeVersion = async () =>{
  try {
    return await tokenContract.version()
  } catch (_: unknown) {
    return '1'
  }
}
```

roman-nazaruk: > why not using Promise.allSettled and check the results? So you avoid calling `tokenContract.nonces(userAddress)` twice. Or at this point extract `tokenContract.version()` in a separate method where you always return something and do not raise an error, like `safeVersion`, boh! So you avoid looping the results of Promise.allSettled
> 
> ```ts
> const safeVersion = async () =>{
>   try {
>     return await tokenContract.version()
>   } catch (_: unknown) {
>     return '1'
>   }
> }
> ```

Updated

src/frontend/src/eth/services/eip2612-permit.services.ts
```diff
@@ -27,11 +27,19 @@ const fetchPermitMetadata = async ({
 	customDeadline,
 	tokenName
 }: FetchPermitMetadataParams): Promise<PermitMetadata> => {
-	const nonce = await tokenContract.nonces(userAddress);
+	const safeVersion = async (): Promise<string> => {
```
AntonioVentilii: some comment here on why we do this would be helpful for posterity

AntonioVentilii (approved):
LGTM tks!

PR #10062, opened 2025-10-29
fix(frontend): Disable buy button in hero if on Nft page

daviddecentage:
# Motivation

Showing the buy button in the hero doesnt make sense if the user is on the Nft page, so we disable them.

# Changes

Added inflowDisabled from hero context to disabled condition of the button

# Tests

Any other page:
image

Nfts page:
image


AntonioVentilii (commented):
There is actually an Hero Context that can be used in this case.

The store `inflowActionsDisabled` can be used in these cases since Buy is an inflow, and Swap is both inflow and outflow

P.S. may you please split it per-button? Better track it separately. I even think that the swap button should have both `inflowActionsDisabled` and `outflowActionsDisabled` in its logic.

AntonioVentilii (approved):

PR #10064, opened 2025-10-29
fix(frontend): Disable nfts tab if unsupported network

daviddecentage:
# Motivation

We want to disable the Nfts pages tab if the selected network is unsupported.

# Changes

- Added disabled prop to tabs
- Added disabled to nfts tab if selected network is nfts unsupported

# Tests

Added tests

https://github.com/user-attachments/assets/ea28f1fe-a9b9-4f0d-9a8f-6f7bca81a18d

### DIFF
--- a/src/frontend/src/icp/services/gldt-stake.services.ts
+++ b/src/frontend/src/icp/services/gldt-stake.services.ts
@@ -4,7 +4,7 @@ import { approve } from '$icp/api/icrc-ledger.api';
 import type { IcToken } from '$icp/types/ic-token';
 import { nowInBigIntNanoSeconds } from '$icp/utils/date.utils';
 import { GLDT_STAKE_CANISTER_ID, NANO_SECONDS_IN_MINUTE } from '$lib/constants/app.constants';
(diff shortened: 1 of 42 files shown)


AntonioVentilii (commented):
I have two doubts now:

- what happens if you are already in the NFT tab with "All networks", and then you select a NTF-unsupported network? May you please put a video of this?

- if we do this, then we should revert the PR https://github.com/dfinity/oisy-wallet/pull/10040? It seems pointless at this point

AntonioVentilii:
I checked the video, it still seems a bit strange... should we maybe automatically redirect the user directly to the Assets page, out of NFT, when he selects a network that does not support NFT?

I mean: either we have a placeholder, or we disable the tab. Having both it is redundant

daviddecentage:
> I checked the video, it still seems a bit strange... should we maybe automatically redirect the user directly to the Assets page, out of NFT, when he selects a network that does not support NFT?
> 
> I mean: either we have a placeholder, or we disable the tab. Having both it is redundant

Stefan specifically said he doesnt want the redirect to tokens page when were already on the Nfts tab, so no not redundant.

PR #10068, opened 2025-10-29
fix(frontend): Hide mobile nav on bottom sheet visible

daviddecentage:
# Motivation

The mobile navigation overlaps bottom sheets in many cases. Since the bottom sheets can appear on any z-index on the page, we cannot solve this by CSS only, so we completely hide the mobile nav as soon as a bottom sheet is visible.

# Changes

- Hide mobile nav if a bottom sheet is opened

# Tests

Adjusted bottom sheet tests


src/frontend/src/lib/components/navigation/MobileNavigationMenu.svelte
```diff
@@ -13,6 +14,7 @@
 <Responsive down="md">
 	<div
 		class="mobile-nav visible fixed right-0 bottom-0 left-0 z-3 flex flex-row border-t-1 border-tertiary bg-primary-inverted-alt md:hidden"
+		class:hidden={$bottomSheetOpenStore}
```
AntonioVentilii, line 17: why not an IF condition?

daviddecentage: I think toggling a class will be smoother UX instead of mounting/unmounting everytime the bottomsheet opens/closes, but up to you.

AntonioVentilii (approved):

PR #10073, opened 2025-10-30
feat(frontend): added correct fee calc for swap gasless

roman-nazaruk:
# Motivation

We need to add new calculation of fee if swap is galsess

# Changes

Adjusted fee calculation

### DIFF
--- a/src/frontend/src/eth/constants/erc20.constants.ts
+++ b/src/frontend/src/eth/constants/erc20.constants.ts
@@ -18,6 +18,7 @@ export const ERC20_ABI = [
 // Spec: https://eips.ethereum.org/EIPS/eip-2612
 export const ERC20_PERMIT_ABI = [
 	'function nonces(address owner) view returns (uint256)',
+	'function version() view returns (string)',
 	'function DOMAIN_SEPARATOR() view returns (bytes32)'
 ];
 
--- a/src/frontend/src/eth/providers/infura-erc20.providers.ts
+++ b/src/frontend/src/eth/providers/infura-erc20.providers.ts
@@ -135,14 +135,14 @@ export class InfuraErc20Provider implements Erc20Provider {
 		contractAddress: string;
 		userAddress: EthAddress;
 	}): Promise<boolean> => {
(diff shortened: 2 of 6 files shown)


src/frontend/src/eth/components/swap/SwapEthWizard.svelte
```diff
@@ -170,7 +170,13 @@
...
 	const isApproveNeeded = $derived<boolean>(
 		$swapAmountsStore?.swaps[0]?.type === VeloraSwapTypes.MARKET &&
-			isNotDefaultEthereumToken($sourceToken)
+			isNotDefaultEthereumToken($sourceToken) &&
+			!$isSourceTokenPermitSupported
+	);
+
+	const isGasless = $derived<boolean>(
+		$swapAmountsStore?.swaps[0]?.type === VeloraSwapTypes.DELTA &&
+			$isSourceTokenPermitSupported === true
```
DenysKarmazynDFINITY: This flag is boolean, right? then let's remove "=== true"

roman-nazaruk: no it can be undefined

roman-nazaruk: Fixed

src/frontend/src/eth/components/swap/SwapEthForm.svelte
```diff
@@ -83,6 +85,14 @@
 					})
 				: ZERO;
 
+		if (isGasless) {
```
DenysKarmazynDFINITY: This change now means that only for `isGasless` it will be checking the `userAmount > parsedSendBalance` condition. I guess what you want is:
```
if (userAmount > parsedSendBalance) {
  return 'insufficient-funds';
}

if (isGasless) {
  return;
}
```

roman-nazaruk: Good suggestion

roman-nazaruk: Updated

PR #10075, opened 2025-10-30
feat(frontend): UI store

daviddecentage:
# Motivation

For an upcoming fix that is purely UI related, we add a store for UI specific states.

# Changes

Added bottomSheetOpenStore

# Tests

Added test for the store

### DIFF
--- a/src/frontend/src/tests/lib/stores/ui.store.spec.ts
+++ b/src/frontend/src/tests/lib/stores/ui.store.spec.ts
@@ -1,0 +1,30 @@
+import { bottomSheetOpenStore } from '$lib/stores/ui.store';
+import { get } from 'svelte/store';
+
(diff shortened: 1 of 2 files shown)


AntonioVentilii (approved):

src/frontend/src/tests/lib/stores/ui.store.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { bottomSheetOpenStore } from '$lib/stores/ui.store';
+import { get } from 'svelte/store';
+
+describe('ui.store', () => {
```
AntonioVentilii, line 4: P.S. i think the tests are not required in this specific case, since there is no additional logic to `writable`, but nice to have!

PR #10094, opened 2025-10-30
fix(frontend): Make the mapped attributes of NFTs more solid

AntonioVentilii:
# Motivation

Some of the attributes may come in different shapes (for example as record, or partial record). So, we need to create an NFT attribute mapper that is a little bit more solid.

# Changes

- Extract the NFT attribute mapper in `AlchemyProvider` class.
- Expand the mapper to consider nullish values AND records.

# Tests

Some NFTs were not loading before:

Screenshot 2025-10-30 at 16 30 41

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -68,7 +68,7 @@
 				"pem-file": "^1.0.1",
 				"postcss": "^8.5.6",
 				"prettier": "^3.5.3",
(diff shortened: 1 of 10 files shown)


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #10108, opened 2025-10-31
refactor(frontend): Remove some loops from `LoaderCollections`

AntonioVentilii:
# Motivation

In component `LoaderCollections` we are doing a few loops that could be avoided, since we can handle the all the ERC tokens in a single loop.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #10127, opened 2025-11-01
feat(frontend): display gldt_stake canister error messages

DenysKarmazynDFINITY:
# Motivation

The gldt_stake canister returns meaningful messages when staking/unstaking action fails on their end. Therefore, it makes sense to display them instead of default strings if we can parse the values.

### DIFF
--- a/src/frontend/src/icp/canisters/gldt_stake.errors.ts
+++ b/src/frontend/src/icp/canisters/gldt_stake.errors.ts
@@ -3,27 +3,35 @@ import { CanisterInternalError } from '$lib/canisters/errors';
 
 export const mapGldtStakeCanisterError = (err: ManageStakePositionError): CanisterInternalError => {
 	if ('StartDissolvingError' in err) {
-		return new CanisterInternalError('Failed to start dissolving');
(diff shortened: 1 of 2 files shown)


src/frontend/src/icp/canisters/gldt_stake.errors.ts
```diff
@@ -3,27 +3,35 @@ import { CanisterInternalError } from '$lib/canisters/errors';
 
 export const mapGldtStakeCanisterError = (err: ManageStakePositionError): CanisterInternalError => {
 	if ('StartDissolvingError' in err) {
-		return new CanisterInternalError('Failed to start dissolving');
+		return new CanisterInternalError(
+			Object.values(err.StartDissolvingError)[0] ?? 'Failed to start dissolving'
```
AntonioVentilii, line 7: `fromNullable`?

DenysKarmazynDFINITY: the value is not really a nullable, but rather a nested object - e.g. `response.StartDissolvingError.DissolvementsLimitReached`

PR #10153, opened 2025-11-03
feat(frontend): scope auth client provider with a class

peterpeterparker:
# Motivation

In an upcoming PR, we would like to transform the custom storage from a variable declared at the root of a module to an instance pattern. That's why, in a first step, we refactor the various auth-client related functions to include them in a class pattern.

In this PR, the various functions get reexported by the module, therefore it should not introduce any other changes that those scoped to the module.


AntonioVentilii (approved):
Nice, tks!!!

PR #10156, opened 2025-11-03
feat(frontend): auth client provider singleton

peterpeterparker:
# Motivation

We are not sure but, we suspect the random IndexedDB issue that happens at boot time (IDB not open or something) might be related to the fact that we started using a custom storage variable for providing and interacting with the `AuthClient`. That is way, instead, we transform the pattern into a singleton that will be initialized once when the `authStore` is synchronized.

# Changes

- Transform `AuthClientProvider` into an old'school Java like singleton.
- Use singleton instead of root module declared functions.


AntonioVentilii (approved):
THank you!!

AntonioVentilii (approved):

PR #10158, opened 2025-11-04
feat(frontend): Send user snapshots for NFT

AntonioVentilii:
# Motivation

We want to start sending snapshots of the NFTs too. However, we still need to adapt a bit the Reward Canister (RC).

For example:

- `Standard`: we cannot tell the RC the difference between NFT and Fungible Token currently. So, for now we create a custom string for the NFT.
- `Balance`: we should be able to send separate NFT for each collection with their specific balance, but for now we just send the number of different NFT per collection as balance.

# Changes

- Map the token symbol of NFT as a string with separators and necessary data.
- Map the token balance of NFT with the number of difference NFT ID (different NFTs).

# Tests

Added tests


src/frontend/src/lib/services/user-snapshot.services.ts
```diff
@@ -155,6 +160,22 @@ const getLastTransactionsByToken = ({
 				: [];
 };
 
+// Currently we have not way of differentiate more details for the token (for example the standard or the ID for NFTs).
+// So, for now, we simply create a custom string for the NFTs, and use the token ID description for the fungible tokens.
+// TODO: Improve the data passing when the Rewards Canister expand the interface for the type `AnyToken`.
+const toTokenSymbol = (token: Token): string => {
+	if (isTokenErc721(token) || isTokenErc1155(token)) {
+		return `nft#${token.name}#${token.address}#${token.section ?? ''}`;
```
DenysKarmazynDFINITY, line 168: shall we check if token.section is defined earlier to prevent having "#" at the end of the string if it's undefined?
`return `nft#${token.name}#${token.address}${nonNullish(token.section) ? `#${token.section}` : ''}`;`

AntonioVentilii: So, I asked how we wanted to receive in the RC, and it is preferable to have the separator `#` even if with empty string after.

DenysKarmazynDFINITY: ok, got it

PR #10167, opened 2025-11-04
feat(frontend): Remove unnecessary NFT store update loops

AntonioVentilii:
# Motivation

I had a deeper look at the `LoaderNft` component and it really seems we are doing a bit of unnecessary loops for the NFT store.

We are already setting the NFT by network, removing all the existing one and overriding them. So, it seems unnecessary to remove or update anything after in the same store.

# Changes

- Remove `handleRemovedNfts` and `handleUpdatedNfts` from `LoaderNft`.
- Remove all utils and methods related to them.

# Test

I tested manually in a test canister, sending and receiving ERC721 and ERC1155 and all worked as expected.


DenysKarmazynDFINITY (approved):
Nice, so much code removed 🙂  let's just keep an eye on it for a while

PR #10173, opened 2025-11-04
fix(frontend): Avoid token contract as symbol for NFTs

AntonioVentilii:
# Motivation

It does not really make sense to fallback to token contracts for each NFT contract. We can have an empty symbol for now since it is very rare and since it is only associated with the amount.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #10200, opened 2025-11-05
feat(frontend): added new isAllowanceSufficient for icrc2 tokens

roman-nazaruk:
# Motivation

We need to check for existing allowance in the swap flow to avoid unnecessary double approvals when slippage is exceeded or when a transaction fails.

# Changes

- Added new method hasSufficientIcrcAllowance that checks if the user has sufficient ICRC-2 allowance before initiating an approve transaction

# Tests

Covered new method with the tests

### DIFF
--- a/src/frontend/src/icp/services/icrc.services.ts
+++ b/src/frontend/src/icp/services/icrc.services.ts
@@ -350,6 +350,10 @@ export const loadDisabledIcrcTokensExchanges = async ({
 /**
  * Checks if the owner has sufficient allowance for the spender to execute a transaction.
  *
+ * The allowanceBuffer ensures the allowance won't expire while the transaction is being processed.
+ * For example, if the allowance expires in 10 seconds but the swap takes 20 seconds to complete,
+ * we should request a new approval instead of starting a transaction that will fail.
+ *
  * @returns `true` if allowance is sufficient and won't expire within the buffer period, `false` otherwise
  *
  */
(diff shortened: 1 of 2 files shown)


src/frontend/src/icp/services/icrc.services.ts
```diff
@@ -343,3 +346,42 @@ export const loadDisabledIcrcTokensExchanges = async ({
...
+		const { allowance, expires_at } = await icrcAllowance({
+			identity,
+			ledgerCanisterId,
+			owner: { owner },
+			spender: { owner: spender }
+		});
+
+		const expiredBuffer = nowInBigIntNanoSeconds() + allowanceBuffer;
+		const hasSufficientAllowance = allowance >= amount;
+		const isNotExpired = nonNullish(expires_at[0]) && expires_at[0] > expiredBuffer;
```
AntonioVentilii: fromNullable

src/frontend/src/icp/services/icrc.services.ts
```diff
@@ -343,3 +346,42 @@ export const loadDisabledIcrcTokensExchanges = async ({
...
+
+		return hasSufficientAllowance && isNotExpired;
+	} catch (_: unknown) {
+		return false;
```
AntonioVentilii: why would we ignore the errors? we should raise them, you never know why the error was created, right?

src/frontend/src/icp/services/icrc.services.ts
```diff
@@ -343,3 +346,42 @@ export const loadDisabledIcrcTokensExchanges = async ({
...
+			ledgerCanisterId,
+			owner: { owner },
+			spender: { owner: spender }
+		});
+
+		const expiredBuffer = nowInBigIntNanoSeconds() + allowanceBuffer;
+		const hasSufficientAllowance = allowance >= amount;
+		const isNotExpired = nonNullish(expires_at[0]) && expires_at[0] > expiredBuffer;
+
+		return hasSufficientAllowance && isNotExpired;
```
AntonioVentilii: i would put the logic of isExpired in another service, and then combine. Furthermore, are you sure that if it is expired, we still receive a non null allowance?

PR #10224, opened 2025-11-06
chore(deps): Update package-lock

AntonioVentilii:
# Motivation

Run `npm i` to keep package-lock updated: we switched to Node 24 and the package-lock needs therefore to be regenerated.


peterpeterparker (approved):
LGTM, thx

PR #10235, opened 2025-11-07
feat(frontend): added new init of plausible events

roman-nazaruk:
# Motivation

We are integrating the new Plausible Analytics events package.

Note:
This library only works in browser environments. When using server-side rendering (SSR), the init and track functions will not work as they rely on browser APIs. Make sure to only initialize and track events on the client side.

# Changes

Added initialization and event tracking using the new Plausible Analytics package.

### DIFF
--- a/src/frontend/src/icp/components/stake/gldt/GldtStakeRewards.svelte
+++ b/src/frontend/src/icp/components/stake/gldt/GldtStakeRewards.svelte
@@ -56,7 +56,7 @@
 {#if nonNullish($gldtStakeStore?.position)}
 	<StakeContentSection>
 		{#snippet title()}
-			{$i18n.stake.text.unclaimed_rewards}
+			<h4>{$i18n.stake.text.unclaimed_rewards}</h4>
 		{/snippet}
 
 		{#snippet content()}
--- a/src/frontend/src/lib/components/stake/StakeContentSection.svelte
+++ b/src/frontend/src/lib/components/stake/StakeContentSection.svelte
@@ -5,18 +5,14 @@
 	interface Props {
 		title: Snippet;
 		content: Snippet;
-		action?: Snippet;
 	}
 
-	let { content, action, title }: Props = $props();
+	let { content, title }: Props = $props();
 </script>
 
-<!-- TODO: add styling according to the designs -->
-<div class="rounded-xl bg-secondary p-4" in:fade>
-	<div class="flex w-full justify-between">
-		<h4>{@render title()}</h4>
-
-		{@render action?.()}
+<div class="rounded-xl bg-surface p-4" in:fade>
+	<div class="relative flex w-full items-start justify-between">
+		{@render title()}
 	</div>
 
 	<div class="mt-2">


src/frontend/src/lib/services/analytics.services.ts
```diff
@@ -1,27 +1,35 @@
+import { browser } from '$app/environment';
 import { PLAUSIBLE_DOMAIN, PLAUSIBLE_ENABLED } from '$env/plausible.env';
 import type { TrackEventParams } from '$lib/types/analytics';
 import { isNullish, nonNullish } from '@dfinity/utils';
-import Plausible from 'plausible-tracker';
 
-let plausibleTracker: ReturnType<typeof Plausible> | null = null;
+// eslint-disable-next-line @typescript-eslint/consistent-type-imports
```
AntonioVentilii: let's try to fit the correct type

```suggestion
```

src/frontend/src/lib/services/analytics.services.ts
```diff
@@ -1,27 +1,35 @@
...
+		nonNullish(plausibleTracker)
+	) {
 		return;
 	}
 
+	// Note: This module must be imported in the browser only.
+	// The latest version of @plausible-analytics/tracker does not work in server (SSR) mode.
+	// Reference: https://www.npmjs.com/package/@plausible-analytics/tracker
+	// Important: This library only works in browser environments. The `init` and `track`
+	// functions rely on browser APIs, so they should only be initialized and called on the client side.
```
AntonioVentilii, line 24: so, just to clarify, we are going to track ONLY browser?

furthermore, may you please check the impact on performance, both on desktop and on mobile? Verify if it uses more or less memory consumption at loading

src/frontend/src/lib/services/analytics.services.ts
```diff
@@ -1,27 +1,35 @@
+import { browser } from '$app/environment';
 import { PLAUSIBLE_DOMAIN, PLAUSIBLE_ENABLED } from '$env/plausible.env';
 import type { TrackEventParams } from '$lib/types/analytics';
 import { isNullish, nonNullish } from '@dfinity/utils';
-import Plausible from 'plausible-tracker';
 
-let plausibleTracker: ReturnType<typeof Plausible> | null = null;
+// eslint-disable-next-line @typescript-eslint/consistent-type-imports
+let plausibleTracker: typeof import('@plausible-analytics/tracker') | null = null;
```
peterpeterparker: For consistency with the pattern of the stores, maybe `undefined` by default? Then either null or tyepof... if loaded or not. WDYT?

src/frontend/src/lib/services/analytics.services.ts
```diff
@@ -1,27 +1,35 @@
...
 
-// eslint-disable-next-line require-await -- We use this service during initialisation among other async functions, so for consistency we force the asynchronicity here.
 export const initPlausibleAnalytics = async () => {
-	if (!PLAUSIBLE_ENABLED || isNullish(PLAUSIBLE_DOMAIN)) {
+	if (
+		!PLAUSIBLE_ENABLED ||
+		isNullish(PLAUSIBLE_DOMAIN) ||
+		!browser ||
+		nonNullish(plausibleTracker)
```
peterpeterparker, line 15: Picky question: is non nullish really equals to being initialized? Those are two differents steps as far as I can tell from the code. Though it probably works in practice.

src/frontend/src/lib/services/analytics.services.ts
```diff
@@ -1,27 +1,35 @@
...
+	// The latest version of @plausible-analytics/tracker does not work in server (SSR) mode.
+	// Reference: https://www.npmjs.com/package/@plausible-analytics/tracker
+	// Important: This library only works in browser environments. The `init` and `track`
+	// functions rely on browser APIs, so they should only be initialized and called on the client side.
 	try {
-		if (isNullish(plausibleTracker)) {
-			plausibleTracker = Plausible({
-				domain: PLAUSIBLE_DOMAIN,
-				hashMode: false,
-				trackLocalhost: false
```
peterpeterparker, line 19: This - `trackLocalhost` - is not required anymore?

src/frontend/src/lib/services/analytics.services.ts
```diff
@@ -1,27 +1,35 @@
...
-			});
-			plausibleTracker.enableAutoPageviews();
```
peterpeterparker, line 21: likewise, enable by default now?

src/frontend/src/lib/services/analytics.services.ts
```diff
@@ -1,27 +1,35 @@
...
+// eslint-disable-next-line @typescript-eslint/consistent-type-imports
+let plausibleTracker: typeof import('@plausible-analytics/tracker') | null = null;
 
-// eslint-disable-next-line require-await -- We use this service during initialisation among other async functions, so for consistency we force the asynchronicity here.
 export const initPlausibleAnalytics = async () => {
-	if (!PLAUSIBLE_ENABLED || isNullish(PLAUSIBLE_DOMAIN)) {
+	if (
+		!PLAUSIBLE_ENABLED ||
+		isNullish(PLAUSIBLE_DOMAIN) ||
+		!browser ||
```
peterpeterparker, line 14: unrelated: there ware now various similar references - `!browser` - in the code base. Maybe it could be handy to introduce a constant like `NOT_BROWSER` or `SSR_BUILD` or something

src/frontend/src/lib/services/analytics.services.ts
roman-nazaruk: Fixed

src/frontend/src/lib/services/analytics.services.ts
roman-nazaruk: Yeah, I need to check it

src/frontend/src/lib/services/analytics.services.ts
roman-nazaruk: Done. Great suggestion

src/frontend/src/lib/services/analytics.services.ts
roman-nazaruk: this property is not supported anymore in new version of library

src/frontend/src/lib/services/analytics.services.ts
roman-nazaruk: Same here

src/frontend/src/lib/services/analytics.services.ts
roman-nazaruk: Good catch, added new variable to check is it really initialized

src/frontend/src/lib/services/analytics.services.ts
roman-nazaruk: Done.

src/frontend/src/lib/constants/app.constants.ts
```diff
@@ -219,3 +220,5 @@ export const NETWORK_BONUS_MULTIPLIER_DEFAULT = 1;
 
 // NFT max filesize limit (10MB)
 export const NFT_MAX_FILESIZE_LIMIT = 1024 * 1024 * 10;
+
+export const NOT_BROWSER = !browser;
```
AntonioVentilii: will you update the rest of the code in a separate PR afterwards?

src/frontend/src/lib/services/analytics.services.ts
```diff
@@ -1,27 +1,39 @@
...
 import { PLAUSIBLE_DOMAIN, PLAUSIBLE_ENABLED } from '$env/plausible.env';
+import { loadPlausibleTracker } from '$lib/services/analytics-wrapper';
 import type { TrackEventParams } from '$lib/types/analytics';
 import { isNullish, nonNullish } from '@dfinity/utils';
-import Plausible from 'plausible-tracker';
+import type { init, track } from '@plausible-analytics/tracker';
 
-let plausibleTracker: ReturnType<typeof Plausible> | null = null;
+let plausibleTracker: { init: typeof init; track: typeof track } | undefined = undefined;
+let plausibleTrackerIsInitialized = false;
```
peterpeterparker: Again a PICKY feedback, so feel free to ignore, but, I'm not a big fan of having a flag and a variable - having two sources of truth for a single state.

I would maybe rather assign the global variable once things are initialized or create a singleton class to encapsulate the logic.

```
const tmp = await loadPlausibleTracker();

		plausibleTracker.init({
			domain: PLAUSIBLE_DOMAIN
		});
		
// Now it's initialized, so we initialize the global varialbe this way
// we now everyhting is done
plausibleTracker = tmp;
```

or go with something similar as we did this week in the `providers` package. Maybe this can even be a provider.

```
class PlausibleTracker {
  static #instance: PlausibleTracker;
  #tracker: { init: typeof init; track: typeof track } | undefined;
  
  private constructor()
  
  static getInstance(): Promise<PlausibleTracker> {
      if (isNullish(this.#instance) {
          this.#instance = new PlausibleTracke();
          await this.#instance.init();
       }
     
        return  this.#instance;
   }
   
   protected async init() {
         this.#tracker = await loadPlausibleTracker();
         
         this.#tracker.init({
			domain: PLAUSIBLE_DOMAIN
		});

// etc.
``` 

ANYWAY it practically work so probably more a picky feedback than anything else.

src/frontend/src/lib/services/analytics.services.ts
roman-nazaruk: TBH, I’ll create a new NOT_BROWSER constant in a separate PR and remove all !browser checks across the project. I think it represents a different context, so it makes sense to separate them.

src/frontend/src/lib/constants/app.constants.ts
roman-nazaruk: TBH, I’ll create a new NOT_BROWSER constant in a separate PR and remove all !browser checks across the project. I think it represents a different context, so it makes sense to separate them.

src/frontend/src/lib/services/analytics.services.ts
roman-nazaruk: Great suggestion! I chose the first option since it doesn’t require rewriting the whole logic, and updating the tests for the other one would be a bit more complicated.

peterpeterparker (approved):
LGTM, thx

PR #10237, opened 2025-11-07
fix(frontend): Optimize spinner

daviddecentage:
# Motivation

After some analysis it turns out the initial spinner is one of the reasons IOS browsers crash frequently. We optimize it to get rid of several issues.

The main problem seems to be that we currently just remove the spinner via direct DOM manipulation without halting the CSS animations, that use 3D transform. When the DOM element is removed the CSS animations can still linger in the GPU being calculated for an inexisting element.

To address all these issues, several adjustments had to be made:

# Changes

- Simple div instead of SVG
- No transforms except rotate for the animation, especially no 3D transforms
- Cleanly remove the animations and after request a seperate animation frame to remove the DOM element
- Limit animation to 40 iterations (30 seconds) to have a fallback when the DOM element isnt removed as safely as expected

# Tests

https://github.com/user-attachments/assets/3e2787e3-0239-4297-ab54-68d361168ae7

### DIFF
--- a/src/frontend/src/app.html
+++ b/src/frontend/src/app.html
@@ -110,26 +110,28 @@
 
 		<style>
 			#app-spinner {
-					--spinner-size: 30px;
+				--spinner-size: 30px;
 
-          width: 30px;
-          height: 30px;
-          border: 3px solid rgb(0, 102, 255);
-          border-top-color: transparent;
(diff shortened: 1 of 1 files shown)


AntonioVentilii (approved):
Super!

PR #10241, opened 2025-11-07
feat(frontend): Load ERC tokens filtered by active network

AntonioVentilii:
# Motivation

When we load ERC tokens from the backend, we indiscriminately load all of them, even if some EVMs networks are disabled.

So, we want a reactive way to load them according to the enabled networks.

# Changes

- Pass down the enabled networks' chain IDs to all the ERC services that loads data from the backend.

# Tests

Adapted tests.


src/frontend/src/eth/services/erc1155.services.ts
```diff
@@ -6,6 +6,8 @@ import { infuraErc1155Providers } from '$eth/providers/infura-erc1155.providers'
 import { erc1155CustomTokensStore } from '$eth/stores/erc1155-custom-tokens.store';
 import type { Erc1155ContractAddress } from '$eth/types/erc1155';
 import type { Erc1155CustomToken } from '$eth/types/erc1155-custom-token';
+import type { Erc721ContractAddress } from '$eth/types/erc721';
```
AntonioVentilii:
```suggestion
import type {
 Erc721ContractAddress           
  } from '$eth/types/erc721';
```

PR #10245, opened 2025-11-07
chore(frontend): Remove unused event `nnsInput`

AntonioVentilii:
# Motivation

We have some left-over event `nnsInput` that are not really used.


daviddecentage (approved):
LGTM

PR #10263, opened 2025-11-10
feat(frontend): added new calculate Value difference

roman-nazaruk:
# Motivation

We need to display the value difference percentage between paid and received amounts in the swap component. This calculation needs to be reusable across multiple components without prop drilling.

# Changes

- Created `calculateValueDifference` utility function in `swap.utils.ts` to calculate the percentage difference between paid and received values based on swap amounts and exchange rates

# Tests

Covered with tests.

### DIFF
--- a/src/frontend/src/eth/services/prepare.services.ts
+++ b/src/frontend/src/eth/services/prepare.services.ts
@@ -1,0 +1,43 @@
+import type { EthSignTransactionRequest } from '$declarations/signer/declarations/signer.did';
+import type { NetworkChainId } from '$eth/types/network';
+import { i18n } from '$lib/stores/i18n.store';
+import type { TransferParams } from '$lib/types/send';
+import { isNullish, toNullable } from '@dfinity/utils';
+import { get } from 'svelte/store';
+
+export const prepare = ({
+	maxPriorityFeePerGas: max_priority_fee_per_gas,
+	maxFeePerGas: max_fee_per_gas,
+	nonce,
+	gas,
+	chainId: chain_id,
+	data,
+	to,
+	amount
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/utils/swap.utils.ts
```diff
@@ -239,3 +240,31 @@ export const findSwapProvider = (
...
+	const receivedValue =
+		nonNullish(receiveAmount) && nonNullish(destinationTokenExchangeRate)
+			? receiveAmount * destinationTokenExchangeRate
+			: undefined;
+
+	if (nonNullish(paidValue) && nonNullish(receivedValue) && paidValue !== 0) {
+		return ((receivedValue - paidValue) / paidValue) * 100;
+	}
+
+	return undefined;
```
AntonioVentilii:
```suggestion
```

src/frontend/src/lib/utils/swap.utils.ts
```diff
@@ -239,3 +240,31 @@ export const findSwapProvider = (
...
+	swapAmount: OptionAmount;
+	receiveAmount?: number;
+	sourceTokenExchangeRate?: number;
+	destinationTokenExchangeRate?: number;
+}): number | undefined => {
+	const paidValue =
+		nonNullish(swapAmount) && nonNullish(sourceTokenExchangeRate)
+			? Number(swapAmount) * sourceTokenExchangeRate
+			: undefined;
+
```
AntonioVentilii, line 259: i would already put a nullish check here

```ts
if (isNullish(paidValue) || paidValue === 0) {
  return
}
```

to avoid more calculation

PR #10265, opened 2025-11-10
feat(frontend): show info message if swap Provider kongswap and slippage exceeded

roman-nazaruk:
# Motivation

We need to display an info message when using KongSwap as the swap provider if the actual value difference exceeds the user's slippage tolerance.

image

### DIFF
--- a/src/frontend/src/tests/icp/components/swap/SwapIcpWizard.spec.ts
+++ b/src/frontend/src/tests/icp/components/swap/SwapIcpWizard.spec.ts
@@ -132,4 +132,251 @@ describe('SwapIcpWizard', () => {
 
 		expect(container).toBeInTheDocument();
 	});
+
+	describe('slippage exceeded validation', () => {
+		const createContextWithKongSwap = ({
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/swap/SwapForm.svelte
```diff
@@ -113,7 +117,8 @@
 			isNullish(receiveAmount) ||
 			isNullish(fee) ||
 			swapAmountsLoading ||
-			(nonNullish(slippageValue) && Number(slippageValue) >= SWAP_SLIPPAGE_INVALID_VALUE)
+			(nonNullish(slippageValue) && Number(slippageValue) >= SWAP_SLIPPAGE_INVALID_VALUE) ||
+			(nonNullish(isSlippageExceeded) && isSlippageExceeded)
```
AntonioVentilii:
```suggestion
			( isSlippageExceeded)
```

AntonioVentilii (approved):
LGTM tks! One small obs

PR #10271, opened 2025-11-11
feat(frontend): auth broadcast singleton

peterpeterparker:
# Motivation

We don't face this issue in OISY, but as I was releasing Juno to production yesterday, I encountered a race condition caused by the authentication broadcast feature. The tab where the new user signed in emitted a message, which resulted in the authentication store being updated twice. This ultimately triggered two identical functions, both attempting to create a wallet.  

Technically, this wasn't an issue, but visually it caused errors to appear for the new user because one call failed while the other succeeded.  

To prevent a similar issue from potentially occurring in OISY, we discussed implementing the same guard - i.e. making the broadcast provider ignore self-messages by using IDs to identify emitters.

# Changes

- Transform `AuthBroadcastChannel` into a singleton
- Generate an ID in the constructor
- Emit message and the ID in the `postMessage`
- On message, discard message received from self
- Rename `close` to `destroy` to reset the singleton

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,6 @@
 				"@solana-program/token-2022": "^0.6.0",
 				"@solana/kit": "^4.0.0",
 				"@velora-dex/sdk": "^9.0.0",
-				"@walletconnect/auth-client": "^2.1.2",
 				"alchemy-sdk": "^3.6.5",
 				"bitcoinjs-lib": "^6.1.7",
 				"browser-image-compression": "^2.0.2",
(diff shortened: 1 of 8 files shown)


src/frontend/src/lib/schema/auth-broadcast.schema.ts
```diff
@@ -0,0 +1,10 @@
+import * as z from 'zod';
+
+// We broadcast an object containing both the message and the emitter.
+// This allows us to ignore the message in the tab that originally sent it.
+// For example, it prevents reloading the auth store if it's already synced,
+// since the postMessage is triggered after login.
+export const BroadcastDataSchema = z.strictObject({
+	msg: z.literal('authClientLoginSuccess'),
```
AntonioVentilii, line 8: I was thinking if we should create an enum for this string `authClientLoginSuccess`, instead of having it inside the class


or for now, we can leave it like this since we don't have more use-cases?

P.S. for another PR probably?

src/frontend/src/lib/providers/auth-broadcast.providers.ts
```diff
@@ -1,35 +1,61 @@
...
 
-	close = () => {
+	destroy = () => {
 		this.#bc.close();
+		AuthBroadcastChannel.#instance = null;
 	};
 
 	postLoginSuccess = () => {
-		this.#bc.postMessage(AuthBroadcastChannel.MESSAGE_LOGIN_SUCCESS);
+		const data: z.infer<typeof BroadcastDataSchema> = {
```
AntonioVentilii, line 50: why not a specific type for this one? it can be in this same PR, or I can do it in another, no worries. Just curious if there is a reason

src/frontend/src/lib/providers/auth-broadcast.providers.ts
```diff
@@ -1,35 +1,61 @@
...
 
 // If the user has more than one tab open in the same browser,
 // there could be a mismatch of the cached delegation chain vs the identity key of the `authClient` object.
 // This causes the `authClient` to be unable to correctly sign calls, raising Trust Errors.
 // To mitigate this, we use a `BroadcastChannel` to notify other tabs when a login has occurred, so that they can sync their `authClient` object.
 export class AuthBroadcastChannel {
+	static #instance: AuthBroadcastChannel | null;
+
 	readonly #bc: BroadcastChannel;
+	readonly #emitterId: string;
```
AntonioVentilii, line 14: for further iterations and to think about: would a `symbol` object for the ID guarantee even more "uniqueness" than `window.crypto.randomUUID()`? AND, can it be broadcasted?

AntonioVentilii (approved):
Thank you! LGTM! I just wrote some observations, but nothing is blocker, just a few doubts I have

src/frontend/src/lib/providers/auth-broadcast.providers.ts
AntonioVentilii: just speculation, since I imagine we don't have the use case with millions and millions of tabs opened

src/frontend/src/lib/schema/auth-broadcast.schema.ts
peterpeterparker: Agree, I was that close to use an enum too. I can change it here or afterwards, as you wish.

src/frontend/src/lib/providers/auth-broadcast.providers.ts
peterpeterparker: Because the interface is used - and should not be used - anywhere else for now. I hesitated scoping the schema within the provider but thought it was worth adding it to schemas in case we extend it in the future.

src/frontend/src/lib/providers/auth-broadcast.providers.ts
peterpeterparker: I don't think a symbol would be transported by `postMessage`. We would need to convert it to string and back to symbol which would not comply with the assertion.

src/frontend/src/lib/providers/auth-broadcast.providers.ts
peterpeterparker: Makes sense?

src/frontend/src/lib/schema/auth-broadcast.schema.ts
AntonioVentilii: let's do it in another PR, no worries, tks

src/frontend/src/lib/providers/auth-broadcast.providers.ts
AntonioVentilii: ah! makes sense! We can change it in the future if we have more

PR #10279, opened 2025-11-11
style(frontend): Remove `Carousel` fading effect

AntonioVentilii:
# Motivation

To be consistent with the other `Carousel` instances, we remove the fading efffect.


src/frontend/src/routes/(app)/+layout.svelte
AntonioVentilii:
```suggestion
```

PR #10299, opened 2025-11-12
feat(frontend): always display APY value with 2 decimals

DenysKarmazynDFINITY:
# Motivation

We'd like to always display staking APY value with 2 decimals.

Screenshot 2025-11-12 at 16 57 43


src/frontend/src/icp/components/stake/gldt/GldtStakeProvider.svelte
```diff
@@ -21,7 +21,7 @@
 			<div class="ml-3 w-full text-sm">
 				<div class="mb-1 font-bold">
 					{replacePlaceholders($i18n.stake.text.current_apy, {
-						$apy: `${$gldtStakeApyStore?.apy}`
+						$apy: `${($gldtStakeApyStore?.apy ?? 0).toFixed(2)}`
```
AntonioVentilii: maybe a small util to format the APY? so we centralize it somewhere in case is needed a change. We have a format.utils module, IIRC

DenysKarmazynDFINITY: yes, totally - added a util for that

AntonioVentilii (approved):

PR #10304, opened 2025-11-13
feat(frontend): Earning card component

daviddecentage:
# Motivation

We add a simple card component for our earning cards.

# Changes

Added component

# Tests

Added tests

### DIFF
--- a/src/frontend/src/tests/lib/components/earning/EarningOpportunityCard.spec.ts
+++ b/src/frontend/src/tests/lib/components/earning/EarningOpportunityCard.spec.ts
@@ -3,7 +3,6 @@ import { createMockSnippet } from '$tests/mocks/snippet.mock';
 import { assertNonNullish } from '@dfinity/utils';
 import { render, screen } from '@testing-library/svelte';
 
-
 describe('EarningOpportunityCard', () => {
 	const props = {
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/earning/EarningOpportunityCard.svelte
```diff
@@ -0,0 +1,31 @@
...
+				>{@render badge()}</span
+			>
+		</span>
+
+		<h3>{@render title()}</h3>
+
+		<span>{@render description()}</span>
+	</div>
+
+	<span>{@render button()}</span>
```
DenysKarmazynDFINITY, line 30: I would probably remove "span" from here since the button snippet will be anyway passed as a `button` tag, hence everything will be already wrapper into an HTML tag

DenysKarmazynDFINITY (approved):

src/frontend/src/lib/components/earning/EarningOpportunityCard.svelte
daviddecentage: That span is needed so it always appears at the bottom of the card. The button itself has flex-1 by default, and since we cant know what classes get passed to the button we wrap it.

PR #10308, opened 2025-11-13
feat(frontend): Earning yearly amount component

daviddecentage:
# Motivation

Adding a component to correctly format yearly amounts for earning values.

# Changes

Added component

# Tests

Added tests

Two varants of it:
image

### DIFF
--- a/src/frontend/src/lib/components/earning/EarningYearlyAmount.svelte
+++ b/src/frontend/src/lib/components/earning/EarningYearlyAmount.svelte
@@ -30,7 +30,7 @@
 	);
 </script>
 
-{#if nonNullish(yearlyAmount)}
+{#if nonNullish(yearlyAmount) && nonNullish(value)}
 	<span
 		class:text-brand-primary={!formatPositiveAmount}
 		class:text-success-primary={formatPositiveAmount && value > 0}


src/frontend/src/lib/components/earning/EarningYearlyAmount.svelte
```diff
@@ -0,0 +1,41 @@
...
+	interface Props {
+		value: number;
+		showPlusSign?: boolean;
+		formatPositiveAmount?: boolean;
+	}
+
+	const { value, showPlusSign = false, formatPositiveAmount = false }: Props = $props();
+
+	const yearlyAmount = $derived(
+		nonNullish(value)
```
DenysKarmazynDFINITY, line 20: The `value` prop is marked as required in `Props`. So let's either remove the nonNullish checks or make the value optional.

daviddecentage: good point, lets make it optional

src/frontend/src/tests/lib/components/earning/EarningYearlyAmount.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+		const { container } = render(EarningYearlyAmount, { value: 5, formatPositiveAmount: false });
+
+		const span = container.querySelector('span');
+
+		expect(span).toHaveClass('text-brand-primary');
+	});
+
+	it('renders nothing if value is null or undefined', () => {
+		const { container, rerender } = render(EarningYearlyAmount, {
+			value: undefined as unknown as number
```
DenysKarmazynDFINITY: since value can now be "undefined", we can remove the casting

daviddecentage: updated

PR #10317, opened 2025-11-13
refactor(frontend): Use built-in util `isIOS` from gix-cmp

AntonioVentilii:
# Motivation

Library `gix-components` already has an `isIOS` util. We can remove the ones we created.


roman-nazaruk (approved):
Thnx

PR #10318, opened 2025-11-13
feat(frontend): All earning cards list

daviddecentage:
# Motivation

A component that maps and renders all the data needed for the earning cards.

# Changes

Added component

# Tests

Added tests

image

### DIFF
--- a/src/frontend/src/icp/services/worker.dip20-wallet.services.ts
+++ b/src/frontend/src/icp/services/worker.dip20-wallet.services.ts
@@ -34,7 +34,13 @@ export class Dip20WalletWorker extends AppWorker implements WalletWorker {
 					| PostMessageDataResponseWalletCleanUp
 				>
 			>) => {
-				const { msg, data } = dataMsg;
+				const { ref, msg, data } = dataMsg;
+
+				// This is an additional guard because it may happen that the worker is initialised as a singleton.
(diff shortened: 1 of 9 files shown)


src/frontend/src/lib/components/earning/AllEarningOpportunityCardList.svelte
```diff
@@ -0,0 +1,67 @@
...
+			[EarningCardFields.TERMS]: $i18n.earning.terms.flexible,
+			action: () => goto(AppPath.EarningGold)
+		}
+	});
+</script>
+
+<div class="mt-5 flex grid grid-cols-2 gap-3 md:flex-row">
+	{#each earningCards as card, i (`${card.id}-${i}`)}
+		{#if card.id === currentReward.id}
+			<RewardsEarningOpportunityCard />
```
DenysKarmazynDFINITY, line 62: maybe just a bit shorter: `<RewardsEarningOpportunityCard ...{card.id === currentReward.id ? { ... } : {} }/>`

daviddecentage: The reward can only be one card currently. As of now, we always display the latest and the if condition is just there to reconize if its the rewards card. Couldve added a `type: "reward"` to the env or something. We will probably rework this once we get new rewards or new types of earnings so Ill leave it for now

PR #10329, opened 2025-11-14
feat(frontend): Make Scheduler `postMessage` type-safe

AntonioVentilii:
# Motivation

To make the Scheduler more type safe, and in general the `postMessage` flow, we should assign a type to the message that is similar/reflected to the one that we expect to receive.

# Changes

- Create sub-function to build a post message schema.
- Use it for `inferPostMessageSchema`.
- Create similar method but without request messaged for `inferPostMessageSchedulerSchema`.
- Create derived type and apply to class `Scheduler`. 

# Tests

Created tests.

### DIFF
--- a/src/frontend/src/eth/components/send/EthSendForm.svelte
+++ b/src/frontend/src/eth/components/send/EthSendForm.svelte
@@ -17,6 +17,7 @@
 	interface Props {
 		amount: OptionAmount;
 		destination?: string;
+		customNonce?: number;
 		nativeEthereumToken: Token;
 		selectedContact?: ContactUi;
 		onBack: () => void;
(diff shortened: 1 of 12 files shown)


src/frontend/src/lib/schedulers/scheduler.ts
```diff
@@ -124,7 +125,9 @@ export class SchedulerTimer {
 		this.setStatus('idle');
 	}
 
-	postMsg<T>(data: { msg: PostMessageResponse; data?: T } | PostMessageDataError) {
+	postMsg<T extends PostMessageDataRequest | PostMessageDataResponseLoose>(
```
AntonioVentilii:
```suggestion
	postMsg<T extends PostMessageDataResponseLoose>(
```

PR #10347, opened 2025-11-18
feat(frontend): Add reference to IC wallet scheduler errors

AntonioVentilii:
# Motivation

We should add the worker reference to the errors too in the IC wallet worker.

### DIFF
--- a/src/frontend/src/tests/icp/workers/ic-wallet-balance-and-transactions.worker.spec.ts
+++ b/src/frontend/src/tests/icp/workers/ic-wallet-balance-and-transactions.worker.spec.ts
@@ -250,6 +250,14 @@ describe('ic-wallet-balance-and-transactions.worker', () => {
 	}): TestUtil => {
 		let scheduler: IcWalletScheduler<PostMessageDataRequest>;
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/icp/schedulers/ic-wallet.scheduler.ts
```diff
@@ -19,7 +19,7 @@ export abstract class IcWalletScheduler<
 		| PostMessageDataRequestDip20
 > implements Scheduler<PostMessageDataRequest>
 {
-	#ref: PostMessageCommon['ref'];
+	protected ref: PostMessageCommon['ref'];
```
DenysKarmazynDFINITY, line 22: Shall we use here the same approach as in the sibling PR, the one with a setter?

DenysKarmazynDFINITY: Or other way around, don't really have a preference

DenysKarmazynDFINITY: nevermind, I mislooked that `ref = ...` was inside a spec file :)

PR #10353, opened 2025-11-18
refactor(frontend): Earning page

daviddecentage:
# Motivation

This PR ties together all the earning card components by displaying them on the earnings page.

We can also remove the temporary cards etc.

# Changes

- Added contexts to earnings page
- Removed existing cards
- Removed EarningCategoryCard
- Removed unneeded translations

# Tests

Coverage should go up since were removing untested components, as well as test added for the root Earnings component.

### DIFF
--- a/src/frontend/src/lib/components/earning/Earning.svelte
+++ b/src/frontend/src/lib/components/earning/Earning.svelte
@@ -1,15 +1,15 @@
 <script lang="ts">
 	import { setContext } from 'svelte';
 	import GldtStakeContext from '$icp/components/stake/gldt/GldtStakeContext.svelte';
-	import AllEarningOpportunityCardList from '$lib/components/earning/AllEarningOpportunityCardList.svelte';
-	import RewardsEligibilityContext from '$lib/components/rewards/RewardsEligibilityContext.svelte';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/earning/Earning.svelte
```diff
@@ -1,51 +1,32 @@
...
 	import { i18n } from '$lib/stores/i18n.store.js';
-</script>
-
-<div class="flex flex-col">
-	<PageTitle>{$i18n.earning.text.title}</PageTitle>
 
-	<!-- Todo: refactor this once the design is clear -->
-	<!-- we know we want to show these 3 cards, thats why the translations have been added already -->
+	setContext<GldtStakeContextType>(GLDT_STAKE_CONTEXT_KEY, {
+		store: initGldtStakeStore()
```
DenysKarmazynDFINITY, line 15: In the follow-up PR, can you please make the gldt stake store global and use it here + on the GLDT stake page? just to avoid loading the respective data twice

daviddecentage: Yes, will do it in a follow up

PR #10360, opened 2025-11-19
feat(backend)!: Add EXT v2 tokens to Custom tokens variants

AntonioVentilii:
# Motivation

We are going to onboard NFTs in IC. One of the most used standard is [EXT v2 Token](https://github.com/Toniq-Labs/ext-v2-token).

We need to include it among the variants for custom tokens, since we are not going to treat them the same as ICRC tokens.

BREAKING CHANGE: The interface of the backend now includes a new variant.


src/shared/src/impls.rs
```diff
@@ -551,8 +552,10 @@ impl Validate for ErcTokenId {
 impl Validate for CustomTokenId {
     fn validate(&self) -> Result<(), candid::Error> {
         match self {
-            CustomTokenId::Icrc(_) => Ok(()), /* This is a principal.  In principle, we could */
-            // check the exact type of principal.
+            CustomTokenId::Icrc(_) | CustomTokenId::ExtV2(_) => Ok(()), /* This is a principal. */
+            // In principle, we
```
DenysKarmazynDFINITY, line 556: Is it linter who broke this into several lines?

AntonioVentilii: yep... i tried to leave it as before, but nothing...

PR #10371, opened 2025-11-19
feat(frontend): added universal scanner modal

roman-nazaruk:
# Motivation

We aim to integrate with with OpenCryptoPay and we need to add new modal and show icon to opne it in Header

# Changes

- Added new Scanner Modal
- Updated Header component
- Added new variable to modal store

### DIFF
--- a/src/frontend/src/lib/components/hero/Header.svelte
+++ b/src/frontend/src/lib/components/hero/Header.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import { nonNullish } from '@dfinity/utils';
 	import { page } from '$app/state';
+	import { UNIVERSAL_SCANNER_ENABLED } from '$env/universal-scanner.env';
 	import AboutWhyOisy from '$lib/components/about/AboutWhyOisy.svelte';
 	import AboutWhyOisyModal from '$lib/components/about/AboutWhyOisyModal.svelte';
 	import HelpMenu from '$lib/components/core/HelpMenu.svelte';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/scanner/Scanner.svelte
```diff
@@ -0,0 +1,22 @@
...
+	import { modalStore } from '$lib/stores/modal.store';
+	import { modalUniversalScannerOpen } from '$lib/derived/modal.derived';
+	import IconScanLine from '$lib/components/icons/IconScanLine.svelte';
+	import { fade } from 'svelte/transition';
+	import { i18n } from '$lib/stores/i18n.store';
+
+	const modalId = Symbol();
+</script>
+
+<button
```
AntonioVentilii: Let's use `ButtonIcon`, just to centralize the styles

src/frontend/src/lib/components/hero/Header.svelte
```diff
@@ -46,6 +60,9 @@
 
 		{#if $authSignedIn}
 			<WalletConnect />
+			{#if UNIVERSAL_SCANNER_ENABLED}
```
AntonioVentilii, line 64:
```suggestion

			{#if UNIVERSAL_SCANNER_ENABLED}
```

src/frontend/src/lib/components/scanner/Scanner.svelte
roman-nazaruk: Resolved

PR #10381, opened 2025-11-19
feat(frontend): added lnurl decoding util

roman-nazaruk:
# Motivation

We aim to integrate with DFX OpencryptoPay and we need to util for decoding lnurl url

# Changes

Added new decodeLNURL and return utf url

# Tests

Covered with tests


src/frontend/src/lib/utils/open-crypto-pay.utils.ts
```diff
@@ -0,0 +1,24 @@
...
+ * @param lnurl - LNURL string (e.g., "LNURL1DP68GURN8GHJ7...")
+ * @returns Decoded URL string
+ * @throws {Error} If LNURL is invalid or decoding failed
+ *
+ */
+export const decodeLNURL = (lnurl: string): string => {
+	const { words } = decode(lnurl, 2000);
+	const bytes = fromWords(words);
+
+	const url = new TextDecoder().decode(new Uint8Array(bytes));
```
AntonioVentilii, line 21: i am assuming that bytes is not returned as Uint8Array already?

roman-nazaruk: no it returns as number[] and it is not Uint8Array

PR #10402, opened 2025-11-19
feat(frontend): added OpenCryptoPay component

roman-nazaruk:
# Motivation

We aim to show OpenCryptoPay Bill details

# Changes

Added new component OpenCryptoPay to show PAyHero and RecipientData

# Tests

### DIFF
--- a/scripts/generate.sh
+++ b/scripts/generate.sh
@@ -95,7 +95,8 @@ generate_declarations() {
     # That's why we have to post-process the results.
     mv "${generatedTsfile}" "${didfolder}"
     mv "${generatedJsfile}" "${didfolder}"
-    rm -r "${generatedFolder}"
+    # TODO: re-remove the generated folder once we adapt all imports to the new "old" location.
+    # rm -r "${generatedFolder}"
   else
     echo "DID file skipped: $didfile"
   fi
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/scanner/OpenCryptoPay.svelte
```diff
@@ -0,0 +1,33 @@
...
+<ContentWithToolbar>
+	<div>
+		{#if nonNullish($data)}
+			<PayHero
+				amount={$data.requestedAmount.amount}
+				asset={$data.requestedAmount.asset}
+				receipt={$data.displayName}
+			/>
+
+			<ReceiptData recipient={$data?.recipient} />
```
AntonioVentilii: wouldn't it better to display it only when recipient is non nullish?

Otherwise I am misunderstanding the `RecipientData` scope

roman-nazaruk: Agree, updated

src/frontend/src/lib/components/scanner/PayReceiptData.svelte
```diff
@@ -6,7 +6,7 @@
 	import { formatAddress } from '$lib/utils/open-crypto-pay.utils';
 
 	interface Props {
-		recipient: Recipient;
+		recipient?: Recipient;
```
AntonioVentilii: no need for this anymore, right?

roman-nazaruk: Yeap, removed

PR #10403, opened 2025-11-19
feat(frontend): added new ScannerCodeInput component

roman-nazaruk:
# Motivation

Added new ScannerCodeInput to show error and placeholder for scanner input

# Changes

Added new ScannerCodeInput to show error and placeholder for scanner input

# Tests

Covered with tests

### DIFF
--- a/src/frontend/src/lib/components/scanner/OpenCryptoPay.svelte
+++ b/src/frontend/src/lib/components/scanner/OpenCryptoPay.svelte
@@ -1,0 +1,35 @@
+<script lang="ts">
+	import { nonNullish } from '@dfinity/utils';
+	import { getContext } from 'svelte';
+	import PayHero from '$lib/components/scanner/PayHero.svelte';
+	import ReceiptData from '$lib/components/scanner/PayReceiptData.svelte';
+	import Button from '$lib/components/ui/Button.svelte';
+	import ButtonGroup from '$lib/components/ui/ButtonGroup.svelte';
+	import ContentWithToolbar from '$lib/components/ui/ContentWithToolbar.svelte';
+	import { PAY_CONTEXT_KEY, type PayContext } from '$lib/stores/open-crypto-pay.store';
+
+	const { data } = getContext<PayContext>(PAY_CONTEXT_KEY);
+</script>
+
+<ContentWithToolbar>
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/components/scanner/ScannerCodeInput.svelte
```diff
@@ -0,0 +1,32 @@
...
+		value: string;
+		error?: string;
+		label: string;
+		placeholder: string;
+		name: string;
+	}
+
+	let { value = $bindable(''), error, label, placeholder, name }: Props = $props();
+
+	let borderColor = $derived(error ? 'var(--color-border-error-solid)' : 'inherit');
```
AntonioVentilii: let's use tailwind classes

```
const isError = $derived(nonNullish(error)) // or notEmptyString(error) -- i see you use it later on too

class:whatever-color={isError}
class:whatever-inherit={!isError}
```

roman-nazaruk: If I’m not mistaken, you can’t set custom CSS properties directly through classes in tailwind, so that’s why I used this approach.

AntonioVentilii: `class:border-error-solid={isError}`

src/frontend/src/lib/components/scanner/ScannerCodeInput.svelte
```diff
@@ -0,0 +1,32 @@
...
+	let borderColor = $derived(notEmptyString(error) ? 'var(--color-border-error-solid)' : 'inherit');
+</script>
+
+<div>
+	<p class="mt-2 mb-1 md:text-center">{label}</p>
+
+	<div style={`--input-custom-border-color: ${borderColor}`}>
+		<InputText {name} {placeholder} bind:value />
+
+		{#if notEmptyString(error)}
```
AntonioVentilii: let's create a `$derived` for `notEmptyString(error)` since you are using it twice

roman-nazaruk: Done

AntonioVentilii (approved):

PR #10404, opened 2025-11-19
feat(frontend): added scanner code component

roman-nazaruk:
# Motivation

We aim to show new Scanner component for universalScan with BottomSheet

# Changes

Created new ScannerCode component

# Tests

Covered with tests

### DIFF
--- a/src/frontend/src/tests/lib/components/scanner/ScannerCode.spec.ts
+++ b/src/frontend/src/tests/lib/components/scanner/ScannerCode.spec.ts
@@ -11,6 +11,18 @@ vi.mock('$lib/services/open-crypto-pay.services', () => ({
 	processOpenCryptoPayCode: vi.fn()
 }));
 
+vi.mock('@dfinity/gix-components', async () => {
+	const actual = await vi.importActual('@dfinity/gix-components');
+	return {
+		...actual,
+		QRCodeReader: vi.fn().mockImplementation(() => ({
+			$$render: () => '<div data-tid="mock-qr-reader">Mocked QR Reader</div>',
+			$$slots: {},
+			$$scope: {}
+		}))
+	};
+});
+
 describe('ScannerCode.svelte', () => {
 	const mockOnNext = vi.fn();
 	const mockSetData = vi.fn();
@@ -86,6 +98,16 @@ describe('ScannerCode.svelte', () => {
 		renderWithContext();
 
 		expect(screen.getByTestId(OPEN_CRYPTO_PAY_ENTER_MANUALLY_BUTTON)).toBeInTheDocument();
+	});
+
+	it('should show input after clicking enter manually', async () => {
+		renderWithContext();
+
+		await openManualEntry();
+
+		await waitFor(() => {
+			expect(screen.getByPlaceholderText(en.scanner.text.enter_or_paste_code)).toBeInTheDocument();
+		});
 	});
 
 	it('should disable continue button when empty', async () => {


src/frontend/src/lib/components/qr/QrCodeScanner.svelte
```diff
@@ -28,7 +29,9 @@
 			onScan(result);
 		}
 
-		onBack();
+		if (nonNullish(onBack)) {
+			onBack();
+		}
```
AntonioVentilii:
```suggestion
			onBack?.();
```

src/frontend/src/lib/components/scanner/ScannerCode.svelte
```diff
@@ -0,0 +1,128 @@
...
+			onNext();
+		} catch (_: unknown) {
+			error = $i18n.scanner.error.code_link_is_not_valid;
+		} finally {
+			busy.stop();
+		}
+	};
+
+	const handleScan = async ({ status, code }: { status: QrStatus; code?: string }) => {
+		if (status !== 'success' || !code) {
```
AntonioVentilii: isNullish(code)

src/frontend/src/lib/components/scanner/ScannerCode.svelte
```diff
@@ -0,0 +1,128 @@
...
+			return;
```
AntonioVentilii, line 52: just to confirm: if there is an error with scanning the code, we do nothing?

src/frontend/src/lib/components/scanner/ScannerCode.svelte
```diff
@@ -0,0 +1,128 @@
...
+				<Button
+					colorStyle="primary"
+					innerStyleClass="flex items-center justify-center"
+					link
+					onclick={() => {
+						uri = '';
+						error = '';
+						openBottomSheet = true;
+					}}
+					testId="button-enter-manually"
```
AntonioVentilii: let's put it among the test id constants

src/frontend/src/lib/components/scanner/ScannerCode.svelte
```diff
@@ -0,0 +1,128 @@
...
+		}
+		await processCode(code);
+	};
+
+	const handleManualConnect = async () => {
+		await processCode(uri);
+	};
+
+	$effect(() => {
+		if (isEmptyString(uri)) {
```
AntonioVentilii: is this the same as `invalid` variable? if yes, shall it be renamed with something like `invalidUri` or `emptyUri`?

src/frontend/src/lib/components/scanner/ScannerCode.svelte
```diff
@@ -0,0 +1,128 @@
...
+			error = '';
```
AntonioVentilii, line 63: may you please add a test that verify that this `$effect` is not triggered by the changes of `uri`?

src/frontend/src/lib/components/scanner/ScannerCode.svelte
AntonioVentilii: I think the component `QrCodeScanner` triggers `onScan` only when the result is successfull

src/frontend/src/lib/components/scanner/ScannerCode.svelte
roman-nazaruk: Yes, we need to show invalid code and keep Scanner open

src/frontend/src/lib/components/scanner/ScannerCode.svelte
roman-nazaruk: Yes, but status from component is export type QrStatus = 'success' | 'cancelled' | 'token_incompatible';

src/frontend/src/lib/components/scanner/ScannerCode.svelte
roman-nazaruk: Added new test

src/frontend/src/lib/components/scanner/ScannerCode.svelte
roman-nazaruk: Good catch, fixed and renamed to isEmptyString

src/frontend/src/lib/components/scanner/ScannerCode.svelte
AntonioVentilii: will the logic to show invalid code be done after? or is already somewhere?

roman-nazaruk: No, if we scan and it can not parse it then we show scanner that the flow

src/frontend/src/lib/components/scanner/ScannerCode.svelte
roman-nazaruk: Done

src/frontend/src/lib/components/scanner/ScannerCode.svelte
roman-nazaruk: Done

src/frontend/src/lib/constants/test-ids.constants.ts
```diff
@@ -313,3 +313,6 @@ export const STAKE_PROVIDER_EXTERNAL_URL = 'stake-provider-external-url';
 // PWA
 export const PWA_INFO_BANNER_TEST_ID = 'pwa-info-banner';
 export const PWA_INFO_BANNER_CLOSE_BUTTON_TEST_ID = 'pwa-info-banner-close-button';
+
+//OpenCryptoPay
```
AntonioVentilii:
```suggestion
// OpenCryptoPay
```

PR #10422, opened 2025-11-20
feat(frontend): create SwapLoader component

DenysKarmazynDFINITY:
# Motivation

The Swap wizard is going to be re-used in 2 different modals. Therefore, we need to update the initial data loading the way that it becomes reusable. To achieve that, we create a component which loads all necessary data (previously, it was done in `Swap.svelte`).

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapLoader.svelte
+++ b/src/frontend/src/lib/components/swap/SwapLoader.svelte
@@ -1,0 +1,74 @@
+<script lang="ts">
+	import { isNullish } from '@dfinity/utils';
+	import type { Snippet } from 'svelte';
+	import { ICP_TOKEN } from '$env/tokens/tokens.icp.env';
+	import {
+		loadDisabledIcrcTokensBalances,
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/swap/SwapLoader.svelte
```diff
@@ -0,0 +1,74 @@
...
+
+	interface Props {
+		button: Snippet<[onclick: (onSwapReady: () => void) => void]>;
+	}
+
+	let { button }: Props = $props();
+
+	const isDisabled = (): boolean => isNullish($kongSwapTokensStore);
+
+	const loadKongSwapTokens = async (): Promise<'ready' | undefined> => {
```
AntonioVentilii, line 25: don't know if useful but we have a `ResultSuccess` type

DenysKarmazynDFINITY: I think we will be aiming to get rid of this loader completely, so I will keep it as it was in Swap.svelte just to avoid any new changes

PR #10424, opened 2025-11-20
feat(frontend): Canister class for `ExtV2TokenCanister`

AntonioVentilii:
# Motivation

Similar to what we do with other third-party canisters, we create a class to interface with an EXT v2 Token canister.

# Changes

- Create class `ExtV2TokenCanister`.
- Add first method `transactions`.

# Tests

Added tests.

### DIFF
--- a/scripts/generate.sh
+++ b/scripts/generate.sh
@@ -95,7 +95,8 @@ generate_declarations() {
     # That's why we have to post-process the results.
     mv "${generatedTsfile}" "${didfolder}"
     mv "${generatedJsfile}" "${didfolder}"
-    rm -r "${generatedFolder}"
+    # TODO: re-remove the generated folder once we adapt all imports to the new "old" location.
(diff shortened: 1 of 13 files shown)


src/frontend/src/icp/canisters/ext-v2-token.canister.ts
```diff
@@ -0,0 +1,35 @@
+import type {
+	_SERVICE as ExtV2TokenService,
+	Transaction
+} from '$declarations/ext_v2_token/declarations/ext_v2_token.did';
+import { idlFactory as idlCertifiedFactoryXtcLedger } from '$declarations/xtc_ledger/xtc_ledger.factory.certified.did';
```
DenysKarmazynDFINITY: I think you importing it from the wrong canister folder

AntonioVentilii: Good catch!

PR #10428, opened 2025-11-20
refactor(frontend): Avoid referring to subfolder for `backend` bindings

AntonioVentilii:
# Motivation

With the new bindings generation script (PR https://github.com/dfinity/oisy-wallet/pull/10409), we prefer to not have the subfolder `declarations` that bindgen creates, but to consolidate the bindings in the same folder.

However, our current imports are using such subfolder.

The plan is:

- Undo the removal of the subfolder (temporarily).
- Adapt all the imports to use the "old" main folder for declarations (split in a few PRs, since it is quite a lot of changes).
- Re-remove the subfolder.

This PR is part of the second step: we undo the reference to subfolder for the `backend` bindings that was done in PR https://github.com/dfinity/oisy-wallet/pull/9565

### DIFF
--- a/src/frontend/src/lib/api/kong_backend.api.ts
+++ b/src/frontend/src/lib/api/kong_backend.api.ts
@@ -1,7 +1,4 @@
-import type {
-	SwapAmountsReply,
-	TokenReply
-} from '$declarations/kong_backend/declarations/kong_backend.did';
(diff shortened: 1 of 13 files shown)


peterpeterparker (approved):
LGTM, thx

PR #10442, opened 2025-11-20
feat(frontend): implement SwapModalWizardSteps component

DenysKarmazynDFINITY:
# Motivation

The next step of making the Swap wizard re-usable is to extract all its steps into a separate component.

### DIFF
--- a/src/frontend/src/lib/components/swap/SwapModalWizardSteps.svelte
+++ b/src/frontend/src/lib/components/swap/SwapModalWizardSteps.svelte
@@ -80,6 +80,7 @@
 			setFilterNetwork($sourceToken.network);
 
 			setAllowedNetworkIds(SUPPORTED_CROSS_SWAP_NETWORKS[$sourceToken.network.id]);
+			return;
 		}
 
 		if (nonNullish($destinationToken) && nonNullish($sourceToken)) {
@@ -95,6 +96,7 @@
 
 		if (nonNullish($sourceToken)) {
 			setFilterNetwork($sourceToken.network);
+			return;
 		}
 
 		if (nonNullish($destinationToken) && isNullish($sourceToken)) {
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/swap/SwapModalWizardSteps.svelte
```diff
@@ -0,0 +1,196 @@
...
+			allNetworksEnabled = false;
+			setAllowedNetworkIds([$sourceToken.network.id]);
+			return;
+		}
+
+		if (isNullish($destinationToken) && nonNullish($sourceToken)) {
+			allNetworksEnabled = false;
+			setFilterNetwork($sourceToken.network);
+
+			setAllowedNetworkIds(SUPPORTED_CROSS_SWAP_NETWORKS[$sourceToken.network.id]);
```
AntonioVentilii, line 82: should there be a `return` here?

src/frontend/src/lib/components/swap/SwapModalWizardSteps.svelte
```diff
@@ -0,0 +1,196 @@
...
+		}
+	};
+
+	const showSourceTokenList = () => {
+		allNetworksEnabled = true;
+		resetAllowedNetworkIds();
+
+		if (nonNullish($sourceToken)) {
+			setFilterNetwork($sourceToken.network);
```
AntonioVentilii, line 98: same question

src/frontend/src/lib/components/swap/SwapModalWizardSteps.svelte
```diff
@@ -0,0 +1,196 @@
...
+
+		goToWizardStep({
+			modal,
+			steps,
+			stepName
+		});
+	};
+</script>
+
+{#if currentStep?.name === WizardStepsSwap.TOKENS_LIST}
```
AntonioVentilii: can you please wrap it all with `#key` ?

src/frontend/src/lib/components/swap/SwapModalWizardSteps.svelte
DenysKarmazynDFINITY: It was like this in Swap.svelte - but no I guess we can add `return` here just for better readability, even if functionally it doesn't change anything

src/frontend/src/lib/components/swap/SwapModalWizardSteps.svelte
DenysKarmazynDFINITY: same as above - added

src/frontend/src/lib/components/swap/SwapModalWizardSteps.svelte
DenysKarmazynDFINITY: sure, lost it while moving code

PR #10443, opened 2025-11-20
feat(frontend): Highest earning APY derived

daviddecentage:
# Motivation

For the upcoming earn page header, we need some summarized values (across all staking options) for certain data like APY, staked amount etc.

# Changes

Added highestApyEarningData derived

# Tests

Added tests for the new derived

### DIFF
--- a/scripts/generate.sh
+++ b/scripts/generate.sh
@@ -93,11 +93,9 @@ generate_declarations() {
     # icp-bindgen generates the files in a `declarations` subfolder
     # using a different suffix for JavaScript as the one we used to use.
     # That's why we have to post-process the results.
-    # TODO: change back `cp` to `mv` once we adapt all imports to the new "old" location.
-    cp "${generatedTsfile}" "${didfolder}"
-    cp "${generatedJsfile}" "${didfolder}"
-    # TODO: re-remove the generated folder once we adapt all imports to the new "old" location.
-    # rm -r "${generatedFolder}"
+    mv "${generatedTsfile}" "${didfolder}"
+    mv "${generatedJsfile}" "${didfolder}"
+    rm -r "${generatedFolder}"
   else
     echo "DID file skipped: $didfile"
   fi


src/frontend/src/lib/derived/earning.derived.ts
```diff
@@ -64,3 +64,37 @@ export const earningData: Readable<EarningData> = derived(
...
 	}
 );
+
+export const highestApyEarningData: Readable<EarningDataRecord | undefined> = derived(
+	[earningData],
+	([$earningData]) => {
+		const entries = Object.values($earningData);
+
+		if (entries.length === 0) {
+			return undefined;
```
DenysKarmazynDFINITY:
```suggestion
			return;
```

src/frontend/src/lib/derived/earning.derived.ts
```diff
@@ -64,3 +64,37 @@ export const earningData: Readable<EarningData> = derived(
...
+		}
+
+		return entries.reduce<EarningDataRecord | undefined>((highest, record) => {
+			const rawCurrent = record[EarningCardFields.APY];
```
DenysKarmazynDFINITY: I would shorten this method a bit (no harm in using Number even if the value is already this type):
```
const currentApy = Number.isFinite(record[EarningCardFields.APY]) ? Number(record[EarningCardFields.APY]) : undefined;
```

src/frontend/src/lib/derived/earning.derived.ts
```diff
@@ -64,3 +64,37 @@ export const earningData: Readable<EarningData> = derived(
...
+			if (!Number.isFinite(currentApy)) {
+				return highest;
+			}
+
+			if (isNullish(highest)) {
+				return record;
+			}
+
+			const rawHighest = highest[EarningCardFields.APY];
+			const highestApy = typeof rawHighest === 'number' ? rawHighest : Number(rawHighest);
```
DenysKarmazynDFINITY: const highestApy = Number(highest[EarningCardFields.APY]);

daviddecentage: optimized the conditions etc.

src/frontend/src/lib/derived/earning.derived.ts
daviddecentage: optimized the conditions etc.

src/frontend/src/lib/derived/earning.derived.ts
daviddecentage: updated

PR #10445, opened 2025-11-20
feat(frontend): All earning positions usd derived

daviddecentage:
# Motivation

For the upcoming earn page header, we need some summarized values (across all staking options) for certain data like APY, staked amount etc.

# Changes

Added derived for a sum of all stake position amounts

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/derived/earning.derived.ts
+++ b/src/frontend/src/lib/derived/earning.derived.ts
@@ -94,8 +94,11 @@ export const highestApyEarningData: Readable<EarningDataRecord | undefined> = de
 );
 
 export const allEarningPositionsUsd: Readable<number> = derived([earningData], ([$earningData]) =>
-	Object.values($earningData).reduce<number>((acc, record) => {
-		const currentEarning = Number(record[EarningCardFields.CURRENT_EARNING] ?? 0);
-		return isNaN(currentEarning) ? acc : acc + currentEarning;
-	}, 0)
+	Object.values($earningData).reduce<number>(
+		(acc, record) =>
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/derived/earning.derived.ts
```diff
@@ -92,3 +92,10 @@ export const highestApyEarningData: Readable<EarningDataRecord | undefined> = de
 		}, undefined);
 	}
 );
+
+export const allEarningPositionsUsd: Readable<number> = derived([earningData], ([$earningData]) =>
+	Object.values($earningData).reduce<number>((acc, record) => {
+		const currentEarning = Number(record[EarningCardFields.CURRENT_EARNING] ?? 0);
+		return isNaN(currentEarning) ? acc : acc + currentEarning;
```
DenysKarmazynDFINITY: You have a fallback (0) therefore no need in isNan

daviddecentage: The fallback is in case of nullish but the isNaN condition is ensuring that no non-number data comes in from `record[EarningCardFields.CURRENT_EARNING]` since it could theoretically be string as well

DenysKarmazynDFINITY: let’s move then the check to currentEarning, so it’s always a number

DenysKarmazynDFINITY (approved):

src/frontend/src/lib/derived/earning.derived.ts
daviddecentage: made it a oneliner :white_check_mark:

PR #10451, opened 2025-11-20
feat(frontend): Earning position card

daviddecentage:
# Motivation

We want to display a summary of all earning methods current performance.

# Changes

Added component

# Tests

Added tests

image

### DIFF
--- a/src/frontend/src/lib/components/earning/EarningPositionCard.svelte
+++ b/src/frontend/src/lib/components/earning/EarningPositionCard.svelte
@@ -1,13 +1,13 @@
 <script lang="ts">
+	import EarningYearlyAmount from '$lib/components/earning/EarningYearlyAmount.svelte';
 	import StakeContentCard from '$lib/components/stake/StakeContentCard.svelte';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/earning/EarningPositionCard.svelte
```diff
@@ -0,0 +1,36 @@
...
+		<div class="my-1 text-lg font-bold sm:text-xl">
+			<EarningYearlyAmount formatPositiveAmount value={$allEarningYearlyAmountUsd}>
+				{#snippet fallback()}
+					<SkeletonText />
+				{/snippet}
+			</EarningYearlyAmount>
+		</div>
+
+		<div class="flex justify-center gap-2 text-sm sm:text-base">
+			<span class="font-bold">
```
DenysKarmazynDFINITY: `span` is probably not needed

daviddecentage: Right, adjusted

PR #10454, opened 2025-11-21
refactor(frontend): Fallback for yearly earning card values

daviddecentage:
# Motivation

On the earning opportunity cards we want to display a hyphen if the value is not yet loaded.

# Changes

Added fallback snippet for the yearly amounts

# Tests

image


src/frontend/src/lib/components/earning/DefaultEarningOpportunityCard.svelte
```diff
@@ -50,7 +50,11 @@
 								value={nonNullish(cardFields[cardField])
 									? Number(cardFields[cardField])
 									: undefined}
-							/>
+							>
+								{#snippet fallback()}
+									-
```
DenysKarmazynDFINITY, line 55: Are there any cases where we don't wanna use fallback? Wondering if we can display "-" by default inside the component itself

daviddecentage: I was thinking the same but we havent displayed a hyphen as a fallback ever so for consistency we maybe shouldve gone for the skeleton as a default? Not sure. If not passed we just dont show any fallback which at least doesnt break anything either. We can certainly revise it the more we use it.

PR #10460, opened 2025-11-21
feat(frontend): implement GetTokenCardContent component

DenysKarmazynDFINITY:
# Motivation

We need to implement a component that is going to show content in the GetTokenModal.

Screenshot 2025-11-21 at 10 32 47

### DIFF
--- a/src/frontend/src/lib/components/get-token/GetTokenCardContent.svelte
+++ b/src/frontend/src/lib/components/get-token/GetTokenCardContent.svelte
@@ -1,0 +1,86 @@
+<script lang="ts">
+	import type { Snippet } from 'svelte';
+	import { fade } from 'svelte/transition';
+	import { GET_TOKEN_MODAL_POTENTIAL_USD_BALANCE } from '$lib/constants/test-ids.constants';
+	import { currentCurrency } from '$lib/derived/currency.derived';
+	import { exchanges } from '$lib/derived/exchange.derived';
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/get-token/GetTokenCardContent.svelte
```diff
@@ -0,0 +1,86 @@
...
+			$i18n.stake.text.active_earning_per_year,
+			{
+				$amount: `${formatCurrency({
+					value: (potentialTokensUsdBalance * currentApy) / 100,
+					currency: $currentCurrency,
+					exchangeRate: $currencyExchangeStore,
+					language: $currentLanguage
+				})}`
+			}
+		)}
```
daviddecentage, line 84: lets maybe add the `EarningYearlyAmount` component here

daviddecentage (approved):
LGTM, approving but I think we can reuse a component there

src/frontend/src/lib/components/get-token/GetTokenCardContent.svelte
DenysKarmazynDFINITY: good idea, I will take a look at it in a separate PR

daviddecentage: Discussed offline: we will refactor it together with other instances in a seperate PR

PR #10464, opened 2025-11-21
feat(frontend): EXT token method to fetch the balance

AntonioVentilii:
# Motivation

We keep adding methods useful to manage the EXT tokens: in this PR we add the [balance](https://github.com/Toniq-Labs/ext-v2-token/blob/main/API-REFERENCE.md#balance-1) method to fetch the balance of a user for a specific token of the collection.

# Changes

- Create error handling for the `balance` request of the EXT v2 token canister (similar to what we do for other canisters).
- Create method in class `ExtV2TokenCanister` to interface with `balance` request.

# Tests

Added tests.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #10477, opened 2025-11-21
feat(frontend): Collapsible list

daviddecentage:
# Motivation

According to design we will use a sort of collapsible list items a lot more for staking UIs, so we create a reusable component for it.

# Changes

- Added list component
- Added list item component

# Tests

Added tests


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #10497, opened 2025-11-24
feat(build): Create script to load the NFT EXT v2 collections

AntonioVentilii:
# Motivation

There is no direct on-chain way to know the list of existing EXT v2 tokens/collections. But we require it, since we will create the same type of worker that we currently have for Ethereum/EVM NFTs: we want to periodically check if the user has any NFT pertaining to a list of collections.

So, as first step we need to maintain such a list (very similar to what we do for SNSes).

For now we found two sources for a extensive list:

- [Toniq exchange](https://toniq.io/);
- [nftGeek](https://t5t44-naaaa-aaaah-qcutq-cai.raw.icp0.io/).

Scraping their network requests, we find the URL to be queried.

# Note

The user will have the possibility to manually add a collection if it is not in the maintained list yet.

# Changes

- Create small script to fetch and parse data from Toniq.
- Same for nftGeek.
- Merge the data and save them: we will not remove the existing collections that are not in the fetched data anymore.
- Create script to run the module.

# Tests

I manually ran the script and save the created file.

### DIFF
--- a/scripts/build.tokens.ext.ts
+++ b/scripts/build.tokens.ext.ts
@@ -160,10 +160,8 @@ const getCollections = async (): Promise<EnvExtToken[]> => {
 	return Object.values(collectionsMap).sort((a, b) => a.canisterId.localeCompare(b.canisterId));
 };
 
-
-
 try {
-const tokens = await getCollections();
+	const tokens = await getCollections();
 
 	writeFileSync(EXT_COLLECTIONS_JSON_FILE, JSON.stringify(tokens, jsonReplacer, 8));
 } catch (err) {


scripts/build.tokens.ext.ts
```diff
@@ -0,0 +1,174 @@
...
+	const existingCollections = readExistingCollections();
+
+	const toniqCollections = await getToniqCollections();
+	const nftGeekCollections = await getNftGeekCollections();
+
+	// We want to keep the existing collections in the file that are not in the fetched lists anymore.
+	const collectionsMap = [
+		...existingCollections,
+		...toniqCollections,
+		...nftGeekCollections
```
AntonioVentilii:
```suggestion
		...nftGeekCollections,
				...toniqCollections,
```

scripts/build.tokens.ext.ts
```diff
@@ -0,0 +1,174 @@
...
+	].reduce<Record<string, EnvExtToken>>((acc, collection) => {
+		acc[collection.canisterId] = collection;
+
+		return acc;
+	}, {});
+
+	return Object.values(collectionsMap).sort((a, b) => a.canisterId.localeCompare(b.canisterId));
+};
+
+const main = async () => {
```
roman-nazaruk: Maybe we should use a meaningful name for tthis function?

roman-nazaruk (approved):

scripts/build.tokens.ext.ts
```diff
@@ -0,0 +1,174 @@
...
+	const tokens = await getCollections();
+
+	writeFileSync(EXT_COLLECTIONS_JSON_FILE, JSON.stringify(tokens, jsonReplacer, 8));
+};
+
+try {
+	await main();
```
AntonioVentilii:
```suggestion
const tokens = await getCollections();

	writeFileSync(EXT_COLLECTIONS_JSON_FILE, JSON.stringify(tokens, jsonReplacer, 8));
```

scripts/build.tokens.ext.ts
```diff
@@ -0,0 +1,174 @@
...
+
+	return Object.values(collectionsMap).sort((a, b) => a.canisterId.localeCompare(b.canisterId));
+};
+
+const main = async () => {
+	const tokens = await getCollections();
+
+	writeFileSync(EXT_COLLECTIONS_JSON_FILE, JSON.stringify(tokens, jsonReplacer, 8));
+};
+
```
AntonioVentilii:
```suggestion



```

scripts/build.tokens.ext.ts
AntonioVentilii: actually, it is not really needed: deconstructed directly eheh

PR #10504, opened 2025-11-24
feat(frontend): Services to load EXT custom tokens

AntonioVentilii:
# Motivation

Similar to what we do for the other variants of custom tokens, we create a service to load the EXT custom tokens.

# Changes

- Create EXT custom tokens store.
- Create service to build the custom tokens from backend, after having filtered the EXT tokens.
- The metadata are currently fetched ONLY via the curated list. 

# Tests

Added tests.

### DIFF
--- a/src/frontend/src/tests/icp/services/ext.service.spec.ts
+++ b/src/frontend/src/tests/icp/services/ext.service.spec.ts
@@ -19,7 +19,7 @@ vi.mock('$lib/api/backend.api', () => ({
 describe('ext.services', () => {
 	const expectedCustomTokens = [
 		{
-			certified: false,
+			certified: true,
 			data: {
 				version: 1n,
 				enabled: true,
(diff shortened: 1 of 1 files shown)


src/frontend/src/icp/stores/ext-custom-tokens.store.ts
```diff
@@ -0,0 +1,4 @@
+import type { ExtToken } from '$icp/types/ext-token';
+import { initCertifiedUserTokensStore } from '$lib/stores/user-tokens.store';
+
+export const extCustomTokensStore = initCertifiedUserTokensStore<ExtToken>();
```
AntonioVentilii:
```suggestion
import { initCertifiedCustomTokensStore } from '$lib/stores/custom-tokens.store';

export const extCustomTokensStore = initCertifiedCustomTokensStore<ExtToken>();
```

PR #10529, opened 2025-11-25
feat(build): Remove nftGeek from sources of EXT canister collections

AntonioVentilii:
# Motivation

We noticed that the list extracted from nftGeek not necessarily contains only "active" collections. Quite a lot of them are inactive canisters (for example out of cycles).

So, we can simply remove it from our curated list for now.


scripts/build.tokens.ext.ts
AntonioVentilii:
```suggestion



```

PR #10547, opened 2025-11-26
feat(frontend): Move GLDT stake context to top level

AntonioVentilii:
# Motivation

We will need the staking data at top level of our app and not only inside the Earn page. So, we move `GldtStakeContext` (and its initialization) to the `Loaders` components.

### DIFF
--- a/src/frontend/src/lib/derived/earning.derived.ts
+++ b/src/frontend/src/lib/derived/earning.derived.ts
@@ -58,15 +58,6 @@ export const earningData: Readable<EarningData> = derived(
 							$exchanges
 						})
 					: undefined,
-				[EarningCardFields.CLAIMABLE_REWARDS]: nonNullish(gldtToken)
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/earning/StakeContext.svelte
```diff
@@ -0,0 +1,23 @@
...
+
+	let { children }: Props = $props();
+
+	setContext<GldtStakeContextType>(GLDT_STAKE_CONTEXT_KEY, {
+		store: gldtStakeStore
+	});
+</script>
+
+<GldtStakeContext>
+	{@render children()}
```
DenysKarmazynDFINITY, line 22: Let's add a tiny test for children snippet, just to keep up with the test coverage number.

DenysKarmazynDFINITY (approved):

PR #10563, opened 2025-11-26
feat(frontend): Service to update NFT media consent

AntonioVentilii:
# Motivation

As suggested by @roman-nazaruk , we create a separate service to upload the NFT media consent, and centralize the logic.

### DIFF
--- a/src/frontend/src/lib/services/nft.services.ts
+++ b/src/frontend/src/lib/services/nft.services.ts
@@ -171,7 +171,7 @@ export const updateNftMediaConsent = async ({
 }): Promise<NonFungibleToken | undefined> => {
 	const saveToken = {
 		...token,
-		enabled: true,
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/services/nft.services.ts
```diff
@@ -157,3 +157,25 @@ export const updateNftSection = async ({
...
+	$ethAddress
+}: {
+	allowMedia: boolean;
+	$authIdentity: OptionIdentity;
+	token: NonFungibleToken;
+	$ethAddress: OptionEthAddress;
+}): Promise<NonFungibleToken | undefined> => {
+	const saveToken = {
+		...token,
+		enabled: true,
```
AntonioVentilii:
```suggestion
		enabled: true, // must be true otherwise we couldn't see it at this point
```

PR #10589, opened 2025-11-27
feat(frontend): New tab for the Earning route

AntonioVentilii:
# Motivation

Together with the Tokens and NFTs lists, we want to show the list of staking positions of the user. So, first we create a tab for that:

Screenshot 2025-11-27 at 16 46 05


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #10601, opened 2025-11-27
feat(frontend): isGoldaoToken util

daviddecentage:
# Motivation

Were adding a small util function to check if a token is GOLDAO.

# Changes

Added goldao leder id constant
Added function

# Tests

Added tests

### DIFF
--- a/src/frontend/src/icp-eth/utils/token.utils.ts
+++ b/src/frontend/src/icp-eth/utils/token.utils.ts
@@ -1,9 +1,9 @@
 import {
 	GLDT_LEDGER_CANISTER_ID,
-	GOLDAO_LEDGER_CANISTER_ID,
 	VCHF_LEDGER_CANISTER_ID,
 	VEUR_LEDGER_CANISTER_ID
 } from '$env/networks/networks.icrc.env';
+import { GOLDAO_LEDGER_CANISTER_ID } from '$env/tokens/tokens.sns.env';
 import { isTokenIcrc } from '$icp/utils/icrc.utils';
 import type { Token } from '$lib/types/token';
 


src/frontend/src/icp-eth/utils/token.utils.ts
```diff
@@ -7,6 +7,9 @@ export const isGLDTToken = (token: Token): boolean =>
 	token.symbol === 'GLDT' &&
 	token.ledgerCanisterId === GLDT_LEDGER_CANISTER_ID;
 
+export const isGoldaoToken = (token: Token): boolean =>
+	isTokenIcrc(token) && token.symbol === 'GOLDAO';
```
AntonioVentilii: better to compare it with canister ID too, for example like we do for GLDT.

I see that we do with symbol in the other utils, but that is not correct. I will change those in another PR.

It is safer to compare canister IDs for ICRC

AntonioVentilii: For example:

- https://github.com/dfinity/oisy-wallet/pull/10606

daviddecentage: updated

src/frontend/src/env/networks/networks.icrc.env.ts
```diff
@@ -443,6 +443,9 @@ export const VCHF_LEDGER_CANISTER_ID: LedgerCanisterIdText =
 export const VEUR_LEDGER_CANISTER_ID: LedgerCanisterIdText =
 	ADDITIONAL_ICRC_PRODUCTION_DATA.VEUR?.ledgerCanisterId ?? 'wu6g4-6qaaa-aaaan-qmrza-cai';
 
+export const GOLDAO_LEDGER_CANISTER_ID: LedgerCanisterIdText =
+	ADDITIONAL_ICRC_PRODUCTION_DATA.GOLDAO?.ledgerCanisterId ?? 'tyyy3-4aaaa-aaaaq-aab7a-cai';
```
AntonioVentilii: I think GOLDAO is in the SNS json and not in `ADDITIONAL_ICRC_PRODUCTION_DATA`

daviddecentage: updated

PR #10605, opened 2025-11-27
feat(frontend): Get GLDT staking transactions util

daviddecentage:
# Motivation

We need to display all Gldt and Goldao transactions on the staking page. For that we create a utils function to get a list of all those.

# Changes

- Added new type `StakingTransactionsUiWithToken`
- Added utils function

# Tests

Added tests

### DIFF
--- a/src/frontend/src/env/tokens/tokens-evm/tokens-polygon/tokens-erc20/tokens.zchf.env.ts
+++ b/src/frontend/src/env/tokens/tokens-evm/tokens-polygon/tokens-erc20/tokens.zchf.env.ts
@@ -1,0 +1,26 @@
+import { POLYGON_MAINNET_NETWORK } from '$env/networks/networks-evm/networks.evm.polygon.env';
+import { ZCHF_TOKEN_GROUP } from '$env/tokens/groups/groups.zchf.env';
+import zchf from '$eth/assets/zchf.webp';
+import type { RequiredEvmBep20Token } from '$evm/types/bep20';
+import type { TokenId } from '$lib/types/token';
+import { parseTokenId } from '$lib/validation/token.validation';
+
+export const ZCHF_DECIMALS = 18;
+
+export const ZCHF_SYMBOL = 'ZCHF';
+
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/utils/stake.utils.ts
```diff
@@ -0,0 +1,77 @@
...
+	ckBtcPendingUtxoTransactions: CertifiedTransaction<IcTransactionUi>[];
+	ckEthPendingTransactions: CertifiedTransaction<IcTransactionUi>[];
+	icExtendedTransactions: CertifiedTransaction<IcTransactionUi>[];
+	icTransactionsStore: IcCertifiedTransactionsData;
+	btcStatusesStore: CertifiedStoreData<BtcStatusesData>;
+	ckBtcMinterInfoStore: CertifiedStoreData<CkBtcMinterInfoData>;
+	ckBtcPendingUtxosStore: CertifiedStoreData<CkBtcPendingUtxosData>;
+	icPendingTransactionsStore: CertifiedStoreData<IcPendingTransactionsData>;
+}): StakingTransactionsUiWithToken[] => {
+	const gldtTxs = getAllIcTransactions({
```
DenysKarmazynDFINITY: I'm not sure if it makes sense to use `getAllIcTransactions` - we know for sure that neither gldt nor golddao token txs will be inside `ckBtcPendingUtxoTransactions` or `ckBtcPendingUtxosStore` or all other unrelated stores. Also, this method uses `getIcExtendedTransactions` under the hood which checks if a token is ckBTC and if yes it tries to extend it with additional data which also seems to be irrelevant to our current case. The only stores where we can potentially find the needed items are `icTransactionsStore` and `icPendingTransactionsStore`. My suggestion we manually loop through them and try to find both tokens at the same time, which will also save us 1 additional loop.

daviddecentage: Makes sense. I updated it to only take data from the two relevant stores

PR #10611, opened 2025-11-28
feat(frontend): Full date prop for transactions

daviddecentage:
# Motivation

On the staking page, we would like to display the full date for transactions.

# Changes

Added prop to display the full date for a transaction

# Tests

Added test


src/frontend/src/lib/components/transactions/Transaction.svelte
```diff
@@ -59,7 +60,8 @@
 		tokenId,
 		children,
 		onClick,
-		approveSpender
+		approveSpender,
+		fullDate = false
```
DenysKarmazynDFINITY: just an idea: let's use the same name as the util - `timeOnly`

daviddecentage: makes sense, updated

PR #10616, opened 2025-11-28
feat(frontend): update gldt-stake.store

DenysKarmazynDFINITY:
# Motivation

We can now update gldt-stake.store with the config value.


AntonioVentilii (approved):
LGTM tks!

PR #10624, opened 2025-11-28
feat(frontend): Stake transaction component

daviddecentage:
# Motivation

We need a generic Staking transaction component that works with AnyTransactionUiWithToken type, that is specialised for staking transactions (is reward, staking, unstakin label, fee on top of value for unstake etc).

# Changes

Added component

# Tests

Added tests

### DIFF
--- a/src/frontend/src/lib/components/stake/StakeTransaction.svelte
+++ b/src/frontend/src/lib/components/stake/StakeTransaction.svelte
@@ -47,11 +47,9 @@
 						('fee' in transaction && nonNullish(transaction.fee) ? transaction.fee : ZERO)) *
 					-1n
 				);
-			} 
-				return transaction.value;
-			
+			}
+			return transaction.value;
 		}
-		
 	};
 
 	const getLabel = (transaction: StakingTransactionsUiWithToken) =>


src/frontend/src/lib/components/stake/StakeTransaction.svelte
```diff
@@ -0,0 +1,73 @@
...
+					: 'unconfirmed';
+		}
+		return 'unconfirmed';
+	};
+
+	const getDisplayAmount = (transaction: StakingTransactionsUiWithToken) =>
+		'incoming' in transaction &&
+		transaction.incoming &&
+		!transaction.isReward &&
+		nonNullish(transaction.value)
```
DenysKarmazynDFINITY: can you please use `if`s in this function to increase the readability + reuse `nonNullish(transaction.value)` check (it is there 2 times with the current approach)

src/frontend/src/lib/components/stake/StakeTransaction.svelte
```diff
@@ -0,0 +1,73 @@
...
+	}
+	const { transaction, onClick }: Props = $props();
+
+	const getTimestamp = (timestampNanoseconds?: bigint | number) =>
+		nonNullish(timestampNanoseconds)
+			? Number(BigInt(timestampNanoseconds) / NANO_SECONDS_IN_SECOND)
+			: undefined;
+
+	const getToAddress = (transaction: StakingTransactionsUiWithToken): string | undefined =>
+		nonNullish(transaction.to) && transaction.to.length > 0
```
DenysKarmazynDFINITY: let's make it `nonNullish(transaction.to) ? ... : undefined`

src/frontend/src/lib/components/stake/StakeTransaction.svelte
daviddecentage: updated

src/frontend/src/lib/components/stake/StakeTransaction.svelte
daviddecentage: upadted

PR #10649, opened 2025-12-01
feat(frontend): added new createPaymentMethodDataMap

roman-nazaruk:
# Motivation
We need to parse and build the supported networks from the OpenCryptoPay response.

# Changes
Added a utility method createPaymentMethodDataMap to normalize and construct the supported network list.

# Tests
Added tests for the new utility function.

### DIFF
--- a/src/frontend/src/lib/stores/modal-tokens-list.store.ts
+++ b/src/frontend/src/lib/stores/modal-tokens-list.store.ts
@@ -128,7 +128,18 @@ export const initModalTokensListContext = (
 			update((state) => ({
 				...state,
 				filterNetworksIds: networksIds
-			}))
+			})),
+		resetFilters: () => {
+			update((state) => ({
+				...state,
+				filterQuery: undefined,
+				filterNetwork: undefined,
+				filterZeroBalance: undefined,
+				sortByBalance: undefined,
+				filterNetworksIds: undefined,
+				filterNfts: undefined
+			}));
+		}
 	};
 };
 
@@ -140,6 +151,7 @@ export interface ModalTokensListContext {
 	setFilterQuery: (query: string) => void;
 	setFilterNetwork: (network: Network | undefined) => void;
 	setFilterNetworksIds: (networksIds: NetworkId[] | undefined) => void;
+	resetFilters: () => void;
 }
 
 export const MODAL_TOKENS_LIST_CONTEXT_KEY = Symbol('modal-tokens-list');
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/utils/open-crypto-pay.utils.ts
```diff
@@ -44,3 +51,59 @@ export const formatAddress = (address?: Address): string => {
...
+		if (nonNullish(pay?.openCryptoPay)) {
+			acc.add(pay.openCryptoPay);
+		}
+		return acc;
+	}, new Set());
+
+	return transferAmounts.filter(
+		(transfer) =>
+			transfer.available && transfer.assets.length > 0 && supportedMethods.has(transfer.method)
+	);
```
AntonioVentilii, line 78: this can be merged in a single loop?

src/frontend/src/lib/utils/open-crypto-pay.utils.ts
```diff
@@ -44,3 +51,59 @@ export const formatAddress = (address?: Address): string => {
...
+const createAssetsMap = (assets: RequestedAmount[]): PaymentAssetsData => {
+	const assetsMap = new Map();
+
+	assets.forEach(({ asset, amount }) => {
+		assetsMap.set(asset, { amount });
+	});
+
+	return assetsMap;
+};
+
```
AntonioVentilii: this loop can be avoided, right? or merged inside another?

AntonioVentilii (commented):
In general, I think that maybe there are too many loops in these functions. may you please have a look if it can be simplified?

src/frontend/src/lib/utils/open-crypto-pay.utils.ts
```diff
@@ -44,3 +45,31 @@ export const formatAddress = (address?: Address): string => {
...
+	networks: Network[];
+}): Map<string, PaymentMethodData> => {
+	const supportedMethods = networks.reduce<Set<string>>((acc, { pay }) => {
+		if (nonNullish(pay?.openCryptoPay)) {
+			acc.add(pay.openCryptoPay);
+		}
+		return acc;
+	}, new Set());
+
+	return transferAmounts.reduce<Map<string, PaymentMethodData>>(
```
AntonioVentilii, line 67: here you could put a `supportedMethods.length>0 ? .... : new Map()` right?

roman-nazaruk: Added condition if (supportedMethods.size === 0) {
		return new Map();
	}

PR #10656, opened 2025-12-02
feat(frontend): Go-to button in Earning tab

AntonioVentilii:
# Motivation

We create a new button in the `Earning` tab of the `Assets` page, that redirects to the `Earn` page.

Screenshot 2025-12-02 at 09 58 06


src/frontend/src/lib/components/earning/GoToEarnButton.svelte
```diff
@@ -0,0 +1,34 @@
...
+		networkUrl({
+			path: AppPath.Earn,
+			networkId: $networkId,
+			usePreviousRoute: true,
+			fromRoute
+		})
+	);
+</script>
+
+<button
```
DenysKarmazynDFINITY: Would the `Button` component with `colorStyle="success"` work here?

AntonioVentilii: Good catch! Yes, done!! It is the same:

Screenshot 2025-12-02 at 14 27 31

PR #10657, opened 2025-12-02
feat(frontend): added new util mapTokenToPayable

roman-nazaruk:
# Motivation

We need to map Tokens and make them as Payable tokens

# Changes

Added new util mapTokenToPayableToken 

# Tests

Covered new logic with tests

### DIFF
--- a/src/frontend/src/lib/utils/open-crypto-pay.utils.ts
+++ b/src/frontend/src/lib/utils/open-crypto-pay.utils.ts
@@ -90,7 +90,7 @@ export const mapTokenToPayableToken = ({
 }: {
 	token: Token;
 	methodDataMap: Map<string, PaymentMethodData>;
-}): Partial<PayableToken> | undefined => {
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/utils/open-crypto-pay.utils.ts
```diff
@@ -77,3 +83,36 @@ export const createPaymentMethodDataMap = ({
...
+		return;
+	}
+
+	const methodData = methodDataMap.get(tokenNetwork);
+
+	if (isNullish(methodData)) {
+		return;
+	}
+
+	const assetData = methodData.assets.get(token.symbol);
```
AntonioVentilii, line 108: please add an explanation on why we cannot use token ID or a unique token identifier (like address, ledger, etc)

roman-nazaruk: Done.

PR #10662, opened 2025-12-02
feat(frontend): Show earning potential in the staking placeholder

AntonioVentilii:
# Motivation

We want to show the potential earning in the staking placeholder, or the APY if the amount is not that relevant.

Screenshot 2025-12-02 at 11 42 20
Screenshot 2025-12-02 at 11 43 17

### DIFF
--- a/src/frontend/src/lib/components/stake/NoStakePlaceholder.svelte
+++ b/src/frontend/src/lib/components/stake/NoStakePlaceholder.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  	import { notEmptyString } from '@dfinity/utils';
+	import { notEmptyString } from '@dfinity/utils';
 	import noEarningBanner from '$lib/assets/earning/no-earning-banner.svg';
 	import Img from '$lib/components/ui/Img.svelte';
 	import { currentCurrency } from '$lib/derived/currency.derived';


src/frontend/src/lib/components/stake/NoStakePlaceholder.svelte
```diff
@@ -1,21 +1,50 @@
...
+	const usdAmountThreshold = 20;
 </script>
 
 <div class="flex flex-col items-center gap-5 px-6 py-10">
 	<!-- TODO: use the placeholder image for the staking or rename the current one to be more generic -->
 	<Img alt={$i18n.stake.alt.placeholder_image} src={noNftsBanner} />
 
 	<div class="flex flex-col items-center gap-2">
-		<h5>{$i18n.stake.text.title_empty}</h5>
+		{#if notEmptyString($i18n.stake.text.title_empty_1)}
```
DenysKarmazynDFINITY, line 25: if the translation key is not there, does an empty h5 introduce some unwanted margins? just wondering if we can skip this check or not

src/frontend/src/lib/components/stake/NoStakePlaceholder.svelte
```diff
@@ -1,21 +1,50 @@
...
 	import { i18n } from '$lib/stores/i18n.store';
+	import { formatCurrency } from '$lib/utils/format.utils';
 
 	interface Props {
 		hideDescription?: boolean;
 	}
 
 	let { hideDescription = false }: Props = $props();
+
+	const usdAmountThreshold = 20;
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: I'm fine with keeping it here, but let's maybe update the naming to highlight that it's a const - `USD_AMOUNT_THRESHOLD`.

src/frontend/src/lib/components/stake/NoStakePlaceholder.svelte
AntonioVentilii: the difference is minimal, but yes, it renders a bit of margin:

### Without check
Screenshot 2025-12-02 at 15 10 56

### With check
Screenshot 2025-12-02 at 15 11 07

PR #10667, opened 2025-12-02
feat(frontend): added new getEthFeeDataWithProvider service method

roman-nazaruk:
# Motivation

We need to extract { feeData, providers, params } for ETH tokens. This PR introduces a new service to centralize and reuse that logic for calculation EthFee for payment, swap, send

# Changes

Added a new getEthFeeDataWithProvider service to extract feeData, providers, and params.

# Tests

Added test coverage for the new service.

### DIFF
--- a/src/frontend/src/tests/eth/services/fee.services.spec.ts
+++ b/src/frontend/src/tests/eth/services/fee.services.spec.ts
@@ -58,20 +58,6 @@ describe('eth-fee-data.services', () => {
 			});
 		});
 
-		it('should use from address as to when to is empty string', async () => {
-			const result = await getEthFeeDataWithProvider({
-				networkId: network.id,
-				chainId: network.chainId,
-				from: fromAddr,
-				to: ''
-			});
-
-			expect(result.params).toEqual({
-				from: fromAddr,
-				to: fromAddr
-			});
-		});
-
 		it('should use from address as to when to is not provided', async () => {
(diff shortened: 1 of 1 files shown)


src/frontend/src/eth/services/fee.services.ts
```diff
@@ -98,3 +102,48 @@ export const getCkErc20FeeData = async ({
...
+export const getEthFeeDataWithProvider = async ({
+	networkId,
+	chainId,
+	from,
+	to = ''
+}: {
+	networkId: NetworkId;
+	chainId: bigint;
+	from: EthAddress;
+	to?: EthAddress;
```
AntonioVentilii: better to pass `to` as non-optional

roman-nazaruk: Done

src/frontend/src/eth/services/fee.services.ts
```diff
@@ -98,3 +103,44 @@ export const getCkErc20FeeData = async ({
...
+	from,
+	to
+}: {
+	networkId: NetworkId;
+	chainId: bigint;
+	from: EthAddress;
+	to: EthAddress;
+}): Promise<{
+	feeData: Omit<TransactionFeeData, 'gas'>;
+	providers: InfuraProvider;
```
AntonioVentilii: this is actually a single provider, we can rename it just `provider`

PR #10673, opened 2025-12-03
style(frontend): Show OISY NFT ID before falling back to normal NFT ID

AntonioVentilii:
# Motivation

In `NftCard`, we want to first show the OISY NFT ID, and then fallback to the normal ID.

### Before

Screenshot 2025-12-03 at 09 32 32

### After

Screenshot 2025-12-03 at 09 32 25

### DIFF
--- a/src/frontend/src/tests/lib/components/nfts/NftCard.spec.ts
+++ b/src/frontend/src/tests/lib/components/nfts/NftCard.spec.ts
@@ -253,7 +253,7 @@ describe('NftCard', () => {
 		expect(getByText(`#${mockValidErc721Nft.id} – ${mockValidErc721Nft.name}`)).toBeInTheDocument();
 	});
 
-	it('should render first hte OISY NFT ID', () => {
+	it('should render first the OISY NFT ID', () => {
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/nfts/NftCard.svelte
```diff
@@ -136,7 +136,7 @@
 			{withCollectionLabel ? nft.collection.name : nft.name}
 		</span>
 		<span class="truncate text-xs" class:text-tertiary={!disabled}>
-			#{nft.id}
+			#{nft.oisyId ?? nft.id}
```
DenysKarmazynDFINITY, line 139: maybe we can later create a util similar to `getTokenDisplaySymbol`, since we already have such a line in a few places

AntonioVentilii: yes!!! I will do it after this one

AntonioVentilii: Done in:

- https://github.com/dfinity/oisy-wallet/pull/10696

PR #10678, opened 2025-12-03
feat(frontend): Add logic in `EarningsList` to prepare for `EarningCard`

AntonioVentilii:
# Motivation

We are going to show the earning positions per provider in the Earning tab of Assets page:

Screenshot 2025-12-03 at 10 40 33

However, we still don't have a complete interface on how to structure the providers and their earning features, since we ha ve a single provider and a single features for now.

Since it is difficult to foresee the exact needs for multiple providers, for now we create an base interface for the providers and expand it on needs. It is quite similar to what we do for `Token` type.

In this PR, we expand the `EarningsList` component to start building such a list of providers.

# Changes

- Create new type `ProviderUi`.
- Change `EarningsList` to build the list of `ProviderUi`, for now with no real data apart from the stake-provider configs.
- Adapt the rendering to show the list instead of the placeholder.

# Tests

Added some tests, but no change for now.

### DIFF
--- a/src/frontend/src/lib/components/earning/EarningsList.svelte
+++ b/src/frontend/src/lib/components/earning/EarningsList.svelte
@@ -35,7 +35,7 @@
 		</div>
 	{/each}
 
-	{#if filteredEarningProviders?.length === 0}
+	{#if filteredEarningProviders.length === 0}
 		<NoStakePlaceholder />
 	{/if}
 </div>


src/frontend/src/lib/types/provider-ui.ts
```diff
@@ -0,0 +1,5 @@
+import type { StakeProviderConfig } from '$lib/types/stake';
+
+export type ProviderUi<T extends StakeProviderConfig = StakeProviderConfig> = T & {
```
DenysKarmazynDFINITY, line 3: How about making it a bit more specific like `StakeProviderUi` or `EarningProviderUi`?

AntonioVentilii: Good point! I was actually going to exams it after: basically staking will be a subset of earning, that will be a subset of providers eheh

AntonioVentilii: Good point! I was actually going to exams it after: basically staking will be a subset of earning, that will be a subset of providers eheh

PR #10680, opened 2025-12-03
feat(frontend): Add base `EarningCard` for Earning tab

AntonioVentilii:
# Motivation

We create a base EarningCard component to show the logo and card title of the Earning provider.

Below an idea mocking a position:

Screenshot 2025-12-03 at 11 35 06

### DIFF
--- a/src/frontend/src/lib/components/earning/EarningCard.svelte
+++ b/src/frontend/src/lib/components/earning/EarningCard.svelte
@@ -15,7 +15,7 @@
 </script>
 
 <div class="flex w-full flex-col">
-	<LogoButton  rounded={false}>
+	<LogoButton rounded={false}>
 		{#snippet logo()}
 			<span class="mr-2 flex">
 				<Logo


src/frontend/src/lib/components/earning/EarningCard.svelte
```diff
@@ -0,0 +1,36 @@
...
+		provider: ProviderUi;
+	}
+
+	let { provider }: Props = $props();
+
+	let { logo: logoSrc, name, cardTitle } = $derived(provider);
+</script>
+
+<div class="flex w-full flex-col">
+	<LogoButton condensed={false} dividers={false} hover={true} rounded={false}>
```
DenysKarmazynDFINITY: You can only keep `rounded={false}`, the rest are component's default values

AntonioVentilii: good catch!

PR #10684, opened 2025-12-03
feat(frontend): added new ExcahngeTokenToPay.svelte

roman-nazaruk:
# Motivation

We aim to show new ExcahngeTokenToPay exchange rate for payment modal

# Changes

Added new ExchangeTokenToPay.svelte component

# Tests

Covered with tests


AntonioVentilii (commented):
As discussed offline, we will check if we can reuse the existing component

PR #10685, opened 2025-12-03
feat(frontend): Map more data for `ProviderUi` in `EarningsList`

AntonioVentilii:
# Motivation

We add a few data for ProviderUi:

- max APY among all earning features;
- total earnings per year, being the sum of all the earning features;
- total position (in USD) of all the money put in earning with the provider;
- list of tokens involved in the earnings.

However, since we have only Gold DAO and staking, for now the sum is a single value...

### DIFF
--- a/src/frontend/src/tests/mocks/providers-ui.mock.ts
+++ b/src/frontend/src/tests/mocks/providers-ui.mock.ts
@@ -1,8 +1,28 @@
+import { ICP_NETWORK } from '$env/networks/networks.icp.env';
+import { GLDT_LEDGER_CANISTER_ID } from '$env/networks/networks.icrc.env';
+import type { IcToken } from '$icp/types/ic-token';
 import { stakeProvidersConfig } from '$lib/config/stake.config';
 import type { ProviderUi } from '$lib/types/provider-ui';
 import { StakeProvider } from '$lib/types/stake';
+import { parseTokenId } from '$lib/validation/token.validation';
+
+const mockGldtToken: IcToken = {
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/earning/EarningsList.svelte
```diff
@@ -1,19 +1,46 @@
...
 
 	// TODO: this logic is very custom-made for the current single provider. Improve it to be more generic: each provider has a feature (for example earning), and each feature a sub-group (for example staking), and each sub-group has a list of objects to which it applies (like tokens).
 	let earningProvidersUi = $derived(
 		Object.entries(stakeProvidersConfig).reduce<ProviderUi[]>((acc, [providerName, config]) => {
 			if (providerName === StakeProvider.GLDT) {
+				const gldtToken = $enabledFungibleTokens.find(isGLDTToken);
+
+				const gldtStakePosition = $gldtStakeStore?.position?.staked ?? ZERO;
+
+				const gldtStakePositionUsd = nonNullish(gldtToken)
```
DenysKarmazynDFINITY, line 24: Just as an idea - did you consider reusing the store David created for the Earn page? Some of the data you calculate here might be already available there.

AntonioVentilii: Good one! I tried but apparently it goes in loop... i am not really sure why, but there is some linked derived store in that case, even if i cannot find it. However, I will try and fix it for sure, but unfortunately it requires quite a bit of refactoring

PR #10686, opened 2025-12-03
feat(frontend): Display more Earning Provider data in the card

AntonioVentilii:
# Motivation

We want to display more data in `EarningCard`.

Screenshot 2025-12-03 at 10 40 33

### DIFF
--- a/src/frontend/src/lib/components/earning/EarningCard.svelte
+++ b/src/frontend/src/lib/components/earning/EarningCard.svelte
@@ -30,7 +30,9 @@
 
 	let testId = $derived(`${EARNING_CARD}-${name}`);
 
-	let tokenSymbols = $derived(new Set(provider.tokens.map((token) => getTokenDisplaySymbol(token))));
+	let tokenSymbols = $derived(
+		new Set(provider.tokens.map((token) => getTokenDisplaySymbol(token)))
+	);
 
 	let networkNames = $derived(new Set(provider.tokens.map(({ network: { name } }) => name)));
 


src/frontend/src/lib/components/earning/EarningCard.svelte
```diff
@@ -12,9 +18,24 @@
...
+		cardTitle,
+		maxApy,
+		totalEarningPerYear,
+		totalPositionUsd
+	} = $derived(provider);
 
 	let testId = $derived(`${EARNING_CARD}-${name}`);
+
+	let tokenSymbols = $derived(
+		new Set(provider.tokens.map(({ oisySymbol, symbol }) => oisySymbol?.oisySymbol ?? symbol))
```
DenysKarmazynDFINITY: let's use `getTokenDisplaySymbol` here

src/frontend/src/lib/components/earning/EarningCard.svelte
```diff
@@ -12,9 +18,24 @@
...
+	);
+
+	let networkNames = $derived(new Set(provider.tokens.map(({ network: { name } }) => name)));
+
+	let descriptionElements = $derived([...tokenSymbols, ...networkNames]);
```
DenysKarmazynDFINITY, line 39: just wondering - do we want to kind of mix token symbols and networks? I would probably expect it to be somehow paired

src/frontend/src/lib/components/earning/EarningCard.svelte
AntonioVentilii: nice! i forgot about it! done!

src/frontend/src/lib/components/earning/EarningCard.svelte
AntonioVentilii: good point... to be honest not sure, since we don't have any other example of multiple tokens/networks. What do you suggest? Shall i separate the rendering?

(i was thinking that it would be something like `GLDT * BONK * Internet Computer * Solana`)

DenysKarmazynDFINITY: I guess we can postpone this decision until we have a real use-case for this question, let's keep as it is for now

PR #10688, opened 2025-12-03
feat(frontend): added new util findNativeToken

roman-nazaruk:
# Motivation

We need to have a util method to get native token for each network

# Changes

Added new findNativeToken util

# Tests

Covered new util with tests

### DIFF
--- a/src/frontend/src/lib/utils/token.utils.ts
+++ b/src/frontend/src/lib/utils/token.utils.ts
@@ -12,6 +12,7 @@ import type { CertifiedStoreData } from '$lib/stores/certified.store';
 import type { OptionBalance } from '$lib/types/balance';
 import type { CanisterIdText } from '$lib/types/canister';
 import type { ExchangesData } from '$lib/types/exchange';
+import type { Network } from '$lib/types/network';
 import type { StakeBalances } from '$lib/types/stake-balance';
 import type { RequiredTokenWithLinkedData, Token, TokenStandard } from '$lib/types/token';
 import type { CardData } from '$lib/types/token-card';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -269,3 +270,11 @@ export const isTokenToggleable = <T extends Token>(token: T): token is TokenTogg
...
 export const filterEnabledToken = <T extends Token>(token: T): boolean =>
 	isTokenToggleable(token) ? token.enabled : true;
+
+export const findNativeToken = ({
+	networkId,
+	enabledTokens
+}: {
+	networkId: Network['id'];
+	enabledTokens: Token[];
+}): Token | undefined => enabledTokens.find(({ network: { id } }) => id === networkId);
```
AntonioVentilii, line 280: i think this will just find the first token with the same network

src/frontend/src/tests/lib/utils/tokens.utils.spec.ts
```diff
@@ -1083,4 +1084,71 @@ describe('tokens.utils', () => {
...
+				networkId: ETHEREUM_NETWORK_ID,
+				enabledTokens: []
+			});
+
+			expect(result).toBeUndefined();
+		});
+
+		it('should return first matching token when multiple tokens have same network', () => {
+			const mockUsdcToken = { ...USDC_TOKEN, network: ETHEREUM_TOKEN.network };
+			const enabledTokens = [ETHEREUM_TOKEN, mockUsdcToken];
```
AntonioVentilii, line 1122: add another test where the position of these two are inverted pls

`const enabledTokens = [mockUsdcToken, ETHEREUM_TOKEN];`

PR #10691, opened 2025-12-03
feat(frontend): added new enrichEthEvmToken utils for balances and usd values

roman-nazaruk:
# Motivation

We need to have util method that will return eth/evm token with sufficient balance and currentExchangeRate

# Changes

Added new enrichEthEvmToken util

# Tests

Covered logic with tests

### DIFF
--- a/src/frontend/src/lib/components/earning/EarningCard.svelte
+++ b/src/frontend/src/lib/components/earning/EarningCard.svelte
@@ -1,10 +1,17 @@
 <script lang="ts">
+	import Divider from '$lib/components/common/Divider.svelte';
 	import Logo from '$lib/components/ui/Logo.svelte';
 	import LogoButton from '$lib/components/ui/LogoButton.svelte';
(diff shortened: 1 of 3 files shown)


src/frontend/src/eth/utils/token.utils.ts
```diff
@@ -16,3 +30,132 @@ export const isNotSupportedErc20TwinTokenId = (tokenId: TokenId): boolean =>
 // ERC20 token addresses are zero-padded to 32 bytes (64 characters) string in hex format, when they need to be ABI-encoded.
 export const tokenAddressToHex = (address: string): string =>
 	zeroPadValue(address.toLowerCase(), 32);
+
+export const hasSufficientBalance = ({
```
AntonioVentilii, line 36: in the interest of time: please add a TODO for each of these methods saying that these functions may be duplicated of code that we have elsewhere

roman-nazaruk: Done

PR #10700, opened 2025-12-03
fix(frontend): Expand NFT images for SVG without size

AntonioVentilii:
# Motivation

Some NFTs are loaded from SVG websites but without given size, so they shrink and do not appear.

### Before

Screenshot 2025-12-03 at 17 47 14

### After

Screenshot 2025-12-03 at 17 47 50

### No change in the rest

Screenshot 2025-12-03 at 17 48 02
Screenshot 2025-12-03 at 17 48 12

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftHero.svelte
+++ b/src/frontend/src/lib/components/nfts/NftHero.svelte
@@ -110,7 +110,7 @@
 									data: nft
 								})}
 						>
-							<Img src={nft.imageUrl} styleClass="max-h-full max-w-full" />
+							<Img src={nft.imageUrl} styleClass="max-h-full max-w-full w-full" />
 						</button>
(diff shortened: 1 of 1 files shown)


DenysKarmazynDFINITY (approved):

src/frontend/src/lib/components/nfts/NftHero.svelte
```diff
@@ -110,7 +110,7 @@
 									data: nft
 								})}
 						>
-							<Img src={nft.imageUrl} styleClass="max-h-full max-w-full" />
+							<Img src={nft.imageUrl} styleClass="max-h-full max-w-full w-full" />
```
DenysKarmazynDFINITY, line 113: looks like max-w-full can now be removed

PR #10712, opened 2025-12-04
feat(frontend): added new TokenFeeValue component

roman-nazaruk:
# Motivation

We need to show proper exchangeRate for TokenFee for payment

# Changes

Added new component ExchangeTokenFee

# Tests

Covered with tests

### DIFF
--- a/src/frontend/src/lib/components/hero/Header.svelte
+++ b/src/frontend/src/lib/components/hero/Header.svelte
@@ -28,6 +28,10 @@
 	let menuOpen = $state(false);
 
 	let nftsCollectionRoute = $derived(isRouteNfts(page) && nonNullish($routeCollection));
+
+	let modalsOpen = $derived($modalWalletConnect || $modalUniversalScannerOpen);
+
+	let biggerOverlay = $derived(menuOpen || networkSwitcherOpen || helpMenuOpen || modalsOpen);
 </script>
 
 <header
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/tokens/TokenFeeValue.svelte
```diff
@@ -0,0 +1,22 @@
+<script lang="ts">
+	import { currentCurrency } from '$lib/derived/currency.derived';
+	import { currentLanguage } from '$lib/derived/i18n.derived';
+	import { currencyExchangeStore } from '$lib/stores/currency-exchange.store';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { formatCurrency } from '$lib/utils/format.utils';
+
+	let { feeInUSD }: { feeInUSD: number } = $props();
+	let exchangeBalance = $derived(
```
AntonioVentilii, line 10:
```suggestion
	let { feeInUSD }: { feeInUSD: number } = $props();
	
	let exchangeBalance = $derived(
```

src/frontend/src/lib/components/tokens/TokenFeeValue.svelte
```diff
@@ -0,0 +1,22 @@
+<script lang="ts">
```
AntonioVentilii, line 1: let's rename the component or move it to a more generic folder: if it is not specific for the scanner for the Pay flow, let's move it there. Or if it is a generic token component, let's move it to token folder

roman-nazaruk: Done

PR #10713, opened 2025-12-04
feat(frontend): Create Pay Dialog component

AntonioVentilii:
# Motivation

We want to add a button that opens a modal (or bottom sheet, depending on the screen size) with information about pay via OISY.

# Changes

- Create new modal opening in modalStore.
- Create `PayDialog` component: it will render a banner with information, and a button that opens the scanner.

# Tests

Screenshot 2025-12-04 at 13 25 41
Screenshot 2025-12-04 at 13 25 50


src/frontend/src/lib/components/pay/PayDialog.svelte
```diff
@@ -0,0 +1,84 @@
...
+		PAY_DIALOG,
+		PAY_DIALOG_BANNER,
+		PAY_DIALOG_PAY_BUTTON
+	} from '$lib/constants/test-ids.constants';
+	import { modalPayDialogOpen } from '$lib/derived/modal.derived';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { modalStore } from '$lib/stores/modal.store';
+	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+
+	let open = $state($modalPayDialogOpen);
```
DenysKarmazynDFINITY: any reason why we don't use the store directly? I imagine there should be a `modalStore.openPayDialog` method, and for closing we could as always use `modalStore.close`. Or am I missing something?

src/frontend/src/lib/components/pay/PayDialog.svelte
```diff
@@ -0,0 +1,84 @@
...
+			<IconScanLine />
+			{replaceOisyPlaceholders($i18n.pay.text.dialog_button)}
+		</div>
+	</Button>
+{/snippet}
+
+<Responsive up="md">
+	<Modal onClose={close} testId={PAY_DIALOG}>
+		{#snippet title()}
+			<div class="flex items-center gap-3 font-bold text-primary">
```
DenysKarmazynDFINITY: since there is just one element, I think we can remove `gap-3`, or replace it with a margin if you used it for spacing purposes

src/frontend/src/lib/components/pay/PayDialogContent.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+	import { Html } from '@dfinity/gix-components';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+</script>
+
+<div>
```
DenysKarmazynDFINITY: can we remove this div?

src/frontend/src/lib/components/pay/PayDialog.svelte
AntonioVentilii: done!

src/frontend/src/lib/components/pay/PayDialogContent.svelte
AntonioVentilii: ah right! I will re-add it when i add the hyperlink

src/frontend/src/lib/components/pay/PayDialog.svelte
AntonioVentilii: good point! i think i lost the line of coding eheh

DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #10717, opened 2025-12-04
feat(frontend): Pay button in the main header

AntonioVentilii:
# Motivation

Creating a button to open the Pay dialog.

Screenshot 2025-12-04 at 15 02 26
Screenshot 2025-12-04 at 15 02 40


src/frontend/src/lib/components/icons/IconPay.svelte
```diff
@@ -0,0 +1,26 @@
+<script lang="ts">
```
roman-nazaruk, line 2: Could you plase add source?

AntonioVentilii: Good catch!

PR #10720, opened 2025-12-04
feat(frontend): custom error message for DissolvementsLimitReached

DenysKarmazynDFINITY:
# Motivation

We want to display a more user-friendly message if gldt_stake canister throws DissolvementsLimitReached. The custom error will be shown only if all required information is available, otherwise we fallback to the one returned by the canister.


src/frontend/src/icp/components/stake/gldt/GldtUnstakeWizard.svelte
```diff
@@ -98,10 +101,24 @@
 				}
 			});
 
-			toastsError({
-				msg: { text: $i18n.send.error.unexpected },
-				err
-			});
+			if (
```
AntonioVentilii, line 104: i imagine we don't have existing tests for this?

DenysKarmazynDFINITY: added new tests for that 👍

PR #10730, opened 2025-12-04
feat(frontned): added new BottomSheetTokenList component for pay

roman-nazaruk:
# Motivation

We aim to show tokens list for payment in Bottomsheet

# Changes

Added new component for payments to show tokens list in BottomSheet

# Tests

Covered component with the tests

image

### DIFF
--- a/src/frontend/src/lib/components/scanner/open-crypto-pay/BottomSheetTokensList.svelte
+++ b/src/frontend/src/lib/components/scanner/open-crypto-pay/BottomSheetTokensList.svelte
@@ -11,7 +11,7 @@
 	let { onClose, visible = $bindable() }: Props = $props();
 </script>
 
-<BottomSheet bind:visible {onClose}>
+<BottomSheet {onClose} bind:visible>
 	{#snippet content()}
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/scanner/open-crypto-pay/BottomSheetTokensList.svelte
```diff
@@ -0,0 +1,24 @@
...
+
+	interface Props {
+		onClose: () => void;
+		visible: boolean;
+	}
+
+	let { onClose, visible = $bindable() }: Props = $props();
+</script>
+
+<div class="fixed inset-0 z-50">
```
AntonioVentilii: why this very high and very custom z-index? this can cause really issues on finding out why certain elements are on top of others, causing a lot of workaround (like we have in the Header of the tokens list)

is it necessary? isn't bottomsheet already on top of its rendered parent?

roman-nazaruk: Good catch, removed.

PR #10743, opened 2025-12-05
feat(frontend): payment modal with tokenlist steps

roman-nazaruk:
# Motivation

We aim to show tokens list wizard step for payments

# Changes

Added TokensList for payments wizard modal

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
+++ b/src/frontend/src/lib/components/nfts/NftCollectionCard.svelte
@@ -20,7 +20,7 @@
 	import type { NftCollectionUi } from '$lib/types/nft';
 	import { replacePlaceholders } from '$lib/utils/i18n.utils.js';
 	import { nftsUrl } from '$lib/utils/nav.utils';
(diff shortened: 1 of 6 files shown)


src/frontend/src/lib/components/scanner/ScannerWizard.svelte
```diff
@@ -35,7 +57,21 @@
...
+			<OpenCryptoPay
+				onSelectToken={() => goToStep(WizardStepsScanner.TOKENS_LIST)}
+				bind:isTokenSelecting
+			/>
+		{:else if currentStep?.name === WizardStepsScanner.TOKENS_LIST && !isTokenSelecting}
+			<TokensList onClose={() => goToStep(WizardStepsScanner.PAY)} />
 		{/if}
 	{/key}
+
+	{#if isTokenSelecting}
```
AntonioVentilii: this one should not be here right? it should be in the consumer of the Wizard, otherwise any other consumer will bring it together

roman-nazaruk: Right, removed

PR #10744, opened 2025-12-05
feat(frontend): add earning potential hint button

DenysKarmazynDFINITY:
# Motivation

We need to add an additional hint button to the Earn page.

Screenshot 2025-12-05 at 10 01 56

### DIFF
--- a/src/frontend/src/lib/components/earning/EarningPotentialCard.svelte
+++ b/src/frontend/src/lib/components/earning/EarningPotentialCard.svelte
@@ -1,5 +1,7 @@
 <script lang="ts">
+	import { slide } from 'svelte/transition';
 	import EarningYearlyAmount from '$lib/components/earning/EarningYearlyAmount.svelte';
+	import IconHelp from '$lib/components/icons/lucide/IconHelp.svelte';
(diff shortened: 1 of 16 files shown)


src/frontend/src/lib/components/earning/EarningPotentialCard.svelte
```diff
@@ -9,11 +11,25 @@
...
+	let infoExpanded = $state(false);
 </script>
 
 <StakeContentCard>
 	{#snippet content()}
-		<div class="text-sm font-bold">{$i18n.stake.text.earning_potential}</div>
+		<div class="flex items-center justify-center gap-0.5">
+			<div class="text-sm font-bold">{$i18n.stake.text.earning_potential}</div>
+
+			<button class="text-tertiary" onclick={() => (infoExpanded = !infoExpanded)}>
```
AntonioVentilii, line 23: Does `Buttonicon` help? Or not applicable?

DenysKarmazynDFINITY: good idea, but unfortunately it comes with some styling that is not suitable for this case, so for simplicity I'll keep it as it is

PR #10798, opened 2025-12-09
feat(frontend): added payProgress to payment flow

roman-nazaruk:
# Motivation

Added payProgress to payment flow to track progress steps when payment is loading

### DIFF
--- a/src/frontend/src/lib/components/manage/AddTokenReviewByNetwork.svelte
+++ b/src/frontend/src/lib/components/manage/AddTokenReviewByNetwork.svelte
@@ -142,7 +142,7 @@
 			}
 		}
 
-		if (ethMetadata.decimals > 0) {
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/scanner/OpenCryptoPay.svelte
```diff
@@ -55,6 +66,34 @@
...
+			await pay({
+				token: $selectedToken,
+				data: $data,
+				from: $ethAddress,
+				identity: $authIdentity,
+				progress
+			});
+		} catch (e: unknown) {
+			// TODO: add steps to redirect to Payment Failed screen and add event
+			console.warn(e);
```
AntonioVentilii: i think you can remove the console warn, since you already added a TODO

PR #10826, opened 2025-12-10
fix(frontend): fixed decoded url types

roman-nazaruk:
# Motivation

We need to change type for numeric values to bigint for decoding qr-code values


src/frontend/src/tests/lib/utils/qr-code.utils.spec.ts
```diff
@@ -179,18 +183,6 @@ describe('decodeUrn', () => {
 
 				expect(result).toBeUndefined();
 			});
-
-			it('should return undefined for empty value', () => {
-				const result = decodeQrCodeUrn(`ethereum:${validAddress}@1?value=`);
-
-				expect(result).toBeUndefined();
```
AntonioVentilii, line 189: can you show the result here instead of removing the test?

src/frontend/src/tests/lib/utils/qr-code.utils.spec.ts
```diff
@@ -179,18 +183,6 @@ describe('decodeUrn', () => {
...
-			});
-
-			it('should return undefined for value with whitespace', () => {
-				const result = decodeQrCodeUrn(`ethereum:${validAddress}@1?value=   `);
-
-				expect(result).toBeUndefined();
```
AntonioVentilii, line 195: same?

src/frontend/src/tests/lib/utils/qr-code.utils.spec.ts
```diff
@@ -16,7 +17,7 @@ vi.mock('@icp-sdk/canisters/ledger/icrc', () => ({
 describe('decodeUrn', () => {
 	const tokenList = get(tokens);
 	const destination = 'some-destination';
-	const amount = 1.23;
+	const amount = 123;
```
AntonioVentilii: let's leave it as 1.23, we should test for decimals

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -36,20 +36,17 @@ export const decodeQrCodeUrn = (urn: string): DecodedUrn | undefined => {
...
-			return { [key]: parseFloat(value) };
+			return { [key]: value };
 		}
+
 		if ((URN_STRING_PARAMS as readonly string[]).includes(key)) {
 			return { [key]: value };
 		}
-		if (!isNaN(parseFloat(value))) {
-			return { [key]: parseFloat(value) };
-		}
```
AntonioVentilii, line 49: why was this removed?

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -36,20 +36,17 @@ export const decodeQrCodeUrn = (urn: string): DecodedUrn | undefined => {
...
+
 		return { [key]: value };
 	};
 
-	const parseQueryString = (
-		qs: string
-	): { [key: string]: string | number | undefined } | undefined => {
+	const parseQueryString = (qs: string): { [key: string]: string | undefined } | undefined => {
```
AntonioVentilii: why was the `number` removed?

src/frontend/src/lib/types/qr-code.ts
```diff
@@ -21,18 +21,24 @@ const DecodedUrnBaseSchema = z.object({
...
+		[param]: z.string().optional()
+	}),
 	{} as Record<(typeof URN_STRING_PARAMS)[number], z.ZodOptional<z.ZodString>>
 );
 
 export const DecodedUrnSchema = DecodedUrnBaseSchema.extend({
 	...NumericParamsSchema,
 	...StringParamsSchema
-}).catchall(z.union([z.string(), z.number()]).optional());
+}).catchall(z.union([z.string(), z.number(), z.bigint()]).optional());
```
AntonioVentilii, line 46: can `z.number()` be removed here?

src/frontend/src/lib/types/qr-code.ts
```diff
@@ -21,18 +21,24 @@ const DecodedUrnBaseSchema = z.object({
...
+	}),
+	{} as Record<(typeof URN_NUMERIC_PARAMS)[number], z.ZodOptional<z.ZodBigInt>>
 );
 
 const StringParamsSchema = URN_STRING_PARAMS.reduce(
-	(acc, param) => ({ ...acc, [param]: z.string().optional() }),
+	(acc, param) => ({
+		...acc,
+		[param]: z.string().optional()
+	}),
```
AntonioVentilii: what changed here? just the formatting? if yes, let's put it back

src/frontend/src/tests/lib/utils/qr-code.utils.spec.ts
roman-nazaruk: I added another tests for it

src/frontend/src/tests/lib/utils/qr-code.utils.spec.ts
roman-nazaruk: same

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -24,7 +26,13 @@ import { decodePayment } from '@icp-sdk/canisters/ledger/icrc';
  * @param {string} urn - The URN string to decode.
  * @returns {DecodedUrn | undefined} The decoded URN object, or undefined if the URN string does not match the expected pattern.
  */
-export const decodeQrCodeUrn = (urn: string): DecodedUrn | undefined => {
+export const decodeQrCodeUrn = ({
+	urn,
+	isDFX = false
```
AntonioVentilii, line 41: let's add a comment (long explanation) in the documentation of the function, explaining the entire issue with `number` and `bigint`, the reference to the specs, etc

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -36,11 +44,18 @@ export const decodeQrCodeUrn = (urn: string): DecodedUrn | undefined => {
 
 	const processParam = ([key, value]: [string, string]) => {
 		if ((URN_NUMERIC_PARAMS as readonly string[]).includes(key)) {
+			if (isDFX) {
+				return { [key]: value };
```
AntonioVentilii, line 59: same here, but here can be a normal comment

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -36,11 +44,18 @@ export const decodeQrCodeUrn = (urn: string): DecodedUrn | undefined => {
...
+			}
 			return { [key]: parseFloat(value) };
 		}
+
 		if ((URN_STRING_PARAMS as readonly string[]).includes(key)) {
 			return { [key]: value };
 		}
+		if (isDFX && !isNaN(parseFloat(value))) {
+			return { [key]: value };
```
AntonioVentilii, line 70: same

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -78,7 +94,18 @@ export const decodeQrCodeUrn = (urn: string): DecodedUrn | undefined => {
 		...params
 	};
 
+	if (isDFX) {
+		const result = DecodedUrnBigIntSchema.safeParse(decodedUrn);
+		if (!result.success) {
+			console.warn('QR code cannot be correctly parsed:', result.error);
+			return undefined;
```
AntonioVentilii:
```suggestion
			return;
```

src/frontend/src/lib/types/qr-code.ts
AntonioVentilii: same question: can `z.number()` be removed?

src/frontend/src/lib/services/open-crypto-pay.services.ts
```diff
@@ -187,7 +188,7 @@ const preparePaymentTransaction = async ({
 
 	progress(ProgressStepsPayment.CREATE_TRANSACTION);
 
-	const decodedData = decodeQrCodeUrn(uri);
+	const decodedData = decodeQrCodeUrn({ urn, isDFX: true }) as DecodedUrnBigInt;
```
AntonioVentilii, line 191: is the casting necessary? if you remove it, does it return a bad type?

src/frontend/src/eth/utils/qr-code.utils.ts
```diff
@@ -120,8 +120,10 @@ export const decodeQrCode = ({
 	}
 
 	const amount =
-		functionName === 'transfer'
-			? uint256
+		functionName === 'transfer' && nonNullish(uint256)
+			? typeof uint256 === 'number'
```
AntonioVentilii, line 126: according to the specs, with `transfer`, we should always ONLY use `uint256`, even if `undefined`:

> Requesting payments in [ERC-20](https://eips.ethereum.org/EIPS/eip-20) tokens involves a request to call the `transfer` function of the token contract with an `address` and a `uint256` typed parameter, containing the beneficiary address and the amount in atomic units, respectively. For example, requesting a Unicorn to address `0x8e23ee67d1332ad560396262c48ffbb01f93d052` looks as follows: `ethereum:0x89205a3a3b2a69de6dbf7f01ed13b2108b2c43e7/transfer?address=0x8e23ee67d1332ad560396262c48ffbb01f93d052&uint256=1`

AntonioVentilii (commented):
Nice! A few comments

src/frontend/src/lib/services/open-crypto-pay.services.ts
roman-nazaruk: Unfortunately, yes we need casting here

src/frontend/src/eth/utils/qr-code.utils.ts
roman-nazaruk: Yes, but since we cast our types, it will come in as undefined, and value will also be undefined here. So it works the same as before, but with one additional condition that resolves the TypeScript issue where uint256 is always either a number or undefined.

const amount =
		functionName === 'transfer' && nonNullish(uint256)
			? typeof uint256 === 'number'
				? uint256
				: Number(uint256)
			: nonNullish(value)
				? +formatToken({ value: BigInt(value), unitName: token.decimals })
				: undefined;

src/frontend/src/lib/utils/qr-code.utils.ts
roman-nazaruk: Added

src/frontend/src/lib/utils/qr-code.utils.ts
roman-nazaruk: Done

src/frontend/src/lib/utils/qr-code.utils.ts
roman-nazaruk: Done

src/frontend/src/eth/utils/qr-code.utils.ts
AntonioVentilii: not really, because if `uin256` is nullish, it will skip completely the `uint256` part, no? and it will analyze `value` (that we cannot guarantee that is `undefined`).

Let's leave it as before, and if needed to change, we can do it in another PR

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -16,6 +18,15 @@ import { decodePayment } from '@icp-sdk/canisters/ledger/icrc';
...
  * /^([a-zA-Z]+):([a-zA-Z0-9\-.]+)(@(\d+))?(\/([a-zA-Z]+))?(\?(.*))?$/
  *
+ *
+ * When the `isDFX` flag is enabled, all numeric values (e.g., `value`, `uint256`)
+ * are expected to be parsed and returned as **BigInt**, not as JavaScript `number`.
+ *
+ * JavaScript `number` cannot safely represent large integers used in blockchain
+ * environments (such as ERC-20 token amounts or values denominated in wei),
+ * because it is limited to 53 bits of precision. Any value exceeding this limit
+ * will be silently rounded and lose accuracy.
```
AntonioVentilii, line 28: can you add that this is an issue of DFX and that the specifics (with URL reference) says otherwise to what the URL of DFX returns? And that until DFX fixes it this is a workaround to adapt it. Plus, add a `TODO:`

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -37,16 +54,26 @@ export const decodeQrCodeUrn = ({ urn }: { urn: string }): DecodedUrn | undefine
...
+				return { [key]: value };
+			}
 			return { [key]: parseFloat(value) };
 		}
 
 		if ((URN_STRING_PARAMS as readonly string[]).includes(key)) {
 			return { [key]: value };
 		}
 
+		// Treat numeric values as bigint when the isDFX flag is true, as described in the JSDoc above.
```
AntonioVentilii, line 68: let's make an explicit specification: why do we treat it as bigint? 

```ts
// Due to the DFX issue described in the JSDoc, we need to parse the `value` params as `bigint`, instead of `number`, because DFX URL provides it incorrectly as `bigint`.
// TODO: Remove this workaround when DFX fixes the issue on their side
```

src/frontend/src/lib/utils/qr-code.utils.ts
```diff
@@ -81,6 +108,16 @@ export const decodeQrCodeUrn = ({ urn }: { urn: string }): DecodedUrn | undefine
 		...params
 	};
 
+	if (isDFX) {
+		const result = DecodedUrnBigIntSchema.safeParse(decodedUrn);
```
AntonioVentilii, line 112: same here: As described....

PR #10831, opened 2025-12-10
refactor(frontend): refactored decoding payment data

roman-nazaruk:
# Motivation

We need to refactor decoding data for DFX for native and erc20 tokens


AntonioVentilii (commented):
Maybe there was some additional commit to this PR? It does not seem only a refactor

PR #10836, opened 2025-12-11
feat(build): Include legacy EXT tokens

AntonioVentilii:
# Motivation

After having thoroughly tested, we can include the legacy standard of EXT tokens.


DenysKarmazynDFINITY (approved):
LGTM, thanks

PR #10837, opened 2025-12-11
feat(frontend): extract value or uint256 from erc681

roman-nazaruk:
# Motivation

We need to extract the value or uint256 from ERC-681 to compare it with our value.
# Changes

Created util method getERC681Value

# Tests

Covered util with tests

### DIFF
--- a/src/frontend/src/lib/utils/open-crypto-pay.utils.ts
+++ b/src/frontend/src/lib/utils/open-crypto-pay.utils.ts
@@ -272,6 +272,16 @@ export const getERC681Value = (uri: string): bigint | undefined => {
 			return;
 		}
 
+		if (value.includes('e') || value.includes('E') || value.includes('.')) {
+			const number = parseFloat(value);
+
+			if (!isFinite(number)) {
+				return;
+			}
+
+			return BigInt(number);
+		}
+
 		return BigInt(value);
 	} catch (_: unknown) {
 		// If it is not parseable, we can handle a nullish value
--- a/src/frontend/src/tests/lib/utils/open-crypto-pay.utils.spec.ts
+++ b/src/frontend/src/tests/lib/utils/open-crypto-pay.utils.spec.ts
@@ -1542,6 +1542,51 @@ describe('open-crypto-pay.utils', () => {
 			});
 		});
 
+		describe('Scientific notation', () => {
+			it('should parse scientific notation with lowercase e', () => {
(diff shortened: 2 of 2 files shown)


src/frontend/src/lib/utils/open-crypto-pay.utils.ts
```diff
@@ -262,3 +262,18 @@ export const validateDecodedData = ({
...
 	};
 };
+
+export const getERC681Value = (uri: string): bigint | undefined => {
+	try {
+		const params = new URLSearchParams(uri.split('?')[1] || '');
+		const value = params.get('value') ?? params.get('uint256');
+
+		if (isEmptyString(value)) {
+			return undefined;
```
roman-nazaruk:
```suggestion
			return;
```

src/frontend/src/lib/utils/open-crypto-pay.utils.ts
```diff
@@ -262,3 +262,18 @@ export const validateDecodedData = ({
 		estimatedGasLimit
 	};
 };
+
+export const getERC681Value = (uri: string): bigint | undefined => {
```
AntonioVentilii, line 266: if this is going to be used in the URN decoder utils, better to put it in the same module

src/frontend/src/lib/utils/open-crypto-pay.utils.ts
```diff
@@ -262,3 +262,18 @@ export const validateDecodedData = ({
...
+		const value = params.get('value') ?? params.get('uint256');
+
+		if (isEmptyString(value)) {
+			return;
+		}
+
+		return BigInt(value);
+	} catch {
+		return undefined;
+	}
```
AntonioVentilii:
```suggestion
	} catch (_: unknown) {
		// If it is not parseable, we can handle a nullish value
	}
```

src/frontend/src/lib/utils/open-crypto-pay.utils.ts
roman-nazaruk: No, only here for dfx

sbpublic: Agree with Antonio.
Such a function belongs to an EVM scope util.
And also, this file is open-crypto-pay named, which is a standard, not a provider. We can have multiple providers that implement this standard, so if we have to do something dfx specific, then it should be in a dfx object, not one for the standard itself.

PR #10838, opened 2025-12-11
feat(frontend): added dfx validation for native and erc20 tokens

roman-nazaruk:
# Motivation

We need to validate native and erc20 tokens with dfx response

# Changes

Added methods to validate 

- validateNativeTransfer 
- validateErc20Transfer 

# Tests

Covered with tests


src/frontend/src/lib/utils/open-crypto-pay.utils.ts
```diff
@@ -262,3 +267,135 @@ export const validateDecodedData = ({
...
+
+		if (isEmptyString(value)) {
+			return undefined;
+		}
+
+		return BigInt(value);
+	} catch {
+		return undefined;
+	}
+};
```
AntonioVentilii, line 294: this is from another PR, right? if yes, let's wait first for the other to merge

may you please put it in draft for now?

PR #10841, opened 2025-12-11
feat(frontend): add env variable for the primary II version

DenysKarmazynDFINITY:
# Motivation

We need to create a new env variable for switching between II versions. It will be removed from the codebase after the current latest version (2.0) becomes a standard in OISY.


.github/workflows/deploy-to-environment.yml
```diff
@@ -349,6 +354,7 @@ jobs:
               echo "VITE_EARNING_ENABLED=$VITE_EARNING_ENABLED_BETA"
               echo "VITE_AI_ASSISTANT_CONSOLE_ENABLED=$VITE_AI_ASSISTANT_CONSOLE_ENABLED_BETA"
               echo "VITE_UNIVERSAL_SCANNER_ENABLED=$VITE_UNIVERSAL_SCANNER_ENABLED_BETA"
+              echo "VITE_PRIMARY_INTERNET_IDENTITY_VERSION=$(get VITE_PRIMARY_INTERNET_IDENTITY_VERSION "$VITE_PRIMARY_INTERNET_IDENTITY_VERSION_BETA")"
```
AntonioVentilii:
```suggestion
              echo "VITE_PRIMARY_INTERNET_IDENTITY_VERSION=$VITE_PRIMARY_INTERNET_IDENTITY_VERSION_BETA"
```

src/frontend/src/env/auth.env.ts
```diff
@@ -0,0 +1,3 @@
+// TODO: to be removed when II 2.0 becomes the default version
+export const PRIMARY_INTERNET_IDENTITY_VERSION =
+	import.meta.env.VITE_PRIMARY_INTERNET_IDENTITY_VERSION ?? '1.0';
```
AntonioVentilii: should we maybe use the util `notEmptyString` ? it checks both for undefined and empty string

I imagine you will handle the mapping elsewhere afterwards

DenysKarmazynDFINITY: Good idea, updated. Yes, the var will be used in the next PRs.

.github/workflows/deploy-to-environment.yml
DenysKarmazynDFINITY: oops, good catch

PR #10860, opened 2025-12-12
feat(frontend): update ButtonAuthenticateWithHelp component

DenysKarmazynDFINITY:
# Motivation

We need to prepare the ButtonAuthenticateWithHelp component for the case when II version env var is 2.0:
* The primary button uses the 2.0 domain
* The secondary button that uses the 1.0 domain appears on the page
* T&C text gets updated

Screenshot 2025-12-12 at 11 52 50


src/frontend/src/lib/components/auth/ButtonAuthenticateWithHelp.svelte
```diff
@@ -36,7 +45,36 @@
...
+				<IconHelp size="18" />
+			</ExternalLink>
+
+			<button
+				aria-label={$i18n.auth.text.sign_in_with_identity_number}
+				onclick={() => onAuthenticate(InternetIdentityDomain.VERSION_1_0)}
+			>
+				{$i18n.auth.text.sign_in_with_identity_number}
+			</button>
+		</div>
```
AntonioVentilii: just for manageability and readability, shall this be put in another component? If we ever want to re-use it too

DenysKarmazynDFINITY: good idea, done

PR #10872, opened 2025-12-12
feat(frontend): New object type for Token standard

AntonioVentilii:
# Motivation

We are going to transform the `standard` field of a `Token` type from an `enum` to an `object` (so that we can extend it with other fields like `version` or `subStandards`).

In this PR, we create the new type.

### DIFF
--- a/src/frontend/src/tests/lib/schema/token.schema.spec.ts
+++ b/src/frontend/src/tests/lib/schema/token.schema.spec.ts
@@ -113,8 +113,8 @@ describe('token.schema', () => {
 			expect(TokenStandardSchema.parse({ code: validStandard })).toEqual({ code: validStandard });
 		});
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/sol/services/worker.sol-wallet.services.ts
```diff
@@ -95,7 +96,7 @@ export class SolWalletWorker extends AppWorker implements WalletWorker {
 			tokenOwnerAddress
 		};
 
-		const worker = await AppWorker.getInstance();
+		const worker = await AppWorker.getInstance({ asSingleton: isIOS() });
```
AntonioVentilii:
```suggestion
		const worker = await AppWorker.getInstance();
```

roman-nazaruk (approved):

PR #10892, opened 2025-12-15
feat(frontend): Service to extract nested media URLs

AntonioVentilii:
# Motivation

There are some NFT media URLs that have nested URLs in their rendering. Those are actually the real NFT media (for example [here](https://rhg63-2aaaa-aaaag-qcwhq-cai.raw.icp0.io/?tokenid=3i4l7-bqkor-uwiaa-aaaaa-buavr-4aqca-aahew-a) and [here](https://deb2y-miaaa-aaaam-qa3rq-cai.raw.icp0.io/?tokenid=hejvp-aykor-uwiaa-aaaaa-deag4-maqca-aaaby-a)).

So, we will inform the users about it, and let they decide.

In this PR, we create a service to extract nested media URLs from the rendering.

### DIFF
--- a/src/frontend/src/lib/services/url.services.ts
+++ b/src/frontend/src/lib/services/url.services.ts
@@ -2,19 +2,19 @@ import { browser } from '$app/environment';
 import { isEmptyString, isNullish, nonNullish } from '@dfinity/utils';
 
 const toAbsUrl = (src: string): string | undefined => {
-	const v = src.trim();
+	const parsedSrc = src.trim();
 
-	if (isEmptyString(v)) {
+	if (isEmptyString(parsedSrc)) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/services/url.services.ts
```diff
@@ -0,0 +1,113 @@
+import { browser } from '$app/environment';
+import { isEmptyString, isNullish, nonNullish } from '@dfinity/utils';
+
+const toAbsUrl = (src: string): string | undefined => {
+	const v = src.trim();
```
DenysKarmazynDFINITY: nitpick: not sure what `v` stands for, can we please adjust it to a full word?

DenysKarmazynDFINITY (approved):

PR #10895, opened 2025-12-15
feat(frontend): Use more complex Token standard

AntonioVentilii:
# Motivation

We want to extend the token standard with more details. So, in this PR, we first use the new `TokenStandard` type everywhere in the code.


DenysKarmazynDFINITY (approved):
Massive job!

PR #10910, opened 2025-12-17
feat(frontend): Show additional media URL for NFTs

AntonioVentilii:
# Motivation

The media URL of some NFTs (for example EXT NFTs like [this one](https://rhg63-2aaaa-aaaag-qcwhq-cai.raw.icp0.io/?tokenid=3i4l7-bqkor-uwiaa-aaaaa-buavr-4aqca-aahew-a)) have nested URLs per-se, making it difficult to really render the correct media.

To have a full disclaimer for the user, we extract such nested URL and show it to the user with a message.

# Changes

- Fetch additional media URL in component `NftMetadataList`.
- Render a message box with a disclaimer in case the additiona URL is non-nullish.

# Tests

Since we are here we created dedicated unittests (mostly copied from existing one of the consumers `NftHero` and `NftCollectionHero`).

Plus a practical test:

Screenshot 2025-12-17 at 06 17 13

### DIFF
--- a/src/frontend/src/lib/components/nfts/NftMetadataList.svelte
+++ b/src/frontend/src/lib/components/nfts/NftMetadataList.svelte
@@ -47,7 +47,7 @@
 
 	const updateAdditionalMediaUrl = async () => {
 		if (isNullish(nft?.imageUrl) || !allowMedia) {
-			additionalMediaUrl = undefined
+			additionalMediaUrl = undefined;
 
 			return;
 		}


src/frontend/src/lib/components/nfts/NftMetadataList.svelte
```diff
@@ -169,14 +197,43 @@
...
+			<div class="mt-2" in:slide={SLIDE_PARAMS}>
+				<MessageBox level="info">
+					<div class="flex flex-col gap-1 text-sm">
+						{$i18n.nfts.text.media_stored_at_different_location}
+
+						<output class="truncate text-tertiary">
+							{additionalMediaUrl}
+						</output>
+
+						<div class="items-end">
```
DenysKarmazynDFINITY, line 213: I think just having `items-end` does not bring any effect - did you mean to add `flex` too? or shall we remove it completely?

AntonioVentilii: I tried without and it was like this

Screenshot 2025-12-17 at 09 43 25

Maybe it can be left there without `flex`, right?

PR #10912, opened 2025-12-17
feat(frontend): Use custom tokens for service `autoLoadErc20Token`

AntonioVentilii:
# Motivation

Since `UserToken` type is deprecated in favour of `CustomToken`, we can effectively change the service `autoLoadErc20Token` to use the latter (similar to what `autoLoadIcrcToken` does).


DenysKarmazynDFINITY (approved):
nice!

PR #10994, opened 2026-01-07
feat(frontend): capitalize date's first letter in grouped txs list

DenysKarmazynDFINITY:
# Motivation

The previous implementation contained a tailwind class that actually does not exist ('first-letter:capitalize'). Instead, to capitalize only the first symbol of the date string, we now use TS.

### DIFF
--- a/src/frontend/src/lib/components/transactions/TransactionsDateGroup.svelte
+++ b/src/frontend/src/lib/components/transactions/TransactionsDateGroup.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 	import { slide } from 'svelte/transition';
+	import { capitalizeFirstLetter } from '../../../tests/utils/string-utils';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/transactions/TransactionsDateGroup.svelte
```diff
@@ -15,15 +15,17 @@
 	}
 
 	let { formattedDate, transactions, testId }: Props = $props();
+
+	let capitalizedFormattedDate = $derived(
```
AntonioVentilii: can the util `capitalizeFirstLetter` be use here?

DenysKarmazynDFINITY: not only can - it should 😄  updated

PR #11000, opened 2026-01-07
feat(frontend): Add destination to the Send context

AntonioVentilii:
# Motivation

We need to start analyzing if the user is sending the token to the minter. To do so, instead of reformulating the information in each component, we can set it in the send context.

So, first, we need to save the destination in the conntext.

### DIFF
--- a/src/frontend/src/lib/stores/send.store.ts
+++ b/src/frontend/src/lib/stores/send.store.ts
@@ -1,5 +1,6 @@
 import { exchanges } from '$lib/derived/exchange.derived';
 import { balancesStore } from '$lib/stores/balances.store';
+import type { Address } from '$lib/types/address';
 import type { OptionBalance } from '$lib/types/balance';
 import type { NetworkId } from '$lib/types/network';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/stores/send.store.ts
```diff
@@ -67,9 +70,10 @@ export interface SendContext {
 	sendTokenId: Readable<TokenId>;
 	sendTokenStandard: Readable<TokenStandard>;
 	sendTokenSymbol: Readable<string>;
-	sendBalance: Readable<OptionBalance>;
 	sendTokenExchangeRate: Readable<number | undefined>;
 	sendTokenNetworkId: Readable<NetworkId>;
+	sendBalance: Readable<OptionBalance>;
+	sendDestination: Writable<string>;
```
DenysKarmazynDFINITY: If you prefer so, we can use `Address` here

AntonioVentilii: good one! changed!

PR #11011, opened 2026-01-07
feat(frontend): display ICP tokens fee in token details

DenysKarmazynDFINITY:
# Motivation

Same as for ICRC tokens, we need to display the token fee for ICP standard tokens.

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokenModalContent.svelte
+++ b/src/frontend/src/lib/components/tokens/TokenModalContent.svelte
@@ -2,7 +2,7 @@
 	import { nonNullish } from '@dfinity/utils';
 	import type { Snippet } from 'svelte';
 	import { isTokenErc20 } from '$eth/utils/erc20.utils';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/tokens/TokenModalContent.svelte
```diff
@@ -173,7 +173,7 @@
 				{/snippet}
 			</ModalListItem>
 
-			{#if isTokenIcrc(token)}
+			{#if isTokenIcrc(token) || isTokenIcp(token)}
```
AntonioVentilii: maybe `isTokenIc` to include dip20?

AntonioVentilii (approved):

src/frontend/src/lib/components/tokens/TokenModalContent.svelte
DenysKarmazynDFINITY: done!

PR #11065, opened 2026-01-12
feat(frontend): add additional Hero info list item

DenysKarmazynDFINITY:
# Motivation

We need to add one more info list item to the Hero component.

Screenshot 2026-01-12 at 17 22 10


src/frontend/src/lib/components/icons/IconSocialLogin.svelte
```diff
@@ -0,0 +1,33 @@
+<svg fill="none" height="24" viewBox="0 0 72 24" width="72" xmlns="http://www.w3.org/2000/svg">
```
AntonioVentilii, line 1: no need for a disclaimer for license here, correct?

DenysKarmazynDFINITY: I think we don't need it since it's a set of branded icons (apple, google, etc)

PR #11157, opened 2026-01-16
feat(frontend): Delete data on logout only once

AntonioVentilii:
# Motivation

In the logout service it makes more sense to either delete all principals data or just the current one.

So, we adjust the parameters to this respect.

### DIFF
--- a/src/frontend/src/lib/services/auth.services.ts
+++ b/src/frontend/src/lib/services/auth.services.ts
@@ -57,6 +57,12 @@ import { isNullish } from '@dfinity/utils';
 import type { Principal } from '@icp-sdk/core/principal';
 import { get } from 'svelte/store';
 
+export enum PrincipalsStorage {
+	CURRENT = 'current',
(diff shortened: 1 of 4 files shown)


src/frontend/src/lib/services/auth.services.ts
```diff
@@ -128,18 +128,18 @@ export const signIn = async (
 
 export const signOut = ({
 	resetUrl = false,
-	clearAllPrincipalsStorages = false,
+	clearPrincipalStorages = 'current',
 	source = ''
 }: {
 	resetUrl?: boolean;
-	clearAllPrincipalsStorages?: boolean;
+	clearPrincipalStorages?: 'current' | 'all' | 'none';
```
DenysKarmazynDFINITY: let's create an enum for that and use it instead of strings

AntonioVentilii: done

PR #11206, opened 2026-01-19
feat(frontend): Cache NFTs store

AntonioVentilii:
# Motivation

NFTs are fetched from the network on every page load, causing a blank/loading state until the API responses come back. Caching NFTs in IndexedDB allows the UI to render instantly from the cache while fresh data is fetched in the background — matching the existing pattern used for balances and custom tokens.

# Changes

- **`idb-nfts.api.ts`** — New IndexedDB API for NFTs using `idb-keyval`. Handles serialization of `Symbol`-based IDs (`NetworkId`, `TokenId`) to strings for storage, and deserialization back to the correct singleton `Symbol` instances by looking up `SUPPORTED_NETWORKS`.
- **`idb-nfts.ts`** — Serializable type definitions (`SerializableNetwork`, `SerializableNftCollection`, `SerializableNft`) that replace `Symbol` IDs with `string` for IndexedDB compatibility.
- **`LoaderNfts.svelte`** — Loads cached NFTs from IDB on first render (before live data arrives) and debounces writes back to IDB when the store updates, following the same pattern as `BalancesIdbSetter`.
- **`auth.services.ts`** — Added `clearIdbNfts` to the sign-out cleanup list.

# Tests

Added tests.


AntonioVentilii:
not a good solution since we rely on unique network symbols

PR #11211, opened 2026-01-20
feat(frontend): Debounce listing custom tokens from backend

AntonioVentilii:
# Motivation

During initialization (or in general during token flows) we call the service that lists the custom tokens from the backend. However this call may happen simultaneously in different parts of the codebase.

To avoid fire several calls for the same function with the same parameters, we want to debounce it.

We should be aware that there are services, like `queryAndUpdate`, that call the function twice as `query` and as `update`. So, in this case, we should not debounce them.

# Changes

- Create debounced `query` call for `listCustomTokens`.
- Create debounced `update` call for `listCustomTokens`.
- Cache the pending calls, so that they can be returned the same value (or the same error) of the last call.
- Finalize the debounced service and use it in the consumers.

# Tests

Created tests and adapted the rest. Note that the tests for the main service `loadNetworkCustomTokens` should not change, expect for the fact that it will be slightly slower because of the debouncing.

Furthermore, manually tested in test canister and the initial calls to the backend method are remarkably reduced.

### DIFF
--- a/src/frontend/src/lib/services/custom-tokens.services.ts
+++ b/src/frontend/src/lib/services/custom-tokens.services.ts
@@ -108,15 +108,7 @@ export const debounceListCustomTokens = (
 	new Promise<CustomToken[]>((resolve, reject) => {
 		const key: DebounceWindowKey = params.certified ? 'update' : 'query';
 
-		const window =
-			windows[key] ??
-			(() => {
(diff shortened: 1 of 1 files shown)


AntonioVentilii:
In PR https://github.com/dfinity/oisy-wallet/pull/11217, we introduced a cache-store for the backend custom tokens, so no need to debounce the value.

PR #11239, opened 2026-01-21
feat(frontend): Delete all OISY-related IDB on signout

AntonioVentilii:
# Motivation

On sign-out, when we want to clear all principals from the IDB, we actually should even delete all possible OISY-related DBs .

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensDisplayHandler.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { nonNullish } from '@dfinity/utils';
-	import { onDestroy, type Snippet, untrack } from 'svelte';
+	import { type Snippet, untrack } from 'svelte';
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/api/idb.api.ts
```diff
@@ -0,0 +1,34 @@
...
+			reject(new Error(`IndexedDB database deletion blocked: ${name}`))
+		);
+	});
+
+export const deleteIdbAllOisyRelated = async () => {
+	try {
+		const dbInfo = await indexedDB.databases();
+
+		const promises = dbInfo.map(async ({ name }) => {
+			if (isNullish(name) || !name.startsWith('oisy-')) {
```
DenysKarmazynDFINITY, line 22: can there be non-oisy databases?

AntonioVentilii: yes, for example, we have the wallet connect and the auth one (the one created by agent-js)

DenysKarmazynDFINITY: got it, thanks

PR #11242, opened 2026-01-21
style(frontend): Use Tailwind `transition-colors` where possible

AntonioVentilii:
# Motivation

For snippets where we change the colors on hovering, we can substitute `transition` with `transition-colors`.

### DIFF
--- a/src/frontend/src/lib/components/tokens/TokensList.svelte
+++ b/src/frontend/src/lib/components/tokens/TokensList.svelte
@@ -130,7 +130,7 @@
 					{:else}
 						{@const { token } = tokenOrGroup}
 
-						<div class="transition duration-300 hover:bg-primary">
+						<div class="transition-colors duration-300 hover:bg-primary">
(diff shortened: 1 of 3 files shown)


src/frontend/src/lib/components/tokens/TokensList.svelte
```diff
@@ -169,8 +169,8 @@
 					{/snippet}
 
 					{#each enableMoreTokensWithKey as { tokenOrGroup, key } (key)}
-						<div class="overflow-hidden rounded-xl">
-							<div class="transition duration-300 hover:bg-primary">
+						<div class="overflow-hidden rounded-xl" animate:flip={flipParams}>
```
DenysKarmazynDFINITY, line 172: this one didn't have the animation before, is this added on purpose here?

AntonioVentilii: yes, all good, i forgot to re-put it before

PR #11250, opened 2026-01-22
feat(frontend): Cache finalized SOL parsed transactions

AntonioVentilii:
# Motivation

If a Solana signature is finalized, it is quite definitive (unless forced changes in the chain). So we can safely cache those values, to reduce calls.


AntonioVentilii:
no need to cache the entire list of parsed transactions, since we have:

- https://github.com/dfinity/oisy-wallet/pull/11249

PR #11260, opened 2026-01-22
feat(frontend): Optional delete IDB storage in `gotoReplaceRoot`

AntonioVentilii:
# Motivation

When we use the service `gotoReplaceRoot` during logout, we want to pass on the information if we need to clear the IDB storage.

### DIFF
--- a/src/frontend/src/tests/lib/utils/nav.utils.spec.ts
+++ b/src/frontend/src/tests/lib/utils/nav.utils.spec.ts
@@ -174,7 +174,7 @@ describe('nav.utils', () => {
 			expect(mockGoTo).toHaveBeenCalledExactlyOnceWith('/', { replaceState: true });
 		});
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/tests/lib/utils/nav.utils.spec.ts
```diff
@@ -163,10 +164,22 @@ describe('nav.utils', () => {
...
+		});
+
 		it('should navigate to "/" with replaceState', async () => {
 			await gotoReplaceRoot();
 
-			expect(mockGoTo).toHaveBeenCalledWith('/', { replaceState: true });
+			expect(mockGoTo).toHaveBeenCalledExactlyOnceWith('/', { replaceState: true });
+		});
+
+		it('should navigate to "/?deleteIdbCache=true" with replaceState when deleteIdbCache is true', async () => {
```
AntonioVentilii:
```suggestion
		it('should navigate with param to delete cache with replaceState when deleteIdbCache is true', async () => {
```

PR #11277, opened 2026-01-23
feat(frontend): Allow service `mapBackendTokens` to receive tokens as inputs

AntonioVentilii:
# Motivation

We will reuse the services to load custom tokens to avoid loading the backend tokens multiple times. To do such a feat, we need to be able to pass a default list of tokens, and if not provided, they are indeed fetched.

### DIFF
--- a/src/frontend/src/tests/lib/services/load-tokens.services.spec.ts
+++ b/src/frontend/src/tests/lib/services/load-tokens.services.spec.ts
@@ -239,7 +239,7 @@ describe('load-tokens.services', () => {
 			expect(toastsError).not.toHaveBeenCalled();
 		});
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/tests/lib/services/load-tokens.services.spec.ts
```diff
@@ -236,14 +239,14 @@ describe('load-tokens.services', () => {
 			expect(toastsError).not.toHaveBeenCalled();
 		});
 
-		it('should return an empty array if no tokens are filtered successfully', async () => {
+		it('should return and empty array if no tokens are filtered successfully', async () => {
```
AntonioVentilii:
```suggestion
		it('should return an empty array if no tokens are filtered successfully', async () => {
```

PR #11308, opened 2026-01-26
style(frontend): Remove the option to log in via the legacy II

sbpublic:
# Motivation

II 1.0 will be sunsetted on 29.01.2026

# Changes
Reverts changes from https://github.com/dfinity/oisy-wallet/pull/11057

- remove the legacy login button
- adjust styles back to 1 button settings
- and its tests
- and localizations

# Tests

| Old | Prod | New |
|---|---|---|
|image|image|image|
|image|image|image|
|image|image|image|

### DIFF
--- a/src/frontend/src/lib/components/auth/ButtonAuthenticateWithHelp.svelte
+++ b/src/frontend/src/lib/components/auth/ButtonAuthenticateWithHelp.svelte
@@ -42,7 +42,6 @@
 >
 	<ButtonAuthenticate
 		{fullWidth}
-		isLandingPage
 		onclick={() => onAuthenticate(InternetIdentityDomain.VERSION_2_0)}
 		styleClass="bg-brand-primary text-primary-inverted"
 		testId={LOGIN_BUTTON}


src/frontend/src/lib/components/ui/ButtonAuthenticate.svelte
```diff
@@ -14,13 +14,12 @@
 </script>
 
 <button
-	class={`login-button flex w-full items-center justify-center gap-2 rounded-2xl py-3 text-lg leading-6 font-bold whitespace-nowrap sm:max-w-none ${styleClass ?? ''}`}
+	class={`login-button flex w-full items-center justify-center gap-4 rounded-2xl py-3 text-lg leading-6 font-bold whitespace-nowrap sm:max-w-none ${styleClass ?? ''}`}
 	class:lg:px-6={!fullWidth}
 	class:max-w-[24rem]={!fullWidth}
 	class:px-12={fullWidth || !isLandingPage}
```
DenysKarmazynDFINITY: Can you please also remove the `isLandingPage ` prop? It (prev. called `isPrimaryIdentityVersion2`) was introduced for backwards compatibility if the feature flag has to be switched back to II1.0, and is not needed anymore. You would need remove it from this component, and from ButtonAuthenticateWithHelp (the only place where it's currently used).

PR #11328, opened 2026-01-27
feat(frontend): Allow tokens with same name

AntonioVentilii:
# Motivation

Since we typically index or list or set the URL based on the token symbol, it does not make sense to avoid tokens with same name.

For now, we really want tokens with different symbol.

### DIFF
--- a/src/frontend/src/icp/services/ext-add-custom-tokens.service.ts
+++ b/src/frontend/src/icp/services/ext-add-custom-tokens.service.ts
@@ -4,7 +4,8 @@ import { i18n } from '$lib/stores/i18n.store';
 import { toastsError } from '$lib/stores/toasts.store';
 import type { OptionIdentity } from '$lib/types/identity';
 import { shortenWithMiddleEllipsis } from '$lib/utils/format.utils';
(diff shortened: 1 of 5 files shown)


src/frontend/src/icp/services/ext-add-custom-tokens.service.ts
```diff
@@ -75,11 +75,7 @@ const assertExistingTokens = ({
 	token: ExtTokenWithoutId;
 }): { valid: boolean } => {
 	if (
-		extTokens.find(
-			({ symbol, name }) =>
-				symbol.toLowerCase() === token.symbol.toLowerCase() ||
-				name.toLowerCase() === token.name.toLowerCase()
-		) !== undefined
+		nonNullish(extTokens.find(({ symbol }) => symbol.toLowerCase() === token.symbol.toLowerCase()))
```
DenysKarmazynDFINITY: shall we create a util for that?

AntonioVentilii: good idea! done!

PR #11346, opened 2026-01-28
refactor(frontend): Extract component to guard media display

AntonioVentilii:
# Motivation

For NFTs, we have a component in place that display (or not) an NFT image based on the consent and/or media status.

Now, we are thinking of having the same flow for custom tokens icon: we will allow the users to decide whether to show or not a token logo (for security reasons).

In this PR, we extract a component that can be re-used as media guard.


src/frontend/src/lib/components/guard/MediaDisplayGuard.svelte
```diff
@@ -0,0 +1,88 @@
...
+	import FilesizeLimitExceededImage from '$lib/components/icons/nfts/FilesizeLimitExceeded.svelte';
+	import InvalidDataImage from '$lib/components/icons/nfts/InvalidData.svelte';
+	import UnsupportedMediaTypeImage from '$lib/components/icons/nfts/UnsupportedMediaType.svelte';
+	import Button from '$lib/components/ui/Button.svelte';
+	import { NftMediaStatusEnum } from '$lib/schema/nft.schema';
+	import { i18n } from '$lib/stores/i18n.store';
+
+	interface Props {
+		type: 'hero-banner' | 'card' | 'card-selectable' | 'nft-display' | 'nft-logo';
+		mediaStatus: NftMediaStatusEnum;
```
DenysKarmazynDFINITY, line 15: probably, your plan is to rename this enum into `MediaStatusEnum` in the follow-up PRs?

AntonioVentilii: correct! and all the other things too: strings, sub-components, etc

I didn't do it in this PR, because it is really a lot of changes

PR #11376, opened 2026-01-30
refactor(frontend): Rename folder and files for env ICRC tokens

AntonioVentilii:
# Motivation

The env tokens files and folder are divided by standard for IC tokens. So it makes more sense to rename them all with the ICRC index and not ICP.


src/frontend/src/icp/utils/ic-send.utils.ts
```diff
@@ -57,7 +57,8 @@ export const isInvalidDestinationIc = ({
 	}
 
 	if (nonNullish(networkId) && isNetworkIdEthereum(networkId)) {
-		return !isEthAddress(destination);
+		return !
+      (destination);
```
AntonioVentilii:
```suggestion
		return !isEthAddress(destination);
```

PR #11395, opened 2026-02-02
feat(frontend): update utxos calculation

DenysKarmazynDFINITY:
# Motivation

While checking the utxos calculation util, I noticed that the a BTC tx base size can be increased a bit in order to produce more precise results.


AntonioVentilii (approved):
Good catch

PR #11411, opened 2026-02-03
feat(frontend): add BTC loader to ScannerModal

DenysKarmazynDFINITY:
# Motivation

We need to add BTC loader component to ScannerModal.

### DIFF
--- a/src/frontend/src/lib/components/scanner/ScannerModal.svelte
+++ b/src/frontend/src/lib/components/scanner/ScannerModal.svelte
@@ -4,6 +4,7 @@
 	import { setContext } from 'svelte';
 	import OpenCryptoPayWizard from '$lib/components/open-crypto-pay/OpenCryptoPayWizard.svelte';
 	import ScannerCode from '$lib/components/scanner/ScannerCode.svelte';
+	import ScannerModalPayDataLoader from '$lib/components/scanner/ScannerModalPayDataLoader.svelte';
 	import { scannerWizardSteps } from '$lib/config/scanner.config';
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/scanner/ScannerModal.svelte
```diff
@@ -72,22 +73,24 @@
...
-	bind:this={modal}
-	disablePointerEvents={currentStep?.name === WizardStepsScanner.TOKENS_LIST}
-	{onClose}
-	{steps}
-	bind:currentStep
->
-	{#snippet title()}
-		{currentStep?.title}
-	{/snippet}
+<ScannerModalPayDataLoader>
```
AntonioVentilii, line 76: shouldn't it wrap only `OpenCryptoPayWizard`? or even go inside it?

Asking because ScannerModal will include other flows (for example WalletConnect) and it probably does not require the loader for all the situations

AntonioVentilii (commented):
one doubt:

src/frontend/src/lib/components/scanner/ScannerModal.svelte
DenysKarmazynDFINITY: There are 2 reasons why I put it there:
1. We need to keep it outside of the wizard component due to the re-rendering issue.
2. We want to start loading BTC data as soon as possible, even if there is a chance that it's not gonna be needed. The reason is to make the pay process as fast as possible, meaning that e.g. user doesn't need to wait in the store while some data is being loaded. If we start loading utxos and etc. when the Pay modal gets open, it's already "too late".

This part will be changed after we load the utxos data via workers.

PR #11416, opened 2026-02-03
feat(frontend): Split WalletConnect button from the listener

AntonioVentilii:
# Motivation

The component WalletConnectListener should be used at top-level to handle all WalletConnect connections/request.

However, before this PR, it pertains to the Header, that even if it is high-up, it is still not top-level.

So, we will split the component in the button-logic-part, and assign the listener-part to the other loaders.

# Changes

- Create new component `WalletConnect`.
- Move the WalletConnect button logic from `WalletConnectListener` to the new component.
- Use the new component in the `Header`.
- Move `WalletConnectListener` to `Loaders`.

# Tests

Tested manually in test canister fe2 both for ETH swaps and SOL swaps. Plus, tested the auto-reconnection when reloading the page.

Furthermore, this fixes the overflow issue of the Proposal modal in Android devices:

### Before

image-20260224-151106

### After

IMG_5252

### DIFF
--- a/src/frontend/src/lib/components/wallet-connect/WalletConnect.svelte
+++ b/src/frontend/src/lib/components/wallet-connect/WalletConnect.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
 	import { nonNullish } from '@dfinity/utils';
-	import { untrack } from 'svelte';
 	import { walletConnectReconnecting } from '$eth/stores/wallet-connect.store';
 	import WalletConnectButton from '$lib/components/wallet-connect/WalletConnectButton.svelte';
 	import { TRACK_COUNT_WALLET_CONNECT_MENU_OPEN } from '$lib/constants/analytics.constants';


src/frontend/src/lib/components/wallet-connect/WalletConnectSession.svelte
```diff
@@ -199,30 +184,10 @@
 	});
 
 	onDestroy(() => walletConnectPaired.set(false));
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: let's add it indeed

src/frontend/src/lib/components/wallet-connect/WalletConnect.svelte
```diff
@@ -0,0 +1,48 @@
+<script lang="ts">
+	import { nonNullish } from '@dfinity/utils';
+	import { untrack } from 'svelte';
```
AntonioVentilii:
```suggestion
```

src/frontend/src/lib/components/wallet-connect/WalletConnect.svelte
```diff
@@ -0,0 +1,48 @@
...
+	import { walletConnectReconnecting } from '$eth/stores/wallet-connect.store';
+	import WalletConnectButton from '$lib/components/wallet-connect/WalletConnectButton.svelte';
+	import { TRACK_COUNT_WALLET_CONNECT_MENU_OPEN } from '$lib/constants/analytics.constants';
+	import { authNotSignedIn } from '$lib/derived/auth.derived';
```
AntonioVentilii:
```suggestion
```

PR #11449, opened 2026-02-11
feat(frontend): remove the legacy icp token on ethereum from the list of curated tokens

sbpublic:
# Motivation

Currently we don't allow multiple tokens with the same symbol or name in a network. And Importing the ICP ethereum version from OneSec fails. Since this legacy icp on ethereum is not needed anymore, we can remove it.

**But we need to ensure the ICP specific logic does not apply to a new imported icp token!**

image

image

# Changes

- remove the token from the curated list
- ensure the current logic that applies to all evm tokens with the symbol icp does not apply to new tokens with that symbol. So make it based on address and network instead of symbol
- remove the fixed icon for tokens with symbol 'icp'. Add it to the curated icons folder instead

# Tests
- User who had the token enabled
  - can still see it, including the icon and the hero text ✅
  - can delete it ✅
- User who don't have it
  - can not see or enable it anymore ✅
  - but can import it, and then it also shows properly ✅
  - can also import a different icp token ✅
    - that does not show special ✅

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -419,9 +419,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.11.0"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b35204fbdc0b3f4446b89fc1ac2cf84a8a68971995d0bf2e925ec7cd960f9cb3"
+checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 dependencies = [
  "serde",
 ]
@@ -833,7 +833,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ab67060fc6b8ef687992d439ca0fa36e7ed17e9a0b16b25b601e8757df720de"
 dependencies = [
  "data-encoding",
- "syn 2.0.114",
+ "syn 1.0.109",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


src/frontend/src/env/tokens/tokens.erc20.env.ts
```diff
@@ -57,13 +57,7 @@ export const ERC20_CONTRACTS_SEPOLIA: Erc20Contract[] = [
...
 ];
 
-export const ERC20_CONTRACT_ICP: Erc20Contract = {
-	// ICP
-	address: '0x054B8f99D15cC5B35a42a926635977d62692F25b',
-	exchange: 'icp'
-};
-
-export const ERC20_CONTRACTS_PRODUCTION: Erc20Contract[] = [ERC20_CONTRACT_ICP];
+export const ERC20_CONTRACTS_PRODUCTION: Erc20Contract[] = [];
```
AntonioVentilii, line 60: just for posterity

```suggestion
// TODO: Check if this list is/ will be needed, and remove it otherwise
export const ERC20_CONTRACTS_PRODUCTION: Erc20Contract[] = [];
```

src/frontend/src/eth/utils/token.utils.ts
```diff
@@ -12,7 +13,10 @@ import { formatToken } from '$lib/utils/format.utils';
 import { zeroPadValue } from 'ethers/utils';
 
 export const isErc20Icp = (token: OptionToken): boolean =>
-	token?.symbol === ERC20_ICP_SYMBOL && isTokenErc20(token);
+	token?.symbol === ERC20_ICP_SYMBOL &&
```
AntonioVentilii, line 16: Shall we can remove this check for the symbol? Or did you want to do it in another PR?

```suggestion
```

sbpublic: I intended to, but then the next check fails because token could be undefined and the isErc20Token can't handle that.
Didn't know how to easily resolve (I would need to check for not nullish first or so), so I kept it.

Somehow it's not wrong to keep it . Before it also checked for the symbol, so my change only makes the constraints more narrow. Removing it would change the behavior if they rename the symbol of the token to "ABC". Before it would not be the icp anymore, and without this condition, it now would keep it as icp...).
And if I remove this condition, we would have to remove the constant as well.

src/frontend/src/env/tokens/tokens.erc20.env.ts
sbpublic: I think I saw once a CI that checks something about those tokens (could not find it now anymore...)

PR #11485, opened 2026-02-12
feat(frontend): add erc4626 derived

DenysKarmazynDFINITY:
# Motivation

We can now implement all erc4626 derived stores.

Note: there is one store that is not standard comparing with other erc tokens - erc4626TokensExchangeData. It's gonna be used later for exchange rate calculation - instead of passing the whole array with tokens, we need to flatten the object since some fields (e.g. network id symbol) cannot be serialized and therefore passed to the worker.


AntonioVentilii (approved):
LGTM tks

PR #11494, opened 2026-02-13
feat(frontend): implement ERC4626 services

DenysKarmazynDFINITY:
# Motivation

We can now implement ERC4626 services.


src/frontend/src/eth/services/erc4626.services.ts
```diff
@@ -0,0 +1,238 @@
...
+	return await isInterfaceErc4626(address);
+};
+
+export const loadErc4626Tokens = async ({
+	identity
+}: {
+	identity: OptionIdentity;
+}): Promise<void> => {
+	loadDefaultErc4626Tokens();
+	await loadCustomErc4626Tokens({ identity, useCache: true });
```
AntonioVentilii, line 46: I see that we use Promise.all in the other similar services. You prefer to make them sequential here?

DenysKarmazynDFINITY: The thing is `loadDefaultErc4626Tokens` does not return a promise - it just pushes default tokens to the store. In ERC20 services, a similar method fetches contracts, therefore it makes sense to have Promise.all there.

DenysKarmazynDFINITY: I kept the "load" naming for the consistency + in case in the future it's gonna be extended (same as erc20), but let me know if you think the naming is confusing

PR #11517, opened 2026-02-13
feat(frontend): add ERC4626 tokens to token utils

DenysKarmazynDFINITY:
# Motivation

We need to add ERC4626 standard to the tokens utils.

### DIFF
--- a/src/frontend/src/lib/utils/token.utils.ts
+++ b/src/frontend/src/lib/utils/token.utils.ts
@@ -46,7 +46,11 @@ export const getMaxTransactionAmount = ({
 }): string => {
 	const value =
 		(balance ?? ZERO) -
-		(tokenStandard.code !== 'erc20' && tokenStandard.code !== 'spl' ? fee : ZERO);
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/utils/token.utils.ts
```diff
@@ -46,7 +46,11 @@ export const getMaxTransactionAmount = ({
 }): string => {
 	const value =
 		(balance ?? ZERO) -
-		(tokenStandard.code !== 'erc20' && tokenStandard.code !== 'spl' ? fee : ZERO);
+		(tokenStandard.code !== 'erc20' &&
+		tokenStandard.code !== 'erc4626' &&
+		tokenStandard.code !== 'spl'
```
AntonioVentilii, line 51: for another PR maybe: don't we have utils for this standard check?

DenysKarmazynDFINITY: good point!

PR #11533, opened 2026-02-13
feat(frontend): extend exchange worker with ERC4626 tokens

DenysKarmazynDFINITY:
# Motivation

Finally, we can extend exchange worker with ERC4626 tokens.


AntonioVentilii (approved):
LGTM tks

PR #11545, opened 2026-02-16
feat(backend)!: Calculate BTC address for method `btc_add_pending_transaction`

AntonioVentilii:
# Motivation

Suggested by prodsec's @tmu0 

For method `btc_add_pending_transaction` we don't really need to pass the BTC address that refers to the pending transactions. We can simply derive it from the principal, like we do in method `btc_select_user_utxos_fee`.

BREAKING CHANGE: The interface of the method will change to have one less param.

# Changes

-Replicate the calculation of the source address done in other methods.

# Tests

Current tests will suffice.


DenysKarmazynDFINITY (approved):
LGTM, thanks!

PR #11618, opened 2026-02-20
refactor(frontend): Flatten `Loaders` component as much as possible

AntonioVentilii:
# Motivation

It doesn't make sense that we nest a bunch of component in `Loaders` component if they are not necessarily guards or context.

### DIFF
--- a/src/frontend/src/lib/components/core/MultipleListeners.svelte
+++ b/src/frontend/src/lib/components/core/MultipleListeners.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-	import type { Snippet } from 'svelte';
 	import { authSignedIn } from '$lib/derived/auth.derived';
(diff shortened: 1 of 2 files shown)


src/frontend/src/lib/components/core/MultipleListeners.svelte
AntonioVentilii:
```suggestion
```

src/frontend/src/lib/components/loaders/LoaderTokens.svelte
AntonioVentilii:
```suggestion
```

src/frontend/src/lib/components/rewards/UserSnapshotWorker.svelte
AntonioVentilii:
```suggestion
	import { onDestroy,  } from 'svelte';
```

PR #11622, opened 2026-02-20
feat(frontend): Component to custom-sort tokens list

AntonioVentilii:
# Motivation

Copying what was done with `NftSortMenu`, we create `TokensSortMenu`, that simply set the tokens sort store according to the specific criteria we decide.


src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -129,7 +129,8 @@ export const sortTokens = <T extends Token>({
 		}
 
 		return (
-			a.symbol.localeCompare(b.symbol) ||
+			// If the choice is to prioritise symbol sorting, it will have already happened before
+			(primarySortStrategy !== 'symbol' && a.symbol.localeCompare(b.symbol)) ||
```
AntonioVentilii:
```suggestion
			( a.symbol.localeCompare(b.symbol)) ||
```

PR #11638, opened 2026-02-23
refactor(frontend): Prepare util `sortTokens` to accept tokens groups

AntonioVentilii:
# Motivation

This PR refactors the token sorting logic to prepare the support both:

* `TokenUi<T>[]`
* `TokenUiOrGroupUi[]`

using a single, shared sorting implementation.

The main change is the introduction of a normalisation step (`unwrapTokenSortFields`) combined with a **decorate → sort → undecorate** pattern to ensure consistent and predictable sorting across both tokens and groups.

Previously, sorting operated directly on `TokenUi<T>` and relied on token-specific fields (e.g. `network.name`, `id`, etc.).

To support groups without duplicating comparator logic or introducing branching throughout the comparator, we:

1. Introduce a normalised sort shape (`TokenSortUnwrapped`).
2. Convert both tokens and groups into that shape before sorting.
3. Run a single comparator over the normalised fields.

This removes the need for token/group-specific logic inside the comparator and keeps the sorting rules defined in one place.

# Note

## Why the “double map” is intentional

The implementation uses:

```ts
$tokens
  .map(token => ({ token, u: unwrapTokenSortFields(item) }))
  .sort((a, b) => comparator(a.u, b.u))
  .map(({ token }) => token);
```

This is deliberate:

* **Avoid repeated normalisation per comparison**
  Sorting performs `O(n log n)` comparisons.
  By precomputing the sort fields once per element, we ensure:

  * `unwrapTokenSortFields` runs exactly **once per item**
  * The comparator only compares primitive fields

* **Predictable performance**
  The list size is small (~100–150 elements), and sorting runs approximately every 30 seconds (derived store refresh).
  The additional linear passes and temporary allocations are negligible at this scale.

* **Simpler comparator**
  The comparator operates only on normalised primitives, without:

  * type branching (token vs group),
  * repeated field extraction,
  * or cache/memoisation logic.

This keeps the comparator easier to reason about and safer to extend.

### 1. Lazy normalisation with `WeakMap`

This would avoid the upfront map but introduce:

* additional cache lookups per comparison,
* more complex comparator logic,
* less predictable execution flow.

Given the small dataset and low frequency, this would be unnecessary complexity.

### 2. Duplicate comparators for tokens and groups

Rejected to avoid:

* logic duplication,
* divergence risk,
* maintenance overhead.

# Changes

- Create small util to "normalise" the items to be sorted: either token UI or tokens group.
- Wrap the sorting logic with decorators: we need to normalise and unnormalise the lists.

# Tests

Current tests should suffice.

### DIFF
--- a/src/frontend/src/lib/components/exchange/ExchangeRateChange.svelte
+++ b/src/frontend/src/lib/components/exchange/ExchangeRateChange.svelte
@@ -8,17 +8,12 @@
 
 	interface Props {
 		usdPriceChangePercentage24h: number | undefined;
-		withBackground?: boolean;
+		background?: 'light' | 'dark';
 		timeFrame?: '24h';
 		fontSize?: 'sm' | 'xs';
 	}
 
-	let {
-		usdPriceChangePercentage24h,
-		withBackground = false,
-		timeFrame,
-		fontSize = 'sm'
-	}: Props = $props();
+	let { usdPriceChangePercentage24h, background, timeFrame, fontSize = 'sm' }: Props = $props();
 
 	let parsedExchangeRateChange = $derived(
 		nonNullish(usdPriceChangePercentage24h)
@@ -49,9 +44,10 @@
 {#if nonNullish(parsedExchangeRateChange)}
 	<span
 		class="px-1 text-xs"
-		class:bg-error-subtle-30={withBackground && exchangeRateChangeSign === 'negative'}
-		class:bg-success-subtle-30={withBackground && exchangeRateChangeSign === 'positive'}
-		class:rounded={withBackground}
+		class:bg-opacity-30={background === 'dark'}
+		class:bg-primary={background === 'light'}
+		class:bg-primary-inverted={background === 'dark'}
+		class:rounded={nonNullish(background)}
 		class:sm:text-sm={fontSize === 'sm'}
 		class:text-error-primary={exchangeRateChangeSign === 'negative'}
 		class:text-success-primary={exchangeRateChangeSign === 'positive'}
--- a/src/frontend/src/lib/components/hero/HeroContent.svelte
+++ b/src/frontend/src/lib/components/hero/HeroContent.svelte
@@ -122,6 +122,17 @@
 	let isGradientToRight = $derived($networkSolana && !isTrumpToken);
 
 	let isGradientToBottomRight = $derived(isGLDTToken || $networkBsc);
+
+	let rateChangeBackground = $derived(
+		$networkICP ||
+			$networkBase ||
+			$networkPolygon ||
+			$networkArbitrum ||
+			isTrumpToken ||
+			isVeurToken
+			? ('dark' as const)
+			: ('light' as const)
+	);
 </script>
 
 <div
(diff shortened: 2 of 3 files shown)


src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -23,34 +23,43 @@ import type { Token, TokenId, TokenToPin } from '$lib/types/token';
...
+		usdBalance: item.usdBalance,
+		usdPriceChangePercentage24h: item.usdPriceChangePercentage24h,
+		usdMarketCap: item.usdMarketCap,
+		balance: item.balance
+	};
+};
+
+type TokenSortUnwrapped = ReturnType<typeof unwrapTokenSortFields>;
+
 const createTokenComparator =
```
AntonioVentilii, line 79:
```suggestion
/**
 * Creates a comparator function for sorting tokens based on multiple criteria:
 *
 * 1. Deprecation status (non-deprecated tokens first).
 * 2. Primary sorting strategy (either performance or symbol, or value by default, based on the provided parameter).
 * 3. USD balance (descending).
 * 4. Explicitly pinned tokens (pinned first, preserving the order provided by `pinIndexById`).
 * 5. Token symbol (ascending, locale-aware).
 * 6. Token name (ascending, locale-aware).
 * 7. Network name (ascending, locale-aware).
 * 8. Token balance (descending).
 * 9. USD market cap (descending).
 *
 * The `primarySortStrategy` parameter allows overriding the default sorting by value with either performance or symbol prioritisation.
 *
 */
const createTokenComparator =
```

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -138,16 +154,15 @@ const createTokenComparator =
...
- * 7. Network name (ascending, locale-aware).
- * 8. Token balance (descending).
- * 9. USD market cap (descending).
+ * 2. USD balance (descending).
+ * 3. Explicitly pinned tokens (pinned first, preserving the order provided in `$tokensToPin`).
+ * 4. Token symbol (ascending, locale-aware).
+ * 5. Token name (ascending, locale-aware).
+ * 6. Network name (ascending, locale-aware).
+ * 7. Token balance (descending).
+ * 8. USD market cap (descending).
```
AntonioVentilii:
```suggestion
 * 2. Primary sorting strategy (either performance or symbol, or value by default, based on the provided parameter).
 * 3. USD balance (descending).
 * 4. Explicitly pinned tokens (pinned first, preserving the order provided in `$tokensToPin`).
 * 5. Token symbol (ascending, locale-aware).
 * 6. Token name (ascending, locale-aware).
 * 7. Network name (ascending, locale-aware).
 * 8. Token balance (descending).
 * 9. USD market cap (descending).
```

src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -138,16 +154,15 @@ const createTokenComparator =
...
  *
- * The `primarySortStrategy` parameter allows overriding the default sorting by value with either performance or symbol prioritisation.
+ * Additionally, if `primarySortStrategy` is set, it overrides the default sorting by value.
```
AntonioVentilii:
```suggestion
 * The `primarySortStrategy` parameter allows overriding the default sorting by value with either performance or symbol prioritisation.
```

DenysKarmazynDFINITY (approved):

PR #11663, opened 2026-02-24
feat(backend)!: Verify II delegation for method `btc_add_pending_transaction`

AntonioVentilii:
# Motivation

The `btc_add_pending_transaction` endpoint can currently be called by anyone with a valid principal, including CLI tools or scripts that are not authenticated through Internet Identity. This opens a potential attack surface where an unauthorised caller could reserve UTXOs for pending transactions without going through the intended authentication flow.

This PR adds a server-side verification step that validates the caller provided a legitimate II delegation chain, ensuring that only users who authenticated through Internet Identity (or canister controllers) can add pending Bitcoin transactions.

BREAKING CHANGE: II-delegation verification for method `btc_add_pending_transaction`, with related parameters and types.

# Changes

- **New `delegation` module** (`src/backend/src/delegation.rs`): Implements `verify_ii_delegation_chain`, which validates an II delegation chain against the caller's principal. The verification checks:
  - Exactly one delegation in the chain (II invariant)
  - No delegation targets set
  - The delegation's public key derives to the caller's principal
  - The signing canister is a known II canister (from `supported_credentials` config)
  - The delegation has not expired
  - The canister signature is valid against the IC root key
- **II delegation verification in `btc_add_pending_transaction`**: The endpoint now requires callers to include an `IIDelegationChain` in the request. The chain is verified before proceeding with the transaction logic.
- **Controller bypass**: Canister controllers skip the delegation check entirely, since they are already trusted and may need to operate without an II session.
- **New dependencies**: Added `ic-signature-verification` and `ic-canister-sig-creation` crates for canister signature verification.
- **Test infrastructure**: 
  - New `IICanister` test helper (`tests/it/utils/ii.rs`) that deploys a real Internet Identity canister on PocketIC, registers identities, and obtains delegation chains for end-to-end testing.
  - New `setup_with_ii()` function that creates a PocketIC environment with NNS + II + fiduciary subnets and configures the backend with the PocketIC root key.
  - Updated `test.backend.sh` to download the II WASM for integration tests.

# Tests

Added new tests and existing rate-limit tests remain passing: the rate limiter runs before the delegation check, so non-controller callers hit `InvalidDelegationChain` instead of succeeding, but the rate-limit assertions still hold.

### DIFF
--- a/src/declarations/backend/backend.did.d.ts
+++ b/src/declarations/backend/backend.did.d.ts
@@ -523,6 +523,7 @@ export interface _SERVICE {
 	 *
 	 * Requires a valid II delegation chain to verify the caller authenticated
 	 * through Internet Identity. This protects against unauthorised CLI callers.
+	 * Controllers bypass this check.
 	 *
 	 * # Errors
 	 * Errors are enumerated by: `BtcAddPendingTransactionError`.


src/backend/src/delegation.rs
```diff
@@ -0,0 +1,62 @@
...
+///
+/// Checks performed:
+/// 1. The delegation chain is non-empty.
+/// 2. The self-authenticating principal derived from the chain's root public key matches the
+///    caller.
+/// 3. None of the delegations in the chain have expired.
+/// 4. The root public key is a valid IC canister signature key (proving the caller authenticated
+///    through a canister, not a raw key pair).
+/// 5. The canister ID embedded in the public key matches a known II canister ID.
+/// 6. If targets are specified in any delegation, the current canister is included.
```
bitdivine: Do you want to check the signature itself here or somewhere else?

AntonioVentilii: correct, included it here! thanks for that!

src/backend/src/delegation.rs
```diff
@@ -0,0 +1,97 @@
...
+    if !known_ii_canister_ids.contains(&canister_sig_pk.canister_id) {
+        return Err(format!(
+            "delegation root key belongs to canister {} which is not a known II canister",
+            canister_sig_pk.canister_id
+        ));
+    }
+
+    let mut signing_key = chain.public_key.clone();
+
+    for signed_del in &chain.delegations {
```
tmu0: number of delegations is unbounded, II delegation chains should always have only 1 delegation and no targets set

src/backend/src/delegation.rs
```diff
@@ -0,0 +1,97 @@
...
+///    caller.
+/// 3. None of the delegations in the chain have expired.
+/// 4. The root public key is a valid IC canister signature key (proving the caller authenticated
+///    through a canister, not a raw key pair).
+/// 5. The canister ID embedded in the public key matches a known II canister ID.
+/// 6. If targets are specified in any delegation, the current canister is included.
+/// 7. Each delegation signature is cryptographically valid. The first delegation is verified as an
+///    IC canister signature against the IC root key; subsequent delegations are verified against
+///    the session key from the preceding delegation.
+pub fn verify_ii_delegation_chain(
```
tmu0, line 16: no need to implement this on your own, we have a library for this: 

- https://github.com/dfinity/internet-identity/blob/3380aca8df9bb9552002222c8788b7c47c9e6008/src/sig-verifier-js/src/lib.rs#L42
- usage: https://github.com/dfinity/control-panel/blob/9c3681e35286fecff3a2ad37568aff8cc75dfa96/src/backend/src/services/auth.rs#L171

AntonioVentilii: we already uses the same underlying crates (`ic-canister-sig-creation` and `ic-signature-verification`), so I can use them directly, WDYT? Same implementation

src/backend/src/delegation.rs
AntonioVentilii: done

src/backend/src/delegation.rs
tmu0: I would always recommend using libraries for security critical code instead of duplicating code. When later it turns out, that it has a vulnerability you just need to patch the library and bump versions everywhere instead of remembering to which projects you have copied the code to.

But for this instance it might be okay for now, because I've heard that the II team is working on a new library and the old one will be deprecated anyway.

PR #11676, opened 2026-02-25
feat(frontend): Sort networks by pinned list first

AntonioVentilii:
# Motivation

Currently, everything else being the same, two tokens are sorted at some point by network name. However, before that, we would like to have consistency and sort by the predefined order that we currently have in the network dropdown choice.

That means that we need to apply for networks a logic similar to the ones of pinned tokens.

# Changes

- Create util for network comparator.
- Add new comparator before the network name comparator.
- Adapt util `sortTokens` to accept a list of pinned network and create indexed list.
- Adapted consumers.

# Tests

Adapted tests.

### DIFF
--- a/src/frontend/src/lib/components/rewards/UserSnapshotWorker.svelte
+++ b/src/frontend/src/lib/components/rewards/UserSnapshotWorker.svelte
@@ -86,9 +86,6 @@
 		if (
 			$authNotSignedIn ||
 			$isBusy ||
-			isNullish($btcAddressMainnet) ||
-			isNullish($ethAddress) ||
-			isNullish($solAddressMainnet) ||
 			isNullish($tokens) ||
 			$exchangeNotInitialized ||
 			$noPositiveBalanceAndNotAllBalancesZero


src/frontend/src/lib/utils/tokens.utils.ts
```diff
@@ -67,8 +68,33 @@ type SortableTokenId = TokenId;
...
 type SortableItem<T extends Token> = TokenUi<T> | TokenUiOrGroupUi;
 
+const createNetworkComparator =
+	({ networkPinIndexById }: { networkPinIndexById: ReadonlyMap<SortableNetworkId, number> }) =>
+	// eslint-disable-next-line local-rules/prefer-object-params -- This is a sort function.
+	(a: SortableNetworkId, b: SortableNetworkId): number => {
+		const aPin = networkPinIndexById.get(a);
+		const bPin = networkPinIndexById.get(b);
+
+		const aPinnedNet = aPin !== undefined;
```
DenysKarmazynDFINITY: isNullish? or you're only interested in handling `undefined`?

AntonioVentilii: good catch! fixed

PR #11682, opened 2026-02-25
fix(frontend): Bad scaling for WalletConnect proposal modal

AntonioVentilii:
# Motivation

The WalletConnect Session Proposal modal overflows on Android phones in full-page mode.

The approve/reject buttons and namespace content are not properly scrollable. The issue does not reproduce on iPhones or on the Connect (QR code) step, which has compact content that fits without scrolling.

On iOS, the `@supports (-webkit-touch-callout: none)` block gives the dialog an explicit height via `-webkit-fill-available`, which allows the nested flex layout chain to properly propagate the height constraint and enable scrolling in `.content`.

On Android, that `@supports` block never fires (it's iOS/WebKit-only), so `--dialog-height` remained `initial` (auto). While `min-height` and `max-height` are both `100dvh`, relying on `height: auto` with min/max constraints doesn't reliably propagate the height through the deeply nested flex layout on Android Chrome, causing content to overflow instead of scrolling.

# Changes

- Set `--dialog-height: 100dvh` as the default for mobile dialogs instead of `initial` (auto).
- Added `--dialog-height: initial` in the desktop media query to preserve the existing auto-sizing behavior for non-mobile dialogs.

# Tests

### Before

image-20260224-151106


AntonioVentilii:
Superseded by:

- https://github.com/dfinity/oisy-wallet/pull/11416

PR #11714, opened 2026-02-27
feat(frontend): POC: loop imports

AntonioVentilii:
# Motivation

# Changes

# Tests


AntonioVentilii:
this was just a test to create a script to check loop imports. We will rely on existing libraries for it

PR #11772, opened 2026-03-03
feat(frontend): extend harvest store

DenysKarmazynDFINITY:
# Motivation

We need to update the harvest store with additional data, as well as to change the way we were verifying if provided address is actually an autopilot vault.

### DIFF
--- a/src/frontend/src/lib/stores/harvest.store.ts
+++ b/src/frontend/src/lib/stores/harvest.store.ts
@@ -1,11 +1,14 @@
-import { ERC4626_TOKENS } from '$env/tokens/tokens.erc4626.env';
+import { HARVEST_AUTOPILOT_ADDRESSES } from '$eth/constants/harvest-autopilots.constants';
 import type { Erc4626ContractAddress } from '$eth/types/erc4626';
 import type { HarvestVault } from '$lib/types/harvest';
 import { formatStakeApyNumber } from '$lib/utils/format.utils';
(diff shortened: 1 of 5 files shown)


src/frontend/src/lib/stores/harvest.store.ts
```diff
@@ -16,18 +19,16 @@ const initHarvestVaultsStore = (): HarvestVaultsStore => {
...
 			const data = vaults.reduce<HarvestVaultsStoreData>((acc, vault) => {
-				if (
-					knownVaultAddresses.has(vault.vaultAddress.toLowerCase()) &&
-					nonNullish(vault.estimatedApy)
-				) {
-					acc[vault.vaultAddress] = {
-						estimatedApy: formatStakeApyNumber(Number(vault.estimatedApy))
+				const address = vault.vaultAddress.toLowerCase();
+
+				if (HARVEST_AUTOPILOT_ADDRESSES.includes(address) && nonNullish(vault.estimatedApy)) {
```
AntonioVentilii, line 28: nitpick: i know that may it is too verbose, but we have the util `areAddressesEqual` that compensate the case-sensitiveness if you specify the network.

DenysKarmazynDFINITY: it's a good point, my only concern is I'll need to loop through `HARVEST_AUTOPILOT_ADDRESSES` in order to use `areAddressesEqual` - `.includes` feels a bit shorter

PR #11804, opened 2026-03-04
feat(frontend): update earning stores

DenysKarmazynDFINITY:
# Motivation

The next step is to update `earning.derived` with Harvest autopilot data. At the same time, the GLDT stuff can be removed from there.


AntonioVentilii (approved):
LGTM tks!

PR #11805, opened 2026-03-04
feat(frontend): implement LoadersHarvestAutopilotBalances component

DenysKarmazynDFINITY:
# Motivation

When entering the Harvest page, we need to show all autopilots, even if they are disabled. To do that, we need to load balances for all disabled vaults.

### DIFF
--- a/src/frontend/src/eth/components/loaders/LoadersHarvestAutopilotBalances.svelte
+++ b/src/frontend/src/eth/components/loaders/LoadersHarvestAutopilotBalances.svelte
@@ -56,11 +56,11 @@
 	});
 
 	onMount(async () => {
-		const principal = $authIdentity?.getPrincipal();
-
-		if (isNullish(principal)) {
+		if (isNullish($authIdentity)) {
 			return;
 		}
+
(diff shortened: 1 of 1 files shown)


src/frontend/src/eth/components/loaders/LoadersHarvestAutopilotBalances.svelte
```diff
@@ -0,0 +1,81 @@
...
+
+		debounceLoad();
+	});
+
+	onMount(async () => {
+		const principal = $authIdentity?.getPrincipal();
+
+		if (isNullish(principal)) {
+			return;
+		}
```
AntonioVentilii: theoretically here you could first check for $authIdentity, right? but same same

AntonioVentilii (approved):

src/frontend/src/eth/components/loaders/LoadersHarvestAutopilotBalances.svelte
DenysKarmazynDFINITY: yes, better! updated

PR #11842, opened 2026-03-05
feat(frontend): Util to calculate IC fee for OpenCryptoPay

AntonioVentilii:
# Motivation

As first step to integrate ICP in the OpenCryptoPay flow, we calculate the fee to return a structured response to the pay service.

NOTE: this util can be applied both to ICP and to ICRC tokens.

### DIFF
--- a/src/frontend/src/icp/services/ic-open-crypto-pay.services.ts
+++ b/src/frontend/src/icp/services/ic-open-crypto-pay.services.ts
@@ -1,0 +1,21 @@
+import type { IcFeeResult } from '$icp/types/pay';
+import { getTokenFee } from '$icp/utils/token.utils';
+import type { PayableToken } from '$lib/types/open-crypto-pay';
+import { isNullish } from '@dfinity/utils';
+
+/**
+ * Calculates the fee for an ICP/ICRC token payment.
+ * The total fee covers two transactions: one for `icrc2_approve` and one for `icrc2_transfer_from`.
+ */
+export const calculateIcFee = (token: PayableToken): IcFeeResult | undefined => {
+	const fee = getTokenFee(token);
+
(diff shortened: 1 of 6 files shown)


src/frontend/src/icp/services/ic-open-crypto-pay.services.ts
```diff
@@ -0,0 +1,21 @@
...
+export const calculateIcFee = (token: PayableToken): IcFeeResult | undefined => {
+	const fee = getTokenFee(token);
+
+	if (isNullish(fee)) {
+		return;
+	}
+
+	return {
+		feePerTransaction: fee,
+		totalFee: fee * 2n
```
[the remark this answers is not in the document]
DenysKarmazynDFINITY: irrelevant in this case

src/frontend/src/icp/services/ic-open-crypto-pay.services.ts
```diff
@@ -0,0 +1,21 @@
...
+import { isNullish } from '@dfinity/utils';
+
+/**
+ * Calculates the fee for an ICP/ICRC token payment.
+ * The total fee covers two transactions: one for `icrc2_approve` and one for `icrc2_transfer_from`.
+ */
+export const calculateIcFee = (token: PayableToken): IcFeeResult | undefined => {
+	const fee = getTokenFee(token);
+
+	if (isNullish(fee)) {
```
DenysKarmazynDFINITY, line 13: unlikely to happen, but we can fallback to OCP `minFee` if `fee` is nullish

AntonioVentilii: As discussed offline, we will proceed like the other networks and ignore the token for now, in the remote case of missing fee

PR #12152, opened 2026-03-25
docs(core): Proposal for architecture split

AntonioVentilii:
# Motivation

# Changes

# Tests


AntonioVentilii:
we are not going to split now

PR #12175, opened 2026-03-26
perf(frontend): Store batching and equality guards

AntonioVentilii:
# Motivation

Split from https://github.com/dfinity/oisy-wallet/pull/11895


AntonioVentilii:
Not needed for now

PR #12370, opened 2026-04-10
feat(frontend): enable to install oisy as PWA

sbpublic:
# Motivation

We once disabled PWA because the signer standard does not work from PWA to PWA, see #9103.
Since we are working on a solution for that, we decided to allow it again. Also since we only show the warning banner in the signer UI and if it's really in standalone mode, the pros are much higher than ten the cons.

# Changes

- enable standalone mode, but keep the warning banner. We changed that to only show when used as signer.
- change the banner text to match the new 'scenario'

# Tests
OISY Wallet 3

IMG_5361

IMG_5362


DenysKarmazynDFINITY (approved):
Nice!

PR #12419, opened 2026-04-14
feat(frontend): update QrCodeReader component

DenysKarmazynDFINITY:
# Motivation

We start the Universal scanner feature with updating the QrCodeReader component:
1. Update design.
2. Switch from CSS to Tailwind.

Screenshot 2026-04-14 at 15 41 20

### DIFF
--- a/src/frontend/src/lib/components/ui/QrCodeReader.svelte
+++ b/src/frontend/src/lib/components/ui/QrCodeReader.svelte
@@ -4,6 +4,12 @@
 	import { isDesktop } from '$lib/utils/device.utils';
 	import { nextElementId } from '$lib/utils/html.utils';
 
+	interface Props {
+		expandedLayout?: boolean;
+	}
+
+	let { expandedLayout = false }: Props = $props();
+
 	const id = nextElementId('qrcode-reader-');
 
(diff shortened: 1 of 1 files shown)


src/frontend/src/lib/components/ui/QrCodeReader.svelte
```diff
@@ -110,51 +116,29 @@
...
+<article {id} class="relative h-full w-full overflow-hidden {mirror ? '-scale-x-100' : ''}">
+	<video
+		bind:this={videoElement}
+		class="block h-full w-full object-cover"
+		autoplay
+		muted
+		playsinline
+	></video>
+
+	<div
```
AntonioVentilii, line 132: Just to be sure, we had an issue in the past with borderless dark-themed inverted QR codes. Do they still work with the new schema?

AntonioVentilii (approved):
LGTM tks! Small doubt:

src/frontend/src/lib/components/ui/QrCodeReader.svelte
DenysKarmazynDFINITY: logic-wise, everything should stay the same, but worth double-checking, thanks!

PR #12425, opened 2026-04-15
feat(frontend): new ScannerInfo component

DenysKarmazynDFINITY:
# Motivation

We need to implement a new component with the info on what is "OISY Scan".

Screenshot 2026-04-15 at 09 27 54


src/frontend/src/lib/components/scanner/ScannerInfo.svelte
```diff
@@ -0,0 +1,67 @@
...
+	import { OISY_SCAN_URL, OISY_PAY_URL } from '$lib/constants/oisy.constants';
+	import {
+		OISY_SCANNER_INFO,
+		OISY_SCANNER_INFO_GOT_IT_BUTTON
+	} from '$lib/constants/test-ids.constants';
+	import { i18n } from '$lib/stores/i18n.store';
+	import { replaceOisyPlaceholders } from '$lib/utils/i18n.utils';
+
+	interface Props {
+		onButtonClick: () => void;
```
AntonioVentilii, line 17: not called `onClick` just to be more specific?

AntonioVentilii (approved):
LGTM tks

src/frontend/src/lib/components/scanner/ScannerInfo.svelte
DenysKarmazynDFINITY: exactly - since there are a few action points in this modal, it's good to keep this prop a bit more specific

PR #12435, opened 2026-04-15
feat(frontend): move Pay/Scan buttons from Header to Menu

DenysKarmazynDFINITY:
# Motivation

We now need to move the Pay and Scanner buttons from Header to Menu. In the header, only the Universal Scanner and connected WC buttons will be displayed.


AntonioVentilii (approved):
NICE tks!

PR #12454, opened 2026-04-16
chore(frontend): Remove cbBTC icrc token

sbpublic:
# Motivation

The 1sec team stated
> _Regarding cbBTC we stopped showing it in our UI as we would rather push for ckBTC._

# Changes

- remove the curated cbBTC on ICP and its icon that were added in #12420 

# Tests
image


AntonioVentilii (approved):
Pity, but tks

PR #12463, opened 2026-04-16
feat(frontend): add back button to NFT detail and collection pages

sbpublic:
# Motivation

The NFT detail and collection pages had no dedicated navigation controls — users were forced to use the breadcrumbs as the only way to go back. This added two complementary navigation elements to both pages: a close (X) button overlaid on the hero image, and an up arrow next to the breadcrumbs, giving users faster and more intuitive ways to move up in the page hierarchy.

# Changes

- **Close (X) button on NFT hero images**: Added a close button (X icon) overlaid on the hero image of both the NFT collection and NFT detail pages, styled with a semi-transparent dark background and hover state
- **Back (arrow) button in breadcrumbs**: Added a back arrow button next to the breadcrumb navigation on both pages that navigates one level up in the hierarchy
- **NFT collection page close behavior**: The X button always navigates directly to the assets/NFT list page
- **NFT detail page close behavior**: The X button uses browser history (`history.back()`) to return to the previous page, falling back to the root route if there is no history
- **Test selector update**: Updated `NftHero` test to use a more specific selector (`.h-64 button.block`) to target the image button, avoiding ambiguity with the new close button
- **Accessibility**: Added i18n keys (`core.alt.up_one_level`) for the up arrow button aria-label across all 15 supported languages

## Spec

**(X) Button**
- NFT Collection Page: go to assets
- NFT Page: goes back in history (so either collection or assets)

**(<) Button**
- NFT Collection Page: go to assets 
- NFT Page: go to collection

# Tests

- **NftCollectionHero**: X button navigates to the assets page; back arrow navigates one level up
- **NftHero**: X button calls `back({ pop: true })` when there is a previous route and `back({ pop: false })` when there is none; back arrow navigates to the parent collection URL

image

image


AntonioVentilii (approved):
LGTM tks

PR #12480, opened 2026-04-17
feat(frontend): add ICRC CHAT token to CHAT token group

sbpublic:
# Motivation

The ICRC CHAT token (OpenChat SNS, ledger canister `2ouva-viaaa-aaaaq-aaamq-cai`) was not part of the CHAT token group, causing it to appear as a standalone entry separate from the Ethereum, Base, and Arbitrum CHAT tokens.

image

# Changes

- Added `groupDataId: "CHAT"` to the ICRC CHAT entry in `tokens.sns.json` so the existing `mapEnvGroupData` mechanism assigns it to the CHAT group at runtime

Note: The group is currently not picking up the price from the ICP token, and price and performance remain empty because the first 3 tokens don't have a price. This will be fixed in a separate PR.

# Tests
image


sbpublic:
What a large PR history for a single change ... 🫣

DenysKarmazynDFINITY (approved):

